PR #2, opened 2013-02-22
added rule to check for selector arguments in sonar

lkwg82:
Hi folks,

this is the pull request for http://jira.codehaus.org/browse/SONARJAVA-106

cheers Lars


benzonico:
Thanks for your contribution. This PR has not been applied and can no longer be applied since the way to write checks has changed (we are now using a strongly typed Syntax Tree and rely on a totally different class). 

I scheduled http://jira.codehaus.org/browse/SONARJAVA-106 to 3.0 and linked it to your implementation.

As a consequence of all this I close this pull request.

PR #8, opened 2014-01-22
Add immutable option on S00115

diorcety:
Provide a way to apply the rule S00115 only on immutable type.
Some people don't want static final mutable variable to be considered as constant.


benzonico:
Thanks for your contribution. 

This issue has been handled by this ticket : http://jira.codehaus.org/browse/SONARJAVA-432 with a different approach.
Therefore I am closing this pull request.

PR #13, opened 2014-07-01
fixed EmptyStackException when trying to check interface with default me...

steve-ulrich-proemion:
...thods and anonymous inner classes


benzonico:
Thanks for your contribution,

This check has been migrated to the strongly typed AST and therefore your fix is no longer relevant so I close this PR. 
Still, there was a lack of test on default method, fixed in commit: https://github.com/SonarSource/sonar-java/commit/bc45a1dff1709bbc3b6039435d3f40598f357d66

Thanks.

PR #16, opened 2014-08-07
Add exceptions for Lombok annotations to UnusedPrivateFieldCheck

nathansgreen:
(S1068). Class- and field-level annotations are checked. [Lombok](http://projectlombok.org/) generates methods and constructors from annotations, eliminating boilerplate code. Source-level analysis should take into account Lombok semantics to prevent false positives.

Test case fails for reasons I don't understand. If I drop the "lombok." prefix on the annotation names, I can make the test pass. Why doesn't the annotation's `IdentifierTree.name()` return a fully-qualified annotation name?

An alternative implementation would be to sublass this check and allow Lombok support to be turned on/off on a per-project basis.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/UnusedPrivateFieldCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/UnusedPrivateFieldCheck.java
@@ -19,6 +19,8 @@
  */
 package org.sonar.java.checks;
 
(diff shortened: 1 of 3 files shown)


benzonico:
Ticket http://jira.codehaus.org/browse/SONARJAVA-73 has been fixed. In consequence I close this PR.

PR #19, opened 2015-01-08
Upgrade dependencies

hansjoachim:
Just upgraded a couple of dependencies to the latest version.

(I tried to upgrade junit version as well, but it broke a bunch of test which presumably required/used the older version of hamcrest)


benzonico:
Hi thanks for your contribution, but there are no need right now to upgrade those dependencies. So I will just close this PR.

PR #22, opened 2015-01-21
Sonarjava 863 rspec 2257 only standard cryptographic algorithms

m-g-sonar:
Added check allowing to control that only standard cryptographic algorithms should be used (RSPEC-2257).
- Created check and unit tests
- Updated sqale model


m-g-sonar:
Canceling pull request. Rule updated.

PR #23, opened 2015-01-21
SONARJAVA-863 RSPEC-2257 Custom cryptographic algorithm check

m-g-sonar:
Custom cryptographic algorithm check:

- Controls use of a non-standard algorithm, as determined attacker may be able to break the algorithm and compromise whatever data has been protected. Standard algorithms like `SHA-256`, `SHA-384`, `SHA-512`, ... should be used instead.
- Tracks any sub-types of `java.security.MessageDigest`.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CustomCryptographicAlgorithmCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CustomCryptographicAlgorithmCheck.java
@@ -1,0 +1,65 @@
+/*
+ * SonarQube Java
+ * Copyright (C) 2012 SonarSource
+ * dev@sonar.codehaus.org
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package org.sonar.java.checks;
+
+import com.google.common.collect.ImmutableList;
+import org.sonar.check.Priority;
+import org.sonar.check.Rule;
+import org.sonar.java.model.declaration.ClassTreeImpl;
+import org.sonar.java.resolve.Symbol.TypeSymbol;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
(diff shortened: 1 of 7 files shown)


java-checks/src/main/java/org/sonar/java/checks/CustomCryptographicAlgorithmCheck.java
```diff
@@ -0,0 +1,65 @@
...
+  private boolean isJavaSecurityMessageDigestSubClass(ClassTree tree) {
+    String qualifiedName = "java.security.MessageDigest";
+    ClassTreeImpl classTreeImpl = (ClassTreeImpl) tree;
+    TypeSymbol classSymbol = classTreeImpl.getSymbol();
+    /*
+     * Corner cases:
+     * - getSymbol is nullable
+     * - A type is a subtype of itself
+     */
+    return classSymbol != null && !classSymbol.getType().is(qualifiedName) && classSymbol.getType().isSubtypeOf(qualifiedName);
```
benzonico: The comment here should be a // comment not a /* and the comment about getSymbol is nullable is unecessary.

java-checks/src/test/files/checks/CustomCryptographicAlgorithmCheck.java
```diff
@@ -0,0 +1,59 @@
+import java.security.MessageDigest;
+import java.util.Observable;
+
+class A { // Compliant
+}
+
+class B extends MessageDigest { // Noncompliant
```
benzonico: You can simplify your test cases by declaring B and C classes abstract (avoid the unecessary implementation of methods in the example)

java-checks/src/main/java/org/sonar/java/checks/CustomCryptographicAlgorithmCheck.java
```diff
@@ -0,0 +1,65 @@
...
+    if (hasSemantic()) {
+      ClassTree classTree = (ClassTree) tree;
+      if (isJavaSecurityMessageDigestSubClass(classTree)) {
+        addIssue(classTree, "Use a standard algorithm instead of creating a custom one.");
+      }
+    }
+  }
+
+  private boolean isJavaSecurityMessageDigestSubClass(ClassTree tree) {
+    String qualifiedName = "java.security.MessageDigest";
```
benzonico: should be a a constant.

java-checks/src/main/java/org/sonar/java/checks/CustomCryptographicAlgorithmCheck.java
```diff
@@ -0,0 +1,65 @@
...
+  public void visitNode(Tree tree) {
+    if (hasSemantic()) {
+      ClassTree classTree = (ClassTree) tree;
+      if (isJavaSecurityMessageDigestSubClass(classTree)) {
+        addIssue(classTree, "Use a standard algorithm instead of creating a custom one.");
+      }
+    }
+  }
+
+  private boolean isJavaSecurityMessageDigestSubClass(ClassTree tree) {
```
benzonico: Parameter of method could be ClassTreeImpl directly, this would allow only one casting in the whole class (when calling the method) and no need for a classTree variable.

benzonico:
see comments on file.

benzonico:
Manually merged. Well done.

PR #24, opened 2015-01-21
SONARJAVA-850 Generics methods read from bytecode.

benzonico:
Some checks have been updated to handle this new semantic information.


pynicolas:
LGTM

PR #26, opened 2015-01-22
SONARJAVA-862 RSPEC-2386 Interface static mutable member check

m-g-sonar:
Interface static mutable member check :

- There is no good reason to have a mutable object as the `public static` member of an `interface`. Such variables should be moved into classes and their visibility lowered.
- This rule checks that interfaces do not have `public static` array, `Collection`, or `Date` members.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/InterfaceStaticMutableMemberCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/InterfaceStaticMutableMemberCheck.java
@@ -31,6 +31,7 @@ import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.Tree.Kind;
 import org.sonar.plugins.java.api.tree.VariableTree;
 
+import java.text.MessageFormat;
 import java.util.List;
 
 @Rule(
@@ -40,9 +41,7 @@ import java.util.List;
 @BelongsToProfile(title = "Sonar way", priority = Priority.CRITICAL)
 public class InterfaceStaticMutableMemberCheck extends SubscriptionBaseVisitor {
 
-  private static final String MESSAGE = "Move \"%s\" to a class and lower its visibility";
-  private static final String COLLECTION_QUALIFIED_NAME = "java.util.Collection";
-  private static final String DATE_QUALIFIED_NAME = "java.util.Date";
+  private static final String MESSAGE = "Move \"{0}\" to a class and lower its visibility";
 
   @Override
   public List<Kind> nodesToVisit() {
@@ -55,7 +54,7 @@ public class InterfaceStaticMutableMemberCheck extends SubscriptionBaseVisitor {
       if (member.is(Kind.VARIABLE)) {
         VariableTreeImpl variableTree = (VariableTreeImpl) member;
         if (isStaticMember(variableTree) && isMutableMember(variableTree)) {
-          addIssue(variableTree, String.format(MESSAGE, variableTree.simpleName().name()));
+          addIssue(variableTree, MessageFormat.format(MESSAGE, variableTree.simpleName().name()));
         }
       }
     }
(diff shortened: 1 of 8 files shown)


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S2386.html
```diff
@@ -0,0 +1,12 @@
+<p>There is no good reason to have a mutable object as the <code>public static</code> member of an <code>interface</code>. Such variables should be moved into classes and their visibility lowered.</p>
+<p>This rule checks that interfaces do not have <code>public static</code> array, <code>Collection</code>, or <code>Date</code> members</p>
+
+<h2>Noncompliant code example</h2>
+
+<pre>
+public interface MyInterface {
+
+  public static String [] strings; // Noncompliant
```
benzonico: why the space between Type and Dim ?

java-checks/src/main/java/org/sonar/java/checks/InterfaceStaticMutableMemberCheck.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public void visitNode(Tree tree) {
+    for (Tree member : ((ClassTree) tree).members()) {
+      if (member.is(Kind.VARIABLE)) {
+        VariableTreeImpl variableTree = (VariableTreeImpl) member;
+        if (variableTree.modifiers().modifiers().contains(Modifier.STATIC)) {
+          Tree variableType = variableTree.type();
+          VariableSymbol symbol = variableTree.getSymbol();
+          if (variableType.is(Kind.ARRAY_TYPE) ||
```
benzonico: The condition is too complex to be let as is : this should probably be externalized in one or several submethod to understand at first sight what is going on.

java-checks/src/main/java/org/sonar/java/checks/InterfaceStaticMutableMemberCheck.java
```diff
@@ -0,0 +1,68 @@
...
+            (variableType.is(Kind.PARAMETERIZED_TYPE) && symbol.getType().isSubtypeOf(COLLECTION_QUALIFIED_NAME)) ||
+            (variableType.is(Kind.IDENTIFIER) && (symbol.getType().is(DATE_QUALIFIER_NAME) || symbol.getType().isSubtypeOf(COLLECTION_QUALIFIED_NAME)))) {
+            addIssue(variableTree, MESSAGE.replace("xxx", variableTree.simpleName().name()));
```
benzonico: Please do not use a string replace for messages: use formatting or concatenation.

m-g-sonar:
Source code updated based on pull request code review.

java-checks/src/main/java/org/sonar/java/checks/InterfaceStaticMutableMemberCheck.java
```diff
@@ -0,0 +1,79 @@
...
+@Rule(
+  key = "S2386",
+  priority = Priority.CRITICAL,
+  tags = {"unpredictable"})
+@BelongsToProfile(title = "Sonar way", priority = Priority.CRITICAL)
+public class InterfaceStaticMutableMemberCheck extends SubscriptionBaseVisitor {
+
+  private static final String MESSAGE = "Move \"%s\" to a class and lower its visibility";
+  private static final String COLLECTION_QUALIFIED_NAME = "java.util.Collection";
+  private static final String DATE_QUALIFIED_NAME = "java.util.Date";
```
benzonico: I would inline the constant here. It just brings an unecessary level of indirection in those cases.

java-checks/src/main/java/org/sonar/java/checks/InterfaceStaticMutableMemberCheck.java
```diff
@@ -0,0 +1,79 @@
...
+    return ImmutableList.of(Kind.INTERFACE);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    for (Tree member : ((ClassTree) tree).members()) {
+      if (member.is(Kind.VARIABLE)) {
+        VariableTreeImpl variableTree = (VariableTreeImpl) member;
+        if (isStaticMember(variableTree) && isMutableMember(variableTree)) {
+          addIssue(variableTree, String.format(MESSAGE, variableTree.simpleName().name()));
```
benzonico: In this case I would prefer MessageFormat as this is just a way to concatenate stuff rather than formatting it.

m-g-sonar:
Source code updated based on pull request code review (2nd pass).

m-g-sonar:
Removed message constant.

benzonico:
LGTM !

PR #27, opened 2015-01-23
SONARJAVA-865 RSPEC-2183 Ints and longs shift check

m-g-sonar:
Ints and longs shift check:

- Since an `int` is a 32-bit variable, shifting by more than (-)31 is confusing at best and an error at worst. Shifting an `int` by 32 is the same as shifting it by 0, and shifting it by 33 is the same as shifting it by 1.
- Similarly, shifting a `long` by (-)64 is the same as shifting it by 0, and shifting it by 65 is the same as shifting it by 1.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
@@ -76,7 +76,7 @@ public class ShiftOnIntOrLongCheck extends SubscriptionBaseVisitor {
 
     if (shift.is(Kind.INT_LITERAL, Kind.LONG_LITERAL)) {
       String value = ((LiteralTree) shift).value();
-      long numberBits = sign * Long.parseLong(value);
+      long numberBits = sign * Integer.decode(value);
       long reducedNumberBits = numberBits % (expectInt ? 32 : 64);
       String message = getMessage(numberBits, reducedNumberBits, expectInt, identifier);
       if (message != null) {
--- a/java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
@@ -24,7 +24,6 @@ import org.sonar.check.BelongsToProfile;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
 import org.sonar.java.model.AbstractTypedTree;
-import org.sonar.java.resolve.Type;
 import org.sonar.plugins.java.api.tree.ArrayAccessExpressionTree;
 import org.sonar.plugins.java.api.tree.AssignmentExpressionTree;
 import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
@@ -52,44 +51,40 @@ public class ShiftOnIntOrLongCheck extends SubscriptionBaseVisitor {
 
   @Override
   public void visitNode(Tree tree) {
-    ExpressionTree target = null;
-    ExpressionTree shift = null;
+    String identifier;
+    ExpressionTree shift;
 
     if (tree.is(Kind.LEFT_SHIFT, Kind.RIGHT_SHIFT)) {
       BinaryExpressionTree binaryExpressionTree = (BinaryExpressionTree) tree;
-      target = binaryExpressionTree.leftOperand();
+      identifier = getIdentifierName(binaryExpressionTree.leftOperand());
       shift = binaryExpressionTree.rightOperand();
     } else {
       AssignmentExpressionTree assignmentExpressionTree = (AssignmentExpressionTree) tree;
-      target = assignmentExpressionTree.variable();
+      identifier = getIdentifierName(assignmentExpressionTree.variable());
       shift = assignmentExpressionTree.expression();
     }
 
-    String identifier = getIdentifierName(target);
-    Type expectedType = ((AbstractTypedTree) target).getSymbolType();
-    boolean expectInt = expectedType.is("int") || expectedType.is("java.lang.Integer");
     int sign = shift.is(Kind.UNARY_MINUS) ? -1 : 1;
-
     if (shift.is(Kind.UNARY_MINUS, Kind.UNARY_PLUS)) {
       shift = ((UnaryExpressionTree) shift).expression();
     }
 
     if (shift.is(Kind.INT_LITERAL, Kind.LONG_LITERAL)) {
-      String value = ((LiteralTree) shift).value();
-      long numberBits = sign * Long.decode(value);
-      long reducedNumberBits = numberBits % (expectInt ? 32 : 64);
-      String message = getMessage(numberBits, reducedNumberBits, expectInt, identifier);
+      int base = getBase(tree);
+      long numberBits = sign * Long.decode(((LiteralTree) shift).value());
+      long reducedNumberBits = numberBits % base;
+      String message = getMessage(numberBits, reducedNumberBits, base, identifier);
       if (message != null) {
         addIssue(tree, message);
       }
     }
   }
 
-  private String getMessage(long numberBits, long reducedNumberBits, boolean expectInt, String identifier) {
+  private String getMessage(long numberBits, long reducedNumberBits, int base, String identifier) {
     if (reducedNumberBits == 0L) {
-      return MessageFormat.format("Remove this useless shift (multiple of {0})", expectInt ? 32 : 64);
-    } else if (tooManyBits(numberBits, expectInt)) {
-      if (expectInt) {
+      return MessageFormat.format("Remove this useless shift (multiple of {0})", base);
+    } else if (tooManyBits(numberBits, base)) {
+      if (base == 32) {
         return MessageFormat.format(
           identifier == null ?
             "Either use a \"long\" or correct this shift to {0}" :
@@ -102,16 +97,23 @@ public class ShiftOnIntOrLongCheck extends SubscriptionBaseVisitor {
     return null;
   }
 
-  private boolean tooManyBits(long numberBits, boolean expectInt) {
-    long value = Math.abs(numberBits);
-    return (expectInt && value >= 32) || (!expectInt && value >= 64);
+  private int getBase(Tree tree) {
+    if (((AbstractTypedTree) tree).getSymbolType().is("int")) {
+      return 32;
+    }
+    return 64;
+  }
+
+  private boolean tooManyBits(long numberBits, int base) {
+    return Math.abs(numberBits) >= base;
   }
 
   private String getIdentifierName(ExpressionTree tree) {
-    ExpressionTree expressionTree = tree;
     if (tree.is(Kind.ARRAY_ACCESS_EXPRESSION)) {
-      expressionTree = ((ArrayAccessExpressionTree) tree).expression();
+      return getIdentifierName(((ArrayAccessExpressionTree) tree).expression());
+    } else if (tree.is(Kind.IDENTIFIER)) {
+      return ((IdentifierTree) tree).name();
     }
-    return expressionTree.is(Kind.IDENTIFIER) ? ((IdentifierTree) expressionTree).name() : null;
+    return null;
   }
 }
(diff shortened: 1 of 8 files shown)


java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private boolean treeTypeIs(ExpressionTree tree, String fullyQualifiedName) {
+    return (tree.is(Kind.IDENTIFIER) && ((IdentifierTreeImpl) tree).getSymbolType().is(fullyQualifiedName)) ||
+      (tree.is(Kind.ARRAY_ACCESS_EXPRESSION) && ((ArrayAccessExpressionTreeImpl) tree).getSymbolType().is(fullyQualifiedName)) ||
+      (tree.is(Kind.PARENTHESIZED_EXPRESSION) && ((ParenthesizedTreeImpl) tree).getSymbolType().is(fullyQualifiedName)) ||
+      (tree.is(Kind.METHOD_INVOCATION) && ((MethodInvocationTreeImpl) tree).getSymbolType().is(fullyQualifiedName));
+  }
+
+  private boolean isShiftEvaluable(ExpressionTree tree, boolean intOrLong) {
+    return intOrLong && tree.is(Kind.INT_LITERAL, Kind.LONG_LITERAL);
```
benzonico: if tree is an INT_LITERAL or a LONG_LITERAL then its type is necessarily int or long so  the boolean parameter is useless here.

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,130 @@
...
+  private boolean treeTypeIsInt(ExpressionTree tree) {
+    return treeTypeIs(tree, "int") || treeTypeIs(tree, "java.lang.Integer");
+  }
+
+  private boolean treeTypeIsLong(ExpressionTree tree) {
+    return treeTypeIs(tree, "long") || treeTypeIs(tree, "java.lang.Long");
+  }
+
+  private boolean treeTypeIs(ExpressionTree tree, String fullyQualifiedName) {
+    return (tree.is(Kind.IDENTIFIER) && ((IdentifierTreeImpl) tree).getSymbolType().is(fullyQualifiedName)) ||
```
benzonico: getSymbolType() is defined on AbstractTypedTree, so only one cast should be necessary thus simplifying a lot this expression.

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,130 @@
...
+    boolean expectLong = treeTypeIsLong(target);
+    boolean expectInt = treeTypeIsInt(target);
+    boolean shiftIsMinus = shift.is(Kind.UNARY_MINUS);
+
+    if (shift.is(Kind.UNARY_MINUS, Kind.UNARY_PLUS)) {
+      shift = ((UnaryExpressionTree) shift).expression();
+    }
+
+    if (isShiftEvaluable(shift, expectInt || expectLong)) {
+      String value = ((LiteralTreeImpl) shift).value();
```
benzonico: No need to cast to LiteralTreeImpl in this case : LiteralTree is sufficient

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,130 @@
...
+  public void visitNode(Tree tree) {
+    ExpressionTree target = null;
+    ExpressionTree shift = null;
+
+    if (tree.is(Kind.LEFT_SHIFT, Kind.RIGHT_SHIFT)) {
+      BinaryExpressionTree binaryExpressionTree = (BinaryExpressionTree) tree;
+      target = binaryExpressionTree.leftOperand();
+      shift = binaryExpressionTree.rightOperand();
+    } else {
+      AssignmentExpressionTreeImpl assignmentExpressionTree = (AssignmentExpressionTreeImpl) tree;
```
benzonico: No need to go to the implementation in this case : working with AssignmentExpressionTree is sufficient

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,130 @@
...
+    boolean expectInt = treeTypeIsInt(target);
+    boolean shiftIsMinus = shift.is(Kind.UNARY_MINUS);
+
+    if (shift.is(Kind.UNARY_MINUS, Kind.UNARY_PLUS)) {
+      shift = ((UnaryExpressionTree) shift).expression();
+    }
+
+    if (isShiftEvaluable(shift, expectInt || expectLong)) {
+      String value = ((LiteralTreeImpl) shift).value();
+      long numberBits = (shiftIsMinus ? -1 : 1) * (expectInt ? Integer.parseInt(value) : Long.parseLong(value));
```
benzonico: Since your are anyway assigning it to a long, you don't care about parsing value with Integer.parseInt and can use directly Long.parseLong.

java-checks/src/test/java/org/sonar/java/checks/ShiftOnIntOrLongCheckTest.java
```diff
@@ -0,0 +1,80 @@
...
+public class ShiftOnIntOrLongCheckTest {
+
+  private static final String REMOVE_SHIFT_32 = "Remove this useless shift (multiple of 32)";
+  private static final String REMOVE_SHIFT_64 = "Remove this useless shift (multiple of 64)";
+  private static final String INT_IDENTIFIER_VALUE = "Either make \"xxx\" a \"long\" or correct this shift to nnn";
+  private static final String INT_VALUE = "Either use a \"long\" or correct this shift to nnn";
+  private static final String LONG_VALUE = "Correct this shift to nnn";
+
+  private String getMessage(String identifier, String value) {
+    return INT_IDENTIFIER_VALUE.replace("xxx", identifier).replace("nnn", value);
```
benzonico: please avoid the string replace

m-g-sonar:
Source code updated based on pull request code review.

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,120 @@
...
+    Type expectedType = ((AbstractTypedTree) target).getSymbolType();
+    boolean expectLong = expectedType.is("long") || expectedType.is("java.lang.Long");
+    boolean expectInt = expectedType.is("int") || expectedType.is("java.lang.Integer");
+    boolean shiftIsMinus = shift.is(Kind.UNARY_MINUS);
+
+    if (shift.is(Kind.UNARY_MINUS, Kind.UNARY_PLUS)) {
+      shift = ((UnaryExpressionTree) shift).expression();
+    }
+
+    if ((expectInt || expectLong) && isShiftValueEvaluable(shift)) {
```
benzonico: The first or is really not needed. If you are working with shift, the result is necessarily an int or a long.

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,120 @@
...
+  key = "S2183",
+  priority = Priority.CRITICAL,
+  tags = {"bug"})
+@BelongsToProfile(title = "Sonar way", priority = Priority.CRITICAL)
+public class ShiftOnIntOrLongCheck extends SubscriptionBaseVisitor {
+
+  private static final String USELESS_SHIFT_MESSAGE = "Remove this useless shift (multiple of {0})";
+  private static final String INTEGER_NO_ID_MESSAGE = "Either use a \"long\" or correct this shift to {0}";
+  private static final String INTEGER_ID_MESSAGE = "Either make \"{1}\" a \"long\" or correct this shift to {0}";
+  private static final String LONG_MESSAGE = "Correct this shift to {0}";
```
benzonico: Definitely : messages should be inlined in code. It is not going to be reused and it makes the code difficult to read as it is indirected if you put them as constants.

m-g-sonar:
Source code updated based on pull request code review (2nd pass).

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,114 @@
...
+
+    if (shift.is(Kind.INT_LITERAL, Kind.LONG_LITERAL)) {
+      String value = ((LiteralTree) shift).value();
+      long numberBits = (shiftIsMinus ? -1 : 1) * Long.parseLong(value);
+      long reducedNumberBits = numberBits % (expectInt ? 32 : 64);
+      insertIssue(tree, numberBits, reducedNumberBits, expectInt, identifier);
+    }
+  }
+
+  private void insertIssue(Tree tree, long numberBits, long reducedNumberBits, boolean expectInt, String identifier) {
```
benzonico: I tend to think, that this method should only compute the  message (thus returning a String).

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,114 @@
...
+    boolean expectInt = expectedType.is("int") || expectedType.is("java.lang.Integer");
+    boolean shiftIsMinus = shift.is(Kind.UNARY_MINUS);
+
+    if (shift.is(Kind.UNARY_MINUS, Kind.UNARY_PLUS)) {
+      shift = ((UnaryExpressionTree) shift).expression();
+    }
+
+    if (shift.is(Kind.INT_LITERAL, Kind.LONG_LITERAL)) {
+      String value = ((LiteralTree) shift).value();
+      long numberBits = (shiftIsMinus ? -1 : 1) * Long.parseLong(value);
```
benzonico: Given that this is the only place where shiftIsMinus is used, maybe the -1 or 1 should be computed directly...

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
m-g-sonar: I'm not sure of it, as the method does not necessarily add issue to the tree. However, I can change it to return a message (potentially null when no message should be added) and test the return value before adding the issue.

benzonico: I would prefer this in fact... thus the method will only compute the message and not do two things.

m-g-sonar:
Source code updated based on pull request code review (3rd pass).

m-g-sonar:
While running integration tests, discovered cases of non-decimal shifts not correctly handled. Fixed in last commit.

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,117 @@
...
+    boolean expectInt = expectedType.is("int") || expectedType.is("java.lang.Integer");
+    int sign = shift.is(Kind.UNARY_MINUS) ? -1 : 1;
+
+    if (shift.is(Kind.UNARY_MINUS, Kind.UNARY_PLUS)) {
+      shift = ((UnaryExpressionTree) shift).expression();
+    }
+
+    if (shift.is(Kind.INT_LITERAL, Kind.LONG_LITERAL)) {
+      String value = ((LiteralTree) shift).value();
+      long numberBits = sign * Integer.decode(value);
```
benzonico: Why not Long.decode ?

m-g-sonar: It's a mistake. I did my first tests with Integer and I forget that i was initially using Long.

m-g-sonar:
Fixed wrong cast.

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -0,0 +1,117 @@
...
+      target = binaryExpressionTree.leftOperand();
+      shift = binaryExpressionTree.rightOperand();
+    } else {
+      AssignmentExpressionTree assignmentExpressionTree = (AssignmentExpressionTree) tree;
+      target = assignmentExpressionTree.variable();
+      shift = assignmentExpressionTree.expression();
+    }
+
+    String identifier = getIdentifierName(target);
+    Type expectedType = ((AbstractTypedTree) target).getSymbolType();
```
benzonico: In fact the expectedType and thus expectInt variable can be computed directly from the type of the variable tree casted to an AbstractTypedTree. This should really simplify this visitNode method.

benzonico:
LGTM !

PR #28, opened 2015-01-26
SONARJAVA-866 RSPEC-2167 CompareTo return value check

m-g-sonar:
"compareTo" return value check

- It is the sign, rather than the magnitude of the value returned from `compareTo` that matters. Returning `Integer.MIN_VALUE` does not convey a higher degree of inequality, and doing so can cause errors because the return value of `compareTo` is sometimes inversed, with the expectation that negative values become positive. However, inversing `Integer.MIN_VALUE` yields `Integer.MIN_VALUE` rather than `Integer.MAX_VALUE`.
- The rule check the return value of the `compareTo` methods and add an issue if `Integer.MIN_VALUE` is returned.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CompareToReturnValueCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CompareToReturnValueCheck.java
@@ -20,17 +20,23 @@
 package org.sonar.java.checks;
 
 import com.google.common.collect.ImmutableList;
+import com.sonar.sslr.api.AstNode;
 import org.sonar.check.BelongsToProfile;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
+import org.sonar.java.model.declaration.MethodTreeImpl;
+import org.sonar.java.model.declaration.VariableTreeImpl;
 import org.sonar.java.model.expression.IdentifierTreeImpl;
+import org.sonar.java.model.statement.ReturnStatementTreeImpl;
+import org.sonar.java.resolve.Type;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodTree;
-import org.sonar.plugins.java.api.tree.PrimitiveTypeTree;
 import org.sonar.plugins.java.api.tree.ReturnStatementTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.Tree.Kind;
+import org.sonar.plugins.java.api.tree.VariableTree;
 
 import java.util.List;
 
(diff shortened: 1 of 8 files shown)


java-checks/src/main/java/org/sonar/java/checks/CompareToReturnValueCheck.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+
+@Rule(
+  key = "S2167",
+  priority = Priority.CRITICAL,
+  tags = {"bug"})
+@BelongsToProfile(title = "Sonar way", priority = Priority.CRITICAL)
+public class CompareToReturnValueCheck extends SubscriptionBaseVisitor {
+
+  private boolean insideCompareTo = false;
```
benzonico: A simple flag will not cover the corner case where an inner class is defined inside a compareTo method of an Outer class.

java-checks/src/main/java/org/sonar/java/checks/CompareToReturnValueCheck.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Override
+  public void leaveNode(Tree tree) {
+    if (tree.is(Tree.Kind.METHOD)) {
+      insideCompareTo = false;
+    }
+  }
+
+  private boolean isCompareToDeclaration(MethodTree tree) {
```
benzonico: We can probably use/improve the AbstractMethodDetection mechanism to find compareTo method ( :warning: Generics are not directly supported in this class right now)

benzonico:
I think that we should probably approach this by detecting the compareTo method then subvisit the body of the method and detecting return statements (and cutting the visit of inner classes to cover the corner case of inner classes defined inside compareTo methods).

m-g-sonar:
Ok I'll try this approach!

m-g-sonar:
Updated code to use BaseTreeVisitor.

java-checks/src/main/java/org/sonar/java/checks/CompareToReturnValueCheck.java
```diff
@@ -0,0 +1,110 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    MethodTree methodTree = (MethodTree) tree;
+    if (isCompareToDeclaration(methodTree)) {
+      methodTree.accept(new ReturnStatementVisitor(methodTree));
+    }
+  }
+
+  private boolean isCompareToDeclaration(MethodTree tree) {
+    return isMethodName(tree, "compareTo") && hasOneNonPrimitiveParameter(tree) && returnsInt(tree);
```
benzonico: Method has only one purpose, find out name of method is "compareTo" so this parameter should be part of the method and the method should be renamed.

java-checks/src/main/java/org/sonar/java/checks/CompareToReturnValueCheck.java
```diff
@@ -0,0 +1,110 @@
...
+    public ReturnStatementVisitor(MethodTree callingMethod) {
+      this.callingMethod = ((MethodTreeImpl) callingMethod).getAstNode();
+    }
+
+    @Override
+    public void visitReturnStatement(ReturnStatementTree tree) {
+      if (isWithinCallingMethodBody(tree) && returnsIntegerMinValue(tree.expression())) {
+        addIssue(tree, "Simply return -1");
+      }
+      super.visitReturnStatement(tree);
```
benzonico: uncessary call to super. No need to visit further.

m-g-sonar:
Code updated.

java-checks/src/main/java/org/sonar/java/checks/CompareToReturnValueCheck.java
```diff
@@ -0,0 +1,103 @@
...
+    if (isCompareToDeclaration(methodTree)) {
+      methodTree.accept(new ReturnStatementVisitor());
+    }
+  }
+
+  private boolean isCompareToDeclaration(MethodTree tree) {
+    return isCalledCompareTo(tree) && hasOneNonPrimitiveParameter(tree) && returnsInt(tree);
+  }
+
+  private boolean isCalledCompareTo(MethodTree tree) {
```
benzonico: named would be better : the term call has a total different meaning in the context of java code.

m-g-sonar:
Method renamed with more appropriate name.

benzonico:
LGTM.

PR #30, opened 2015-01-28
HardCodedCredentialCheck should not consider empty string literal a violation

torakiki:
SONARJAVA-788 RSPEC-2068:  HardCodedCredentialCheck should not consider empty string literal as a violation.

I have few false positives where a mutable variable called "password" is initialized to the empty string and it's considered a critical violation. I think the empty string can hardly be seen as hard coded credential and it may happen to have password variable assigned to it.


benzonico:
Hi and thanks for your contribution, 

In fact we disagree with the fact that empty string is necessarily not an issue. It can also be seen as a really bad (I mean really _really_ bad) way to hardcode an empty password. 
As for security rules we tend to raise potential issue (so we are less regarding towards false positive) I am afraid that your approach will likely hide problems rather than prevent noise.

So in the current state of your pull request I am going to reject it. 
Sorry.

PR #32, opened 2015-01-30
SONARJAVA-862 RSPEC-2386 Exclude Immutable Collection

benzonico:
Exclusion of immutable collections from guava. 
The added .class file is required to correctly unit test this exclusion.


m-g-sonar:
LGTM.

PR #33, opened 2015-01-30
SONARJAVA-821 SONARJAVA-834 RSPEC-1452

benzonico:
Remove false positive on overriden method and class literals on RSPEC-1452


m-g-sonar:
LGTM.

PR #37, opened 2015-02-02
SONARJAVA-861 sonar.squid.analyse.property now taken into account in Public API metric

m-g-sonar:
Added management of the property `sonar.squid.analyse.property` in the computation of the Public API metric. If the property is set to `true`, then `getters` and `setters` are now ignored.

### DIFF
--- a/java-squid/src/main/java/org/sonar/java/ast/visitors/PublicApiChecker.java
+++ b/java-squid/src/main/java/org/sonar/java/ast/visitors/PublicApiChecker.java
@@ -76,7 +76,7 @@ public class PublicApiChecker extends BaseTreeVisitor {
   private final Deque<Tree> currentParents = new LinkedList<Tree>();
   private double publicApi;
   private double documentedPublicApi;
-  private boolean separateAccessorsFromMethods;
+  private final boolean separateAccessorsFromMethods;
   private final AccessorVisitor accessorVisitor;
 
   public static PublicApiChecker newInstanceWithAccessorsHandledAsMethods() {
--- a/java-squid/src/main/java/org/sonar/java/ast/visitors/PublicApiChecker.java
+++ b/java-squid/src/main/java/org/sonar/java/ast/visitors/PublicApiChecker.java
@@ -79,11 +79,11 @@ public class PublicApiChecker extends BaseTreeVisitor {
   private boolean separateAccessorsFromMethods;
   private final AccessorVisitor accessorVisitor;
 
-  public static PublicApiChecker newPublicApiCheckerAccessorsHandledAsMethods() {
+  public static PublicApiChecker newInstanceWithAccessorsHandledAsMethods() {
     return new PublicApiChecker(false);
   }
 
-  public static PublicApiChecker newPublicApiCheckerAccessorsSeparatedFromMethods() {
+  public static PublicApiChecker newInstanceWithAccessorsSeparatedFromMethods() {
     return new PublicApiChecker(true);
   }
 
@@ -298,9 +298,4 @@ public class PublicApiChecker extends BaseTreeVisitor {
     }
     return ParsingUtils.scaleValue(documentedPublicApi / publicApi * 100, 2);
   }
-
-  public boolean separatesAccessorsFromMethods() {
-    return separateAccessorsFromMethods;
-  }
-
 }
--- a/java-squid/src/main/java/org/sonar/java/ast/visitors/PublicApiChecker.java
+++ b/java-squid/src/main/java/org/sonar/java/ast/visitors/PublicApiChecker.java
@@ -44,6 +44,7 @@ import org.sonar.plugins.java.api.tree.Tree.Kind;
 import org.sonar.plugins.java.api.tree.VariableTree;
 
 import javax.annotation.Nullable;
+
 import java.util.Deque;
 import java.util.LinkedList;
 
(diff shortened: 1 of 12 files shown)


java-squid/src/main/java/org/sonar/java/ast/visitors/PublicApiChecker.java
```diff
@@ -75,6 +75,13 @@
   private final Deque<Tree> currentParents = new LinkedList<Tree>();
   private double publicApi;
   private double documentedPublicApi;
+  private boolean analyseAccessors;
+  private final AccessorVisitor accessorVisitor;
+
+  public PublicApiChecker(boolean analyseAccessors) {
```
benzonico: Boolean trap : http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html
Having a constructor with a boolean parameter does not seem the best option here. Maybe two methods to create the instance with correct setting and this constructor goes private.

java-squid/src/test/java/org/sonar/java/ast/visitors/PublicApiCheckerTest.java
```diff
@@ -43,17 +43,20 @@
 
   private PublicApiChecker publicApiChecker;
   private CompilationUnitTree cut;
+  private boolean analyseAccessors;
```
benzonico: Why a field if it is always set to true ?

m-g-sonar: It is not always set to true, see line 113. It is changed before re-doing the test.

java-squid/src/test/java/org/sonar/java/SquidUserGuideTest.java
```diff
@@ -144,7 +144,8 @@ public void measures_on_project() throws Exception {
     assertThat(metrics.get("statements").intValue()).isEqualTo(12047);
     assertThat(metrics.get("complexity").intValue()).isEqualTo(8475 - 80 /* SONAR-3793 */- 2 /* SONAR-3794 */);
     assertThat(metrics.get("comment_lines").intValue()).isEqualTo(17908);
-    assertThat(metrics.get("public_api").intValue()).isEqualTo(3221);
+    // 69: SONARJAVA-861 analyseAccessors property of the measurer is set to true. Getters and setters ignored.
+    assertThat(metrics.get("public_api").intValue()).isEqualTo(3221 - 69);
```
benzonico: We should probably have two tests: one with the value of the property set to true and the other one to false, to actually see the difference.

java-squid/src/test/java/org/sonar/java/StrutsTest.java
```diff
@@ -89,7 +89,8 @@ public void measures_on_project() throws Exception {
     assertThat(metrics.get("statements").intValue()).isEqualTo(6403);
     assertThat(metrics.get("complexity").intValue()).isEqualTo(3957 - 145 /* SONAR-3793 */ - 1 /* SONAR-3794 */);
     assertThat(metrics.get("comment_lines").intValue()).isEqualTo(7605);
-    assertThat(metrics.get("public_api").intValue()).isEqualTo(1340);
```
benzonico: We should probably have two tests: one with the value of the property set to true and the other one to false, to actually see the difference.

java-squid/src/test/java/org/sonar/java/ast/visitors/PublicApiCheckerTest.java
```diff
@@ -99,7 +107,12 @@ public void leaveNode(Tree tree) {
           methodTrees.pop();
         }
       }
-    }.scanTree(cut);
+    };
+
+    visitor.scanTree(cut);
```
benzonico: better split into two tests.

m-g-sonar:
Code updated based on previous comments.

java-squid/src/main/java/org/sonar/java/ast/visitors/PublicApiChecker.java
```diff
@@ -280,4 +298,9 @@ public double getDocumentedPublicApiDensity() {
     }
     return ParsingUtils.scaleValue(documentedPublicApi / publicApi * 100, 2);
   }
+
+  public boolean separatesAccessorsFromMethods() {
```
benzonico: This method is there only for tests and test code can be reworked to remove it.

java-squid/src/main/java/org/sonar/java/ast/visitors/PublicApiChecker.java
```diff
@@ -75,6 +76,21 @@
   private final Deque<Tree> currentParents = new LinkedList<Tree>();
   private double publicApi;
   private double documentedPublicApi;
+  private boolean separateAccessorsFromMethods;
```
benzonico: should be final ? or updated in instance methods to avoid boolean constructor ?

java-squid/src/test/java/org/sonar/java/SquidUserGuideTest.java
```diff
@@ -144,6 +143,39 @@ public void measures_on_project() throws Exception {
...
+    if (metrics.get("public_api").intValue() != 0) {
+      density = (metrics.get("public_api") - metrics.get("public_undocumented_api")) / metrics.get("public_api");
+    }
+    assertThat(density).isEqualTo(0.64, Delta.delta(0.01));
+  }
+
+  @Test
+  public void measures_on_project_accessors_handled_as_methods() throws Exception {
+    initAndScan(false);
+    ArgumentCaptor<Measure> captor = ArgumentCaptor.forClass(Measure.class);
```
benzonico: The capture of metrics and some assertions can probably be factorized and not duplicated.

java-squid/src/test/java/org/sonar/java/StrutsTest.java
```diff
@@ -89,8 +88,33 @@ public void measures_on_project() throws Exception {
...
+    // 48: SONARJAVA-861 analyseAccessors property of the measurer is set to true. Getters and setters not counted as public api.
+    assertThat(metrics.get("public_api").intValue()).isEqualTo(1340 - 48);
   }
 
+  @Test
+  public void measures_on_project_accessors_handled_as_methods() throws Exception {
+    initAndScan(false);
+    ArgumentCaptor<Measure> captor = ArgumentCaptor.forClass(Measure.class);
```
benzonico: Capture of arguments and some metrics can be factorized.

benzonico:
LGTM. Don't forget integration tests.

PR #38, opened 2015-02-02
SONARJAVA-878 Relax validation on sonar.java.binaries

benzonico:
Validate only when sources files are presents to be backward compatible with multi module analysis when using sonar-maven-plugin 2.2


m-g-sonar:
LGTM.

PR #39, opened 2015-02-02
Fix typo in DependenciesVisitor.java

lzilber:


benzonico:
Thanks for your contribution.

There is actually a (bad) test on this unused method. The fix for this would be to remove this alltogether. I'll handle this like that so I'll close your PR when this is done.

benzonico:
Fixed with commit : https://github.com/SonarSource/sonar-java/commit/6bf4fe5cae212d7d314cb0bac31c5bc15cf58229

lzilber:
thanks!

PR #41, opened 2015-02-03
SONARJAVA-867 RSPEC-2165 Finalize fields set check

m-g-sonar:
Finalize fields set check :
- Added new rule to check `null` assignment on fields in `finalize` methods.
- There is no point in setting class fields to `null` in a finalizer. If this this is a hint to the garbage collector, it is unnecessary - the object will be garbage collected anyway - and may actually cause extra work for the garbage collector.


m-g-sonar:
Code updated with respect to comments from code review.

java-checks/src/main/java/org/sonar/java/checks/FinalizeFieldsSetCheck.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private boolean isFieldAssignment(AssignmentExpressionTree tree) {
+      ExpressionTree variable = tree.variable();
+      if (variable.is(Kind.MEMBER_SELECT)) {
+        variable = ((MemberSelectExpressionTree) variable).expression();
+      }
+      if (variable.is(Kind.IDENTIFIER)) {
+        Symbol variableSymbol = getSemanticModel().getReference((IdentifierTree) variable);
+        return variableSymbol.owner().isKind(Symbol.TYP);
```
benzonico: This is not sufficient if you have an expression refering a field of another instance  or from another class : 
class A  {
  private int field;
  protected void finalize() {
    new A().field = null;
    new B().field = null;
  }
}

java-checks/src/main/java/org/sonar/java/checks/FinalizeFieldsSetCheck.java
```diff
@@ -0,0 +1,114 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    MethodTree methodTree = (MethodTree) tree;
+    if (isFinalizeDeclaration(methodTree)) {
+      methodTree.accept(new AssignmentVisitor());
+    }
+  }
+
+  private boolean isFinalizeDeclaration(MethodTree tree) {
+    return hasSemantic() && isMethodNamedFinalize(tree) && isOverriding(tree) && hasNoParameters(tree);
```
benzonico: Check for overriding is uncessary.

java-checks/src/main/java/org/sonar/java/checks/FinalizeFieldsSetCheck.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private boolean isNullAssignment(AssignmentExpressionTree tree) {
+      return tree.expression().is(Kind.NULL_LITERAL);
+    }
+
+    private String getFieldName(AssignmentExpressionTree tree) {
+      ExpressionTree variable = tree.variable();
+      if (variable.is(Kind.MEMBER_SELECT)) {
+        variable = ((MemberSelectExpressionTree) variable).expression();
```
benzonico: The expression won't contain the field name.

java-checks/src/main/java/org/sonar/java/checks/FinalizeFieldsSetCheck.java
```diff
@@ -0,0 +1,116 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    MethodTree methodTree = (MethodTree) tree;
+    if (isFinalizeDeclaration(methodTree)) {
+      methodTree.accept(new AssignmentVisitor());
+    }
+  }
+
+  private boolean isFinalizeDeclaration(MethodTree tree) {
+    return hasSemantic() && isMethodNamedFinalize(tree) && hasNoParameters(tree);
```
benzonico: hasSemantic has nothing to do in this method directly.

benzonico:
LGTM

PR #49, opened 2015-02-06
SONARJAVA-868 RSPEC-2175 Inappropriate collection check

m-g-sonar:
Check on inappropriate "Collection" calls:

- Calls of methods `contains` and `remove` of any sub-type of `java.util.Collection` are now checked to control compatibility of provided argument (manage boxing and unknown type).
- Added new methods in the `ParametrizedTypeType` class to access more easily type parameters.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CollectionInappropriateCallsCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CollectionInappropriateCallsCheck.java
@@ -20,6 +20,7 @@
 package org.sonar.java.checks;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Maps;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
@@ -43,6 +44,7 @@ import javax.annotation.Nullable;
 
 import java.text.MessageFormat;
 import java.util.List;
+import java.util.Map;
 
 @Rule(
   key = "S2175",
@@ -53,6 +55,8 @@ import java.util.List;
 @SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.DATA_RELIABILITY)
 @SqaleConstantRemediation("15min")
 public class CollectionInappropriateCallsCheck extends AbstractMethodDetection {
+
+  private final Map<String, String> primitiveToPrimitiveWrapper = buildPrimitiveMapper();
 
   @Override
   protected List<MethodInvocationMatcher> getMethodInvocationMatchers() {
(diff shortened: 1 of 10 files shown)


java-checks/src/main/java/org/sonar/java/checks/CollectionInappropriateCallsCheck.java
```diff
@@ -0,0 +1,147 @@
...
+  private boolean isArgumentCompatible(Type argumentType, Type collectionParameterType) {
+    return (isSubtypeOf(argumentType, collectionParameterType) || isBoxingPossible(argumentType, collectionParameterType));
+  }
+
+  private boolean isSubtypeOf(Type argumentType, Type collectionParameterType) {
+    return argumentType.isSubtypeOf(collectionParameterType.getSymbol().getFullyQualifiedName());
+  }
+
+  private boolean isKnown(Type type) {
+    return type != null && !type.isTagged(Type.UNKNOWN);
```
benzonico: You should not have to test for null and so inline this method.

java-checks/src/main/java/org/sonar/java/checks/CollectionInappropriateCallsCheck.java
```diff
@@ -0,0 +1,147 @@
...
+  private boolean isBoxingPossible(Type argumentType, Type collectionParameterType) {
+    return argumentType.isPrimitive() && collectionParameterType.isPrimitiveWrapper() && boxingMatch(argumentType, collectionParameterType);
+  }
+
+  private boolean boxingMatch(Type primitiveType, Type primitiveWrapperType) {
+    String primitiveWrapperFullyQualifiedName = primitiveToPrimitiveWrapper.get(primitiveType.getSymbol().getFullyQualifiedName());
+    return primitiveWrapperFullyQualifiedName.equals(primitiveWrapperType.getSymbol().getFullyQualifiedName());
+  }
+
+  private Map<String, String> buildPrimitiveMapper() {
```
benzonico: should be a static immutable final map and more probably in Type object.

java-checks/src/test/files/checks/CollectionInappropriateCallsCheck.java
```diff
@@ -0,0 +1,114 @@
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+class A {
```
benzonico: Lacks the case of List<Number> l = Lists.<Integer>newArrayList(1);  l.contains(1);

java-checks/src/main/java/org/sonar/java/checks/CollectionInappropriateCallsCheck.java
m-g-sonar: For raw types such as `List list = Lists.newArrayList();`, the test is required.

java-squid/src/main/java/org/sonar/java/resolve/Type.java
```diff
@@ -136,6 +142,24 @@ public String toString() {
     return symbol == null ? "" : symbol.toString();
   }
 
+  @Nullable
+  public Type primitiveType() {
+    return primitiveType;
+  }
+
+  protected void setPrimitiveType(Type primitiveType) {
```
benzonico: we truly don't need setters. The property should be package visible.

java-squid/src/test/java/org/sonar/java/resolve/SymbolsTest.java
```diff
@@ -43,41 +43,57 @@ public void builtin_types() {
     assertThat(symbols.byteType.symbol.name).isEqualTo("byte");
```
benzonico: What about a test on a type that is NOT a primitive.

benzonico:
LGTM!

PR #50, opened 2015-02-09
SONARJAVA-892 RSPEC-2447 Boolean method return value check

m-g-sonar:
Boolean methid return value check:

- New check verifying return statements of methods returning `Boolean`. 
- While `null` is technically a valid `Boolean` value, that fact, and the distinction between `Boolean` and `boolean` is easy to forget. So returning `null` from a `Boolean` method is likely to cause problems with callers' code.


benzonico:
LGTM!

PR #53, opened 2015-02-10
SONARJAVA-890 RSPEC-2437 Silly bit operation check

m-g-sonar:
New check on silly bit operation.
- Certain bit operations are just silly and should not be performed because their results are predictable.
- Specifically, using `& -1` with any value will always result in the original value, as will `anyValue ^ 0` and `anyValue | 0`.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/SillyBitOperationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/SillyBitOperationCheck.java
@@ -23,6 +23,7 @@ import com.google.common.collect.ImmutableList;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
+import org.sonar.java.model.LiteralUtils;
 import org.sonar.plugins.java.api.tree.AssignmentExpressionTree;
 import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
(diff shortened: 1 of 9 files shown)


java-checks/src/main/java/org/sonar/java/checks/SillyBitOperationCheck.java
```diff
@@ -0,0 +1,115 @@
...
+      Kind.OR,
+      Kind.OR_ASSIGNMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    Long identityElement = getBitwiseOperationIdentityElement(tree);
+    Long evaluatedExpression = evaluateExpression(getExpression(tree));
+
+    if (evaluatedExpression != null && evaluatedExpression == identityElement) {
```
benzonico: use equals instead of ==

java-checks/src/main/java/org/sonar/java/checks/SillyBitOperationCheck.java
```diff
@@ -0,0 +1,115 @@
...
+      expression = ((UnaryExpressionTree) expression).expression();
+    }
+
+    if (expression.is(Kind.INT_LITERAL, Kind.LONG_LITERAL)) {
+      return sign * Long.decode(trimLongSuffix(((LiteralTree) expression).value()));
+    }
+    return null;
+  }
+
+  private String trimLongSuffix(String longString) {
```
benzonico: This should probably find its way in LiteralUtils class if it is already duplicated.

java-squid/src/test/java/org/sonar/java/model/LiteralUtilsTest.java
```diff
@@ -55,4 +55,12 @@ public void intValue() throws Exception {
     }
   }
 
+  @Test
+  public void testTrimLongSuffix() throws Exception {
+    String longValue = "12345";
```
benzonico: What If I give the empty string as an input of that method ?

benzonico: What if I pass null, What if Pass a null LiteralTree ?

benzonico:
LGTM!

PR #55, opened 2015-02-10
Allow custom adjustments

nrainer:
Allow adding further exclusions in sub classes by changing the method modifiers to protected.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CompareObjectWithEqualsCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CompareObjectWithEqualsCheck.java
@@ -51,6 +51,10 @@ public class CompareObjectWithEqualsCheck extends BaseTreeVisitor implements Jav
 
   private JavaFileScannerContext context;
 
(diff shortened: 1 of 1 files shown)


benzonico:
Hi @nrainer and thanks for your contribution.

I will not accept your pull request as for this is a hack to access symbol table and it breaks encapsulation. We have in our plan to work on an API to access semantic information from custom check but this won't be available right away. 

I am therefore closing this pull request.

nrainer:
Hi @benzonico 
I understand your concerns. I did not want to exploit the access to semantic information.

The idea was to allow for excluding further classes:
In our project, instances of certain (singletone) classes are supposed to be compared with == to improve the readability of the code and to avoid NPEs (eg for UIDs). A SonarQube rule checks the code for this convention, however the rule obviously collides with the CompareObjectWithEqualsCheck rule.

Would it be possible to add a parameter to the rule which contains comma-separated qualified names of classes to be excluded from the analysis? (like `isJavaLangClassComparison` does for instances of `Class`)

PR #56, opened 2015-02-10
SONARJAVA-888 RSPEC-2160 Equals not overriden in subclass check

m-g-sonar:
New check on equals methods not overridden in subclass:

- Extend a class that overrides `equals` and add fields without overriding `equals` in the subclass, and you run the risk of non-equivalent instances of your subclass being seen as equal, because only the superclass fields will be considered in the equality test.
- The check verifies that any class extending a class implementing `equals` (which is not `java.lang.Object`) implements the `equals` method if it has new fields.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverriddenInSubclassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverriddenInSubclassCheck.java
@@ -1,0 +1,123 @@
+/*
+ * SonarQube Java
+ * Copyright (C) 2012 SonarSource
+ * dev@sonar.codehaus.org
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02
+ */
+package org.sonar.java.checks;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+import org.sonar.api.server.rule.RulesDefinition;
+import org.sonar.check.Priority;
+import org.sonar.check.Rule;
+import org.sonar.java.model.AbstractTypedTree;
+import org.sonar.java.model.declaration.ClassTreeImpl;
+import org.sonar.java.resolve.Symbol;
+import org.sonar.java.resolve.Symbol.MethodSymbol;
+import org.sonar.java.resolve.Symbol.TypeSymbol;
+import org.sonar.java.resolve.Type;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.Modifier;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.Tree.Kind;
+import org.sonar.plugins.java.api.tree.VariableTree;
+import org.sonar.squidbridge.annotations.ActivatedByDefault;
+import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
+import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
+
+import java.util.List;
+
+@Rule(
+  key = "S2160",
+  name = "Subclasses that add fields should override \"equals\"",
+  tags = {"bug"},
+  priority = Priority.MAJOR)
+@ActivatedByDefault
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.LOGIC_RELIABILITY)
+@SqaleConstantRemediation("30min")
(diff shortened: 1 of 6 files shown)


java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverriddenInSubclassCheck.java
```diff
@@ -0,0 +1,123 @@
...
+    for (Tree member : classTree.members()) {
+      if (isField(member)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private boolean isField(Tree tree) {
+    return tree.is(Kind.VARIABLE) && !((VariableTree) tree).modifiers().modifiers().containsAll(Lists.newArrayList(Modifier.FINAL, Modifier.STATIC));
```
benzonico: What about a static member ? 
class A extends B {
    public static String staticMember;
}

java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverriddenInSubclassCheck.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private boolean implementsEquals(ClassTree classTree) {
+    return hasEqualsMethod(((ClassTreeImpl) classTree).getSymbol());
+  }
+
+  private boolean parentClassImplementsEquals(ClassTree tree) {
+    Tree superClass = tree.superClass();
+    if (superClass != null) {
+      TypeSymbol superClassSymbol = ((AbstractTypedTree) superClass).getSymbolType().getSymbol();
```
benzonico: Minor detail but you can check type and then assign a TypeSymbol variable.

java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverriddenInSubclassCheck.java
```diff
@@ -0,0 +1,123 @@
...
+      while (!superClassSymbol.getType().is("java.lang.Object")) {
+        if (hasEqualsMethod(superClassSymbol)) {
+          return true;
+        }
+        Type superClassType = superClassSymbol.getSuperclass();
+        if (superClassType == null) {
```
benzonico: Why this null check ? it deserves at least a comment as this case is not supposed to happen.

java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverriddenInSubclassCheck.java
```diff
@@ -0,0 +1,123 @@
...
+    List<Symbol> equalsMembers = superClassSymbol.members().lookup("equals");
+    for (Symbol symbol : equalsMembers) {
+      if (isEqualsMethod(symbol)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private boolean isEqualsMethod(Symbol symbol) {
```
benzonico: I am pretty sure there is already an implementation somewhere of this method. This should probably be factorised somewhere (probably in Symbol class)

java-checks/src/test/files/checks/EqualsNotOverriddenInSubclassCheck.java
```diff
@@ -0,0 +1,60 @@
...
+}
+
+class B extends A { // Noncompliant; instances will use A's equals method
+  String s2;
+}
+
+abstract class C {
+  boolean equals;
+  
+  public abstract boolean equals(Object obj);
```
benzonico: Does this compile ?

m-g-sonar: Yes it does. It allows you to force all the sub-classes to implement equals...

java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverriddenInSubclassCheck.java
m-g-sonar: I cannot find methods in other check which are based on the `symbol`. They all use the `Tree`s (`MethodTree`, `ExpressionTree`, ...) as parameter. Do I refactor them to only use this method instead?

benzonico: So, no. it was a mistake on my side, detection of method based on symbol is not the same thing as detecting it with tree.

java-checks/src/test/files/checks/EqualsNotOverriddenInSubclassCheck.java
benzonico: ok.

java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverriddenInSubclassCheck.java
```diff
@@ -0,0 +1,115 @@
...
+
+  private boolean implementsEquals(ClassTree classTree) {
+    return hasEqualsMethod(((ClassTreeImpl) classTree).getSymbol());
+  }
+
+  private boolean parentClassImplementsEquals(ClassTree tree) {
+    Tree superClass = tree.superClass();
+    if (superClass != null) {
+      Type superClassType = ((AbstractTypedTree) superClass).getSymbolType();
+      while (!superClassType.erasure().isTagged(Type.UNKNOWN) && !superClassType.is("java.lang.Object")) {
```
benzonico: Why call to erasure() ?

benzonico:
LGTM

m-g-sonar:
About getting the symbol to get the type again: This is a workaround to avoid issues when handling parametric types. There is a JIRA ticket related to this problem: [SONARJAVA-901](http://jira.codehaus.org/browse/SONARJAVA-901)

PR #64, opened 2015-02-16
SONARJAVA-900 RSPEC-2440 Fields ignored

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/ClassWithOnlyStaticMethodsInstantiationCheck.java
```diff
@@ -84,15 +85,6 @@ private boolean hasOnlyStaticMethod(TypeSymbol typeSymbol) {
     return hasStaticMethod;
```
benzonico: Can be changed to a simple return true;

benzonico:
LGTM

PR #65, opened 2015-02-16
SONARJAVA-896 Subtyping of arrays

benzonico:
Fix subtyping of Arrays and update CollectionInnapropriate calls


m-g-sonar:
LGTM!

PR #66, opened 2015-02-16
SONARJAVA-904 RSPEC-2444 lazy static field initialization : Exclude primitives

benzonico:
Fix subtyping of Arrays and update CollectionInnapropriate calls


m-g-sonar:
LGTM.

PR #67, opened 2015-02-16
SONARJAVA-886 RSPEC-2154 Primitive wrappers in ternary operator check

m-g-sonar:


java-checks/src/test/files/checks/PrimitiveWrappersInTernaryOperatorCheck.java
```diff
@@ -0,0 +1,15 @@
+class A {
+  void foo() {
```
benzonico: Isn't this missing a test case with a primitive ? 
int i;
long l;
true ? i : l; //Compliant ?

m-g-sonar: done.

benzonico:
LGTM!

PR #71, opened 2015-02-17
SONARJAVA-900 RSPEC-2440 Fix rule to ignore constructors

m-g-sonar:


benzonico:
LGTM! I don't like the hasOnlyStaticMethod method but I can't think of a really simpler way.

benzonico:
LGTM!

PR #72, opened 2015-02-17
exclude method names from check by regular expressions

fjakop:
We use a very old bytecode-enhancing technology (Kodo JDO) which produces getter and setter in our bytecode for intercepting access to persistent objects.
Some of the generated methods are not called (an they're private), so we get false positives with UnusedPrivateMethodCheck rule.
I added a string parameter to that rule which can hold a regular expression for method names to include (all our false positive methods match to '^jdo[GS]et\w+').
Test is included.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/UnusedPrivateMethodCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/UnusedPrivateMethodCheck.java
@@ -19,12 +19,14 @@
  */
 package org.sonar.java.checks;
 
-import com.google.common.base.Joiner;
-import com.google.common.collect.Lists;
+import java.util.List;
+import java.util.regex.Pattern;
+
 import org.apache.commons.lang.StringUtils;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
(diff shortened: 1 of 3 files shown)


benzonico:
Hi and thanks for your contribution. 

This rule uses the old BytecodeVisitor and therefore is soon to be replaced by our new way to write rules and use the semantic analysis. This was not done yet as we were lacking quite some type and method resolution yet. Once this is done, the rule will not rely on the method found in bytecode but on what is in the source code and thus you won't have to exclude things like this if you enhance your bytecode.

As a consequence, I am closing this pull request.

Side note though : aren't those generated method flagged as synthetic ?

fjakop:
No, unfortunately they are not. Maybe this is because our kodo-jdo implementation is of 2005?
When can we expect this new way to be available to the world, because we're in progress of migrating all our old findbugs, checkstyle, pmd maven profiles to sonar?

benzonico:
No commitment here, but I do hope to have this migrated for java 3.1




*Nicolas PERU | SonarSource*
*Senior Developer*
http://sonarsource.com
https://twitter.com/benzonico


2015-02-20 12:12 GMT+01:00 fjakop <notifications@github.com>:

PR #74, opened 2015-02-18
SONARJAVA-910 RSPEC-881 Fix fp on return statements

benzonico:


java-checks/src/main/java/org/sonar/java/checks/IncrementDecrementInSubExpressionCheck.java
```diff
@@ -51,31 +52,34 @@
...
     super.visitUnaryExpression(tree);
-
     if (isIncrementOrDecrement(tree)) {
       context.addIssue(tree, RULE, "Extract this increment or decrement operator into a dedicated statement.");
     }
   }
 
+  @Override
+  public void visitReturnStatement(ReturnStatementTree tree) {
+    if(tree.expression()== null || !isIncrementOrDecrement(tree.expression())) {
```
m-g-sonar: Maybe use a variable instead of calling 3 times `tree.expression()`

m-g-sonar:
LGTM!

PR #75, opened 2015-02-18
SONARJAVA-106 RSPEC-2301 Selector argument check

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/SelectorMethodArgumentCheck.java
```diff
@@ -0,0 +1,136 @@
...
+  public void visitNode(Tree tree) {
+    MethodTree methodTree = (MethodTree) tree;
+    List<VariableSymbol> booleanParameterSymbols = getBooleanParametersAsSymbol(methodTree.parameters());
+    BlockTree blockTree = methodTree.block();
+
+    if (isPublic(methodTree) && blockTree != null && !booleanParameterSymbols.isEmpty()) {
+      for (VariableSymbol variable : booleanParameterSymbols) {
+        Collection<IdentifierTree> usages = getSemanticModel().getUsages(variable);
+        if (usages.size() == 1) {
+          blockTree.accept(new ConditionnalStatementVisitor(variable.getName(), getSingleUsage(usages), tree));
```
benzonico: Typo : only one 'n' in conditional

java-checks/src/main/java/org/sonar/java/checks/SelectorMethodArgumentCheck.java
```diff
@@ -0,0 +1,136 @@
...
+    for (VariableTree variableTree : parameters) {
+      if (isBooleanVariable(variableTree)) {
+        booleanParameters.add(((VariableTreeImpl) variableTree).getSymbol());
+      }
+    }
+    return booleanParameters;
+  }
+
+  private IdentifierTree getSingleUsage(Collection<IdentifierTree> usages) {
+    return Iterables.get(usages, 0);
```
benzonico: In this case I would probably inline this method.

java-checks/src/main/java/org/sonar/java/checks/SelectorMethodArgumentCheck.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  private boolean isBooleanVariable(VariableTree variableTree) {
+    if (!hasSemantic()) {
```
benzonico: There is no point of even executing this rule if we don't have semantic, so this should be at the top of visitNode method.

java-checks/src/main/java/org/sonar/java/checks/SelectorMethodArgumentCheck.java
```diff
@@ -0,0 +1,136 @@
...
+
+  private boolean isPublic(MethodTree methodTree) {
+    return methodTree.modifiers().modifiers().contains(Modifier.PUBLIC);
+  }
+
+  private List<VariableSymbol> getBooleanParametersAsSymbol(List<VariableTree> parameters) {
+    List<VariableSymbol> booleanParameters = Lists.newLinkedList();
+    for (VariableTree variableTree : parameters) {
+      if (isBooleanVariable(variableTree)) {
+        booleanParameters.add(((VariableTreeImpl) variableTree).getSymbol());
```
benzonico: Design suggestion: The symbol could be extracted and the isBoolean could take the symbol as parameter.

java-checks/src/main/java/org/sonar/java/checks/SelectorMethodArgumentCheck.java
```diff
@@ -0,0 +1,136 @@
...
+      checkParameterUsage(tree.condition());
+    }
+
+    @Override
+    public void visitConditionalExpression(ConditionalExpressionTree tree) {
+      checkParameterUsage(tree.condition());
+    }
+
+    private void checkParameterUsage(ExpressionTree condition) {
+      if (condition.is(Kind.IDENTIFIER) && usage.equals((IdentifierTree) condition)) {
```
benzonico: no need to check for kind as you are using equals right after that.

java-checks/src/main/java/org/sonar/java/checks/SelectorMethodArgumentCheck.java
```diff
@@ -0,0 +1,136 @@
...
+        reportIssue();
```
benzonico: No need for this method. Inline it.

java-checks/src/test/files/checks/SelectorMethodArgumentCheck.java
```diff
@@ -0,0 +1,73 @@
...
+      }
+    }
+    return 0;
+  }
+
+  public int tuc(int a, boolean b) { // Noncompliant
+    return b ? a+1 : a-1;
+  }
+
+  private int hal(int a, boolean b) { // Compliant
```
benzonico: Maybe precise why this is compliant in the comment.

benzonico:
LGTM!

PR #76, opened 2015-02-18
SONARJAVA-913 RSPEC-2438 Runnable as parameter of thread check

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/RunnableAsThreadArgumentCheck.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  public List<Kind> nodesToVisit() {
+    return ImmutableList.of(Kind.NEW_CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    NewClassTree newClassTree = (NewClassTree) tree;
+    List<ExpressionTree> arguments = newClassTree.arguments();
+    if (arguments.size() > 0 && isOfTypeThread(newClassTree) && isSubTypeOfThread(arguments.get(0))) {
```
benzonico: please use !arguments.isEmpty()

java-checks/src/main/java/org/sonar/java/checks/RunnableAsThreadArgumentCheck.java
```diff
@@ -0,0 +1,88 @@
...
+  tags = {"multi-threading", "pitfall"},
+  priority = Priority.CRITICAL)
+@ActivatedByDefault
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.INSTRUCTION_RELIABILITY)
+@SqaleConstantRemediation("15min")
+public class RunnableAsThreadArgumentCheck extends SubscriptionBaseVisitor {
+
+  @Override
+  public List<Kind> nodesToVisit() {
+    return ImmutableList.of(Kind.NEW_CLASS);
```
benzonico: Why only new class ? This rule should also apply to method calls.

java-checks/src/main/java/org/sonar/java/checks/RunnableAsThreadArgumentCheck.java
benzonico: What if I have a method expecting a Thread as an argument and I pass a thread to it ? This will trigger an issue whereas it should not.

m-g-sonar:
Branch outdated. Please look at https://github.com/SonarSource/sonar-java/pull/82

PR #78, opened 2015-02-19
SONARJAVA-888 RSPEC-2160 Fix FP when equals is final in superclass

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverriddenInSubclassCheck.java
```diff
@@ -99,7 +99,7 @@ private boolean parentClassImplementsEquals(ClassTree tree) {
   private boolean hasEqualsMethod(TypeSymbol superClassSymbol) {
```
benzonico: please at least rename the method : hasNotFinalEqualsMethod.

benzonico:
LGTM!

PR #82, opened 2015-02-20
SONARJAVA 913 RSPEC 2438 Thead instead of runnable argument check

m-g-sonar:
Replace pull request https://github.com/SonarSource/sonar-java/pull/76


java-checks/src/main/java/org/sonar/java/checks/ThreadAsRunnableArgumentCheck.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+    // FIXME semantic - symbol should never be null!
+    if (!arguments.isEmpty() && methodSymbol != null && methodSymbol.isKind(Symbol.MTH)) {
+      checkArgumentsTypes(arguments, (MethodSymbol) methodSymbol);
+    }
+  }
+
+  private void checkArgumentsTypes(List<ExpressionTree> arguments, MethodSymbol methodSymbol) {
+    List<Type> parametersTypes = methodSymbol.getParametersTypes();
+    // FIXME semantic - if there is arguments and method symbol has been resolved, the list of parameters should not be empty!
```
benzonico: Can you produce a case ? why could the parametersTypes be empty ? this should lead to opening a jira ticket.

java-checks/src/main/java/org/sonar/java/checks/ThreadAsRunnableArgumentCheck.java
```diff
@@ -0,0 +1,118 @@
...
+    if (tree.is(Kind.NEW_CLASS)) {
+      NewClassTreeImpl nct = (NewClassTreeImpl) tree;
+      methodSymbol = getSemanticModel().getReference(nct.getConstructorIdentifier());
+      arguments = nct.arguments();
+    } else {
+      MethodInvocationTreeImpl mit = (MethodInvocationTreeImpl) tree;
+      methodSymbol = mit.getSymbol();
+      arguments = mit.arguments();
+    }
+    // FIXME semantic - symbol should never be null!
```
benzonico: Better to precise a case in which it could be null

benzonico:
LGTM! but please create requested jira tickets for emptyness or nullness of symbols.

PR #84, opened 2015-02-20
SONARJAVA-901 Fix ParametrizedTypeType tag when unknown rawType

m-g-sonar:


java-squid/src/main/java/org/sonar/java/resolve/Type.java
```diff
@@ -278,6 +278,10 @@ public Type erasure() {
       super(symbol);
       this.rawType = symbol.getType();
       this.typeSubstitution = typeSubstitution;
+      if (rawType.isTagged(UNKNOWN)) {
```
benzonico: I don't think this should be done at this level. I tend to think that when constructing a ParametrizedTypeType using the ParametrizedTypeCache we should return the unknown type altogether rather than an unknown ParametrizedTypeType which can (IMO) lead to confusion.

benzonico:
Close this one in favor of : https://github.com/SonarSource/sonar-java/pull/87

PR #86, opened 2015-02-23
SONARJAVA-914 fix fp

benzonico:
Exclude cases with no ambiguity (no method matching name in outer class) and cases where innerClass extends OuterClass


m-g-sonar:
LGTM!

PR #89, opened 2015-03-06
Explained limitation of current calculation

sat4j:
The current calculation does not take into account empty lines added within the case statements.

This is not a real pull request because I do not provide the code to fix this. I use the PR mechanism because I found no other way (JIRA, GitHub issues) to point out that issue.


benzonico:
Please use mailing list rather than PR for this kind of things : http://www.sonarqube.org/get-support/

PR #91, opened 2015-03-10
TypeAndReferenceSolver: extend resolution of annotations to fields and m...

merciesa:


java-squid/src/main/java/org/sonar/java/resolve/AnnotationInstance.java
```diff
@@ -37,6 +37,10 @@ public void addValue(AnnotationValue annotationValue) {
     values.add(annotationValue);
   }
 
+  public Symbol.TypeSymbol getTypeSymbol() {
```
benzonico: @merciesa not sure this method is actually useful, for now requires at least an @visibleForTesting

benzonico:
In this very case the test won't be less precise at all. Let's keep things encapsulated for now and see how it goes.

benzonico:
LGTM : please create a ticket to associate to this commit and schedule it for version 3.1.

PR #94, opened 2015-03-12
Expression type api

benzonico:


java-squid/src/main/java/org/sonar/java/resolve/Type.java
```diff
@@ -29,7 +29,7 @@
 import java.util.Map;
 import java.util.Set;
 
-public class Type {
+public class Type implements org.sonar.plugins.java.api.semantic.Type {
```
m-g-sonar: I'm not really happy about having the same class name for implementation of the new semantic interface.

m-g-sonar:
Except the Type implementation class having the same name, everything LGTM!

PR #96, opened 2015-03-13
SONARJAVA-928 RSPEC-2259 Null pointers should not be dereferenced

merciesa:
Add check for null arguments in method invocation

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
@@ -27,14 +27,18 @@ import org.sonar.java.model.expression.MethodInvocationTreeImpl;
 import org.sonar.java.resolve.AnnotationInstance;
 import org.sonar.java.resolve.SemanticModel;
 import org.sonar.java.resolve.Symbol;
+import org.sonar.java.resolve.Symbol.MethodSymbol;
 import org.sonar.plugins.java.api.JavaFileScanner;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
 import org.sonar.plugins.java.api.tree.ArrayAccessExpressionTree;
 import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -72,13 +76,26 @@ public void visitMemberSelectExpression(MemberSelectExpressionTree tree) {
...
   }
 
+  @Override
+  public void visitMethodInvocation(MethodInvocationTree tree) {
+    Symbol symbol = ((MethodInvocationTreeImpl) tree).getSymbol();
+    if (symbol.isKind(Symbol.MTH)) {
+      MethodSymbol methodSymbol = (MethodSymbol) ((MethodInvocationTreeImpl) tree).getSymbol();
+      for (int i = 0; i < tree.arguments().size(); i += 1) {
+        List<Symbol> parameters = methodSymbol.getParameters().scopeSymbols();
+        if (!canBeNull(parameters.get(i < parameters.size() ? i : parameters.size() - 1))) {
```
benzonico: I would probably not rely on a ternary operator in this case, just to make it clear for further reading that you are doing so to handle problems of varargs (at the very least a small commentary could be nice)

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -72,13 +76,26 @@ public void visitMemberSelectExpression(MemberSelectExpressionTree tree) {
     super.visitMemberSelectExpression(tree);
   }
 
+  @Override
+  public void visitMethodInvocation(MethodInvocationTree tree) {
+    Symbol symbol = ((MethodInvocationTreeImpl) tree).getSymbol();
+    if (symbol.isKind(Symbol.MTH)) {
+      MethodSymbol methodSymbol = (MethodSymbol) ((MethodInvocationTreeImpl) tree).getSymbol();
```
benzonico: Why not cast symbol variable directly ?

benzonico:
LGTM except for some minor issue. see comments.

PR #99, opened 2015-03-16
don't mark lombok's Getter/Setter as unused private field

liudongmiao:
See https://jira.codehaus.org/browse/SONARJAVA-73

http://projectlombok.org/ introduce `Getter/Setter`, which simplify `getter/setter` use annotation. This patch won't mark these annotation as unsed private field.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/UnusedPrivateFieldCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/UnusedPrivateFieldCheck.java
@@ -126,8 +126,29 @@ public class UnusedPrivateFieldCheck extends SubscriptionBaseVisitor {
     }
   }
 
+  private boolean isLombok(IdentifierTree tree) {
+    String name = tree.name();
+    return ("Getter".equals(name) || "Setter".equals(name));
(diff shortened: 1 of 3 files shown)


benzonico:
Hi and thanks for your contribution, 

As you have noticed, the check is already taking into account lombok `Getter` as per (https://jira.codehaus.org/browse/SONARJAVA-73 which is fixed by https://github.com/SonarSource/sonar-java/commit/1605af75f648e372a126dcb7b3991505398acb7c). If you want to add support for lombok `Setter` you should first make clear that this is your intent, have a closer look at what is already supported, and probably reuse what's already in place instead of adding your own logic to detect lombok `Getter` and `Setter` annotations (which can be flawed as it relies solely on identifiers and not on symbol table).

In the current state, I cannot accept this contribution as you duplicate a functionality.

Thanks again.

liudongmiao:
Your response is so quick...I just want to close this pull request.

liudongmiao:
I just find Getter/Setter is not enouch, and there are other annotation.
So I think, it would be a better solution, use the "delombok".

PR #100, opened 2015-03-16
Sonarjava 928

merciesa:
I created a dedicated commit for the comments.
Added:
- conditional (ternary) expressions
- merge of contexts in case of if and conditional expressions

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
@@ -70,7 +70,6 @@ public class NullPointerCheck extends BaseTreeVisitor implements JavaFileScanner
   public static final String KEY = "S2259";
   private static final RuleKey RULE_KEY = RuleKey.of(CheckList.REPOSITORY_KEY, KEY);
 
-  private final ConditionVisitor conditionVisitor = new ConditionVisitor();
   private JavaFileScannerContext context;
   private State currentState;
   private SemanticModel semanticModel;
@@ -122,7 +121,7 @@ public class NullPointerCheck extends BaseTreeVisitor implements JavaFileScanner
   public void visitConditionalExpression(ConditionalExpressionTree tree) {
     State trueState = new State(currentState);
     State falseState = new State(currentState);
-    conditionVisitor.visitCondition(tree.condition(), trueState, falseState);
+    visitCondition(tree.condition(), trueState, falseState);
     currentState = trueState;
     tree.trueExpression().accept(this);
     currentState = falseState;
@@ -134,7 +133,7 @@ public class NullPointerCheck extends BaseTreeVisitor implements JavaFileScanner
   public void visitIfStatement(IfStatementTree tree) {
     State trueState = new State(currentState);
     State falseState = tree.elseStatement() != null ? new State(currentState) : null;
-    conditionVisitor.visitCondition(tree.condition(), trueState, falseState);
+    visitCondition(tree.condition(), trueState, falseState);
     currentState = trueState;
     tree.thenStatement().accept(this);
     if (tree.elseStatement() != null) {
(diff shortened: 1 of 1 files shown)


benzonico:
A small description of the PR could be nice :)

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -71,12 +158,64 @@ public void visitArrayAccessExpression(ArrayAccessExpressionTree tree) {
...
+      currentState.setVariableValue(identifierSymbol, value);
+    }
+    super.visitAssignmentExpression(tree);
+  }
+
+  @Override
+  public void visitClass(ClassTree tree) {
+    // state required for assignments in class body (e.g. static initializers, and int a, b, c = b = 0;)
+    State oldState = currentState;
+    currentState = new State(null);
```
benzonico: what about a constructor with no parameters rather than passing null value explicitely ?

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -52,7 +65,81 @@
...
+
+    @Override
+    public void visitBinaryExpression(BinaryExpressionTree tree) {
+      Symbol identifierSymbol;
+      if (tree.leftOperand().is(Tree.Kind.NULL_LITERAL) && tree.rightOperand().is(Tree.Kind.IDENTIFIER)) {
+        identifierSymbol = semanticModel.getReference((IdentifierTreeImpl) tree.rightOperand());
+      } else if (tree.leftOperand().is(Tree.Kind.IDENTIFIER) && tree.rightOperand().is(Tree.Kind.NULL_LITERAL)) {
+        identifierSymbol = semanticModel.getReference((IdentifierTreeImpl) tree.leftOperand());
+      } else {
+        return;
```
benzonico: Might worth to specify via the tests that a case like `if(a==null && b==null)` won't be covered.

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -52,7 +65,81 @@
...
+  public enum AbstractValue {
+    // value is known to be not null.
+    NOTNULL,
+    // value is known to be null.
+    NULL,
+    // value is unknown (could be null or not null).
+    UNKNOWN
+  }
+
+  private class ConditionVisitor extends BaseTreeVisitor {
```
benzonico: private class and enum should be at end of file for readability matters.

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -52,7 +65,81 @@
...
+        Preconditions.checkState(tree.is(Tree.Kind.NOT_EQUAL_TO));
+        trueState.setVariableValue(identifierSymbol, AbstractValue.NOTNULL);
+        if (falseState != null) {
+          falseState.setVariableValue(identifierSymbol, AbstractValue.NULL);
+        }
+      }
+    }
+  }
+
+  public static class State {
```
benzonico: why public ?

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -52,7 +65,81 @@
...
+        AbstractValue result = variables.get(variable);
+        if (result != null) {
+          return result;
+        }
+      }
+      return AbstractValue.UNKNOWN;
+    }
+
+    // sets the value of the variable in the current state.
+    public void setVariableValue(Symbol variable, AbstractValue value) {
```
benzonico: why not a VariableSymbol ?

merciesa: fixed

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
merciesa: changed to private

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
merciesa: ok

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
merciesa: ok

benzonico:
LGTM!

PR #101, opened 2015-03-17
SONARJAVA-932 RSPEC-2696 Static field update check

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/StaticFieldUpateCheck.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
+      return;
+    }
+
+    ClassTree classTree = (ClassTree) tree;
+    if (isNonStatic(classTree)) {
```
benzonico: Why the static of the class matters here ?

benzonico:
LGTM: see comments.

PR #103, opened 2015-03-17
SONARJAVA-921 Look for methods settings credentials

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/HardCodedCredentialsCheck.java
```diff
@@ -68,12 +70,34 @@ public void visitNode(Tree tree) {
...
+    return arguments.size() == 2 && argumentsAreLiterals(arguments) && isPassword((LiteralTree) arguments.get(0));
+  }
+
+  private boolean isPassword(LiteralTree argument) {
+    return argument.is(Tree.Kind.STRING_LITERAL) && PASSWORD_VARIABLE_PATTERN.matcher(LiteralUtils.trimQuotes(argument.value())).matches();
+  }
+
+  private boolean argumentsAreLiterals(List<ExpressionTree> arguments) {
+    for (ExpressionTree argument : arguments) {
+      if (!(argument instanceof LiteralTree)) {
```
benzonico: You should not use InstanceOf but Tree.Kinds and in this very case STRING_LITERAL

benzonico:
LGTM: see comment.

PR #105, opened 2015-03-17
SONARJAVA-933 Fix bitwise operands evaluation

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -90,27 +89,15 @@ public void visitNode(Tree tree) {
   }
 
   private boolean isZeroMaskShift(BinaryExpressionTree binaryExpressionTree) {
-    return isLiteralValue(binaryExpressionTree.leftOperand(), 1) && isLiteralValue(binaryExpressionTree.rightOperand(), 0);
+    return hasValue(binaryExpressionTree.leftOperand(), 1) && hasValue(binaryExpressionTree.rightOperand(), 0);
```
benzonico: int Promoted to long, better with the L explicitely.

java-checks/src/main/java/org/sonar/java/checks/ShiftOnIntOrLongCheck.java
```diff
@@ -90,27 +89,15 @@ public void visitNode(Tree tree) {
...
-        // is necessarily not equals.
-        // Note that Long.MAX_VALUE = "0x7FFF_FFFF_FFFF_FFFFL", but it is possible to write larger numbers in hexadecimal
-        // to be used as mask in bitwise operation. For instance:
-        // 0x8000_0000_0000_0000L (MAX_VALUE + 1),
-        // 0xFFFF_FFFF_FFFF_FFFFL (only ones),
-        // 0xFFFF_FFFF_FFFF_FFFEL (only ones except least significant bit), ...
-      }
-    }
-    return false;
+  private boolean hasValue(ExpressionTree tree, long value) {
```
benzonico: name is not so great: this check that tree is a literal equal to the correct value.

benzonico:
LGTM ! see small comments.

PR #106, opened 2015-03-17
CompareObjectWithEqualsCheck: allow exclusion of class types

nrainer:
The possibility to exclude certain class types from the comparison would be highly appreciated.

We have certain classes in our project for which a comparison with == and a comparison with the equals method gives the same result (e.g. key classes which are implemented as singletons per value). We prefer to compare instances of these classes with == (reasons are avoidance of NPEs and better readability) and have a custom SonarQube rule to ensure the adherence to this convention. However, our custom rule is understandably in conflict with the CompareObjectWithEqualsCheck for the these classes. As we would like to keep the "compare with equals"-check, a way to exclude certain class types would be desirable.

This pull request introduces a new rule parameter named "exclusions". The string parameter holds the fully qualified names of classes separated by ";". It is used to check if comparisons shall be skipped depending on the type. Currently, only comparisons with java.lang.Class are skipped.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CompareObjectWithEqualsCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CompareObjectWithEqualsCheck.java
@@ -61,26 +61,17 @@ public class CompareObjectWithEqualsCheck extends BaseTreeVisitor implements Jav
   @RuleProperty(
       key = "exclusions",
       defaultValue = "" + DEFAULT_EXCLUSIONS)
-  private String exclusions;
+  private String exclusions = DEFAULT_EXCLUSIONS;
   
   private Set<String> exclusionSet;
   
-  public CompareObjectWithEqualsCheck() {
-	  exclusionSet = new HashSet<String>();
-	  setExclusions(DEFAULT_EXCLUSIONS);
-  }
-
-  public void setExclusions(String exclusions) {
-	  this.exclusions = exclusions;
-	  exclusionSet.clear();
-	  
-	  if (exclusions != null) {
+  private synchronized Set<String> getExclusionSet() {
+	  if (exclusionSet == null) {
+		  exclusionSet = new HashSet<String>();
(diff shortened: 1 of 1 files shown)


benzonico:
Hi and thanks for your contribution.

We are not very keen on adding rule parameters without a very good reason. 
Here I definitely cannot spot why this should not be handled with equals (readability is a matter of opinion and as for NPE, if this is singleton then I don't see why this should be an issue). So we are not in favor of implementing this feature.

On a technical side note : there is absolutely no test in your pull request (which would have make me reject it) and (small detail but still) you use  ';' as a separator whereas every rule parameter uses ','.

As a consequence of all the previous I am closing this.

Once again, thanks for your contribution.

Cheers,

nrainer:
Hi and thank you for your response.

I agree with you regarding the technical side. I will change the character of the parameter separator to the default one. Moreover, I gladly will add test cases to ensure that the changed logic is working as expected.

I also agree with you that readability is a matter of opinion and do understand that you prefer not to add new rule parameters if not really necessary.
Though, I am wondering why the rule allows the comparison of instances of ``java.lang.Class`` using == without raising an issue... The same applies to enums. UIDs, in fact, are kind of enum values.
The difference to the enum construct is that the CodeType construct (provided by Eclipse Scout) (which in our project consists of a set of UIDs) can be subclassed.
NullPointerExceptions can still occur because we do not only compare the declared values but also other fields that may contain any UID or be null.

The following code might clarify the use of UIDs:

``if (person.getSalutationUid() == SalutationCodeType.MrCode.UID) {
 //...
}``

``//may throw an NPE if compared with equals
if (communication1.getChannelUid() == communication2.getChannelUid() {
}``

I hope I was able to convince you why this change would be appreciated by us. Does my pull request have a chance if I incorporate your technical comments or will it still be refused?
Please let me know.

Thank you and best regards,
Rainer

benzonico:
Hi, can you please resume this discussion and forward it to debate on the user mailing list. (http://www.sonarqube.org/get-support/ point 2) 

This will allow the community to weigh in one way or another. 

Thanks.

PR #109, opened 2015-03-18
SONARJAVA-928 RSPEC-2259 implementation of SONARJAVA-945

merciesa:
Fix false positive: constraints in subexpressions
visitCondition, visitorConditionalAnd and visitConditionalOr are utterly hard to understand. Feel free to suggest a better implementation

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
@@ -20,7 +20,6 @@
 package org.sonar.java.checks;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import org.sonar.api.rule.RuleKey;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
@@ -43,6 +42,7 @@ import org.sonar.plugins.java.api.tree.ClassTree;
 import org.sonar.plugins.java.api.tree.CompilationUnitTree;
 import org.sonar.plugins.java.api.tree.ConditionalExpressionTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.ForStatementTree;
 import org.sonar.plugins.java.api.tree.IfStatementTree;
 import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodInvocationTree;
@@ -50,6 +50,7 @@ import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.TryStatementTree;
 import org.sonar.plugins.java.api.tree.VariableTree;
+import org.sonar.plugins.java.api.tree.WhileStatementTree;
 import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
 import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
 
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -208,6 +235,14 @@ public void visitVariable(VariableTree tree) {
     }
   }
 
+  @Override
+  public void visitWhileStatement(WhileStatementTree tree) {
+    ConditionalState conditionalState = visitCondition(tree.condition());
+    currentState = conditionalState.trueState;
+    scan(tree.statement());
+    currentState = currentState.parentState.merge(conditionalState.trueState, conditionalState.falseState);
```
benzonico: aren't we gonna end up with no valuable information with a merge like this one ?

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -269,34 +304,97 @@ private void checkForIssue(Tree tree) {
...
+    if (currentConditionalState != null) {
+      // merges the information for the false state.
+      // the false state of the right operand also contains the information from the true state of the left operand.
+      // first, discards the information learned from the left operand.
+      leftConditionalState.falseState.merge(leftConditionalState.trueState, null);
+      leftConditionalState.falseState.merge(rightConditionalState.falseState, null);
+      // merges the information for the true state.
+      leftConditionalState.trueState.merge(rightConditionalState.trueState, null);
+      // applies changes in the parent state.
+      currentConditionalState.falseState.apply(leftConditionalState.falseState);
```
benzonico: This should probably be part of ConditionalState.

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
merciesa: if a=null before the loop, and a!=null in the condition, then the merge will set the value of a to UNKNOWN after the loop, which is valuable. I'll add a comment in that sense.

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -331,29 +441,33 @@ public void setVariableValue(VariableSymbol variable, AbstractValue value) {
       variables.put(variable, value);
     }
 
-    public State merge(State trueState, @Nullable State falseState) {
+    // copies the values from state to this.
+    public State apply(State state) {
```
benzonico: As discussed this should be renamed to "copieValuesFrom"

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -331,29 +441,33 @@ public void setVariableValue(VariableSymbol variable, AbstractValue value) {
...
+      for (VariableSymbol variable : state.variables.keySet()) {
+        this.setVariableValue(variable, state.getVariableValue(variable));
+      }
+      return this;
+    }
+
+    // merges the values from state1 and state2 (or this if state2 is null) into this.
```
benzonico: Comment should be JavaDoc.

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -269,34 +308,105 @@ private void checkForIssue(Tree tree) {
...
+      falseState = new State(currentState);
+      trueState = new State(currentState);
+    }
+
+    void mergeConditionalAnd(ConditionalState leftConditionalState, ConditionalState rightConditionalState) {
+      // merges the information for the false state.
+      // the false state of the right operand also contains the information from the true state of the left operand.
+      // first, discards the information learned from the left operand.
+      leftConditionalState.falseState.merge(leftConditionalState.trueState, null);
+      leftConditionalState.falseState.merge(rightConditionalState.falseState, null);
```
benzonico: I am really doubtful about this "merge" in fact. 

in the case of something like : 
```java
if(a !=null && b!=null) {
//...
}  else { 
   a.foo(); //we don't raise issue here because we cancel constraints on "a" (which is fine)
   b.foo(); //shall we raise an issue here ? we have no idea which path  was taken and so it can b that b is perfectly fine, case is not different from a.
}
```

benzonico:
LGTM! Commit from second review provides way better readability).

PR #110, opened 2015-03-18
SONARJAVA-947 RSPEC-2674 InputStream return value check

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/IgnoredStreamReturnValueCheck.java
```diff
@@ -0,0 +1,88 @@
...
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
+      return;
+    }
+
+    ExpressionTree statement = ((ExpressionStatementTree) tree).expression();
+    if (statement.is(Kind.METHOD_INVOCATION)) {
+      ExpressionTree methodSelect = ((MethodInvocationTree) statement).methodSelect();
+      if (methodSelect.is(Kind.MEMBER_SELECT)) {
+        MethodSymbol symbol = (MethodSymbol) getSemanticModel().getReference(((MemberSelectExpressionTree) methodSelect).identifier());
```
benzonico: You should not need to go through getSemanticModel but you can use directly the getSymbol from MethodInvocationTreeImpl (and even the upcoming symbol() method from the interface).

m-g-sonar:
PR outdated.

PR #113, opened 2015-03-20
SONARJAVA-928 RSPEC-2259 implementation of SONARJAVA-949

merciesa:
Fix false positive: constraints before a loop should be invalidated

### DIFF
--- a/java-checks/src/test/files/checks/NullPointerCheck.java
+++ b/java-checks/src/test/files/checks/NullPointerCheck.java
@@ -258,18 +258,68 @@ class NullPointerTest {
     boolean b2 = str != null || str.length() == 0; // Noncompliant
   }
 
+  public void testDoWhileLoop(boolean condition) {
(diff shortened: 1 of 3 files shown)


java-checks/src/test/files/checks/NullPointerCheck.java
```diff
@@ -258,18 +258,68 @@ public void testLogicalOr(String str) {
...
+      object2.hashCode(); // Noncompliant
+     }
+    object1.hashCode(); // False negative
+    object2.hashCode(); // Compliant
+    object2.hashCode(); // Compliant
+  }
+
+  public void testHoistedLoop() {
+    Object a = null, b = null;
+    if (b == null) {
```
benzonico: typo ? should be b ?

benzonico:
Otherwise looks good to me. The test cases looks really better with a clear separation by method.

PR #114, opened 2015-03-20
Sonarjava 928

merciesa:
implementation of SONARJAVA-951
    Fix false positive: switch should be bypassed

implementation of SONARJAVA-950
    Fix false positive: issue should be reported only on local variables, but not on instance or class fields


benzonico:
LGTM! See two comments on styling the condition.

PR #115, opened 2015-03-23
SONARJAVA-928 RSPEC-2259 implementation of SONARJAVA-954

merciesa:
Fix false positive: constraints on method parameters should not be ignored when set to unknown


benzonico:
LGTM! Please update commit title (not comment) to mention you are introducing the UNSET value.

PR #117, opened 2015-03-23
SONARJAVA-928 SONARJAVA-941

merciesa:
Improve constraint tracking in visitAssignmentExpression to match visitVariable

### DIFF
--- a/java-checks/src/test/files/checks/NullPointerCheck.java
+++ b/java-checks/src/test/files/checks/NullPointerCheck.java
@@ -336,13 +336,21 @@ class NullPointerTest {
     str.length(); // False negative
   }
 
-  public void test(@Nullable Object o) {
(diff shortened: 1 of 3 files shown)


java-checks/src/test/files/checks/NullPointerCheck.java
```diff
@@ -345,6 +345,14 @@ public void test(@Nullable Object o) {
     a.hashCode(); // Compliant
   }
 
+  public void testAssignNullableMethod() {
+    Object object;
+    object = nullableMethod();
+    if(object.hashCode()); // issue
```
benzonico: ouch! the  ; for the empty assignement is really less readable than an empty block IMO.

benzonico:
LGTM

PR #118, opened 2015-03-23
SONARJAVA-928 RSPEC-2259 SONARJAVA-949

merciesa:
Invalidate only the symbols updated in loops

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
@@ -154,10 +154,10 @@ public class NullPointerCheck extends BaseTreeVisitor implements JavaFileScanner
 
   @Override
   public void visitDoWhileStatement(DoWhileStatementTree tree) {
-    currentState.invalidateValuesOfHierarchy();
+    currentState.invalidateVariables(new AssignmentVisitor().findAssignedVariables(tree.statement()));
     currentState = new State(currentState);
     scan(tree.statement());
-    visitCondition(tree.condition());
+    scan(tree.condition());
     restorePreviousState();
   }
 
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -409,6 +414,32 @@ private void visitRelationalNotEqualTo(BinaryExpressionTree tree) {
...
+      for (Tree tree : trees) {
+        tree.accept(this);
+      }
+      return assignedSymbols;
+    }
+
+    @Override
+    public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+      // currently we only handle assignment of the form: identifier =
+      // TODO(merciesa): array[expr] = and mse.identifier = requires advanced tracking and are left outside of the scope
```
benzonico: mse.identifier is not a local variable so it is outside of scope. 
However array[index] = ... ; can easily be tracked. But for me this is not relevant for what we are currently doing : array should be initialized alltogether or array access will raise NPE. 
Therefore, I think this comment is not relevant here.

java-checks/src/main/java/org/sonar/java/checks/NullPointerCheck.java
```diff
@@ -409,6 +414,32 @@ private void visitRelationalNotEqualTo(BinaryExpressionTree tree) {
...
+      if (tree.variable().is(Tree.Kind.IDENTIFIER)) {
+        assignedSymbols.add((VariableSymbol) semanticModel.getReference((IdentifierTreeImpl) tree.variable()));
```
benzonico: What if the variable is (for some reason) not resolved ? you will add null to the set, how will your code react to this ? You should have a testcase for this.

java-checks/src/test/files/checks/NullPointerCheck.java
```diff
@@ -353,6 +354,48 @@ public void testAssignNullableMethod() {
...
   }
 
+  public void testComplexLoop(@Nullable Object nullableObject) {
+    Object object1 = null, object11 = null, object12 = null;
+    for(int i = 0; object11 == null; i += 1) {
+      object11.hashCode(); // False negative
+      object12.hashCode(); // Noncompliant
+      nullableObject.hashCode(); // Noncompliant
+      if(i == 1) {
+        object1.hashCode(); // False negative
```
benzonico: This is not a false negative.

java-checks/src/test/files/checks/NullPointerCheck.java
```diff
@@ -353,6 +354,48 @@ public void testAssignNullableMethod() {
...
+    object1.hashCode(); // False negative
+
+    Object object2 = null, object21 = null, object22 = null;
+    int i = 0;
+    while(object21 == null) {
+      object21.hashCode(); // False negative
+      object22.hashCode(); // Noncompliant
+      nullableObject.hashCode(); // Noncompliant
+      if(i == 1) {
+        object2.hashCode(); // False negative
```
benzonico: This is not a false negative.

benzonico:
See comments.

benzonico:
LGTM

PR #119, opened 2015-03-23
SONARJAVA-956 Refactor check to use AbstractMethodDetection

m-g-sonar:
Refactor from `BytecodeVisitor` to `AbstractMethodDetection`. 
Note that check only looks at method invocations from classes having `java.lang.Runnable` as super-type, as `java.lang.Thread` implements `java.lang.Runnable`.


benzonico:
Cool ! LGTM

PR #121, opened 2015-03-24
Sonarjava 928

merciesa:
Set constraints after assignment instead of before. this prevents null constraints from being propagated too early:
e.g. current = current.nullableParent; // False positive

Remove propagation of constraints from instance and static fields.
e.g.
local = nullableField
local.field; // False positive


benzonico:
LGTM

PR #124, opened 2015-03-24
SONARJAVA-957 RSPEC-2695 PreparedStatement and ResultSet check

m-g-sonar:
The PR is in two parts:
- First part improves `MethodInvocationMatcher` by adding new features:
  - It is now possible to set a (new) `NameCriteria` to match method name, instead of only using exact name. Currently provided `NameCriteria` handle the following situations: 
    - exact name
    - prefix  (which can be used to target getter or setter for instance) 
  - Parameters can now be handled using `TypeCriteria`, allowing for instance to target methods having a parameter being a sub-type of a given fully qualified name.
  - A new `TypeCriteria` have been introduced, which corresponds to "*any Type*". It can be used as a wildcard for a method parameter.
- Second part adds new check verifying usages of `PreparedStatement` and `ResultSet` methods. It implements [RSPEC-2695] (http://jira.sonarsource.com/browse/RSPEC-2695)

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CheckList.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CheckList.java
@@ -304,7 +304,8 @@ public final class CheckList {
       NullPointerCheck.class,
       StaticFieldUpateCheck.class,
       IgnoredStreamReturnValueCheck.class,
-      DateUtilsTruncateCheck.class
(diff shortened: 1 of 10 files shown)


m-g-sonar:
As discussed, NameCriteria has been updated to only use exact name and prefix.

benzonico:
LGTM!

PR #125, opened 2015-03-25
SONARJAVA-515 Rule: "Object" should not be extended

merciesa:
I will review the title of the rule and description with Ann (extends list sounds a bit strange to me)

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/UselessExtendsCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/UselessExtendsCheck.java
@@ -19,17 +19,19 @@
  */
 package org.sonar.java.checks;
 
-import org.sonar.api.rule.RuleKey;
+import com.google.common.collect.ImmutableList;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
 import org.sonar.plugins.java.api.JavaFileScanner;
-import org.sonar.plugins.java.api.JavaFileScannerContext;
-import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.Tree.Kind;
 import org.sonar.plugins.java.api.tree.TypeTree;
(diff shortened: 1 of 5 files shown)


java-checks/src/main/java/org/sonar/java/checks/UselessExtendsCheck.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    this.context = context;
+    scan(context.getTree());
+  }
+
+  @Override
+  public void visitClass(ClassTree tree) {
+    TypeTree superClass = tree.superClass();
+    if (superClass != null && "java.lang.Object".equals(superClass.symbolType().fullyQualifiedName())) {
```
benzonico: You are better off to use superClass.symbolType.is("java.lang.Object").

java-checks/src/test/files/checks/UselessExtendsCheck.java
```diff
@@ -0,0 +1,12 @@
...
+
+  public class Class1 { // Compliant
+  }
+
+  public class Class2 extends Object { // Noncompliant
+  }
+
+}
+
+public class Class3 extends java.lang.Object { // Noncompliant
```
benzonico: Lack a testcase where the symbol in the extend clause is not known (just use a dummy name)

java-checks/src/main/java/org/sonar/java/checks/UselessExtendsCheck.java
```diff
@@ -0,0 +1,62 @@
...
+import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
+
+@Rule(
+  key = UselessExtendsCheck.RULE_KEY,
+  name = "extends list entries should not be redundant",
+  tags = {"clumsy"},
+  priority = Priority.MINOR)
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.UNDERSTANDABILITY)
+@SqaleConstantRemediation("1min")
+public class UselessExtendsCheck extends BaseTreeVisitor implements JavaFileScanner {
```
benzonico: Why not a subscription visitor ? Really more appropriate in this case IMO : you will only visit the Class and Interface (skipping enum and annotations alltogether) 
Moreover, it will let you remove some boiler plate code still required for BaseTreeVisitor.

benzonico:
LGTM

PR #127, opened 2015-03-25
Sonarjava 515

merciesa:
Added redundant interface in list (e.g. implements I1, I1), commit 6cbbfc1
Added redundant interface in hierarchy, commit eb81a88


benzonico:
The package org.sonar.java.resolve should not be used in checks. Please update your code accordingly. If you have a usecase not covered, let me know about it.

PR #128, opened 2015-03-25
SONARJAVA-923 RSPEC-2112 URL hashCode and equals check

m-g-sonar:


benzonico:
LGTM! I really have to do something about those TypeParameter.

PR #133, opened 2015-03-26
SONARJAVA-515 

merciesa:
Fix false positive in case two or more interfaces are not resolved properly


java-checks/src/test/files/checks/UselessExtendsCheck.java
```diff
@@ -25,7 +25,7 @@
 implements I2 {
 }
 
-class Class4 extends UnknownClass implements UnknownInterface { // Compliant
+class Class4 extends UnknownClass1 implements UnknownInterface1, UnknownInterface2 { // Compliant
```
benzonico: What about : 
```
class Class4 implements UnknownInterface1, UnknownInterface1 //Same identifier, we don't know about type but in this case issue should probably be raised ?
```

benzonico:
LGTM!

PR #135, opened 2015-03-27
SONARJAVA-927 Annotations on enum constant is not available in SyntaxTree

merciesa:


benzonico:
Should I review it already ?

merciesa:
Yes please, apart from the commit message ;)

merciesa:
This is indeed an interesting case. There are two constructs, AnnotationTree and ModifierTree.

from the types: AnnotationTree extends ModifierTree
but: List<AnnotationTree> and List<ModifierTree> are unrelated

from the grammar: EnumConstant can only have a list of annotations (not modifiers)
Still, annotations in EnumConstant are represented by ModifiersTree.

Hence, a cast is required because ModifiersTree expects a List<ModifierTree> and not a List<AnnotationTree>.

The only risk here is if a ModifierTree is added to the list and accessed with the original type. But in practice this should not happen, because the original type is lost during construction of ModifiersTree.

merciesa:
But we could make a copy if you feel safer:
ModifiersTreeImpl modifiers = modifiers(annotations.isPresent() ? Optional.<List<ModifierTree>>of(new ArrayList(annotations.get())) : Optional.<List<ModifierTree>>absent());

EnumConstantTreeImpl result = new EnumConstantTreeImpl(modifiers, identifier, newClass);

benzonico:
Let's go for this one, but that would probably require some refactoring at some point.

PR #136, opened 2015-03-30
SONARJAVA-953 Fix FP with too many lines

m-g-sonar:
Line of the case statement is not counted anymore when it is not followed by trivias or statement.


benzonico:
LGTM!

PR #138, opened 2015-03-30
SONARJAVA-971 Fix BytecodeCompleter to add unknown superclass to unknown classes

m-g-sonar:


benzonico:
See minor comments. Otherwise LGTM!

PR #141, opened 2015-04-02
SONARJAVA-966 SONARJAVA-981 Support symbolic values on boolean variables

merciesa:


benzonico:
LGTM! I did not fully reviewed the functional side of this, i let this rely on @fmallet shoulders.

PR #146, opened 2015-04-10
SONARJAVA 989 Limit the number of execution states per method

merciesa:


benzonico:
As discussed, I am closing this PR in order to have another approach based on merge of states more suitable for this check.

PR #148, opened 2015-04-14
SONARJAVA-966 Fix false positive: invalidate constraint on assignments and prefix/postfix increment/decrement

merciesa:


java-checks/src/main/java/org/sonar/java/symexec/ExecutionState.java
```diff
@@ -220,4 +223,29 @@ private void mergeBooleanConstraints(Iterable<ExecutionState> states) {
...
+      findRelationPairs(current, pairs);
+    }
+    for (Map.Entry<Symbol.VariableSymbol, Symbol.VariableSymbol> cell : pairs.entries()) {
+      if (getRelation(cell.getKey(), cell.getValue()) != SymbolicRelation.UNKNOWN) {
+        setRelation(cell.getKey(), SymbolicRelation.UNKNOWN, cell.getValue());
+      }
+    }
+  }
+
+  private static void findRelationPairs(ExecutionState state, Multimap<Symbol.VariableSymbol, Symbol.VariableSymbol> pairs) {
```
benzonico: IMO this should return a Multimap that should be put in result using putAll method. Hence it will not modify input parameter (which will be only an execution state) and you will have a pure function.

benzonico: And so this method could actually be part of the ExecutionState.

java-checks/src/main/java/org/sonar/java/symexec/ExecutionState.java
```diff
@@ -220,4 +223,29 @@ private void mergeBooleanConstraints(Iterable<ExecutionState> states) {
     return result;
   }
 
+  void invalidateRelationsOnSymbol(VariableSymbol symbol) {
```
benzonico: If I am not mistaking, the parameter is unused.

java-checks/src/main/java/org/sonar/java/symexec/ExecutionState.java
```diff
@@ -220,4 +222,26 @@ private void mergeBooleanConstraints(Iterable<ExecutionState> states) {
     return result;
   }
 
+  void invalidateRelationsOnSymbol(Symbol.VariableSymbol symbol) {
+    Multimap<Symbol.VariableSymbol, Symbol.VariableSymbol> pairs = HashMultimap.create();
+    for (ExecutionState current = this; current != null; current = current.parentState) {
+      pairs.putAll(current.findRelationPairs(symbol));
+    }
+    for (Map.Entry<Symbol.VariableSymbol, Symbol.VariableSymbol> cell : pairs.entries()) {
+      if (getRelation(cell.getKey(), cell.getValue()) != SymbolicRelation.UNKNOWN) {
```
benzonico: Why check the relation ?

benzonico:
See comment, otherwise LGTM

PR #150, opened 2015-04-15
SONARJAVA-994 Execute rules specific to test on test files

benzonico:
I have an IT to prove all this works together. IT still need a fix to retrieve junit jar properly.


java-squid/src/main/java/org/sonar/java/AbstractJavaClasspath.java
```diff
@@ -0,0 +1,192 @@
...
+  protected abstract void init();
+
+  protected List<File> getFilesFromProperty(String property) {
+    List<File> result = Lists.newArrayList();
+    String fileList = settings.getString(property);
+    if (StringUtils.isNotEmpty(fileList)) {
+      Iterable<String> fileNames = Splitter.on(SEPARATOR).omitEmptyStrings().split(fileList);
+      for (String pathPattern : fileNames) {
+        boolean isLibraryProperty = property.endsWith("libraries");
+        List<File> libraryFilesForPattern = getFilesForPattern(fs.baseDir(), pathPattern, isLibraryProperty);
```
m-g-sonar: As the `baseDir` won't change during loop execution, it may be more interesting to use a variable initialized before the "for" statement instead of accessing it every time.

Same remark for the `isLibraryProperty`.

java-squid/src/main/java/org/sonar/java/JavaSquid.java
```diff
@@ -76,8 +79,11 @@ public JavaSquid(JavaConfiguration conf,
       Iterable<CodeVisitor> measurers = Arrays.asList((CodeVisitor)measurer);
       visitorsToBridge =  Iterables.concat(visitorsToBridge, measurers);
     }
-
-    VisitorsBridge visitorsBridge = new VisitorsBridge(visitorsToBridge, sonarComponents);
+    List<File> classpath = Lists.newArrayList();
+    if(sonarComponents != null) {
```
m-g-sonar: The test `sonarComponents != null` is repeated 4 times in a row. It can be simplified.

java-squid/src/main/java/org/sonar/java/JavaClasspath.java
```diff
@@ -19,81 +19,50 @@
...
-    initalized = false;
   }
 
-  private void init() {
+  protected void init() {
     if (!initalized) {
       initalized = true;
       validateLibraries = project.getModules().isEmpty();
       FilePredicates predicates = fs.predicates();
       hasJavaSources = fs.hasFiles(predicates.and(predicates.hasLanguage("java"), predicates.hasType(InputFile.Type.MAIN)));
```
m-g-sonar: `hasJavaSources` is set in both `JavaClasspath` and `JavaTestClasspath`, with the only difference being the  `InputFile` type (MAIN or TEST). A method using only the type as parameter can probably be put directly in the   `AbstractJavaClasspath` class.

java-squid/src/main/java/org/sonar/java/AbstractJavaClasspath.java
```diff
@@ -0,0 +1,199 @@
...
+  protected List<File> getFilesFromProperty(String property) {
+    List<File> result = Lists.newArrayList();
+    String fileList = settings.getString(property);
+    if (StringUtils.isNotEmpty(fileList)) {
+      Iterable<String> fileNames = Splitter.on(SEPARATOR).omitEmptyStrings().split(fileList);
+      File baseDir = fs.baseDir();
+      boolean isLibraryProperty = property.endsWith("libraries");
+      for (String pathPattern : fileNames) {
+        List<File> libraryFilesForPattern = getFilesForPattern(baseDir, pathPattern, isLibraryProperty);
+        if (validateLibraries && libraryFilesForPattern.isEmpty() && hasJavaSources()) {
```
m-g-sonar: hasJavaSources() will be called for each file... Should be called sooner.

m-g-sonar:
LGTM!

PR #151, opened 2015-04-16
SONARJAVA-966 SONARJAVA-1001 Add support of field reference using explicit "super" and "this"

merciesa:


java-checks/src/main/java/org/sonar/java/symexec/SymbolicEvaluator.java
```diff
@@ -141,6 +145,20 @@ public final void visitBinaryExpression(BinaryExpressionTree tree) {
       }
       return null;
     }
+
+    final boolean isSuperOrThisMemberSelect(ExpressionTree tree) {
+      if (tree.is(Tree.Kind.MEMBER_SELECT)) {
+        MemberSelectExpressionTree memberSelectTree = (MemberSelectExpressionTree) tree;
+        if (memberSelectTree.expression().is(Tree.Kind.IDENTIFIER)) {
+          IdentifierTree identifierExpression = (IdentifierTree) memberSelectTree.expression();
+          if ("super".equals(identifierExpression.name()) || "this".equals(identifierExpression.name())) {
```
benzonico: if can be removed.

benzonico:
And what about member of the same class but from another instance ?
like 
class A { 
String s;

void compare(A a1) {
  if(s==a2.s) {
    if(s==a2.s) {
    }
  }
}
}

benzonico:
I guess this last commit answer my question. LGTM

PR #152, opened 2015-04-16
SONARJAVA-966 SONARJAVA-1000 Primitive values should be considered as symbol values

merciesa:


benzonico:
See comment but otherwise, LGTM.

PR #153, opened 2015-04-17
SONARJAVA-966 SONARJAVA-1006 Fix false positive in try-catch-finally blocks

merciesa:


java-checks/src/main/java/org/sonar/java/symexec/SymbolicEvaluator.java
```diff
@@ -704,6 +719,12 @@ static PackedStatementStates instantiate() {
       return new PackedStatementStates();
     }
 
+    static PackedStatementStates instantiateWithState(ExecutionState state) {
+      PackedStatementStates result = new PackedStatementStates();
```
benzonico: this should probably be just a wrapper to call the next method (with the list argument)

benzonico:
LGTM

PR #155, opened 2015-04-20
SONARJAVA-1002 Remove rule: "if" statement conditions should not unconditionally evaluate to "true" or to "false"

merciesa:


java-checks/src/main/java/org/sonar/java/checks/IfConditionAlwaysTrueOrFalseCheck.java
```diff
@@ -34,9 +33,8 @@
 @Rule(
   key = "S1145",
   name = "\"if\" statement conditions should not unconditionally evaluate to \"true\" or to \"false\"",
-  tags = {"bug", "cwe", "misra", "security"},
-  priority = Priority.MAJOR)
-@ActivatedByDefault
+  priority = Priority.MAJOR,
+  status = "DEPRECATED")
```
benzonico: isn't there a constant in some API to code that ?

benzonico:
LGTM, please double check my comment.

PR #161, opened 2015-04-21
SONARJAVA-1020 RSPEC-2677 "read", "readLine", and "next" return values should be used

merciesa:


java-checks/src/main/java/org/sonar/java/checks/UnusedReturnedDataCheck.java
```diff
@@ -0,0 +1,137 @@
...
+
+@Rule(
+  key = "S2677",
+  name = "\"read\", \"readLine\", and \"next\" return values should be used",
+  tags = {"bug"},
+  priority = Priority.BLOCKER)
+@ActivatedByDefault
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.DATA_RELIABILITY)
+@SqaleConstantRemediation("5min")
+public class UnusedReturnedDataCheck extends BaseTreeVisitor implements JavaFileScanner {
```
benzonico: Why not use a subscriptionBaseVisitor ?

java-checks/src/main/java/org/sonar/java/checks/UnusedReturnedDataCheck.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private static final List<MethodInvocationMatcher> CHECKED_METHODS = ImmutableList.of(
+    MethodInvocationMatcher.create()
+      .typeDefinition(TypeCriteria.subtypeOf("java.io.BufferedReader"))
+      .name("readLine")
+      .withNoParameterConstraint(),
+    MethodInvocationMatcher.create()
+      .typeDefinition(TypeCriteria.subtypeOf("java.util.Iterator"))
+      .name("next")
+      .withNoParameterConstraint(),
```
benzonico: the call to `withNoParameterConstraint` is not required and should be removed. This call will actually match any method called next with whatever the parameters whereas it should match the next method with no parameters in your case.

java-checks/src/main/java/org/sonar/java/checks/UnusedReturnedDataCheck.java
```diff
@@ -0,0 +1,137 @@
...
+    MethodInvocationMatcher.create()
+      .typeDefinition(TypeCriteria.subtypeOf("java.io.Reader"))
+      .name("read")
+      .withNoParameterConstraint());
+
+  @Nullable
+  private MethodInvocationTree currentMethodTree;
```
benzonico: unused field ?

benzonico:
LGTM

PR #162, opened 2015-04-21
SONARJAVA-1014 Junit test cases should call super methods

benzonico:


java-checks/src/main/java/org/sonar/java/checks/CallSuperInTestCaseCheck.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private boolean isWithinJunit3TestCase(Symbol.MethodSymbol methodSymbol) {
+    Type type = methodSymbol.owner().type();
+    return type.isSubtypeOf(JUNIT_FRAMEWORK_TEST_CASE) && !type.symbol().superClass().is(JUNIT_FRAMEWORK_TEST_CASE);
+  }
+
+  private boolean isSetUpOrTearDown(Symbol.MethodSymbol methodSymbol) {
+    return (methodSymbol.name().equals("setUp") || methodSymbol.name().equals("tearDown"))
+        && methodSymbol.parameterTypes().size() == 0 && methodSymbol.returnType().type().is("void");
```
m-g-sonar: You don't need to test the return type as with the same name and number of parameters you are already sure of the override.

m-g-sonar:
Except from the comment about the return type. LGTM!

PR #163, opened 2015-04-21
SONARJAVA-1016 Fix reporting on wrong line for return and throw expressions / Add exceptions to reduce number of issues

merciesa:


benzonico:
LGTM technically. I did not test it extensively functionnaly though.

PR #166, opened 2015-04-22
SONARJAVA-1023 Self contained tests

m-g-sonar:
- Fix management of non-compliant line using @-X
- Change verifier behavior to fail only at the end of the verification process when encountering unexpected issue(s). It now provides a list lines associated with unexpected issues.


benzonico:
LGTM

PR #167, opened 2015-04-22
SONARJAVA-1019 RSPEC-2387 Add support of case insensitive comparison

merciesa:


java-checks/src/main/java/org/sonar/java/checks/ChildClassShadowFieldCheck.java
```diff
@@ -65,26 +65,29 @@ public void visitNode(Tree tree) {
...
           }
         }
       }
     }
   }
 
-  @CheckForNull
-  private Symbol.TypeSymbol getDefiningClassInHierarchyForField(Symbol.TypeSymbol classSymbol, String fieldName) {
+  private void checkForIssue(Symbol.TypeSymbol classSymbol, Tree memberTree, String fieldName) {
+    String upperCaseFieldName = fieldName.toUpperCase();
```
benzonico: This should probably be uppercased directly in the condition that tests it, not prior to this.

benzonico:
Excpet for comment improving readability LGTM.

PR #168, opened 2015-04-22
SONARJAVA-1021 RSPEC-2786 Nested "enum"s should not be declared static

merciesa:


java-checks/src/main/java/org/sonar/java/checks/NestedEnumStaticCheck.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.ENUM);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    Symbol.TypeSymbol symbol = ((ClassTree) tree).symbol();
+    if (symbol.owner().isTypeSymbol() && symbol.isStatic()) {
```
benzonico: This check should not rely on semantic !!!

merciesa: switched to syntax tree

java-checks/src/test/files/checks/NestedEnumStatic.java
```diff
@@ -0,0 +1,14 @@
+enum MyEnum { // Compliant
+}
+
+// should be rejected during compilation
```
benzonico: Should not be required to be tested but ok.

benzonico:
LGTM, please open relevant issue on symbol table if nested enums are not declared implicitely static.

PR #170, opened 2015-04-23
SONARJAVA-1020 RSPEC-2677 Fix FP: Remove Iterator.next() from the scope of the rule

merciesa:


benzonico:
LGTM. But I can see no trace whatsoever of your conversation with Ann to reduce the scope of the rule. Shouldn't this be in a comment on the relevant RSPEC ?

PR #176, opened 2015-04-27
SONARJAVA-1025 Nested enums are implicitly static

m-g-sonar:


SonarTech:

java-squid/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
```diff
@@ -253,7 +253,7 @@ public void InterfaceDeclaration() {
 
     typeSymbol = (JavaSymbol.TypeJavaSymbol) result.symbol("NestedEnum");
     assertThat(typeSymbol.owner()).isSameAs(interfaceSymbol);
-    assertThat(typeSymbol.flags()).isEqualTo(Flags.PUBLIC | Flags.ENUM);
+    assertThat(typeSymbol.flags()).isEqualTo(Flags.PUBLIC | Flags.ENUM | Flags.STATIC);
```
benzonico: Is there some tests to verify that an enum is NOT static in some cases ?

benzonico:
See comment, otherwise LGTM

java-squid/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
m-g-sonar: No, there is no such test. I'll add them.

PR #186, opened 2015-04-30
SONARJAVA-1047 RSPEC-2676 Neither "Math.abs" nor negation should be used on numbers that could be "MIN_VALUE"

merciesa:


SonarTech:

java-checks/src/main/java/org/sonar/java/checks/AbsOnNegativeCheck.java
```diff
@@ -0,0 +1,122 @@
...
+      }
+    } else {
+      ExpressionTree operand = ((UnaryExpressionTree) tree).expression();
+      checkForIssue(operand);
+    }
+  }
+
+  private void checkForIssue(ExpressionTree tree) {
+    MethodInvocationTree nestedTree = extractMethodInvocation(tree);
+    if (nestedTree != null && NEGATIVE_METHODS.anyMatch(nestedTree)) {
```
m-g-sonar: And what about testing if the expression is directly Integer.MIN_VALUE ?

SonarTech:

merciesa:
I won't add support for explicit values now, because LiteralUtils fails for the two specific values.

m-g-sonar:
LGTM!

PR #205, opened 2015-05-08
SONARJAVA-1040 Improve Rule: S864 Report issues only once per statement

merciesa:


SonarTech:

m-g-sonar:
Due to a limitation of the platform and as discussed, it would be better to raise issue on the operator, but at maximum once by line.

java-checks/src/main/java/org/sonar/java/checks/OperatorPrecedenceCheck.java
```diff
@@ -136,7 +149,7 @@ public void visitBinaryExpression(BinaryExpressionTree tree) {
     Tree.Kind peek = stack.peek();
     Tree.Kind kind = getKind(tree);
     if (requiresParenthesis(peek, kind)) {
-      raiseIssue(tree);
+      raiseIssue();
```
m-g-sonar: Raise the issue on the line of the operator, which can be accessed directly using `tree.operatorToken().line()`

merciesa:
We raise the issue per expression, if the line of the operator has not been already reported.

m-g-sonar:
LGTM!

PR #208, opened 2015-05-12
add 'aar' to file suffix filter

dodocat:
Android library archive is now widely used. aar suffix is very common. The filter should include this suffix.

And I am wondering if we should filter the file if user passed a absolute file which is not a directory?

Maybe before filtering the file list, we should check like this:

``` java
      for (String pathPattern : fileNames) {
        File file = new File(pathPattern);
        if (file.exists() && !file.isDirectory()) {
          result.add(file);
          continue;
        }
        List<File> libraryFilesForPattern = getFilesForPattern(baseDir, pathPattern, isLibraryProperty);
      ......
      }
```

If this is proper I will add it to this PR or create a new one.

### DIFF
--- a/java-squid/src/main/java/org/sonar/java/AbstractJavaClasspath.java
+++ b/java-squid/src/main/java/org/sonar/java/AbstractJavaClasspath.java
@@ -127,7 +127,10 @@ public abstract class AbstractJavaClasspath implements BatchExtension {
     if (libraryProperty) {
       if (pattern.endsWith("*")) {
         fileFilter = new AndFileFilter((IOFileFilter) fileFilter,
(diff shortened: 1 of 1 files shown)


benzonico:
Hi @dodocat ,
Sorry for late reply : 
Thanks for the contribution, would you mind describing a little bit more in depth why you need the .aar file suffix added to the classpath ? As I am no android expert, is that a dependency of your android project that you need to solve to provide bytecode to the analyzer ?

dbmeneses:
Closing this as it's missing unit tests, ITs and the actual loading of classes from aar packages by the SquidClassLoader.

PR #212, opened 2015-05-13
SONARJAVA-839 ClassVariableVisibilityCheck: final fields should not raise an issue.

merciesa:


java-checks/src/main/java/org/sonar/java/checks/ClassVariableVisibilityCheck.java
```diff
@@ -84,10 +81,11 @@ private boolean isClass() {
   }
 
   private static boolean isConstant(List<Modifier> modifiers) {
-    return !modifiers.isEmpty() && modifiers.contains(Modifier.FINAL) && modifiers.contains(Modifier.STATIC);
+    return modifiers.contains(Modifier.FINAL);
```
m-g-sonar: Then method should probablybe renamed isFinal()...

m-g-sonar:
LGTM except for comment.

PR #242, opened 2015-05-28
SONARJAVA-1091

benzonico:
Allow to read two format of JaCoCo binary report by shading required classes older version of JaCoCo core.


benzonico:
@Godin please review. To be noted, there seems to be no need to distinguish analyzer to be used when reading coverage per tests.

java-jacoco/src/main/java/org/sonar/plugins/jacoco/JaCoCoReportMerger.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+
+/**
+ * Utility class to merge JaCoCo reports.
+ *
+ * This class handles two versions of JaCoCo binary format to merge.
+ */
+public class JaCoCoReportMerger {
```
SonarTech: MAJOR Add a private constructor to hide the implicit public one. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1118)

SonarTech:
SonarQube analysis reported 1 new issue:
* 1 major

SonarTech:
SonarQube analysis reported 1 new issue:
* 1 major

benzonico:
Even though it is merged, feel free to review @Godin

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
```diff
@@ -182,9 +177,62 @@ public final void readExecutionData(File jacocoExecutionData, SensorContext cont
...
+      dis = new DataInputStream(new FileInputStream(jacocoExecutionData));
+      byte firstByte = dis.readByte();
+      Preconditions.checkState(firstByte == ExecutionDataWriter.BLOCK_HEADER);
+      Preconditions.checkState(dis.readChar() == ExecutionDataWriter.MAGIC_NUMBER);
+      char version = dis.readChar();
+      return version == ExecutionDataWriter.FORMAT_VERSION;
+    } finally {
+      if (dis != null) {
+        dis.close();
+      }
```
Godin: @benzonico use Java 7 try-with-resource

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
```diff
@@ -182,9 +177,62 @@ public final void readExecutionData(File jacocoExecutionData, SensorContext cont
...
+    boolean currentReportFormat = true;
+    try {
+      inputStream = new BufferedInputStream(new FileInputStream(jacocoExecutionData));
+      currentReportFormat = isCurrentReportFormat(jacocoExecutionData);
+      if(currentReportFormat) {
+        ExecutionDataReader reader = new ExecutionDataReader(inputStream);
+        reader.setSessionInfoVisitor(sessionInfoStore);
+        reader.setExecutionDataVisitor(executionDataVisitor);
+        reader.read();
+      }else {
```
Godin: @benzonico missing space before "else"

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
```diff
@@ -272,6 +327,19 @@ private CoverageBuilder analyze(ExecutionDataStore executionDataStore) {
...
+  private void analyzeClassFile(org.jacoco.previous.core.analysis.Analyzer analyzer, File classFile) {
+    InputStream inputStream = null;
+    try {
+      inputStream = new FileInputStream(classFile);
+      analyzer.analyzeClass(inputStream, classFile.getPath());
+    } catch (IOException e) {
+      // (Godin): in fact JaCoCo includes name into exception
+      JaCoCoExtensions.LOG.warn("Exception during analysis of file " + classFile.getAbsolutePath(), e);
+    } finally {
+      Closeables.closeQuietly(inputStream);
```
Godin: @benzonico use Java 7 try-with-resource

java-jacoco/src/main/java/org/sonar/plugins/jacoco/JaCoCoReportMerger.java
```diff
@@ -0,0 +1,86 @@
...
+    for (File report : reports) {
+      if (report.isFile()) {
+        try {
+          isCurrentVersionFormat = AbstractAnalyzer.readJacocoReport(report, dataStore, infoStore);
+        } catch (IOException e) {
+          throw new SonarException(String.format("Unable to read %s", report.getAbsolutePath()), e);
+        }
+      }
+    }
+    return isCurrentVersionFormat;
```
Godin: @benzonico result of merge and subsequent result of analysis will be correct only if all files for a merge have the same version, whereas this is not enforced by this method

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
```diff
@@ -182,9 +177,62 @@ public final void readExecutionData(File jacocoExecutionData, SensorContext cont
     }
   }
 
+  public static boolean isCurrentReportFormat(File jacocoExecutionData) throws IOException {
```
Godin: @benzonico might be better to extract this method from this class

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
```diff
@@ -182,9 +177,62 @@ public final void readExecutionData(File jacocoExecutionData, SensorContext cont
...
+        reader.setExecutionDataVisitor(executionDataVisitor);
+        reader.read();
+      }
+    } finally {
+      Closeables.closeQuietly(inputStream);
+    }
+    return currentReportFormat;
+  }
+
   private boolean readCoveragePerTests(SensorContext context, ExecutionDataVisitor executionDataVisitor) {
```
Godin: @benzonico why you think that analysis of coverage per test should not depend on version? I strongly believe in exact opposite. And thus maybe there are lack of tests?

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
```diff
@@ -140,30 +142,23 @@ private void populateClassFilesCache(File dir, String path) {
...
-        inputStream = new BufferedInputStream(new FileInputStream(jacocoExecutionData));
-        ExecutionDataReader reader = new ExecutionDataReader(inputStream);
-        reader.setSessionInfoVisitor(executionDataVisitor);
-        reader.setExecutionDataVisitor(executionDataVisitor);
-        reader.read();
-      } finally {
-        Closeables.closeQuietly(inputStream);
+        useCurrentFormat = readJacocoReport(jacocoExecutionData, executionDataVisitor);
+      } catch (IOException e) {
+        throw new SonarException(e);
```
Godin: @benzonico the way of propagation of IOException from JaCoCo here is inconsistent with the way in class JaCoCoReportMerger

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
```diff
@@ -182,9 +177,62 @@ public final void readExecutionData(File jacocoExecutionData, SensorContext cont
...
+   * @return true if binary format is the latest one.
+   * @throws IOException in case of error or binary format not supported.
+   */
+  public static boolean readJacocoReport(File jacocoExecutionData, IExecutionDataVisitor executionDataVisitor, ISessionInfoVisitor sessionInfoStore) throws IOException {
+    JaCoCoExtensions.LOG.info("Analysing {}", jacocoExecutionData);
+    InputStream inputStream = null;
+    boolean currentReportFormat = true;
+    try {
+      inputStream = new BufferedInputStream(new FileInputStream(jacocoExecutionData));
+      currentReportFormat = isCurrentReportFormat(jacocoExecutionData);
```
Godin: @benzonico IMO possible to use just one stream instead of opening file multiple times - see methods ``mark`` and ``reset`` in BufferedInputStream, but this might be considered as minor optimization

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
benzonico: I was also surprised by this, so there are tests using coverage per tests with both versions.

java-jacoco-previous/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+                  <shadedPattern>org.jacoco.previous.core.internal</shadedPattern>
+                </relocation>
+              </relocations>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+</project>
```
Godin: @benzonico I personally prefer newline at the end of all textual files, and we supposed to have an active rule for Java files, but unfortunately not for xml

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
Godin: Then simply test is incomplete. Based on changes between versions, differences should be observable for example in case of consecutive method calls placed on different lines:
```
void example() {
  method();
  method();
}
```
previous version will insert two probes - at the beginning of the method and at the next, current version will insert three probes - additional will be between method invocations.

java-jacoco/src/main/java/org/sonar/plugins/jacoco/AbstractAnalyzer.java
benzonico: I am not sure I get the point of this comment.

Godin: @benzonico yes, but in this case without additional message and in another case with

PR #243, opened 2015-05-29
SONARJAVA-1018 Rule: S2325 "private" methods that don't access instance data should be "static"

mpaladin:


java-checks/src/main/java/org/sonar/java/checks/StaticMethodCheck.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private boolean referenceInstance(Symbol symbol) {
+    return symbol.isUnknown() || (!symbol.isStatic() && fromInstance(symbol.owner()));
+  }
+
+  private boolean fromInstance(Symbol owner) {
+    for (Symbol outerClass : outerClasses) {
+      Type ownerType = owner.type();
+      if (owner.equals(outerClass) || (ownerType != null && outerClass.type().isSubtypeOf(ownerType))) {
```
m-g-sonar: Tthe type of the owner should never be null. Did you encounter one?

mpaladin: @m-g-sonar that branch is covered by unit test.

java-checks/src/test/files/checks/StaticMethodCheck.java
```diff
@@ -0,0 +1,58 @@
+package foo;
+
+class Utilities {
```
m-g-sonar: could you add examples with usages of `super` and `this` to also show that these cases  are covered?

m-g-sonar:
If you created a ticket for cases where owner is null, LGTM!

mpaladin:
@m-g-sonar ticket created: http://jira.sonarsource.com/browse/SONARJAVA-1100.

PR #245, opened 2015-05-29
SONARJAVA-547 Add query, super and extends tokens to WildcardTree

m-g-sonar:


m-g-sonar:
PR is now part of https://github.com/SonarSource/sonar-java/pull/251

PR #247, opened 2015-05-30
Implemented exclude filter for try-with-resource variables

renewolfert:
Rule S1481 gives a false positive on the following snippet
```java
java.io.FileOutputStream fileOutputStream = new java.io.FileOutputStream();
try (java.nio.channels.FileLock fileLock = fileOutputStream.getChannel().lock()) {
  // write content to fileOutputStream
  fileOutputStream.flush();
}
```

The `fileLock` variable is required in a try-with-resource statement but there may be no need to do anything with the `FileLock` variable.

This features adds logic to skip the variable-usage check for try-with-resource variables based on a configurable list of classes. The exclude filter can also be used to skip the check for custom AutoCloseable locking implementations.


benzonico:
Thanks for your pull request:  I have created a Jira ticket to address the issue you raised: http://jira.sonarsource.com/browse/SONARJAVA-1101 

I don't want to solve this by adding a parameter but rather excluding FileLock from tryWithResource directly within the check. 

As a consequence I am closing this pull request and will address the issue.

PR #252, opened 2015-06-01
SONARJAVA-715 Rule: A "while" loop should be used instead of a "for" loop

mpaladin:


java-checks/src/test/files/checks/ForLoopUsedAsWhileLoopCheck.java
```diff
@@ -0,0 +1,24 @@
...
+  private void noncompliant() {
+    int i = 0;
+    for (; i < 5;) { // Noncompliant
+      i++;
+    }
+  }
+
+  private void compliant() {
+    int i = 0;
+    while (i < 5) {
```
merciesa: what is the purpose of testing a while loop? :)

merciesa:
LGTM

java-checks/src/test/files/checks/ForLoopUsedAsWhileLoopCheck.java
mpaladin: @merciesa it was just the rspec compliant code, removed.

PR #256, opened 2015-06-02
S00119: use ^[A-Z][1-9]?$ as default regex for type parameters.

jbgi:
using type parameter like V1, V2 is pretty common to indicate
types parameters that have a closely related role in an algorithm.
eg.
```java
static <K,V1,V2> Map<K,V2> transformValues(Map<K,V1> fromMap,
                                  Function<? super V1,V2> function)
```
from guava ```com.google.common.collect.Maps```.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/BadTypeParameterName_S00119_Check.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/BadTypeParameterName_S00119_Check.java
@@ -45,7 +45,7 @@ import java.util.regex.Pattern;
 @SqaleConstantRemediation("10min")
 public class BadTypeParameterName_S00119_Check extends SubscriptionBaseVisitor {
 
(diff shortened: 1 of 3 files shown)


benzonico:
Hi, Thanks for your contribution. 

I have created the following ticket http://jira.sonarsource.com/browse/SONARJAVA-1102  and will proceed with the merging of your pull request right away.

jbgi:
Thanks. By the way, how can I open issues if I don't have an account on http://jira.sonarsource.com/?

benzonico:
To discuss/report an issue best way is the google group and we'll open it, see : http://www.sonarqube.org/get-support/

benzonico:
Merged in https://github.com/SonarSource/sonar-java/commit/dc880219d6e2e7341fbe0fa71c65f3495b936ea2 Thanks.

benzonico:
Oh and by the way, regexp has been changed to support also 0.

PR #257, opened 2015-06-02
SONARJAVA-990 S1068 Exclude more lombok annotations

benzonico:


m-g-sonar:
LGTM!

erikvaningen:
fantastic, thank you for excluding the Lombok annotations

davotoula:
This is great, we are waiting for this enhancement. Is there a way to find out which release this will be part of?
thanks :+1:

benzonico:
As you can see on this commit : https://github.com/SonarSource/sonar-java/commit/244610d51b1a90cb65550461cb737eb62dc19e0c this PR has been merged and is tagged with 3.4 and 3.5 

Moreover this PR refers to ticket : https://jira.sonarsource.com/browse/SONARJAVA-990 marked as fixed in 3.4

So this enhancement is already available from the latest two release of the Sonar-java plugin. To have more detail please see release notes on jira : https://jira.sonarsource.com/jira/secure/ReleaseNote.jspa?projectId=10973&version=11899

davotoula:
Got it, thanks!

l-e-X:
I have recently updated SonarQube to version 4.5.4 and the Java plugin to version 3.5. We have classes annotated with `@Data`, but it seems that the rule `squid:S1068` doesn't handle this "special" annotations. Altough they should be ignored since version 3.4. Please see attached screenshot. Do i need to configure something???
image

benzonico:
Please not use a merged pull request to report/discuss an issue : http://www.sonarqube.org/get-support/

PR #263, opened 2015-06-03
RSPEC-2208 : Wildcard imports should not be used

axel3rd:
Origin : [SUPPORT-550](http://jira.sonarsource.com/servicedesk/customer/portal/1/SUPPORT-550).

Rule key like checkstyle / old rules and not _Sxxxx_ ... but I don't know the naming strategy :(.


benzonico:
Hi Alix thanks for the contribution ! 
You are suggesting an implementation on a rule we want to address our way, that is why I can make the huge following feedback : 
As I think you do want to contribute more rules, here is a list of things that could be improved in order for me to accept your pull request :

- key and implementation : we usually respecify rules to improve documentation : http://jira.sonarsource.com/browse/RSPEC-2208 so for us, this implementation should comply with what is described in the ticket and the key should match this ticket number (so key should be S2208 )

- description should match what is specified in the java ticket for this rule : http://jira.sonarsource.com/browse/RSPEC-2860

- tags, sqale properties and such should match what was specified in RSPEC.

- Test : please use JavaCheckVerifier, see https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/test/java/org/sonar/java/checks/InnerStaticClassesCheckTest.java and its associated test file to understand how it works. It really simplify your test code as the test file becomes seflcontained. I know this is still lacking from SonarExample repo.

- We definitely want to avoid to have too many rules properties when defining a new rule. I know that here you sticked to Checkstyle way of doing thing which is a bit different than our approach. Once again the number of parameter we want is described in the specification ticket : http://jira.sonarsource.com/browse/RSPEC-2208 and we should have none in this case (and accept static star import by default).


Do not hesitate to update this pull request. I will also add some feedback directly in your code.

And again, thanks for the contribution.

java-checks/src/main/java/org/sonar/java/checks/AvoidStarImportCheck.java
```diff
@@ -0,0 +1,119 @@
...
+  @RuleProperty(
+    key = "Exclude",
+    description = "Comma separated list of packages where star imports are allowed (Note that this property is not recursive, subpackages of excluded packages are not automatically excluded)",
+    defaultValue = "" + DEFAULT_EXCLUDE)
+  public String exclude = DEFAULT_EXCLUDE;
+  private HashSet<String> excludePackages = new HashSet<>();
+
+  @Override
+  public List<Kind> nodesToVisit() {
+    if (StringUtils.isNotBlank(exclude)) {
```
benzonico: This method is not really designed to handle initialiazation stuff. Moreover, adding behaviour introduces side effect in a method that should not have any. It is better to override `scanFile` method (and do not forget to call `super` implementation)

java-checks/src/main/java/org/sonar/java/checks/AvoidStarImportCheck.java
```diff
@@ -0,0 +1,119 @@
...
+      for (String p : exclude.split(",")) {
+        excludePackages.add(p + ".*");
+      }
+    }
+    return ImmutableList.of(Tree.Kind.IMPORT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    // Visit import only => cast
```
benzonico: IMO this comment is useless and should be removed.

java-checks/src/main/java/org/sonar/java/checks/AvoidStarImportCheck.java
```diff
@@ -0,0 +1,119 @@
...
+    description = "whether to allow starred static member imports like import static org.junit.Assert.*",
+    defaultValue = "" + DEFAULT_ALLOW)
+  public boolean allowStaticMemberImports = DEFAULT_ALLOW;
+
+  @RuleProperty(
+    key = "Exclude",
+    description = "Comma separated list of packages where star imports are allowed (Note that this property is not recursive, subpackages of excluded packages are not automatically excluded)",
+    defaultValue = "" + DEFAULT_EXCLUDE)
+  public String exclude = DEFAULT_EXCLUDE;
+  private HashSet<String> excludePackages = new HashSet<>();
```
benzonico: I am not really convinced by this exclusion property : I think it should be removed which will simplify the implementation (no need to compute fully qualified name anymore). 

If it is kept however it should not be a `HashSet` but a `Set`

java-checks/src/main/java/org/sonar/java/checks/AvoidStarImportCheck.java
```diff
@@ -0,0 +1,119 @@
...
+    ImportTree importTree = (ImportTree) tree;
+    String importName = fullQualifiedName(importTree.qualifiedIdentifier());
+
+    // There is a star in the import
+    if (importName.endsWith(".*")) {
+      boolean allowed = allowClassImports && !importTree.isStatic();
+      boolean allowedStatic = allowStaticMemberImports && importTree.isStatic();
+      boolean excluded = !excludePackages.isEmpty() && excludePackages.contains(importName);
+      if (!allowed && !allowedStatic && !excluded) {
+        addIssue(importTree, String.format("Using the '.*' form of import should be avoided - %s", importName));
```
benzonico: as message should comply with http://jira.sonarsource.com/browse/RSPEC-2208 and with removal of package exclusion property this should allow to drop computation of name, simplifying the check.

java-checks/src/main/java/org/sonar/java/checks/AvoidStarImportCheck.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  /**
+   * Get full qualified name for tree
+   *
+   * @param tree Tree
+   * @return String
+   */
+  public String fullQualifiedName(Tree tree) {
```
benzonico: This should probably be private (and even static) and thus javadoc is a little bit overkill in this context.

axel3rd:
Hi Nicolas, 

Thanks for this feedback, I didn't search/see RSPEC-2208 / RSPEC-2860 before starting to code :-(.
I will address these points ASAP.

Best Regards.

axel3rd:
New proposal with feedback.

Some items about I'm not sure in (in [RSPEC-2208](http://jira.sonarsource.com/browse/RSPEC-2208)) :
* `Completeness:Full` is the `@ActivatedByDefault` annotation usage ?
* `Applicability:Sources, Tests` requires something or rules are applied on src/test by default ?

benzonico:
Thanks Alix for your work : 
- `Completeness` refers to wether the spec is done or not. So here you should not worry about it.
- `@ActivatedByDefault` maps to the Default Profile field in RSPEC (which is not here in this ticket).
- Regarding applicability, it is all happening in the CheckList class but for now, please let it apply to source only.

benzonico:
I'll try to merge your PR next week.

axel3rd:
ok, available for any complements if required.

benzonico:
Hi Alix, 

your pr was squashed and merged in https://github.com/SonarSource/sonar-java/commit/9d300091259503af5cd6ed9b468600d92c5fa09e 

One thing though:  please configure your git repository to use core.autocrlf=true so you won't commit windows line endings. Thanks.

Thanks again for your work.

PR #278, opened 2015-06-09
Fix false negative in 'useless parentheses' check on assignment of instance variables

renewolfert:
The UselessParenthesesCheck raises an issue on the return statement in the ```getValue``` (getter) method but not on the assignment of the instance variable in the ```setValue``` (setter) method. The UselessParenthesesCheck did not check for parent expressions of the type ```ASSIGNMENT```.

```java
public Date getValue() {
  return (this.value != null ? (Date) this.value.clone() : null);
}

public void setValue(Date newValue) {
  this.value = (newValue != null ? (Date) newValue.clone() : null);
}
```


m-g-sonar:
Hey René!

Thanks for your contribution and for spotting that missing case. On top of that, thank you a lot for completing unit tests as well! I'll merge your commit with the master as soon as possible.

m-g-sonar:
Manually merged. I close your PR.

PR #281, opened 2015-06-10
RSPEC-1659 : Multiple variables should not be declared on the same line

axel3rd:
Proposal for [RSPEC-1659](http://jira.sonarsource.com/browse/RSPEC-1659), note about it :
* No tag => I choose **convention**
* **description** can't be apply for Java language => some liberties in *S1659.html*
* sqale cost to **5min** is not too much ?

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/OneDeclarationPerLineCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/OneDeclarationPerLineCheck.java
@@ -20,13 +20,14 @@
 package org.sonar.java.checks;
 
 import com.google.common.collect.ImmutableList;
-import java.util.ArrayList;
 import java.util.List;
-import org.apache.commons.lang.StringUtils;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.tree.BlockTree;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.StatementTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.Tree.Kind;
 import org.sonar.plugins.java.api.tree.VariableTree;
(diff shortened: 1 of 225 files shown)


java-checks/src/main/java/org/sonar/java/checks/OneDeclarationPerLineCheck.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public void visitNode(Tree tree) {
+    VariableTree varTree = (VariableTree) tree;
+    int varLineCurrent = varTree.endToken().line();
+    if (varLineCurrent == varLineLast) {
+      varsOnSameLine.add(varTree.symbol().name());
+    } else {
+      if (!varsOnSameLine.isEmpty()) {
+        addIssue(varLineLast, String.format("Declare \"%s\" on a separate line.", StringUtils.join(varsOnSameLine, "\", \"")));
```
m-g-sonar: With that approach you will miss some cases, in particular the case of the last variable declaration. For instance, your current implementation, you will miss the issue in the following code:

```java
class A {
  int a; // Compliant
  int b, c; // Noncompliant and detected
  int d, e; // Noncompliant but not detected
}
```

java-checks/src/test/files/checks/OneDeclarationPerLineCheck.java
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * Test MultipleVariableDeclarations inside a method
+   * @return int
+   */
+  public int method() {
+    int var1 = 0, var2 = -1; // Noncompliant {{Declare "var2" on a separate line.}}
+    int varOk = 1; // Compliant, only one on the line
+    return varFieldInt1 + varFieldInt2 + varFieldInt3 + var1 + var2 + varOk;
+  }
+
```
m-g-sonar: I'm wondering about corner cases here, what should be the status of following example?

```java
int a; int b; // issue?
int c,
    d;  // issue ?

**Comment by m-g-sonar** [comment]
Hello Alix!

First, thank you for your contribution, participating to rules implementation is really appreciated!

Regarding your comments.
- Your html description is perfectly fine, thx for the adaptation to java. Maybe adding corner cases  could be a plus;
- `Convention` looks fine as tag;
- The remediation cost is indeed a bit high, I'll ask @ganncamp .

I added some comments related to your implementation. At the moment you are however missing some cases. You will probably have to slightly change your approach in order to solve that issue. Maybe only focusing on the `,` as `endToken` could help.

Regarding the target of the rule,  I wonder if we should only focus on single type declaration or only single lines ? (see corner cases from my comment) Keep in mind that the test file used by the UT is somehow a kind of documentation on how the rule react. Adding all possible case allows to be perfectly clear to identify how its behavior.

On that point, as the rule is going to deprecate the one from checkstyle, it could be useful to look at how they are reacting to such cases. 

I'll go back to you as soon as you update the PR!

Thanks! :)

**Comment by axel3rd** [comment]
Hi Michael, 

Thanks for the feedback and snippet code.
That will teach me to code in my couch in front of TV ...
I will test these (not corner) cases with checkstyle and come back with a better implementation.

Best regards.

**Comment by axel3rd** [comment]
NB about previous commit :
- Strategy change : One issue by default, for techdebt coherence (and like checkstyle behavior)
- In addition of comma check, use last var line check for case : ``int a; int b;``

**Comment by axel3rd** [comment]
Quite good for me ... available for any changes if requested.

**Comment by ganncamp** [comment]
@m-g-sonar the implementation matches the spec as far as . You want to downgrade it to 2min?
Also, I've added the 'convention' tag to the spec.

**Comment by m-g-sonar** [comment]
@ganncamp Yep 2min is probably better! Thanks for updating the RSPEC with the `Convention` tag.

**Comment by ganncamp** [comment]
Done @m-g-sonar

File: java-checks/src/main/java/org/sonar/java/checks/OneDeclarationPerLineCheck.java [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+    return ImmutableList.of(Tree.Kind.VARIABLE);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    VariableTree varTree = (VariableTree) tree;
+    if (varSameDeclaration || lastVarLine == varTree.simpleName().identifierToken().line()) {
+      addIssue(tree, String.format("Declare \"%s\" on a separate line.", varTree.symbol().name()));
+    }
+    varSameDeclaration = ",".equals(varTree.endToken().text());
```
m-g-sonar: `endToken()` is `Nullable`, you may have NPE here.

m-g-sonar:
Hey Alix, 

Sorry, I did not spot it last time I reviewed your code... But the current approach won't work as expected.

- First, variables does not have always an `endToken`, so you may encounter NPE during execution (in particular for the case of the last parameter of a method).
- Second, your code is currently exploring **all** the `VariableTree`s. However, your code aim only variable declarations. Unfortunately, `VariableTree` can be found at multiple other places.

At the moment, the current code snippet would raise problems:

```java
  public void foo(int a, int b) {  // NPE (2nd variable w/o endToken) but it would also have raised an issue when none is expected
    try {
      
    } catch(Exception e) { // NPE (variable w/o endToken)
      
    }
    
    for (Object o : Lists.newLinkedList()) { // NPE (variable w/o endToken)
    }
  }
```

At this point, I think your best option is to change your approach and extends a `BaseTreeVisitor` instead of a `SubscriptionBaseVisitor`. This will allow you to totally control the exploration and therefore only focus on variables you really want to investigate (more precisely, avoid exploring some part of trees). You will for instance avoid exploring method parameters, catch block parameters, `ForEachStatement` variable, etc.

If you need some examples, you can look at the checks `ArrayEqualsCheck` or `AssignmentInSubExpressionCheck` for instances.

Don't hesitate to ask any question.

axel3rd:
Hi Michael, 

Thanks for your feedback.
I excuse myself for my previous implementation.
I will review it and come with better code, consider this previous snippet.

Best regards.

benzonico:
Hi, sorry to weigh in like this but I don't think you would need to change the approach that much and use a BaseTreeVisitor here. 
If my understanding is correct, you want to focus only on variable tree that are field declaration. Then just register to all Kind that implements ClassTree (ie: INTERFACE, ENUM, ANNOTATION_TYPE (which is not needed here as you won't be able to declare any variable in annotation type) and CLASS)  and then iterate through its members and check with your current logic only the members that are of kind VARIA

axel3rd:
Hi Nicolas, 

Ok, I will use ``BaseTreeVisitor`` and visit ``ClassTree``/``MethodTree`` (for [variables declaration in a method](https://github.com/SonarSource/sonar-java/pull/281/files#diff-c16207051951ea9aac82d9b14a7c08f5R10)) and add @m-g-sonar snippets in test file ... who expose some cases that I did thought in first implementation.
I will try to do that at beginning of next week.

Best regards.

benzonico:
Hi Alix, My point was in fact that you actually don't need the BaseTreeVisitor here, subscribing to CLASS ENUM and INTERFACE kind should do the trick.

axel3rd:
Grrr, bad copy/paste sorry. 

We are in phase : using ``SubscriptionBaseVisitor``, same check logic, but update visitors strategy to catch only wanted variables.

axel3rd:
>  I will try to do that at beginning of next week.

My week starts now ^^

I hope this last push will working as expected ...

benzonico:
Merged in https://github.com/SonarSource/sonar-java/commit/6c7a3caa4f8172ccd1e11082e3333f37ee1eb9a1 see http://jira.sonarsource.com/browse/SONARJAVA-1148

PR #287, opened 2015-06-12
use svg icon of travis ci build

stas-vilchik:


benzonico:
I have absolutely no pblm to actually merge this but could you explain why is this better ?

stas-vilchik:
svg icons look much better on high pixel density screens (like retina)

PR #288, opened 2015-06-15
RSPEC-3052 : Fields should not be initialized to default values - TO DELETE

axel3rd:
Proposal for [RSPEC-3052](http://jira.sonarsource.com/browse/RSPEC-3052).


java-checks/src/main/java/org/sonar/java/checks/FieldsShouldNotBeInitializedToDefaultValuesCheck.java
```diff
@@ -0,0 +1,90 @@
...
+    for (Tree member : classTree.members()) {
+      if (member.is(Tree.Kind.VARIABLE)) {
+        checkField((VariableTree) member);
+      }
+    }
+  }
+
+  public void checkField(VariableTree fieldTree) {
+    // No initialization or not literal => go out
+    if (!(fieldTree.initializer() instanceof LiteralTree)) {
```
benzonico: You should not be using `instance of` operator but rather check for nullity and use `is()` method and Kinds.

java-checks/src/main/java/org/sonar/java/checks/FieldsShouldNotBeInitializedToDefaultValuesCheck.java
```diff
@@ -0,0 +1,90 @@
...
+      return;
+    }
+
+    boolean explicitInit = false;
+    LiteralTree initTree = (LiteralTree) fieldTree.initializer();
+    try {
+      if (!initTree.symbolType().isPrimitive() && "null".equals(initTree.value())) {
```
benzonico: This should probably be tested using LITERAL_NULL kind.

java-checks/src/main/java/org/sonar/java/checks/FieldsShouldNotBeInitializedToDefaultValuesCheck.java
```diff
@@ -0,0 +1,90 @@
...
+        explicitInit = true;
+      } else if (initTree.symbolType().isPrimitive() && initTree.symbolType().isNumerical()
```
benzonico: isNumerical is sufficient here, first part of the condition is redundant.

java-checks/src/main/java/org/sonar/java/checks/FieldsShouldNotBeInitializedToDefaultValuesCheck.java
```diff
@@ -0,0 +1,90 @@
...
+        && NumberFormat.getInstance().parse(initTree.value()).shortValue() == 0) {
+        explicitInit = true;
+      } else if (initTree.symbolType().isPrimitive(Primitives.BOOLEAN) && !Boolean.parseBoolean(initTree.value())) {
+        explicitInit = true;
+      }
+    } catch (ParseException e) {
```
benzonico: I am not a found of surrounding the whole else if with a try catch for only some cases of this else if.

Moreover, I tend to think that, as this else if is only setting a variable to true, the conditions should be extracted in well labeled methods returning a boolean to express behaviour properly and thus reducing the main code to only one if with a or condition based on those extracted methods.

benzonico:
Hi @axel3rd , 
Thanks for the good work.
See my comments in code. And let me know of any question. Can you please squash all your commits into a single one ? that would help review in this case.

axel3rd:
Thanks @benzonico for this feedback. 
I will do these points tomorrow (and check if a 0.0000001 initialization could be a false positive with this logic).

axel3rd:
@benzonico : Required job done ... but not sure of cause of _merge conflict_ :-(

benzonico:
@axel3rd have a closer look at rebase command in git and especially at interactive rebase.

axel3rd:
Ok ... I will propose a new PR ... I mess up last action.

axel3rd:
New PR for this feature : [here](https://github.com/SonarSource/sonar-java/pull/297).

PR #290, opened 2015-06-16
SONARJAVA-1117 FP in S109 (MagicNumber): Numbers used in constructor of enum constants should be ignored

merciesa:


java-checks/src/main/java/org/sonar/java/checks/MagicNumberCheck.java
```diff
@@ -57,6 +58,11 @@ public void scanFile(JavaFileScannerContext context) {
   }
 
   @Override
+  public void visitEnumConstant(EnumConstantTree tree) {
+    // Ignore literals within enum constant
```
benzonico: and what about a case like : 
```
enum Foo {
  PLOP(100) {
   int i = 100;
  void foo(int i) {
    System.out.println(42);
  }
};
void foo(int i) {}
}
```

merciesa: Do you see a case where the initializer of the enum could be something else than a NewClassTree ?

benzonico: Check in the construction of the EnumConstantTreeImpl : https://github.com/SonarSource/sonar-java/blob/master/java-squid/src/main/java/org/sonar/java/model/declaration/EnumConstantTreeImpl.java#L32

java-checks/src/main/java/org/sonar/java/checks/MagicNumberCheck.java
```diff
@@ -57,6 +59,11 @@ public void scanFile(JavaFileScannerContext context) {
   }
 
   @Override
+  public void visitEnumConstant(EnumConstantTree tree) {
+    scan(((NewClassTree) tree.initializer()).classBody());
```
benzonico: I created ticket and fixed it : https://github.com/SonarSource/sonar-java/pull/291

benzonico:
LGTM. Can you wait until https://github.com/SonarSource/sonar-java/pull/291 is merged so you can actually remove the useless cast.

PR #295, opened 2015-06-16
SONARJAVA-1146 + SONARJAVA-1147

m-g-sonar:


benzonico:
I am not sure of the first commit : it is actually reverting your modification so this is not per say a solution to SONARJAVA-1146

PR #321, opened 2015-07-06
SONARJAVA-1181 : False negative Undocumented API check when javadoc is empty on class or methods

axel3rd:
Proposal for [SONARJAVA-1181](http://jira.sonarsource.com/browse/SONARJAVA-1181) (Origin : [SonarQube Google Group](https://groups.google.com/forum/#!topic/sonarqube/hT_e1VAP7FM)).

NB : Implementation checks empty description on parameters (not in specification, but logic IMO).

Many changes on file test [UndocumentedApi.java](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/test/files/checks/UndocumentedApi.java), update with [JavaCheckVerifier](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/test/java/org/sonar/java/checks/verifier/JavaCheckVerifier.java).


benzonico:
Thanks for contribution : fixed in https://github.com/SonarSource/sonar-java/pull/367

PR #324, opened 2015-07-17
Make compatible with surefire reuseForks=false

fabdouglas:
JacocoController.getInstance() is executed too soon in case of surefire option reuseForks=false.
In this configuration, the listener is created without javaagent enabled. It is enabled only when the VM is forked while starting a test.


fabdouglas:
In addition I've fixed a small issue in case  of broken test. Previous implementation saved the session id just before the dump, at the end of the tests.
This PR sets the SessionId immediatly,

stepanv:
Any update on this? For admins, note that this fixes SONARJAVA-728. Am I right fabdouglas?

fabdouglas:
@stepanv Yes, it relates to this SONARJAVA-728 but I had not comment right on this Jira.

Godin:
@stepanv I'll try to allocate time to look into that.
@fabdouglas First of all - thanks for your contribution. Could you please clean up history a bit (squash + rebase on top of current master)? According to your comments multiple issues were fixed, so could you please make commit messages more explicit to be able to relation between problem and change?

Godin:
@fabdouglas ping

paulmfarrar:
Hi @fabdouglas & @Godin, any chance we could get some movement on this change? I need reuseForks to work with Sonar analysis and without this change it constantly fails. Happy to help if I can. Thanks.

fabdouglas:
Indeed, the history is a bit polluted... 
I'm sorry, I had not much time to make this rebase and merge the new conflicts.
Maybe the next week.

paulmfarrar:
Was this merged?

PR #334, opened 2015-07-29
SONARJAVA-1191S1312 should not raise issues for maven loggers

mpaladin:
http://jira.sonarsource.com/browse/SONARJAVA-1191


benzonico:
Adding this exception to the rule should probably lead to an update of the rule description to mention it.

PR #335, opened 2015-07-30
SONARJAVA-1187 and SONARJAVA-1189

mpaladin:
http://jira.sonarsource.com/browse/SONARJAVA-1187
http://jira.sonarsource.com/browse/SONARJAVA-1189


mpaladin:
Two tickets on the same pr because they are on the same rule.

PR #340, opened 2015-08-03
SONARJAVA-1140 Expand S1158 to also cover "compareTo"

mpaladin:
http://jira.sonarsource.com/browse/SONARJAVA-1140

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/ToStringUsingBoxingCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ToStringUsingBoxingCheck.java
@@ -24,6 +24,7 @@ import com.google.common.collect.ImmutableSet;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/ToStringUsingBoxingCheck.java
```diff
@@ -86,11 +89,15 @@ private static String getNewlyCreatedClassName(MethodInvocationTree mit) {
     return "";
   }
 
-  private static boolean isCallingToString(MethodInvocationTree mit) {
-    if (mit.methodSelect().is(Tree.Kind.MEMBER_SELECT)) {
-      MemberSelectExpressionTree mset = (MemberSelectExpressionTree) mit.methodSelect();
-      return "toString".equals(mset.identifier().name());
+  private static String isCallingToStringOrCompareTo(ExpressionTree methodSelect) {
```
m-g-sonar: Even if it's a private method, I really think that you should add the `@CheckForNull` annotation.

PR #355, opened 2015-08-14
SONARJAVA-1190

m-g-sonar:
http://jira.sonarsource.com/browse/SONARJAVA-1190


java-checks/src/main/java/org/sonar/java/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -54,7 +60,7 @@ public void scanFile(JavaFileScannerContext context) {
 
   @Override
   public void visitLambdaExpression(LambdaExpressionTree tree) {
-    if (isMethodInvocation(tree.body()) || isBlockInvokingMethod(tree.body())) {
+    if ((isMethodInvocation(tree.body()) && isUsingLambdaParamAsArg(tree)) || isBlockInvokingMethod(tree.body())) {
```
mpaladin: @m-g-sonar I would move `isMethodInvocation(tree.body())` inside `isUsingLambdaParamAsArg` method because right now the cast to `MethodInvocationTree` happens inside `isUsingLambdaParamAsArg` which makes the rule a little unclear.

PR #359, opened 2015-08-17
SONARJAVA-1224 Handle implicit variables "this" and "super"

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/MembersDifferOnlyByCapitalizationCheck.java
```diff
@@ -171,13 +172,18 @@ private static String getSymbolTypeName(Symbol symbol) {
...
+      if ((variableButNotHiddenMember(symbol) || methodButNotConstructor(symbol)) && !(symbol.isPrivate() && ignorePrivate)) {
         results.add(symbol);
       }
     }
     return results;
   }
 
+  private static boolean variableButNotHiddenMember(Symbol symbol) {
+    String name = symbol.name();
+    return symbol.isVariableSymbol() && !"this".equals(name) && !"super".equals(name);
```
benzonico: Several things : 
-  as you handle the "this" symbol, it should be explicitely tested
- it should be explicitely mentioned in jira ticket
- This is not hidden members... implicit eventually but hidden has a total different meaning in java semantic : http://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.8.2

m-g-sonar:
Merged.

PR #363, opened 2015-08-19
SONARJAVA-1171 Allow custom rules for test files.

benzonico:


m-g-sonar:
I think that integration tests are not complete regarding the real use of test-oriented rules as extension.

If the new check focusing on tests has indeed been added in the {{java-extension-plugin}} project, the actual tests in {{com.sonar.it.java.suite.JavaExtensionsTest}} does not check for any results.

PR #367, opened 2015-08-19
SONARJAVA-1181 Report issue on empty javadoc comments

benzonico:


java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
```diff
@@ -139,6 +139,12 @@ private void visitNode(Tree tree) {
     }
   }
 
+  private static boolean isEmptyJavadoc(String javadoc) {
+    //remove start and end of doc as well as line feed and stars.
+    String cleanedupJavadoc = javadoc.trim().substring(3).replace("\n", "").replace("*/", "").replace("*", "").trim();
```
m-g-sonar: this won't work if the javadoc contains other new line characters (`\r` for instance).

PR #368, opened 2015-08-19
SONARJAVA-1201 Fix false positive when referencing local variables

benzonico:


benzonico:
merged in https://github.com/SonarSource/sonar-java/commit/f0821a4f2f4dee625305933f5bcd5c1d00ef2392

PR #370, opened 2015-08-20
SONARJAVA-1217 Refine raised issues

mpaladin:
https://jira.sonarsource.com/browse/SONARJAVA-1217

When the iterable is an identifier raise an issue only if its owner
is a method to skip fields.

When the iterable is not an identifier raise issues only for
consecutives foreachs.

### DIFF
--- a/java-checks/src/test/files/checks/LoopsOnSameSetCheck.java
+++ b/java-checks/src/test/files/checks/LoopsOnSameSetCheck.java
@@ -7,17 +7,22 @@ public class A {
     return null;
   }
 
+  private List<String> globalStrings;
+
   public void doSomethingToAList(List<String> strings, List<String> strings2) {
     for (String str : strings);
-    for (String str : strings); // Noncompliant {{Combine this loop with the one that starts on line 11.}}
+    foo();
+    for (String str : strings); // Noncompliant {{Combine this loop with the one that starts on line 13.}}
     strings = null;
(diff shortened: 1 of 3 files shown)


java-checks/src/test/files/checks/LoopsOnSameSetCheck.java
```diff
@@ -7,17 +7,22 @@
...
-    for (String str : strings); // Noncompliant {{Combine this loop with the one that starts on line 11.}}
+    foo();
+    for (String str : strings); // Noncompliant {{Combine this loop with the one that starts on line 13.}}
     strings = null;
     for (String str : strings); // strings get reassigned
     for (String str : getList());
-    for (String str : getList()); // Noncompliant {{Combine this loop with the one that starts on line 15.}}
+    for (String str : getList()); // Noncompliant {{Combine this loop with the one that starts on line 18.}}
+    foo();
+    for (String str : getList()); // no issue if something in between for loops
```
m-g-sonar: Doesn't this contradict the non-compliant issue of line 15 ?

mpaladin: @m-g-sonar nope, when the iterable is an identifier which is not a field there can be things in between the for loops as long as they don't reference it.

m-g-sonar: Okay!

PR #372, opened 2015-08-22
Regular expression rule template for checking for architectural constraints

cardil:
A source code comply to an architectural model when it fully adheres to a set of architectural constraints. A constraint set with regular expression allows to deny various code conventions, or to force them.

You can for instance use this rule to :

 * Force using given format of error codes ex.: `new ErrCode\((?!"[A-Z]{2}-[0-9]{6}").*\);`
 * Deny using given code ex.: `@ManagedBean`

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CheckList.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CheckList.java
@@ -344,6 +344,7 @@ public final class CheckList {
       .add(MembersDifferOnlyByCapitalizationCheck.class)
       .add(LoopsOnSameSetCheck.class)
       .add(PublicStaticMutableMembersCheck.class)
+      .add(RegularExpressionSimpleCheck.class)
       .build();
   }
 
--- a/java-checks/src/main/java/org/sonar/java/checks/RegularExpressionSimpleCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/RegularExpressionSimpleCheck.java
@@ -1,0 +1,216 @@
+/*
+ * SonarQube Java
+ * Copyright (C) 2012 SonarSource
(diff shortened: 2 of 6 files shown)


Cosaquee:
:+1:

MattAgile:
:+1:

adko-pl:
:+1:

benzonico:
Thanks for your pull request. 

Unfortunately I don't see a big value out of this rule as a rule for the java plugin. This check has absolutely no dependency on the java language in fact. I tend to think this could eventually be valuable as a rule template in a dedicated plugin rather than a java rule in the java plugin.  By doing so you would be able to actually configure on which kind of file you want to match the searched expression and not only limit that to java file. 

Moreover, such generic rules goes against what we are trying to achieve in terms of zero configuration by providing a rule that is way too generic.

With those arguments in mind I am closing this pull request. 
Thanks again for your contribution.

cardil:
That's unfortunate. That's rule is in fact a template and shouldn't  be active by default like architectural constraints rule that is present now in sonar. It's meant to be replacement for checkstyle regex rule. I can't finally dump chceckstyle unless a rule template like this.

Yup. This can be generic rule but easy of configuration for java file can speak to place it here.

If you are not convinced I will just publish a micro plugin for sonar with this rule

benzonico:
When we reimplemented checkstyle rules we had a look at those regexp rule and decided to skip it as it is too generic in our opinion and not particularly dedicated to Java language. 

Therefore we really think there would be value by having those as a dedicated plugin. More especially given that would allow to actually define this rule for several languages and not only Java. 

Feel free to implement this as a separate plugin and reach the sonarqube google group ( https://groups.google.com/forum/#!forum/sonarqube ) to announce your contribution to SonarQube ecosystem. 

Cheers,

PR #375, opened 2015-09-01
SONARJAVA-1267 S1948 SerializableFieldInSerializableClassCheck : Support of java.util.Map

gronono:
Hi guys,

Maps should be treated in the same way as Collections in S1948.

Regards.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/SerializableFieldInSerializableClassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/SerializableFieldInSerializableClassCheck.java
@@ -32,6 +32,7 @@ import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.Modifier;
 import org.sonar.plugins.java.api.tree.ParameterizedTypeTree;
 import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.TypeArguments;
 import org.sonar.plugins.java.api.tree.TypeTree;
 import org.sonar.plugins.java.api.tree.VariableTree;
 import org.sonar.plugins.java.api.tree.WildcardTree;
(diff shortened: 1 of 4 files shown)


benzonico:
Hi thanks for the contribution. 

Could you develop a little bit your argumentation in order for us to accept the pull request ? 
Moreover, can you have a look at the travis and appveyor build so they pass. Those are failing on ruling test (which can be run using `mvn clean install` in the its/ruling directory using a jdk 8) because you modified behaviour of this rule. As such the expected results should be updated to match this modification.

java-checks/src/main/java/org/sonar/java/checks/SerializableFieldInSerializableClassCheck.java
```diff
@@ -129,7 +143,7 @@ private static boolean implementsSerializable(@Nullable Type type) {
       return implementsSerializable(((Type.ArrayType) type).elementType());
     }
     if (type.isClass() || ((JavaType) type).isTagged(JavaType.TYPEVAR)) {
-      return type.erasure().isSubtypeOf("java.io.Serializable");
+      return type.erasure().isSubtypeOf(Serializable.class.getName());
```
benzonico: This has nothing to do with your contribution. I would keep the class name as a String as it is clearer and avoid to import Serializable which is not required. Same remark for the other arguments.

java-checks/src/main/java/org/sonar/java/checks/SerializableFieldInSerializableClassCheck.java
```diff
@@ -41,7 +42,10 @@
 
 import javax.annotation.Nullable;
 
+import java.io.Serializable;
+import java.util.Collection;
 import java.util.List;
+import java.util.Map;
```
benzonico: As mentioned in other comment, those imports should not be required just to resolve the name of the class and thus be removed.

benzonico:
Ok so this is in addition to the following tickets : https://jira.sonarsource.com/browse/SONARJAVA-1225 (which should then include Map if we accept your PR) and https://jira.sonarsource.com/browse/SONARJAVA-1061 (which fixed the exclusion of Collections)

gronono:
yes it is the same as the task SONARJAVA-1061 but for the Map.
In SONARJAVA-1225, they suggest to look at the concrete type. In case of DTO, it will be very difficult because setters will take a Map argument.

gronono:
I fail to run JavaRulingTest locally. I fixed the json file by examining the cases in hand, hoping that it is good. I also took the opportunity to do a rollback on the automatic organization of imports on the class SerializableFieldInSerializableClassCheck

benzonico:
When build is ok, what you can do is squash your commits into one and force push on your branch to update the PR nicely.

gronono:
I merged the commits. And builds should be ok.

gronono:
A JIRA bug was created: https://jira.sonarsource.com/browse/SONARJAVA-1267

benzonico:
Merged in https://github.com/SonarSource/sonar-java/commit/ebb2e78f53733f78305e10d3687c0b3426db257d Thanks for contribution !

PR #376, opened 2015-09-01
SONARJAVA-1249 S1996 : Files should contain only one class or interface each

mpaladin:
https://jira.sonarsource.com/browse/SONARJAVA-1249

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/OneClassInterfacePerFileCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/OneClassInterfacePerFileCheck.java
@@ -1,0 +1,57 @@
+/*
+ * SonarQube Java
+ * Copyright (C) 2012 SonarSource
+ * sonarqube@googlegroups.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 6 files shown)


java-checks/src/main/java/org/sonar/java/checks/OneClassInterfacePerFileCheck.java
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  public List<Kind> nodesToVisit() {
+    return ImmutableList.of(Kind.COMPILATION_UNIT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    int types = ((CompilationUnitTree) tree).types().size();
+    if (types > 1) {
+      context.addIssue(-1, OneClassInterfacePerFileCheck.this, "There are " + types + " types in this file; move all but one of them to other files.", (double) (types - 1));
```
benzonico: no need for the call on the context object, you can call `addIssue` with int and message (see `IssuableSubscriptionVisitor`)

benzonico: Oh nevermind, I did not spot the sqale effort provided.

mpaladin: @benzonico it's because of the `effortToFix`.

PR #380, opened 2015-09-02
SONARJAVA-1240 Rule S2972: Inner classes should not have too many lines

mpaladin:
https://jira.sonarsource.com/browse/SONARJAVA-1240


java-checks/src/main/java/org/sonar/java/checks/InnerClassTooManyLinesCheck.java
```diff
@@ -0,0 +1,76 @@
...
+    return ImmutableList.of(Kind.CLASS, Kind.ENUM, Kind.INTERFACE, Kind.ANNOTATION_TYPE);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree node = (ClassTree) tree;
+    Symbol.TypeSymbol symbol = node.symbol();
+    Symbol owner = symbol.owner();
+    Type ownerType = owner.type();
+    if (ownerType != null && ownerType.isClass() && owner.owner().type() == null) {
```
benzonico: I don't easily get what the last bit of the condition is required for ?

mpaladin: @benzonico you will like it more now.

PR #383, opened 2015-09-03
SONARJAVA-1241 Rule S2973: Escaped Unicode characters should not be used

mpaladin:
https://jira.sonarsource.com/browse/SONARJAVA-1241


java-checks/src/main/java/org/sonar/java/checks/EscapedUnicodeCharactersCheck.java
```diff
@@ -0,0 +1,86 @@
...
+
+  @Override
+  public List<Kind> nodesToVisit() {
+    return ImmutableList.of(Kind.STRING_LITERAL);
+  }
+
+  @Override
+  public void visitNode(Tree node) {
+    String value = ((LiteralTree) node).value().replaceAll("\\\\\\\\u", "\\\\z");
+    value = value.substring(1, value.length() - 1);
```
benzonico: Please use trimQuotes method in LiteralUtils : https://github.com/SonarSource/sonar-java/blob/master/java-squid/src/main/java/org/sonar/java/model/LiteralUtils.java#L94-94

mpaladin:
@benzonico thanks for spotting .class file, comments fixed.

PR #389, opened 2015-09-09
SONARJAVA-1253 Rule S2221: "Exception" should not be caught when not required by called methods

mpaladin:
https://jira.sonarsource.com/browse/SONARJAVA-1253


java-checks/src/main/java/org/sonar/java/checks/CatchExceptionCheck.java
```diff
@@ -0,0 +1,118 @@
...
+
+  private boolean catchesExceptionCheck(BlockTree block, TypeTree catchType) {
+    return isJavaLangException(catchType.symbolType()) && !helperVisitor.containsExplicitThrowsException(block);
+  }
+
+  private static boolean isJavaLangException(Type type) {
+    return type.is("java.lang.Exception");
+  }
+
+  private static class HelperVisitor extends BaseTreeVisitor {
```
benzonico: Detail: HelperVisitor is a crappy name (too generic) : what about methodThrowingExceptionFinder ? or something alike ?

PR #392, opened 2015-09-09
SONARJAVA-1251 S1228 Package-Info raise false positive when analyzing a single file

mpaladin:
https://jira.sonarsource.com/browse/SONARJAVA-1251

### DIFF
--- a/java-squid/src/main/java/org/sonar/java/SonarComponents.java
+++ b/java-squid/src/main/java/org/sonar/java/SonarComponents.java
@@ -24,9 +24,11 @@ import com.google.common.collect.Lists;
 import org.sonar.api.BatchExtension;
 import org.sonar.api.batch.fs.FileSystem;
 import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.batch.fs.InputPath;
 import org.sonar.api.batch.rule.CheckFactory;
 import org.sonar.api.batch.rule.Checks;
 import org.sonar.api.component.ResourcePerspectives;
+import org.sonar.api.issue.Issuable;
 import org.sonar.api.measures.FileLinesContext;
 import org.sonar.api.measures.FileLinesContextFactory;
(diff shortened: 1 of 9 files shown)


mpaladin:
@benzonico code removed from `ChecksBridge` and added the ability to report issues on a file different from the one analysed, feedback is welcome as I don't fully like it.

java-squid/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -98,6 +108,10 @@ public Highlightable highlightableFor(File file) {
     return resourcePerspectives.as(Highlightable.class, inputFromIOFile(file));
   }
 
+  public Issuable issuableFor(File file) {
```
benzonico: Test ?

java-squid/src/main/java/org/sonar/java/model/VisitorsBridge.java
```diff
@@ -242,6 +250,37 @@ public String getFileKey() {
       return sourceFile.getKey();
     }
 
+    @CheckForNull
+    private RuleKey getRuleKey(JavaCheck check) {
```
benzonico: This is not related to this ticket, but doing this should probably lead to remove completely the ChecksBridge class and report directly from the context class. If you share the feeling can you please create the ticket ?

mpaladin: @benzonico I fully share, right now the index is per project, it should be at least per file or on the go, ticket created: https://jira.sonarsource.com/browse/SONARJAVA-1256

java-squid/src/main/java/org/sonar/java/SonarComponents.java
mpaladin: @benzonico test would not be realistic with `ResourcePerspective`.

PR #397, opened 2015-09-15
SONARJAVA-1150 Fix incorrect resolution of methods in generic classes

mpaladin:
https://jira.sonarsource.com/browse/SONARJAVA-1150

### DIFF
--- a/java-squid/src/main/java/org/sonar/java/resolve/Types.java
+++ b/java-squid/src/main/java/org/sonar/java/resolve/Types.java
@@ -69,8 +69,7 @@ public class Types {
           for (JavaType.ClassJavaType classType : t.getSymbol().superTypes()) {
             erasedTypes.add(classType.erasure());
           }
-
-          result = erasedTypes.contains(s);
+          result = erasedTypes.contains(s.erasure());
           break;
         case JavaType.BOT:
           result = s.tag == JavaType.BOT || s.tag == JavaType.CLASS || s.tag == JavaType.ARRAY;
--- a/java-squid/src/test/files/sym/references/MethodCall.java
+++ b/java-squid/src/test/files/sym/references/MethodCall.java
@@ -1,4 +1,4 @@
-package references;
+package references; import java.util.List;
 
 @SuppressWarnings("all")
 class MethodCall extends Parent {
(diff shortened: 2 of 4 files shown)


java-squid/src/main/java/org/sonar/java/resolve/Resolve.java
```diff
@@ -23,7 +23,6 @@
 import com.google.common.collect.ImmutableList;
 
 import javax.annotation.Nullable;
-
```
benzonico: Detail but this has nothing to do with this commit.

java-squid/src/main/java/org/sonar/java/resolve/Types.java
```diff
@@ -65,12 +65,11 @@ public boolean isSubtype(JavaType t, JavaType s) {
           }
 
           //FIXME work on erased types while generics method is not implemented/read from bytecode.
-          Set<JavaType> erasedTypes = Sets.newHashSet();
+          Set<String> erasedTypes = Sets.newHashSet();
           for (JavaType.ClassJavaType classType : t.getSymbol().superTypes()) {
-            erasedTypes.add(classType.erasure());
+            erasedTypes.add(classType.erasure().fullyQualifiedName());
```
benzonico: I might be wrong but I am really wondering if you did not, somehow shortcut erasure by relying on fully qualified name and therefore hacked a solution instead of fixing the real issue... We would need to discuss in depth what is going on here.

java-squid/src/test/files/sym/references/MethodCall.java
```diff
@@ -181,3 +181,29 @@ void varargs_usage_2() {
...
+    private void complexGenericMethod(T argument) {
+    }
+  }
+
+  public void test() {
+    NestedGenericClass<List<Object>> nestedGenericClass = new NestedGenericClass<List<Object>>(new java.util.ArrayList<Object>());
+    nestedGenericClass.genericMethod(new java.util.LinkedList<Object>());
+    ComplexNestedGenericClass<List<Object>> complexNestedGenericClass = new ComplexNestedGenericClass<List<Object>>(new java.util.ArrayList<Object>());
+    complexNestedGenericClass.complexGenericMethod(new java.util.LinkedList<Object>());
+  }
```
m-g-sonar: Not sure that we are testing all the possible situations there.
On my side I also tested with the following cases, without encountering any issues:
```
    ComplexNestedGenericClass v1 = new ComplexNestedGenericClass(new java.util.ArrayList<Object>());
    ComplexNestedGenericClass<U> v2 = new ComplexNestedGenericClass<U>(myU);
    ComplexNestedGenericClass<? extends List<Object>> v3 = new ComplexNestedGenericClass<List<Object>>(new java.util.ArrayList<Object>());
```

PR #314, opened 2015-06-26
SONARJAVA-1079 : S109 - Magic Numbers: add an "authorized numbers" parameter

axel3rd:
Proposal for [SONARJAVA-1079](http://jira.sonarsource.com/browse/SONARJAVA-1079)

PR #399, opened 2015-09-15
SONARJAVA-1079 : "authorized numbers" parameter  & SONARJAVA-1235 : exclude final only

axel3rd:
Hi,

Proposal for : 
* [SONARJAVA-1079](http://jira.sonarsource.com/browse/SONARJAVA-1079) following #314
* [SONARJAVA-1235](https://jira.sonarsource.com/browse/SONARJAVA-1235) (included in this PR, 1 code line updated in check)

Best regards.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/MagicNumberCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/MagicNumberCheck.java
@@ -19,28 +19,32 @@
  */
 package org.sonar.java.checks;
 
+import java.math.BigDecimal;
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+import java.text.ParseException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
+import org.sonar.check.RuleProperty;
 import org.sonar.java.model.ModifiersUtils;
 import org.sonar.plugins.java.api.JavaFileScanner;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
 import org.sonar.plugins.java.api.tree.AnnotationTree;
 import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.EnumConstantTree;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.LiteralTree;
 import org.sonar.plugins.java.api.tree.Modifier;
-import org.sonar.plugins.java.api.tree.ModifiersTree;
+import org.sonar.plugins.java.api.tree.NewArrayTree;
 import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.Tree.Kind;
 import org.sonar.plugins.java.api.tree.VariableTree;
 import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
 import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
-
-import java.math.BigDecimal;
-import java.text.DecimalFormat;
-import java.text.DecimalFormatSymbols;
-import java.text.ParseException;
-import java.util.Locale;
 
 @Rule(
   key = "S109",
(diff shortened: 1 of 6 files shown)


benzonico:
Hi Alix, before we can have a look at this pull request, it has to pass the checks in place. That means that you would have to make pass the ruling (see what is under its/ruling directory) and update the expected files (checking that what you update is correct). 

In order to keep the commit history clean, you can amend your commit and then force push on your branch : that will automatically update this PR and relaunch the automatic checks.

axel3rd:
@benzonico : Thanks for explanation. [squid-S109.json](https://github.com/SonarSource/sonar-java/blob/master/its/ruling/src/test/resources/commons-beanutils/squid-S109.json) udpated, [WrapDynaClass.java](https://svn.apache.org/repos/asf/commons/proper/beanutils/trunk/src/main/java/org/apache/commons/beanutils/WrapDynaClass.java) (commons-beanutils 1.9.3-SNAPSHOT) is *final* => not a MagicNumber.

But I didn't found 4 differences as IT build mentions it in *issues-report.html*.
Sorry if this PR requires some other builds :(

axel3rd:
Sorry for the delay, I had some difficulties to run [orchestrator](https://github.com/SonarSource/orchestrator) locally (behind corporate proxy & windows maven renamed = tricky).

Currently the PR implementation is bad about ``final Object[] params = new Object[2];`` (issue not raised). I will try to do better quickly.

m-g-sonar:
No worries @axel3rd, I will go back to your PR as soon as you tell me it's OK!

axel3rd:
@m-g-sonar : Running ITS on Guava (many differences found, thanks !), there are questions about specifications of [SONARJAVA-1235](https://jira.sonarsource.com/browse/SONARJAVA-1235) :-(.

An issue must be raised (or not) in these cases ?

*Constructor*
```
// Checkstyle : no
final List<Object> a = new ArrayList<>(42);
final BigInteger bi = new BigInteger("16a09e667f3bcc908b2fb1366ea957d3e3adec17512775099da2f590b0667322a", 16);
```
*Array initialization*
```
// Checkstyle : yes
final Object[] p1 = new Object[42]; 
final char[] c = new char[1024];
```
*Static type*
```
private static final MyType MY_TYPE = new MyType() {
    int magic = 42;  // Checkstyle : yes
  };
```
*Primitive array (without new)*
```
final long[] la = {3L, 4L}; // Checkstyle : no
```
*Prrimitive initialization with operation*
```
final long l = 3L * 2;  // Checkstyle : no
```
*Initialisation with method*
```
final ByteBuffer b = ByteBuffer.allocateDirect(8); // Checkstyle : no
```
*Return*
```
return new char[1024]; // checkstyle : yes (but no doubt a bug in current MagicNumber implementation)
```
Thanks

m-g-sonar:
Let's try to keep the implementation simple. I would suggest to ignore variables as soon as the `final` modifier is present, at the exception of array dimensions. That way it seems that we will follow the checkstyle implementation. For the return statements however (interesting case!), it seems indeed smarter to not raise issue.

axel3rd:
@m-g-sonar : ok for a *checkstyle like* implementation.

*return statements* : IMO it should raise a issue because : 
```
byte[] r = new byte[1024]; // raise an issue, not final
return r;

// very different of ?
return new byte[1024];
```

But I let you the final decision.

m-g-sonar:
@axel3rd: I totally misread your previous comment. Yep, we should totally raise an issue in that case. Note that current implementation (v 3.5) of the MagicNumberCheck already raise one for such cases.

axel3rd:
@m-g-sonar : ITs updated. 
Know limitation of implementation : The check in a final class inside another not done ; considered like Object (see. [UT](https://github.com/SonarSource/sonar-java/pull/399/files#diff-b99616f564432d41576462038dd149efR74)).

java-checks/src/main/java/org/sonar/java/checks/MagicNumberCheck.java
```diff
@@ -100,9 +118,9 @@ public void visitAnnotation(AnnotationTree annotationTree) {
 
   @Override
   public void visitVariable(VariableTree tree) {
-    // skip static final variables
-    ModifiersTree modifiers = tree.modifiers();
-    if (!(ModifiersUtils.hasModifier(modifiers, Modifier.STATIC) && ModifiersUtils.hasModifier(modifiers, Modifier.FINAL))) {
+    ExpressionTree initializer = tree.initializer();
+    boolean newArray = initializer != null && initializer.is(Kind.NEW_ARRAY) && ((NewArrayTree) initializer).newKeyword() != null;
```
m-g-sonar: Instead of looking for the presence of the `new` keyword, you should check for presence of initializers.
Currently, you will have an issue with the following code :
```
final long[] array = new long[] {3L, 4L}; // Should be compliant
```

m-g-sonar:
@axel3rd: Okay for the limitations, as long as they are documented in the UT, it's fine for me!
Except for the small issue regarding new array initializers, everything looks fine with current version!

axel3rd:
> small issue regarding new array initializers

Good catch, thanks, fixed.

m-g-sonar:
Perfect @axel3rd, I'll now proceed with the merge!

m-g-sonar:
Merged with master! I consequently close this PR. See your commit here: https://github.com/SonarSource/sonar-java/commit/6777e2703ef81d009f41dbe00de114dc836914d7

Thanks a lot for your contribution! I'll also close the 2 related JIRA ticket.

PR #400, opened 2015-09-15
SONARJAVA-1250 Surefire Parser: do not try to read time attribute for skipped tests

mpaladin:
https://jira.sonarsource.com/browse/SONARJAVA-1250

### DIFF
--- a/java-surefire/src/main/java/org/sonar/plugins/surefire/data/SurefireStaxHandler.java
+++ b/java-surefire/src/main/java/org/sonar/plugins/surefire/data/SurefireStaxHandler.java
@@ -87,7 +87,8 @@ public class SurefireStaxHandler implements XmlStreamHandler {
     detail.setName(name);
 
     String status = UnitTestResult.STATUS_OK;
-    long duration = getTimeAttributeInMS(testCaseCursor);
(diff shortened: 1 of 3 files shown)


java-surefire/src/main/java/org/sonar/plugins/surefire/data/SurefireStaxHandler.java
```diff
@@ -87,7 +87,8 @@ private static UnitTestResult parseTestResult(SMInputCursor testCaseCursor) thro
     detail.setName(name);
 
     String status = UnitTestResult.STATUS_OK;
-    long duration = getTimeAttributeInMS(testCaseCursor);
+    String time = testCaseCursor.getAttrValue("time");
```
m-g-sonar: Variable `time` should be only defined in block `if(duration == null) { ... }`

mpaladin: @m-g-sonar it needs to be defined before, `testCaseCursor` moves forward in ` testCaseCursor.descendantElementCursor()` and won't be able to return time attribute later.

PR #411, opened 2015-09-25
SONARJAVA-1237 Extract Reassignment Finder class and fix NPE

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/helpers/ReassignmentFinder.java
```diff
@@ -0,0 +1,99 @@
...
+import org.sonar.plugins.java.api.tree.Tree;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+
+public class ReassignmentFinder extends BaseTreeVisitor {
```
benzonico: Why externalize the class if it is not tested on its own ? Moreover I think that this class should be more documented to at least clearly expose its purpose.

m-g-sonar: I'll cut the modifications in two separated commits and add dedicated unit tests.

PR #414, opened 2015-09-30
SONARJAVA-1268 RSPEC-2083 Values used in path traversal should be neutralized

m-g-sonar:


m-g-sonar:
Rule has not been merged in plugin. Work on the topic is postponed this next iteration targeting the owasp benchmark.

PR #456, opened 2015-10-22
SONARJAVA-1266 FP in S1948: Fields having unknown symbol type should not be considered as being not serializable

mpaladin:
https://jira.sonarsource.com/browse/SONARJAVA-1266


java-checks/src/main/java/org/sonar/java/checks/SerializableFieldInSerializableClassCheck.java
```diff
@@ -132,7 +137,8 @@ private static boolean isSerializable(Tree tree) {
 
   private static boolean implementsSerializable(@Nullable Type type) {
     if (type == null || type.isUnknown()) {
-      return false;
+      // do not raise an issue if type is unknown
+      return true;
     }
     if (type.isPrimitive()) {
```
m-g-sonar: you could have merged the expression of this `if` with the previous one.

PR #461, opened 2015-10-23
Java version compatible checks

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/SAMAnnotatedCheck.java
```diff
@@ -66,6 +70,11 @@
       build();
 
   @Override
+  public boolean isCompatibleWithJavaVersion(Integer version) {
+    return JavaVersionHelper.java8Compatible(version);
```
mpaladin: @m-g-sonar why compatible and not guaranteed?

m-g-sonar: Indeed if the rule is enabled by default, having the adequate version should be guaranteed.

PR #491, opened 2015-11-19
SONARJAVA-1384 - A switch statement without default case must allow for

DidierBesset:
non execution of any case.

### DIFF
--- a/java-squid/src/main/java/org/sonar/java/cfg/CFG.java
+++ b/java-squid/src/main/java/org/sonar/java/cfg/CFG.java
@@ -31,6 +31,7 @@ import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
 import org.sonar.plugins.java.api.tree.BlockTree;
 import org.sonar.plugins.java.api.tree.BreakStatementTree;
 import org.sonar.plugins.java.api.tree.CaseGroupTree;
+import org.sonar.plugins.java.api.tree.CaseLabelTree;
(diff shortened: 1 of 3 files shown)


java-squid/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -578,10 +580,16 @@ private void buildSwitchStatement(SwitchStatementTree switchStatementTree) {
     // process body
     currentBlock = createBlock(switchSuccessor);
     breakTargets.addLast(switchSuccessor);
+    boolean defaultCaseMet = false;
```
benzonico: I would rename it : hasDefault or hasDefaultCase

java-squid/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -578,10 +580,16 @@ private void buildSwitchStatement(SwitchStatementTree switchStatementTree) {
...
     if (!switchStatementTree.cases().isEmpty()) {
       CaseGroupTree firstCase = switchStatementTree.cases().get(0);
       for (CaseGroupTree caseGroupTree : Lists.reverse(switchStatementTree.cases())) {
         build(caseGroupTree.body());
+        for (CaseLabelTree caseLabel : caseGroupTree.labels()) {
+          if ("default".equals(caseLabel.caseOrDefaultKeyword().text())) {
+            defaultCaseMet = true;
```
benzonico: Once found you can break the loop.

PR #505, opened 2015-11-25
SONARJAVA-1399 Execute naively catch blocks

benzonico:


java-squid/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -723,12 +725,23 @@ private void buildForStatement(ForStatementTree tree) {
...
+    Block beforeFinally = createBlock(currentBlock);
+    List<Block> catches = new ArrayList<>();
+    for (CatchTree catchTree : tryStatementTree.catches()) {
+      currentBlock = createBlock(finallyOrEndBlock);
+      build(catchTree.block());
+      catches.add(currentBlock);
+    }
+    currentBlock = beforeFinally;
     build(tryStatementTree.block());
+    for (Block catchBlocks : catches) {
```
m-g-sonar: Should be ``catchBlock``.

PR #508, opened 2015-11-26
SONARJAVA-tbd: Rule improvement: Ignore 'throws Exception' for @Test methods

vorburger:
Bonjour les amis à Genève (je vous écris depuis Lausanne),

How about this one? Good idea, no?

I would have opened a JIRA on SONARJAVA, but you seem to have blocked New Issue creation for new user accounts - perhaps you'd like to grant that JIRA permission to me? (Or am I just too dumb to see the button?) New JIRA should probably be linked to https://jira.sonarsource.com/browse/SONARJAVA-125, commented in https://jira.sonarsource.com/browse/RSPEC-112, like I saw in https://jira.sonarsource.com/browse/SONARJAVA-225 or https://jira.sonarsource.com/browse/SONARJAVA-125.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/RawException_S00112_Check.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/RawException_S00112_Check.java
@@ -19,7 +19,9 @@
  */
 package org.sonar.java.checks;
 
-import com.google.common.collect.ImmutableSet;
+import java.util.List;
+import java.util.Set;
+
 import org.apache.commons.lang.BooleanUtils;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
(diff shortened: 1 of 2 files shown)


benzonico:
Bonjour Lausanne ;) ! and thanks for the contribution ! 

About jira : this is on purpose : we want to discuss what goes as a ticket before we actually create them : main channel to discuss features with java team would be through googlegroup : sonarqube@googlegroups.com (that you can use as an email) 

Then regarding your contribution, I have nothing to say about your implementation which is fine but as for functionality I have two issues : 
- This assume that you run this rule against test code which goes against what we have in mind in a mid term plan (and already exists for some rules) to actually have rule dedicated for tests and rule for main code. Ideally we would like to make things tunable from quality profile. But as for now, we assume that this rule will be run against main code only and as such your feature would be irrelevant. 

- Even without this in mind, I tend to think that there is no reason for your test method to actually throw Exception. I do know that this is a classical template generated by IDEs but IMO it should not and you should only throw specific exception if required and even maybe just handle the exceptions within your tests without "leaking" to the signature of the method. 

Given those two reasons I am closing this pull request as I don't think this is the direction we want to go. 

Once again though, thanks for your contribution. 

Cheers,

PR #512, opened 2015-12-01
SONARJAVA-1383 Introducing JavaVersion

m-g-sonar:


java-squid/src/main/java/org/sonar/plugins/java/api/JavaVersion.java
```diff
@@ -0,0 +1,41 @@
...
+
+@Beta
+public interface JavaVersion {
+
+  boolean isJava7Compatible();
+
+  boolean isJava8Compatible();
+
+  @Nullable
+  Integer asInteger();
```
benzonico: Why ?

java-squid/src/main/java/org/sonar/java/model/JavaVersionImpl.java
```diff
@@ -0,0 +1,100 @@
...
+  private static final int JAVA_8 = 8;
+  private final int javaVersion;
+  private final boolean initialized;
+
+  public JavaVersionImpl() {
+    this.javaVersion = -1;
+    this.initialized = false;
+  }
+
+  public JavaVersionImpl(int javaVersion) {
```
benzonico: Should deserve a @VisibleForTesting  ?

m-g-sonar: No, as it is actually used by {{JavaCheckVerifier}} (allow testing behavior of rules with given version of java).

PR #522, opened 2015-12-04
SONARJAVA-1375 Fix overriding test for method symbols

benzonico:


java-squid/src/main/java/org/sonar/java/resolve/JavaSymbol.java
```diff
@@ -493,6 +493,9 @@ public void setMethodType(JavaType.MethodJavaType methodType) {
 
     @CheckForNull
     public MethodJavaSymbol overriddenSymbol() {
+      if(isStatic()) {
```
mpaladin: @benzonico formatting: ``if(`` -> ``if (``

PR #526, opened 2015-12-08
Do not use Chocolatey in Appveyor

Godin:
Check of default Chocolatey feed fails from time to time. And this check was done for each build, even if Maven was already installed.

Also by removing usage of Chocolatey we receive less files for cache, and can automatically select Apache mirror.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,14 +6,21 @@ branches:
 
 cache:
   # Note that we don't use cache for local Maven repository - see https://github.com/SonarSource/sonar-java/pull/525
-  - C:\ProgramData\chocolatey\bin -> appveyor.yml
-  - C:\ProgramData\chocolatey\lib -> appveyor.yml
-  - C:\bin\apache-maven-3.2.5 -> appveyor.yml
+  - C:\Users\appveyor\maven -> appveyor.yml
 
 install:
   - set MAVEN_VERSION=3.2.5
-  - choco install maven -version %MAVEN_VERSION%
-  - set PATH=%PATH%;C:\bin\apache-maven-%MAVEN_VERSION%\bin
+  # Note that indentation is important here:
+  - ps: |
+      Add-Type -AssemblyName System.IO.Compression.FileSystem
+      if (!(Test-Path -Path "C:\Users\appveyor\maven" )) {
+        (new-object System.Net.WebClient).DownloadFile(
+          "http://www.apache.org/dyn/closer.cgi?action=download&filename=maven/maven-3/$($env:MAVEN_VERSION)/binaries/apache-maven-$($env:MAVEN_VERSION)-bin.zip",
+          "C:\Users\appveyor\maven.zip"
+        )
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -6,15 +6,22 @@ branches:
...
+  - C:\Users\appveyor\maven -> appveyor.yml
 
 install:
   - git submodule update --init --recursive
   - set MAVEN_VERSION=3.2.5
-  - choco install maven -version %MAVEN_VERSION%
-  - set PATH=%PATH%;C:\bin\apache-maven-%MAVEN_VERSION%\bin
+  # Note that indentation is important here:
+  - ps: |
+      Add-Type -AssemblyName System.IO.Compression.FileSystem
```
dbolkensteyn: @Godin we can achieve the same in an easier way, using `curl` and `unzip`, see https://github.com/dbolkensteyn/nemo-roslyn/blob/master/appveyor.yml#L14

appveyor.yml
```diff
@@ -6,15 +6,22 @@ branches:
 
 cache:
   # Note that we don't use cache for local Maven repository - see https://github.com/SonarSource/sonar-java/pull/525
-  - C:\ProgramData\chocolatey\bin -> appveyor.yml
-  - C:\ProgramData\chocolatey\lib -> appveyor.yml
-  - C:\bin\apache-maven-3.2.5 -> appveyor.yml
+  - C:\Users\appveyor\maven -> appveyor.yml
```
dbolkensteyn: @Godin is this worth doing in the first place? is Apache download servers are fast & stable, then probably not. Perhaps this was initially introduced exactly because of Chocolatey's own instability.

dbolkensteyn: In sonar-csharp, we're not caching Maven itself: see https://github.com/SonarSource/sonar-csharp/blob/master/appveyor.yml

But then, sonar-csharp gets build much less frequently than sonar-java, so stability issues are harder to spot

appveyor.yml
Godin: @dbolkensteyn You're right, I was thinking about this: I suppose that recent Windows machines do have PowerShell by default, while according to http://www.appveyor.com/docs/installed-software#tools `unzip` comes from 7-zip and `curl` comes as a transitive dependency of Git. So this variant can be copy-pasted easily into another environment, while another not. From here it is probably up to Java team to decide which way they'd like to maintain ( cc @benzonico ).

appveyor.yml
Godin: @dbolkensteyn if this caching of Maven was introduced to overcome instability of Chocolatey, then this didn't worked - see description of PR: Chocolatey anyway always was trying to install Maven.

The main point here is to remove Chocolatey, primarily because of instability of its public feed, and the rest was preserved as it was. Note that it is even recommended to use own feed -  https://github.com/chocolatey/choco/wiki/ChocolateyFAQs#should-my-organization-depend-on-use-the-community-feed-httpschocolateyorgpackages
So whether caching of Maven helps a lot or not is IMO another story, but this can be easily measured by timing Appveyor phases.

PR #527, opened 2015-12-08
SONARJAVA-1414 - Opened resources passed as argument to instance crea…

DidierBesset:
…tion should no longer be tracked.

### DIFF
--- a/java-squid/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
+++ b/java-squid/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
@@ -33,6 +33,7 @@ import org.sonar.plugins.java.api.JavaFileScanner;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
 import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.Arguments;
 import org.sonar.plugins.java.api.tree.AssignmentExpressionTree;
 import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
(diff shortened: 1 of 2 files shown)


java-squid/src/test/files/se/UnclosedResourcesCheck.java
```diff
@@ -240,4 +240,18 @@ public void correctHandlingOfJarFile(String fileName) {
       closeJar(jar);
     }
   }
+  
+  public InputStream getStreamSimon(@Nonnull String obj) throws IOException {
```
benzonico: Method name is quite odd...

java-squid/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -217,11 +220,15 @@ public void visitMethodInvocation(MethodInvocationTree syntaxNode) {
...
-        final List<SymbolicValue> values = programState.peekValues(syntaxNode.arguments().size() + 1);
-        final List<SymbolicValue> argumentValues = values.subList(1, values.size());
-        for (SymbolicValue target : argumentValues) {
-          programState = closeResource(programState, target);
-        }
+        attemptToCloseArguments(syntaxNode.arguments(), 1);
+      }
+    }
+
+    private void attemptToCloseArguments(final Arguments arguments, int stackOffset) {
```
benzonico: Why is that an attempt ?  I am not sure of the stackOffset argument, it appears like a magic number to me.

DidierBesset: Stack for Method invocation: SV that will contain the value returned by the method + values of the arguments
Stack for NewClass: only values of the arguments.
Alternative is to use a boolean (newClass or methodInvocation)

PR #537, opened 2015-12-15
Fix ClassCastException in TreeFactory

MarkZ3:
An instance of ArrayTypeTreeImpl could be wrongly cast to
PrimitiveTypeTreeImpl.

See also https://groups.google.com/forum/#!topic/sonarqube/VTsAYxhGRiA


benzonico:
As discussed in the google group thread : https://groups.google.com/forum/#!topic/sonarqube/VTsAYxhGRiA  I am closing this PR because it lacks test and moreover the fix end up ignoring the annotation on type which can lead to odd behaviour in rules afterwards.

PR #551, opened 2015-12-17
SONARJAVA-1236 S2187 - FP when abstract parent class has the tests annotated methods

mpaladin:


java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
```diff
@@ -68,60 +69,74 @@ public boolean apply(SymbolMetadata.AnnotationInstance input) {
...
+        checkJunit4TestClass(simpleName, symbol, members);
       }
     }
-    reportIssue(classTree.simpleName(), "Add some tests to this class.");
   }
 
-  private void checkJunit3TestClass(ClassTree tree) {
-    if (tree.symbol().type().isSubtypeOf("junit.framework.TestCase")) {
-      checkMethods(tree, false);
+  private void checkTestNGmembers(IdentifierTree identifierTree, Iterable<Symbol> members) {
```
m-g-sonar: @mpaladin I would rename `identifierTree` to `className` to explicit from where it comes from!

java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
```diff
@@ -68,60 +69,74 @@ public boolean apply(SymbolMetadata.AnnotationInstance input) {
...
+      }
     }
+    reportIssue(identifierTree, "Add some tests to this class.");
   }
 
-  private void checkJunit4TestClass(ClassTree tree) {
-    IdentifierTree name = tree.simpleName();
-    if (name != null && name.name().endsWith("Test")) {
-      checkMethods(tree, true);
+  private void checkJunit3TestClass(IdentifierTree simpleName, JavaSymbol.TypeJavaSymbol symbol, Iterable<Symbol> members) {
```
m-g-sonar: @mpaladin I would prefer using a more meaningful name for the identifier tree from the class name.

java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
```diff
@@ -68,60 +69,74 @@ public boolean apply(SymbolMetadata.AnnotationInstance input) {
...
+    if (symbol.type().isSubtypeOf("junit.framework.TestCase")) {
+      checkMethods(simpleName, members, false);
     }
   }
 
-  private void checkMethods(ClassTree classTree, boolean forJunit4) {
-    for (Tree member : classTree.members()) {
-      if (member.is(Kind.METHOD) && isTestMethod(forJunit4, (MethodTree) member)) {
+  private void checkJunit4TestClass(IdentifierTree simpleName, JavaSymbol.TypeJavaSymbol symbol, Iterable<Symbol> members) {
```
m-g-sonar: @mpaladin same comment about the name of the variable.

PR #559, opened 2015-12-21
IT Ruling: Remove unused dependency on diffutils

dbolkensteyn:
This `diffutils` dependency doesn't seem to be used to me - but let's leave the ITs verify this!


benzonico:
I guess you're right, let's findout !

PR #563, opened 2015-12-22
Migrate some rules to precise issue location

mpaladin:


benzonico:
LGTM that will be a nice update on this topic.

PR #573, opened 2015-12-30
SONARJAVA-1273 Rule : S3346 "assert" should only be used with boolean variables

mpaladin:


java-checks/src/main/java/org/sonar/java/checks/AssertOnBooleanVariableCheck.java
```diff
@@ -0,0 +1,98 @@
...
+      .addAll(SIDE_EFFECT_KIND)
+      .build();
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree.is(Kind.ASSERT_STATEMENT)) {
+      withinAssert = true;
+      containsSideEffect = false;
+    } else if (withinAssert && !containsSideEffect) {
```
benzonico: Right side of the && is pretty useless IMO

java-checks/src/main/java/org/sonar/java/checks/AssertOnBooleanVariableCheck.java
```diff
@@ -0,0 +1,98 @@
...
+  key = "S3346",
+  name = "\"assert\" should only be used with boolean variables",
+  priority = Priority.MAJOR,
+  tags = {Tag.SUSPICIOUS})
+@ActivatedByDefault
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.INSTRUCTION_RELIABILITY)
+@SqaleConstantRemediation("5min")
+public class AssertOnBooleanVariableCheck extends SubscriptionBaseVisitor {
+
+  private static final List<Kind> SIDE_EFFECT_KIND = ImmutableList.of(
```
benzonico: Ok it is really an edge case and probably not a real world one but what about NewClassTree ?

```
assert new Comparator<String>() {
      @Override
      public int compare(String s, String t1) {
        return 0;
      }
    }.compare("foo", "bar") <0;
```

java-checks/src/main/java/org/sonar/java/checks/AssertOnBooleanVariableCheck.java
mpaladin: @benzonico it spares a memory assignment if `containsSideEffect` was already set to true.

PR #582, opened 2016-01-06
Fix parsing of Scala classes containing symbols

jpeel:
Scala classes are allowed to have symbols like +-~/: in them.
When compiled, these become Java class names by replacing the symbol
with a $ and a word. + becomes $plus. The code for parsing class names
for inner classes didn't handle these cases and would throw errors
saying that classes didn't exist when sonar was run. This fixes the
parsing.

### DIFF
--- a/java-squid/src/main/java/org/sonar/java/resolve/Convert.java
+++ b/java-squid/src/main/java/org/sonar/java/resolve/Convert.java
@@ -20,10 +20,33 @@
 package org.sonar.java.resolve;
 
 import com.google.common.base.Preconditions;
+import java.util.HashSet;
 import org.apache.commons.lang.StringUtils;
 
 public class Convert {
+  private static final HashSet<String> scalaSymbols;
 
+  static {
+    scalaSymbols = new HashSet<String>();
+    scalaSymbols.add("$tilde");
+    scalaSymbols.add("$eq");
+    scalaSymbols.add("$less");
+    scalaSymbols.add("$greater");
+    scalaSymbols.add("$bang");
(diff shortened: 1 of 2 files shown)


jpeel:
I ran into this with a project using Akka Persistence due to a class having an object with the name of '->'.

jpeel:
Any thoughts/suggestions on this pull request? What else needs to be done to get this integrated because it seems necessary for analysis of any Java code that relies on Scala libraries.

objectmastery:
Running into a similar problem with a project using the Play Framework. It would be great if this pull request could be looked into.

Is there anything that can be done to help it along i.e. is there anything missing, source code that needs updating, etc.?

benzonico:
The problem I have with this fix is that it breaks some valid cases to handle a special case : if we apply it like it is done that means we break valid cases of inner classes name tilde. 
The correct way to handle that would be to ensure that inner classes with a name starting with a $ are properly supported : could you provide a class where this error is happening so we can reproduce, create a ticket and solve it in a general way ? 
Thanks

benzonico:
Hi, 

Thanks to this thread : 
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/sonarqube/YVqM8aON2pQ/wdFG50qNDgAJ

I was finally able to reproduce the issue and took some time to figure it out.

I created a ticket to handle it : https://jira.sonarsource.com/browse/SONARJAVA-2043 

The correct and general solution is not to add exceptions related to scala compiler but rather than properly solve all inner class names starting with $ (or outer class name ending with $). 

Thanks for making the feedback and the contribution.

PR #584, opened 2016-01-07
SONARJAVA-1457 Only consider int and long from remainder of 1 check

benzonico:


java-checks/src/main/java/org/sonar/java/checks/ConstantMathCheck.java
```diff
@@ -113,7 +114,7 @@ public void visitNode(Tree tree) {
     if (hasSemantic()) {
       if (tree.is(Tree.Kind.REMAINDER)) {
         BinaryExpressionTree remainderTree = (BinaryExpressionTree) tree;
-        if (isIntegralOne(remainderTree.rightOperand())) {
+        if (isIntegralOne(remainderTree.rightOperand()) && isIntOrLong(remainderTree)) {
```
mpaladin: @benzonico you already have `isIntegral(` method in this class that you can reuse. False negative in case the type is a boxed type, ie. `java.lang.Integer`.

PR #585, opened 2016-01-08
Added support for String constants as parameter to @SuppressWarnings

mbring:
Since I'm new to the code there might be a smarter/cleaner way to solve the problem and I would happily do any changes that you suggest.

I also modified the existing test file JavaSuppressWarning.java but if you would rather want a separate file (with a separate test) I can fix that.

### DIFF
--- a/java-squid/src/main/java/org/sonar/java/JavaFilesCache.java
+++ b/java-squid/src/main/java/org/sonar/java/JavaFilesCache.java
@@ -33,6 +33,7 @@ import org.sonar.java.signature.MethodSignatureScanner;
 import org.sonar.java.syntaxtoken.LastSyntaxTokenFinder;
 import org.sonar.plugins.java.api.JavaFileScanner;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.tree.AnnotationTree;
 import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.ClassTree;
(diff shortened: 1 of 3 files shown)


benzonico:
Hi and thanks for the contribution, 

This is a nice idea but two remarks on your implementation : 

  -  You are not handling the case where the constant is used with its fully qualified name: `MyClass.FOO`

- This will only work if the constant is defined in the same source file as the one it is used (if not the declaration of variable will be null). We could think of having this implementation with this limitation but I think this might be rather confusing for end user (suppress warnings will partially work with a constant). 

The "ideal" way would be to get the value of a constant when we solve it but this requires to improve semantic analysis on the reading of bytecode part (which is a bit more complex part).

mbring:
Thanks for reviewing this!

The idea was of course that constants in other files could be used, otherwise it's kind of pointless and I agree that including a partially working solution is just confusing. I was hoping that my solution would somehow work even for these cases but since I'm not so familiar with the code my implementation fell short...

The first issue I could of course fix but it doesn't really matter if the second part does not work.

Is the "ideal" solution something that I could fix? I have no idea where to start so I would need a push in the right direction.

benzonico:
@mbring Well to improve this we should probably add a value property for static final variables and complete them when reading from bytecode (see class `BytecodeCompleter`) or from the source (see class `FirstPass` ) 

This is not something we have planned right now (but that definitely could be of value). 
If you have more questions regarding this, feel free to ask. But as a consequence of the limitation we have raised in the preceding conversation I am closing this pull request.

mbring:
Thanks, for the feedback! I might look into it when I get a chance.

alban-auzeill:
Hi @fupgang,
I can not reproduce. String constants with `@SuppressWarnings` should be supported. It has been part of this improvement: [SONARJAVA-2492](https://sonarsource.atlassian.net/browse/SONARJAVA-2492).
You are probably facing a different issue. Could you try to double-check your issue and if needed open a new topic on [Sonar community](https://community.sonarsource.com/c/clean-code/fp/7)?

PR #595, opened 2016-01-14
SONARJAVA-1472 Catch only rethrowing

benzonico:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S2737.html
```diff
@@ -0,0 +1,14 @@
+<p>A <code>catch</code> clause that only rethrows the caught exception has the same effect as omitting the <code>catch</code> altogether and letting it bubble up automatically, but with more code and the additional detrement of leaving maintainers scratching their heads. </p>
```
m-g-sonar: description should be changed.

PR #605, opened 2016-01-19
SONARJAVA-1483 RSPEC-3282 EJB interceptor exclusions should be declared as annotations

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/xml/ejb/InterceptorExclusionsCheck.java
```diff
@@ -0,0 +1,73 @@
...
+    try {
+      NodeList interceptorBindings = context.evaluateXPathExpression("ejb-jar/assembly-descriptor/interceptor-binding");
+      for (int i = 0; i < interceptorBindings.getLength(); i++) {
+        Node interceptorBinding = interceptorBindings.item(i);
+        if (!isDefaultInterceptor(context, interceptorBinding)) {
+          checkExclusions(context, interceptorBinding);
+        }
+      }
+    } catch (XPathExpressionException e) {
+      throw new IllegalArgumentException("[S3282] Unable evaluate xpath expression for file " + context.getFile().getAbsolutePath(), e);
```
benzonico: Why illegal argument ?? ?

java-checks/src/main/java/org/sonar/java/checks/xml/ejb/InterceptorExclusionsCheck.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+  }
+
+  private void checkExclusions(XmlCheckContext context, Node interceptorBinding) throws XPathExpressionException {
+    NodeList exclusions = context.evaluateXPathExpressionFromNode(
+      interceptorBinding,
+      "*[self::exclude-default-interceptors[text()=\"true\"] or self::exclude-class-interceptors[text()=\"true\"]]");
+    for (int i = 0; i < exclusions.getLength(); i++) {
```
benzonico: Why not a for each loop ?

PR #612, opened 2016-01-20
SONARJAVA-1458 Ignore fields updated in init methods

benzonico:


java-checks/src/main/java/org/sonar/java/checks/ServletInstanceFieldCheck.java
```diff
@@ -47,17 +53,56 @@
...
 
   @Override
   public void visitNode(Tree tree) {
-    VariableTree variable = (VariableTree) tree;
-    if (hasSemantic() && isOwnedByAServlet(variable) && !isStaticOrFinal(variable)) {
+    if (tree.is(Kind.METHOD) && isServletInit((MethodTree) tree)) {
+      tree.accept(new AssignmentVisitor());
+    } else if (tree.is(Kind.VARIABLE)) {
+      VariableTree variable = (VariableTree) tree;
+      if (hasSemantic() && isOwnedByAServlet(variable) && !isStaticOrFinal(variable)) {
```
m-g-sonar: checking for the semantic should be done before checking the type of the tree, as semantic is used to get the parameter type of the method.

PR #620, opened 2016-01-25
SONARJAVA-1462 Log message of report not found for surefire only when…

benzonico:
… property has been set by user

### DIFF
--- a/java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java
+++ b/java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java
@@ -58,19 +58,20 @@ public class SurefireJavaParser implements BatchExtension {
     this.javaResourceLocator = javaResourceLocator;
   }
 
-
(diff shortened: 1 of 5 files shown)


java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java
```diff
@@ -58,19 +58,20 @@ public SurefireJavaParser(ResourcePerspectives perspectives, JavaResourceLocator
     this.javaResourceLocator = javaResourceLocator;
   }
 
-
-  public void collect(SensorContext context, File reportsDir) {
-    File[] xmlFiles = getReports(reportsDir);
+  public void collect(SensorContext context, File reportsDir, boolean reportDirSetByUser) {
```
m-g-sonar: In order to avoid a method with a flag parameter, I would prefer to have 2 public methods such has "collectFromUserReportDir" and "collectFromDefault" for instance, and make this method private.

PR #629, opened 2016-01-26
SONARJAVA-1490 Fix exit path evaluation for finally with branches

benzonico:


DidierBesset:
I would prefer to have the field "exitPath" hidden and accessed via getter/setter. This is easier to track changes since it is not final

PR #631, opened 2016-01-26
SONARJAVA-1436 RSPEC-3421 Deprecated "${pom}" properties should not be used

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/xml/maven/DeprecatedPomPropertiesCheck.java
```diff
@@ -0,0 +1,78 @@
...
+          String property = extractPropertyName(text);
+          reportIssue(textNode, "Replace \"pom." + property + "\" with \"project." + property + "\".");
+          text = skipFirstProperty(text);
+        }
+      }
+    }
+  }
+
+  private static String skipFirstProperty(String text) {
+    String newText = text.substring(text.indexOf(POM_PROPERTY_PREFIX));
```
benzonico: no need to do two substring, you should search for index of suffix starting at index of prefix

PR #638, opened 2016-01-28
SONARJAVA-1299 ignores parameters annotated with @Observes

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
```diff
@@ -79,6 +87,15 @@ public void visitNode(Tree tree) {
     }
   }
 
+  private static boolean hasAuthorizedAnnotation(List<AnnotationTree> annotations) {
+    for (AnnotationTree annotationTree : annotations) {
```
benzonico: use isAnnotatedWith

PR #639, opened 2016-01-28
SONARJAVA-1488-qualflaw - Coverage improved, bug fixed in combination of equals relation

DidierBesset:


java-squid/src/main/java/org/sonar/java/se/symbolicvalues/BinaryRelation.java
```diff
@@ -159,7 +159,7 @@ protected RelationState resolveState(Collection<BinaryRelation> knownRelations,
    * @return a SymbolicValueRelation or null if the receiver and the supplied relation cannot be combined
    */
   @CheckForNull
-  private BinaryRelation combineUnordered(BinaryRelation relation) {
+  protected BinaryRelation combineUnordered(BinaryRelation relation) {
```
m-g-sonar: add ``@VisibleForTest``.

PR #648, opened 2016-02-09
SONARJAVA-1521 Fix message of S3398 for anonymous classes

benzonico:


java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
```diff
@@ -92,12 +94,22 @@ private boolean isPrivateMethodOfOuterClass(Symbol symbol) {
...
-          }
+          reportIssueOnMethod((MethodTree) methodUsed.declaration());
         }
       }
     }
+
+    private void reportIssueOnMethod(@Nullable MethodTree declaration) {
+      if (declaration != null) {
+        String message = "Move this method into ";
+        if (classSymbol.name().isEmpty()) {
```
m-g-sonar: I would really prefer to identify the class as being an anonymous class in the constructor of {{MethodInvocationVisitor}}, with a dedicated {{boolean}} variable, instead of rechecking every time an issue should be reported.

PR #664, opened 2016-02-16
SONARJAVA-1526 - SE: Assignement with operator should generate a new …

DidierBesset:
…SV and not behave like a standard assignement


java-squid/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -67,6 +67,7 @@
 import org.sonar.plugins.java.api.tree.WhileStatementTree;
 
 import javax.annotation.Nullable;
+
```
benzonico: Almost perfect commit ! :) (but ok, I'm being picky here).

PR #665, opened 2016-02-16
SONARJAVA-1516 - FP in AssignmentInSubExpressionCheck : assignement i…

DidierBesset:
…n while statement are not ignored


java-checks/src/main/java/org/sonar/java/checks/AssignmentInSubExpressionCheck.java
```diff
@@ -130,6 +131,11 @@ private static boolean isRelationalExpression(Tree tree) {
   }
 
   @Override
+  public void visitWhileStatement(WhileStatementTree tree) {
```
benzonico: This is way too broad and will results in false negative :

{code}
while(true) {
   if ((a = b = 0) != null) { // Noncompliant but not detected.
 }
{code}

PR #672, opened 2016-02-21
S1118 Add support for UtilityClass lombok annotations

wuwen5:
https://projectlombok.org/features/experimental/UtilityClass.html


m-g-sonar:
Hello @wuwen5,

First, thank you for your contribution and the quality of the code you submitted!

Unfortunately, we can not accept your PR at the moment. If we usually try to take into account lombok annotations in our rules, we only rely on validated features. The `UtilityClass` annotation is still experimental, and consequently subject to changes. As those potential changes may have possible impact on the expected behavior of the rule, we simply cannot add it yet.

Don't hesitate to come back to us when the feature will be validated!

Regards,
Michael

PR #676, opened 2016-02-23
SONARJAVA-1551 RSPEC-3034 Raw byte values should not be used in bitwise operations in combination with shifts

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/RawByteBitwiseOperationsCheck.java
```diff
@@ -0,0 +1,107 @@
...
+  List<ExpressionTree> shifts = new LinkedList<>();
+  List<ExpressionTree> byteSecuring = new LinkedList<>();
+
+
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    this.context = context;
+    scan(context.getTree());
+    shifts.clear();
+    byteSecuring.clear();
```
DidierBesset: How about byteContainments? Especially an "s" is missing.

java-checks/src/main/java/org/sonar/java/checks/RawByteBitwiseOperationsCheck.java
```diff
@@ -0,0 +1,107 @@
...
+    if (isIntegerOrLongExpected(tree.symbolType())) {
+      ExpressionTree leftOperand = ExpressionsHelper.skipParentheses(tree.leftOperand());
+      ExpressionTree rightOperand = ExpressionsHelper.skipParentheses(tree.rightOperand());
+      checkShiftWithoutByteSecuring(leftOperand, rightOperand);
+      checkShiftWithoutByteSecuring(rightOperand, leftOperand);
+    }
+  }
+
+  private static boolean isShifting(BinaryExpressionTree tree) {
+    return tree.is(Tree.Kind.LEFT_SHIFT, Tree.Kind.RIGHT_SHIFT, Tree.Kind.UNSIGNED_RIGHT_SHIFT);
```
DidierBesset: Not needed for unsigned shifts?

m-g-sonar: Yes, because the issue is not with the shift itself, but with the addition of a signed byte , not protected, afterward.

PR #681, opened 2016-02-25
SONARJAVA-1538 - CFG SE : nested statements in try catch end up with …

DidierBesset:
…wrong CFG


benzonico:
We shall reopen this PR when we actually work on the ticket.

PR #685, opened 2016-02-29
SONARJAVA-716 Use EnumSet instead of Set of Enums

benzonico:


java-checks/src/main/java/org/sonar/java/checks/EnumSetCheck.java
```diff
@@ -0,0 +1,82 @@
...
+import org.sonar.plugins.java.api.tree.Tree.Kind;
+import org.sonar.plugins.java.api.tree.VariableTree;
+import org.sonar.squidbridge.annotations.SqaleConstantRemediation;
+import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
+
+import java.util.List;
+
+@Rule(
+  key = "S1641",
+  name = "Sets with keys that are enum values should be replaced with EnumSet",
```
m-g-sonar: I wonder if it should not rather be ``elements`` instead of ``keys``.

PR #722, opened 2016-03-11
SONARJAVA-1583 Add scope for switch statements

benzonico:


java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
```diff
@@ -765,6 +765,12 @@ public void try_with_resources() {
     assertThat(result.symbol("foo3", 7)).isSameAs(result.reference(8, 7));
   }
 
+  @Test
+  public void switch_statement() {
+    Result result = Result.createFor("SwitchStatement");
+
```
m-g-sonar: This test should contain assertions. At least to verify the existence of the variable a in each scope.

PR #728, opened 2016-03-14
SONARJAVA-1225 improve handling of collections

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -79,14 +83,47 @@ public void visitNode(Tree tree) {
...
+  }
+
+  private static boolean implementsCollectionButNotSerializable(Type type) {
+    return !type.symbol().isInterface() && isSubtypeOfCollectionApi(type) && !implementsSerializable(type);
+  }
+
+  private void checkCollectionAssignments(List<IdentifierTree> usages) {
+    for (IdentifierTree usage : usages) {
+      Tree parentTree = usage.parent();
+      if (parentTree instanceof AssignmentExpressionTree) {
```
benzonico: ... I understand the why but this goes against the use of kinds.

benzonico: and so should be fixed.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -79,14 +83,47 @@ public void visitNode(Tree tree) {
...
+        reportIssue(simpleName);
+      }
     }
   }
 
+  private static boolean initializerNotImplementingSerializable(ExpressionTree initializer) {
+    return initializer != null && implementsCollectionButNotSerializable(initializer.symbolType());
+  }
+
+  private static boolean implementsCollectionButNotSerializable(Type type) {
```
benzonico: what about isUnserializableCollection ? bit shorter.

PR #734, opened 2016-03-17
SONARJAVA-1571 Handle spring assertions 'ResulltActions.andExpect()'

m-g-sonar:


java-checks/src/test/java/org/sonar/java/checks/AssertionsInTestsCheckTest.java
```diff
@@ -24,24 +24,31 @@
 
 public class AssertionsInTestsCheckTest {
 
+  private static final AssertionsInTestsCheck CHECK = new AssertionsInTestsCheck();
```
benzonico: Why static final ?

PR #751, opened 2016-03-24
SONARJAVA-1613 do not run check without semantic

m-g-sonar:


benzonico:
Title of PR is really ambiguous: I thought at first that this was general and not only for one check.

benzonico:
I would use "run check" rather than "perform analysis" which seems global.

m-g-sonar:
Indeed! Sorry for the misleading title of the PR/commit!

PR #752, opened 2016-03-25
Handle lombok annotations generating equals() and hashCode()

thomasturrell:
These classes should be compliant since Lombok annotation generate equals() and hashCode()

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverridenWithCompareToCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverridenWithCompareToCheck.java
@@ -23,9 +23,10 @@ import com.google.common.collect.ImmutableList;
 import org.sonar.api.server.rule.RulesDefinition;
 import org.sonar.check.Priority;
 import org.sonar.check.Rule;
-import org.sonar.java.model.declaration.MethodTreeImpl;
 import org.sonar.java.tag.Tag;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.Symbol;
+import org.sonar.plugins.java.api.semantic.SymbolMetadata;
 import org.sonar.plugins.java.api.semantic.Type;
 import org.sonar.plugins.java.api.tree.ClassTree;
 import org.sonar.plugins.java.api.tree.MethodTree;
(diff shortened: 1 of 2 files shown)


m-g-sonar:
Hello @thomasturrell,

Thanks for the contribution, as well as your comments on our [google group](https://groups.google.com/d/topic/sonarqube/bxWTWS5tnrk/discussion). 
This PR is now related to JIRA ticket : [SONARJAVA-1614](https://jira.sonarsource.com/browse/SONARJAVA-1614)

If you want to push a little bit further your contribution, for instance by trying fixing the rule directly, you may want to have a look at what was done for resolution of SONARJAVA-1513. The same approach as in the following PR https://github.com/SonarSource/sonar-java/pull/684 cab be used to solve the issue! Otherwise, we will deal with your PR when handling the ticket.

Cheers,
Michael

thomasturrell:
Hi @m-g-sonar 

My first thought is that to avoid duplicating code I should move the methods shouldImplementEquals and generatesEquals from EqualsNotOverriddenInSubclassCheck to somewhere else. Do you have a preference where? I was thinking of adding them to org.sonar.java.checks.helpers.MethodsHelper.

Failing that I can simply duplicate the code no problems.

m-g-sonar:
For the time being, duplicating the code won't be a problem.

On our side, we are investigating other ways to handle framework such as lombok. The idea being to ignore issues from multiple rules, instead of having to modify their implementation each time. Of course, as long as this isn't implemented, rules will still have to be modified.

java-checks/src/main/java/org/sonar/java/checks/EqualsNotOverridenWithCompareToCheck.java
```diff
@@ -56,35 +63,58 @@
...
 
-      if (compare != null && !hasEquals) {
+      if (compare != null && !generatesEquals(classTree) && !implementsEquals(classTree)) {
         reportIssue(compare.simpleName(), "Override \"equals(Object obj)\" to comply with the contract of the \"compareTo(T o)\" method.");
       }
     }
   }
 
+  private static boolean implementsEquals(ClassTree classTree) {
+      return hasNotFinalEqualsMethod(classTree.symbol());
```
m-g-sonar: As you removed the part where we look into the hierarchy of the current class (originally needed in `EqualsNotOverridenInSubclassCheck`), I think it's simpler to stay with the initial approach based on the `hasEquals` test.

m-g-sonar:
hey @thomasturrell,

I did some comments on your PR that you should take into account.
Nice to see that the CI is now green and everything is going fine.

Please squash your commits into a single commit, come back to the initial approach to check equals method, and I'll merge it!

m-g-sonar:
Completed in https://github.com/SonarSource/sonar-java/pull/752

PR #760, opened 2016-03-31
SONARJAVA-1621 resolve methods of bounds of a type variable

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/resolve/JavaSymbol.java
```diff
@@ -640,14 +641,26 @@ public TypeVariableJavaSymbol(String name, JavaSymbol owner) {
     @Override
     @Nullable
     public JavaType getSuperclass() {
-      // FIXME : should return upper bound or Object if no bound defined.
+      JavaType firstBound = bounds().get(0);
```
benzonico: No need to invoke the method here.

java-frontend/src/main/java/org/sonar/java/resolve/JavaSymbol.java
```diff
@@ -640,14 +641,26 @@ public TypeVariableJavaSymbol(String name, JavaSymbol owner) {
...
+      if (!firstBound.getSymbol().isInterface()) {
+        return firstBound;
+      }
       return null;
```
benzonico: shouldn't it be Object instead of null ? as you can call toString or hashcode on a variable typed by a type parameter

m-g-sonar:
merged.

PR #771, opened 2016-04-06
SONARJAVA-1595 readResolve should be visible to subclasses

benzonico:


java-checks/src/main/java/org/sonar/java/checks/serialization/PrivateReadResolveCheck.java
```diff
@@ -0,0 +1,70 @@
...
+import org.sonar.squidbridge.annotations.SqaleSubCharacteristic;
+
+import java.util.List;
+
+@Rule(
+  key = "S2062",
+  name = "\"readResolve\" methods should be inheritable",
+  priority = Priority.MAJOR,
+  tags = {Tag.PITFALL})
+@SqaleSubCharacteristic(RulesDefinition.SubCharacteristics.DATA_RELIABILITY)
```
m-g-sonar: The rule should be activated by default.

m-g-sonar: The rule should be activated by default.

java-checks/src/main/java/org/sonar/java/checks/serialization/PrivateReadResolveCheck.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  private static boolean isPrivateReadResolve(MethodTree method) {
+    return "readResolve".equals(method.simpleName().name()) && method.parameters().isEmpty() && ModifiersUtils.hasModifier(method.modifiers(), Modifier.PRIVATE);
+  }
+
+  private static boolean isOwnedByExtensibleClass(MethodTree method) {
+    Symbol owner = method.symbol().owner();
+    return !owner.isPrivate() && !owner.isFinal();
```
m-g-sonar: I would still check if the owner implements serializable. It would be sad to raise FP, even if they are unlikely.

PR #787, opened 2016-04-14
SONARJAVA-958 Unused private methods should be migrated to semantic API

benzonico:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -19,58 +19,39 @@
...
-          if (StringUtils.isEmpty(paramName)) {
-            paramName = MethodSignatureScanner.getReadableType(param.getJvmJavaType());
-          }
-          params.add(paramName + (param.isArray() ? "[]" : ""));
+  public void visitNode(Tree tree) {
+    MethodTree node = (MethodTree) tree;
+    Symbol symbol = node.symbol();
+    if (symbol.isPrivate() && symbol.usages().isEmpty()) {
+      if (node.is(Tree.Kind.CONSTRUCTOR)) {
+        if (!node.parameters().isEmpty()) {
```
m-g-sonar: this test should be part of the previous if condition.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -19,58 +19,42 @@
...
-          params.add(paramName + (param.isArray() ? "[]" : ""));
+  public void visitNode(Tree tree) {
+    if(!hasSemantic()) {
+      return;
+    }
+    MethodTree node = (MethodTree) tree;
+    Symbol symbol = node.symbol();
+    if (node.modifiers().annotations().isEmpty() && symbol.isPrivate() && symbol.usages().isEmpty()) {
+      if (node.is(Tree.Kind.CONSTRUCTOR)) {
+        if (!node.parameters().isEmpty()) {
```
m-g-sonar: this test should be part of the previous if condition.

java-checks/src/test/files/checks/UnusedPrivateMethod.java
```diff
@@ -0,0 +1,131 @@
...
+class Lambdas {
+  void method(){
+    IntStream.range(1, 5)
+      .map((x)-> x*x )
+      .map(x -> x * x)
+      .map((int x) -> x * x)
+      .map((x)-> x*x )
+    ;
+  }
+
```
m-g-sonar: I would also add a test with a method reference.

PR #792, opened 2016-04-19
SONARJAVA-1648 - NPE on S2637 when return statement has no expression

DidierBesset:


java-frontend/src/main/java/org/sonar/java/se/checks/NonNullSetToNullCheck.java
```diff
@@ -223,6 +224,9 @@ public void visitReturnStatement(ReturnStatementTree tree) {
     }
 
     private boolean isLocalExpression(ExpressionTree expression) {
+      if (expression == null) {
```
m-g-sonar: If you really want to put the test within the method, then add the `@Nullable`  annotation to the parameter.

PR #803, opened 2016-05-02
SONARJAVA-1594 New rule: getter and setter should be sync in pairs

benzonico:


java-checks/src/main/java/org/sonar/java/checks/SyncGetterAndSetterCheck.java
```diff
@@ -0,0 +1,129 @@
...
+    checkMethodTree(methodTree, GETTER_BOOLEAN, SETTER, "set");
+    checkMethodTree(methodTree, SETTER, GETTER, "get");
+    checkMethodTree(methodTree, SETTER, GETTER_BOOLEAN, "is");
+  }
+
+  private void checkMethodTree(MethodTree methodTree, GetSetPredicate ownPredicate, GetSetPredicate pairPredicate, String startsWith) {
+    if (isSynchronized(methodTree) && ownPredicate.apply(methodTree)) {
+      // Synchronized getter, lookup the setter.
+      Symbol.TypeSymbol owner = (Symbol.TypeSymbol) methodTree.symbol().owner();
+      Collection<Symbol> pairedMethods = owner.lookupSymbols(startsWith + methodTree.symbol().name().substring(ownPredicate.getStartName().length()));
```
m-g-sonar: the string `startsWith` can be removed to only rely on the `pairPredicate.getStartName()`, which does exactly the same.

PR #804, opened 2016-05-03
SONARJAVA-1664 Fix fp on RSPEC-1994 when condition is empty

benzonico:


java-checks/src/test/files/checks/ForLoopIncrementAndUpdateCheck.java
```diff
@@ -37,6 +37,11 @@ void foo() {
     for (Enumeration serverIds = db.serverTable.keys(); serverIds.hasMoreElements();){} // Compliant
     for (ExecutionState current = state; !current.equals(this); current = current.parentState) {} // Compliant
     for (Integer integer = 0; integer.longValue() < 100L ; integer += 1) {} // Compliant
+    for (int i = 2; ; i++) { // false positive
```
m-g-sonar: Should be `Compliant` rather than `false positive`.

PR #814, opened 2016-05-06
Added rule for slf4j log messages checks

SPeeSimon:
Hi,

This new rule is to check if a log message for SLF4j Logger:
- Is not a method call
- Is not a String.format() method call
- Is not an Object concatenating

### DIFF
--- a/java-checks/pom.xml
+++ b/java-checks/pom.xml
@@ -163,6 +163,12 @@
                   <type>jar</type>
                 </artifactItem>
                 <artifactItem>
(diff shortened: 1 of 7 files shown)


benzonico:
Thanks for the pull request : I created the ticket to implement the rule on java plugin side: https://jira.sonarsource.com/browse/SONARJAVA-1788 

As the rule will cover way more than what your implementation is doing I am closing this pull request but still link it to the ticket for help/inspiration of testcases. 

Thanks again for the contribution.

PR #818, opened 2016-05-09
SONARJAVA-1599 RSPEC-3020 'toArray' should be passed an array of the proper type

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/ToArrayCheck.java
```diff
@@ -0,0 +1,69 @@
...
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.TypeCastTree;
+
+import java.util.List;
+
+@Rule(key = "S3020")
+public class ToArrayCheck extends IssuableSubscriptionVisitor {
```
benzonico: Should probably use a AbstractMethodInvocationMatcher

PR #844, opened 2016-05-30
SONARJAVA-1701 SONARJAVA-1702 Fix generics issues

benzonico:


java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
```diff
@@ -105,7 +105,8 @@ public void Generics() {
     JavaSymbol applyMethod = result.symbol("apply");
     assertThat(result.reference(83, 12)).isSameAs(applyMethod);
     assertThat(result.reference(89, 61)).isSameAs(applyMethod);
-    assertThat(applyMethod.usages()).hasSize(2);
+    // A method that does not compile is resolved whereas it should not.
```
m-g-sonar: I would add an assertions with the reference line, to indicate which method call is wrongly resolved.

PR #848, opened 2016-06-02
Adding support for Guava VisibleForTesting annotation

cardil:
This PR adds support for for Guava's `@VisibleForTesting` annotation and will no longer raise issues on protected members that have been annotated with this annotation as well as default visibility (package). This annotation indicates that the visibility of a type or member has been purposely relaxed to make the code testable.

```java
public final class MyFinalClass {

  @VisibleForTesting
  protected Logger logger; // Compliant, no S2156 rule failed
  @VisibleForTesting
  protected int calculateSomethingComplex(String input) { // Compliant, no S2156 rule failed
   // ...
  }
}

class Cone {
  @VisibleForTesting
  Logger logger; // Compliant, no S2039 rule failed
}

class MyFinalClassTest {
  @Test
  public void test() {
    MyFinalClass my = new MyFinalClass();
    my.logger = mock(Logger.class);
    assertThat(my.calculateSomethingComplex(INPUT_JSON)).isEqualTo(42);
  }
}
```

More info: http://eng.wealthfront.com/2011/12/02/beyond-javas-access-control-visibility


cardil:
It's strange - it seams to pass tests on "My" Travis - here: https://travis-ci.org/wavesoftware/sonar-java/builds/137741688

m-g-sonar:
Hey @cardil,

Thanks for your contribution. Your job is passing because on "your" travis, because the "Ruling" is not activated. If you look at all the other jobs logs, except for the CI, you'll see this line:
```
================ plugin or ruling tests are only run on pull requests! =================
```
In our jobs, we are executing special ITs which are playing the rules against several java projects, and comparing the results (the RULING jobs). Have a look at the [README](https://github.com/SonarSource/sonar-java/blob/master/README.md) in order to get the it source projects and try execute the ruling locally. You'll have to validate the differences for each of the project.

.editorconfig
```diff
@@ -0,0 +1,12 @@
+# EditorConfig is awesome: http://EditorConfig.org
```
m-g-sonar: Please remove this file from your commit.

java-checks/src/main/java/org/sonar/java/checks/predicates/VisibleForTestingPredicate.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * This is a predicate that finds if given modifiers tree contains {@link VisibleForTesting} annotation.
+ *
+ * @author <a href="mailto:krzysztof.suszynski@coi.gov.pl">Krzysztof Suszynski</a>
+ * @since 07.06.16
+ */
+public class VisibleForTestingPredicate implements Predicate<ModifiersTree> {
+
+  public static final String GUAVA_FQCN = VisibleForTesting.class.getName();
```
m-g-sonar: Please use directly the fully qualified name as String.

java-checks/src/main/java/org/sonar/java/checks/FieldModifierCheck.java
```diff
@@ -31,10 +32,12 @@
 import org.sonar.plugins.java.api.tree.VariableTree;
 
 import java.util.List;
+import java.util.function.Predicate;
 
 @Rule(key = "S2039")
 public class FieldModifierCheck extends IssuableSubscriptionVisitor {
 
+  private static final Predicate<ModifiersTree> MODIFIERS_TREE_PREDICATE = new VisibleForTestingPredicate().negate();
```
m-g-sonar: Could you write your predicate using a lambda instead?
Even if used at 2 different places, it would be cleaner from our point of view. And no need to introduce a new class with its unit test, as it can be unit tested using the check verification.

java-checks/src/main/java/org/sonar/java/checks/predicates/VisibleForTestingPredicate.java
cardil: Yup. I've noticed that when pass it via String test validator can have `!UnknownSymbol` instead of correct `AnnotationTree` structure (while debugging the test). It must be due to yours custom lexer, which for some reason do not load `VisibleForTesting` class when encountered at test file like here: https://github.com/SonarSource/sonar-java/pull/848/files#diff-02c2ce19b6675f863c2cbecc70b95c23R15

java-checks/src/main/java/org/sonar/java/checks/FieldModifierCheck.java
cardil: I'm not a fan of integration tests. I mean by that to test more then one class at a time. It's just the approach you are using `JavaCheckVerifier`. That way is hard to cover all routes on your code and it's also less strait forward. Decoupling the parts of the logic the way I've written is easier to test.

m-g-sonar: Tests of rules are not integration tests, but unit test. Each rule is tested unitarilly, with expected behavior. You don't need to try every possible combination of modifiers, but only the presence or not of the annotation.

I can however tell you that mocking absolutely everything which is generated by the parsing is not a better approach in order to test the behavior of your predicate (which could be a simple lambda expression on a stream of ModifierTree).

cardil: There are integration test just because they use multiple classes. In fact the whole java parser/lexer is used. That's not unit testing.

I don't mind you like integration tests better then unit testing. In fact many companies/developers like that. It's most effective to reach 80% coverage  in that way,  bit extreamly difficult for 100% of conditions. Unit testing let you reach 100% quite easily because you focus on every edge case in single class so you know that class is surly working. But still you need integration test to verify yourself between multiple classes. 

I mind copy pasting of my code. Even if it's 10 lines. I'm pretty much sure sonar could detect that as copy paste. :-/

m-g-sonar: OK, let's put aside the unit/integration part of the discussion.

After thinking a bit more about your approach, I realized that there is a simpler way to check presence of the annotation, by relying 100% on the semantic API (and fully unit tested! OK, I stop.)

In the `isNotVisibleForTesting` method, instead of using your predicate to test the presence of the annotation, you should do the exact same thing using the symbol metadata. I'll add a comment there.

java-checks/src/main/java/org/sonar/java/checks/FieldModifierCheck.java
```diff
@@ -44,17 +47,27 @@
...
+
   private static boolean hasNoVisibilityModifier(VariableTree member) {
     ModifiersTree modifiers = member.modifiers();
     return !(ModifiersUtils.hasModifier(modifiers, Modifier.PUBLIC)
       || ModifiersUtils.hasModifier(modifiers, Modifier.PRIVATE)
       || ModifiersUtils.hasModifier(modifiers, Modifier.PROTECTED));
   }
+
+  private static boolean isNotVisibleForTesting(VariableTree member) {
+    return MODIFIERS_TREE_PREDICATE.test(member.modifiers());
```
m-g-sonar: you can replace this call to your predicate by this: `!member.symbol().metadata().isAnnotatedWith("com.google.common.annotations.VisibleForTesting")`

Cosaquee:
+1

cardil:
Tests are now passing

travis.sh
```diff
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 function configureTravis {
-  mkdir ~/.local
+  mkdir -p ~/.local
```
m-g-sonar: Why the `-p`?

cardil: `-p` will create the directory path recursively making sure the whole path is created and also will not fail if directory is already created. It's more safe and cross machine this way. `~/.local` directory is being used by other software and in that way I can run this on my machine for example.

cardil:
Is there something else to fix with this PR, or everything is okey now?

m-g-sonar:
Everything is alright, I'll merge it today. I just have to update the specifications of the rules accordingly to your changes.

m-g-sonar:
I rebased your PR and squashed your commits into a single one. I also slightly reworked the descriptions of the rules. Your commit is now part of this PR: https://github.com/SonarSource/sonar-java/pull/896

As soon as our CI and QA are green, I'll merge and close this PR.

Thanks a lot for your contribution. Your  improvement of the rule will be part of version 4.0 of the Java Plugin, soon to be released.

cardil:
@m-g-sonar Awesome! :smile:

m-g-sonar:
Merged! Your contribution is now on master, in this commit: https://github.com/SonarSource/sonar-java/commit/7a47f7ab9b329a0b853ceb6baa0efe99900b4245

PR #849, opened 2016-06-03
SONARJAVA-1712 fix reporting on lambdas

m-g-sonar:


java-frontend/src/main/java/org/sonar/plugins/java/api/JavaFileScannerContext.java
```diff
@@ -75,7 +76,9 @@
 
   void reportIssue(JavaCheck javaCheck, Tree tree, String message, List<Location> secondaryLocations, @Nullable Integer cost);
 
-  void reportIssue(JavaCheck javaCheck, Tree starTree, Tree endTree, String message);
+  void reportIssue(JavaCheck javaCheck, Tree startTree, Tree endTree, String message);
+
+  void reportIssue(JavaCheck javaCheck, Tree startTree, Tree endTree, String message, List<Location> secondaryLocations);
```
benzonico: This is debatable API design as things are getting a bit messed up with the other methods... Moreover, it seems you add a method you don't really need right ahead.

m-g-sonar: Ok, adding `@Nullable` cost to match other API. First method already exists, I just fixed a typo.

PR #864, opened 2016-06-13
SONARJAVA-1722 empty substitution from parameter is not invalid

benzonico:


java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
```diff
@@ -973,7 +973,7 @@ public void Lambdas() throws Exception {
 
 
     JavaSymbol sym = result.symbol("o");
-    assertThat(sym.type.toString()).isEqualTo("!Defered type!"); // unknown while we do the analysis with java 7
+    assertThat(sym.type.toString()).isEqualTo("Object");
```
m-g-sonar: It would be better to use `.is("java.lang.Object")).isTrue();` rather than testing return value of the `toString()` method.

PR #868, opened 2016-06-14
SONARJAVA-1500 No constraints on volatile fields

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -179,12 +180,22 @@ public String toString() {
 
   public ProgramState addConstraint(SymbolicValue symbolicValue, Constraint constraint) {
     PMap<SymbolicValue, Constraint> newConstraints = constraints.put(symbolicValue, constraint);
-    if (newConstraints != constraints) {
+    if (newConstraints != constraints && !isSymbolicValueOfVolatileField(symbolicValue)) {
```
benzonico: Wondering if it wouldn't not be better to test for "volatileness" of SV prior to adding the constraints.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -179,12 +180,22 @@ public String toString() {
...
       return new ProgramState(this, newConstraints);
     }
     return this;
   }
 
+  private boolean isSymbolicValueOfVolatileField(SymbolicValue symbolicValue) {
```
benzonico: What is tested here is that there are no volatile symbol that can reach this SV whereas this is not exactly what the ticket is supposed to solve : the ticket is about a specific symbol that is volatile and thus its SV may have changed (but not the constraints on the currently assigned SV).

benzonico:
```
class VolatileFields {
   private volatile boolean volatileField = false;
   private boolean a = false;
 
   void foo() {
     a = volatileField;
     if (volatileField) {
       return;
     }
     while (true) {
       if (a) { // NonCompliant as this field is nonvolatile, it cannot be modified by another thread
         System.out.println();
       }
     }
   }
 }
```

PR #869, opened 2016-06-14
Fix quality flaw: update Sensor to new API in XmlFileSensor

m-g-sonar:


sonar-java-plugin/src/main/java/org/sonar/plugins/java/XmlFileSensor.java
```diff
@@ -43,22 +42,28 @@ public XmlFileSensor(SonarComponents sonarComponents, FileSystem fs) {
   }
 
   @Override
-  public boolean shouldExecuteOnProject(Project project) {
-    return fs.hasFiles(xmlFilePredicate);
+  public String toString() {
```
benzonico: Why adding a toString and not inlining it in the descriptor ??

sonar-java-plugin/src/main/java/org/sonar/plugins/java/XmlFileSensor.java
```diff
@@ -43,22 +42,28 @@ public XmlFileSensor(SonarComponents sonarComponents, FileSystem fs) {
...
-    sonarComponents.registerCheckClasses(CheckList.REPOSITORY_KEY, CheckList.getXmlChecks());
-    new XmlAnalyzer(sonarComponents, sonarComponents.checkClasses()).scan(getXmlFiles());
+  public void describe(SensorDescriptor descriptor) {
+    descriptor.name(this.toString());
   }
 
-  private Iterable<File> getXmlFiles() {
-    return fs.files(xmlFilePredicate);
+  @Override
+  public void execute(org.sonar.api.batch.sensor.SensorContext context) {
```
benzonico: no need for fully qualified name.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/XmlFileSensor.java
m-g-sonar: The `toString()` method was already there (see line 61). I just moved it. As in `JavaSquidSensor` we kept it, I did the same here.

PR #875, opened 2016-06-20
SONARJAVA-1728 Only targets printStackTrace on default output

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/PrintStackTraceCalledWithoutArgumentCheck.java
```diff
@@ -38,6 +43,10 @@
 
   private final Deque<Symbol.TypeSymbol> enclosingClass = new LinkedList<>();
   private JavaFileScannerContext context;
+  private static final MethodMatcher PRINT_STACK_TRACE_MATCHER = MethodMatcher.create()
```
benzonico: What happens if no semantic is available ?

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S1148_java.html
```diff
@@ -23,3 +23,6 @@
 }
 </pre>
 
+<h2>Exceptions</h2>
+<p>This rule ignores <code>printStackTrace</code> calls with <code>Stream</code> arguments.</p>
```
benzonico: This is slightly different than what you implemented as you kept the printStackTrace(System.err) calls as non compliant.

java-checks/src/main/java/org/sonar/java/checks/PrintStackTraceCalledWithoutArgumentCheck.java
```diff
@@ -77,19 +71,7 @@ private boolean enclosingClassExtendsThrowable() {
...
-        ExpressionTree expression = mset.expression();
-        String identifier = mset.identifier().name();
-        return expression.is(Tree.Kind.IDENTIFIER)
-          && "System".equals(((IdentifierTree) expression).name())
-          && ("err".equals(identifier) || "out".equals(identifier));
-      }
-      return false;
-    }
-    return true;
+  private static boolean calledOnTypeInheritedFromThrowable(MethodInvocationTree tree) {
```
benzonico: I would pass directly the methodSelect to this method, that should be clearer than the safe implicit cast.

PR #877, opened 2016-06-20
SONARJAVA-1730 Handle unknown symbols when checking methods

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -20,6 +20,7 @@
 package org.sonar.java.se.checks;
 
 import com.google.common.collect.Lists;
+
```
benzonico: unecessary change... :)

PR #885, opened 2016-06-21
SONARJAVA-1739 Sample SonarLint IT

henryju:


m-g-sonar:
New ITs targeting integration with SonarLint are passing only with refactor of how `SensorContext` is used.

PR #906, opened 2016-06-27
SONARJAVA-398 Remove distinction between accessors and other methods

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/ast/visitors/ComplexityVisitor.java
```diff
@@ -126,16 +121,11 @@ public void visitNode(Tree tree) {
 
   private void computeMethodComplexity(MethodTree methodTree) {
     BlockTree block = methodTree.block();
-    if (block != null && (classTrees.isEmpty() || !isAccessor(methodTree))) {
+    if (block != null) {
```
benzonico: Not really sure you should remove the whole second operand. (what about `classTrees.isEmpty` ??) 
If it is correct then I guess the cleanup can go deeper and the `classTrees` field can be removed.

m-g-sonar: Removed `classTrees`, which was useless.

PR #913, opened 2016-06-30
SONARJAVA-1735 Fix default regExp to match surefire/failsafe defaults

benzonico:


java-checks/src/test/files/checks/naming/BadTestClassNameCheck.java
```diff
@@ -15,3 +15,28 @@ void foo() {}
...
+class AnIT {
+  @Test
+  void foo() {
+  }
+}
+class AnITCase {
+  @Test
+  void foo() {
+  }
+}
```
m-g-sonar: add new line

PR #917, opened 2016-07-01
SONARJAVA-1691: Reduce FN amount.

Johnnei:

- Couldn't get these to run properly, I'll see what the CI results say 😇 Hm, looks like the local results weren't weird then 😃 

Methods annotated with @SuppressWarning for unchecked and rawtypes should
not cause method variables to be considered 'needed' by external systems.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
@@ -30,20 +30,26 @@ import org.sonar.java.model.declaration.MethodTreeImpl;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
 import org.sonar.plugins.java.api.semantic.Symbol;
+import org.sonar.plugins.java.api.tree.AnnotationTree;
 import org.sonar.plugins.java.api.tree.BlockTree;
 import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.LiteralTree;
 import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.Modifier;
+import org.sonar.plugins.java.api.tree.NewArrayTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.VariableTree;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 
 @Rule(key = "S1172")
 public class UnusedMethodParameterCheck extends IssuableSubscriptionVisitor {
 
   private static final String AUTHORIZED_ANNOTATION = "javax.enterprise.event.Observes";
+  private static final String SUPPRESS_WARNINGS_ANNOTATION = "java.lang.SuppressWarnings";
+  private static final Collection<String> EXCLUDED_WARNINGS_SUPPRESSIONS = ImmutableList.of("\"rawtypes\"", "\"unchecked\"");
   private static final MethodMatcherCollection SERIALIZABLE_METHODS = MethodMatcherCollection.create(
     MethodMatcher.create().name("writeObject").addParameter("java.io.ObjectOutputStream"),
     MethodMatcher.create().name("readObject").addParameter("java.io.ObjectInputStream"));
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
```diff
@@ -80,7 +89,34 @@ private static boolean isExcluded(MethodTree tree) {
...
+    }
+
+    return tree.arguments().stream().allMatch(UnusedMethodParameterCheck::containsOnlyExcludedWarnings);
+  }
+
+  private static boolean containsOnlyExcludedWarnings(Tree tree) {
+    if (tree.is(Tree.Kind.STRING_LITERAL)) {
+      return EXCLUDED_WARNINGS_SUPPRESIONS.contains(((LiteralTree) tree).value());
+    } else if (tree.is(Tree.Kind.NEW_ARRAY)) {
+      return ((JavaTree) tree).getChildren().stream()
```
benzonico: There are two problems here : usage of stream API is a bit overkill and JavaTree is not part of the API but rather an implementation detail, so I would suggest to cast tree to NewArrayTree and have a for loop over its initializers (or use stream API here). That will end up in the same result but more explicit and without the use of downcast to JavaTree.

benzonico:
Hi, Thanks for the contribution. See my inline comments to improve it, we'll most probably integrate it as part of the ongoing sprint for version 4.1 once improved. I'll create a ticket linked to this improvement rather than reusing an old closed one.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
```diff
@@ -25,25 +25,34 @@
...
+import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 
 @Rule(key = "S1172")
 public class UnusedMethodParameterCheck extends IssuableSubscriptionVisitor {
 
   private static final String AUTHORIZED_ANNOTATION = "javax.enterprise.event.Observes";
+  private static final String SUPPRESS_WARNINGS_ANNOTATION = "java.lang.SuppressWarnings";
+  private static final Collection<String> EXCLUDED_WARNINGS_SUPPRESIONS = Collections.unmodifiableCollection(Arrays.asList("\"rawtypes\"", "\"unchecked\""));
```
benzonico: Typo in constant name.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
```diff
@@ -80,7 +89,34 @@ private static boolean isExcluded(MethodTree tree) {
...
+  private static boolean isExcludedSuppressWarningAnnotation(AnnotationTree tree) {
+    if (!tree.annotationType().symbolType().is(SUPPRESS_WARNINGS_ANNOTATION)) {
+      return false;
+    }
+
+    return tree.arguments().stream().allMatch(UnusedMethodParameterCheck::containsOnlyExcludedWarnings);
+  }
+
+  private static boolean containsOnlyExcludedWarnings(Tree tree) {
+    if (tree.is(Tree.Kind.STRING_LITERAL)) {
```
benzonico: I tend to think that this should be written as a single method (`isExcludedLiteral` maybe ?) that could be used as a method ref in the unwrapping of the newArrayTree and would avoid you write `return false`  but `return isExcludedLiteral(tree);` (bit more compact)

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
benzonico: As we are already using it here : https://github.com/SonarSource/sonar-java/pull/917/files#diff-f8c834c0aca82abca36725beddf6ab85R62 I think using ImmutableList.of would be a better choice in this case.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
```diff
@@ -80,7 +89,34 @@ private static boolean isExcluded(MethodTree tree) {
   }
 
   private static boolean isAnnotated(MethodTree tree) {
-    return !tree.modifiers().annotations().isEmpty();
+    if (tree.modifiers().annotations().isEmpty()) {
```
benzonico: As per documentation of java stream api : https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#anyMatch-java.util.function.Predicate- : this is not required as anyMatch will return false is stream is empty.

Johnnei:
@benzonico I've added a second commit to show the changes I made based on your review. If you're OK with these changes and got the ticket number I'll squash the the commits and reword the commit message.

benzonico:
@Johnnei Thanks for the fix: this LGTM. Ticket created : https://jira.sonarsource.com/browse/SONARJAVA-1770 please update the commit message, squash and I'll merge it. 
Thanks for the contribution.

Johnnei:
@benzonico Squashing done! Glad to be able to help out 😄

PR #922, opened 2016-07-12
SONARJAVA-1759 'null' should not be used with 'Optional'

ivandalbosco:


java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+
+  }
+
+  /**
+   * Returns <code>true</code> iff the specified method signature returns an <code>Optional</code>.
+   */
+  private static boolean returnsOptional(MethodTree method) {
+    TypeTree returnType = method.returnType();
+    if (returnType != null) {
+      return returnType.symbolType().is(Optional.class.getName());
```
m-g-sonar: please use the fully qualified name `java.util.Optional` instead of using `Optional.class.getName()`.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+    return false;
```
m-g-sonar: You are not covering this case in your test file (constructors).

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    public void visitClass(ClassTree tree) {
+      // don't visit subclasstree, methods in there will be visited by outer class
+    }
+
+  }
+
+  /**
+   * Returns <code>true</code> iff the specified method signature returns an <code>Optional</code>.
```
m-g-sonar: No need for the javadoc, the name of the method is explicit enough.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  private static boolean isComparisonOperator(SyntaxToken token) {
+    return JavaPunctuator.EQUAL.getValue().equals(token.text()) ||
+           JavaPunctuator.NOTEQUAL.getValue().equals(token.text());
+  }
+
+  private static boolean isOptional(ExpressionTree expression) {
+    // warning: "null" is an Optional!
+    return expression.symbolType().is(Optional.class.getName()) && !expression.is(Tree.Kind.NULL_LITERAL);
```
m-g-sonar: Please remove usage of `.class`, for the same reason as above.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+    super.visitMethod(method);
+  }
+
+  @Override
+  public void visitBinaryExpression(BinaryExpressionTree binaryExpression) {
+    String message = "Remove this null-check of an \"Optional\".";
+    if (isComparisonOperator(binaryExpression.operatorToken())) {
+      ExpressionTree left = binaryExpression.leftOperand();
+      ExpressionTree right = binaryExpression.rightOperand();
+      if (isOptional(left) && isNull(right) ||
```
SonarTech: MAJOR Add parentheses to make the operator precedence explicit. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS864)

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+          isNull(left) && isOptional(right)) {
```
SonarTech: MAJOR Add parentheses to make the operator precedence explicit. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS864)

SonarTech:
SonarQube analysis reported 2 issues
* MAJOR 2 major

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+      }
+    }
+
+    super.visitBinaryExpression(binaryExpression);
+  }
+
+  /**
+   * A visitor that raises an issue on any <code>return null</code> statement. 
+   */
+  private class ReturnVisitor extends BaseTreeVisitor {
```
m-g-sonar: You are going to raise false positives in case of a lambda defined in the method, returning null. You should skip them too.

```
class A {
  Optional<String> foo(List<String> myList) {
    myList.stream().map(s -> {
      if (s.length() > 42) {
        return null; // will raise an issue.
      }
      return s;
    });
    return Optional.empty();
  }
}
```

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+          isNull(left) && isOptional(right)) {
+        context.reportIssue(this, binaryExpression, message);
+      }
+    }
+
+    super.visitBinaryExpression(binaryExpression);
+  }
+
+  /**
+   * A visitor that raises an issue on any <code>return null</code> statement.
```
m-g-sonar: Rename your class "ReturnNullVisitor" and this javadoc can be removed.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * A visitor that raises an issue on any <code>return null</code> statement. 
+   */
+  private class ReturnVisitor extends BaseTreeVisitor {
+
+    @Override
+    public void visitReturnStatement(ReturnStatementTree returnStatement) {
+      ExpressionTree expression = returnStatement.expression();
+      if (expression.is(Kind.NULL_LITERAL)) {
```
m-g-sonar: The following case will raise an issue:
```
Optional<String> foo(List<String> myList) {
  return myList.isEmpty() ? Optional.empty() : null:
}
```

You probably needs to visit the return statement expression again.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+      method.accept(new ReturnVisitor());
+    }
+
+    super.visitMethod(method);
+  }
+
+  @Override
+  public void visitBinaryExpression(BinaryExpressionTree binaryExpression) {
+    String message = "Remove this null-check of an \"Optional\".";
+    if (isComparisonOperator(binaryExpression.operatorToken())) {
```
m-g-sonar: I would prefer to simply check the kind of the binaryExpression, with:
```
binaryExpression.is(tree.Kind.EQUAL_TO, Tree.Kind.NOT_EQUAL_TO)
```

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
m-g-sonar: This issue should have been spotted by sonarlint. Are you using it?

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,134 @@
...
+    return false;
+  }
+
+  private static boolean isComparisonOperator(SyntaxToken token) {
+    return JavaPunctuator.EQUAL.getValue().equals(token.text()) ||
+           JavaPunctuator.NOTEQUAL.getValue().equals(token.text());
+  }
+
+  private static boolean isOptional(ExpressionTree expression) {
+    // warning: "null" is an Optional!
```
m-g-sonar: No need for the warning. The `<nullType>` always return `true` when checking if it is another type. It is an expected behavor!

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: OK!

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: OK, removed.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: OK.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: I think my alignment was clearer than extra parentheses, but it's OK.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: Noted, thanks.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: Great, thanks.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: OK

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: Thanks a lot for this one!

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: My understanding is that your case if a false negative. Do you agree?

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: Noted. Thanks.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Override
+  public void visitMethod(MethodTree method) {
+    if (!method.is(Tree.Kind.CONSTRUCTOR) && returnsOptional(method)) {
+
+      // check that the method is not annotated with @Nullable
+      ModifiersTree modifiers = method.modifiers();
+      for (AnnotationTree annotation : modifiers.annotations()) {
+        Type type = annotation.annotationType().symbolType();
```
m-g-sonar: As discussed, this visit of annotations should be generalized to also catch parameters, field and variables.
The following cases should raise issues:

```
class A {
  @Nullable // Noncompliant
  Optional<String> myField;

  void foo(@Nullable Optional<String> s) { // Noncompliant - Optional should not be null
  }
}
```

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Override
+  public void visitBinaryExpression(BinaryExpressionTree binaryExpression) {
+    String message = "Remove this null-check of an \"Optional\".";
+    if (binaryExpression.is(Tree.Kind.EQUAL_TO, Tree.Kind.NOT_EQUAL_TO)) {
+      ExpressionTree left = binaryExpression.leftOperand();
+      ExpressionTree right = binaryExpression.rightOperand();
+      if ((isOptional(left) && isNull(right)) || (isNull(left) && isOptional(right))) {
+        context.reportIssue(this, binaryExpression, message);
```
m-g-sonar: you can put directly the message into the call to `reportIssue`, no need to use a dedicated variable.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+  }
+
+  private static boolean returnsOptional(MethodTree method) {
+    return method.returnType().symbolType().is(OPTIONAL);
+  }
+
+  private static boolean isOptional(ExpressionTree expression) {
+    return expression.symbolType().is(OPTIONAL) && !expression.is(Tree.Kind.NULL_LITERAL);
```
m-g-sonar: `expression.is(Tree.Kind.NULL_LITERAL)` should be replaced by `isNull(expression)` in order to reuse the method.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,132 @@
...
+      ExpressionTree expression = returnStatement.expression();
+      if (!checkNull(expression) && expression.is(Kind.CONDITIONAL_EXPRESSION)) {
+        ConditionalExpressionTree conditionalExpression = (ConditionalExpressionTree)expression; 
+        checkNull(conditionalExpression.trueExpression());
+        checkNull(conditionalExpression.falseExpression());
+      }
+    }
+    
+    private boolean checkNull(ExpressionTree expression) {
+      if (expression.is(Kind.NULL_LITERAL)) {
```
m-g-sonar: `expression.is(Tree.Kind.NULL_LITERAL)` should be replaced by `isNull(expression)` in order to reuse the method.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,132 @@
...
+        context.reportIssue(NullShouldNotBeUsedWithOptionalCheck.this, expression.firstToken(), "Methods with an \"Optional\" return type should never return null.");
```
m-g-sonar: Why reporting on the first token? You should probably report directly on the whole expression.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public void visitReturnStatement(ReturnStatementTree returnStatement) {
+      ExpressionTree expression = returnStatement.expression();
+      if (!checkNull(expression) && expression.is(Kind.CONDITIONAL_EXPRESSION)) {
+        ConditionalExpressionTree conditionalExpression = (ConditionalExpressionTree)expression; 
+        checkNull(conditionalExpression.trueExpression());
+        checkNull(conditionalExpression.falseExpression());
+      }
+    }
```
m-g-sonar: You should call `super.visitReturnStatement(returnStatement)`  before leaving the method, or you won't catch cases such as :
```
Optional<String> s = getValue();
return (s != null) ? optional.isEmpty() : s; // Noncompliant, the null check on s is wrong!
```

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,132 @@
...
+    super.visitBinaryExpression(binaryExpression);
+  }
+
+  private class ReturnNullVisitor extends BaseTreeVisitor {
+
+    @Override
+    public void visitReturnStatement(ReturnStatementTree returnStatement) {
+      ExpressionTree expression = returnStatement.expression();
+      if (!checkNull(expression) && expression.is(Kind.CONDITIONAL_EXPRESSION)) {
+        ConditionalExpressionTree conditionalExpression = (ConditionalExpressionTree)expression;
```
m-g-sonar: You should extract this part into the `visitConditionalExpression` method, and start the method by checking if the type of the expression is `Optional`.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: OK...

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: OK

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: OK

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: OK

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: Why? I do catch your case.
Anyway, I add your case to the unit test.

m-g-sonar: Oups, correct. You will catch this one. However, you won't catch chained conditional expressions:
```
return myList.isEmpty() ? Optional.of("hello") : (myList.size() > 10 ? Optional.of("world") : null);
```

ivandalbosco: In my own life I never wrote such a line and I will probably never do! I guess that if I do, SonarQube will urge me to split my line! But OK, a super.visitReturnStatement will be added.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: Thanks, visitConditionalExpression is a pretty good hint. 
The expression is Optional, as guarded by the enclosing class (L 55).

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,149 @@
...
+
+  private static boolean isOptional(ExpressionTree expression) {
+    return expression.symbolType().is(OPTIONAL) && !isNull(expression);
+  }
+
+  private static boolean isNull(ExpressionTree expression) {
+    return expression.is(Tree.Kind.NULL_LITERAL);
+  }
+
+  private void checkNullableOptionalAnnotation(ModifiersTree modifiers, String message) {
```
m-g-sonar: This method should be called `checkNullableAnnotation`, as you already know that it is applied on a `Nullable` method or variable.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,149 @@
...
+    @Override
+    public void visitReturnStatement(ReturnStatementTree returnStatement) {
+      checkNull(returnStatement.expression());
+
+      super.visitReturnStatement(returnStatement);
+    }
+
+    @Override
+    public void visitConditionalExpression(ConditionalExpressionTree conditionalExpression) {
+      checkNull(conditionalExpression.trueExpression());
```
m-g-sonar: This is a bit too strong. You should perform the verification of the true and false expressions only if the type of the expression is Optional. With the current state, the following code will trigger an issue:
```
  public Optional<String> bar(booleab b) {
    Object myObject = b ? null : new Object();
    return Optional.empty();
  }
```

With the following, it should work:
```
 if (conditionalExpression.symbolType().is(OPTIONAL)) {
    checkNull(conditionalExpression.trueExpression());
    checkNull(conditionalExpression.falseExpression());
  }
```

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -0,0 +1,149 @@
...
+      checkNull(conditionalExpression.falseExpression());
+
+      super.visitConditionalExpression(conditionalExpression);
+    }
+
+    private boolean checkNull(ExpressionTree expression) {
```
m-g-sonar: I'm not sure that you still need to return a `boolean` here.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: Done

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: Thanks, great.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
ivandalbosco: Correct!

PR #926, opened 2016-07-14
SONARJAVA-372 Consider only fields used in one method

vilchik-elena:


benzonico:
LGTM. You could have reopened the ticket. Please wait for QA to finish before merged.

vilchik-elena:
@benzonico Could you check again this PR? I've added one commit to not build LVA for each method in the class

java-checks/src/main/java/org/sonar/java/checks/PrivateFieldUsedLocallyCheck.java
```diff
@@ -59,41 +57,57 @@
   @Override
   public void visitNode(Tree tree) {
     TypeSymbol classSymbol = ((ClassTree) tree).symbol();
-    initFieldsMetaData(classSymbol);
+    Set<Symbol> fieldsReadOnAnotherInstance = FieldsReadOnAnotherInstanceVisitor.getFrom(classSymbol);
```
benzonico: The getFrom method should take the classTree not the symbol.

java-checks/src/main/java/org/sonar/java/checks/PrivateFieldUsedLocallyCheck.java
```diff
@@ -59,41 +57,57 @@
...
     return memberSymbol.isPrivate() && memberSymbol.isVariableSymbol();
   }
 
-  private static boolean isUsedOutsideMethods(Symbol privateFieldSymbol, TypeSymbol classSymbol) {
+  /**
+   * If private field used in several methods then returns null, otherwise returns the method where it's used
+   */
+  @CheckForNull
+  private static MethodTree usedInOneMethodOnly(Symbol privateFieldSymbol, TypeSymbol classSymbol) {
+    Set<MethodTree> methods = new HashSet<>();
```
benzonico: No need for a set here you can just use a variable and check its nullity.

PR #935, opened 2016-07-20
add support for Hamcrest assertion with three arguments (squid:S2699)

echipachenko:
Add support for Hamcrest assertion with three arguments for rule "Tests should include assertions - squid:S2699"


m-g-sonar:
Hey! Thanks a lot for your contribution!

PR #937, opened 2016-07-21
SONARJAVA-1782 Add handling of PasswordAuthentication with hardcoded credential

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/HardCodedCredentialsCheck.java
```diff
@@ -43,38 +47,79 @@
...
+
+  private void handleAssignement(AssignmentExpressionTree tree) {
+    ExpressionTree variable = tree.variable();
+    if (isStringLiteral(tree.expression()) && isPasswordVariable(variable)) {
+      reportIssue(variable);
+    }
+  }
+
+  private void handleConstructor(NewClassTree tree) {
+    if (PASSWORD_AUTHENTICATION_CONSTRUCTOR.matches(tree)) {
```
tvsonar: It's a matter of style, but I would reverse the conditions to not nest 3 `if` statements.

java-checks/src/test/files/checks/HardCodedCredentialsCheck.java
```diff
@@ -30,15 +32,25 @@ private void a() {
...
-    
+
     MyUnknownClass.myUnknownMethod("password", "xxxxx"); // Noncompliant
+
+    PasswordAuthentication pa;
+    pa = new PasswordAuthentication("userName", "1234".toCharArray());  // Noncompliant
+    pa = new PasswordAuthentication("userName", pwd); // Compliant
+    pa = new PasswordAuthentication("userName", getPwd(var)); // Compliant
+    pa = new PasswordAuthentication("userName", var.toCharArray()); // Compliant
+  }
```
tvsonar: `pa = new SomeOtherAuthentication("username", "1234".toCharArray());` could also be tested to show that there are three conditions that needs to be satisfied.

m-g-sonar: good point.

PR #940, opened 2016-07-21
SONARJAVA-1723 Remove handling of PrintWriter constructors using streams

m-g-sonar:


m-g-sonar:
Rejected. Constructors of PrintWriter should still be considered by the rule.

PR #941, opened 2016-07-22
SONARJAVA-1753 Handle empty final arrays

m-g-sonar:


java-checks/src/test/files/checks/PublicStaticMutableMembersCheck.java
```diff
@@ -34,6 +34,15 @@
...
 
   public static final int[] data = new int[5]; // Noncompliant
+  public static final int[] data_unknown_size = new int[getDim()]; // Noncompliant
+
+  public static int getDim() {
+    return 42;
+  }
+
+  public static final int[] EMPTY_DATA_1 = new int[0]; // Compliant
+  public static final String[][] EMPTY_DATA_2 = {new String[0], {}}; // Compliant
```
tvsonar: As discussed, coming from C# this should be noncompliant.

java-checks/src/main/java/org/sonar/java/checks/PublicStaticMutableMembersCheck.java
```diff
@@ -153,24 +158,45 @@ public void leaveNode(Tree tree) {
...
+    if (!expression.is(Tree.Kind.NEW_ARRAY)) {
+      return false;
+    }
+    NewArrayTree nat = (NewArrayTree) expression;
+    return nat.dimensions().stream().allMatch(PublicStaticMutableMembersCheck::isZeroDimension) && nat.initializers().isEmpty();
+  }
+
+  private static boolean isZeroDimension(ArrayDimensionTree dim) {
+    ExpressionTree expression = dim.expression();
+    return expression == null || (expression.is(Tree.Kind.INT_LITERAL) && "0".equals(((LiteralTree) expression).value()));
```
tvsonar: The method name doesn't match this behavior. If `expression == null` that doesn't mean that the dimension of the array is zero. It means that there has to be an initializer that defines the size.

java-checks/src/main/java/org/sonar/java/checks/PublicStaticMutableMembersCheck.java
```diff
@@ -153,24 +158,45 @@ public void leaveNode(Tree tree) {
...
     }
     return true;
   }
 
+  private static boolean isEmptyArray(ExpressionTree expression) {
+    if (!expression.is(Tree.Kind.NEW_ARRAY)) {
+      return false;
+    }
+    NewArrayTree nat = (NewArrayTree) expression;
+    return nat.dimensions().stream().allMatch(PublicStaticMutableMembersCheck::isZeroDimension) && nat.initializers().isEmpty();
```
tvsonar: I expect to have the following check here: `hasEmptyInitializer(nat) || allDimensionsAreZero(nat)`

PR #942, opened 2016-07-22
SONARJAVA-1757 Ignore serializable classes

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/UtilityClassWithPublicConstructorCheck.java
```diff
@@ -43,26 +44,27 @@
...
+    }
   }
 
   private static boolean anonymousClass(ClassTree classTree) {
     return classTree.simpleName() == null;
   }
 
-  private static boolean extendsAnotherClass(ClassTree classTree) {
-    return classTree.superClass() != null;
+  private static boolean extendsAnotherClassOrImplementsSerialiozable(ClassTree classTree) {
```
tvsonar: Typo in method name

m-g-sonar: thx for noticing this!

PR #139, opened 2015-03-31
SONARJAVA-966 SONARJAVA-979 Support relations between symbolic values deduced from relational operators

merciesa:

PR #943, opened 2016-07-26
SONARJAVA-372 Handle constants and annotated fields

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/PrivateFieldUsedLocallyCheck.java
```diff
@@ -61,11 +63,17 @@ public void visitNode(Tree tree) {
...
     classSymbol.memberSymbols().stream()
       .filter(PrivateFieldUsedLocallyCheck::isPrivateField)
+      .filter(s -> !(s.isFinal() && s.isStatic()))
+      .filter(s -> !useAuthorizedAnnotation(s))
       .filter(s -> !s.usages().isEmpty())
       .filter(s -> !fieldsReadOnAnotherInstance.contains(s))
       .forEach(s -> checkPrivateField(s, classSymbol));
   }
 
+  private static boolean useAuthorizedAnnotation(Symbol s) {
```
tvsonar: `hasAuthorizedAnnotation` or `useAllowedAnnotation` sounds better to me.

java-checks/pom.xml
```diff
@@ -184,11 +184,6 @@
                   <type>jar</type>
                 </artifactItem>
                 <artifactItem>
-                  <groupId>javax</groupId>
```
tvsonar: Is this change required?

m-g-sonar: It's a duplicate of line [#139](https://github.com/SonarSource/sonar-java/pull/943/files/9791754d17d30557a03aa303e233cda221181131#diff-598a04ed7e994c371420982332fc73caR139), that I discovered while I wanted to add the jar containing `@Inject` annotation.

PR #946, opened 2016-07-28
Make MeasurerTest compatible with Windows.

Johnnei:

The Windows file.seperator is '\' which causes a mismatch between the project key constructed in the assertions compared to one in DefaultInputFile.

This will allow 'mvn clean install' to cleanly run out of the box on windows.


benzonico:
Thanks for the contribution. We'll try to add a small build on windows to ensure we don't encounter issues like this.

PR #949, opened 2016-08-07
SONARJAVA-1810 Array or Collection of Serializable objects are Serializable

janosgyerik:

- [ ] ITs should pass : To run ITs locally, checkout the [README](https://github.com/SonarSource/sonar-java/blob/master/README.md) of the project. -> still not working in master

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableObjectInSessionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableObjectInSessionCheck.java
@@ -79,7 +79,9 @@ public class SerializableObjectInSessionCheck extends AbstractMethodDetection {
   }
 
   private boolean isSerializableParametrized(ParametrizedTypeJavaType type) {
-    return type.isSubtypeOf("java.io.Serializable")
+    // note: this is assuming that custom implementors of Collection
+    // have the good sense to make it serializable just like all implementations in the JDK
+    return (type.isSubtypeOf("java.io.Serializable") || type.isSubtypeOf("java.util.Collection"))
       && type.typeParameters().stream().allMatch(t -> {
       JavaType javaType = type.substitution(t);
       assert javaType != null;
--- a/java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableObjectInSessionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableObjectInSessionCheck.java
@@ -24,6 +24,7 @@ import org.sonar.check.Rule;
 import org.sonar.java.checks.methods.AbstractMethodDetection;
 import org.sonar.java.matcher.MethodMatcher;
 import org.sonar.java.matcher.TypeCriteria;
+import org.sonar.java.resolve.ArrayJavaType;
 import org.sonar.plugins.java.api.semantic.Type;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodInvocationTree;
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableObjectInSessionCheck.java
```diff
@@ -44,8 +47,45 @@
...
+    return type instanceof ParametrizedTypeJavaType;
+  }
+
+  private boolean isSerializableParametrized(ParametrizedTypeJavaType type) {
+    // note: this is assuming that custom implementors of Collection
+    // have the good sense to make it serializable just like all implementations in the JDK
+    return (type.isSubtypeOf("java.io.Serializable") || type.isSubtypeOf("java.util.Collection"))
+      && type.typeParameters().stream().allMatch(t -> {
+      JavaType javaType = type.substitution(t);
+      assert javaType != null;
```
m-g-sonar: Please remove the `assert` statement and replace the whole lambda expression by:
```
t -> isSerializable(type.substitution(t));
```

The situation where we are not able to retrieve the substitution for one of its own type parameter should never occur. However, if it occurs, we don't want the rule to fail the analysis with an assertion error, we want it to fail badly. It means that we are facing a problematic situation and we want a feedback from an afraid user on it. It is the responsibility of the semantic analysis to fail if it encounter an impossible situation, not the rules.

From a coverage point of view, you also won't be able to cover such cases, as you can't build it.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableObjectInSessionCheck.java
```diff
@@ -44,8 +47,45 @@
...
     Type type = argument.symbolType();
-    if (!type.isPrimitive() && !type.isSubtypeOf("java.io.Serializable")) {
-      reportIssue(argument, "Make \"" + type + "\" serializable or don't store it in the session.");
+    if (!isSerializable(type)) {
+      String andParameters = isParametrized(type) ? " and its parameters" : "";
+      reportIssue(argument, "Make \"" + type + "\"" + andParameters + " serializable or don't store it in the session.");
     }
   }
+
+  private boolean isSerializable(Type type) {
```
m-g-sonar: This method can be static, as all the methods it calls.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableObjectInSessionCheck.java
janosgyerik: In production the `assert` statement should have no effect, so users should see the same kind of crash with or without it. It would only be triggered during tests, other than that it's just documentation.

In terms of coverage, for what it's worth, IntelliJ shows the method fully covered when I run `SerializableObjectInSessionCheckTest`.

I removed it anyway because I did not like the tediousness anyway, but I let IntelliJ slap a `//noinspection ConstantConditions` on the statement to make the warning go away, with an explanatory comment.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableObjectInSessionCheck.java
```diff
@@ -44,8 +46,44 @@
...
+  private static boolean isParametrized(Type type) {
+    return type instanceof ParametrizedTypeJavaType;
+  }
+
+  private static boolean isSerializableParametrized(ParametrizedTypeJavaType type) {
+    // note: this is assuming that custom implementors of Collection
+    // have the good sense to make it serializable just like all implementations in the JDK
+    //
+    // note: type.substitution(t) should never be null
+    //noinspection ConstantConditions
```
m-g-sonar: No need for the IDE-specific comment. we prefer to deal with the warnings manually.

janosgyerik: Alright, removed.

m-g-sonar:
Thanks for your contribution @janosgyerik, your PR has been squashed and merged on master!

PR #951, opened 2016-08-08
SONARJAVA-1772 Remove UnusedProtectedMethod rule and its dependencies

benzonico:


java-frontend/src/main/java/org/sonar/java/DefaultJavaResourceLocator.java
```diff
@@ -77,8 +75,7 @@ public String findSourceFileKeyByClassName(String className) {
     return sourceFileByClass.get(name);
   }
 
-  @Override
-  public Collection<String> classKeys() {
+  private Collection<String> classKeys() {
```
m-g-sonar: Is the ordering really required? Because if it's not, we should simply drop the method and use `resourcesByClass.keySet()` at line 85 instead.

PR #952, opened 2016-08-08
SONARJAVA-1809 Fix NPE on EnumConstant with body

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/DITCheck.java
```diff
@@ -58,23 +58,30 @@ public void scanFile(JavaFileScannerContext context) {
...
+        dit++;
+        superClass = superClass.symbol().superClass();
+      }
+      if (dit > max) {
+        Tree reportTree = tree.simpleName();
+        if (tree.parent().is(Tree.Kind.NEW_CLASS)) {
+          reportTree = ((NewClassTree) tree.parent()).newKeyword();
+        }
+        context.reportIssue(this, reportTree, "This class has " + dit + " parents which is greater than " + max + " authorized.",
+          new ArrayList<JavaFileScannerContext.Location>(), dit - max);
```
benzonico: Type param can be removed.

java-checks/src/test/files/checks/Dit.java
```diff
@@ -13,3 +13,14 @@ void foo() {
 class Two {
 
 }
+
+enum Any {
+  enumConst { // Compliant - enum constant are not considered (level=3: enumConst -> Any -> Enum<Any> -> Object)
```
benzonico: lacking nested issue to ensure the visit of enum constant body.

PR #956, opened 2016-08-09
SONARJAVA-1808 discard previous modifications on BinarySymbolicValue and use wrapped symbolic values

m-g-sonar:


java-frontend/src/test/files/se/DivisionByZeroCheck.java
```diff
@@ -371,4 +371,27 @@ long hashSymbol(byte[] buf) {
     }
     return h & 0xFFFFFFFFL;
   }
+
+  private void decodeBigInteger(int value) throws Exception {
```
benzonico: Test case can probably be simplified.

m-g-sonar: simplified

PR #964, opened 2016-08-15
SONARJAVA-1819 Complete type of method references based on their retu…

benzonico:


java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -524,34 +525,50 @@ public void visitLambdaExpression(LambdaExpressionTree tree) {
...
       // TODO compute return type for lambda with block and return statements
       return;
     }
     JavaType refinedReturnType = (JavaType) ((AbstractTypedTree) lambdaExpressionTree.body()).symbolType();
-    if (refinedReturnType != capturedReturnType && lambdaType.isParameterized()) {
+    refineType(lambdaExpressionTree, lambdaType, capturedReturnType, refinedReturnType);
+  }
+
+  private void refineType(AbstractTypedTree expression, JavaType expressionType, JavaType capturedReturnType, JavaType refinedReturnType) {
+    if (refinedReturnType != capturedReturnType && expressionType.isParameterized()) {
```
tvsonar: Statement nesting can be avoided by inverting this `if` and returning early from the method.

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -524,34 +525,50 @@ public void visitLambdaExpression(LambdaExpressionTree tree) {
...
       // found a lambda return type different from the one infered : update infered type
-      TypeSubstitution typeSubstitution = ((ParametrizedTypeJavaType) lambdaType).typeSubstitution;
+      TypeSubstitution typeSubstitution = ((ParametrizedTypeJavaType) expressionType).typeSubstitution;
       for (Map.Entry<TypeVariableJavaType, JavaType> entry : typeSubstitution.substitutionEntries()) {
         if (entry.getValue() == capturedReturnType) {
```
tvsonar: Nesting can be removed by moving this condition into a stream filtering.

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -524,34 +525,50 @@ public void visitLambdaExpression(LambdaExpressionTree tree) {
...
           TypeSubstitution refinedSubstitution = new TypeSubstitution(typeSubstitution).add(entry.getKey(), refinedReturnType);
-          JavaType refinedLambdaType = parametrizedTypeCache.getParametrizedTypeType(lambdaType.symbol, refinedSubstitution);
-          lambdaExpressionTree.setType(refinedLambdaType);
+          JavaType refinedType = parametrizedTypeCache.getParametrizedTypeType(expressionType.symbol, refinedSubstitution);
+          expression.setType(refinedType);
           break;
```
tvsonar: `break` could be removed by converting this foreach loop into a `findAny` or `findFirst`

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -524,34 +525,50 @@ public void visitLambdaExpression(LambdaExpressionTree tree) {
...
 
+  @CheckForNull
   private static JavaSymbol.MethodJavaSymbol getSamMethod(JavaType lambdaType) {
     for (Symbol member : lambdaType.symbol().memberSymbols()) {
       if (member.isMethodSymbol() && member.isAbstract()) {
         return (JavaSymbol.MethodJavaSymbol) member;
       }
     }
-    throw new IllegalStateException("Method not found in SAM Interface : "+lambdaType.fullyQualifiedName());
+    for (ClassJavaType type : lambdaType.symbol.superTypes()) {
```
tvsonar: Nesting could be simplified to a single call chain.

tvsonar: This change should be mentioned in the ticket, or add a dedicated ticket for it) It's not related to returning the return type, but to walk up the inheritance chain.

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -524,34 +525,50 @@ public void visitLambdaExpression(LambdaExpressionTree tree) {
...
+      for (Symbol member : type.symbol().memberSymbols()) {
+        if (member.isMethodSymbol() && member.isAbstract()) {
+          return (JavaSymbol.MethodJavaSymbol) member;
+        }
+      }
+    }
+    return null;
```
tvsonar: Is this expected? Shouldn't we throw instead?

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -914,14 +931,31 @@ public void visitMethodReference(MethodReferenceTree methodReferenceTree) {
       if("new".equals(methodReferenceTree.method().name())) {
         resolution = resolve.findMethod(methodEnv, getType(methodReferenceTree.expression()), "<init>", samMethodArgs);
         associateReference(methodReferenceTree.method(), resolution.symbol());
+        return;
       } else {
```
tvsonar: The `else` is redundant here.

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -914,14 +931,31 @@ public void visitMethodReference(MethodReferenceTree methodReferenceTree) {
...
         resolveMethodSymbol(methodReferenceTree.method(), methodEnv, samMethodArgs, ImmutableList.<JavaType>of());
       }
+      JavaType methodRefType = (JavaType) methodRefTree.symbolType();
+      if (methodRefTree.symbolType().isUnknown() || methodRefType.isTagged(JavaType.DEFERRED)) {
```
tvsonar: why not use `methodRefType` instead of calling `methodRefTree.symbolType()` again?

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -914,14 +931,31 @@ public void visitMethodReference(MethodReferenceTree methodReferenceTree) {
...
+        return;
+      }
+      JavaSymbol.MethodJavaSymbol samMethod = getSamMethod(methodRefType);
+      if(samMethod == null) {
+        return;
+      }
+      JavaType samReturnType = (JavaType) samMethod.returnType().type();
+      JavaType capturedReturnType = resolve.resolveTypeSubstitution(samReturnType, methodRefType);
```
tvsonar: `capturedReturnType` is only used inside the `if`, could this declaration be moved there?

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
benzonico: done in pull request of SONARJAVA-1820

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
benzonico: It is completely related to the computing the return type, this method is only called from return type computation . it was simply not covered by lambdas.

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
benzonico: It can be expected for method reference : foo(a::toString) the to String is not a sam method.

PR #978, opened 2016-08-19
SONARJAVA-1794: Add parameter to toggle UndocumentedAPI behaviour

Johnnei:

The new parameters allows to switch between inclusion and exclusion mode.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
@@ -55,6 +55,7 @@ public class UndocumentedApiCheck extends BaseTreeVisitor implements JavaFileSca
   private static final Kind[] METHOD_KINDS = PublicApiChecker.methodKinds();
 
   private static final String DEFAULT_FOR_CLASSES = "**.api.**";
+  private static final String DEFAULT_EXCLUSION = "**.internal.**";
 
   @RuleProperty(
     key = "forClasses",
@@ -62,7 +63,15 @@ public class UndocumentedApiCheck extends BaseTreeVisitor implements JavaFileSca
     defaultValue = DEFAULT_FOR_CLASSES)
   public String forClasses = DEFAULT_FOR_CLASSES;
 
-  private WildcardPattern[] patterns;
+  @RuleProperty(
+    key = "exclusion",
+    description = "Pattern of classes which are excluded from adhering to this constraint.",
+    defaultValue = DEFAULT_EXCLUSION)
+  public String exclusion = DEFAULT_EXCLUSION;
+
+  private WildcardPattern[] inclusionPatterns;
+  private WildcardPattern[] exclusionPatterns;
+
   private final Deque<ClassTree> classTrees = Lists.newLinkedList();
   private final Deque<Tree> currentParents = Lists.newLinkedList();
 
@@ -160,7 +169,7 @@ public class UndocumentedApiCheck extends BaseTreeVisitor implements JavaFileSca
   }
 
   private boolean isExcluded(Tree tree) {
-    return !isPublicApi(tree) || isAccessor(tree) || !isMatchingPattern();
+    return !isPublicApi(tree) || isAccessor(tree) || !isMatchingInclusionPattern() || isMatchingExclusionPattern();
   }
 
   private boolean isAccessor(Tree tree) {
@@ -185,8 +194,12 @@ public class UndocumentedApiCheck extends BaseTreeVisitor implements JavaFileSca
     return publicApiChecker.isPublicApi(currentParent, tree);
   }
 
-  private boolean isMatchingPattern() {
-    return WildcardPattern.match(getPatterns(), className());
+  private boolean isMatchingInclusionPattern() {
+    return WildcardPattern.match(getInclusionPatterns(), className());
+  }
+
+  private boolean isMatchingExclusionPattern() {
+    return WildcardPattern.match(getExclusionPatterns(), className());
   }
 
   private String className() {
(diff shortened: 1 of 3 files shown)


benzonico:
@Johnnei could you please rebase your PR on master : I deactivated PR analysis so, at least CI can run on travis. Thanks.

m-g-sonar:
hey @Johnnei,

Your approach won't work if you want to raise issues on all the undocumented API in package matching `*api*` but not on internal APIs.
More explicitly, I can't see how to have the following behavior with your approach:
- `org.foo.bar` : no issues, not API package
- `org.foo.api.bar` : issues on all the undocumented API
- `org.foo.internal.bar` : no issues, not API package
- `org.foo.internal.api.bar` : skipped, as being internal API

Johnnei:
Hey @m-g-sonar,

I've chosen this approach as I felt like combining both an `*.api.*` and  a `*.internal.*` in the same project would be very confusing for the consumer of the API.
In your example when using both `*.api.*` and `*.internal*` the `org.foo.bar` package for me is wild ground: Would it be safe to use it? It's not part of the API but it's also not part of the internals, what is it?

As another consideration: I feel like having both an inclusion and exclusion field would be more confusing to configure. For example: What should I configure if I only want the exclusions on `*.internal.*` or only the inclusions on `*.api.*`? (Or maybe I'm just a dumb dumb and not thinking about the simple way 😄)

m-g-sonar:
Okay, my example was a little dumb, so I just looked at the reason we created the ticket. The SonarQube platform define multiple APIs, for instance [sonar-plugin-api](https://github.com/SonarSource/sonarqube/tree/master/sonar-plugin-api/src/main/java/org/sonar/api). This API is developed in order to allow creation of external sonar plugin. In this API, however, there is parts which should not be used externally, or are subject to change.

Back to my previous package examples, I would then expect to have

- `org.foo.bar` : no issues, not API package
- `org.foo.api.bar` : issues on all the undocumented API
- `org.foo.api.internal.bar` : no issues, excluded from API

Regarding the rule behavior, that's not that complicated:
```
  forClasses    exclusion     behavior
  *api*         nothing       all classes in any 'api' package
    *           *internal*    all classes in any package, except if they are in an internal package
  *api*         *internal*    all classes which match *api*, as long as they are not matching *internal*
```

Johnnei:
@m-g-sonar I've taken a quick look through the repository and it seems like that even though SonarQube uses the api prefix it doesn't actually use the `org.foo.bar` example and thus only defining `**.internal.**` as exclusion would raise exactly the same issues as with having both defined `**.api.**` and `**.internal.**` (or am I missing something?).

However with your given configuration I can agree that it wouldn't be all that complex to configure (I was thinking too complex on their interactions).
So as a final confirmation:
- There should be both an inclusion field and exclusion field.
- I'd say: Always enforce the user to explicitly define the 'forClasses' and optionally the exclusions (Will have to look into on how to enforce that on the Web-side if possible)

Is that correct for a SonarQube point of view?

m-g-sonar:
@Johnnei Yep, I confirm.

Regarding enforcing usages of  the rule parameters, we can't define mandatory/optional parameters in the UI for a normal rule (would have been different for template rule, which has to be instantiated with manual configuration). I think that the description of the parameter is enough. The rule should however now be able to handle case where forClasses/exception parameter has not been defined (empty value).

Johnnei:
@m-g-sonar Alright! Final (I think) question: What should be the behaviour when the `forClasses` is not supplied (aka: empty)? Should it fallback to the default for example?

m-g-sonar:
@Johnnei Let's consider empty `forClasses` as being a wildcard (matches everything).

Johnnei:
@m-g-sonar, Here's the reworked version 😄

java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
```diff
@@ -198,11 +211,26 @@ private String className() {
...
+
+      inclusionPatterns = PatternUtils.createPatterns(forClasses);
+    }
+    return inclusionPatterns;
+  }
+
+  private WildcardPattern[] getExclusionPatterns() {
+    if (exclusionPatterns == null) {
+      if (StringUtils.isEmpty(exclusion)) {
+        return new WildcardPattern[0];
```
m-g-sonar: You should also lazily initialize the exclusionPatterns with an empty array when exclusion is not provided, as it will slightly improve performance when calling the rule on multiple files during analysis.

Johnnei: Maybe I'm not understanding it right, but in every case the exclusionsPatterns will end up with a non-null value if it was null. If it's either an empty string (maybe isBlank would be better?) or null it will become an empty array, otherwise the pattern will be parsed and the patterns assigned. So the first time this method is called the exlcusionPatterns will be null and after the call it can't be null anymore.

m-g-sonar: Well, as you are not setting the `exclusionPatterns` field, but directly returning in case of empty `exclusion` param, it will remains `null`.

Johnnei: Whoops! You're correct, my brain somehow convinced me it was being assigned. Will correct this later today.

Johnnei:
@m-g-sonar Fixed the Quality flaw. I've also removed one of the test cases as it didn't add any more coverage (both line-wise and functional-wise).

m-g-sonar:
Merged! Thanks again for your contribution @Johnnei ! :)

Johnnei:
@m-g-sonar Glad to be of help! Also great to see that we can have functional discussions :smile:

PR #983, opened 2016-08-30
SONARJAVA-1751 fix FN on method calls without parameter on lambda single argument

m-g-sonar:


benzonico:
isSingleMethodInvocationUsingLambdaParamAsArg should be renamed.

PR #988, opened 2016-09-01
SONARJAVA-1828/SONARJAVA-1822 Handle MessageFormat.format() with arrays and LocalDate

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/PrintfCheck.java
```diff
@@ -341,8 +341,11 @@ private void checkTimeConversion(MethodInvocationTree mit, String param, Type ar
   }
 
   private void checkTimeTypeArgument(MethodInvocationTree mit, Type argType) {
-    if (!(argType.isNumerical() || argType.is("java.lang.Long") || argType.isSubtypeOf("java.util.Date") || argType.isSubtypeOf("java.util.Calendar"))) {
-      reportIssue(mit, "Time argument is expected (long, Long, Date or Calendar).");
+    if (!(argType.isNumerical()
+      || isOneOf(argType, "java.lang.Long", "java.time.LocalDate")
```
tvsonar: I'd rename this method to `isTypeOfAny` and introduce `isSubtypeOfAny` for the following two lines. Also I tend to no use `!(a || b || c)` when it's too long, use `!a && !b && !c` instead.

java-checks/src/main/java/org/sonar/java/checks/PrintfCheck.java
```diff
@@ -121,14 +122,30 @@ private void handlePrintfFormat(MethodInvocationTree mit, String formatString, L
   private void handleMessageFormat(MethodInvocationTree mit, String formatString, List<ExpressionTree> args) {
     String newFormatString = cleanupDoubleQuote(formatString);
     Set<Integer> indexes = getMessageFormatIndexes(newFormatString);
+    List<ExpressionTree> newArgs = args;
+    if (newArgs.size() == 1) {
+      ExpressionTree firstArg = args.get(0);
```
tvsonar: This is correct, but it looks strange that you check `newArgs.size()` and then access `args.get(0)`, use `newArgs.get(0)` instead.

PR #999, opened 2016-09-09
fix usage of MethodMatcher#withNoParameterConstraint()

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/WaitInWhileLoopCheck.java
```diff
@@ -84,7 +85,7 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
       MethodMatcher.create().name("wait"),
       MethodMatcher.create().name("wait").addParameter("long"),
       MethodMatcher.create().name("wait").addParameter("long").addParameter("int"),
-      MethodMatcher.create().typeDefinition("java.util.concurrent.locks.Condition").name("await").withNoParameterConstraint()
+      MethodMatcher.create().typeDefinition("java.util.concurrent.locks.Condition").name(NameCriteria.startsWith("await")).withNoParameterConstraint()
```
m-g-sonar: method `withNoParameterConstraint` is indeed somehow confusing. In my case, I'm always looking at the implementation to have an idea which I should use... I think that we should go a bit further and modify the MethodMatcher in order to have the following methods regarding handling of parameter:

- `withoutParameter` : the method has no parameter
- `withAnyParameters`: the method parameters are not considered and can have any parameter

We could also modify the method matcher in order to fail if there is no parameter added and `withoutParameter` or `withAnyParameters` has not been called.

PR #1001, opened 2016-09-12
SONARJAVA-1838 NPE on S1149 when type of lambda parameter is synchronized

alban-auzeill:


java-checks/src/test/files/checks/SynchronizedClassUsageCheck.java
```diff
@@ -1,7 +1,20 @@
-import java.util.Hashtable;     // Compliant
-import java.util.Vector;        // Compliant
+import java.util.ArrayList;     // Compliant
```
m-g-sonar: This comment is irrelevant. You can remove it, or add `// Compliant` comments on the import which are related to the targeted types.

PR #1003, opened 2016-09-16
SONARJAVA-1843 Rule S2959: Unnecessary semicolons should be omitted

alban-auzeill:

### DIFF
--- a/java-checks/src/test/files/checks/UnnecessarySemicolonCheck.java
+++ b/java-checks/src/test/files/checks/UnnecessarySemicolonCheck.java
@@ -1,0 +1,28 @@
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public class UnnecessarySemicolonCheck {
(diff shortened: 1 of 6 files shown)


java-checks/src/test/files/checks/UnnecessarySemicolonCheck.java
```diff
@@ -0,0 +1,29 @@
...
+  private static InputStream input(String path) throws IOException {
+    return Files.newInputStream(Paths.get(path));
+  }
+
+  public static void main(String[] args) throws IOException {
+
+    try(InputStream i1 = input("i1");) { // Noncompliant {{Remove this extraneous semicolon.}}
+    }
+
+    try(InputStream i1 = input("i1");
```
m-g-sonar: For this test case, put the two input streams on the same line and use precise issue location testing, it will help guarantee that we are only raising issue on the last semi-colon.

You can use precise issue location like this:
```// Noncompliant [[sc=14;ec=15]]```

PR #1006, opened 2016-09-21
Improve contrib guidelines

janos-ss:


benzonico:
Thanks for the contribution @janos-ss

PR #1015, opened 2016-09-28
SONARJAVA-1858 Propagate type substitution of call site to formals paramaters when resolving methods

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/resolve/GenericsTest.java
```diff
@@ -853,46 +854,15 @@ public void test_method_resolution_with_substitution_in_type_hierarchy() {
...
-        + "class A {}"
-        + "class B extends A {}");
+  public void parametrized_method_resolution_with_bounded_type_variable() {
+    List<Type> elementTypes = declaredTypesFromFile("src/test/files/resolve/GenericMethodsBoundedTypeVariables.java");
 
     JavaType type = (JavaType) elementTypes.get(0);
     JavaSymbol.MethodJavaSymbol methodSymbol = getMethodSymbol(type, "foo");
-    // FIXME should be 12
-    assertThat(methodSymbol.usages()).hasSize(0);
+    // FIXME should be 12 - explicit type arguments not handled
```
benzonico: there is already a ticket about that missing piece. could be great to reference it.

PR #1020, opened 2016-10-03
SONARJAVA-1813 Update rule description and use type from json files

benzonico:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S139_java.html
```diff
@@ -11,9 +11,4 @@
 // This very long comment is better placed before the line of code
 int a2 = b + c;
 </pre>
-<h2>Exceptions</h2>
```
m-g-sonar: As we didn't removed the parameter, this should not be changed.

PR #1026, opened 2016-10-05
SOANRJAVA-1869 handle deferred type variables during inference

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/resolve/TypeInferenceSolver.java
```diff
@@ -102,6 +103,18 @@ private TypeSubstitution inferTypeSubstitution(MethodJavaSymbol method, TypeSubs
     return result;
   }
 
+  private static JavaType getUninferedType(DeferredType deferredType) {
```
benzonico: This should probably be in deferredType class.

PR #1027, opened 2016-10-10
SONARJAVA-1872 Fix how deferred and primitive types are handled in conditionnal expressions

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
```diff
@@ -1334,4 +1335,11 @@ public void double_lambda_type_propagation() {
     JavaSymbol my = result.symbol("my");
     assertThat(my.usages()).hasSize(2);
   }
+
+  @Test
+  public void conditional_expression_in_lambda() {
+    Result result = Result.createFor("ConditionalExpressionInLambda");
+    JavaSymbol foo = result.symbol("foo");
+    assertThat(foo.usages()).hasSize(1);
```
benzonico: Please make a direct test : this one is indirect. You should test the expected type of the conditional expression.

java-frontend/src/main/java/org/sonar/java/resolve/LeastUpperBound.java
```diff
@@ -104,7 +105,13 @@ private Type cachedLeastUpperBound(Set<Type> types) {
       }
     }
     return erasedBest;
+  }
 
+  private static Set<Type> primitiveWrappers(Set<Type> types) {
+    if (!types.stream().allMatch(Type::isPrimitive)) {
```
benzonico: I would just not negate the condition.

java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
```diff
@@ -1341,5 +1341,8 @@ public void conditional_expression_in_lambda() {
     Result result = Result.createFor("ConditionalExpressionInLambda");
     JavaSymbol foo = result.symbol("foo");
     assertThat(foo.usages()).hasSize(1);
+
+    JavaSymbol bar = result.symbol("bar");
```
benzonico: Please test the expected type of the conditional expression.

PR #1036, opened 2016-10-24
SONARJAVA-1903 Fix XmlCheckVerifier in order to disable DTD validation

m-g-sonar:


java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/XmlCheckVerifier.java
```diff
@@ -96,6 +106,7 @@ protected static void retrieveExpectedIssuesFromFile(File xmlFile, CheckVerifier
           checkVerifier.collectExpectedIssues(text, line);
         }
       }
+      reader.close();
```
benzonico: no need to close this reader as the underlying input stream is closed

PR #1039, opened 2016-10-24
SONARJAVA-1899 Handle null checks

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -65,10 +69,56 @@ public void scanFile(JavaFileScannerContext context) {
...
+      BinaryExpressionTree bet = (BinaryExpressionTree) expr;
+      if (nullAgainstParam(bet.leftOperand(), bet.rightOperand(), tree) || nullAgainstParam(bet.rightOperand(), bet.leftOperand(), tree)) {
+        return Optional.of(expr.is(Tree.Kind.EQUAL_TO) ? "isNull" : "nonNull");
+      }
+    }
+    return Optional.empty();
+  }
+
+  private static boolean nullAgainstParam(ExpressionTree o1, ExpressionTree o2, LambdaExpressionTree tree) {
+    if (o1.is(Tree.Kind.NULL_LITERAL) && (o2.is(Tree.Kind.IDENTIFIER))) {
```
benzonico: what about : 
```
(f) == null 
```

PR #1045, opened 2016-10-26
Replace maven-antrun-plugin with jaxb2-maven-plugin for using xjc

vincenzolaudizio:
The module called java-maven-model uses the maven-antrun-plugin for generating the Maven model.
The plugin is configured to execute the xjc tool that needs to be available in the path.

It would it better to remove the "system dependency" in favour of the specific maven plugin jaxb2-maven-plugin.

PR #1046, opened 2016-10-26
Replace maven-antrun-plugin with jaxb2-maven-plugin for using xjc

m-g-sonar:


m-g-sonar:
Clone of https://github.com/SonarSource/sonar-java/pull/1045 in order to trigger internal QA to run ITs.

PR #1052, opened 2016-10-28
SONARJAVA-1914

benzonico:


SonarTech:
* MINOR 1 minor

java-frontend/src/main/java/org/sonar/java/se/checks/CustomUnclosedResourcesCheck.java
```diff
@@ -136,8 +135,10 @@ protected void closeResource(@Nullable SymbolicValue target) {
       }
     }
 
-    protected void openResource(Tree syntaxNode) {
-      SymbolicValue sv = programState.peekValue();
+    protected void openResource(@Nullable SymbolicValue sv, Tree syntaxNode) {
+      if(sv == null) {
```
m-g-sonar: I would prefer to remove the `@Nullable` annotation and the null check, and provide directly the peeked value when calling the method in `visitNewClass` and `visitMethodInvocation`.

PR #1057, opened 2016-11-01
SONARJAVA-1883 Make S131 take into account complete enums when checking for default case

m-g-sonar:


its/ruling/src/test/resources/jdk6/squid-SwitchLastCaseIsDefaultCheck.json
```diff
@@ -24,9 +24,6 @@
 4187,
 4228,
 ],
-'jdk6:java/awt/Dialog.java':[
```
m-g-sonar: Covers all the 4 cases of the enum.

its/ruling/src/test/resources/jdk6/squid-SwitchLastCaseIsDefaultCheck.json
```diff
@@ -162,27 +159,6 @@
 'jdk6:java/io/StreamTokenizer.java':[
 644,
 ],
-'jdk6:java/lang/CharacterData00.java':[
```
m-g-sonar: The check now requires semantic, so all the issues within `java.lang` package are lost.

java-checks/src/main/java/org/sonar/java/checks/SwitchLastCaseIsDefaultCheck.java
```diff
@@ -42,35 +46,57 @@
...
   private static CaseLabelTree getLastLabel(SwitchStatementTree switchStatementTree) {
-    if (!switchStatementTree.cases().isEmpty()) {
-      List<CaseLabelTree> labels = switchStatementTree.cases().get(switchStatementTree.cases().size() - 1).labels();
-      if (!labels.isEmpty()) {
-        return labels.get(labels.size() - 1);
-      }
-    }
-    return null;
+    List<CaseGroupTree> caseGroups = switchStatementTree.cases();
+    List<CaseLabelTree> lastCaseLabels = caseGroups.get(caseGroups.size() - 1).labels();
```
benzonico:
```
  switch ("") {

    } 
``` 
is a valid compiling code that can blew up this line :boom:

java-checks/src/main/java/org/sonar/java/checks/SwitchLastCaseIsDefaultCheck.java
```diff
@@ -42,35 +46,57 @@
...
+  }
+
+  private static boolean missingCasesOfEnum(SwitchStatementTree switchStatementTree) {
+    return numberConstants(switchStatementTree) > allLabels(switchStatementTree).size();
+  }
+
+  private static List<CaseLabelTree> allLabels(SwitchStatementTree switchStatementTree) {
+    return switchStatementTree.cases().stream()
+      .flatMap(caseGroup -> caseGroup.labels().stream())
+      .collect(Collectors.toList());
```
benzonico: Given that you use it solely as a stream or to gets its size you could return a Stream here... no need to collect that as a list...

java-checks/src/main/java/org/sonar/java/checks/SwitchLastCaseIsDefaultCheck.java
```diff
@@ -42,35 +46,57 @@
...
+    }
     SwitchStatementTree switchStatementTree = (SwitchStatementTree) tree;
-    CaseLabelTree defaultLabel = getDefaultLabel(switchStatementTree);
-    CaseLabelTree lastLabel = getLastLabel(switchStatementTree);
-    if (defaultLabel == null) {
-      reportIssue(switchStatementTree.switchKeyword(), "Add a default case to this switch.");
-    } else if (!defaultLabel.equals(lastLabel)) {
-      reportIssue(defaultLabel, "Move this default to the end of the switch.");
+    Optional<CaseLabelTree> defaultLabel = getDefaultLabel(switchStatementTree);
+    if (!defaultLabel.isPresent()) {
```
benzonico: This could be reworked to avoid having ! in  nearly every condition to make the different cases cleaner to read.

java-checks/src/main/java/org/sonar/java/checks/SwitchLastCaseIsDefaultCheck.java
m-g-sonar: This method will only be called with a non-empty list of case groups.

PR #1064, opened 2016-11-04
SONARJAVA-1909 Empty statement ignored when used to define empty enum

m-g-sonar:


java-checks/src/test/files/checks/EmptyStatementUsageCheck.java
```diff
@@ -81,3 +81,14 @@ void foo (MyEnum scope) {
 }
 
 ; // Noncompliant {{Remove this empty statement.}}
+
+enum EmptyEnum {
+  // This is my empty enum full of emptyness
```
benzonico: This is a rather depressed enum.

PR #1072, opened 2016-11-07
SONARJAVA-1846 Indentation Check : Fix false positive

benzonico:


java-checks/src/main/java/org/sonar/java/checks/IndentationCheck.java
```diff
@@ -74,119 +60,98 @@
...
-    return BLOCK_TYPES;
-  }
+  private JavaFileScannerContext context;
 
   @Override
   public void scanFile(JavaFileScannerContext context) {
     expectedLevel = 0;
     isBlockAlreadyReported = false;
     lastCheckedLine = 0;
     isInAnonymousClass.clear();
```
m-g-sonar: If you clear the list at the end of the scan, no need to clear it before.

PR #1076, opened 2016-11-10
SONARJAVA-1942 do not stack TRUE/FALSE for conditional expressions

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -385,6 +395,42 @@ private void handleBranch(CFG.Block programPosition, Tree condition, boolean che
...
 
+  private static boolean isRequiredForExpression(Tree tree) {
+    Tree current;
+    Tree parent = tree;
+    do {
+      current = parent;
+      parent = parent.parent();
+    } while (parent.is(Tree.Kind.PARENTHESIZED_EXPRESSION, Tree.Kind.CONDITIONAL_AND, Tree.Kind.CONDITIONAL_OR));
+    Optional<Tree> condition = getCondition(parent);
+    return !condition.isPresent() || !current.equals(condition.get());
```
benzonico: use map.

PR #1091, opened 2016-11-16
Increase coverage

saberduck:
fix quality flaw: increase coverage

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/synchronization/DoubleCheckedLockingCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/synchronization/DoubleCheckedLockingCheck.java
@@ -35,11 +35,11 @@ import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
 import org.sonar.plugins.java.api.tree.StatementTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
-import javax.annotation.CheckForNull;
 import java.util.Collection;
 import java.util.Deque;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Optional;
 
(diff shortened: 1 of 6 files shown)


java-checks/src/main/java/org/sonar/java/checks/synchronization/DoubleCheckedLockingCheck.java
```diff
@@ -123,31 +125,27 @@ private boolean insideCriticalSection() {
...
 
-  private static boolean isField(ExpressionTree expressionTree) {
-    Symbol symbol = symbolFromVariable(expressionTree);
-    if (symbol == null) {
-      return false;
-    }
-    return symbol.isVariableSymbol() && symbol.owner().isTypeSymbol();
+  private static Optional<Symbol> isField(ExpressionTree expressionTree) {
+    return symbolFromVariable(expressionTree)
+      .filter(s -> s.isVariableSymbol() && s.owner().isTypeSymbol());
```
m-g-sonar: You didn't covered the case where it's not a variable:
```
  if (MyClass.class == null) {
      synchronized (this) {
        if (MyClass.class == null) {
          // ...
```

java-checks/src/test/files/checks/WriteObjectTheOnlySynchronizedMethodCheck.java
```diff
@@ -16,7 +15,6 @@ public void bounce(float angle, float velocity) {
   }
 
   private synchronized void writeObject(ObjectOutputStream stream) throws IOException { // Compliant, maybe FN
```
m-g-sonar: why `maybe` ? if it's not considered because of the other synchronized method, from the inner class, then it should be documented for that behavior. Otherwise, just cut the exploration of classtree bodies in the `SynchronizationVisitor`.

PR #1094, opened 2016-11-17
SONARJAVA-1918 Fix wrong reporting of nesting penalty

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/CognitiveComplexityMethodCheck.java
```diff
@@ -116,10 +118,12 @@ private void increaseComplexityByOne(Tree tree) {
     private void increaseComplexity(Tree tree, int increase) {
       complexity += increase;
       if(!ignored.contains(tree)) {
```
benzonico: I might be wrong but wouldn't adding the succeeding if tree to the ignored set achieve the purpose simplier ?

m-g-sonar: It won't work, as it will then report nothing for that `if` tree, while we still have to report the `+1`, but without the nesting penalty.

PR #1101, opened 2016-11-22
SONARJAVA-1788 Rule S2629 Preconditions and logging arguments should …

saberduck:
Implements SONARJAVA-1788 Rule S2629


java-checks/src/main/java/org/sonar/java/checks/methods/AbstractMethodDetection.java
```diff
@@ -22,6 +22,9 @@
 import com.google.common.collect.ImmutableList;
 import org.sonar.java.matcher.MethodMatcher;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
```
SonarTech: MINOR Remove this unused import 'org.sonar.plugins.java.api.tree.IdentifierTree'. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

java-checks/src/main/java/org/sonar/java/checks/methods/AbstractMethodDetection.java
```diff
@@ -22,6 +22,9 @@
...
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
```
SonarTech: MINOR Remove this unused import 'org.sonar.plugins.java.api.tree.MemberSelectExpressionTree'. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,175 @@
...
+public class LazyArgEvaluationCheck extends BaseTreeVisitor implements JavaFileScanner {
+
+  private static final String JAVA_LANG_STRING = "java.lang.String";
+
+  static final class Methods {
+    static final class Guava {
+      static final MethodMatcher PRECONDITIONS = MethodMatcher.create()
+        .name("checkState").withAnyParameters();
+    }
+    static final class Logging {
```
SonarTech: MAJOR Add a private constructor to hide the implicit public one. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS1118)

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,175 @@
...
+      }
+    }
+  }
+
+  @Override
+  public void visitCatch(CatchTree tree) {
+    // cut the visit on catch blocks, because we don't mind some performance loss on exceptional paths
+  }
+
+  private List<MethodMatcher> getMethodInvocationMatchers() {
```
SonarTech: MAJOR Make "getMethodInvocationMatchers" a "static" method. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS2325)

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Rule(key = "S2629")
+public class LazyArgEvaluationCheck extends BaseTreeVisitor implements JavaFileScanner {
+
+  private static final String JAVA_LANG_STRING = "java.lang.String";
+
+  static final class Methods {
```
SonarTech: MAJOR Add a private constructor to hide the implicit public one. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS1118)

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,175 @@
...
+    static final class Guava {
```
SonarTech: MAJOR Add a private constructor to hide the implicit public one. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS1118)

SonarTech:
SonarQube analysis reported 7 issues
* MAJOR 4 major
* MINOR 3 minor

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,183 @@
...
+      if (methodMatcher.matches(tree)) {
+        onMethodInvocationFound(tree);
+        break;
+      }
+    }
+  }
+
+  @Override
+  public void visitCatch(CatchTree tree) {
+    // cut the visit on catch blocks, because we don't mind some performance loss on exceptional paths
```
m-g-sonar: Okay so that's the reason why you don't want to rely on `AbstractMethodDetection`. But then what about this case (nested code in catch blocks):
```
MyClass myField;

public void exceptionalPaths() {
  try {
    myField = tryGetMyField();
  } catch (Exception e) {
    LOG.info("Something went wrong: " + message); // ignored, - OK -
    myField = new MyClass() {
      @Overidde
      void doSomethingAllTheTime() {
        LOG.info("logging all the time consuming resources for nothing " + computeValue() + generateStuff()); // consequently ignored, but which should raise an issue
      }
    };
  }
}
```

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,183 @@
...
+    this.context = context;
+    if (context.getSemanticModel() == null) {
+      return;
+    }
+    scan(context.getTree());
+  }
+
+  @Override
+  public void visitMethodInvocation(MethodInvocationTree tree) {
+    for (MethodMatcher methodMatcher : getMethodInvocationMatchers()) {
```
m-g-sonar: You should use `MethodMatcherCollection.anyMatch()`

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+  }
+
+  @Override
+  public void visitCatch(CatchTree tree) {
+    // cut the visit on catch blocks, because we don't mind some performance loss on exceptional paths
+  }
+
+  private static List<MethodMatcher> getMethodInvocationMatchers() {
```
m-g-sonar: You should define a `private static final MethodMatcherCollection` instead, containing all your method matchers, and use its methods for checking.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.stream.Stream;
+
+@Rule(key = "S2629")
+public class LazyArgEvaluationCheck extends BaseTreeVisitor implements JavaFileScanner {
+
+  private static final String JAVA_LANG_STRING = "java.lang.String";
+
+  static final class Methods {
+
+    static final class Guava {
```
m-g-sonar: Do you really think it worth the effort to declare static classes for `guava` and `logging` ?
It clearly sort things out, but the need to have private empty constructors on top of them makes it a bit absurd.

I would only make a bunch of static variables, but it's ok for me if you prefer to let it that way.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,183 @@
...
+      static final MethodMatcher PRECONDITIONS = MethodMatcher.create()
+        .name("checkState").withAnyParameters();
```
m-g-sonar: What about the owner of the method?! Basically you are targeting here any method called `checkState` , from any class, while it should only target methods from  `com.google.common.base.Preconditions`.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,183 @@
...
+      static final MethodMatcher TRACE = slf4jPrototype().name("trace");
+      static final MethodMatcher DEBUG = slf4jPrototype().name("debug");
+      static final MethodMatcher INFO = slf4jPrototype().name("info");
+      static final MethodMatcher WARN = slf4jPrototype().name("warn");
+      static final MethodMatcher ERROR = slf4jPrototype().name("error");
+
+      private Logging() {}
+
+      static MethodMatcher slf4jPrototype() {
+        return MethodMatcher.create().addParameter(JAVA_LANG_STRING).addParameter(TypeCriteria.anyType());
```
m-g-sonar: Same remark about owner of the methods here, are you sure that you want to target any method called `trace/debug/info...`. ? 
The FP rate will be pretty high on this I guess. You should  configure the method matcher to cover only methods from `org.slf4j.Logger`.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
saberduck: I was thinking to be bit more generic at first, but you are right about FPs

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
saberduck: Refactored

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
saberduck: This is a corner case IMO, not worth to bother with. It's a misuse if exception is used in normal logic flow.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
saberduck: Refactored

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
m-g-sonar: I agree that it's a corner case. However, even as being a misuse it does not justify simply ignoring the case. We can not rely on the fact that our users will always code nicely.

Catching such cases can be done using a simple stack in which we push every time we enter a catchTree/classTree/methodTree/lambdaTree, and pop every time we leave the node. Simply looking at the top of the stack when handling method invocation should tell us if it's OK or not.

Johnnei:
As a small notice: Please be aware of the 'is[level]Enabled' gaurding style for non lazy-evaluated statements:
[SLF4J Variant](http://www.slf4j.org/apidocs/org/slf4j/Logger.html#isDebugEnabled())
[Log4J2 Variant](http://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Logger.html#isDebugEnabled())

Example:
`    if (LOGGER.isDebugEnabled()) {
        LOGGER.debug("my" + "expensive" + "logging");
    }`

I didn't really verify if this rule will flag for those cases but as it's also not mentioned in the RSPEC so I'd mention it. Better safe than sorry :) @saberduck @m-g-sonar

m-g-sonar:
Thanks for the feedback @Johnnei . 
We already discussed the subject internally, as we also had these cases in mind. This PR contains a first naive approach of the rule, but we definitely want it to cover such cases. That's notably one of the reasons why we didn't merge the rule already! :)
I also think that we should mention them in the RSPEC.

saberduck:
We account for Guava `Preconditions` , SLF4J `Logger` and JDK `Logger` now. Log4J is not accounted for.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,287 @@
...
+  private static final MethodMatcher PRECONDITIONS = MethodMatcher.create()
+    .typeDefinition("com.google.common.base.Preconditions")
+    .name("checkState")
+    .withAnyParameters();
+
+  private static final MethodMatcher JUL_LOG = MethodMatcher.create()
+    .typeDefinition(JUL_LOGGER)
+    .name("log")
+    .addParameter(TypeCriteria.anyType())
+    .addParameter(String.class.getCanonicalName());
```
m-g-sonar: Reuse the `STRING` constant.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+    if (visitor.hasBinaryExpression) {
+      msg.append("Use the built-in formatting to construct this argument.");
+    }
+    return new JavaFileScannerContext.Location(msg.toString(), stringArgument);
+  }
+
+  private static Stream<ExpressionTree> findStringArg(MethodInvocationTree mit) {
+    return mit.arguments().stream()
+      .filter(arg -> arg.symbolType().is(String.class.getCanonicalName()));
```
m-g-sonar: Reuse the `STRING` constant

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    enum SLF4J_LEVELS implements LogLevels {
+      TRACE,
+      DEBUG,
+      INFO,
+      WARN,
+      ERROR;
+
+      final List<MethodMatcher> log = varArgVariants(SLF4J_LOGGER, toString().toLowerCase(Locale.ROOT));
```
m-g-sonar: The `varArgVariants` method is only called from here, so why not put it in this enum as private static method instead? And removing the toLowerCase from the method call to include in it the method body.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,287 @@
...
+  @Override
+  public void visitMethodInvocation(MethodInvocationTree tree) {
+    if (LAZY_ARG_METHODS.anyMatch(tree) && !insideCatchStatement() && !insideLevelTest()) {
+      onMethodInvocationFound(tree);
+    }
+  }
+
+  @Override
+  public void visitIfStatement(IfStatementTree ifTree) {
+    Slf4jTestVisitor slf4jTestVisitor = new Slf4jTestVisitor();
```
m-g-sonar: I'm not sure the name of the visitor is appropriate, as from what I can see, we are not only checking for SLF4J level test methods, but also "JUL".

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  @Override
+  public void visitCatch(CatchTree tree) {
+    stackAndContinue(tree, super::visitCatch);
+  }
+
+  @Override
+  public void visitMethod(MethodTree tree) {
+    stackAndContinue(tree, super::visitMethod);
```
m-g-sonar: May worth a comment why you need to stack a method, as you never check for it in the stack, even if I guess it's done in order to avoid FPs in anonymous classes.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,287 @@
...
+    enum JUL_LEVELS implements LogLevels {
+      SEVERE,
+      WARNING,
+      INFO,
+      CONFIG,
+      FINE,
+      FINER,
+      FINEST;
+
+      final List<MethodMatcher> log = ImmutableList.of(loggerProto(JUL_LOGGER).name(toString().toLowerCase(Locale.ROOT)));
```
m-g-sonar: I'm not convinced about the usage of the log/test fields in both enums. I wonder if they are really useful. From what I can see, the `log()` and `test()` methods will be called only once, in the static initializers of the rule, so maybe we can win some place and simply put their initializers directly in the implementations of the methods. It won't make it less clear.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,294 @@
...
+
+  interface LogLevels {
+    List<MethodMatcher> log();
+
+    MethodMatcher test();
+
+    static MethodMatcher levelTestMatcher(String typeDefinition, String level) {
+      return MethodMatcher.create()
+        .typeDefinition(typeDefinition)
+        .name(String.format("is%c%sEnabled", level.charAt(0), level.toLowerCase(Locale.ROOT).substring(2)))
```
m-g-sonar: Are you sure about `.substring(2)` ? With `INFO`, I'll get `isIfoEnabled`...

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,294 @@
...
+
+      @Override
+      public MethodMatcher test() {
+        return levelTestMatcher(SLF4J_LOGGER, toString());
+      }
+
+      private static List<MethodMatcher> slf4jVariants(Supplier<MethodMatcher> prototype) {
+        return ImmutableList.of(
+          prototype.get().parameters(STRING),
+          prototype.get().parameters(STRING, TypeCriteria.anyType()),
```
m-g-sonar: I would make a static import of `TypeCriteria.anyType()` to reduce a bit the verbosity of that block.

java-checks/src/test/files/checks/LazyArgEvaluationCheck.java
```diff
@@ -0,0 +1,142 @@
...
+  void jul() {
+    logger.finest("Unable to open file " + csvPath);  // Noncompliant {{Use the built-in formatting to construct this argument.}}
+    logger.finer("Unable to open file " + csvPath);  // Noncompliant {{Use the built-in formatting to construct this argument.}}
+    logger.fine("Unable to open file " + csvPath);  // Noncompliant {{Use the built-in formatting to construct this argument.}}
+    logger.config("Unable to open file " + csvPath);  // Noncompliant {{Use the built-in formatting to construct this argument.}}
+    logger.info("Unable to open file " + csvPath);  // Noncompliant {{Use the built-in formatting to construct this argument.}}
+    logger.warning("Unable to open file " + csvPath);  // Noncompliant {{Use the built-in formatting to construct this argument.}}
+    logger.severe("Unable to open file " + csvPath);  // Noncompliant {{Use the built-in formatting to construct this argument.}}
+
+    if (logger.isFinestEnabled()) {
```
m-g-sonar: There is something fishy with your implementation, and in particular your if-condition visitor (especially since the name is actually not matching).
The following code does NOT raise an issue, where I would expect one (you actually don't test this case...):
```
  void foo(boolean b) {
    if (b) {
      logger.finest("Unable to open file " + csvPath);  // Noncompliant
    }
  }
```

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
saberduck: It's a bug, thanks for finding this!

java-checks/src/test/files/checks/LazyArgEvaluationCheck.java
saberduck: Added testcase

SonarTech:
* MAJOR 1 major

java-checks/src/test/files/checks/LazyArgEvaluationCheck.java
```diff
@@ -116,29 +116,13 @@ void jul() {
...
-      logger.info("Unable to open file " + csvPath);  // Compliant - inside if test
-    }
-    if (logger.isWarningEnabled()) {
-      logger.warning("Unable to open file " + csvPath);  // Compliant - inside if test
-    }
-    if (logger.isSevereEnabled()) {
-      logger.severe("Unable to open file " + csvPath);  // Compliant - inside if test
+
+    if (logger.isLoggable(Level.INFO)) {
+      logger.trace("Unable to open file " + csvPath);  // Compliant - FP, we don't verify that level in "if" matches actual level used in logging
```
m-g-sonar: If we are not raising an issue where we could/should, then it's a FN, not a FP.

PR #1105, opened 2016-11-23
SONARJAVA-1964 Fix false positive on S1699 when called method is in t…

mjimenezp:
…he body of an anonymous class or lambda

This pull request contains the fix for the FP when a call to a method is in the body of an anonymous class or lambda. 

- [X] Unit tests are passing and you provided a unit test for your fix
- [X] ITs should pass : To run ITs locally, checkout the [README](https://github.com/SonarSource/sonar-java/blob/master/README.md) of the project.
- [X] If there is a [Jira](http://jira.sonarsource.com/browse/SONARJAVA) ticket available, please make your commits and pull request start with the ticket number (SONARJAVA-1964)

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/ConstructorCallingOverridableCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ConstructorCallingOverridableCheck.java
@@ -33,6 +33,8 @@ import org.sonar.plugins.java.api.tree.IdentifierTree;
 import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodInvocationTree;
 import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.LambdaExpressionTree;
+import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.Tree.Kind;
 
@@ -86,11 +88,21 @@ public class ConstructorCallingOverridableCheck extends IssuableSubscriptionVisi
       }
       if (isInvocationOnSelf) {
         Symbol symbol = methodIdentifier.symbol();
-        if (isOverridableMethod(symbol) && isMethodDefinedOnConstructedType(symbol) &&!isInsideAnonymousClass(methodIdentifier)) {
+        if (isOverridableMethod(symbol) && isMethodDefinedOnConstructedType(symbol)) {
           reportIssue(methodIdentifier, "Remove this call from a constructor to the overridable \"" + methodIdentifier.name() + "\" method.");
         }
       }
       super.visitMethodInvocation(tree);
+    }
+
+    @Override
+    public void visitLambdaExpression(LambdaExpressionTree lambdaExpressionTree) {
+      // skip lambdas
+    }
+
+    @Override
+    public void visitNewClass(NewClassTree tree) {
+      // skip new class
     }
 
     private boolean is(ExpressionTree expression, String match) {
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/ConstructorCallingOverridableCheck.java
```diff
@@ -87,7 +86,7 @@ public void visitMethodInvocation(MethodInvocationTree tree) {
       }
       if (isInvocationOnSelf) {
         Symbol symbol = methodIdentifier.symbol();
-        if (isOverridableMethod(symbol) && isMethodDefinedOnConstructedType(symbol)) {
+        if (isOverridableMethod(symbol) && isMethodDefinedOnConstructedType(symbol) &&!isInsideAnonymousClass(methodIdentifier)) {
```
m-g-sonar: There is a simpler approach than what you implemented. It works perfectly, but it does not take advantage of the `ConstuctorBodyVisitor` being a `BaseTreeVisitor`.

I would remove the condition you added here, which require exploring parents in order to determine if you are part of a lambda or a new class tree, and replace it by simply overriding the tree visitors of the lambda and newClassTree visitors, skipping their analysis:

```
   @Override
      // skip lambdas
    }

    @Override
    public void visitNewClass(NewClassTree tree) {
      // skip new class
    }
```

Once done, no need to add extra logic, the exploration of these trees will be skipped, avoiding raising FPs in the process. The rest of your code can also be removed.

java-checks/src/test/files/checks/ConstructorCallingOverridableCheck.java
```diff
@@ -138,3 +138,68 @@ void foo(Easy easy) {
...
+    yaml = memoize(() -> loadYamlConfig("_config.yml")); // Compliant
+
+    data = memoize(() -> getResourceList() // Compliant
+      .stream()
+      .filter(path -> path.startsWith("_data/"))
+      .collect(toMap(Site::nameWithoutExtension, this::readYaml))
+    );
+  }
+
+  @SuppressWarnings("unchecked")
```
m-g-sonar: The annotation is not relevant to the test and should be removed.

java-checks/src/test/files/checks/ConstructorCallingOverridableCheck.java
```diff
@@ -138,3 +138,68 @@ void foo(Easy easy) {
...
+  private Map<String, Object> loadYamlConfig(String configFile) {
+    Path configPath = Paths.get(configFile);
```
m-g-sonar: You don't need to fill the body of that method, you just need it to use it as lambda.

java-checks/src/test/files/checks/ConstructorCallingOverridableCheck.java
```diff
@@ -138,3 +138,68 @@ void foo(Easy easy) {
...
+class G{
+  G(int value) {
+    this.profileActivator = new ProfileActivator() {
+      public void activate() throws ProfileDataException {
+        activateDeferredProfile();  // Compliant
+      }
+    };
+    ProfileDeferralMgr.registerDeferral(this.profileActivator);
+  }
+  void activateDeferredProfile()  throws ProfileDataException{
```
m-g-sonar: Replace the exception by `Exception`, so your test is self contained.

mjimenezp:
I just push the commit with the required fixes, can you give it a look to check if all is right,please?

Thanks

m-g-sonar:
Everything is alright now @mjimenezp! I'll merge the commit with our master as soon as we released version 4.3. Your fix will then be part of verison 4.4 of the SonarQube Java Analyzer.

Cheers,
Michael

PR #1106, opened 2016-11-24
SONARJAVA-1976 Display program states and multiple parents

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/ExplodedGraph.java
```diff
@@ -105,12 +105,14 @@ public String toString() {
     public final ProgramState programState;
     @Nullable
     public Node parent;
+    public final List<Node> cacheHits;
```
benzonico: This is poorly named : cache hit has absolutely no significance here... Those are parents nodes.

java-frontend/src/test/java/org/sonar/java/viewer/EGViewer.java
```diff
@@ -42,6 +42,7 @@
 
   private static final ActionParser<Tree> PARSER = JavaParser.createParser(StandardCharsets.UTF_8);
   private final Viewer viewer;
+  private static final boolean SHOW_CACHE = true;
```
benzonico: Naming! notion of cache should probably not appear anywhere except for the walker : this is implementation detail.

java-frontend/src/main/java/org/sonar/java/resolve/JavaSymbol.java
```diff
@@ -645,6 +645,11 @@ public MethodTree declaration() {
     public boolean isParametrized() {
       return !typeVariableTypes.isEmpty();
     }
+
+    @Override
```
benzonico: missing a test.

java-frontend/src/test/java/org/sonar/java/viewer/EGViewer.java
m-g-sonar: This is an implementation details that I want to see appearing in the EG Viewer, which is a debugging tool. Knowing that a link come from the cache or not seems then important to me.

java-frontend/src/main/java/org/sonar/java/resolve/JavaSymbol.java
m-g-sonar: tested

PR #1109, opened 2016-11-25
SONARJAVA-1959 Fix recursion issue

benzonico:


java-frontend/src/main/java/org/sonar/java/se/checks/SECheck.java
```diff
@@ -94,17 +94,18 @@ public void reportIssue(Tree tree, String message, Set<List<JavaFileScannerConte
     Symbol lastEvaluated = currentNode.programState.getLastEvaluated();
     while (node != null) {
       ExplodedGraph.Node finalNode = node;
+      node = node.parent;
```
m-g-sonar: I would have let that line at the end of the while() block, as you are not using `node` anymore before reassigning it to the parent.

PR #1111, opened 2016-11-26
SONARJAVA-1972 - S2095 Allow users to specify types to be excluded fr…

fergarrui:
* Added a comma separated list as a parameter to exclude types in the check.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
@@ -19,15 +19,9 @@
  */
 package org.sonar.java.se.checks;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import javax.annotation.Nullable;
+import org.apache.commons.lang.StringUtils;
 import org.sonar.check.Rule;
 import org.sonar.check.RuleProperty;
-import org.sonar.java.cfg.CFG;
 import org.sonar.java.matcher.MethodMatcher;
 import org.sonar.java.model.ExpressionUtils;
 import org.sonar.java.se.CheckerContext;
@@ -44,12 +38,19 @@ import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.IdentifierTree;
 import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodInvocationTree;
-import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.ReturnStatementTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.TryStatementTree;
 import org.sonar.plugins.java.api.tree.VariableTree;
+
+import javax.annotation.Nullable;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 
 @Rule(key = "S2095")
 public class UnclosedResourcesCheck extends SECheck {
@@ -58,13 +59,12 @@ public class UnclosedResourcesCheck extends SECheck {
     OPENED, CLOSED
   }
 
-  private static final String DEFAULT_EXCLUDED_TYPES = "";
-  
   @RuleProperty(
-    key = "Excluded types",
-    description = "Comma separated list of the fully qualified name of excluded types. Example: org.apache.hadoop.fs.FileSystem",
-    defaultValue = "" + DEFAULT_EXCLUDED_TYPES)
-  public String excludedTypes = DEFAULT_EXCLUDED_TYPES;
+    key = "excludedResourceTypes",
+    description = "Comma separated list of the excluded resource types, using fully qualified names (example: \"org.apache.hadoop.fs.FileSystem\")",
+    defaultValue = "")
+  public String excludedTypes = "";
+  private final List<String> excludedTypesList = new ArrayList<>();
   
   private static final String JAVA_IO_AUTO_CLOSEABLE = "java.lang.AutoCloseable";
   private static final String JAVA_IO_CLOSEABLE = "java.io.Closeable";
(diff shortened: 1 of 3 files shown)


java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -72,9 +81,21 @@
...
   };
+  
+  private static final List<String> excludedTypesList = new ArrayList<>();
+  
   private static final MethodMatcher[] CLOSEABLE_EXCEPTIONS = new MethodMatcher[] {
     MethodMatcher.create().typeDefinition("java.nio.file.FileSystems").name("getDefault").withoutParameter()
   };
+  
+  @Override
+  public void init(MethodTree methodTree, CFG cfg) {
```
m-g-sonar: Please do not override the `init()` method. Symbolic execution based checks are sensitive, and in this case we don't want to touch anything in the initialization phase.

The best approach in order to retrieve the list of excluded resource types is to use a dedicated (private) getter, which lazily initialize the list (which then should not be static, with all the side effects you noticed).

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -116,34 +137,39 @@ private static boolean needsClosing(Type type) {
         return false;
       }
     }
+    for (String excludedType : excludedTypesList) {
```
m-g-sonar: If the list is `static`, then you don't have to remove the static modifier everywhere. However, I would prefer you to use a dedicated getter lazily loading the configuration of excluded types.

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -36,27 +44,28 @@
...
 @Rule(key = "S2095")
 public class UnclosedResourcesCheck extends SECheck {
 
   private enum Status {
     OPENED, CLOSED
   }
 
+  private static final String DEFAULT_EXCLUDED_TYPES = "";
+  
+  @RuleProperty(
```
m-g-sonar: I updated the description of the rule in order to include the parameter. Please follow the description from here for keys and description: [RSPEC-3591](https://jira.sonarsource.com/browse/RSPEC-3591)

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -72,9 +81,21 @@
     "com.sun.org.apache.xml.internal.security.utils.UnsyncByteArrayOutputStream",
     "org.springframework.context.ConfigurableApplicationContext"
   };
+  
+  private static final List<String> excludedTypesList = new ArrayList<>();
```
m-g-sonar: Put this list (which should not be static) next to the other rule property-related field. Grouping will help.

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -116,34 +137,39 @@ private static boolean needsClosing(Type type) {
...
+    for (String excludedType : excludedTypesList) {
+      if (type.is(excludedType)) {
+        return false;
+      }
+    }
     return isCloseable(type);
   }
 
-  private static boolean isCloseable(Type type) {
+  private boolean isCloseable(Type type) {
```
m-g-sonar: I'm not sure you need to remove all of these `static` modifiers. Some of them are probably still valid (this one notably)

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -19,7 +19,15 @@
  */
 package org.sonar.java.se.checks;
 
+import java.util.ArrayList;
```
m-g-sonar: Please discard these changes, and only limit the modifications to the new imports.

fergarrui:
Hi, I did the changes request:
* Discarded the changes in the imports (I'm using Eclipse formatter suggested by Sonar, that's why they were reordered, I'll keep in mind not to reorder them again).
* Rule properties updated to match [RSPEC-3591](https://jira.sonarsource.com/browse/RSPEC-3591)
* Grouped parameter-related properties. Also removed static modifier from the list holding the excluded types.
* Removed init() override and added a lazy load to populate the list.
* Added back some static modifiers, but I can't add most of them.

Thanks for your suggestions and patience! Please let me know if you want more improvements/changes.

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -116,29 +125,43 @@ private static boolean needsClosing(Type type) {
...
+    if ( excludedTypesList.isEmpty() && !StringUtils.isBlank(excludedTypes)) {
+      for (String excludedType : excludedTypes.split(",")) {
+        excludedTypesList.add(excludedType.trim());
+      }
+    }
+    return excludedTypesList;
+  }
 
-  private static boolean isCloseable(Type type) {
+  private boolean isCloseable(Type type) {
```
m-g-sonar: This one can still be static.

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -116,29 +125,43 @@ private static boolean needsClosing(Type type) {
...
-  private static boolean isOpeningResource(NewClassTree syntaxNode) {
+  private boolean isOpeningResource(NewClassTree syntaxNode) {
     if (isWithinTryHeader(syntaxNode)) {
       return false;
     }
     return needsClosing(syntaxNode.symbolType());
   }
 
-  private static boolean isWithinTryHeader(Tree syntaxNode) {
+  private boolean isWithinTryHeader(Tree syntaxNode) {
```
m-g-sonar: And when `isTryStatementResource` is static, this one can be static as well.

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -116,29 +125,43 @@ private static boolean needsClosing(Type type) {
...
     final Tree parent = syntaxNode.parent();
     if (parent.is(Tree.Kind.VARIABLE)) {
       return isTryStatementResource((VariableTree) parent);
     }
     return false;
   }
 
-  private static boolean isTryStatementResource(VariableTree variable) {
+  private boolean isTryStatementResource(VariableTree variable) {
```
m-g-sonar: This one can be static.

fergarrui:
Changes done :)

PR #1116, opened 2016-11-29
Add option to dump unittest test files with assertion comments

saberduck:
This PR adds possibility to dump unittests's test file with `noncompliant` and `flow@id` comments added as reported from the check. This can be used to compare and merge with expected comments. This is enabled with System property `dump.unittests` .

To run with unittest dumping enabled execute build like this 
`mvn -Ddump.unittests  test -pl java-frontend -am`
dumped files will be available under `java-frontend/target/actual`

### DIFF
--- a/java-frontend/src/test/java/org/sonar/java/se/JavaCheckVerifier.java
+++ b/java-frontend/src/test/java/org/sonar/java/se/JavaCheckVerifier.java
@@ -22,8 +22,11 @@ package org.sonar.java.se;
 import com.google.common.annotations.Beta;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.HashMultiset;
+import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multiset;
 import org.apache.commons.io.FileUtils;
@@ -35,6 +38,11 @@ import org.sonar.java.model.VisitorsBridgeForTests;
 import org.sonar.plugins.java.api.JavaFileScanner;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -49,6 +57,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
 
 import static java.util.stream.Collectors.joining;
 import static org.fest.assertions.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


java-frontend/src/test/java/org/sonar/java/se/JavaCheckVerifier.java
```diff
@@ -188,7 +198,60 @@ private void scanFile(String filename, JavaFileScanner[] checks, Collection<File
...
+      .filter(m -> Objects.nonNull(m.getLine()))
+      .flatMap(m -> Stream.concat(
+        Stream.of(Maps.immutableEntry(m.getLine(), String.format("// Noncompliant {{%s}}", m.getMessage()))),
+        flowComments(m)))
+      .forEach(result::put);
+    return result.build();
+  }
+
+  private static Stream<Map.Entry<Integer, String>> flowComments(AnalyzerMessage m) {
+    return m.flows.stream().flatMap(f -> f.stream().map(fm -> Maps.immutableEntry(fm.getLine(), String.format("// flow@id {{%s}}", fm.getMessage()))));
```
m-g-sonar: and so, how do you differentiate the flows?

java-frontend/src/test/java/org/sonar/java/se/JavaCheckVerifier.java
```diff
@@ -188,7 +198,60 @@ private void scanFile(String filename, JavaFileScanner[] checks, Collection<File
...
+
+    try {
+      ListMultimap<Integer, String> comments = computeComments(issues);
+      List<String> lines = Files.readAllLines(source).stream()
+        .map(line -> line.replaceAll("// (Noncompliant|flow).*$", ""))
+        .collect(Collectors.toList());
+
+      List<String> dump = IntStream.range(0, lines.size())
+        .mapToObj(idx -> new Line(idx, lines.get(idx)))
+        .map(l ->  l.s + comments.get(l.idx + 1).stream().collect(joining(",")))
```
m-g-sonar: Maybe there is a nicer way to dump the results: 
```
if (null == a && b == a)// flow@id {{}},// flow@id {{}},// flow@id {{}}
```

java-frontend/src/test/java/org/sonar/java/se/JavaCheckVerifier.java
```diff
@@ -188,7 +198,60 @@ private void scanFile(String filename, JavaFileScanner[] checks, Collection<File
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static ListMultimap<Integer, String> computeComments(Set<AnalyzerMessage> issues) {
+    ImmutableListMultimap.Builder<Integer, String> result = ImmutableListMultimap.builder();
+    issues.stream()
+      .filter(m -> Objects.nonNull(m.getLine()))
+      .flatMap(m -> Stream.concat(
+        Stream.of(Maps.immutableEntry(m.getLine(), String.format("// Noncompliant {{%s}}", m.getMessage()))),
```
m-g-sonar: Like for the flows, it would be nice to have a way to identify flows

m-g-sonar:
I'm not convinced by the feature. To me it's a lot of code for not a lot at the end.
I would prefer an option added to the verifier in order to force flow validation, with a nice log in case of fail, rather than a dump and compare manually.

saberduck:
OK, I decided that this PR is perhaps not the best way how to approach flow unit testing

PR #1124, opened 2016-12-01
SONARJAVA-1963 remove syntaxNode from Constraint

saberduck:
Removes `syntaxNode` from `Constraint`. This changes the identity (`hashcode`) of `Constraint` and thus identity of `Node`. Which has the effect that the cache is used more often because we can internalize more nodes of the EG. 
However this change lead to two issues
1. when reporting the issue, flows need to be computed using all parents of the node. 
2. when calling endOfExecutionPath (EEP) , not all of the parents of the node are calculated

To solve #1 I first moved `SECheck#flow` to separate class and simplified the algorithm to use just recursion instead of mix of custom stack and recursive approaches. After I modified to use all parents to compute flow.
To solve #2 I deffer the call of EEP until the EG is fully explored or exploration is interrupted. This is achieved by maintaining a list of exit nodes in EGWalker.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
@@ -81,6 +81,7 @@ import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Deque;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -126,7 +127,7 @@ public class ExplodedGraphWalker {
   ConstraintManager constraintManager;
   private boolean cleanup = true;
   MethodBehavior methodBehavior;
-  private List<ExplodedGraph.Node> endOfExecutionPath;
+  private Set<ExplodedGraph.Node> endOfExecutionPath;
 
   public static class ExplodedGraphTooBigException extends RuntimeException {
 
@@ -189,7 +190,7 @@ public class ExplodedGraphWalker {
     methodTree = tree;
     constraintManager = new ConstraintManager();
     workList = new LinkedList<>();
-    endOfExecutionPath = new ArrayList<>();
+    endOfExecutionPath = new HashSet<>();
     if(DEBUG_MODE_ACTIVATED) {
       LOG.debug("Exploring Exploded Graph for method " + tree.simpleName().name() + " at line " + ((JavaTree) tree).getLine());
     }
--- a/java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
@@ -187,6 +187,7 @@ public class ExplodedGraphWalker {
     methodTree = tree;
     constraintManager = new ConstraintManager();
     workList = new LinkedList<>();
+    List<ExplodedGraph.Node> endOfExecutionPath = new ArrayList<>();
     if(DEBUG_MODE_ACTIVATED) {
       LOG.debug("Exploring Exploded Graph for method " + tree.simpleName().name() + " at line " + ((JavaTree) tree).getLine());
     }
@@ -207,7 +208,7 @@ public class ExplodedGraphWalker {
       programPosition = node.programPoint;
       programState = node.programState;
       if (programPosition.block.successors().isEmpty()) {
-        handleEndOfExecutionPath();
+        endOfExecutionPath.add(node);
         continue;
       }
       try {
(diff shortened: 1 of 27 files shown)


saberduck:
WIP - needs additional tests and cleanup

SonarTech:

Note: The following issues were found on lines that were not modified in the pull request. Because these issues can't be reported as line comments, they are summarized here:

1. CRITICAL [DivisionByZeroCheck.java#L109](https://github.com/SonarSource/sonar-java/blob/a5b0dcdb4cea95fc96a27eb3eba71519868d8596/java-frontend/src/main/java/org/sonar/java/se/checks/DivisionByZeroCheck.java#L109): The Cyclomatic Complexity of this method "visitBinaryExpression" is 11 which is greater than 10 authorized. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

saberduck:
Looks like git orders commits by date and commits are mixed up after rebase. https://github.com/SonarSource/sonar-java/pull/1124/commits/d2d83b1beb64d6b2a62a9395b4155dd9175fe609  Ruling should be last commit

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -187,6 +190,7 @@ private void execute(MethodTree tree) {
     methodTree = tree;
     constraintManager = new ConstraintManager();
     workList = new LinkedList<>();
+    endOfExecutionPath = new HashSet<>();
```
m-g-sonar: Why not use a set which preserve the order, like `LinkedHashSet`? It would allow to keep the current test working and keep predictability.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -89,7 +88,7 @@ public Pop(ProgramState programState, List<SymbolicValue> result) {
 
   private ProgramState(PMap<Symbol, SymbolicValue> values, PMap<SymbolicValue, Integer> references,
     PMap<SymbolicValue, Constraint> constraints, PMap<ExplodedGraph.ProgramPoint, Integer> visitedPoints,
-    Deque<SymbolicValue> stack, SymbolicValue returnSymbolicValue) {
+    Deque<SymbolicValue> stack, @Nullable SymbolicValue returnSymbolicValue) {
```
saberduck: I agree this is noise, and shouldn't be part of PR

java-frontend/src/test/files/se/LocksNotUnlockedCheck.java
```diff
@@ -189,7 +190,7 @@ public void while_statement() {
       lock.tryLock();
     }
     lock.unlock();
-    while (foo) {
+    while (Random.nextBoolean()) {
```
saberduck: This was temporary and I forgot to put test back to original state after fixing the problem with flow paths. Thanks for catching it.

java-frontend/src/test/java/org/sonar/java/se/MethodYieldTest.java
```diff
@@ -245,8 +250,19 @@ public void catch_class_cast_exception() throws Exception {
     assertThat(behaviorCache.values()).hasSize(1);
     MethodBehavior methodBehavior = behaviorCache.values().iterator().next();
     assertThat(methodBehavior.yields()).hasSize(2);
-    assertThat(methodBehavior.yields().get(1).resultIndex).isEqualTo(0);
-    assertThat(methodBehavior.yields().get(0).resultConstraint.isNull()).isTrue();
+    MethodYield[] expected = new MethodYield[] {
+      buildMethodYield(0, null, (Constraint) null),
```
m-g-sonar: As far as i can tell you only use this method with a varargs being a single null constraint. So remove it from the parameter and hardcode the array in `buildMethodYield`.

java-frontend/src/test/java/org/sonar/java/se/MethodYieldTest.java
```diff
@@ -245,8 +250,19 @@ public void catch_class_cast_exception() throws Exception {
...
+      buildMethodYield(-1, new ObjectConstraint(true, false, null), (Constraint) null)};
```
m-g-sonar: Replace this constructor call by `ObjectConstraint.nullConstraint()`.

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
m-g-sonar: Ok, I'll agree on that, we don't need ordered yields, nor ordered endOfExecutionpath. So let's keep it that way.

PR #1126, opened 2016-12-04
SONARJAVA-1798 - Rule S3688: Disallowed classes should not be used

fergarrui:
* Added a parameterized rule to ban the usage of a class or package.


java-checks/src/test/files/checks/DisallowedClassCheck.java
```diff
@@ -0,0 +1,19 @@
...
+  String method1() { // Noncompliant [[sc=3;ec=9]]
+  }
+void method(String param) { // Noncompliant [[sc=13;ec=19]]
+  B var = new B(); // Compliant, no subtypes checked
+  String str = new String(); // Noncompliant [[sc=3;ec=9]]
+  System.out.println(var);
+  var.doSomething(new Integer(1));
+}
+}
+class B extends String { // Compliant, can extend
```
m-g-sonar: To me this case should also be excluded. An extension should be considered as an usage as well.
If it's making too much noise, we will remove the case in a second time, but I would go for also banning extension in the first implementation of the rule.

java-checks/src/test/files/checks/DisallowedClassCheckRegex.java
```diff
@@ -0,0 +1,19 @@
...
+  private Integer disallowed2; // Noncompliant [[sc=11;ec=18]]
+  private java.util.Date date;
+  String method1() { // Noncompliant [[sc=3;ec=9]]
+  }
+void method(String param) { // Noncompliant [[sc=13;ec=19]]
+  B var = new B(); // Compliant, no subtypes checked
+  String str = new String(); // Noncompliant [[sc=3;ec=9]]
+  System.out.println(var);
+  var.doSomething(new Integer(1)); // Noncompliant  [[sc=19;ec=33]]
+}
```
m-g-sonar: You may want to add a case where we are not able to resolve the type (for example a method returning an Unknown type, not defined anywhere or not imported)

java-checks/src/main/java/org/sonar/java/checks/DisallowedClassCheck.java
```diff
@@ -0,0 +1,89 @@
...
+import org.sonar.plugins.java.api.JavaFileScanner;
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.NewClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = DisallowedClassCheck.KEY)
+public class DisallowedClassCheck extends BaseTreeVisitor implements JavaFileScanner {
```
fergarrui: I saw that `Template Rule: Yes`, but I had no idea about that annotation, thanks! I will add it.

java-checks/src/test/files/checks/DisallowedClassCheck.java
fergarrui: I excluded it just in case a developer wants to forbid a certain class, for example java.util.Date but not a subclass of it that he implemented to make it thread safe.
Anyway I will add that exclusion :)

fergarrui:
Hi,
Made the requested changes, please have a look to my comments just in case you want me to revert anything.

By the way, I saw that in the rule specification, it is mentioned that the rule should run both in Java code and Test cases. I thought that could be done by adding it to the test checklist, but seems that I was wrong, because doing that makes the build fail (defining a rule twice), how can I do that?

I also rebased my branch but still showing the conflict.

Thanks!

m-g-sonar:
Great! Thanks for the rebase/update.
You could have let the key of the rule as a constant, I forgot about its usage in the log while reviewing. But this current solutions works fine as well. I'll deal with the conflict when merging.

About the rule targeting both main and test sources, I'm gonna look at what we can do.

m-g-sonar:
Okay, so for the time being, let's only focus on main source files, so your implementation is already complete.

Today, a rule is targeting either main or test files, but not both. Regarding rule templates, their is no clear view if we are going to continue supporting them in the future. Consequently, I would not put too much energy on the subject.

To answer your question about how to target both sources types, a correct approach would then be to have a sister rule targeting only test files, sharing parts of the implementation... But let's drop that part of the spec for the moment.

Cheers,
Michael

java-checks/src/main/java/org/sonar/java/checks/CheckList.java
```diff
@@ -451,6 +451,7 @@ private CheckList() {
       .add(FilesExistsJDK8Check.class)
       .add(StaticImportCountCheck.class)
       .add(ClassFieldCountCheck.class)
+      .add(DisallowedClassCheck.class)
```
m-g-sonar: I'm not sure you rebased correctly your branch. Are you sure the master branch you are looking at is up to date?

The conflict comes from this line. In our master branch, there is 4 more check after `ClassFieldCountCheck.class`.
See here: [CheckList.java#L457](https://github.com/SonarSource/sonar-java/blob/4.3-RC1/java-checks/src/main/java/org/sonar/java/checks/CheckList.java#L457)

m-g-sonar:
@fergarrui can you just check the rebase of this PR? From what I have seen, our master is a few commit ahead of your PR, and new rules have been added, causing the conflict.

m-g-sonar:
Hey @fergarrui. I rebased your branch on PR #1131 and fixed the conflict. I also squash-merged the other branch into master. Resulting commit on master: https://github.com/SonarSource/sonar-java/commit/aba360243ac5c37ff313a09b2424e0939231aca4

As it is now merged, I close this PR. Thanks again for your work!

++
Michael

PR #1129, opened 2016-12-06
SONARJAVA-1984 Fix reading of array types in bytecode to include dimensions

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/matcher/MethodMatcherTest.java
```diff
@@ -222,6 +222,22 @@ public void detected() {
     assertThat(matches.get(objectToStringWithAnyParam)).containsExactly(6, 10, 11, 14);
     assertThat(matches.get(integerToString)).containsExactly(19);
     assertThat(matches.get(callSiteIsTest)).containsExactly(6, 10, 11, 14, 18, 22);
+
+  }
+
+  @Test
+  public void arrays_in_bytecode() {
+    MethodMatcher arrayInByteCode = MethodMatcher.create()
+      .typeDefinition("javax.swing.table.DefaultTableModel")
```
benzonico: I would not rely on external method defined in JDK which might change or be deprecated at some point. This test is indirect and the case is tested directly : so either remove this test or rely on something we control. Moreover given that you have ruling raising a new issue I think it should be removed.

PR #1133, opened 2016-12-09
SONARJAVA-1992 Fix S2309 to raise issue on file instead of last line and check for empty issue locations

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/AnalyzerMessage.java
```diff
@@ -118,12 +132,14 @@ public String toString() {
...
 
   private static AnalyzerMessage.TextSpan textSpanBetween(SyntaxToken firstSyntaxToken, SyntaxToken lastSyntaxToken) {
-    return new AnalyzerMessage.TextSpan(
+    AnalyzerMessage.TextSpan location = new AnalyzerMessage.TextSpan(
       firstSyntaxToken.line(),
       firstSyntaxToken.column(),
       lastSyntaxToken.line(),
       lastSyntaxToken.column() + lastSyntaxToken.text().length()
     );
+    Preconditions.checkState(!location.isEmpty(), "Issue location should not be empty");
```
benzonico: I tend to think that the message should be a bit more helpful than that... if it blows, message is not really helpful.

m-g-sonar: @benzonico : I tried different things but I was not convinced. I agree that we should be more helpful at this point. What do you think of : 
```
Preconditions.checkState(!location.isEmpty(), "Invalid issue location: Issue location is empty when trying reporting on "
      + "(" + firstSyntaxToken.line() + "," + firstSyntaxToken.column() +").");
```

benzonico: You have the check, might worth to also report it ?

PR #1140, opened 2016-12-15
SONARJAVA-1998 NPE issue location is also item of the flow

saberduck:
* Refactored Expectations.Parser to allow more robust parsing of comments in test
 * added messages to the flows for NPE check

### DIFF
--- a/java-frontend/src/test/files/se/NPE_reporting.java
+++ b/java-frontend/src/test/files/se/NPE_reporting.java
@@ -1,21 +1,39 @@
+package javax.annotation;
+
+@interface CheckForNull {}
+
+@interface Nonnull {}
+
+@interface Nullable {}
+
 class A {
   void null_assigned(Object b, Object c) {
     if(b == null);
     if(c == null);
     Object a = null; // flow@fl1 {{...}}
-    a.toString(); // Noncompliant [[flows=fl1]]
+    a.toString(); // Noncompliant [[flows=fl1]] flow@fl1 {{a is dereferenced}}
+  }
+
+  private Object getA() {
+    return null;
+  }
+
+  void null_assigned(Object b, Object c) {
+    if(b == null);
+    if(c == null);
+    getA().toString(); // Noncompliant [[flows=mnull]] flow@mnull {{Result of getA() is dereferenced}} flow@mnull {{...}}
   }
 
   void reassignement() {
     Object a = null; // flow@reass {{...}}
     Object b = new Object();
     b = a; // flow@reass {{...}}
-    b.toString(); // Noncompliant [[flows=reass]]
+    b.toString(); // Noncompliant [[flows=reass]] flow@reass {{b is dereferenced}}
   }
 
   void relationshipLearning(Object a) {
     if (a == null) { // flow@rela {{...}}
-      a.toString(); // Noncompliant [[flows=rela]]
+      a.toString(); // Noncompliant [[flows=rela]] flow@rela {{a is dereferenced}}
     }
   }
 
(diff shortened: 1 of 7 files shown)


java-frontend/src/test/java/org/sonar/java/se/Expectations.java
```diff
@@ -106,43 +118,15 @@
...
-      final int offset;
-
-      RelativeLineRef(int offset) {
-        this.offset = offset;
-      }
-
-      @Override int getLine(int ref) {
-        return ref + offset;
-      }
+  private static class IssueImpl extends EnumMap<IssueAttribute, Object> implements Issue {
```
benzonico: Not sure to see the point of having Issue and IssueImpl here... Why Issue can't extends EnumMap and have the create static method ? So no need for IssueImpl at all.

java-frontend/src/test/java/org/sonar/java/se/Expectations.java
```diff
@@ -235,109 +230,191 @@ public void visitTrivia(SyntaxTrivia syntaxTrivia) {
...
     }
 
-    private void collectExpectedIssues(String comment, int line) {
-      if (comment.startsWith(NONCOMPLIANT_COMMENT)) {
-        parseIssue(comment, line);
+    @VisibleForTesting
+    void collectExpectedIssues(String comment, int line) {
+      if (NONCOMPLIANT_COMMENT.matcher(comment).find()) {
+        IssueAndFlows issueAndFlows = parseIssue(comment, line);
+        if (issueAndFlows.parsed()) {
```
benzonico: Given : 
- how the constant `IssueAndFlow.NULL` is declared and used
- IssueAndFlow class is not implementing equals and hashcode

I don't see how this condition can be false and thus I don't see how it is tested.
I would go for the suggestion to remove it completely.

java-frontend/src/test/java/org/sonar/java/se/Expectations.java
```diff
@@ -235,109 +230,191 @@ public void visitTrivia(SyntaxTrivia syntaxTrivia) {
...
       attributes.put(MESSAGE, message);
-      Arrays.stream(flowIds)
-        .map(id -> new FlowComment(id, line, attributes))
-        .forEach(f -> flows.put(f.id, f));
+      return ids.stream().map(id -> new FlowComment(id, line, attributes));
     }
 
-    private void parseIssue(String comment, int line) {
-      String cleanedComment = StringUtils.remove(comment, NONCOMPLIANT_COMMENT);
+    static class IssueAndFlows {
```
benzonico: Dodgy naming : ParsedComment ?

java-frontend/src/test/files/se/NPE_reporting.java
```diff
@@ -58,4 +76,16 @@ private Object foo(boolean a) {
...
+
+  class A {}
+
+  public void testMemberSelect(A a1, @CheckForNull A a2, @Nullable A a3) {
+    a1.hashCode(); // No issue
+    a2.hashCode(); // Noncompliant [[flows=a2]] {{NullPointerException might be thrown as 'a2' is nullable here}} flow@a2 {{a2 is dereferenced}}
+    a3.hashCode(); // Noncompliant [[flows=a3]] {{NullPointerException might be thrown as 'a3' is nullable here}} flow@a3 {{a3 is dereferenced}}
+
+    a1.a.hashCode(); // No issue
+    a1.b.hashCode(); // False negative
```
saberduck: This was copy&pasted from `java-frontend/src/test/files/se/NullDereferenceCheck.java` , because there were some issues with the flow in this. You're right, that this is not needed in this test. I am not sure of the original purpose.

java-frontend/src/test/java/org/sonar/java/se/Expectations.java
saberduck: Indeed, `parseIssue` always succeeds because we have at least the line. Thanks for catching this.

PR #1141, opened 2016-12-15
SONARJAVA-1870 Fix false negative of fully qualified name usage

benzonico:


java-checks/src/main/java/org/sonar/java/checks/SimpleClassNameCheck.java
```diff
@@ -20,73 +20,66 @@
...
-    if (importTree.qualifiedIdentifier().is(Kind.MEMBER_SELECT)) {
-
-      IdentifierTree identifier = ((MemberSelectExpressionTree) importTree.qualifiedIdentifier()).identifier();
-      if ("*".equals(identifier.name())) {
-        fileContainsStarImport = true;
-        return;
+    CompilationUnitTree cut = (CompilationUnitTree) tree;
+    cut.types().stream().filter(NOT_EMPTY_STATEMENT).map(t -> ((ClassTree) t).symbol()).forEach(this::checkSymbol);
+    List<ImportTree> imports = cut.imports().stream().filter(NOT_EMPTY_STATEMENT).map(t -> (ImportTree) t).collect(Collectors.toList());
+    boolean fileContainsStarImport = false;
```
saberduck: Is there a reason not to use streams? I would prefer to use one style or the other, but not mixing them

java-checks/src/main/java/org/sonar/java/checks/SimpleClassNameCheck.java
```diff
@@ -56,16 +56,10 @@ public void visitNode(Tree tree) {
     CompilationUnitTree cut = (CompilationUnitTree) tree;
     cut.types().stream().filter(NOT_EMPTY_STATEMENT).map(t -> ((ClassTree) t).symbol()).forEach(this::checkSymbol);
     List<ImportTree> imports = cut.imports().stream().filter(NOT_EMPTY_STATEMENT).map(t -> (ImportTree) t).collect(Collectors.toList());
```
saberduck: It's unnecessary to collect here , and stream again few lines below. The same goes for `checkImports` argument. Just pass the `Stream` around

benzonico: Not possible in a simple manner : anyMatch and forEach are terminal operations.

PR #1146, opened 2016-12-16
SONARJAVA-1999 flow reporting DivByZero

saberduck:


saberduck:
Look at the last commit. I will rebase later.

java-frontend/src/main/java/org/sonar/java/se/checks/Flows.java
```diff
@@ -0,0 +1,37 @@
...
+import org.sonar.plugins.java.api.tree.Tree;
+
+import java.util.List;
+import java.util.Set;
+
+class Flows {
+
+  private Flows() {}
+
+  static Set<List<JavaFileScannerContext.Location>> singleton(String msg, Tree tree) {
```
benzonico: This should be part of FlowComputation class and properly unit tested. No need to introduce another class here.

java-frontend/src/main/java/org/sonar/java/se/checks/DivisionByZeroCheck.java
```diff
@@ -211,8 +211,18 @@ public SymbolicValue wrappedValue() {
...
+      String flowMessage;
+      if (expression.is(Tree.Kind.IDENTIFIER)) {
+        String name = ((IdentifierTree) expression).name();
+        expressionName = "'" + name + "'";
+        flowMessage = name + " is divided by zero";
+      } else {
+        expressionName = "this expression";
+        flowMessage = "this expression contains division by zero";
+      }
+      context.reportIssue(expression, DivisionByZeroCheck.this, String.format("Make sure %s can't be zero before doing this %s.", expressionName, operation),
```
benzonico: Please use concatenation here.  No benefit from format.

PR #1148, opened 2016-12-17
SONARJAVA-2013 - FP S1488 when the variable is used to hold @SuppressWarnings

fergarrui:
Simple fix to remove the FP when the variable has to be declared to be annotated with @SuppressWarnings.


benzonico:
Thanks for the contribution @fergarrui : I handled this issue differently by not relying on the semantic (and tbh I implemented it without realizing there was this PR). Fix was merged with https://github.com/SonarSource/sonar-java/pull/1151 

As a consequence I am closing this PR.

PR #1153, opened 2016-12-19
SONARJAVA-2012 Update rule description

benzonico:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S2737_java.html
```diff
@@ -18,6 +18,23 @@
   return sb.toString();
 }
 </pre>
+<h2>Compliant Solution</h2>
```
m-g-sonar: There is something fishy with this modification. This is C# code, and it does not correspond to the correct solution from java [RSPEC-2737](https://jira.sonarsource.com/browse/RSPEC-2737). I guess it's because the java solution is labelled as being "Compliant Code Example", instead of "Compliant Solution".

PR #1158, opened 2016-12-21
SONARJAVA-2029 Flows are not reported in correct order

saberduck:


java-frontend/src/test/java/org/sonar/java/se/JavaCheckVerifierTest.java
```diff
@@ -331,11 +337,11 @@ public void verify_flow_locations() {
       .issueWithFlow(20)
         .flow(17, "msg", 19, null)
         .add();
-    try {
-      JavaCheckVerifier.verify("src/test/files/JavaCheckVerifierFlows.java", fakeVisitor);
```
saberduck: There is a missing assert to make this fail when no issue is detected, that's why I reworked it to the new AssertJ style

java-frontend/src/test/java/org/sonar/java/se/JavaCheckVerifier.java
```diff
@@ -207,6 +205,7 @@ private void assertMultipleIssue(Set<AnalyzerMessage> issues) throws AssertionEr
     Preconditions.checkState(!issues.isEmpty(), "At least one issue expected");
     List<Integer> unexpectedLines = Lists.newLinkedList();
     Multimap<Integer, Expectations.Issue> expected = expectations.issues;
+    expectations.reverseFlows(); // FIXME temporary workaround
```
m-g-sonar: Do you have a ticket for that? If not, create one and add a reference in the code. So we will not forget about it.

saberduck: rewrote comment to remove fixme

PR #1163, opened 2016-12-22
SONARJAVA-2022 Fix S2187 to handle 'cucumber' runner

m-g-sonar:


java-checks/src/test/files/checks/NoTestInTestClassCheck.java
```diff
@@ -87,7 +88,13 @@ public void test2() {
 public class MyNewTest { // should not raise an issue
 }
 @org.junit.runner.RunWith(Enclosed.class)
-public class MyNewTest2 { // no issue
+public class MyNew2Test { // no issue
+}
+@org.junit.runner.RunWith(cucumber.api.junit.Cucumber.class)
```
vilchik-elena: @m-g-sonar Not sure it's relevant but what about testing `@RunWith...` (without full name)?

java-checks/src/test/files/checks/NoTestInTestClassCheck.java
```diff
@@ -87,7 +88,13 @@ public void test2() {
...
+public class MyCucumberTest { // should not raise an issue
+}
+@org.junit.runner.RunWith(Cucumber.class)
```
vilchik-elena: @m-g-sonar You don't test unknown runners

java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
```diff
@@ -99,16 +100,19 @@ private void checkJunit3TestClass(IdentifierTree className, Iterable<Symbol> mem
   }
 
   private void checkJunit4AndAboveTestClass(IdentifierTree className, JavaSymbol.TypeJavaSymbol symbol, Iterable<Symbol> members) {
-    if (symbol.name().endsWith("Test") && !runWithEnclosedRunner(symbol)) {
+    if (symbol.name().endsWith("Test") && !runWithRunner(symbol)) {
```
vilchik-elena: @m-g-sonar `runWithRunner` this name is a bit funny :)

java-checks/src/test/files/checks/NoTestInTestClassCheck.java
m-g-sonar: Indeed @vilchik-elena we are currently specifically targeting `Cucumber.class` and `Enclosed.class`

PR #1164, opened 2016-12-26
SONARJAVA-2025 - S3400 'Methods should not return constants' should n…

fergarrui:
…ot be applied to default methods

* The rule is not raising issues on 'default' interface methods.


java-checks/src/main/java/org/sonar/java/checks/ConstantMethodCheck.java
```diff
@@ -62,6 +65,15 @@ public void visitNode(Tree tree) {
     }
   }
 
+  private static boolean hasDefaultKeyword(MethodTreeImpl methodTree) {
+    Symbol.MethodSymbol symbol = methodTree.symbol();
```
m-g-sonar: There is a way simpler approach to check if it's a default method (see here: [UnusedMethodParameterCheck.java#L119](https://github.com/SonarSource/sonar-java/blob/4.4.0.8066/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java#L119)):
```
ModifiersUtils.hasModifier(modifiers, Modifier.DEFAULT)
```

fergarrui: Alright, was trying to find some method like that but I couldn't, did the change, thanks!

m-g-sonar:
Thanks for the update @fergarrui ! Merged with master!

PR #1167, opened 2017-01-03
SONARJAVA-1850 FN on S2110 when using static imports for date parts name

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/InvalidDateValuesCheck.java
```diff
@@ -119,12 +119,16 @@ private static String getName(MethodInvocationTree mit, String name) {
...
-      MemberSelectExpressionTree mse = (MemberSelectExpressionTree) argument;
-      Symbol reference = mse.identifier().symbol();
-      if (reference.owner().type().is(JAVA_UTIL_CALENDAR) && Threshold.getThreshold(reference.name()) != null) {
-        return reference.name();
-      }
+      reference = ((MemberSelectExpressionTree) argument).identifier().symbol();
+    } else if (argument.is(Tree.Kind.IDENTIFIER)) {
+      reference = ((IdentifierTree) argument).symbol();
+    }
+    if (reference != null &&
```
benzonico: The condition where reference is null won't be tested

alban-auzeill: I've added a line in the unit test with a not supported INT_LITERAL to ensure that reference is tested with null.

PR #1168, opened 2017-01-04
SONARJAVA-1926: Enforce that rule with constant remediation function …

saberduck:
- use Gson to parse rule's metadata
- check when issue with cost is raised and rule has constant cost

### DIFF
--- a/java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/CheckVerifier.java
+++ b/java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/CheckVerifier.java
@@ -28,7 +28,7 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multiset;
-import com.google.common.io.Resources;
+import com.google.gson.Gson;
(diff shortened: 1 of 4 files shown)


java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/CheckVerifier.java
```diff
@@ -182,36 +181,71 @@ private void assertMultipleIssue(Set<AnalyzerMessage> issues) throws AssertionEr
...
+        case "Linear":
+          return RemediationFunction.LINEAR;
+        case "Constant/Issue":
+          return RemediationFunction.CONST;
+        default:
+          return null;
+      }
+    } catch (IOException e) {
+      // Failed to open json file, as this is not part of API yet, we should not fail because of this
+      return null;
```
benzonico: I think it should rethrow with a correct message rather than defaulting to some behavior. Better to fail when something unexpected happens (and test that).

saberduck: So is this a part of an API now? I kept the current behavior because of the comment, but if the comment is not valid anymore, I agree.

PR #1169, opened 2017-01-04
SONARJAVA-1971 IndentationCheck with lambda as arg of method on separate line

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/IndentationCheck.java
```diff
@@ -131,14 +131,17 @@ public void visitMethodInvocation(MethodInvocationTree tree) {
...
   public void visitLambdaExpression(LambdaExpressionTree lambdaExpressionTree) {
-    int previousTokenLine = getPreviousToken(lambdaExpressionTree).line();
-    int lambdaFirstTokenLine = lambdaExpressionTree.firstToken().line();
-    if (previousTokenLine != lambdaFirstTokenLine) {
-      expectedLevel += indentationLevel;
-    }
-    super.visitLambdaExpression(lambdaExpressionTree);
-    if (previousTokenLine != lambdaFirstTokenLine) {
-      expectedLevel -= indentationLevel;
+    // scan lambdaExpressionTree.parameters() is useless
```
alban-auzeill: Because parameter of lambda can not raise issue, it's just: type, identifier and no initializer. Indentation is checked only for member of "class", "block", "switch". A parameter of lambda can not contains one of those. Should I change the comment?

benzonico: Yes, that was actually the point : the comment is useless if you don't explain the why.

PR #1172, opened 2017-01-05
SONARJAVA-2032: UnclosedResourcesCheck FP on try-with-resources

saberduck:


java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -166,8 +166,13 @@ private boolean isOpeningResource(NewClassTree syntaxNode) {
   }
 
   private static boolean isWithinTryHeader(Tree syntaxNode) {
-    final Tree parent = syntaxNode.parent();
-    if (parent.is(Tree.Kind.VARIABLE)) {
+    Tree parent = syntaxNode;
+    while ((parent = parent.parent()) != null) {
```
benzonico: Please don't do assignement and comparison within the same expression. 

You can also avoid the break by testing nullness and kind in while condition and doing the assignement in the body of the while.

PR #1174, opened 2017-01-10
SONARJAVA-2036 FP when statement is initialized indirectly

benzonico:


java-checks/src/main/java/org/sonar/java/checks/PreparedStatementAndResultSetCheck.java
```diff
@@ -102,7 +103,17 @@ private static Integer getPreparedStatementNumberOfParameters(@Nullable Expressi
     if (tree != null && tree.is(Tree.Kind.METHOD_INVOCATION)) {
       Arguments arguments = ((MethodInvocationTree) tree).arguments();
       if (!arguments.isEmpty()) {
-        return getNumberQuery(arguments.get(0));
+        ExpressionTree methodSelect = ((MethodInvocationTree) tree).methodSelect();
```
saberduck: Why not use `MethodMatcher` for this?

PR #1159, opened 2016-12-21
SONARJAVA-1995 Report exceptional paths in invoked methods

m-g-sonar:

PR #1189, opened 2017-01-18
SONARJAVA-1995 Selected yields on edges of exploded graph for exceptional paths

m-g-sonar:


m-g-sonar:
Rework of https://github.com/SonarSource/sonar-java/pull/1159

java-frontend/src/main/java/org/sonar/java/se/MethodYield.java
```diff
@@ -120,7 +121,7 @@ public String toString() {
     if (resultConstraint != null) {
       stateStream = stateStream.map(s -> s.addConstraint(sv, resultConstraint));
     }
-    return stateStream.distinct();
+    return stateStream.distinct().map(ps -> new Pair<>(this, ps));
```
benzonico: This is a rather complicated hack to try to keep using stream api against all odds in EGWalker. I would try to get rid of Pair class and do a proper old loop.

java-frontend/src/main/java/org/sonar/java/se/checks/FlowComputation.java
```diff
@@ -117,14 +122,51 @@ private void run(@Nullable final ExplodedGraph.Node currentNode, @Nullable final
     final ExplodedGraph.Node parent = currentNode.parent();
     learnedConstraints.stream()
       .filter(addToFlow.and(Objects::nonNull))
-      .forEach(lc -> flow.add(location(parent, learnedConstraintMessage(lc, parent))));
+      .forEach(lc -> {
+        flow.add(location(parent, learnedConstraintMessage(lc, currentNode, parent)));
+        if (parent.programPoint.syntaxTree().is(Tree.Kind.METHOD_INVOCATION)) {
+          List<MethodYield> usedMethodYields = currentNode.usedMethodYields(parent);
+          if (!usedMethodYields.isEmpty()) {
```
benzonico: I tend to think that this check should not be required if flowFromMethodInvocation return an empty list.

java-frontend/src/main/java/org/sonar/java/se/checks/FlowComputation.java
```diff
@@ -117,14 +122,51 @@ private void run(@Nullable final ExplodedGraph.Node currentNode, @Nullable final
...
+  }
+
+  private static int correspondingArgumentIndex(SymbolicValue candidate, ExplodedGraph.Node invocationNode) {
+    MethodInvocationTree mit = (MethodInvocationTree) invocationNode.programPoint.syntaxTree();
+    List<SymbolicValue> arguments = argumentsUsedForMethodInvocation(invocationNode, mit);
+    return arguments.indexOf(candidate);
+  }
+
+  private static List<SymbolicValue> argumentsUsedForMethodInvocation(ExplodedGraph.Node invocationNode, MethodInvocationTree mit) {
+    List<SymbolicValue> values = new ArrayList<>(invocationNode.programState.peekValues(mit.arguments().size()));
```
benzonico: This is not properly tested : you only have invocations with one argument in your tests. AFAIK (pblm raised by tibor) there might be an issue with indexes when you are in the case of exceptions (because in execptional path return value is on the stack whereas otherwise it is not). So I would add more complex cases to be 100% sure things are working properly and/or document if it works only on the context of exceptional paths.

java-frontend/src/main/java/org/sonar/java/se/checks/FlowComputation.java
```diff
@@ -117,14 +122,51 @@ private void run(@Nullable final ExplodedGraph.Node currentNode, @Nullable final
...
+  private static String learnedConstraintMessage(Constraint lc, ExplodedGraph.Node currentNode, ExplodedGraph.Node parent) {
+    Tree nodeTree = parent.programPoint.syntaxTree();
     String name = SyntaxTreeNameFinder.getName(nodeTree);
     if (nodeTree.is(Tree.Kind.METHOD_INVOCATION)) {
+      List<MethodYield> usedMethodYields = currentNode.usedMethodYields(parent);
+      if (!usedMethodYields.isEmpty()) {
+        // TODO use other yields as alternative flows ?
+        Type exceptionType = usedMethodYields.get(0).exceptionType();
+        // Only considering exceptional flows
+        if (exceptionType != null) {
```
benzonico: Wondering about this limitation in fact. Maybe to be addressed in another ticket.

java-frontend/src/main/java/org/sonar/java/se/checks/FlowComputation.java
m-g-sonar: Not yet taken into account.

java-frontend/src/main/java/org/sonar/java/se/MethodYield.java
```diff
@@ -57,6 +63,15 @@ public MethodYield(int arity, boolean varArgs) {
     this.exceptionType = null;
   }
 
+  private static List<List<JavaFileScannerContext.Location>> initFlows(int arity) {
+    // n parameters + 1 for the return value
+    List<List<JavaFileScannerContext.Location>> result = new ArrayList<>(arity + 1);
+    for (int i = 0; i < arity + 1; i++) {
+      result.add(0, ImmutableList.of());
```
benzonico: not a big deal but as the list is empty anyway, no real need to add at the beginning. Moreover I think you could express this method in a one liner with IntStream.range.mapToObj.collect(toList)

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraph.java
```diff
@@ -113,7 +117,7 @@ public Tree syntaxTree() {
     public final ProgramPoint programPoint;
     @Nullable
     public final ProgramState programState;
-    private final List<Node> parents;
+    private final Map<Node, MethodYield> parents;
```
benzonico: This implies one property : A node can only be reached from a parent via a unique method yield. I am not really sure this is true.

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraph.java
```diff
@@ -154,17 +158,17 @@ private void addConstraint(SymbolicValue sv, @Nullable Constraint constraint) {
...
-    public void addParent(Node node) {
-      parents.add(node);
+    public void addParent(Node node, @Nullable MethodYield methodYield) {
+      parents.put(node, methodYield);
     }
 
     @Nullable
     public Node parent() {
-      return parents.isEmpty() ? null : parents.get(0);
+      return parents.isEmpty() ? null : Lists.newArrayList(parents.keySet()).get(0);
```
benzonico: getParents().get(0) would be better here.

java-frontend/src/test/files/se/XProcExceptionalBranchingReporting.java
```diff
@@ -0,0 +1,57 @@
...
+
+  void tst3(Object o) {
+    if (o != null) {
+      Object o2 = returnNullIfParamNotNull(o); // flow@npe3 {{'returnNullIfParamNotNull' returns null}} flow@npe3 {{'o2' is assigned null}}
+      o2.toString(); // Noncompliant [[flows=npe3]] {{NullPointerException might be thrown as 'o2' is nullable here}} flow@npe3 {{'o2' is dereferenced}}
+    }
+  }
+}
+
+class MyException1 extends Exception {}
```
benzonico: missing newline

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraph.java
```diff
@@ -243,5 +247,10 @@ public boolean equals(Object obj) {
     public String toString() {
       return "B" + programPoint.block.id() + "." + programPoint.i + ": " + programState;
     }
+
+    @CheckForNull
+    public MethodYield usedMethodYield(Node from) {
```
benzonico: I tend to prefer "selected" over "used" in this context.

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -993,7 +999,15 @@ public void enqueue(ExplodedGraph.ProgramPoint programPoint, ProgramState progra
...
 
+  public void enqueue(ExplodedGraph.ProgramPoint programPoint, ProgramState programState, @Nullable MethodYield methodYield) {
+    enqueue(programPoint, programState, false, methodYield);
+  }
+
   public void enqueue(ExplodedGraph.ProgramPoint newProgramPoint, ProgramState programState, boolean exitPath) {
+    enqueue(newProgramPoint, programState, exitPath, null);
+  }
+
+  public void enqueue(ExplodedGraph.ProgramPoint newProgramPoint, ProgramState programState, boolean exitPath, @Nullable MethodYield methodYield) {
```
benzonico: Thinking outloud (in writing) more than actually a request for change : 
I am wondering if it is a good thing to have this methodYield parameter added to this method as this is kinda corner case compared to all the other cases... not sure what would be an alternative but addition of this parameter is bothering me a bit (more a gut feeling than a real thing here)

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraph.java
m-g-sonar: I would correct it by saying that:
> A node can only be reached from a parent via **at most** one method yield.
MethodYields are optionals.

Now, I didn't noticed any cases where it was not the case, because of uniqueness of yields, which always imply learning something different. I didn't enforced that property with a preconditions, but it maybe worth it. WDYT?

m-g-sonar: Ok, after further investigation, there is still some cases where multiple yields can conduct to the same node, when considering variadic methods.

java-frontend/src/main/java/org/sonar/java/se/checks/FlowComputation.java
m-g-sonar: modified tests in order to introduce more parameters.

PR #1191, opened 2017-01-23
SONARJAVA-1537 Support assert statement in CFG and SE

benzonico:


java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -561,13 +562,27 @@ private void visit(Tree tree, @Nullable Tree terminator) {
...
     checkerDispatcher.executeCheckPostStatement(tree);
     clearStack(tree);
   }
 
+  private void executeAssertStatement(Tree tree) {
+    // After an assert statement we know that the evaluated expression is true.
+    ProgramState.Pop pop = programState.unstackValue(1);
+    pop.values.forEach(v -> v.setConstraint(pop.state, BooleanConstraint.TRUE)
+      .forEach(ps -> {
+        checkerDispatcher.syntaxNode = tree;
```
saberduck: not directly related to your change, but why `org.sonar.java.se.CheckerDispatcher#syntaxNode` is a field instead of just being a parameter of `org.sonar.java.se.CheckerDispatcher#addTransition`  ?

benzonico: If it would be then it would be propagated through method calls within the class => field

PR #1195, opened 2017-01-24
SONARJAVA-2066 hardcode Objects.requireNonNull behavior

benzonico:


java-frontend/src/main/java/org/sonar/java/se/SymbolicExecutionVisitor.java
```diff
@@ -60,20 +61,61 @@ public void visitNode(Tree tree) {
...
+          if (declaration != null) {
+            SymbolicExecutionVisitor.this.execute(declaration);
+          }
+        }
+      }
+      return behaviors.get(symbol);
+    }
+
+    private boolean isObjectsRequireNonNullMethod(Symbol symbol) {
+      return symbol.owner().type().is("java.util.Objects") && "requireNonNull".equals(symbol.name());
```
benzonico: Because this requires to refactor methodMatcher to match on symbols which is not available now. Let's implement a bit more of required methods before going for change on MethodMatcher

java-frontend/src/main/java/org/sonar/java/se/SymbolicExecutionVisitor.java
```diff
@@ -60,20 +61,61 @@ public void visitNode(Tree tree) {
...
+      happyYield.exception = false;
+      happyYield.parametersConstraints[0] = ObjectConstraint.notNull();
+      happyYield.resultIndex = 0;
+      happyYield.resultConstraint = happyYield.parametersConstraints[0];
+      behavior.addYield(happyYield);
+
+      MethodYield exceptionalYield = new MethodYield(symbol.parameterTypes().size(), false);
+      exceptionalYield.exception = true;
+      exceptionalYield.parametersConstraints[0] = ObjectConstraint.nullConstraint();
+      behavior.addYield(exceptionalYield);
```
benzonico: As discussed : for the time being we are not even sure we actually want to provide the exceptional yield for this case (to limit noise and/or FP).

PR #1198, opened 2017-01-25
SONARJAVA-2026: Flow for local variables should be reported on initia…

saberduck:
…lization not declaration

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
@@ -740,19 +740,24 @@ public class ExplodedGraphWalker {
     return !symbol.isUnknown() && symbol.metadata().isAnnotatedWith("javax.annotation.Nonnull");
   }
 
+  /**
+   * We do nothing for local variables without initializers, for other variables we use default value or initializer if there is one
+   *
+   * See <a href="http://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.12.5">JLS 4.12.5</a> for details
+   */
   private void executeVariable(VariableTree variableTree, @Nullable Tree terminator) {
-    ExpressionTree initializer = variableTree.initializer();
-    if (initializer == null) {
+    if (variableTree.initializer() == null) {
       SymbolicValue sv = null;
       Type variableType = variableTree.symbol().type();
+      boolean isField = ProgramState.isField(variableTree.symbol());
       if (terminator != null && terminator.is(Tree.Kind.FOR_EACH_STATEMENT)) {
         sv = constraintManager.createSymbolicValue(variableTree);
-      } else if (variableType.is("boolean")) {
+      } else if (isField && variableType.is("boolean")) {
         sv = SymbolicValue.FALSE_LITERAL;
       } else if (variableTree.parent().is(Tree.Kind.CATCH)) {
         sv = getCaughtException(variableType);
         programState = programState.addConstraint(sv, ObjectConstraint.notNull());
-      } else if (!variableType.isPrimitive()) {
+      } else if (isField && variableType.isClass()) {
         sv = SymbolicValue.NULL_LITERAL;
       }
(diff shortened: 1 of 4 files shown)


java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -734,19 +734,24 @@ private static boolean isNonNullMethod(Symbol symbol) {
     return !symbol.isUnknown() && symbol.metadata().isAnnotatedWith("javax.annotation.Nonnull");
   }
 
+  /**
+   * We do nothing for local variables without initializers, for other variables we use default value or initializer if there is one
```
benzonico: I would use passive form here. "Nothing is done..." as you and me are not really doing the computation :wink:

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -734,19 +734,24 @@ private static boolean isNonNullMethod(Symbol symbol) {
...
+   *
+   * See <a href="http://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.12.5">JLS 4.12.5</a> for details
```
benzonico: really minor but in order to lighten the read we usually refer to JLS8 4.12.5 without the link...

java-frontend/src/test/files/se/SeEngineTest.java
```diff
@@ -109,3 +109,38 @@ boolean equals(Object obj) {
   }
 
 }
+
+class DefaultValues {
+
+  void localVars() {
+    boolean a; // this is basically noop
```
benzonico: I tend to disagree, this is not a noop : for me noop implies there is an operation while there is none here. It is a declaration. If I'm correct what you meant to say is more about the fact that this line has no impact on constraint on value. Really going for detail here I admit but I would drop that comment : useless and confusing.

java-frontend/src/test/files/se/NullDereferenceCheck.java
```diff
@@ -714,3 +714,11 @@ private static void bar(Object o, byte[] b) {
     FooBar.bar(null, new byte[10]); // Compliant
   }
 }
+
+class Params {
+
+  void foo(int a) {
```
benzonico: Not sure how this testcase actually relates to the ticket. And I am not even sure how this case actually demonstrate anything related to primitive : you can write the same test with Object instead of int and it will pass...

java-frontend/src/test/files/se/SeEngineTest.java
```diff
@@ -108,3 +109,38 @@ boolean equals(Object obj) {
...
+    Object b;
+    manyStatementsLater();
+    inTheGalaxyFarAway();
```
benzonico: Missing a "Far" https://sewforthandsewon.files.wordpress.com/2015/12/a-long-time-ago-sw.png?w=300&h=198 :wink:

java-frontend/src/test/files/se/SeEngineTest.java
```diff
@@ -31,6 +31,7 @@ void test_null_assignement() {
 
   void local_variable() {
     Object a;
+    a = null;
```
benzonico: wondering if this should be fixed or removed altogether given the testcase added.

saberduck: you're right, this was duplicated by the added test. I removed it

java-frontend/src/test/files/se/SeEngineTest.java
```diff
@@ -108,3 +104,38 @@ boolean equals(Object obj) {
...
 
 }
+
+class DefaultValues {
+
+  void localVars() {
+    boolean a;  // no SV is created on local variable declaration, value is undefined as per JLS
+    Object b;
+    manyStatementsLater();
+    inTheGalaxyFarFarAway();
```
benzonico: :+1:

PR #1202, opened 2017-01-26
SONARJAVA-2071 support guava Preconditions behavior

benzonico:


java-frontend/src/main/java/org/sonar/java/se/SymbolicExecutionVisitor.java
```diff
@@ -104,6 +107,11 @@ public MethodBehavior get(Symbol.MethodSymbol symbol) {
       return behaviors.get(symbol);
     }
 
+    private boolean isGuavaPrecondition(Symbol.MethodSymbol symbol) {
+      return symbol.owner().type().is("com.google.common.base.Preconditions")
+        && ("checkNotNull".equals(symbol.name()) || "checkArgument".equals(symbol.name()) || "checkState".equals(symbol.name()) );
```
m-g-sonar: symbol.name() called 3 times? maybe worth a dedicated variable.

PR #1207, opened 2017-01-27
Alow custom rules to externalize metadata

ushatil:
Move resource loading into protected methods, so custom rules can
override JavaRulesDefinition and externalize metadata.

Currently, if custom rule definitions try to extend JavaRulesDefinition, then
they can't externalize metadata the same way builtin rules do, because
that class only loads resources from its own library.

This small change would allow subclasses to load resources from their
own libraries.

Signed-off-by: Uri Shatil <uri.shatil@gmail.com>

### DIFF
--- a/sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
+++ b/sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
@@ -61,6 +61,28 @@ public class JavaRulesDefinition implements RulesDefinition {
     repository.done();
   }
 
+  /**
+   * Override this method to externalize rule metadata
+   *
+   * @param metadataKey the rule key for a given rule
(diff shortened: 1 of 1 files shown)


benzonico:
Hi @ushatil , 

Thanks for the contribution. 
The class you modified is in no way designed for extension for custom rules. This is not part of any API and the exemples of custom rules plugin don't rely on it : https://github.com/SonarSource/sonar-custom-rules-examples/blob/master/java-custom-rules/src/main/java/org/sonar/samples/java/MyJavaRulesDefinition.java but rather implement their own RulesDefinition interface. While I completely understand this is indeed quite some boiler plate code that could be simplified, this requires a bit more thinking and prioritizing on our side. 

As a consequence of this I am closing this PR. Feel free to contact us through sonarqube@googlegroups.com if you want to open a discussion for helping us to make progress on that topic.

Cheers,

PR #1208, opened 2017-01-30
Store EG Nodes in methodYield instead of precomputing flows

benzonico:


java-frontend/src/main/java/org/sonar/java/se/MethodYield.java
```diff
@@ -193,12 +194,11 @@ public boolean equals(Object obj) {
...
-  public void flowForParameter(int parameterIndex, List<JavaFileScannerContext.Location> flow) {
-    flowByParameters.set(parameterIndex + 1, flow);
-  }
-
   public List<JavaFileScannerContext.Location> flow(int parameterIndex) {
-    return flowByParameters.get(parameterIndex + 1);
+    if(node == null || behavior == null) {
+      return Lists.newArrayList();
+    }
+    return FlowComputation.flow(node, behavior.parameters().get(parameterIndex));
```
m-g-sonar: I expect this to fail badly if we try to track the return value of the method, as it's not part of the parameters.  (indexOutOfBound?) We need to get the adequate symbolic value in such case (No failing test?! :-1: for me)

PR #1224, opened 2017-02-06
SONARJAVA-2095: Avoid creating temporary lists when inspecting stack

saberduck:


java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -417,4 +422,9 @@ public SymbolicValue exitValue() {
   public boolean exitingOnRuntimeException() {
     return exitSymbolicValue instanceof SymbolicValue.ExceptionalSymbolicValue && ((SymbolicValue.ExceptionalSymbolicValue) exitSymbolicValue).exceptionType() == null;
   }
+
+  public boolean stackIsEmpty() {
```
m-g-sonar: I'm not convinced by the interest of this one method. You are basically creating it only to use it in one place. See other comment.

java-frontend/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -43,26 +43,25 @@
 
   @Override
   public ProgramState checkPreStatement(CheckerContext context, Tree syntaxNode) {
-    SymbolicValue currentVal = context.getState().peekValue();
-    if (currentVal == null) {
+    if (context.getState().stackIsEmpty()) {
```
m-g-sonar: Why not simply create a local method `stackIsEmpty(ProgramState state)` doing the null-check?
As it's the only place you need it, creating a method in ProgramState seems over-engineering to me.

java-frontend/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -43,26 +43,25 @@
...
     if (syntaxNode.is(Tree.Kind.METHOD_INVOCATION)) {
       MethodInvocationTree methodInvocation = (MethodInvocationTree) syntaxNode;
-      toCheck = methodInvocation.methodSelect();
-      int numberArguments = methodInvocation.arguments().size();
-      List<SymbolicValue> values = context.getState().peekValues(numberArguments + 1);
-      currentVal = values.get(numberArguments);
+      Tree methodSelect = methodInvocation.methodSelect();
+      if (methodSelect.is(Tree.Kind.MEMBER_SELECT)) {
+        SymbolicValue dereferencedSV = context.getState().peekValue(methodInvocation.arguments().size());
+        return checkConstraint(context, methodSelect, dereferencedSV);
```
m-g-sonar: nice improvement :+1:

java-frontend/src/main/java/org/sonar/java/collections/PStack.java
```diff
@@ -41,6 +43,14 @@
   E peek();
 
   /**
+   *
+   * @param i - index of element to be returned, 0 means top of the stack
+   * @return i-th element from top of the stack or null, if stack doesn't have so many elements
+   */
+  @CheckForNull
+  E peek(int i);
```
m-g-sonar: I would prefer to have the method throw an `IllegalStateException` if we try to access something out of the stack. When using this method, it means that you perfectly know what you are doing, trying to access a given element from the stack. If you can't I would prefer to have it failing fast.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
saberduck: I removed the method for now, but I think I will reintroduce it once we make also ProgramState.peek to throw an exception

java-frontend/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
saberduck: see first comment

PR #1226, opened 2017-02-07
SONARJAVA-2097 move ProgramPoint to top level

saberduck:


java-frontend/src/test/java/org/sonar/java/se/SECheckTest.java
```diff
@@ -38,8 +35,8 @@
   @Test(timeout = 3000)
   public void flow_from_exit_node_should_not_lead_to_infinite_recursion() throws Exception {
     CFG cfg = CFGTest.buildCFG("void foo(boolean a) { if(a) {foo(true);} foo(false); }");
-    ExplodedGraph.Node node = new ExplodedGraph.Node(new ProgramPoint(cfg.blocks().get(3), 0), mock(ProgramState.class));
-    node.addParent(new ExplodedGraph.Node(new ProgramPoint(cfg.blocks().get(2), 2), mock(ProgramState.class)), null);
+    ExplodedGraph.Node node = new ExplodedGraph.Node(new ProgramPoint(cfg.blocks().get(3)), mock(ProgramState.class));
+    node.addParent(new ExplodedGraph.Node(new ProgramPoint(cfg.blocks().get(2)), mock(ProgramState.class)), null);
```
m-g-sonar: This is not equivalent to what was written before, should be the 3rd element.

java-frontend/src/test/java/org/sonar/java/se/ProgramPointTest.java
```diff
@@ -30,9 +30,9 @@
   public void test_to_string_method() throws Exception {
     // ToString method of program point is used by viewer.
     CFG cfg = CFGTest.buildCFG("void foo() {foo();}");
-    ProgramPoint pp = new ProgramPoint(cfg.blocks().get(0), 0);
+    ProgramPoint pp = new ProgramPoint(cfg.blocks().get(0));
     assertThat(pp.toString()).isEqualTo("B1.0  IDENTIFIER1");
-    pp = new ProgramPoint(cfg.blocks().get(0), 2);
+    pp = pp.next().next();
```
m-g-sonar: Not really convinced by this notation. I would prefer to have the constructor still being visible, but annotated with `@VisibleForTesting`.

java-frontend/src/main/java/org/sonar/java/se/ProgramPoint.java
```diff
@@ -27,11 +29,23 @@
...
+  }
+
+  private ProgramPoint(CFG.Block block, int i) {
     this.block = block;
     this.i = i;
   }
 
+  ProgramPoint next() {
+    // pointing one element behind is allowed, it represents terminator of the CFG block
+    Preconditions.checkState(i + 1 <= block.elements().size() + 1,
```
m-g-sonar: use a variable for `block.elements().size()`

m-g-sonar:
All your commits are prefixed with "SONARJAVA-2095" when it should be "SONARJAVA-2097"

saberduck:
Fixed wrong JIRA ticket number

java-frontend/src/test/java/org/sonar/java/se/ProgramPointTest.java
saberduck: As discussed, there would be very few tests using explicit constructor and I prefer to encapsulate the inner state of the class.

java-frontend/src/main/java/org/sonar/java/se/ProgramPoint.java
saberduck: I added more detailed comment to this one, including one FIXME

PR #1249, opened 2017-02-21
Fix quality flaw : improve coverage

benzonico:


java-frontend/src/main/java/org/sonar/java/se/checks/OptionalGetBeforeIsPresentCheck.java
```diff
@@ -47,15 +47,7 @@
     "NoSuchElementException will be thrown when invoking method %s() without verifying Optional parameter.");
 
   private enum OptionalConstraint implements Constraint {
-    PRESENT, NOT_PRESENT;
-
-    @Override
-    public String valueAsString() {
```
m-g-sonar: Instead of removing that code I would add a test, because the default behavior of `valueAsString()` from the `Constraint` interface returns an empty String.

java-frontend/src/test/java/org/sonar/java/se/constraint/ConstraintTest.java
```diff
@@ -0,0 +1,38 @@
...
+  public void ObjectConstraint_should_be_valid_with_correct_constraints() throws Exception {
+    assertThat(ObjectConstraint.NULL.isValidWith(null)).isTrue();
+    assertThat(ObjectConstraint.NOT_NULL.isValidWith(null)).isTrue();
+    assertThat(ObjectConstraint.NULL.isValidWith(ObjectConstraint.NULL)).isTrue();
+    assertThat(ObjectConstraint.NOT_NULL.isValidWith(ObjectConstraint.NOT_NULL)).isTrue();
+    assertThat(ObjectConstraint.NOT_NULL.isValidWith(ObjectConstraint.NULL)).isFalse();
+    assertThat(ObjectConstraint.NULL.isValidWith(ObjectConstraint.NOT_NULL)).isFalse();
+    assertThat(ObjectConstraint.NULL.isValidWith(BooleanConstraint.TRUE)).isFalse();
+  }
+}
```
m-g-sonar: missing new line.

java-frontend/src/main/java/org/sonar/java/se/checks/OptionalGetBeforeIsPresentCheck.java
benzonico: well, as it is in fact some dead code...

m-g-sonar: This method from is used when computing flows, though the `Constraint` interface. I'm wondering if it's wise to remove it here, as it may have impact when flows are encountering the OptionalConstraints.

benzonico: This one won't even be used by flows...

benzonico: and in order to test it I would have to change its visibility. So I don't see the point of keeping something unused and badly designed.

java-frontend/src/test/java/org/sonar/java/se/constraint/ConstraintTest.java
benzonico: and updated my formatter...

PR #1250, opened 2017-02-21
Update EGViewer in order to correctly display method yields and exceptional branching

m-g-sonar:
Update the viewer with respect of recent evolution of the SE engine.

- Everytime a yield is used, the edge is now **purple**, and clickable to display the yield
- Everytime an exception is on the stack for a given node, the edge leading to that node is **red**
- Display in **red** the new nodes from the graph which have no parent, but are not the starting point (currently coming from checks creating exceptional yields)

Two examples:

image

image

### DIFF
--- a/java-frontend/src/test/java/org/sonar/java/se/ProgramStateDataProvider.java
+++ b/java-frontend/src/test/java/org/sonar/java/se/ProgramStateDataProvider.java
@@ -1,0 +1,80 @@
+/*
+ * SonarQube Java
+ * Copyright (C) 2012-2017 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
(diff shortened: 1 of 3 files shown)


java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -497,4 +497,28 @@ private static void addLearnedConstraint(ImmutableSet.Builder<LearnedConstraint>
     return result.build();
   }
 
+  /**
+   * To be used ONLY by ExplodedGraph viewer
+   */
+  @VisibleForTesting
```
benzonico: This is a rather bold change : adding API on programState for debugging purpose is a show stopper for me.

java-frontend/src/main/java/org/sonar/java/se/xproc/MethodYield.java
```diff
@@ -158,7 +158,7 @@ public boolean equals(Object obj) {
       .isEquals();
   }
 
-  static Stream<Constraint> pmapToStream(@Nullable  PMap<Class<? extends Constraint>, Constraint> pmap) {
+  public static Stream<Constraint> pmapToStream(@Nullable  PMap<Class<? extends Constraint>, Constraint> pmap) {
```
benzonico: This one should also probably not have an extended visibility but rather be reimplemented.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
benzonico: I tend to think that we may want to provide a class (in the tests) that will do the job of converting a PS to a String rather than adding an API in production code.

m-g-sonar: agreed.

PR #1255, opened 2017-02-22
SONARJAVA-2076 use flows from all yields

saberduck:


java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -143,41 +143,59 @@ private ExecutionPath(@Nullable ExplodedGraph.Edge edge, PSet<ExplodedGraph.Edge
...
-      if (learnedAssociation.isPresent()) {
-        LearnedAssociation la = learnedAssociation.get();
-        for (Class<? extends Constraint> domain : domains) {
-          Constraint constraint = parent.programState.getConstraint(la.symbolicValue(), domain);
-          if (constraint != null) {
-            flowBuilder.add(location(parent, String.format("'%s' is assigned %s.", la.symbol().name(), constraint.valueAsString())));
-          }
-        }
-        return parent.programState.getLastEvaluated();
+      if (learnedAssociation(edge).isPresent()) {
```
benzonico: `return learnedAssociation(edge).map(...).orElse(...);` ?

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -195,47 +204,39 @@ boolean notVisited(ExplodedGraph.Edge e) {
...
+        return ImmutableList.of(location(parent, String.format("Constructor implies '%s'.", constraint.valueAsString())));
       }
       String name = SyntaxTreeNameFinder.getName(nodeTree);
       String message = name == null ? constraint.valueAsString() : String.format("Implies '%s' is %s.", name, constraint.valueAsString());
-      return Stream.of(location(parent, message));
+      return ImmutableList.of(location(parent, message));
     }
 
-    private Stream<JavaFileScannerContext.Location> methodInvocationFlow(Constraint learnedConstraint, ExplodedGraph.Edge edge) {
+    private ImmutableList<JavaFileScannerContext.Location> methodInvocationFlow(Constraint learnedConstraint, ExplodedGraph.Edge edge) {
```
benzonico: I would rather rely on the interface, even for a private method.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -152,12 +152,22 @@ ExecutionPath addEdge(ExplodedGraph.Edge edge) {
       Symbol newTrackSymbol = newTrackedSymbol(edge);
 
       Set<LearnedConstraint> learnedConstraints = learnedConstraints(edge);
-      List<JavaFileScannerContext.Location> lcFlow = learnedConstraints.stream()
+      List<JavaFileScannerContext.Location> lcFlow = flowFromLearnedConstraints(edge, learnedConstraints);
+      flowBuilder.addAll(lcFlow);
+
+      if (isMethodInvocationNode(edge.parent)) {
```
benzonico: I would have this test within the flow from yield method in fact, so the algorithm is pretty straightforward and the method to compute flow from yield will only do so if required... (and return an emtpy list otherwise). (In fact if I'm correct you actually don't need that test at all...)

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -255,6 +265,14 @@ private boolean learnedConstraintForDomains(LearnedConstraint lc) {
...
+      return flowBuilder.build();
+    }
+
+    private List<JavaFileScannerContext.Location> flowFromYields(ExplodedGraph.Edge edge) {
+      ExplodedGraph.Node parent = edge.parent;
+      List<Integer> argumentIndices = correspondingArgumentIndices(symbolicValues, parent);
+      SymbolicValue returnSV = edge.child.programState.peekValue();
+      ImmutableList.Builder<JavaFileScannerContext.Location> flowBuilder = ImmutableList.builder();
       MethodYield selectedMethodYield = edge.child.selectedMethodYield(parent);
       if (selectedMethodYield != null) {
```
benzonico: See previous comment on the file but this test should be nearly the first thing in the method and that will remove a useless and you'l compute argument indices only if the yield is not null.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -268,24 +270,22 @@ private boolean learnedConstraintForDomains(LearnedConstraint lc) {
       return flowBuilder.build();
     }
 
-    private List<JavaFileScannerContext.Location> flowFromYields(ExplodedGraph.Edge edge) {
-      ExplodedGraph.Node parent = edge.parent;
-      List<Integer> argumentIndices = correspondingArgumentIndices(symbolicValues, parent);
+    private Set<List<JavaFileScannerContext.Location>> flowFromYields(ExplodedGraph.Edge edge) {
+      Set<MethodYield> methodYields = edge.yields();
+      if (methodYields.isEmpty()) {
+        return ImmutableSet.of(ImmutableList.of());
```
benzonico: Why not simply an empty set  ?

saberduck: Empty set would made `addEdge` more complicated, because no yield would need to be handled as a special case. Like this we can treat no yield and yield with no flow the same way

benzonico: That would deserve a comment and a testcase that would fail if I remove the empty list.

saberduck: Added comment explaining why having empty flow is necessary. Current tests already cover this.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
saberduck: I do not find that more readable, because returned value depends only on presence of learned association, so the lambda inside the `map` would not have use its parameter.

PR #1257, opened 2017-02-23
SONARJAVA-2100 Reference to 'this' and 'super' is non-null

saberduck:


java-frontend/src/test/files/se/ThisSuperNotNull.java
```diff
@@ -0,0 +1,23 @@
+class A {
+
+  void f() {
+    if (this == null) { // Noncompliant {{Change this condition so that it does not always evaluate to "false"}}
+
+    }
+    if (super == null) { // Noncompliant {{Change this condition so that it does not always evaluate to "false"}}
```
m-g-sonar: I'm pretty sure `super` can not be used as simple identifier, see [JLS-11.11.2](http://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.11.2).

It does not impacts the fix itself, but this code does not compile and it's bothering me a bit.

saberduck: Indeed, you are right this is not compilable. I replaced it with another usage of `super` identifier, even if that one is not explicitly demonstrating the change, since it is not directly possible to demonstrate an issue using `super` .

PR #1262, opened 2017-02-24
FQF remove methods from EG node

saberduck:


java-frontend/src/main/java/org/sonar/java/se/checks/ConditionAlwaysTrueOrFalseCheck.java
```diff
@@ -76,7 +76,7 @@ public void checkEndOfExecution(CheckerContext context) {
 
   private static Set<List<JavaFileScannerContext.Location>> flowFromNode(ExplodedGraph.Node node, boolean conditionIsAlwaysTrue) {
     List<Class<? extends Constraint>> domains = Lists.newArrayList(ObjectConstraint.class, BooleanConstraint.class);
-    Set<List<JavaFileScannerContext.Location>> flows = FlowComputation.flow(node.parent(), node.programState.peekValue(), domains);
+    Set<List<JavaFileScannerContext.Location>> flows = FlowComputation.flow(node, node.programState.peekValue(), domains);
```
benzonico: This modification looks a bit suspicious without a testcase and definitely does not sound like a quality flaw fix.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -287,7 +288,7 @@ private boolean hasConstraintForDomain(LearnedConstraint lc, Class<? extends Con
         // but this is necessary so path is returned in #addEdge and stays in the worklist in #run
         return ImmutableSet.of(ImmutableList.of());
       }
-
+      Preconditions.checkState(isMethodInvocationNode(edge.parent), "Yield present on edge where parent is not MIT");
```
benzonico: I am not really sure this is the place where we actually want to check this and I don't really see what's the risk you address with this invariant.

java-frontend/src/main/java/org/sonar/java/se/checks/ConditionAlwaysTrueOrFalseCheck.java
saberduck: reverted the change

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
saberduck: invariant check moved to `ExplodedGraph.Node.addParent`

PR #1264, opened 2017-02-27
SONARJAVA-2140 LocksLockedCheck : report issue only on methodInvocati…

benzonico:
…on trees

### DIFF
--- a/java-frontend/src/test/files/se/LocksNotUnlockedCheck.java
+++ b/java-frontend/src/test/files/se/LocksNotUnlockedCheck.java
@@ -1,6 +1,5 @@
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
-
 public class MyClass {
 
   Lock l1 = new ReentrantLock();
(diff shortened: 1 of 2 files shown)


java-frontend/src/test/files/se/LocksNotUnlockedCheck.java
```diff
@@ -1,6 +1,7 @@
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
-
+import java.util.concurrent.ConcurrentHashMap;
```
saberduck: I don't think this is necessary to reproduce the scenario

java-frontend/src/test/files/se/LocksNotUnlockedCheck.java
```diff
@@ -324,4 +325,23 @@ public void lock() {
...
+      if (locks.get(id) == lock) {
+        return lock;
+      }
+      return null;
+    } finally {
+      lock.unlock();
+    }
+  }
+
+}
```
saberduck: Same as previous comment, this can be much more simplified. We need only binary expression in the if statement to reproduce the problem

PR #1278, opened 2017-03-02
Fix coverage

benzonico:


java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
```diff
@@ -86,7 +86,7 @@ public void setJavaVersion(JavaVersion javaVersion) {
 
   public void visitFile(@Nullable Tree parsedTree) {
     semanticModel = null;
-    CompilationUnitTree tree = new JavaTree.CompilationUnitTreeImpl(null, Lists.newArrayList(), Lists.newArrayList(), null);
+    CompilationUnitTree tree = new JavaTree.CompilationUnitTreeImpl(null, new ArrayList<>(), new ArrayList<>(), null);
```
m-g-sonar: This has impact on coverage?!

PR #1280, opened 2017-03-03
SONARJAVA-2144 Static field update in constructor check

m-g-sonar:


its/ruling/src/test/resources/sonar-server/squid-S3010.json
```diff
@@ -0,0 +1,5 @@
+{
+'org.codehaus.sonar:sonar-server:src/main/java/org/sonar/server/platform/MasterServletFilter.java':[
+48,
```
m-g-sonar: wonder if the rule should be smart... and identify singleton patterns (by maybe ignoring fields called `singleton`, `instance` or `uniqueInstance`).

saberduck: I would say no, this is not very standard singleton pattern.

PR #1286, opened 2017-03-07
SONARJAVA-2142 Subclass references in class initialization check

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/SubClassReferenceInitalizationCheck.java
```diff
@@ -0,0 +1,100 @@
...
+
+    checkVariables(members, classType);
+    checkStaticInitilizers(members, classType);
+  }
+
+  private void checkVariables(List<Tree> members, Type classType) {
+    members.stream()
+      .filter(member -> member.is(Tree.Kind.VARIABLE))
+      .map(VariableTree.class::cast)
+      .filter(SubClassReferenceInitalizationCheck::isStaticVariable)
```
benzonico: Why only static variables ? seems to me that initialization issue can also occur during instanciation initialization. http://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.4.2

java-checks/src/main/java/org/sonar/java/checks/SubClassReferenceInitalizationCheck.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+      // skip the variable
+      scan(tree.expression());
+    }
+
+    @Override
+    public void visitIdentifier(IdentifierTree tree) {
+      Type type = tree.symbolType();
+      if (!classType.equals(type) && type.isSubtypeOf(classType)) {
```
benzonico: If I'm correct this will raise an issue for anonymous classes ?
```
class Foo {
static Foo foo = new Foo() { // issue ?
 .. . .
   FooSubclass fun() { // issue
     ... 
   }
}
```

java-checks/src/main/java/org/sonar/java/checks/SubClassReferenceInitalizationCheck.java
m-g-sonar: Yep, but §12.4 mention explicitly that the initialization process only relates to static fields and static initializers.

java-checks/src/main/java/org/sonar/java/checks/SubClassReferenceInitalizationCheck.java
m-g-sonar: Indeed. forgot about the anonymous classes.

java-checks/src/main/java/org/sonar/java/checks/SubClassStaticReferenceCheck.java
```diff
@@ -50,15 +50,18 @@ public void visitNode(Tree tree) {
     Type classType = classTree.symbol().type();
     List<Tree> members = classTree.members();
 
-    checkVariables(members, classType);
+    // JLS 12.4. Initialization of Classes and Interfaces:
+    // Initialization of a class consists of executing its static initializers and the initializers for static fields (class variables)
+    // declared in the class.
+    checkStaticVariables(members, classType);
     checkStaticInitilizers(members, classType);
```
benzonico: TYPO !

PR #1302, opened 2017-03-10
SONARJAVA-2177: @RequestMapping methods should be public

agigleux:
Coverage 100%
Use IssuableSubscriptionVisitor
Use MethodSymbol
0 SonarLint Issue


java-checks/src/main/java/org/sonar/java/checks/spring/RequestMappingMethodPublicCheck.java
```diff
@@ -0,0 +1,66 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.METHOD);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodTree methodTree = (MethodTree) tree;
+    Symbol.MethodSymbol methodSymbol = methodTree.symbol();
+
+    if (methodSymbol != null
```
benzonico: This null check should in fact be avoided. The only case where a method symbol from a method tree can be null is when semantic is not available. In the ruling that happens for package java.lang of JDK. If it is not available, then it makes actually no sense to run this check so, in fact at the beginning of the visitNode you can have a : 
```
if(!hasSemantic()) {
  return;
}
```

You can test this by using `JavaCheckVerifier.verifyNoIssueWithoutSemantic` (which you already use).

java-checks/src/main/java/org/sonar/java/checks/spring/RequestMappingMethodPublicCheck.java
```diff
@@ -0,0 +1,66 @@
...
+      && isClassController(methodSymbol)
+      && isRequestMappingAnnotated(methodSymbol)
+      && !isPublicMethod(methodSymbol)) {
```
benzonico: Is it ok to raise an issue if the method has package visibility or is protected ?

java-checks/src/main/java/org/sonar/java/checks/spring/RequestMappingMethodPublicCheck.java
```diff
@@ -0,0 +1,66 @@
...
+import org.sonar.plugins.java.api.semantic.SymbolMetadata;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S3751")
+public class RequestMappingMethodPublicCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.METHOD);
```
benzonico: Better to use Collections.singletonList rather than guava for this.

java-checks/src/main/java/org/sonar/java/checks/spring/RequestMappingMethodPublicCheck.java
agigleux: I thought the same as you and finally convinced myself yes this is OK to raise an issue for package or protected. We want here to enforce all methods with @RequestMapping annotation to be 'public' to clearly show to the developer any other visibility will have no effect.

java-checks/src/main/java/org/sonar/java/checks/spring/RequestMappingMethodPublicCheck.java
agigleux: Done

java-checks/src/main/java/org/sonar/java/checks/spring/RequestMappingMethodPublicCheck.java
agigleux: Is it just because at some point you want to get rid of Guava or is there another reason?

PR #1304, opened 2017-03-13
SONARJAVA-2159 Only target method invocation with side effects

m-g-sonar:


java-checks/src/test/files/checks/AssertOnBooleanVariableCheck.java
```diff
@@ -1,22 +1,45 @@
...
     boolean removed = myList.remove(myList.get(0));
     assert removed;
-    assert(removed);
+
     int i = 0;
-    assert (removed = i); // Noncompliant
-    assert bar() == 1; // Noncompliant
-    assert i++; // Noncompliant
+    assert (removed = i); // Compliant - no call to side-effect method
+    assert i++; // Compliant
```
saberduck: This doesn't compile, `i++` is not boolean expr, also L31 doesn't compile for same reason

PR #1314, opened 2017-03-16
SONARJAVA-2161 Handle default initializer values

m-g-sonar:


java-checks/src/test/files/checks/DeadStoreCheck.java
```diff
@@ -246,3 +247,82 @@ void assertStatement(boolean x) {
...
+    y = i;
+
+    int z1 = -(1); // Compliant
+    z1 = 0;        // Noncompliant
+    z1 = 1;        // Noncompliant
+    z1 = i;
+
+    int z2 = -1;   // Compliant
+    z2 = 0;        // Noncompliant
+    z2 = 1;        // Noncompliant
```
benzonico: I can understand the test of assigning 0 or 1 not in declaration once but why repeating it over and over ? this is noise.

java-checks/src/main/java/org/sonar/java/checks/DeadStoreCheck.java
```diff
@@ -176,12 +178,38 @@ private static void handleIdentifier(Set<Symbol> out, Set<Tree> assignmentLHS, I
...
-      createIssue(((VariableTreeImpl) localVar).equalToken(), localVar.initializer(), symbol);
+    ExpressionTree initializer = localVar.initializer();
+    if (initializer != null && !isUsualDefaultValue(initializer) && !out.contains(symbol)) {
+      createIssue(((VariableTreeImpl) localVar).equalToken(), initializer, symbol);
     }
     out.remove(symbol);
   }
 
+  private static boolean isUsualDefaultValue(ExpressionTree initializer) {
+    ExpressionTree expr = ExpressionUtils.skipParentheses(initializer);
```
benzonico: Wondering if using a switch on kind wouldn't help to make things a bit clearer here.

java-checks/src/main/java/org/sonar/java/checks/DeadStoreCheck.java
```diff
@@ -176,12 +178,38 @@ private static void handleIdentifier(Set<Symbol> out, Set<Tree> assignmentLHS, I
 
   private void handleVariable(Set<Symbol> out, VariableTree localVar) {
     Symbol symbol = localVar.symbol();
-    if (localVar.initializer() != null && !out.contains(symbol)) {
-      createIssue(((VariableTreeImpl) localVar).equalToken(), localVar.initializer(), symbol);
+    ExpressionTree initializer = localVar.initializer();
+    if (initializer != null && !isUsualDefaultValue(initializer) && !out.contains(symbol)) {
```
benzonico: arguable but maybe passing the initializer and returning false if it is null could make things a bit clearer.

m-g-sonar: I won't change that part. I prefer to have the explicit null check of the initializer before checking for usual default value (which has been simplified).

java-checks/src/test/files/checks/DeadStoreCheck.java
```diff
@@ -246,3 +247,69 @@ void assertStatement(boolean x) {
...
+
+    int b = (0);  // Compliant
+    b = -1;       // Noncompliant - Only taken into consideration when used in initializer
+    b = 0;        // Noncompliant
+    b = 1;        // Noncompliant
+    int c = +1;   // Compliant
+    int d = -(1); // Compliant
+    int e = -1;   // Compliant
+
+    long myLong = -1L; // Noncompliant - Only target int literals
```
benzonico: confusing comment : it's the EXCLUSION that targets only int literals.

PR #1317, opened 2017-03-16
Make compatible with surefire reuseForks=false (round 2)

fabdouglas:
JacocoController.getInstance() is executed too soon in case of surefire option reuseForks=false.
In this configuration, the listener is created without javaagent enabled. It is enabled only when the VM is forked while starting a test.
In addition I've fixed a small issue in case of broken test : the data was tagged as anonymous. Previous implementation saved the session id just before the dump, at the end of the tests.
This PR sets the SessionId immediatly when the tests starts.
This is cleaner/rebase version of #324

Tests have been updated, documentation added, and the whole has been tested for JUnit4 with both modes : reuse= false and reuse = true.

### DIFF
--- a/sonar-jacoco-listeners/src/main/java/org/sonar/java/jacoco/JUnitListener.java
+++ b/sonar-jacoco-listeners/src/main/java/org/sonar/java/jacoco/JUnitListener.java
@@ -27,24 +27,53 @@ import org.junit.runner.notification.RunListener;
  */
 public class JUnitListener extends RunListener {
 
-  protected final JacocoController jacoco;
+  /**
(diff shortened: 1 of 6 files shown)


benzonico:
Hi and thanks for the contribution. 

You PR mix two problems and so as such I won't merge it as is. Inspired by your contribution I tackled the problem of reuseForks with this PR : https://github.com/SonarSource/sonar-java/pull/1371 
Note, I added an integration test to verify listener is properly executed when using reuseForks=true. 
As a consequence, I am closing this PR. 

As per the issue with failing tests, could you reach out to sonarqube@googlegroups.com to qualify the problem with us. So we can open a ticket and get it fixed properly. 

Thanks again for the contribution !

PR #1321, opened 2017-03-20
SONARJAVA-2197 Improve log message when JSON file for rule is not found.

saberduck:


java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/CheckVerifier.java
```diff
@@ -239,7 +239,9 @@ private static RemediationFunction remediationFunction(AnalyzerMessage issue) {
       }
     } catch (IOException | JsonParseException e) {
```
SonarTech: MAJOR Either log or rethrow this exception. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS1166)

SonarTech:
* ![MAJOR][MAJOR] 1 major

[MAJOR]: https://raw.githubusercontent.com/SonarCommunity/sonar-github/master/images/severity-major.png 'Severity: MAJOR'

java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/CheckVerifier.java
saberduck: Exception is ignored intentionally.

java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/CheckVerifier.java
```diff
@@ -239,7 +239,9 @@ private static RemediationFunction remediationFunction(AnalyzerMessage issue) {
...
       // Failed to open json file, as this is not part of API yet, we should not fail because of this
-      LOG.warn("Exception parsing JSON for rule " + ruleKey, e);
+      LOG.warn("JSON metadata for rule " + ruleKey + " couldn't be loaded. Constant remediation function will be assumed."
+        + " Provide rule JSON file if you want to customize metadata for tests.");
```
m-g-sonar: We defined nowhere what is the expected format of these JSON files, this is solely the approach we chose to read metadata. Currently, this is the responsibility of the custom plugin implementor to load rules metatada when registering rules (see https://github.com/SonarSource/sonar-java/blob/4.7-RC1/sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java#L98). 

It has previously been done using annotations, but using JSON files is not the standard way, as we didn't defined any standard. Problem is that a "smart developer" may create a JSON file with only the remediation metadata, which will pass the tests, but if he does not load correctly the rules when implementing the RulesDefinition interface, remediation cost will still be missing.

Moreover, remediation function and cost are pretty advanced features, not even mentioned in the tutorial. Consequently, I would:
- not mention any JSON file, but simply mention that `Remediation function/cost is not provided, and assumed constant (default)`.
- put the whole log in `debug` rather than `warning` .

PR #1325, opened 2017-03-20
SONARJAVA-2108 Update S3423 to use secondaries on lines

m-g-sonar:


java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/PomCheckVerifier.java
```diff
@@ -104,7 +105,12 @@ public void reportIssue(PomCheck check, LocatedTree tree, String message) {
     public void reportIssue(PomCheck check, int line, String message, List<Location> secondary) {
       AnalyzerMessage analyzerMessage = new AnalyzerMessage(check, getFile(), line, message, 0);
       for (Location location : secondary) {
-        AnalyzerMessage secondaryLocation = new AnalyzerMessage(check, getFile(), location.tree.startLocation().line(), location.msg, 0);
+        AnalyzerMessage secondaryLocation;
+        if (location.onLine()) {
+          secondaryLocation = new AnalyzerMessage(check, getFile(), location.line, location.msg, 0);
```
saberduck: You can move the logic of getting the line from Location into the Location class and that way no need to duplicate constructor call here

java-frontend/src/main/java/org/sonar/java/xml/maven/PomCheckContext.java
```diff
@@ -37,11 +40,25 @@
...
+      this.line = null;
+    }
+
+    public Location(String msg, int line) {
+      this.msg = msg;
+      this.tree = null;
+      this.line = line;
+    }
+
+    public boolean onLine() {
```
saberduck: As mentioned above, instead of this method, you can have method giving you the line number either from tree or the line

java-frontend/src/main/java/org/sonar/java/xml/maven/PomCheckContextImpl.java
```diff
@@ -67,15 +68,28 @@ public void reportIssue(PomCheck check, int line, String message, List<Location>
   }
 
   @VisibleForTesting
-  static AnalyzerMessage getSecondaryAnalyzerMessage(JavaCheck check, File file, Location location) {
-    XmlLocation startLocation = location.tree.startLocation();
-    XmlLocation endLocation = location.tree.endLocation();
-    TextSpan ts = new TextSpan(
-      startLocation.line(), offsetFromColumn(startLocation.column()),
-      endLocation.line(), offsetFromColumn(endLocation.column()));
+  AnalyzerMessage getSecondaryAnalyzerMessage(JavaCheck check, File file, Location location) {
```
saberduck: This logic can be moved to Location class too

m-g-sonar: nop. let's keep it minimal. I don't want to introduce notion of file+check+analyzerMessage into the location class.

saberduck: Sorry, comment should have been one line below. You only need `TextSpan` in the location class. Up to you if this is too much to be there.

m-g-sonar: Unfortunately in order to compute the line size and get the correct text span I still need the access the `sonarComponents` object, which I don't want to see in the `Location` class. I'll let it that way. No need to rewrite the whole context for that simple case.

saberduck: OK, I haven't noticed

PR #1327, opened 2017-03-21
SONARJAVA-2198 Update list of deprecated algorithms and targeted APIs

saberduck:


java-checks/src/main/java/org/sonar/java/checks/DeprecatedHashAlgorithmCheck.java
```diff
@@ -34,20 +35,67 @@
...
+    boolean match(String algorithm) {
+      String normalizedName = algorithm.replaceAll("-", "").toLowerCase(Locale.ROOT);
+      return normalizedName.contains(name().toLowerCase(Locale.ROOT));
+    }
+  }
+
+  /**
+   * These APIs have static getInstance method to get an implementation of some crypto algorithm.
+   * javax.crypto.Cipher is missing from this list, because it is covered by rule S2278 {@link AvoidDESCheck}
+   * Details can be found here <a href="http://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html">Security Standard Names</a>
```
m-g-sonar: Why not using java 8 doc? Did you check it if there is some differences?

java-checks/src/main/java/org/sonar/java/checks/DeprecatedHashAlgorithmCheck.java
```diff
@@ -34,20 +35,67 @@
...
+    DSA {
+      @Override
+      boolean match(String algorithm) {
+        // exact match required for DSA, so it doesn't match ECDSA
+        return "DSA".equals(algorithm);
+      }
+    };
+
+    boolean match(String algorithm) {
+      String normalizedName = algorithm.replaceAll("-", "").toLowerCase(Locale.ROOT);
```
m-g-sonar: Use `Locale.ENGLISH` (the one which should be used everywhere, don't know why Locale.ROOT is used in some other parts).

java-checks/src/main/java/org/sonar/java/checks/DeprecatedHashAlgorithmCheck.java
```diff
@@ -90,27 +141,27 @@
...
     if (defaultPropertyValue != null) {
       expectedAlgorithm = defaultPropertyValue;
     }
     if (expectedAlgorithm.is(Tree.Kind.STRING_LITERAL)) {
-      String algo = LiteralUtils.trimQuotes(((LiteralTree) expectedAlgorithm).value());
-      return algo.replaceAll("-", "");
+      String algorithmName = LiteralUtils.trimQuotes(((LiteralTree) expectedAlgorithm).value());
+      return Arrays.stream(InsecureAlgorithm.values())
+        .filter(alg -> alg.match(algorithmName))
+        .findFirst().orElse(null);
```
m-g-sonar: Maybe change the method to return an `Optional<InsecureAlgorithm>` rather than mixing null and stream api?

saberduck: OK, however I didn't find an elegant way to get rid of nulls in the caller anyway.

PR #1331, opened 2017-03-22
SONARJAVA-2149 Ignore break and continue in finally when inside loop

saberduck:


java-checks/src/main/java/org/sonar/java/checks/ReturnInFinallyCheck.java
```diff
@@ -52,53 +58,131 @@ public void visitTryStatement(TryStatementTree tree) {
...
+  public void visitSwitchStatement(SwitchStatementTree tree) {
+    isInFinally.push(tree.kind());
+    super.visitSwitchStatement(tree);
+    isInFinally.pop();
+  }
+
+  @Override
   public void visitReturnStatement(ReturnStatementTree tree) {
-    reportIssue(tree.returnKeyword(), "return");
+    reportIssue(tree.returnKeyword(), tree.kind());
```
m-g-sonar: All your first parameters are `SyntaxToken`, just modify the method to have a `SyntaxToken` as first parameter instead of a tree. It would allow you to directly access it's value, instead of using that useless `statement()` method

java-checks/src/main/java/org/sonar/java/checks/ReturnInFinallyCheck.java
```diff
@@ -52,53 +58,131 @@ public void visitTryStatement(TryStatementTree tree) {
...
+      case THROW_STATEMENT:
+        return "throw";
+      default:
+        throw new IllegalArgumentException("Unsupported jump statement!");
+    }
   }
 
+  private boolean isAbruptFinallyBlock(Tree.Kind jumpKind) {
+    if (isInFinally.isEmpty()) {
+      return false;
```
m-g-sonar: Are you covering this branch? in your tests?

java-checks/src/main/java/org/sonar/java/checks/ReturnInFinallyCheck.java
```diff
@@ -52,53 +58,131 @@ public void visitTryStatement(TryStatementTree tree) {
...
-      context.reportIssue(this, tree, "Remove this "+statement+" statement from this finally block.");
+  private void reportIssue(Tree tree, Tree.Kind jumpKind) {
+    if (isAbruptFinallyBlock(jumpKind)) {
+      context.reportIssue(this, tree, "Remove this " + statement(jumpKind) + " statement from this finally block.");
     }
   }
 
-  private boolean isInFinally() {
-    return !isInFinally.isEmpty() && isInFinally.peek();
+  private static String statement(Tree.Kind kind) {
```
m-g-sonar: method is useless, as long as you provide a `SyntaxToken`  in `reportIssue`

java-checks/src/main/java/org/sonar/java/checks/ReturnInFinallyCheck.java
```diff
@@ -23,21 +23,27 @@
...
+import org.sonar.plugins.java.api.tree.WhileStatementTree;
 
 import java.util.Deque;
 import java.util.LinkedList;
 
 @Rule(key = "S1143")
 public class ReturnInFinallyCheck extends BaseTreeVisitor implements JavaFileScanner {
 
-  private final Deque<Boolean> isInFinally = new LinkedList<>();
+  private final Deque<Tree.Kind> isInFinally = new LinkedList<>();
```
m-g-sonar: I would change the variable name as it's counterintuitive now. `nestedTreeKind`?

java-checks/src/main/java/org/sonar/java/checks/ReturnInFinallyCheck.java
saberduck: It was covered even before here https://github.com/SonarSource/sonar-java/blob/97306da9245167ec3f6038b1ed0dc3d193990e5a/java-checks/src/test/files/checks/ReturnInFinallyCheck.java#L64-L64  . It is bit hidden though.

m-g-sonar: sorry missede it! :+1:

java-checks/src/main/java/org/sonar/java/checks/ReturnInFinallyCheck.java
saberduck: good point, I went for `treeKindStack`

PR #1336, opened 2017-03-24
SONARJAVA-2083  Fix usages of annotations in S1128 (PR includes SONARJAVA-2205 and SONARJAVA-2206)

m-g-sonar:


java-checks/src/test/files/checks/UselessImportCheck/WithPackage.java
```diff
@@ -54,3 +58,14 @@ void test() throws MyException, MyException2 {
   }
   void foo(@Nullable int x){}
 }
+
+class A {
+  byte @MyAnnotation2 [] table = null;
```
benzonico: `byte @MyAnnotation2 [] @MyAnnotation4 [] table = null `
Are we sure that annotations on all dimensions are correctly detected ?

PR #1341, opened 2017-03-27
SONARJAVA-2128 Avoid resolution of java.lang.Enum constructor

benzonico:


java-frontend/src/test/files/sym/EnumConstructor.java
```diff
@@ -0,0 +1,11 @@
+enum EnumConstructor {
+  EXAMPLE("String", 0);
+
+  private String string;
+  private Integer i;
+
+  EnumConstructor(String string, Integer i) { // unused private constructor because the constructor from the superclass was picked (strict invocation over loose invocation).
```
m-g-sonar: remove the comment, we resolve now correctly the constructor.

PR #1348, opened 2017-03-30
SONARJAVA-2213 SONARJAVA-2214 Improve flow computation

benzonico:
Avoid useless computation and a add a safety mechanism when computation goes out of hand.


java-frontend/src/test/java/org/sonar/java/se/checks/FlowComputationTest.java
```diff
@@ -72,6 +71,10 @@ public void test_multiple_paths() {
   @Test
   public void test_multiple_paths_xproc() {
     JavaCheckVerifier.verify("src/test/files/se/FlowComputationMultiplePathXProc.java", new NullDereferenceCheck(), new ConditionAlwaysTrueOrFalseCheck());
+  }
 
+  @Test
+  public void test_trigger_yield_flow_computation_only_on_relevant_yields() throws Exception {
+    JavaCheckVerifier.verify("src/test/files/se/UselessFlowComputation.java",  new NullDereferenceCheck(), new ConditionAlwaysTrueOrFalseCheck());
```
m-g-sonar: This test is annoying me. If I understand what it is indirectly testing, we provide no way to verify that no flow has been computed for `fun2` and `fun3`. If we change this behavior, and restart computing flow for all the methods, this test won't fail. Drop it?

java-frontend/src/main/java/org/sonar/java/se/xproc/MethodYield.java
```diff
@@ -175,6 +179,15 @@ public boolean equals(Object obj) {
     if(node == null || behavior == null) {
       return Collections.emptySet();
     }
+    String key = parameterIndices.stream().sorted().map(Object::toString).reduce("", String::concat);
```
m-g-sonar: I'm not sure about that generating the key like this will be unique: from how I see it,
if we call the method with [1,2], we will end up with the same key (`"12"`) as asking for the 12th parameter with the list [12]. (I know this is pretty unlikely, but still...)

benzonico: Let's add a "," in between then. indeed.

PR #1352, opened 2017-03-31
SONARJAVA-1071 Update S1130 to raise issue on throws which are not thrown in the body of the method

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -37,32 +56,166 @@
 @RspecKey("S1130")
 public class RedundantThrowsDeclarationCheck extends IssuableSubscriptionVisitor {
 
+  public static final Set<String> SERIALIZABLE_CONTRACT_METHODS = ImmutableSet.of(
```
benzonico: Why public ? this is also a duplicate of  : ` org.sonar.java.checks.serialization.SerializableContract.SERIALIZABLE_CONTRACT_METHODS` is it intentional ?

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -37,32 +56,166 @@
...
+    }
+
+    return thrownExceptions.stream().noneMatch(t -> t.isSubtypeOf(exceptionType));
+  }
+
+  private static boolean isOverridingOrDesignedForExtension(MethodTree methodTree) {
+    // we need to be sure that it's not an override
+    return !Boolean.FALSE.equals(((MethodTreeImpl) methodTree).isOverriding())
+      || SERIALIZABLE_CONTRACT_METHODS.contains(methodTree.simpleName().name())
+      // kill the noise - usually designed for extension
```
benzonico: Unecessary comment IMO : I don't get it right away anyway. And I think if it deserves a comment it should be on the method itself, not the call. 
Moreover : kill the noise is a term used internally. should not be part of code IMO.

java-checks/src/test/files/checks/RedundantThrowsDeclaration.java
```diff
@@ -58,3 +58,105 @@ public String plop() {
...
+
+  void foo7() throws MyException { // Compliant - False-Negative - studying control flow is not done
+    try {
+      mok();
+    } catch(MyException e) {
+      // do nothing
+    }
+  }
+
+  void foo8(Throwable t) throws java.io.IOException { // Compliant - kill the noise due to SONARJAVA-1778
```
benzonico: Again : I tend to think that : false negative is better suited than kill the noise for comments in code.

PR #1354, opened 2017-04-03
SONARJAVA-2204 Do not raise issue if method with same name is not resolved

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -20,41 +20,105 @@
...
+    } else {
+      return ((IdentifierTree) typeTree).name();
+    }
+  }
+
+  private void checkIfUnused(MethodTree methodTree) {
+    Symbol symbol = methodTree.symbol();
+    if (methodTree.modifiers().annotations().isEmpty() && symbol.isPrivate() && symbol.usages().isEmpty()) {
+      if (methodTree.is(Tree.Kind.CONSTRUCTOR)) {
+        if (!methodTree.parameters().isEmpty()) {
```
benzonico: I would probably have a dedicated isConstructorWithNoParam method and have only one call to .add

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -20,41 +20,105 @@
...
     }
-    MethodTree node = (MethodTree) tree;
-    Symbol symbol = node.symbol();
-    if (node.modifiers().annotations().isEmpty() && symbol.isPrivate() && symbol.usages().isEmpty()) {
-      if (node.is(Tree.Kind.CONSTRUCTOR)) {
-        if (!node.parameters().isEmpty()) {
-          reportIssue(node.simpleName(), "Remove this unused private \"" + node.simpleName().name() + "\" constructor.");
+    if (tree.is(Tree.Kind.METHOD, Tree.Kind.CONSTRUCTOR)) {
+      checkIfUnused((MethodTree) tree);
+    } else {
```
benzonico: I would make explicit tests for every kind (or switch) and have overloaded method for those. That will make classCast less likely to happen. Here there is a lot of implicit going on. Let's make thing explicit.

checkIfUnknown(MethodInvocationTree ) 
checkIfUnknown(NewClasstree  )

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -20,41 +20,105 @@
...
+    }
+  }
+
+  private static String constructorName(TypeTree typeTree) {
+    if (typeTree.is(Tree.Kind.PARAMETERIZED_TYPE)) {
+      return constructorName(((ParameterizedTypeTree) typeTree).type());
+    } else if (typeTree.is(Tree.Kind.MEMBER_SELECT)) {
+      return ((MemberSelectExpressionTree) typeTree).identifier().name();
+    } else {
+      return ((IdentifierTree) typeTree).name();
```
benzonico: trade off here but I would trade the possible ClassCastException against one uncovered case (that should throw an IllegalStateException )

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
m-g-sonar: I didn't wanted to change this part. I simply extracted the existing method. but OK.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -20,41 +20,138 @@
...
+  private static String constructorName(TypeTree typeTree) {
+    switch (typeTree.kind()) {
+      case PARAMETERIZED_TYPE:
+        return constructorName(((ParameterizedTypeTree) typeTree).type());
+      case MEMBER_SELECT:
+        return ((MemberSelectExpressionTree) typeTree).identifier().name();
+      case IDENTIFIER:
+        return ((IdentifierTree) typeTree).name();
+      default:
+        throw new IllegalStateException("Unexpected Type tree not as constructor.");
```
benzonico: maybe we can use english ?

PR #1358, opened 2017-04-04
SONARJAVA-2211 Compute behavior only once

benzonico:


java-frontend/src/main/java/org/sonar/java/se/SymbolicExecutionVisitor.java
```diff
@@ -64,14 +64,15 @@ public void visitNode(Tree tree) {
...
+    ExplodedGraphWalker walker = egwFactory.createWalker(behaviorCache, (SemanticModel) context.getSemanticModel());
     try {
       Symbol.MethodSymbol methodSymbol = methodTree.symbol();
-      ExplodedGraphWalker walker = egwFactory.createWalker(behaviorCache, (SemanticModel) context.getSemanticModel());
       if (methodCanNotBeOverriden(methodSymbol)) {
-        MethodBehavior methodBehavior = new MethodBehavior(methodSymbol);
-        behaviorCache.add(methodSymbol, methodBehavior);
-        methodBehavior = walker.visitMethod(methodTree, methodBehavior);
-        methodBehavior.completed();
+        MethodBehavior methodBehavior = behaviorCache.add(methodSymbol);
```
m-g-sonar: I don't like the name of this method, it makes more than "adding" the symbol to the cache.

benzonico: what's your suggestion ?

m-g-sonar: `methodBehavior(...)` ?
`methodBehaviorForSymbol(...)` ?

PR #1365, opened 2017-04-05
SONARJAVA-2221 Do not raise issue if unknown identifier has same name

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateFieldCheck.java
```diff
@@ -62,63 +68,120 @@
...
+      case IDENTIFIER:
+        collectUnknownIdentifier((IdentifierTree) tree);
+        break;
+      default:
+        throw new IllegalStateException("Unexpected subscribed tree.");
+    }
+  }
+
+  private void collectUnknownIdentifier(IdentifierTree identifier) {
+    if (!isVariableIdentifier(identifier) && !isMethodIdentifier(identifier) && identifier.symbol().isUnknown()) {
```
benzonico: I would test first if the symbol is unknown, faster test and then only add if it is not var/method id... 

And by the way : while I can understand the exclusion of unknown method identifier, how can a variable identifier be unknown ? If I'm correct, the isVariableIdentifier can be entirely dropped if you test first isUnknown

m-g-sonar: OK for first part.
For the second part: I'm not quite sure if it's a bug or a feature... From the `VariableTree`, you can access the variable symbol (using `symbol()`), but if you try to get the symbol from the `IdentifierTree` corresponding to the `simpleName()` of the `VariableTree`, you get _unknown_.

java-checks/src/main/java/org/sonar/java/checks/CallToDeprecatedMethodCheck.java
```diff
@@ -75,6 +76,9 @@ public void leaveNode(Tree tree) {
   }
 
   private void checkIdentifierIssue(IdentifierTree identifierTree) {
+    if (isVariableIdentifier(identifierTree)) {
```
benzonico: maybe juste renaming this method : isVariableDeclarationSimpleName ?

PR #1367, opened 2017-04-05
SONARJAVA-2224 Fix flow from yields for varargs methods

benzonico:


java-frontend/src/main/java/org/sonar/java/se/xproc/MethodYield.java
```diff
@@ -189,13 +189,16 @@ public boolean equals(Object obj) {
 
   private Set<SymbolicValue> getSymbolicValues(List<Integer> parameterIndices) {
     ImmutableSet.Builder<SymbolicValue> parameterSVs = ImmutableSet.builder();
-    for (Integer parameterIndex : parameterIndices) {
+    parameterIndices.stream()
+      // Ignore parameters for varargs
```
m-g-sonar: that comment is wrong. We should only ignore the last parameters, starting from the variadic index.

PR #1368, opened 2017-04-05
SONARJAVA-2202 Update rule description and bug/code smell

benzonico:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S1815_java.html
```diff
@@ -6,4 +6,6 @@
   <li> Protected methods which override a parent class method. </li>
   <li> Protected methods of an abstract class. </li>
 </ul>
+<h2>Deprecated</h2>
+<p>This rule is deprecated, and will eventually be removed.</p>
```
m-g-sonar: well, we don't have the rule anymore in the analyzer as far as I can tell. Maybe we can get rid of the `html`and `json` files as well instead.

PR #1373, opened 2017-04-09
SONARJAVA-2230 - Add restassured support for S2699

etheld:
it includes: body, content, contentType, cookie, cookies, header,
headers, spec, specfication, statusCode, statusLine, time

### DIFF
--- a/java-checks/pom.xml
+++ b/java-checks/pom.xml
@@ -222,6 +222,16 @@
                   <version>0.29</version>
                 </artifactItem>
                 <artifactItem>
+                  <groupId>org.hamcrest</groupId>
+                  <artifactId>hamcrest-library</artifactId>
+                  <version>1.3</version>
+                </artifactItem>
+                <artifactItem>
+                   <groupId>io.rest-assured</groupId>
+                   <artifactId>rest-assured</artifactId>
+                   <version>3.0.2</version>
+                </artifactItem>
+                <artifactItem>
                   <groupId>com.fasterxml.jackson.core</groupId>
                   <artifactId>jackson-annotations</artifactId>
                   <version>2.8.4</version>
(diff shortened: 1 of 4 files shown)


m-g-sonar:
Hey @etheld ,

Thanks for opening this PR! You are obviously quite aware of it, but this is currently still incomplete. Next step would be to add a new test case in test class [`AssertionsInTestsCheckTest`](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/test/java/org/sonar/java/checks/AssertionsInTestsCheckTest.java), which will target rest-assured. The simplest way would be to create a new file in `src/test/files/checks/` including some tests using the framework, and verify that they are correctly detected as being valid.

**Important note:** As the rule is relying on semantic, it requires rest-assured bytecode to correctly detect method invocations. Consequently, you will also need to provide it. To do so, add a dependency to the test-jars of the java-checks project, by simply adding rest-assured artifact item to the already existing list (see [java-checks/pom.xml#L85](https://github.com/SonarSource/sonar-java/blob/master/java-checks/pom.xml#L85) for items already used in our tests). Once added, simply re-compile the project, which will download the artifact and allow tests to read the rest-assured bytecode.

Cheers,
Michael

etheld:
Hey @m-g-sonar thanks for the help. Yes, I was too eager to open a pull request, that's why I marked with WIP later on. I was looking for this, I tested it locally with some sample code. I will add those test cases and find a way to run IT tests. I could not get them to build yet.

m-g-sonar:
Okay, no worries. I'll keep an eye on your PR and wait for your next commit(s).

For this very specific PR, however, you most probably won't have to modify the results of the "ruling" (ITs playing all the rules on some sample projects), as I'm pretty sure that none of them is using the rest-assured framework. A complete test-file will then be quite usefull! :)

etheld:
Sorry for the long delay, had limited access to my computer. I did provide tests for rest assured, it uses hamcrest matchers so I had to add hamcrest-library to the project as well. I did not use hamcrest too much, so if there is a better way to do that, let me know.

The tests are passing(at least locally :) ). I could not get IT tests running, but as you suggested it should be okay. I did install eclipse formatter( I use intellij), so I hope the formatting is fine. 

I ammended the last commit(so it is force pushed), so now it is a single commit.

Let me know what you think

m-g-sonar:
Great, LGTM! Nice work.
I'm going to open a new PR with your commit in order to trigger our internal QA. If everything goes well, we will merge it.

etheld:
Cool, thanks. One more question, I only added restassured 2.0 to the list, do you think it is worth to add the previous versions? or it should be included in the documentation that sonarqube only supports restassured 2.x?

Let me know how the other pull request goes.

m-g-sonar:
I'm not sure that we need to specify it anywhere, but I'm going to ask internally. Note that the current rule description does not specify all the test frameworks which are detected. Nevertheless, if we change it in order to list them, I'll make sure to explicitly mention that it's only targeting rest-assured 2.0.

m-g-sonar:
Hey, QA green on our side, so I merged your work on Master. Thanks again for your contribution!

After discussing internally, we will also update the description of the rule to mention the frameworks which are supported, now with rest-assured, thanks to you!

Cheers,
Michael

PR #1378, opened 2017-04-11
SONARJAVA-2131 Classes with main are not considered as utility classes

saberduck:


java-checks/src/main/java/org/sonar/java/checks/UtilityClassWithPublicConstructorCheck.java
```diff
@@ -125,4 +139,16 @@ private static boolean hasPublicModifier(MethodTree methodTree) {
...
   }
 
+  private static class SpringBootVisitor extends BaseTreeVisitor {
+
+    private static final MethodMatcher SPRING_BOOT = MethodMatcher.create().typeDefinition("org.springframework.boot.SpringApplication").name("run").withAnyParameters();
+    boolean springBootDetected;
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree tree) {
+      if (SPRING_BOOT.matches(tree)) {
```
benzonico: What if  `run` method is called within an anonymous class ? do we want to raise an issue ? This probably deserves a testcase.

```
void foo() {
   new Anonymous() {
        void plop() {
                SpringApplication.run(); // shall we raise an issue or not ?
         }
  }
}

File: java-checks/src/main/java/org/sonar/java/checks/UtilityClassWithPublicConstructorCheck.java [hunk 2]
```diff
@@ -79,6 +87,12 @@ private static boolean hasOnlyStaticMembers(ClassTree classTree) {
     return true;
   }
 
+  private static boolean isSpringBootApplication(ClassTree classTree) {
+    SpringBootVisitor springBootVisitor = new SpringBootVisitor();
```
benzonico: I am not a found of this subvisitor... but other solution are worst IMO : We could approach things by using leavenode on classTree and stack of boolean and visit Class and MethodInvocation. But i tend to think that'll make things more complicated. So maybe keep thing as is for now.

saberduck: OK

java-checks/src/main/java/org/sonar/java/checks/UtilityClassWithPublicConstructorCheck.java
saberduck: I think it is better not to raise the issue, it is quite unlikely that such construct will be used in a utility class, which is the target for the rule. Let's assume user knows what he is doing. I added the testcase for completeness sake.

java-checks/src/main/java/org/sonar/java/checks/UtilityClassWithPublicConstructorCheck.java
```diff
@@ -59,6 +62,23 @@ public void visitNode(Tree tree) {
...
+  private static boolean containsMainMethod(ClassTree classTree) {
+    return classTree.members().stream()
+      .filter(member -> member.is(Tree.Kind.METHOD))
+      .anyMatch(method -> isMainMethod((MethodTree) method));
+  }
+
+  private static boolean isMainMethod(MethodTree method) {
+    return MAIN.matches(method)
+      && method.returnType().symbolType().is("void")
+      && ModifiersUtils.hasModifier(method.modifiers(), Modifier.PUBLIC);
```
benzonico: What about static modifier ?

saberduck: It is checked before by `hasOnlyStaticMembers` , but having a second thought about it, it will be perhaps more consistent to check for it again. I'll fix that

saberduck: I found `MethodTreeImpl.isMainMethod` , so I replaced this function with that.

PR #1384, opened 2017-04-12
SONARJAVA-2134 Builder pattern and method returning field with the sa…

saberduck:
…me name should be excepted in S1845


java-checks/src/main/java/org/sonar/java/checks/MembersDifferOnlyByCapitalizationCheck.java
```diff
@@ -96,7 +96,32 @@ private static boolean isOverriding(Symbol symbol) {
...
-      return differentTypes(currentMember, knownMember) ? sameVisibilityNotPrivate(currentMember, knownMember) : !sameName(currentMember, knownMember);
+      return differentTypes(currentMember, knownMember) ? invalidMethodAndVariable(currentMember, knownMember) : !sameName(currentMember, knownMember);
+    }
+    return false;
+  }
+
+  private static boolean invalidMethodAndVariable(Symbol currentMember, Symbol knownMember) {
+    Symbol methodSymbol = currentMember.isMethodSymbol() ? currentMember : knownMember;
+    Symbol variableSymbol = methodSymbol == currentMember ? knownMember : currentMember;
+    return sameVisibilityNotPrivate(currentMember, knownMember)
```
m-g-sonar: I would first start the method by 
```
if(!sameVisibilityNotPrivate(currentMember, knownMember)) {
  return false;
}
```
as this operation does not requires knowing which one is the method or the variable.

java-checks/src/main/java/org/sonar/java/checks/MembersDifferOnlyByCapitalizationCheck.java
```diff
@@ -96,7 +96,32 @@ private static boolean isOverriding(Symbol symbol) {
...
+
+  private static boolean methodReturningVariableWithSameName(Symbol methodSymbol, Symbol variableSymbol) {
+    if (!sameName(variableSymbol, methodSymbol)) {
+      return false;
+    }
+    Tree declaration = methodSymbol.declaration();
+    if (declaration != null) {
+      ReturnVisitor returnVisitor = new ReturnVisitor(variableSymbol);
+      declaration.accept(returnVisitor);
+      return returnVisitor.singleReturnWithVariableSymbol();
```
m-g-sonar: you may want to exclude exploring annonymous classes in this visitor:

```
void foo() {
  MyObject o = new MyObject() {
    Object bar() {
      return foo;
    }
  };
}
```

java-checks/src/main/java/org/sonar/java/checks/MembersDifferOnlyByCapitalizationCheck.java
```diff
@@ -203,4 +228,28 @@ private static boolean isVariableToExtract(Symbol symbol) {
...
+    public void visitReturnStatement(ReturnStatementTree tree) {
+      returnCount++;
+      ExpressionTree returnExpression = tree.expression();
+      if (returnExpression != null && returnExpression.is(Tree.Kind.IDENTIFIER)) {
+        returnsVariable = ((IdentifierTree) returnExpression).symbol().equals(variableSymbol);
+      }
+    }
+
+    boolean singleReturnWithVariableSymbol() {
+      return returnCount == 1 && returnsVariable;
```
m-g-sonar: OK, so it implicitly means that you can have `n` other statements in the body of the method, as long as you return the variable, it's ok? Is it what you really want to achieve? (can't see a test case related to that).

If you want to target explicitly single-return-statement methods, it should be easier to simply look at the body, which has to be of size 1, with a single return statement returning the variable. And no need for a visitor for that.

java-checks/src/main/java/org/sonar/java/checks/MembersDifferOnlyByCapitalizationCheck.java
saberduck: Good point! I excluded anonymous classes and lambdas, and added corresponding tests.

java-checks/src/main/java/org/sonar/java/checks/MembersDifferOnlyByCapitalizationCheck.java
saberduck: Yes, there can be multiple statement as far as there is only one return statement returning the field. I added the test case, thanks for spotting that. However, I think the methods should not have more than 1 return statement, and I find using the visitor as easiest way to assure that.

m-g-sonar:
Update is OK regarding rule implementation, but there is still a missing case where you return a variable in the body of an anonymous class. From my point of view, even if such classes are skipped, the test file has to be used as documenting edge cases as well.

```
class A {
  Object foo;

  MyObject foo() {
    MyObject o = new MyObject() {
      Object bar() {
        return foo; // Compliant
      }
    };
    return o;
  }
```

PR #1387, opened 2017-04-13
SONARJAVA-2239 Fix failing on empty files

benzonico:


java-checks/src/main/java/org/sonar/java/checks/MissingNewLineAtEndOfFileCheck.java
```diff
@@ -31,10 +31,16 @@
 
   @Override
   public void scanFile(JavaFileScannerContext context) {
-    String content = context.getFileContent();
-    char lastChar = content.charAt(content.length() - 1);
-    if( lastChar != '\n' && lastChar != '\r') {
+    if (isEmptyOrNotEndingWithNewLine(context.getFileContent())) {
```
saberduck: It seems bit wasteful to read whole file into memory to just verify the last byte of it.  As far as I could check it seems that this is not cached in any way. Maybe you can read just the last byte and verify emptiness using `context.getFile()` ?

benzonico: Please have a deeper look at how content of file is accessed.

benzonico: As discussed we don't want to break abstraction provided by Scanner. Therefore we read all file through their abstraction.

PR #1395, opened 2017-04-19
SONARJAVA-2249 ignore classes without static members

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/UtilityClassWithPublicConstructorCheck.java
```diff
@@ -80,17 +80,22 @@ private static boolean extendsAnotherClassOrImplementsSerializable(ClassTree cla
   }
 
   private static boolean hasOnlyStaticMembers(ClassTree classTree) {
-    if (classTree.members().isEmpty()) {
+    List<Tree> members = classTree.members();
+    if (members.isEmpty() || noStaticMember(members)) {
```
saberduck: `members.isEmpty()` is redundant here, because `noStaticMember` will also be true when there are no classes.
Whole method can be perhaps replaced if logic is cleaned up. We need to test that when there is at least one member, which is not constructor nor empty statement, all of them are static.
```
private static boolean hasOnlyStaticMembers(ClassTree classTree) {
    List<Tree> members = classTree.members().stream()
      .filter(m -> !m.is(Tree.Kind.CONSTRUCTOR, Tree.Kind.EMPTY_STATEMENT))
    return !members.isEmpty() && members.stream().allMatch(UtilityClassWithPublicConstructorCheck::isStatic);
  }
```

m-g-sonar: I'm not really convinced. the `or` operator is short-circuit, so if members are empty, we won't call `noStaticMember`. I also would like to avoid collecting and creating a new list.

java-checks/src/main/java/org/sonar/java/checks/UtilityClassWithPublicConstructorCheck.java
```diff
@@ -81,15 +81,10 @@ private static boolean extendsAnotherClassOrImplementsSerializable(ClassTree cla
...
+    if (noStaticMember(members)) {
       return false;
     }
-    for (Tree member : members) {
-      if (!isConstructor(member) && !isStatic(member) && !member.is(Tree.Kind.EMPTY_STATEMENT)) {
-        return false;
-      }
-    }
-    return true;
+    return !members.stream().anyMatch(member -> !isConstructor(member) && !isStatic(member) && !member.is(Tree.Kind.EMPTY_STATEMENT));
```
saberduck: Sorry to bother again with this, but this is very negative  :-) How about 
```
return members.stream().allMatch(member -> isConstructor(member) || isStatic(member) || member.is(Tree.Kind.EMPTY_STATEMENT));
```

PR #1396, opened 2017-04-19
SONARJAVA-2193 S1121 should not raise issue for chained assignments

saberduck:


java-checks/src/main/java/org/sonar/java/checks/AssignmentInSubExpressionCheck.java
```diff
@@ -58,23 +59,27 @@ public void visitAnnotation(AnnotationTree annotationTree) {
 
   @Override
   public void visitLambdaExpression(LambdaExpressionTree lambdaExpressionTree) {
-    //skip lambda if body is an assignement
+    //skip lambda if body is an assignment
     if(!lambdaExpressionTree.body().is(Kind.ASSIGNMENT)) {
```
m-g-sonar: I know that it's not a part you worked on, but this part looks strange to me... I though that any variable used in the body of a lambda should be effectively `final`, as stated in [JLS §15.27.2](http://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.27.2) : 

> Any local variable, formal parameter, or exception parameter used but not declared in a lambda expression must either be declared final or be effectively final

So from my point of view, it means that you can not make any assignment of variable from outside the body of the lambda there. I doubt such code compile, then.

saberduck: The body can be assignment to a field, which is allowed to not be final even if used in lambda. But I agree this is not at all obvious, I had to google the expression to find old thread on ML http://sonarqube-archive.15.x6.nabble.com/Assignment-within-sub-expression-rule-for-lambdas-td5029472.html 

I made another commit making this test more explicit

m-g-sonar: :+1:

PR #1402, opened 2017-04-21
SONARJAVA-2180 S2150 update - add more methods with linear complexity

saberduck:


java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
```diff
@@ -0,0 +1,142 @@
...
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodInvocationTree mit = (MethodInvocationTree) tree;
+    if (!invocationInMethod(mit)) {
```
m-g-sonar: Maybe add a comment to make explicit that you do not consider invocation in constructors, initializers and static initializers.

Now, I would first check if the method is matching any of the matcher before checking if the invocation is part of a method. Intuitively, I have the feeling that we will then do that computation less time, as the probability that a method invocation matches any of the methods is way less than the the probability of being part of a constructor.

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
```diff
@@ -0,0 +1,142 @@
...
+  private static boolean invocationInMethod(MethodInvocationTree mit) {
+    Tree parent = mit.parent();
+    while (parent != null && !parent.is(Tree.Kind.METHOD)) {
+      parent = parent.parent();
+    }
+    return parent != null;
+  }
+
+  private static boolean isField(Symbol symbol) {
+    return symbol.isVariableSymbol() && symbol.owner().isTypeSymbol() && !"this".equals(symbol.name()) && !"super".equals(symbol.name());
```
m-g-sonar: Do you fully cover these cases in your test file? Can't see any `this.contains(...)`, or `super.get(...)`.

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
```diff
@@ -0,0 +1,142 @@
...
+    if (methodSelect.is(Tree.Kind.MEMBER_SELECT)) {
+      ExpressionTree methodSelectExpression = ((MemberSelectExpressionTree) methodSelect).expression();
+      if (methodSelectExpression.is(Tree.Kind.IDENTIFIER)) {
+        return ((IdentifierTree) methodSelectExpression).symbol();
+      }
+    }
+    return null;
+  }
+
+  private static Type actualType(Symbol symbol) {
```
m-g-sonar: I thought we agreed on the fact that we should only focus on invocation where we are 100% sure about the type. This won't work. Only looking at declarations does NOT guarantee that the actual type on which you will call a `get` or `remove` will be this one. If the variable is reassigned straight ahead, or forbidden methods are used on branches where the type has been changed, then we will raise FPs.

As a remainder, we agreed on targeting only the invocations on which the type is guaranteed. To do so, please consider types of the initializer **and all the re-assignments** that you can get from symbol usages. They all have to be of forbidden types in order to allow us to raise issues.

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
```diff
@@ -0,0 +1,142 @@
...
+
+  private static MethodMatcher collectionMethodMatcher() {
+    return MethodMatcher.create().typeDefinition(TypeCriteria.subtypeOf("java.util.Collection"));
+  }
+
+  private static final Map<MethodMatcher, Set<String>> matcherActualTypeMap;
+  static {
+    ImmutableMap.Builder<MethodMatcher, Set<String>> builder = ImmutableMap.builder();
+
+    MethodMatcher collectionContains = collectionMethodMatcher().name("contains").addParameter(JAVA_LANG_OBJECT);
```
saberduck: I want to target only that specific signature, why we should be less precise, when we can be exact?

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
saberduck: OK, check is done only when invocation is matched.

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
saberduck: Good point, I added the tests.

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
```diff
@@ -114,7 +114,16 @@ private static boolean isField(Symbol symbol) {
   }
 
   private static boolean matchesActualType(Symbol invocationTarget, Set<String> actualTypes) {
-    return actualTypes.contains(actualType(invocationTarget).fullyQualifiedName());
+    Type declaredType = invocationTarget.type();
+    if (actualTypes.contains(declaredType.fullyQualifiedName())) {
+      return true;
+    }
+    if (invocationTarget.isPrivate() || invocationTarget.isFinal()) {
```
m-g-sonar: This part worth a comment to explain _why_ we focus only on private or final fields (type cannot be changed from the outside)

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
```diff
@@ -129,14 +138,34 @@ private static Symbol invocationTarget(MethodInvocationTree mit) {
...
+  private static Stream<AssignmentExpressionTree> usageInAssignment(IdentifierTree usage) {
+    Tree parent = usage.parent();
+    while (parent != null && !parent.is(Tree.Kind.ASSIGNMENT)) {
+      parent = parent.parent();
+    }
+    if (parent == null) {
+      return Stream.empty();
+    }
+    AssignmentExpressionTree assignment = (AssignmentExpressionTree) parent;
+    if (ExpressionUtils.skipParentheses(assignment.variable()).equals(usage)) {
```
m-g-sonar: This won't work if assignment is done with member access: `this.myField = new CopyOnWriteArrayList<String>();`

You have to check that while going through parents, you are coming from the `variable()` side of the assignment, and not the `expression()` side. Checking only for `IdentifierTree` is not precise enough.

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
```diff
@@ -0,0 +1,173 @@
...
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+import javax.annotation.CheckForNull;
+
+import java.util.ArrayList;
```
SonarTech: MINOR Remove this unused import 'java.util.ArrayList'. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
```diff
@@ -0,0 +1,173 @@
...
+package org.sonar.java.checks;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+
+import org.sonar.check.Rule;
+import org.sonar.java.checks.helpers.MethodsHelper;
+import org.sonar.java.matcher.MethodMatcher;
+import org.sonar.java.matcher.TypeCriteria;
+import org.sonar.java.model.ExpressionUtils;
```
SonarTech: MINOR Remove this unused import 'org.sonar.java.model.ExpressionUtils'. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

SonarTech:
SonarQube analysis reported 2 issues
* ![MINOR][MINOR] 2 minor

[MINOR]: https://raw.githubusercontent.com/SonarCommunity/sonar-github/master/images/severity-minor.png 'Severity: MINOR'

java-checks/src/main/java/org/sonar/java/checks/CollectionMethodsWithLinearComplexityCheck.java
saberduck: Good point, I modified the loop to check if we are coming from LHS of assignment and it only considers if LHS is identifier or member select.

PR #1405, opened 2017-04-25
SONARJAVA-2008 method invocation type should not always be the target…

benzonico:


java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -219,7 +219,11 @@ public void visitMethodInvocation(MethodInvocationTree tree) {
         }
       );
       List<JavaType> typeParamTypes = getParameterTypes(tree.typeArguments());
-      inferReturnTypeFromInferedArgs(tree, methodEnv, argTypes, typeParamTypes, (JavaType) mit.symbolType(), typeSubstitution);
+      JavaType resultType = ((MethodJavaType) mit.symbol().type()).resultType;
+      if(resultType.symbol.owner == mit.symbol()) {
```
m-g-sonar: Certainly worth adding a comment, or extract the condition to a dedicated method with explicit name, such as: `returnsTypeVarFromMethod(resultType, mit.symbol())`

PR #1409, opened 2017-04-26
SONARJAVA-1901 Fix substitution in thrown types

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
```diff
@@ -1222,8 +1223,7 @@ public void infer_stream_types_on_chained_map() {
 
     IdentifierTree map = result.referenceTree(63, 8);
     MethodJavaType methodJavaType = (MethodJavaType) map.symbolType();
-    // expression type is correctly infered but method type is not recomputed and thus is still defered
-    assertThat(methodJavaType.resultType.isTagged(JavaType.DEFERRED)).isTrue();
+    assertThat(methodJavaType.resultType.isTagged(JavaType.DEFERRED)).isFalse();
```
benzonico: Maybe we can also have a positive assertion ? testing what is expected not only what is not expected ?

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -279,6 +279,8 @@ private void inferReturnTypeFromInferedArgs(MethodInvocationTree tree, Resolve.E
         MethodJavaType methodType = (MethodJavaType) resolution.type();
         if(!methodType.resultType.isTagged(JavaType.DEFERRED)) {
           registerType(tree, resolve.applySubstitution(methodType.resultType, typeSubstitution));
+          // change type associated to identifier as it may have been inferred in the process
```
benzonico: "in the process" -> no idea what this means.

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
```diff
@@ -279,6 +279,8 @@ private void inferReturnTypeFromInferedArgs(MethodInvocationTree tree, Resolve.E
...
+          registerType(identifier, resolution.type());
```
benzonico: shouldn't  resolution.type() be replaced by methodType ?

benzonico: Am I correct in saying that you ended up associating methodType to the identifier of the method in order to access it from check ? That is far from obvious and deserve at least a comment (I am even wondering about a fixme here, this sounds a bit like a hack to shove the methodType into the semantic API)

benzonico: Ok, so this was already done before, but please update tests properly !

java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
benzonico: This is actually _really_ important  : this is the kind of uncomplete udpate that can lead to unfailing tests after some changes. This test is now lacking assertions.

java-frontend/src/main/java/org/sonar/java/resolve/TypeAndReferenceSolver.java
m-g-sonar: Correct. Was already done for method resolution. I'll update the tests.

java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
m-g-sonar: agreed. I took a shortcut here.

java-frontend/src/test/files/sym/ThrownTypeVariables.java
```diff
@@ -1,7 +1,7 @@
 class A<T extends Throwable> {
   void foo() throws T {
-  }
 
+  }
```
benzonico: useless change.

java-frontend/src/test/files/sym/ThrownTypeVariables.java
```diff
@@ -24,4 +24,3 @@ void test(MyInterface<java.util.NoSuchElementException> s) {
     new B<String>().bar(java.io.IOException::new);
   }
 }
-
```
benzonico: useless change

PR #1410, opened 2017-04-27
SONARJAVA-2262 cache hashcode in immutable objects used in collections

benzonico:


java-frontend/src/main/java/org/sonar/java/se/ExplodedGraph.java
```diff
@@ -139,6 +142,7 @@ private Edge(Node child, Node parent) {
       Preconditions.checkState(!Objects.equals(child, parent));
```
m-g-sonar: Maybe worth removing the call to `Objects.equals()` here too, as both child and parent can not be null.

PR #1413, opened 2017-04-28
SONARJAVA-2208 Split printf check

benzonico:


java-checks/src/main/java/org/sonar/java/checks/PrintfFailCheck.java
```diff
@@ -0,0 +1,245 @@
...
+        return;
+      }
+      verifyParameters(mit, args, params);
+    }
+  }
+  protected void handleMessageFormat(MethodInvocationTree mit, String formatString, List<ExpressionTree> args) {
+    String newFormatString = cleanupDoubleQuote(formatString);
+    Set<Integer> indexes = getMessageFormatIndexes(newFormatString);
+    List<ExpressionTree> newArgs = args;
+    if (newArgs.size() == 1) {
```
saberduck: Same as above, this argument handling could be shared between the checks.

java-checks/src/main/java/org/sonar/java/checks/PrintfMisuseCheck.java
```diff
@@ -0,0 +1,202 @@
...
+        return;
+      }
+      unusedArgs.remove(args.get(index));
+    }
+    reportUnusedArgs(mit, args, unusedArgs);
+  }
+
+  private void reportUnusedArgs(MethodInvocationTree mit, List<ExpressionTree> args, List<ExpressionTree> unusedArgs) {
+    for (ExpressionTree unusedArg : unusedArgs) {
+      int i = args.indexOf(unusedArg);
```
saberduck: Not related to the split, although this loop should be done with classic for-loop, if index is needed.

java-checks/src/main/java/org/sonar/java/checks/AbstractPrintfChecker.java
```diff
@@ -0,0 +1,155 @@
...
+
+  protected static boolean isNewArrayWithInitializers(ExpressionTree expression) {
+    return expression.is(Tree.Kind.NEW_ARRAY) && ((NewArrayTree) expression).openBraceToken() != null;
+  }
+
+  protected static String cleanupDoubleQuote(String formatString) {
+    return formatString.replaceAll("\'\'", "");
+  }
+
+  protected boolean checkAndReport(MethodInvocationTree mit, boolean shouldReport, String message) {
```
saberduck: This is used only two times, once in each check. Imo, it will be more clear if inlined and removed from abstract parent.

java-checks/src/main/java/org/sonar/java/checks/PrintfMisuseCheck.java
benzonico: not relevant, args might not be in the correct order.

java-checks/src/main/java/org/sonar/java/checks/PrintfFailCheck.java
```diff
@@ -0,0 +1,245 @@
...
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    ExpressionTree formatStringTree;
+    List<ExpressionTree> args;
+    boolean isMessageFormat = MESSAGE_FORMAT.matches(mit);
+    if (isMessageFormat && !mit.symbol().isStatic()) {
+      // only consider the static method
+      return;
+    }
+    // Check type of first argument:
+    if (mit.arguments().get(0).symbolType().is("java.lang.String")) {
```
saberduck: something like this, also cleanupDoubleQuote and getting the array initializer in `handleMessageFormat ` could be moved into this shared class.

```
  static class FormatAndArgs {
    private ExpressionTree formatStringTree;
    private List<ExpressionTree> args;

    public FormatAndArgs(MethodInvocationTree mit) {
      if (mit.arguments().get(0).symbolType().is("java.lang.String")) {
        formatStringTree = mit.arguments().get(0);
        args = mit.arguments().subList(1, mit.arguments().size());
      } else {
        // format method with "Locale" first argument, skip that one.
        formatStringTree = mit.arguments().get(1);
        args = mit.arguments().subList(2, mit.arguments().size());
      }
    }
```

benzonico: This tends to be overcomplicated for what we want to achieve. And it will introduce too much coupling between the checks for now. I'll stick with the duplication.

PR #1414, opened 2017-04-28
SONARJAVA-2232 Rewrite S1994 to only target for-loop increments being outside updates

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/ForLoopIncrementAndUpdateCheck.java
```diff
@@ -48,41 +58,90 @@
...
 
-  private static Collection<Symbol> getUpdatedSymbols(ForStatementTree forStatementTree) {
+  private static String getMessage(Set<Symbol> updates) {
+    return String.format("Move the update of %s into this loop's update clause.", symbolNames(updates));
+  }
+
+  private static Collection<Symbol> symbolsFromConditionsNotUpdated(ForStatementTree forStatementTree) {
+    Collection<Symbol> conditionSymbols = getConditionSymbols(forStatementTree.condition());
+    Collection<Symbol> updateSymbols = getUpdatedSymbols(forStatementTree.update());
+    return conditionSymbols.stream().filter(s -> !updateSymbols.contains(s)).collect(Collectors.toSet());
```
benzonico: Use Collection.removeAll...

PR #1419, opened 2017-05-01
SONARJAVA-2155 fix quality flaw: simplify and increase coverage

saberduck:


java-checks/src/main/java/org/sonar/java/checks/UselessParenthesesCheck.java
```diff
@@ -34,57 +34,16 @@
...
-  };
-
-
   @Override
   public void visitNode(Tree tree) {
     ParenthesizedTree parenthesizedTree = (ParenthesizedTree) tree;
-    if (uselessParentheses(parenthesizedTree)) {
-      reportIssue(parenthesizedTree.openParenToken(),
+    if (parenthesizedTree.expression().is(Kind.PARENTHESIZED_EXPRESSION)) {
+      reportIssue(((ParenthesizedTree) parenthesizedTree.expression()).openParenToken(),
```
m-g-sonar: It's indeed easier like this.
I however wonder about redundant issues that we may raise, for example in the situation described below we will raise 2 issues:
```
  return (((a + b)));
```
It is expected, but I wonder if we want to improve it or not.

saberduck: I tend to think that it's better to keep it simple and not to add features in the RFF period. Also this way we are consistent with SonarJS and RSPEC. We can always change later.

PR #1425, opened 2017-05-03
SONARJAVA-2184 Fix fp on array access cast

benzonico:


java-checks/src/test/files/checks/RedundantTypeCastCheck.java
```diff
@@ -166,4 +166,12 @@ String foo() {
     }
   }
 
+  void fun() {
+    Object test = new Object[]{"1"};
+    String[] test2 = new Object[]{"1"};
```
m-g-sonar: Does not compile, but not really important.

PR #1428, opened 2017-05-04
SONARJAVA-2257 Use the symbol corresponding to the parameter causing exception for flow

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -56,7 +56,7 @@
 public class NullDereferenceCheck extends SECheck {
 
   private static final ExceptionalYieldChecker EXCEPTIONAL_YIELD_CHECKER = new ExceptionalYieldChecker(
-    "NullPointerException will be thrown when invoking method %s().");
+    "\"NullPointerException\" will be thrown when invoking method \"%s()\".");
```
saberduck: I think `OptionalGetBeforeIsPresentCheck` message ought to be updated too, if we want to have quotes around exception and method.

m-g-sonar: I updated the messages to be coherent with the "normal" null dereference issue message. 
I also wanted to add a dot at the end of the normal message, but it was too many updates not related to the fix.

If it's OK for you I'm going to update the messages for other rules using exception on method invocation messages, and add a second commit to this PR with the update, in order to not pollute the fix.

saberduck: Sure, that's fine

PR #1429, opened 2017-05-04
SONARJAVA-2265 unused private classes

benzonico:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateClassCheck.java
```diff
@@ -0,0 +1,48 @@
...
+    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.ANNOTATION_TYPE, Tree.Kind.ENUM);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if(!hasSemantic()) {
+      return;
+    }
+    ClassTree classTree = (ClassTree) tree;
+    if(classTree.symbol().isPrivate() && classTree.symbol().usages().isEmpty()) {
```
m-g-sonar: Extract `classTree.symbol()` to a dedicated variable rather than calling it thrice.

PR #1436, opened 2017-05-08
SONARJAVA-2049 Track symbols from binary symbolic values

saberduck:


java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -93,22 +93,25 @@ public Pop(ProgramState programState, List<SymbolicValue> result) {
...
     this.values = values;
     this.references = references;
     this.constraints = constraints;
     this.visitedPoints = visitedPoints;
     this.stack = stack;
     this.exitSymbolicValue = exitSymbolicValue;
+    this.lastAssociatedSymbol = lastAssociatedSymbol;
     constraintSize = 3;
   }
   private ProgramState(Symbol symbol, PMap<Symbol, SymbolicValue> values, PMap<SymbolicValue, Integer> references,
```
SonarTech: MAJOR Constructor has 8 parameters, which is greater than 7 authorized. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS00107)

SonarTech:
* ![MAJOR][MAJOR] 1 major

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -235,15 +245,33 @@ private boolean shouldTerminate(Set<LearnedConstraint> learnedConstraints) {
...
+              newTrackedSymbols = newTrackedSymbols.add(newSymbol);
+            }
+          }
+        }
+        return newTrackedSymbols;
+      }).orElse(trackedSymbols);
+    }
+
+    private boolean isLiteralAssignment(Tree tree) {
+      return tree.is(Tree.Kind.ASSIGNMENT) && ((AssignmentExpressionTree) tree).expression().is(Tree.Kind.NULL_LITERAL, Tree.Kind.BOOLEAN_LITERAL);
```
m-g-sonar: This method is annoying me a bit, we are manually excluding the three constant SVs that we add in the program state... So it makes a hidden link between this part of the code and the default empty PS. 

I'm not sure to understand why we need to exclude them as well. To not report meaningless flows such as "implies null is null" ?

saberduck: This is needed because literals reuse same symbolic value. This means that if we lookup what was last assigned symbol from program state, it can give us some completely unrelated symbol which happens to hold `null`, `true` or `false`. 

Another option would be to create new symbolic value for every literal, we have a ticket here https://jira.sonarsource.com/browse/SONARJAVA-2164

m-g-sonar: Okay cool, so it would be probably nice to reference the ticket here, and maybe add a comment in the ticket as well.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -196,20 +186,21 @@ public boolean equals(Object o) {
     return Objects.equals(values, that.values) &&
       Objects.equals(constraints, that.constraints) &&
       Objects.equals(exitSymbolicValue, that.exitSymbolicValue) &&
-      Objects.equals(stack, that.stack);
+      Objects.equals(stack, that.stack) &&
+      Objects.equals(lastAssociatedSymbol, that.lastAssociatedSymbol);
```
m-g-sonar: Why do you need it in the equality? `lastEvaluated` symbol was not used for it, and I fear that it will destroy caching of PS.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -81,36 +81,29 @@ public Pop(ProgramState programState, List<SymbolicValue> result) {
...
-
-  @Nullable
-  Symbol lastEvaluated;
-
   private final PStack<SymbolicValue> stack;
   private final PMap<SymbolicValue, Integer> references;
   private SymbolicValue exitSymbolicValue;
   final PMap<Symbol, SymbolicValue> values;
   final PMap<SymbolicValue, PMap<Class<? extends Constraint>, Constraint>> constraints;
+  final PMap<SymbolicValue, Symbol> lastAssociatedSymbol;
```
m-g-sonar: I think we should add the plural to this variable name: `lastAssociatedSymbols`.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -196,20 +186,21 @@ public boolean equals(Object o) {
...
-      Objects.equals(stack, that.stack);
+      Objects.equals(stack, that.stack) &&
+      Objects.equals(lastAssociatedSymbol, that.lastAssociatedSymbol);
   }
 
   @Override
   public int hashCode() {
     if (hashCode == 0) {
-      hashCode = Objects.hash(values, constraints, peekValue());
+      hashCode = Objects.hash(values, constraints, peekValue(), lastAssociatedSymbol);
```
m-g-sonar: Same remarks as for equals. I doubt about the interest of having the last associated symbols taken into account here. LastEvaluated symbol was not part of it.

java-frontend/src/test/java/org/sonar/java/se/ProgramStateDataProvider.java
```diff
@@ -64,8 +64,7 @@ public String stack() {
   }
 
   public String lastEvaluatedSymbol() {
```
m-g-sonar: Rename this method, even if it's only used for the viewer. let's keep it coherent.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
saberduck: I added reference to ticket to comment on line #252

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
saberduck: Removed `lastAssociatedSymbols` from equality

PR #1444, opened 2017-05-11
SONARJAVA-2219: Spring Annotations - adding missing one

aepfli:
@m-g-sonar sorry fr so many pull requests, i hope this is now done correctly

Hey guys!

With my last commit i think we did miss on the `@Value` annotation. I added it with this Pull-Request, i think that is it, i checked the JSR-250 and spring annotation, and did not find another field-based injection annotation!

pasted image at 2017_05_10 11_10 pm

Br simon

- [x] ITs should pass : To run ITs locally, checkout the [README](https://github.com/SonarSource/sonar-java/blob/master/README.md) of the project. - did not check

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentWithNonAutowiredMembersCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentWithNonAutowiredMembersCheck.java
@@ -47,14 +47,15 @@ public class SpringComponentWithNonAutowiredMembersCheck extends IssuableSubscri
         .map(m -> (VariableTree) m)
         .filter(v -> !v.symbol().isStatic())
         .filter(v -> !isSpringInjectionAnnotated(v.symbol().metadata()))
-        .forEach(v -> reportIssue(v.simpleName(), "Annotate this member with \"@Autowired\", \"@Resource\", or \"@Inject\", or remove it."));
+        .forEach(v -> reportIssue(v.simpleName(), "Annotate this member with \"@Autowired\", \"@Resource\", \"@Inject\", or \"@Value\", or remove it."));
     }
   }
 
   private static boolean isSpringInjectionAnnotated(SymbolMetadata metadata) {
     return metadata.isAnnotatedWith("org.springframework.beans.factory.annotation.Autowired")
       || metadata.isAnnotatedWith("javax.inject.Inject")
-      || metadata.isAnnotatedWith("javax.annotation.Resource");
+      || metadata.isAnnotatedWith("javax.annotation.Resource")
+      || metadata.isAnnotatedWith("org.springframework.beans.factory.annotation.Value");
   }
 
   private static boolean isSpringComponent(SymbolMetadata clazzMeta) {
--- a/java-checks/src/test/files/checks/spring/SpringComponentWithNonAutowiredMembersCheck.java
+++ b/java-checks/src/test/files/checks/spring/SpringComponentWithNonAutowiredMembersCheck.java
@@ -5,15 +5,16 @@ import org.springframework.stereotype.Service;
 import org.springframework.stereotype.Repository;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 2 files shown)


m-g-sonar:
Thanks @aepfli for reopening it with a clear history. It seems that we have an issue internally with PR analysis external from our organization. We are looking at the issue.

Regarding your change and adding `@Value`, can you provide a few more information, such as links where to find description of the annotation?

Thx!

aepfli:
Well i hope spring doc is enough documentation. Based on the package summary we have no covered all field based annotations from spring

- http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/beans/factory/annotation/Value.html
- http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/beans/factory/annotation/package-summary.html

I mean this annotation is somehow not a Dependency Injection, as it just injects a value, which you can define via Value expression - but it is a commonly used Annotation!

IMHO:
This rule also does not cover all cases - what if the field is not annotated but the getters and setters. Checking for Getters and Setters of a field would be the next iteration to improve that rule

aepfli:
@m-g-sonar is this information sufficient enough?

SideNote: if you guys also think my Imho needs a quicker fix, if you point me into the right direction (if a field has a getter oder setter), i surely got a jump start on that part :D

m-g-sonar:
Hey @aepfli, its looks good to me with the update of the rule. I'm going to update the rule description to include `@Value`.

Your remark is indeed correct regarding injection which could be done as well using setter/getter. This part is not covered by the rule... As far as I can tell, injection can also be performed by annotating the constructor (case also not covered).

So let's proceed in 2 phases:
1. Merge your current contribution
2. If motivated, open a new PR with handling of setters/getters/constructors.

Like last time, I'm going to open a new PR with your change to trigger our internal QA. If everything goes fine, I'm going to merge it with current changes.

Once done, I'll invite you to try to update the rule to cover cases of constructors and getters/setters if you still want to give it a try.

The rule will requires semantic, so add a semantic check at the beginning of the method (see example [here](https://github.com/SonarSource/sonar-java/blob/4.9.0.9858/java-checks/src/main/java/org/sonar/java/checks/CatchRethrowingCheck.java#L43)). Then, as this probably won't be trivial, here are my guidelines. There is multiple approaches to handle this, but here is the one I suggest:

- From the field, you can jump to all its usages (`symbol.usages()`). It will return a list of `identiferTree`.
- For each of these identifiers, check that it is part of an assignments (by looping on `tree.parent()` and checking the parent kind). If it is, check that you are coming (traversing the tree up) from the variable side, and not the expression side.
- If it is part of an assignment, check the enclosing method of the identifier. If the method is annotated, then consider the field as being OK, and let assume that it is correctly injected.

I recommend taking shortcuts here, as we will always prefer having False Negative (FN) than False Positive (FP). It's going to be a pain if you want to be really sure that the given field is really assigned correctly in the constructor or within the setter, and not only in a given path. Trust me, you don't want to explore the control flow graph of each method for that.

HTH,
Michael

m-g-sonar:
QA green on sister-PR. I'll proceed with the merge of this PR and let you open a new one if you want to continue improving the rule.

sudr:
I'm motivated to look into supporting this rule for constructor injection. Should I create a new SonarSource issue or how should we go about this?

aepfli:
@sudr are you planning to do something here? if so let me know, we encountered multiple missing features, which i would love to discuss, if you do not have time, this will probably my yearly goal for 2018.

Things which are encountered

* Fields set via spring configuration -> how to handle those? i mean they could be intentionally optional, but still a field
* Fields which are set in a postconstruct -> some fields will be set in a post construct, eg. for initialising a map with some values.

benzonico:
Hey @sudr @aepfli , I tend to think those constructive conversations should probably happen at sonarqube@googlegroups.com for the whole community to weigh in eventually. 
A closed PR might not be the most suitable places to discuss. 

Thanks !

sudr:
Agreed. Here is the link to the google groups thread for the specific need I had:
https://groups.google.com/d/msg/sonarqube/fHn3F5WtN5c/hbKKDIzSAgAJ

For other requirements @aepfli please start a new thread.

PR #1448, opened 2017-05-15
SONARJAVA-2275 Handle flow messages "symbol 'can'/'can not' be null"

m-g-sonar:

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
@@ -314,11 +314,14 @@ public class FlowComputation {
       PMap<Class<? extends Constraint>, Constraint> allConstraints = node.programState.getConstraints(learnedAssociation.symbolicValue());
       Collection<Constraint> constraints = filterByDomains(allConstraints);
       for (Constraint constraint: constraints) {
+        String symbolName = learnedAssociation.symbol().name();
         String msg;
-        if (assigningNullFromTernary(constraint, node)) {
-          msg = String.format(IMPLIES_CAN_BE_NULL_MSG, learnedAssociation.symbol().name());
+        if (ObjectConstraint.NULL == constraint && assigningNullFromTernary(node)) {
+          msg = String.format(IMPLIES_CAN_BE_NULL_MSG, symbolName);
+        } else if (assigningFromMethodInvocation(node) && assignedFromYieldWithUncertainResult(constraint, node)) {
+          msg = String.format("Implies '%s' can be %s.", symbolName, constraint.valueAsString());
         } else {
-          msg = String.format("'%s' is assigned %s.", learnedAssociation.symbol().name(), constraint.valueAsString());
+          msg = String.format("'%s' is assigned %s.", symbolName, constraint.valueAsString());
         }
         flowBuilder.add(location(node, msg));
       }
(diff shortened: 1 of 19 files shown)


java-frontend/src/test/files/se/FlowMessagesBranch.java
```diff
@@ -0,0 +1,26 @@
...
+      || null == b) {   // flow@npe2 {{Implies 'b' can be null.}}
+      // Noncompliant@+1 [[flows=npe1]] {{A "NullPointerException" could be thrown; "a" is nullable here.}}
+      a.toString();     // flow@npe1 {{'a' is dereferenced.}}
+      // Noncompliant@+1 [[flows=npe2]] {{A "NullPointerException" could be thrown; "b" is nullable here.}}
+      b.toString();     // flow@npe2 {{'b' is dereferenced.}}
+    }
+  }
+
+  void bar(boolean a) {
+    if (a == true) { // flow@cot {{Implies 'a' is true.}} flow@cot {{Implies 'a' is true.}}
```
m-g-sonar: this is strange. I'll have another look.

m-g-sonar: Ok, this is still strange, but not related to my ticket and was already present.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -227,6 +234,10 @@ public int hashCode() {
 
       boolean endOfPath = visitedAllParents(edge) || shouldTerminate(learnedConstraints);
 
+      if (endOfPath) {
+        flowBuilder.addAll(flowForNullableMethodParameters(edge.parent));
```
saberduck: I think this should be done outside of the path exploration as a separate step in `FlowComputation.run` method, because this is not dependent on the specific path and should be added to all flows.

java-frontend/src/test/files/se/FlowMessagesBranch.java
saberduck: Have you created ticket for this? It seems to be related to tracking relational symbolic values I did recently.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -247,6 +258,40 @@ private boolean isConstraintFromObjectDomain(@Nullable Constraint constraint) {
       return constraint instanceof ObjectConstraint;
     }
 
+    private List<Location> flowForNullableMethodParameters(ExplodedGraph.Node node) {
+      if (node.parent() != null || !domains.contains(ObjectConstraint.class)) {
+        return ImmutableList.of();
+      }
+      ImmutableList.Builder<JavaFileScannerContext.Location> flowBuilder = ImmutableList.builder();
```
m-g-sonar: Agreed, but I added it by mistake (see line [261](https://github.com/SonarSource/sonar-java/pull/1448/files#diff-ce6001fad62a230360636a1563f880ecR261)). I'm going to remove it. Don't want to add changes everywhere.

java-frontend/src/test/files/se/FlowMessagesBranch.java
m-g-sonar: Created [SONARJAVA-2300](https://jira.sonarsource.com/browse/SONARJAVA-2300)!

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
m-g-sonar: Does not work, I need to be on the last node and still things to learn.

PR #1450, opened 2017-05-15
SONARJAVA-2271 Remove redundant message from object constraint

saberduck:


saberduck:
Note that branch is based on SONARJAVA-2049_2 branch, I will rebase on master afterwards.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -229,6 +234,19 @@ public int hashCode() {
...
+      Map<SymbolicValue, Long> constraintsBySV = learnedConstraints.stream()
+        .filter(lc -> lc.constraint() != null)
+        .collect(Collectors.groupingBy(LearnedConstraint::symbolicValue, Collectors.counting()));
+      return learnedConstraints.stream()
+        .flatMap(lc -> isConstraintFromObjectDomain(lc.constraint()) && constraintsBySV.get(lc.symbolicValue()) > 1 ? Stream.empty() : Stream.of(lc))
+        .collect(Collectors.toSet());
+    }
+
+    private boolean isConstraintFromObjectDomain(@Nullable Constraint constraint) {
+      return constraint != null && constraint.getClass().isAssignableFrom(ObjectConstraint.class);
```
m-g-sonar: why not simply use `constraint instanceof ObjectConstraint` ? if null, it will return `false`.

PR #1454, opened 2017-05-16
SONARJAVA-2296 Fix access with 'this' when building issues messages

m-g-sonar:


java-frontend/src/test/files/se/NullDereferenceCheck.java
```diff
@@ -761,3 +761,12 @@ void bar(A a) {
     }
   }
 }
+
+class finalFieldSetToNull {
+  final Object field = null;
+
+  void foo() {
+    // Noncompliant@+1 {{A "NullPointerException" could be thrown; "field" is nullable here.}}
+    this.field.toString();
```
saberduck: I think we should also test flow messages here, as they are broken in the reported case.

PR #1457, opened 2017-05-17
Fix ruling and template activation

benzonico:


its/ruling/src/test/java/org/sonar/java/it/JavaRulingTest.java
```diff
@@ -291,7 +294,7 @@ private static void instantiateTemplateRule(String ruleTemplateKey, String insta
     }
     activatedRuleKeys.add(instantiationKey);
     SonarClient sonarClient = orchestrator.getServer().adminWsClient();
-    sonarClient.post("/api/rules/create", ImmutableMap.<String, Object>builder()
+    String post2 = sonarClient.post("/api/rules/create", ImmutableMap.<String, Object>builder()
```
m-g-sonar: Useless variable `post2`

PR #1459, opened 2017-05-17
SONARJAVA-2179 Fix handling of final fields when reporting flow

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -449,17 +449,50 @@ private boolean hasConstraintForDomain(LearnedConstraint lc, Class<? extends Con
...
         return ImmutableList.of(location(parent, String.format("Constructor implies '%s'.", constraint.valueAsString())));
       }
-      String name = SyntaxTreeNameFinder.getName(nodeTree);
-      if (name == null) {
-        return ImmutableList.of();
-      }
       String msg;
-      if (ObjectConstraint.NULL == constraint && isNullCheck(nodeTree)) {
-        msg = String.format(IMPLIES_CAN_BE_NULL_MSG, name);
+      Tree reportTree = parent.programPoint.syntaxTree();
```
saberduck: This variable has same value as `nodeTree` already present. I think it would be better to use that one to test if the tree is final field tree and to directly return instead of nesting if-else constructs, otherwise it's becoming bit messy.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -412,7 +412,7 @@ public void accept(SymbolicValue symbolicValue, PMap<Class<? extends Constraint>
   public ProgramState resetFieldValues(ConstraintManager constraintManager) {
     final List<VariableTree> variableTrees = new ArrayList<>();
     values.forEach((symbol, symbolicValue) -> {
-      if (isField(symbol)) {
+      if (isField(symbol) && !symbol.isFinal()) {
```
saberduck: :+1: I am bit surprised this didn't lead to more issues being detected in ruling

PR #1465, opened 2017-05-18
SONARJAVA-2282 Exclude field with @EJB annotation

benzonico:


java-checks/src/main/java/org/sonar/java/checks/ServletInstanceFieldCheck.java
```diff
@@ -71,7 +71,7 @@ public void visitNode(Tree tree) {
   }
 
   private static boolean isExcluded(VariableTree variable) {
-    return isStaticOrFinal(variable) || variable.symbol().metadata().isAnnotatedWith("javax.inject.Inject");
+    return isStaticOrFinal(variable) || variable.symbol().metadata().isAnnotatedWith("javax.inject.Inject") || variable.symbol().metadata().isAnnotatedWith("javax.ejb.EJB");
```
m-g-sonar: if injected 2 times, calling two times `variable.symbol().metadata()` is not efficient. Please use a dedicated intermediate variable.

PR #1476, opened 2017-05-22
SONARJAVA-2309 Symbolic values for resources should be added transiti…

saberduck:
…vely

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
@@ -51,6 +51,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class ProgramState {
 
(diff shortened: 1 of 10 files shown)


java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -259,6 +260,22 @@ public String toString() {
     return "{" + values.toString() + "}  {" + constraints.toString() + "}" + " { " + stack.toString() + " }";
   }
 
+  public ProgramState addConstraintTransitively(SymbolicValue symbolicValue, Constraint constraint) {
+    final ProgramState[] ps = {addConstraint(symbolicValue, constraint)};
+    knownRelations()
+      .filter(rsv -> rsv.isEquality() && (rsv.getLeftOp() == symbolicValue || rsv.getRightOp() == symbolicValue))
+      .map(rsv -> rsv.getLeftOp() == symbolicValue ? rsv.getRightOp() : rsv.getLeftOp())
+      .forEach(sv -> ps[0] = ps[0].addConstraint(sv, constraint));
```
benzonico: I am pretty sure what we are looking for here is a collector acting as reduction : https://docs.oracle.com/javase/tutorial/collections/streams/reduction.html

While this final array works it is not really nice.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -259,6 +260,22 @@ public String toString() {
     return "{" + values.toString() + "}  {" + constraints.toString() + "}" + " { " + stack.toString() + " }";
   }
 
+  public ProgramState addConstraintTransitively(SymbolicValue symbolicValue, Constraint constraint) {
+    final ProgramState[] ps = {addConstraint(symbolicValue, constraint)};
```
benzonico: For sake of clarity i would separate array initializer and method call.

java-frontend/src/main/java/org/sonar/java/se/symbolicvalues/RelationalSymbolicValue.java
```diff
@@ -430,7 +430,7 @@ private boolean isCommutative() {
     }
   }
 
-  private boolean isEquality() {
+  public boolean isEquality() {
```
benzonico: I would add a dedicated test if this is made public.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
saberduck: Unfortunately, stream API doesn't provide fold-left operation, and reduce method would require combiner to allow merging of two intermediate ProgramStates. But I understand that abusing arrays for this is bit of a hack, so I rewrote it by having values collected.

PR #1481, opened 2017-05-23
SONARJAVA-2274 Remove "is assigned to" and use "is" / "can be" flow messages

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -181,20 +183,66 @@ private FlowComputation(Set<SymbolicValue> symbolicValues, Predicate<Constraint>
...
   }
 
-  Stream<ExecutionPath> startPath(ExplodedGraph.Edge edge, PSet<Symbol> trackedSymbols) {
-    return new ExecutionPath(null, PCollections.emptySet(), trackedSymbols, ImmutableList.of(), false).addEdge(edge);
+  Stream<ExecutionPath> startPath(ExplodedGraph.Edge edge, PSet<Symbol> trackedSymbols, SameConstraints sameConstraints) {
+    return new ExecutionPath(null, PCollections.emptySet(), trackedSymbols, sameConstraints, ImmutableList.of(), false).addEdge(edge);
+  }
+
+  private static class SameConstraints {
+    private final Map<Symbol, Boolean> sameConstraintsBySymbol;
```
saberduck: Instead of `Map<Symbol, Boolean>` you can use `PSet<Symbol>`  where only symbols with same constraints are kept. Then you can test with `contains`. Also PSet is persistent, so you can share it without copying.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -230,6 +278,7 @@ public int hashCode() {
       flowBuilder.addAll(laFlow);
 
       PSet<Symbol> newTrackSymbols = newTrackedSymbols(edge);
+      SameConstraints newSameConstaints = new SameConstraints(sameConstraints, newTrackSymbols, domains);
```
saberduck: Typo in variable name.  You can test if there is any change in `trackedSymbols` and create instance only if necessary. It will be often the case that symbols won't be changed.

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -181,20 +183,66 @@ private FlowComputation(Set<SymbolicValue> symbolicValues, Predicate<Constraint>
...
+      SymbolicValue sv = programState.getValue(symbol);
+      if (sv == null) {
+        return false;
+      }
+      Constraint constraint = programState.getConstraint(sv, domain);
+      if (constraint == null) {
+        return false;
+      }
+      Collection<Node> siblingNodes = currentNode.siblings();
+      return siblingNodes.isEmpty() || siblingNodes.stream()
```
saberduck: `allMatch` returns `true` for empty stream, so separate test for emptiness is not necessary

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
m-g-sonar: Thanks for the idea. I tried with a `PMap` and lost myself in some details when I could have simply use a `PSet`.

PR #1494, opened 2017-05-29
SONARJAVA-2111 report multiple flows for SQ >= 6.4

saberduck:


java-frontend/src/main/java/org/sonar/java/model/DefaultJavaFileScannerContext.java
```diff
@@ -123,8 +123,16 @@ public void reportIssue(JavaCheck javaCheck, Tree syntaxNode, String message, Li
 
   @Override
   public void reportIssueWithFlow(JavaCheck javaCheck, Tree syntaxNode, String message, Iterable<List<Location>> flows, @Nullable Integer cost) {
-    // FIXME SONARJAVA-2111 all flows should be reported for SE checks
-    Iterable<List<Location>> reportedFlows = javaCheck instanceof SECheck ? Iterables.limit(flows, 1) : flows;
+    if(sonarComponents == null) {
```
m-g-sonar: Are you sure we need that condition? Are you able to cover it by test case?

The `null`-test was not present before and I can't see any test related to that.

saberduck: Indeed, field itself is not marked as @Nullable, I removed the null-check

PR #1495, opened 2017-05-29
SONARJAVA-2318 + SONARJAVA-2320 Fix handling of parametrized methods and resolution of constructors

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/SubClassStaticReferenceCheck.java
```diff
@@ -99,7 +99,7 @@ public void visitClass(ClassTree tree) {
     @Override
     public void visitIdentifier(IdentifierTree tree) {
       Type type = tree.symbolType();
-      if (!classType.equals(type) && type.isSubtypeOf(classType)) {
+      if (!classType.is(type.fullyQualifiedName()) && type.isSubtypeOf(classType)) {
```
benzonico: Given that classType is (surprise) the type of a class ( https://github.com/SonarSource/sonar-java/pull/1495/commits/a93673052cd6a7d49956b269e986800ec78024bb#diff-5a39a775425a584a6a9e7a42296dadb0R50 )  I struggle to see the point of that fix :  Do you have a failing test case ?

java-frontend/src/test/java/org/sonar/java/resolve/MethodJavaTypeTest.java
```diff
@@ -24,61 +24,116 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class MethodJavaTypeTest {
 
-  private static final Symbols SYMBOLS = new Symbols(new BytecodeCompleter(new SquidClassLoader(Collections.emptyList()), new ParametrizedTypeCache()));
+  private static final ParametrizedTypeCache parametrizedTypeCache = new ParametrizedTypeCache();
+  private static final BytecodeCompleter bytecodeCompleter = new BytecodeCompleter(new SquidClassLoader(Collections.emptyList()), parametrizedTypeCache);
+  private static final Symbols SYMBOLS = new Symbols(bytecodeCompleter);
+  // only used to correctly initialize the parameterized type cache with a type substitution solver
+  private static final TypeSubstitutionSolver typeSubstitutionSolver = new TypeSubstitutionSolver(parametrizedTypeCache, SYMBOLS);
```
benzonico: IMO, drop the comment and suffix the variable with "stub"

java-checks/src/main/java/org/sonar/java/checks/SubClassStaticReferenceCheck.java
m-g-sonar: Yep, failing test cases on current test file if I don't do the change: https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/test/files/checks/SubClassStaticReferenceCheck.java#L10

With the changes that I've done, using equals can not work anymore, as we will have "classType" compared in case of constructors to the "MethodJavaType" associated with the identifier, which will never be equals, even if the constructor does return the same type. However, the "is" method handle the case properly, as we will use the symbol fully qualified name, which is the correct one.

benzonico: This bothers me as it sounds rather like working around the API rather than adressing the real issue. As you tried to improve things with method type, then the is method should most probably be improved to handle those cases... But I do wonder if we actually really want "is" to answer true when we compare classtype and methodtype. 

IMO best course of action would be to have an explicit instanceof here. As you dealt with in other checks

PR #1500, opened 2017-06-02
SONARJAVA-2334 Support Java 9 bytecode for semantic resolution

saberduck:


Godin:
:thumbsup: this fixes a lot of test failures when they executed against JDK 9 EA b171

PR #1506, opened 2017-06-09
SONARJAVA-2337 Update highlighting to cover new restricted keywords

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/ast/visitors/SyntaxHighlighterVisitor.java
```diff
@@ -43,15 +45,18 @@
   private final SonarComponents sonarComponents;
   private final Map<Tree.Kind, TypeOfText> typesByKind;
   private final Set<String> keywords;
+  private final Set<String> restrictedKeywords;
+  private static final String INTERFACE_KEYWORD_VALUE = JavaKeyword.INTERFACE.getValue();
```
saberduck: I found only one usage of this constant, imo `JavaKeyword.INTERFACE` could be used directly.

java-frontend/src/main/java/org/sonar/java/ast/visitors/SyntaxHighlighterVisitor.java
```diff
@@ -43,15 +45,18 @@
...
+  private static final String TRANSITIVE_RESTRICTED_KEYWORD_VALUE = JavaRestrictedKeyword.TRANSITIVE.getValue();
```
saberduck: Same as above, only one usage for a constant.

java-frontend/src/test/java/org/sonar/java/JavaSquidTest.java
```diff
@@ -21,6 +21,7 @@
 
 import com.google.common.collect.Lists;
 import com.google.common.io.Files;
+
```
saberduck: why the whitespace here?

PR #1514, opened 2017-06-15
SONARJAVA-2329 Ignore Non-Null constraints when reporting x-proc issues

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/ExceptionalYieldChecker.java
```diff
@@ -98,18 +101,37 @@ private void reportIssue(ExplodedGraph.Node node, ExceptionalCheckBasedYield yie
...
+        argSymbolicValues.add(sv);
+      } else {
+        PMap<Class<? extends Constraint>, Constraint> constraints = programState.getConstraints(sv);
+        if (constraints == null || (domainsFromConstraints(constraints).size() == 1 && constraints.get(ObjectConstraint.class) == ObjectConstraint.NOT_NULL)) {
+          // skip SV without constraints or only having a single NOT_NULL constraint
+          continue;
+        }
+        argSymbolicValues.add(sv);
+      }
+      if (symbol != null) {
```
saberduck: Symbols should be tracked iff symbolic value is being tracked, this will cause tracking of symbols we are not interested in. I am bit surprised that this wasn't in the tests already, add test if this is missing.

m-g-sonar: That's a wrong assumption. The "continue" statement discard addition of the symbol if the SV is not tracked.

PR #1517, opened 2017-06-16
Do not create LearnedConstraint with null constraint on binary SV

saberduck:


java-frontend/src/main/java/org/sonar/java/se/LearnedConstraint.java
```diff
@@ -22,16 +22,11 @@
...
-
 public class LearnedConstraint {
   final SymbolicValue sv;
-
-  @Nullable
   final Constraint constraint;
 
-  public LearnedConstraint(SymbolicValue sv, @Nullable Constraint constraint) {
+  public LearnedConstraint(SymbolicValue sv, Constraint constraint) {
     this.sv = sv;
```
m-g-sonar: Maybe adding a `Preconditions.checkNotNull(constraint)` test in here, as we are changing the behavior, to be sure that no further null constraints are going to be created.

PR #1520, opened 2017-06-19
Encapsulate constraints in dedicated class

saberduck:


java-frontend/src/main/java/org/sonar/java/se/checks/ExceptionalYieldChecker.java
```diff
@@ -124,30 +123,24 @@ private void reportIssue(ExplodedGraph.Node node, ExceptionalCheckBasedYield yie
...
-    PMap<Class<? extends Constraint>, Constraint> constraints = ps.getConstraints(sv);
+    ConstraintsByDomain constraints = ps.getConstraints(sv);
     return constraints != null && !hasOnlyNonNullConstraint(constraints);
   }
 
-  private static boolean hasOnlyNonNullConstraint(PMap<Class<? extends Constraint>, Constraint> constraints) {
-    return domainsFromConstraints(constraints).size() == 1 && constraints.get(ObjectConstraint.class) == ObjectConstraint.NOT_NULL;
+  private static boolean hasOnlyNonNullConstraint(ConstraintsByDomain constraints) {
+    List<Class<? extends Constraint>> domains = constraints.domains().collect(Collectors.toList());
+    return domains.size() == 1 && constraints.get(ObjectConstraint.class) == ObjectConstraint.NOT_NULL;
```
m-g-sonar: why not directly `constraints.domains().count() == 1` ?

java-frontend/src/test/java/org/sonar/java/se/constraint/ConstraintsByDomainTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.junit.Test;
+
+import java.util.function.BiConsumer;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ConstraintsByDomainTest {
+
+  @Test
+  public void test() throws Exception {
```
m-g-sonar: What are you testing here ?! Please change test name.

java-frontend/src/main/java/org/sonar/java/se/constraint/ConstraintsByDomain.java
```diff
@@ -0,0 +1,95 @@
...
+  private ConstraintsByDomain(PMap<Class<? extends Constraint>, Constraint> constraintPMap) {
+    this.constraintPMap = constraintPMap;
+  }
+
+  public static ConstraintsByDomain empty() {
+    return EMPTY;
+  }
+
+  public ConstraintsByDomain remove(Class<? extends Constraint> domain) {
+    PMap<Class<? extends Constraint>, Constraint> remove = constraintPMap.remove(domain);
```
m-g-sonar: Similarly to the `put` method, I think it would be more judicious to check if after the remove operation the new PMap is similar to the previous one and returning it directly if it's the case, rather than systematically create a new on. That would correspond to a case of removing a domain which is not part of the current domains.

PR #1527, opened 2017-06-21
SONARJAVA-2355 S2095 should support javax.sql.DataSource and java.sql…

saberduck:
….PreparedStatement

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
@@ -88,8 +88,6 @@ public class UnclosedResourcesCheck extends SECheck {
   private static final String JAVA_IO_CLOSEABLE = "java.io.Closeable";
   private static final String JAVA_SQL_STATEMENT = "java.sql.Statement";
   private static final String JAVA_SQL_CONNECTION = "java.sql.Connection";
-  private static final String JAVAX_SQL_DATA_SOURCE = "javax.sql.DataSource";
-  private static final String JAVA_SQL_PREPARED_STATEMENT = "java.sql.PreparedStatement";
 
   private static final MethodMatcherCollection JDBC_RESOURCE_CREATIONS = MethodMatcherCollection.create(
     MethodMatcher.create().typeDefinition(JAVA_SQL_CONNECTION).name("createStatement").withAnyParameters(),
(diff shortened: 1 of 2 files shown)


java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -86,7 +87,21 @@ public String valueAsString() {
   private static final String JAVA_IO_AUTO_CLOSEABLE = "java.lang.AutoCloseable";
   private static final String JAVA_IO_CLOSEABLE = "java.io.Closeable";
   private static final String JAVA_SQL_STATEMENT = "java.sql.Statement";
-  private static final String[] JDBC_RESOURCE_CREATIONS = {"java.sql.Connection", JAVA_SQL_STATEMENT};
+  private static final String JAVA_SQL_CONNECTION = "java.sql.Connection";
+  private static final String JAVAX_SQL_DATA_SOURCE = "javax.sql.DataSource";
```
m-g-sonar: Not sure you need dedicated constants for `javax.sql.DataSource` and `java.sql.PreparedStatement`, as they are used only once.

java-frontend/src/test/files/se/JdbcResourcesTestFile.java
```diff
@@ -108,3 +111,32 @@ void foo() {
   abstract Connection getConnection();
 }
 
+class DataSourceTest {
+
+  void test(DataSource dataSource) throws SQLException {
```
m-g-sonar: I would prefer a compiling code here:

```
  void test(DataSource dataSource) {
    Connection connection = null;
    try {
      connection = dataSource.getConnection(); // Noncompliant
    } catch (SQLException e) {
      // do something
    }
  }
```

java-frontend/src/test/files/se/JdbcResourcesTestFile.java
```diff
@@ -108,3 +111,32 @@ void foo() {
...
+      connection = dataSource.getConnection();
+    } finally {
+      if (connection != null) {
+        connection.close();
+      }
+    }
+  }
+}
+
+class S2095FP {
```
m-g-sonar: This is not a FP anymore.

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -404,24 +419,20 @@ public void visitMethodInvocation(MethodInvocationTree syntaxNode) {
       }
       if (syntaxNode.methodSelect().is(Tree.Kind.MEMBER_SELECT) && needsClosing(syntaxNode.symbolType())) {
         final ExpressionTree targetExpression = ((MemberSelectExpressionTree) syntaxNode.methodSelect()).expression();
+        boolean jdbcResourceCreation = isJdbcResourceCreation(syntaxNode);
```
m-g-sonar: No need for a dedicated method anymore, I would inline it as it's the only remaining usage. I also think that calling `JDBC_RESOURCE_CREATIONS.anyMatch(syntaxNode)` is explicit enough with the assignment.

PR #1528, opened 2017-06-21
SONARJAVA-2228 Fix S3655 to handle 'Optional.of(...)' and 'Optional.ofNullable(...)'

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/OptionalGetBeforeIsPresentCheck.java
```diff
@@ -71,10 +75,33 @@ public ProgramState checkPostStatement(CheckerContext context, Tree syntaxNode)
   }
 
   private static List<ProgramState> setNotPresentConstraint(CheckerContext context, Tree syntaxNode) {
```
saberduck: Name of the method is misleading now, because you are setting `PRESENT` constraint in some cases.

java-frontend/src/main/java/org/sonar/java/se/checks/OptionalGetBeforeIsPresentCheck.java
```diff
@@ -71,10 +75,33 @@ public ProgramState checkPostStatement(CheckerContext context, Tree syntaxNode)
...
-    if(syntaxNode.is(Tree.Kind.METHOD_INVOCATION) && OPTIONAL_EMPTY.matches(((MethodInvocationTree) syntaxNode))) {
-      return context.getState().peekValue().setConstraint(context.getState(), OptionalConstraint.NOT_PRESENT);
+    ProgramState programState = context.getState();
+    if (syntaxNode.is(Tree.Kind.METHOD_INVOCATION)) {
+      return handleMethodInvocation(context, (MethodInvocationTree) syntaxNode);
```
saberduck: I believe introducing new method is unnecessary, since this is the only case. Better to return early when it's not mit and inline the logic from `handleMethodInvocation` here.

PR #1533, opened 2017-06-22
SONARJAVA-2153 Non-zero constraint should be added after division

saberduck:


java-frontend/src/test/files/se/MethodInvocationLeadingToArithmeticException.java
```diff
@@ -14,10 +14,8 @@ void foo2() {
 
   void foo3(int j) {
     int i = 42; // flow@foo3 [[order=1]] {{Implies 'i' is non-zero.}}
-    divByZeroIfArg1Zero(i, j); // FIXME there is no yield produced due to SONARJAVA-2153
```
m-g-sonar: Rather than purely removing it, I would prefer a new case, which does not raise any issue after the method invocation, as `j` can not be zero (we learn it from the yield).

PR #1534, opened 2017-06-22
SONARJAVA-2313 Consider nested new resources has being close

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -272,18 +276,30 @@ public void visitNewClass(NewClassTree syntaxNode) {
...
       }
     }
 
+    private boolean isNestedNewResourceCreation(ExpressionTree expr) {
+      if (expr.is(Tree.Kind.NEW_CLASS)) {
+        Arguments arguments = ((NewClassTree) expr).arguments();
+        if (arguments.stream().noneMatch(UnclosedResourcesCheck::isCloseable)) {
+          return true;
+        }
+        return arguments.stream().filter(UnclosedResourcesCheck::isCloseable).allMatch(this::isNestedNewResourceCreation);
```
benzonico: all match returns true if the stream is empty so no need for the previous if.

PR #1537, opened 2017-06-23
SONARJAVA-1072 ignore methods once lock has been encountered

benzonico:


java-checks/src/main/java/org/sonar/java/checks/StaticFieldInitializationCheck.java
```diff
@@ -36,6 +37,9 @@
 public class StaticFieldInitializationCheck extends AbstractInSynchronizeChecker {
 
   private Deque<Boolean> withinStaticInitializer = new LinkedList<>();
+  private Deque<Boolean> methodUsesLocks = new LinkedList<>();
+  private MethodMatcher lockMethod = MethodMatcher.create().typeDefinition("java.util.concurrent.locks.Lock").name("lock").withoutParameter();
```
m-g-sonar: I would use a `private static final MethodMatcherCollection LOCK_METHODS` instead of declaring two distinct matchers.

java-checks/src/main/java/org/sonar/java/checks/StaticFieldInitializationCheck.java
```diff
@@ -45,15 +49,26 @@
...
+    methodUsesLocks.clear();
   }
 
   @Override
   public void visitNode(Tree tree) {
+    if (tree.is(Tree.Kind.METHOD)) {
+      methodUsesLocks.push(false);
+    }
+    if (tree.is(Tree.Kind.METHOD_INVOCATION)
+      && (lockMethod.matches((MethodInvocationTree) tree) || tryLockMethod.matches((MethodInvocationTree) tree))
```
m-g-sonar: `LOCK_METHODS.anyMatch((MethodInvocationTree) tree)`

PR #1538, opened 2017-06-23
SONARJAVA-2344 RSPEC-4032 Useless package-info check

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/CheckList.java
```diff
@@ -20,6 +20,7 @@
 package org.sonar.java.checks;
 
 import com.google.common.collect.ImmutableList;
+
```
benzonico: ?

PR #1541, opened 2017-06-26
SONARJAVA-1377 Update rule to cover more cases (main description, thrown exception, placeholders)

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
```diff
@@ -298,8 +317,87 @@ private boolean hasNonVoidReturnType(Tree tree) {
...
+      if (javadoc == null) {
+        return Collections.emptyList();
+      }
+      String cleanupJavaDoc = cleanupJavaDoc(javadoc);
+      String[] lines = cleanupJavaDoc.split("\\r?\\n");
+      return Arrays.stream(lines).map(String::trim).collect(Collectors.toList());
+    }
+
+    private static String cleanupJavaDoc(String javadoc) {
+      return javadoc.trim().substring(3).replace("*/", "").replace("*", "").trim();
```
benzonico: The replace "*" by "" seems a bit strong : you are going to strip any * character from the javadoc. Given what is verified I think it is fine but that wouldn't it deserve a comment ?

java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
```diff
@@ -298,8 +317,87 @@ private boolean hasNonVoidReturnType(Tree tree) {
...
+    Javadoc(@Nullable String javadoc) {
+      List<String> lines = lines(javadoc);
+      mainDescription = extractMainDescription(lines);
+      parameters = extractToMap(lines, PARAMETER_JAVADOC_PATTERN);
+      thrownExceptions = extractToMap(lines, EXCEPTION_JAVADOC_PATTERN);
+      returnDescription = extractReturnDescription(lines);
+    }
+
+    private List<String> getUndocumentedParameters(List<String> parameterNames) {
+      return parameterNames.stream().filter(name -> Javadoc.isEmptyDescription(parameters.get(name))).collect(Collectors.toList());
```
benzonico: styling :  No need for the Javadoc. prefix here.

java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
```diff
@@ -298,8 +317,87 @@ private boolean hasNonVoidReturnType(Tree tree) {
...
+    }
+
+    private List<String> getUndocumentedThrownExceptions(List<String> exceptionNames) {
+      return exceptionNames.stream().filter(name -> Javadoc.isEmptyDescription(thrownExceptions.get(name))).collect(Collectors.toList());
```
benzonico: styling :  No need for the Javadoc. prefix here.

java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
m-g-sonar: I simply extracted the method which was previously already used, but I'm going to add a comment indeed.

m-g-sonar: Previous location: https://github.com/SonarSource/sonar-java/pull/1541/files/17d41c35d9f3d8693511f9c0a2e6148cb04163cb..5c5d8b55647fb6719d61c417a9e4c4b031808de9#diff-1e6f1b64c748464c9b3cfd6c481c756aL154

PR #1542, opened 2017-06-27
SONARJAVA-2361 handle analysis cancellation

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/ast/JavaAstScanner.java
```diff
@@ -58,12 +58,17 @@ public void scan(Iterable<File> files) {
...
+          cancelled = true;
+          break;
+        }
         simpleScan(file);
         progressReport.nextFile();
       }
-      successfullyCompleted = true;
+      successfullyCompleted = !cancelled;
     } finally {
       if (successfullyCompleted) {
```
henryju: Does it mean that when analysis is cancelled, the progress report is not stopped (I mean the background thread)?

m-g-sonar: Well, from what I can see (2 lines after), it is cancelled. see `progressReport.cancel();`

henryju: Arg, thanks GitHub for hiding interesting part :)

PR #1547, opened 2017-07-03
SONARJAVA-2169 fail analysis if sonar.java.binaries is empty

benzonico:
Note : analysis do not fail if project contains only one source file. For ease of testing and because bytecode is not required in those cases.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/AbstractJavaClasspath.java
+++ b/java-frontend/src/main/java/org/sonar/java/AbstractJavaClasspath.java
@@ -20,6 +20,7 @@
 package org.sonar.java;
 
 import com.google.common.base.Splitter;
+import com.google.common.collect.Iterables;
 import org.apache.commons.lang.StringUtils;
 import org.sonar.api.batch.BatchSide;
 import org.sonar.api.batch.fs.FileSystem;
@@ -100,6 +101,10 @@ public abstract class AbstractJavaClasspath {
 
   protected boolean hasJavaSources() {
     return fs.hasFiles(fs.predicates().and(fs.predicates().hasLanguage("java"), fs.predicates().hasType(fileType)));
+  }
+
+  protected boolean hasMoreThanOneJavaFile() {
+    return Iterables.size(fs.files(fs.predicates().and(fs.predicates().hasLanguage("java"), fs.predicates().hasType(fileType)))) > 1;
   }
 
   private Set<File> getFilesForPattern(Path baseDir, String pathPattern, boolean libraryProperty) {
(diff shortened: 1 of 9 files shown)


java-frontend/src/main/java/org/sonar/java/JavaClasspath.java
```diff
@@ -52,6 +52,9 @@ protected void init() {
         throw new AnalysisException(
           "sonar.binaries and sonar.libraries are not supported since version 4.0 of sonar-java-plugin, please use sonar.java.binaries and sonar.java.libraries instead");
       }
+      if (binaries.isEmpty() && hasMoreThanOneJavaFile()) {
+        throw new AnalysisException("Please provide compiled classes of your project with sonar.java.binaries property");
```
m-g-sonar: I would put quote around `sonar.java.binaries` to highlight it and make it explicit in the middle of the error message.

benzonico: The other messages don't have quotes...

java-frontend/src/main/java/org/sonar/java/AbstractJavaClasspath.java
```diff
@@ -102,6 +103,10 @@ protected boolean hasJavaSources() {
     return fs.hasFiles(fs.predicates().and(fs.predicates().hasLanguage("java"), fs.predicates().hasType(fileType)));
   }
 
+  protected boolean hasMoreThanOneJavaFile() {
+    return Iterables.size(fs.files(fs.predicates().and(fs.predicates().hasLanguage("java"), fs.predicates().hasType(fileType)))) > 1;
```
m-g-sonar: What about test binaries (for `org.sonar.java.JavaTestClasspath`)? Should we also fail if `sonar.java.test.binaries` is not provided when there is some test sources or we only issue a warning for this case?

benzonico: Problem is that you can perfectly analyze only sources (while analyzing only test is possible, it is less likely).

m-g-sonar: Okay, so let's continue to post warning in case of missing byte code for test sources, but start failing on main sources.

java-frontend/src/main/java/org/sonar/java/JavaClasspath.java
m-g-sonar: Okay! fine by me.

java-frontend/src/main/java/org/sonar/java/AbstractJavaClasspath.java
henryju: Iterables.size(fs.files(p)) > 1 can be replaced by fs.hasFiles(p)

m-g-sonar: Thanks @henryju !

m-g-sonar: Oups, nop, that's not the same. We are testing for explicitly MORE THAN 1 file, not for existence of any file.

henryju: oh, sorry, I missed that.

PR #1552, opened 2017-07-04
SONARJAVA-2353 Prefer nio.Files.delete rather than its file counterpart

benzonico:


java-checks/src/test/files/checks/NioFileDeleteCheck.java
```diff
@@ -0,0 +1,10 @@
+import java.io.File;
+import java.io.Path;
```
saberduck: wrong import, should have been `java.nio.file.Path`

PR #1559, opened 2017-07-06
SONARJAVA-2372 RSPEC-2203 "collect" should be used with "Streams" instead of "list::add"

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/CollectInsteadOfForeachCheck.java
```diff
@@ -0,0 +1,66 @@
...
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodReferenceTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+import java.util.List;
+
+@Rule(key = "S2203")
+public class CollectInsteadOfForeachCheck extends AbstractMethodDetection {
+
+  private static final MethodMatcher FOREACH = MethodMatcher.create().typeDefinition("java.util.stream.Stream").name("forEach").withAnyParameters();
```
benzonico: more of an edgecase I admit but there is also a foreach on Iterable : https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html#forEach-java.util.function.Consumer-

java-checks/src/main/java/org/sonar/java/checks/CollectInsteadOfForeachCheck.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return ImmutableList.of(FOREACH);
```
benzonico: I would use Collections.singletonList rather than guava for this.

java-checks/src/main/java/org/sonar/java/checks/CollectInsteadOfForeachCheck.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    ExpressionTree firstArgument = mit.arguments().get(0);
+    if (!firstArgument.is(Tree.Kind.METHOD_REFERENCE)) {
+      return;
+    }
+    MethodReferenceTree mrt = (MethodReferenceTree) firstArgument;
```
benzonico: I would try to cover simple cases where the add method is invoked directly as well. 

```
list.forEach(s -> otherlist.add(s));

```

java-checks/src/main/java/org/sonar/java/checks/CollectInsteadOfForeachCheck.java
m-g-sonar: I don't think it worth the effort, nor will be relevant in the context of this rule. The rule is explicitly targeting Lists, so it would mean targeting situations such as:
```
myList.forEach(myOtherList::add)
```
When encountered, what should we report? The current implementation would recommend to "_collect myList_" . It's not relevant as we are not dealing with streams, and therefore we are expanding scope of the rule a bit too much from my point of view.
Facing this situation, the correct answer should be to use directly `myOtherList.addAll(myList)`, but I would rather use another rule for that than expending this one.

PR #1561, opened 2017-07-07
SONARJAVA-3269 RSPEC-3552 Classes used as "HashMap" keys should implement "Comparable"

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/HashMapComparableKeyCheck.java
```diff
@@ -0,0 +1,147 @@
...
+  @CheckForNull
+  private static String nonComparableKeyTypeName(Type symbolType) {
+    Type type = symbolType;
+    if (type instanceof MethodJavaType) {
+      type = ((MethodJavaType) type).resultType();
+    }
+    if (type instanceof ParametrizedTypeJavaType) {
+      ParametrizedTypeJavaType parametrizedTypeJavaType = (ParametrizedTypeJavaType) type;
+      JavaType keyType = parametrizedTypeJavaType.substitution(parametrizedTypeJavaType.typeParameters().get(0));
+      if (!keyType.isTagged(JavaType.TYPEVAR) && !keyType.isSubtypeOf("java.lang.Comparable")) {
```
m-g-sonar: I wonder we should not skip interfaces as well. To me, issues when interfaces are used as key are still legit, as we need to be sure that `Comparable` is implemented. So the correct solution would be that the interface which is used should also force implementation of `Comparable`. I would appreciate another point of view.

java-checks/src/main/java/org/sonar/java/checks/HashMapComparableKeyCheck.java
```diff
@@ -0,0 +1,147 @@
...
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.TypeArguments;
+import org.sonar.plugins.java.api.tree.TypeTree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+
+@Rule(key = "S3552")
+public class HashMapComparableKeyCheck extends BaseTreeVisitor implements JavaFileScanner, JavaVersionAwareVisitor {
```
benzonico: I am unsure why the use of a BaseTreeVisitor here : visit is stopped but I don't understand why from the test.

java-checks/src/main/java/org/sonar/java/checks/HashMapComparableKeyCheck.java
benzonico: I tend to agree with your POV : let's be extra careful with feedback on peach.

java-checks/src/main/java/org/sonar/java/checks/HashMapComparableKeyCheck.java
```diff
@@ -0,0 +1,147 @@
...
+        if (name != null) {
+          reportIssue(newClassTree, name);
+        }
+      }
+    }
+  }
+
+  private void reportIssue(NewClassTree newClassTree, String typeName) {
+    context.reportIssue(this, reportTree(newClassTree),
+      String.format("Implement \"Comparable\" in \"%s\" or switch key type.%s",
```
benzonico: I don't understand the part with "or switch key type" tbh.

java-checks/src/main/java/org/sonar/java/checks/HashMapComparableKeyCheck.java
m-g-sonar: I stop the visit to not raise redundant issues, by exploring initializers multiple times. I slightly reworked the visiteor and added a test case which makes the case more explicit.

java-checks/src/main/java/org/sonar/java/checks/HashMapComparableKeyCheck.java
m-g-sonar: Ok, I'll work on another wording.

java-checks/src/main/java/org/sonar/java/checks/HashMapComparableKeyCheck.java
m-g-sonar: So, so we let the issues on interfaces for the time being, and have a look at the results once deployed on peach.

PR #1566, opened 2017-07-12
Extends scopes of S2699

m-g-sonar:


java-checks/src/test/java/org/sonar/java/checks/AssertionsInTestsCheckTest.java
```diff
@@ -76,6 +76,11 @@ public void jMock() {
     verify("JMock");
   }
 
+  @Test
+  public void wireMock() {
```
benzonico: I think it might worth it to look for parameterizedTest for this one.

PR #1567, opened 2017-07-12
SONARJAVA-2169 Analysis should not fail in sonarlint context

benzonico:


sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaPlugin.java
```diff
@@ -47,10 +48,14 @@ public void define(Context context) {
       builder.addAll(SurefireExtensions.getExtensions());
       builder.addAll(JaCoCoExtensions.getExtensions(sonarQubeVersion));
       builder.add(JavaSonarWayProfile.class);
+      builder.addAll(JavaClasspathProperties.getProperties());
+      builder.add(JavaClasspath.class);
+    } else {
+      builder.addAll(JavaClasspathProperties.getProperties());
```
m-g-sonar: I can get that `javaClasspath` needs properties read from `JavaClasspathProperties`, so the order may be important and it should be needed _before_ adding `JavaClasspath`... But why adding it in both side of `if` and not moving it before it then?

henryju: Order is not important, pico collect all classes, and then instantiate them depending on the dependencies.

m-g-sonar: Okay, Thanks @henryju, so this change is useless.

PR #1570, opened 2017-07-13
SONARJAVA-2370 RSPEC-3436 Value-based classes should not be used for locking

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/synchronization/ValueBasedObjectUsedForLockCheck.java
```diff
@@ -0,0 +1,85 @@
...
+
+  private static boolean isValueBasedType(Type type) {
+    if (type.isUnknown() || type.is(JAVA_TIME_CLOCK)) {
+      return false;
+    }
+    if (VALUE_BASED_TYPES.stream().anyMatch(type::is)) {
+      return true;
+    }
+    String fullyQualifiedName = type.fullyQualifiedName();
+    return fullyQualifiedName.startsWith(JAVA_TIME_PACKAGE)
```
benzonico: I would rather go with a regexp `java\.time\.[^\.]` rather than startWith AND a regexp. This would allow to have a simpler code as well : `return anyMatch... && pattern.matcher(type.fqn()).matches();`

java-checks/src/main/java/org/sonar/java/checks/synchronization/ValueBasedObjectUsedForLockCheck.java
```diff
@@ -0,0 +1,85 @@
...
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.SYNCHRONIZED_STATEMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ExpressionTree expression = ((SynchronizedStatementTree) tree).expression();
+    if (isValueBasedType(expression.symbolType())) {
+      reportIssue(expression, "Synchronize on some other object." + context.getJavaVersion().java8CompatibilityMessage());
```
benzonico: message seems rather unprecise.

PR #1571, opened 2017-07-13
SONARJAVA-2375 ThreadLocal with initial value check

benzonico:


java-checks/src/main/java/org/sonar/java/checks/ThreadLocalWithInitialCheck.java
```diff
@@ -0,0 +1,67 @@
...
+      return;
+    }
+    List<Tree> members = classTree.members();
+    if (members.size() != 1) {
+      return;
+    }
+    members.stream()
+      .filter(tree -> tree.is(Tree.Kind.METHOD))
+      .map(t -> ((MethodTree) t))
+      .filter(t -> "initialValue".equals(t.simpleName().name()))
```
m-g-sonar: I know that it's probably I pretty nasty corner case, but as you are not checking the parameters, you may end up with FP when the `initialValue` method is not the override. Up to you if you want to fix it or not.

```
  ThreadLocal<List<String>> threadLocal = new ThreadLocal<List<String>>() {
      protected java.util.List<String> initialValue(String s) { // FP - not the override method but a different method
        List<String> init = initialValue();
        init.add(s);
        return init;
      };
    };
```

java-checks/src/main/java/org/sonar/java/checks/ThreadLocalWithInitialCheck.java
```diff
@@ -0,0 +1,67 @@
...
+      .findFirst().ifPresent(
+        t -> reportIssue(newClassTree.identifier(), "Replace this anonymous class with a call to \"ThreadLocal.withInitial\".")
```
m-g-sonar: Missing compatibility version message if java version is not set.

PR #1574, opened 2017-07-13
SONARJAVA-2390 S2187 Support JUnit 5 Testable meta-annotation

saberduck:


java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
```diff
@@ -38,21 +39,46 @@
...
 import java.util.stream.Stream;
 
 @Rule(key = "S2187")
 public class NoTestInTestClassCheck extends IssuableSubscriptionVisitor {
 
-  private static final Predicate<SymbolMetadata.AnnotationInstance> PREDICATE_ANNOTATION_TEST_OR_UNKNOWN = input -> {
-    Type type = input.symbol().type();
-    return type.isUnknown() || isTestAnnotation(type);
-  };
+  private final Map<String, Boolean> testAnnotations;
```
benzonico: As discussed : while unlikely to happen in this case , this is a memory leak in disguise and goes against the usual way to write checks that are running on a file by file basis.

java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
```diff
@@ -38,21 +39,46 @@
...
+
+  public NoTestInTestClassCheck() {
+    testAnnotations = new HashMap<>();
```
benzonico: This map should probably be replaced by two sets : "encouteredAnnotations" and "testableAnnotations" : this should make the code simpler in the end.

java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
```diff
@@ -38,21 +38,38 @@
...
-      || type.symbol().metadata().isAnnotatedWith("org.junit.jupiter.api.Test");
+  private boolean isTestAnnotation(Type type) {
+    return testAnnotations.contains(type.fullyQualifiedName()) || isJUnitTestableMetaAnnotated(type);
+  }
+
+  private boolean isJUnitTestableMetaAnnotated(Type type) {
+    if (seenAnnotations.contains(type.fullyQualifiedName())) {
+      return false;
+    }
+    seenAnnotations.add(type.fullyQualifiedName());
```
benzonico: You can skip contains call by relying on result of add.

saberduck: I am not really big fan of using `add` that way, I find it easier to read, when there are no side-effects in conditions.

PR #1578, opened 2017-07-14
SONARJAVA-775 Resolve correctly type of clone method on arrays

benzonico:


java-frontend/src/main/java/org/sonar/java/resolve/Symbols.java
```diff
@@ -199,14 +199,13 @@ public Symbols(BytecodeCompleter bytecodeCompleter) {
       entry.getValue().primitiveType = entry.getKey();
     }
 
-    // TODO comment me
```
m-g-sonar: :+1:

PR #1581, opened 2017-07-17
SONARJAVA-2368 RSPEC-3824 SE-based implementation of S3824

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/MapComputeIfAbsentOrPresentCheck.java
```diff
@@ -0,0 +1,180 @@
...
+
+      this.getInvocation = getInvocation;
+      this.putInvocation = putInvocation;
+
+      this.value = value;
+      this.valueConstraint = valueConstraint;
+    }
+
+    private boolean isOnlyPossibleConstraintOnValueForAllPaths() {
+      return node.siblings().stream().allMatch(sibling -> valueConstraint.equals(sibling.programState.getConstraint(value, ObjectConstraint.class)));
```
benzonico: no need to go through siblings, we should keep this verification of raising issue only one kind of constraint contained in this check.

PR #1583, opened 2017-07-17
SONARJAVA-2386 Update S00115 and S00117 to target local constants

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/naming/BadConstantNameCheck.java
```diff
@@ -89,19 +114,50 @@ private void checkName(VariableTree variableTree) {
...
-        isFinal = true;
+  private static boolean isStaticFinal(Symbol variable) {
+    return (variable.isStatic() && variable.isFinal())
+      || (variable.owner().isMethodSymbol() && variable.isFinal());
+  }
+
+  private static class VariableFromMethodVisitor extends BaseTreeVisitor {
+
+    private final MethodTree method;
+    private final BadConstantNameCheck check;
```
benzonico: you could make the inner class not static instead of passing around the check, that's the purpose of inner class.

PR #1585, opened 2017-07-18
SONARJAVA-2364 correct CFG when exception thrown in catch block

benzonico:


java-frontend/src/test/files/se/UnclosedResourcesCheck.java
```diff
@@ -497,3 +497,30 @@ InputStream testChained2() {
 abstract class Utils {
   abstract OutputStream newOutputStream();
 }
+class nestedFinally {
+  void foo() {
+    try {
+      FileInputStream fis = null;
+      try {
+        fis = new FileInputStream("foobar.txt"); // compliant because of the outer finally.
```
m-g-sonar: There is no outer finally!

PR #1589, opened 2017-07-19
SONARJAVA-2391 Ignore thread which use runnables

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/ThreadOverridesRunCheck.java
```diff
@@ -72,4 +87,45 @@ private static boolean overridesRunMethod(Symbol.TypeSymbol classSymbol) {
...
+  private static class SuperRunnableVisitor extends BaseTreeVisitor {
+
+    private boolean callSuperWithRunnable = false;
+
+    private static final MethodMatcher SUPER_THREAD = MethodMatcher.create().typeDefinition("java.lang.Thread").name("<init>").withAnyParameters();
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree tree) {
+      if (SUPER_THREAD.matches(tree) && ThreadOverridesRunCheck.hasRunnableArgument(tree.arguments())) {
+        callSuperWithRunnable = true;
```
benzonico: micro optimisation but : no need to visit further...

java-checks/src/main/java/org/sonar/java/checks/ThreadOverridesRunCheck.java
```diff
@@ -72,4 +87,45 @@ private static boolean overridesRunMethod(Symbol.TypeSymbol classSymbol) {
...
+      }
+      super.visitMethodInvocation(tree);
+    }
+
+    @Override
+    public void visitClass(ClassTree tree) {
+      // skip inner classes
```
benzonico: Not sure I actually spotted the testcase where this skip of inner class is required.

m-g-sonar: There is one. [Line#47](https://github.com/SonarSource/sonar-java/pull/1589/files#diff-d93a2ec6a1ed7e239cc98988f8d389b5R47) would raise two issue if classes were not skipped.

m-g-sonar: modified my test case (see [Line#49](https://github.com/SonarSource/sonar-java/pull/1589/files#diff-d93a2ec6a1ed7e239cc98988f8d389b5R49))

PR #1595, opened 2017-07-20
FQF: improve coverage

saberduck:


benzonico:
Please use explicit words rather than abbreviation in commit title : this one is far from standard and not related to domain of work .

PR #1608, opened 2017-08-14
Fix viewer coverage

benzonico:


java-viewer/src/test/java/org/sonar/java/viewer/ViewerTest.java
```diff
@@ -77,4 +92,75 @@ public void values_with_error() {
...
+  public void start_server_and_test_requests() throws Exception {
+    ServerSocket serverSocket = new ServerSocket(0);
+    int localPort = serverSocket.getLocalPort();
+    serverSocket.close();
+    String uri = "http://localhost:" + localPort + "/";
+    Viewer.startWebServer(localPort, "class A {void fun() {}}");
+    try(CloseableHttpClient client = HttpClients.createMinimal()) {
+      CloseableHttpResponse resp = client.execute(new HttpGet(uri));
+      assertThat(resp.getStatusLine().getStatusCode()).isEqualTo(200);
+      assertThat(EntityUtils.toString(resp.getEntity())).isEqualTo(IOUtils.toString(new FileInputStream(new File("src/test/resources/viewer_result1.html"))));
```
saberduck: Testing for equality of the whole input seems bit fragile to me. I think it will be enough to test that response contains the code which was provided. Also this means that any minor change to the html output requires changing 3 test resources.

java-viewer/src/test/resources/viewer_result3.html
```diff
@@ -0,0 +1,234 @@
...
+  at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:228)<br/>
+  at org.sonar.java.viewer.Viewer.getValues(Viewer.java:117)<br/>
+  at org.sonar.java.viewer.Viewer.generate(Viewer.java:102)<br/>
+  at org.sonar.java.viewer.Viewer.generate(Viewer.java:96)<br/>
+  at org.sonar.java.viewer.Viewer.lambda$startWebServer$2(Viewer.java:85)<br/>
+  at spark.RouteImpl$1.handle(RouteImpl.java:61)<br/>
+  at spark.http.matching.Routes.execute(Routes.java:61)<br/>
+  at spark.http.matching.MatcherFilter.doFilter(MatcherFilter.java:126)<br/>
+  at spark.embeddedserver.jetty.JettyHandler.doHandle(JettyHandler.java:50)<br/>
+  at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:189)<br/>
```
saberduck: As mentioned above, this is too fragile. Actually I had the test failing because of different path taken in jetty internals.

benzonico: Changed this test regarding stacktrace and its potential instability. As per equality of what is render : as nothing is tested on that side, this is killing two bird with one stone : test should be improved when we improve the rendering, at least now this is tested.

PR #1613, opened 2017-08-17
Goto and iconst_x

benzonico:


java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
```diff
@@ -143,6 +151,19 @@ void executeInstruction(BytecodeCFGBuilder.Instruction instruction) {
       return;
     }
     switch (instruction.opcode) {
+      case ICONST_0:
+        SymbolicValue svZero = constraintManager.createSymbolicValue(instruction);
+        programState = programState.stackValue(svZero).addConstraint(svZero, DivisionByZeroCheck.ZeroConstraint.ZERO);
```
saberduck: BooleanConstraint.FALSE should also be added, to have correct behavior for boolean literals.

java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
```diff
@@ -143,6 +151,19 @@ void executeInstruction(BytecodeCFGBuilder.Instruction instruction) {
...
+        break;
+      case ICONST_M1:
+      case ICONST_1:
```
saberduck: `ICONST_1` should have `BooleanConstraint.TRUE`

PR #1625, opened 2017-08-22
Support STORE opcodes

saberduck:


java-frontend/src/test/java/org/sonar/java/bytecode/se/BytecodeEGWalkerExecuteTest.java
```diff
@@ -180,6 +180,17 @@ public void test_nullness_check() throws Exception {
     assertThat(hasConstraint(thisSv, programState, ObjectConstraint.NOT_NULL)).isTrue();
   }
 
+  @Test
+  public void test_store() throws Exception {
+    int[] storeOpcodes = new int[] {Opcodes.ISTORE, Opcodes.LSTORE, Opcodes.FSTORE, Opcodes.DSTORE, Opcodes.ASTORE};
+    SymbolicValue sv = new SymbolicValue();
+    ProgramState startState = ProgramState.EMPTY_STATE.stackValue(sv);
+    for (int opcode : storeOpcodes) {
+      ProgramState programState = execute(new Instruction(opcode, 0), startState);
```
benzonico: I would prefer to have an index different than zero... just for sake of testing that it works with 42 or 67 (because everyone loves prime).

PR #1627, opened 2017-08-22
SONARJAVA-2406 RSPEC-4087 "close()" calls should not be redundant

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/RedundantCloseCheck.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+@Rule(key = "S4087")
+public class RedundantCloseCheck extends IssuableSubscriptionVisitor {
+
+  private static final MethodMatcher AUTOCLOSABLE_CLOSE = MethodMatcher.create().typeDefinition(TypeCriteria.subtypeOf("java.lang.AutoCloseable")).name("close").withoutParameter();
```
saberduck: Typo in variable name, should be `AUTOCLOSEABLE_CLOSE`

PR #1631, opened 2017-08-24
SONARJAVA-2401 Add more immutable types to RSPEC-2201

benzonico:


saberduck:
As discussed, this is in fact API abuse, so fine by me.

PR #1632, opened 2017-08-24
SONARJAVA-2408 S4143 Map values should not be replaced unconditionally

saberduck:


java-checks/src/main/java/org/sonar/java/checks/OverwrittenKeyCheck.java
```diff
@@ -0,0 +1,173 @@
...
+import org.sonar.plugins.java.api.tree.ExpressionStatementTree;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.StatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+import javax.annotation.Nullable;
```
SonarTech: MINOR Remove this unused import 'javax.annotation.Nullable'. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

SonarTech:
* ![MINOR][MINOR] 1 minor

[MINOR]: https://sonarsource.github.io/sonar-github/severity-minor.png 'Severity: MINOR'

java-checks/src/main/java/org/sonar/java/checks/OverwrittenKeyCheck.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+@Rule(key = "S4143")
+public class OverwrittenKeyCheck extends IssuableSubscriptionVisitor {
+
+  private static final MethodMatcher MAP_PUT = MethodMatcher.create().typeDefinition("java.util.Map").name("put").parameters(TypeCriteria.anyType(), TypeCriteria.anyType());
```
m-g-sonar: TypeDefinition should be `TypeCriteria.subtypeOf("java.util.Map")`, or you are never going to catch stuff like:
```
  HashMap<Object, Object> map2;
  void foo() {
    map2.put(2, "foo");
    map2.put(2, "bar"); // Noncompliant
  }
```

java-checks/src/main/java/org/sonar/java/checks/OverwrittenKeyCheck.java
```diff
@@ -0,0 +1,177 @@
...
+    return Collections.singletonList(Tree.Kind.BLOCK);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
+      return;
+    }
+
+    Map<CollectionAndKey, Tree> usedKeys = new HashMap<>();
```
m-g-sonar: It seems to me that you don't need a map here but you may use only a `set<CollectionAndKey>`. If there is an issue detected, you can use the trees from the new `CollectionAndKey` and the one already existing from the set.

java-checks/src/main/java/org/sonar/java/checks/OverwrittenKeyCheck.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+      }
+    }
+    return null;
+  }
+
+  private void handleKey(Map<CollectionAndKey, Tree> keys, CollectionAndKey collectionAndKey) {
+    if (keys.containsKey(collectionAndKey)) {
+      Tree previousTree = keys.get(collectionAndKey);
+      String indexOrKey = collectionAndKey.keyTree.parent().is(Tree.Kind.ARRAY_DIMENSION) ? "index" : "key";
```
m-g-sonar: I would encode the type of the collection into the Object, and put a method there, or precompute it when creating the object.

java-checks/src/main/java/org/sonar/java/checks/OverwrittenKeyCheck.java
saberduck: As discussed, you are right that object in the Set would have the right tree, but there is no API to get the object which was inserted from the Set .

PR #1633, opened 2017-08-25
SONARJAVA-2432 expose equalToken in VariableTree API

m-g-sonar:


benzonico:
Adding this means that we should cleanup calls to implementation method with unecessary cast : see `DeadStoreCheck` and `DefaultJAvaFileScannerContextTest` and this will validate your implementation which lacks tests.

PR #1640, opened 2017-08-29
SONARJAVA-2416 rewrite rule to raise issue only on unread reassigned parameters

benzonico:


java-checks/src/main/java/org/sonar/java/checks/ParameterReassignedToCheck.java
```diff
@@ -43,60 +46,83 @@
...
   @Override
   public void visitMethod(MethodTree tree) {
+    BlockTree block = tree.block();
+    if(block == null) {
+      return;
+    }
+    CFG cfg = CFG.build(tree);
+    LiveVariables analyze = LiveVariables.analyze(cfg);
+    Set<Symbol> live = analyze.getIn(cfg.entry());
     for (VariableTree parameterTree : tree.parameters()) {
```
m-g-sonar: Why not rely on stream?
```
tree.parameters().stream()
  .map(VariableTree::symbol)
  .filter(symbol -> !live.contains(symbol))
  .forEach(variables::add)
```

java-checks/src/main/java/org/sonar/java/checks/ParameterReassignedToCheck.java
```diff
@@ -43,60 +46,83 @@
...
-      variables.add(parameterTree.symbol());
+      if(!live.contains(parameterTree.symbol())) {
+        variables.add(parameterTree.symbol());
+      }
     }
     super.visitMethod(tree);
     for (VariableTree parameterTree : tree.parameters()) {
```
m-g-sonar: same comment about usage of stream.

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -81,7 +81,13 @@
 
 public class CFG {
 
-  private final Symbol.MethodSymbol methodSymbol;
+  private final boolean ignoreBreakAndContinue;
+
+  public void setMethodSymbol(Symbol.MethodSymbol methodSymbol) {
```
m-g-sonar: Even if [S1213](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS1213) is not enabled, I would prefer to have this setter a bit lower in the members declarations (at least after constructors).

m-g-sonar:
There is only indirect tests (tested from the rule) about how you made validation of `Break` and `Continue` statement in CFG optional. As any new feature, it needs proper testing in `CFGTest`, at least to show why it has been added.

java-checks/src/main/java/org/sonar/java/checks/ParameterReassignedToCheck.java
benzonico: I see no added value for stream here... neither from readability nor perf...

m-g-sonar: Okay, I was simply wondering about calling 4 times `parameterTree.symbol()`

PR #1641, opened 2017-08-29
SONARJAVA-2409 Raise issue for sync on new instance

benzonico:


java-checks/src/test/files/checks/SynchronizedFieldAssignmentCheck.java
```diff
@@ -109,4 +109,10 @@ void fun1(Object param) {
 
     }
   }
+
+  void fun2() {
+    synchronized (new Object()) { // Noncompliant
```
m-g-sonar: The message used here is different from the other ones and should be tested.

PR #1646, opened 2017-08-29
SONARJAVA-2414 RSPEC-4165 Assignments should not be redundant

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/RedundantAssignmentsCheck.java
```diff
@@ -0,0 +1,124 @@
...
+
+@Rule(key = "S4165")
+public class RedundantAssignmentsCheck extends SECheck {
+
+  private static final Set<String> STREAM_TYPES = ImmutableSet.of(
+    "java.util.stream.Stream",
+    "java.util.stream.IntStream",
+    "java.util.stream.LongStream",
+    "java.util.stream.DoubleStream");
+  private final Deque<Multimap<AssignmentExpressionTree, Assignment>> contexts = new LinkedList<>();
```
benzonico: This is a bit too generic name. Confusing with actual JavaFileScannerContext and CheckerContext. I would actually try to prefix it at the very least.

java-frontend/src/main/java/org/sonar/java/se/checks/RedundantAssignmentsCheck.java
```diff
@@ -0,0 +1,124 @@
...
+
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    contexts.clear();
```
benzonico: This one should not be necessary... safe to keep it but a bit defensive.

java-frontend/src/main/java/org/sonar/java/se/checks/RedundantAssignmentsCheck.java
```diff
@@ -97,28 +106,39 @@ public void checkEndOfExecution(CheckerContext context) {
...
+    private final Symbol assignedSymbol;
+    @Nullable
+    private final Symbol fromSymbol;
     private final SymbolicValue oldValue;
     private final SymbolicValue newValue;
+    private final ExplodedGraph.Node node;
 
-    public Assignment(Symbol symbol, SymbolicValue oldValue, SymbolicValue newValue) {
-      this.symbol = symbol;
+    public Assignment(Symbol assignedSymbol, @Nullable SymbolicValue oldValue, SymbolicValue newValue, @Nullable Symbol fromSymbol, ExplodedGraph.Node node) {
```
benzonico: Name of this data class is a bit dodgy... hard to get its just a data class from its name only. Not sure how to refine it though.

PR #1648, opened 2017-08-30
SONARJAVA-2422 Assertion argument order

benzonico:


java-checks/src/main/java/org/sonar/java/checks/AssertionArgumentOrderCheck.java
```diff
@@ -0,0 +1,82 @@
...
+    if (isConstant(argToCheck)) {
+      context.reportIssue(this, argToCheck, "Make sure these 2 arguments are in the correct order: expected value, actual value.");
+    }
+  }
+
+  private static ExpressionTree getActualArgument(MethodInvocationTree mit) {
+    int arity = mit.arguments().size();
+    ExpressionTree arg = mit.arguments().get(arity - 1);
+    // Check for assert equals method with delta
+    if (arity > 2 && (arity == 4 || mit.arguments().stream().allMatch(AssertionArgumentOrderCheck::isDouble))) {
```
m-g-sonar: There is also an `assertEquals(float expected, float actual, float delta)` method which only used float and has a delta. It won't be caught by this.

PR #1652, opened 2017-08-31
SONARJAVA-2415 RSPEC-4142 Duplicate values should not be passed as arguments

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/DuplicateArgumentCheck.java
```diff
@@ -0,0 +1,104 @@
...
+  public void visitNode(Tree tree) {
+    MethodInvocationTree mit = (MethodInvocationTree) tree;
+    Arguments arguments = mit.arguments();
+    int nbrArgument = arguments.size();
+    if (nbrArgument <= 1) {
+      return;
+    }
+    Set<ExpressionTree> reported = new HashSet<>();
+    for (int i = 0; i < nbrArgument; i++) {
+      ExpressionTree arg = arguments.get(i);
```
benzonico: Arguable but I would fo for ExpressionUtils.skipParentheses(arguments.get(i)) to get cases like foo(point.x, (point.y)) and or exclude faster cases like foo(("bla"), "flah");

java-checks/src/main/java/org/sonar/java/checks/DuplicateArgumentCheck.java
```diff
@@ -0,0 +1,104 @@
...
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodInvocationTree mit = (MethodInvocationTree) tree;
+    Arguments arguments = mit.arguments();
+    int nbrArgument = arguments.size();
```
benzonico: I would name this arity

PR #1663, opened 2017-09-04
Fix quality flaw: S4143 improve code coverage

saberduck:


java-checks/src/test/files/checks/OverwrittenKeyCheck.java
```diff
@@ -74,4 +74,18 @@ void unknownSymbols(Map<X<?>, X<?>> x) {
     x.put(UNKNOWN_2, BLA);
   }
 
+  int[] fieldArr;
+
+  void coverage(int i, int[] arr) {
```
benzonico: Pleas don't name your cases like this : too generic, it covers a specific case please name it accordingly, otherwise 2 weeks from now we will have no idea what this one is about.

benzonico: And actually, I have NO idea from this PR, what this modification is about...

PR #1668, opened 2017-09-06
SONARJAVA-2188 Fix handling of Exceptional SV leading to correct yield creation

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/constraint/ConstraintsByDomain.java
```diff
@@ -92,6 +93,11 @@ public boolean equals(Object o) {
   }
 
   @Override
+  public String toString() {
+    return "[" + stream().map(Constraint::toString).sorted().collect(Collectors.joining(",")) + "]";
```
saberduck: Minor: you can provide suffix and prefix args to `Collectors.joining` and avoid adding brackets with `+` 

I really like this idea :)

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -793,7 +799,14 @@ private void executeVariable(VariableTree variableTree, @Nullable Tree terminato
       if (terminator != null && terminator.is(Tree.Kind.FOR_EACH_STATEMENT)) {
         sv = constraintManager.createSymbolicValue(variableTree);
       } else if (variableTree.parent().is(Tree.Kind.CATCH)) {
-        sv = getCaughtException(variableTree.symbol().type());
+        SymbolicValue thrownSV = getCaughtException(variableTree.symbol().type());
+        // replace the exceptional sv by a sv wrapping it, because the exception will be used as a variable within the block
+        sv = new SymbolicValue() {
```
saberduck: I think it will be better to have dedicated `static` class for this type of symbolic value, to avoid capturing implicit reference to the outer `ExplodedGraphWalker` . Also I would prefer not to override `wrappedValue` but provide specific method for this purpose.  You should use `constraintManager` and symbolic value factory mechanism to create instance of SV.

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -793,7 +799,14 @@ private void executeVariable(VariableTree variableTree, @Nullable Tree terminato
       if (terminator != null && terminator.is(Tree.Kind.FOR_EACH_STATEMENT)) {
         sv = constraintManager.createSymbolicValue(variableTree);
       } else if (variableTree.parent().is(Tree.Kind.CATCH)) {
-        sv = getCaughtException(variableTree.symbol().type());
+        SymbolicValue thrownSV = getCaughtException(variableTree.symbol().type());
```
saberduck: I think more apt name would be `caughtSV` . Not directly related, but `getCaughtException` is needlessly complex, it can be simplified to following
```
 private SymbolicValue getCaughtException(Type caughtType) {
    SymbolicValue sv = programState.peekValue();
    // FIXME SONARJAVA-2069 every path conducting to a catch block should have an exceptional symbolic value on top of the stack
    if (sv instanceof SymbolicValue.ExceptionalSymbolicValue && !((SymbolicValue.ExceptionalSymbolicValue) sv).exceptionType().isUnknown()) {
      return sv;
    }
    return constraintManager.createExceptionalSymbolicValue(caughtType);
  }
```

java-frontend/src/main/java/org/sonar/java/se/constraint/ConstraintsByDomain.java
m-g-sonar: Didn't know that one. Nice!

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
m-g-sonar: Almost, forgot the {{null}} case, but I got the idea. :+1:

m-g-sonar: Which in fact is not equivalent to the previous one. Exception was consumed by unstacking.

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
m-g-sonar: I'll rethink a bit that, indeed.

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -819,22 +812,24 @@ public SymbolicValue wrappedValue() {
...
   }
 
-  private SymbolicValue getCaughtException(Type caughtType) {
-    SymbolicValue sv = null;
+  private SymbolicValue handleException(Type caughtType) {
+    SymbolicValue.ExceptionalSymbolicValue sv = null;
     Type exceptionType = null;
     // FIXME SONARJAVA-2069 every path conducting to a catch block should have an exceptional symbolic value on top of the stack
     if (programState.peekValue() instanceof SymbolicValue.ExceptionalSymbolicValue) {
+      // clear the stack by consuming the exception
```
saberduck: Actually, stack should be really cleared by calling `programState.clearStack()` not just popping one SV. See https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html#jvms-6.5.athrow  . This might be important for being able to hit cached node in EG. Also it should be cleared regardless if the type is known or not (so perhaps should be done in `executeVariable` together with setting constraint)

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -819,22 +812,24 @@ public SymbolicValue wrappedValue() {
     }
   }
 
-  private SymbolicValue getCaughtException(Type caughtType) {
-    SymbolicValue sv = null;
+  private SymbolicValue handleException(Type caughtType) {
```
saberduck: I would name this `handleCatchVariable`  , `handleException` is bit unclear

PR #1670, opened 2017-09-11
SONARJAVA-2449 Handle filters on Optional

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/OptionalGetBeforeIsPresentCheck.java
```diff
@@ -168,6 +172,11 @@ public void visitMethodInvocation(MethodInvocationTree tree) {
         context.addExceptionalYield(peek, programState, "java.util.NoSuchElementException", check);
         reportIssue(tree);
         programState = null;
+      } else if (OPTIONAL_FILTER.matches(tree)) {
+        // filter has one parameter, so optional is next item on stack
+        SymbolicValue optionalSV = programState.peekValue(1);
+        // reuse the same optional - will cause FN as we make filtering a no-op
+        constraintManager.setValueFactory(() -> optionalSV);
```
saberduck: I think we can actually avoid this FN, by creating new optional SV and setting the constraints accordingly

java-frontend/src/test/files/se/OptionalGetBeforeIsPresentCheck.java
```diff
@@ -81,6 +81,22 @@ private void usingOfNullableWithTest(@Nullable Object o) {
...
+  }
+
+  private void usingFilter2(Optional<String> op) {
+    if (!op.filter(this::testSomething).isPresent()) {
+      return;
+    }
+    op.get(); // Compliant - FN
+  }
+
+  abstract boolean testSomething(String s);
```
saberduck: you can use `String::isEmpty` for the same effect and no need to make class abstract

PR #1672, opened 2017-09-11
SONARJAVA-2444 Add flow message when exception is thrown

saberduck:


java-frontend/src/test/java/org/sonar/java/se/Expectations.java
```diff
@@ -171,6 +169,10 @@ String message() {
       return MESSAGE.get(attributes);
     }
 
+    int line() {
```
m-g-sonar: Seems to me that this is unused, or I can't see the change.

java-frontend/src/test/java/org/sonar/java/se/Expectations.java
```diff
@@ -198,31 +198,39 @@ public String toString() {
   }
 
 
-  Optional<String> containFlow(List<AnalyzerMessage> flow) {
-    if (flowLines == null) {
-      flowLines = computeFlowLines();
+  Optional<String> containFlow(List<AnalyzerMessage> actual) {
+    List<Integer> actualLines = flowToLines(actual, AnalyzerMessage::getLine);
+    List<Map.Entry<String, SortedSet<FlowComment>>> expectedFlows = Multimaps.asMap(flows).entrySet().stream()
```
m-g-sonar: That's ugly (and a huge pain to follow). What about a dedicated type to encode that?

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -315,6 +318,18 @@ public int hashCode() {
         .map(f -> new ExecutionPath(edge, visited.add(edge), newTrackSymbols, newSameConstraints, f, endOfPath));
     }
 
+    private Optional<JavaFileScannerContext.Location> flowFromThrownException(ExplodedGraph.Edge edge) {
+      if (isMethodInvocationNode(edge.parent)) {
+        SymbolicValue peekValue = edge.child.programState.peekValue();
+        if (peekValue instanceof SymbolicValue.ExceptionalSymbolicValue) {
+          Type type = ((SymbolicValue.ExceptionalSymbolicValue) peekValue).exceptionType();
+          String msg = type == null ? "Exception is thrown." : ("'" + type.name() + "' is thrown.");
```
m-g-sonar: More a question of style, but I think I would prefer something such as:
```
String msg = String.format("%s is thrown.", (type == null ? "Exception" : "'" + type.name() + "'"));
```

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -67,6 +66,8 @@
 @Rule(key = "S2095")
 public class UnclosedResourcesCheck extends SECheck {
 
+  private static final ArrayList<Class<? extends Constraint>> RESOURCE_CONSTRAINT_DOMAIN = Lists.newArrayList(ResourceConstraint.class);
```
m-g-sonar: `ArryList` ? Isn't `List` working as well?

java-frontend/src/test/java/org/sonar/java/se/checks/FlowComputationTest.java
```diff
@@ -146,4 +153,15 @@ public void test_method_invocations_without_flows() throws Exception {
...
   }
+
+  @Test
+  public void test_first_flow_location() {
+    List<JavaFileScannerContext.Location> flow1 = ImmutableList
+      .of(new JavaFileScannerContext.Location("last", mock(Tree.class)), new JavaFileScannerContext.Location("first", mock(Tree.class)));
+    List<JavaFileScannerContext.Location> collect = FlowComputation.firstFlowLocation(flow1).collect(Collectors.toList());
+    assertThat(collect).hasSize(1);
+    assertThat(collect.get(0).msg).isEqualTo("first");
+    Stream<JavaFileScannerContext.Location> empty = FlowComputation.firstFlowLocation(Collections.emptyList());
```
m-g-sonar: maybe an empty line in between these two cases to differentiate them.

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
saberduck: srry, this was stupid refactoring done too fast

java-frontend/src/test/java/org/sonar/java/se/Expectations.java
saberduck: I don't want to introduce dedicated type, but I simplified this a lot, it was unnecessarily complex

PR #1676, opened 2017-09-14
SONARJAVA-2451 Ignore caught 'IllegalArgumentException' when 'Enum.valueOf(...)' is used

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/CatchUsesExceptionWithContextCheck.java
```diff
@@ -78,11 +83,42 @@ public void scanFile(JavaFileScannerContext context) {
...
+  }
+
+  private static boolean containsEnumValueOf(Tree tree) {
+    EnumValueOfVisitor visitor = new EnumValueOfVisitor();
+    tree.accept(visitor);
+    return visitor.hasEnumValueOf;
+  }
+
+  private static class EnumValueOfVisitor extends BaseTreeVisitor {
+    private static final MethodMatcher ENUM_VALUE_OF = MethodMatcher.create().typeDefinition("java.lang.Enum").name("valueOf").withAnyParameters();
```
saberduck: Matcher should also match subtypes of enum otherwise we are not going to catch usages like `MyEnum.valueOf("C");`

java-checks/src/main/java/org/sonar/java/checks/CatchUsesExceptionWithContextCheck.java
```diff
@@ -78,11 +83,42 @@ public void scanFile(JavaFileScannerContext context) {
...
+      tree.catches().stream()
+        .filter(c -> c.parameter().symbol().type().is("java.lang.IllegalArgumentException"))
+        .findAny()
+        .ifPresent(excludedCatchTrees::add);
+    }
+    super.visitTryStatement(tree);
+  }
+
+  private static boolean containsEnumValueOf(Tree tree) {
+    EnumValueOfVisitor visitor = new EnumValueOfVisitor();
```
saberduck: Visitor should probably cut on anonymous classes and lambda bodies

m-g-sonar: Correct, I have been lazy on this one.

PR #1677, opened 2017-09-14
SONARJAVA-2459 Fix possible NPE when computing flows

saberduck:


java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -331,7 +331,7 @@ public int hashCode() {
         return Optional.empty();
       }
       SymbolicValue peekValue = edge.child.programState.peekValue();
-      if (peekValue instanceof SymbolicValue.ExceptionalSymbolicValue) {
+      if (peekValue instanceof SymbolicValue.ExceptionalSymbolicValue && edge.parent.programPoint.syntaxTree() != null) {
```
m-g-sonar: I would like a comment here. Especially if we plan to remove this once we fix clearing of exceptional SV when reaching `finally` block.

PR #1680, opened 2017-09-15
SONARJAVA-2461 Fix NPE of true successor

benzonico:


java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
```diff
@@ -315,10 +316,10 @@ public void visitTypeInsn(int opcode, String type) {
     public void visitJumpInsn(int opcode, Label label) {
       if(opcode == GOTO || opcode == JSR) {
         currentBlock.terminator = new Instruction(opcode);
-        blockByLabel.computeIfAbsent(label, l -> currentBlock.createSuccessor());
+        currentBlock.successors = Collections.singletonList(blockByLabel.computeIfAbsent(label, l -> currentBlock.createSuccessor()));
```
saberduck: I think it will be cleaner to make `createSuccessor` take a label as an argument.
Here you are setting `successors` in the `createSuccessor`, when the label is missing, and same field is being immediately overwritten, which makes logic bit hard to follow. Also probably all other usage of `computeIfAbsent` could be replaced by such method.

java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
```diff
@@ -315,10 +316,10 @@ public void visitTypeInsn(int opcode, String type) {
...
         return;
       }
-      blockByLabel.computeIfAbsent(label, l -> currentBlock.createTrueSuccessor());
+      currentBlock.trueBlock = blockByLabel.computeIfAbsent(label, l -> currentBlock.createTrueSuccessor());
```
saberduck: Same as above, you can set `trueBlock` with result of `createSuccessor` taking label as argument, and this will actually eliminate the need for dedicated method

java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
```diff
@@ -315,12 +306,13 @@ public void visitTypeInsn(int opcode, String type) {
     public void visitJumpInsn(int opcode, Label label) {
       if(opcode == GOTO || opcode == JSR) {
         currentBlock.terminator = new Instruction(opcode);
-        blockByLabel.computeIfAbsent(label, l -> currentBlock.createSuccessor());
+        currentBlock.successors = Collections.singletonList(blockByLabel.computeIfAbsent(label, l -> currentBlock.createSuccessor()));
         return;
       }
-      blockByLabel.computeIfAbsent(label, l -> currentBlock.createTrueSuccessor());
+      currentBlock.trueBlock = blockByLabel.computeIfAbsent(label, l -> currentBlock.createSuccessor());
```
saberduck: If the label is already associated with the block, returned block will be set as a `trueBlock` but it will be not part of `successors` of `currentBlock`

PR #1686, opened 2017-09-19
SONARJAVA-2473 Ignore exception messages when computing flows for S2095, S2222 and S3546

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -254,15 +269,17 @@ public boolean hasAlwaysSameConstraint(@Nullable Symbol symbol) {
     final PSet<ExplodedGraph.Edge> visited;
     final Flow flow;
     final boolean finished;
+    final boolean skipExceptionMessages;
```
saberduck: You don't need field in `ExecutionPath`, it never changes during graph traversal, you can use field from `FlowComputation`

java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -254,15 +269,17 @@ public boolean hasAlwaysSameConstraint(@Nullable Symbol symbol) {
...
 
     private ExecutionPath(@Nullable ExplodedGraph.Edge edge, PSet<ExplodedGraph.Edge> visited, PSet<Symbol> trackedSymbols, SameConstraints sameConstraints,
-                          Flow flow, boolean finished) {
+      Flow flow, boolean finished, boolean skipExceptionMessages) {
```
saberduck: see above, parameter is not needed

PR #1688, opened 2017-09-20
SONARJAVA-2462 Also reset fields not declared in files on method invocations

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/se/ExplodedGraphWalkerTest.java
```diff
@@ -440,6 +441,19 @@ public void init(MethodTree methodTree, CFG cfg) {
     assertThat(visitedMethods).containsExactly("test", "privateMethod");
   }
 
+  @Test
+  public void constraints_on_field_reset() {
+    List<File> classpath = Collections.singletonList(new File("target/test-classes"));
```
saberduck: `target/test-classes` is by default on the classpath, see https://github.com/SonarSource/sonar-java/blob/c08b6132063c570f76342f06204638ba895240c2/java-frontend/src/test/java/org/sonar/java/se/JavaCheckVerifier.java#L182  , so this is not necessary

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -451,33 +452,38 @@ public void accept(SymbolicValue symbolicValue, ConstraintsByDomain constraintPM
...
-      SymbolicValue oldValue = newValues.get(symbol);
-      if (oldValue != null) {
+    for (Symbol field : fields) {
+      SymbolicValue oldValue = newValues.get(field);
+      if (field != null) {
         newReferences = decreaseReference(newReferences, oldValue);
       }
-      SymbolicValue newValue = constraintManager.createSymbolicValue(variableTree);
-      newValues = newValues.put(symbol, newValue);
+      Tree declarationOrFirstUsage = declarationOrFirstUsage(field);
```
saberduck: I think it is not necessary to search for a tree of a symbol. This was always `VariableTree` before which leads to call of `constraintManager.createDefaultSymbolicValue`. Now it is possible that this will be `IdentifierTree` , which can lead to some FPs when we accidentally trigger some logic in constraint manager like contrived example below

```
class B {
 Boolean TRUE;
}
class A extends B {

  void testUsage() {
    TRUE = Boolean.TRUE;
    TRUE = Boolean.FALSE;
    doSomething();
    if (TRUE == Boolean.FALSE) {  // FP condition always false

    }
  }
}
```
So I think it will be correct to make sure we are creating "fresh" symbolic value using `constraintManager.createDefaultSymbolicValue()`  with added benefit that we don't need the `Tree` for the symbol.

java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -451,33 +452,38 @@ public void accept(SymbolicValue symbolicValue, ConstraintsByDomain constraintPM
...
     }
     PMap<Symbol, SymbolicValue> newValues = values;
     PMap<SymbolicValue, Integer> newReferences = references;
-    for (VariableTree variableTree : variableTrees) {
-      Symbol symbol = variableTree.symbol();
-      SymbolicValue oldValue = newValues.get(symbol);
-      if (oldValue != null) {
+    for (Symbol field : fields) {
+      SymbolicValue oldValue = newValues.get(field);
+      if (field != null) {
```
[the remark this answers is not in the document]
m-g-sonar: wow, I went a bit too fast here. Thx.

PR #1694, opened 2017-09-21
SONARJAVA-2474 Binary expressions are always NOT_NULL

saberduck:


java-frontend/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -429,25 +429,29 @@ private boolean shouldTerminate(Set<LearnedConstraint> learnedConstraints) {
...
-          } else {
-            msg = IMPLIES_IS_MSG;
-          }
-          flowBuilder.add(location(node, String.format(msg, symbolName, constraint.valueAsString())));
+        return Flow.of(location(node, String.format(IMPLIES_SAME_VALUE, learnedAssociation.symbol().name(), rhsSymbol.name())));
+      }
+      Flow.Builder flowBuilder = Flow.builder();
+      ConstraintsByDomain allConstraints = programState.getConstraints(learnedAssociation.symbolicValue());
+      Collection<Constraint> constraints = filterByDomains(allConstraints);
+      boolean isBoolean = learnedAssociation.symbol.type().is("boolean");
```
m-g-sonar: I think I would prefer here to test `type().isPrimitive(Primitives.BOOLEAN)`, or even  `type().isPrimitive()` , as none of the primitive type can be null.

java-frontend/src/test/files/se/BinaryExpressionNotNull.java
```diff
@@ -0,0 +1,16 @@
...
+
+  void test(Object o) {
+    Boolean b = o != null;
+    if (b == null) { // Noncompliant {{Change this condition so that it does not always evaluate to "false"}}
+
+    }
+
+    String s1 = null;
+    String s2 = "test";
+    String s = s1 + s2;
```
m-g-sonar: Can you add a comment here that a NPE is expected? with a link to MMF-859.

saberduck: String concatenation will not fail with null value, result will be `"nulltest"` ,  but good point, I will add another case using `Integer`

PR #1703, opened 2017-09-26
SONARJAVA-2470 Handle method references terminating streams

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/ThreadOverridesRunCheck.java
```diff
@@ -35,12 +36,13 @@
 import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
-import java.util.Collection;
 import java.util.List;
 
 @Rule(key = "S2134")
 public class ThreadOverridesRunCheck extends IssuableSubscriptionVisitor {
 
+  private static final MethodMatcher RUN = MethodMatcher.create().typeDefinition(TypeCriteria.anyType()).name("run").withoutParameter();
```
saberduck: type definition should be that type is subtype of `java.lang.Thread`

java-frontend/src/test/files/se/StreamNotConsumedCheck.java
```diff
@@ -97,6 +97,32 @@ void exceptionalPath3(Stream<Object> stream) {
     stream.forEach(System.out::println);
   }
 
+  boolean useMethodInvocation1(List<Boolean> list) {
```
saberduck: should the name be `useMethodReference` ?

java-frontend/src/test/files/se/StreamNotConsumedCheck.java
```diff
@@ -97,6 +97,32 @@ void exceptionalPath3(Stream<Object> stream) {
...
+  boolean useMethodInvocation3(List<Boolean> list) {
+    Stream<Boolean> filter = list.stream().filter(Boolean::booleanValue); // Compliant
+    uses(filter);
+    return consumes(filter::iterator);
+  }
+
+  boolean useMethodInvocation4(List<Boolean> list) {
+    Stream<Boolean> filter = list.stream().filter(Boolean::booleanValue); // Compliant
+    java.util.Iterator<Boolean> itr = filter.iterator();
+    return consumes(filter::iterator);
```
saberduck: This should raise issue on S3959  because stream is already consumed. This works when lines are switched, and call using method reference is before the actual invocation.

java-frontend/src/main/java/org/sonar/java/se/checks/StreamConsumedCheck.java
```diff
@@ -128,6 +132,20 @@ private ProgramState handleMethodInvocation(CheckerContext context, MethodInvoca
...
   }
 
+  private static ProgramState handleMethodReference(CheckerContext context, MethodReferenceTree mrt) {
+    ProgramState programState = context.getState();
+    if (TERMINAL_OPERATIONS.anyMatch(mrt.method().symbol())) {
+      Tree expression = mrt.expression();
+      if (expression.is(Tree.Kind.IDENTIFIER)) {
+        SymbolicValue ownerSV = programState.getValue(((IdentifierTree) expression).symbol());
+        if (ownerSV != null && programState.getConstraint(ownerSV, StreamPipelineConstraint.class) != null) {
+          return Iterables.getOnlyElement(ownerSV.setConstraint(programState, StreamPipelineConstraint.CONSUMED));
```
saberduck: There should be a check if the SV is already consumed and if it's the case issue should be raised for S3959, see also comment in tests

java-checks/src/main/java/org/sonar/java/checks/ClassWithoutHashCodeInHashStructureCheck.java
```diff
@@ -34,6 +36,9 @@
 @Rule(key = "S2141")
 public class ClassWithoutHashCodeInHashStructureCheck extends IssuableSubscriptionVisitor {
 
+  private static final MethodMatcher EQUALS_MATCHER = MethodMatcher.create().typeDefinition(TypeCriteria.anyType()).name("equals").parameters("java.lang.Object");
```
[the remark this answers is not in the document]
m-g-sonar: Nice catch. I was sure `typeDefinition` was mandatory, and I didn't check.

PR #1707, opened 2017-09-27
SONARJAVA-2480 Detect already declared constants in S1192

saberduck:


java-checks/src/main/java/org/sonar/java/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -33,57 +32,82 @@
...
+import java.util.stream.Collectors;
 
 @Rule(key = "S1192")
 public class StringLiteralDuplicatedCheck extends BaseTreeVisitor implements JavaFileScanner {
 
   private static final int DEFAULT_THRESHOLD = 3;
 
-  private static final Integer MINIMAL_LITERAL_LENGTH = 7;
+  // String literals include quotes, so this means length 5 as defined in RSPEC
+  private static final int MINIMAL_LITERAL_LENGTH = 7;
```
m-g-sonar: :+1:

java-checks/src/main/java/org/sonar/java/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -33,57 +32,82 @@
...
     if (tree.is(Tree.Kind.STRING_LITERAL)) {
       String literal = tree.value();
+      isConstant(tree).ifPresent(constant -> constants.putIfAbsent(literal, constant));
       if (literal.length() >= MINIMAL_LITERAL_LENGTH) {
         occurrences.put(literal, tree);
       }
     }
   }
 
+  private static Optional<VariableTree> isConstant(LiteralTree tree) {
```
m-g-sonar: I think it would have been easier to simply visit variableTrees directly, and once checked that they are `static` and `final`, just check for an STRING_LITERAL initializer.

PR #1710, opened 2017-09-28
SONARJAVA-2455 Ignore methods 'String.intern()' and 'Character.toChars(int,char[],int)'

m-g-sonar:


saberduck:
Ticket number in the commit message is wrong, it should be SONARJAVA-2455

PR #1726, opened 2017-10-05
Fix for EqualsNotOverridenInSubclassCheck when a parent has final implementation but super-parents don't

alowaniak:
Added a "class with equals in-between" Object and a "final equals" class in the test case.
Adjusted the check so it checks if a parentImplementsNonFinalEquals.
Thus if it encounters a parent that implements a final equals it doesn't travel up the tree any further.


benzonico:
Hi, 
Thanks for the contribution. 
If you want us to have a closer look at your contribution, please  check your commit so the integration tests passes. 
As you are changing a behavior of the rule, can you please explain with detailed example what is the goal of your fix ? 

Thanks.

java-checks/src/test/files/checks/EqualsNotOverriddenInSubclassCheck.java
```diff
@@ -76,11 +76,16 @@ public boolean equals(Object obj) {
...
+
+class N extends M {
+  @Override
   public final boolean equals(Object obj) {
     return false;
   }
 }
 
-class N extends M { // Compliant - M.equals() is final
+class O extends N { // Compliant - M.equals() is final
```
benzonico: Please update comment accordingly to your fix. Moreover, please add a dedicated testcase rather than modifying an existing one as this prevents to test for regression.

alowaniak:
@benzonico thanks for replying and my apologies for my rather messy PR, it's my first time making a PR and so I am basically just trying it out.

Regarding the detailed example of the goal of my fix:
Where should this be provided, in these comments?
The test-case is a pretty good example IMO and I'm having a hard time explaining it but:
Before this fix children (that add a field and) that extend a parent with a final equals method will only not be flagged(*) as an issue (because they can obviously not override the equals method) **iff** said parent does not extend a class which also overrides the equals method.
As soon as some class in a hierarchy makes equals final all classes lower in the hierarchy can not override it anymore (regardless of what happens higher in the hierarchy) and this change fixes the rule so it won't mark any of those lower classes.


Regarding the IT's: I can't really get them to run locally (not even without my modifications, I get `ExceptionInInitializerError: null` in `JavaTestSuite`) and I didn't really expect my change to impact anything in IT's (the travis builds look like they error on Authorization and github user?).
So TBH I'm not sure what I can do to fix that.

java-checks/src/test/files/checks/EqualsNotOverriddenInSubclassCheck.java
alowaniak: Fixed it

alowaniak:
Closed because I haven't had the time for a while to look more into it.
Might start via [more proper channels](https://groups.google.com/forum/#!forum/sonarqube) at a later time.

PR #1729, opened 2017-10-09
SONARJAVA-2512 handle INVOKEDYNAMIC instruction

saberduck:


java-frontend/src/test/java/org/sonar/java/bytecode/se/BytecodeEGWalkerExecuteTest.java
```diff
@@ -698,6 +698,21 @@ public void test_multianewarray() throws Exception {
...
+  public void test_invoke_dynamic() throws Exception {
+    SymbolicValue lambdaArg = new SymbolicValue();
+    ProgramState programState = execute(new Instruction.InvokeDynamicInsn("(I)Ljava/util/function/Supplier;"), ProgramState.EMPTY_STATE.stackValue(lambdaArg));
+    assertThat(programState.peekValue()).isNotNull();
+    assertThat(programState.peekValue()).isNotEqualTo(lambdaArg);
+
+    programState = execute(new Instruction.InvokeDynamicInsn("()Ljava/util/function/Supplier;"), ProgramState.EMPTY_STATE);
+    assertThat(programState.peekValue()).isNotNull();
+
+    // this is theoretical, in practice lambda will always evaluate to target
```
benzonico: I don't understand this comment : what is target in this context ? What are the following lines testing? edge case ?

java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
```diff
@@ -385,7 +385,9 @@ void executeInstruction(Instruction instruction) {
         }
         break;
       case INVOKEDYNAMIC:
-        // TODO SONARJAVA-2512
+        Instruction.InvokeDynamicInsn invokeDynamicInsn = (Instruction.InvokeDynamicInsn) instruction;
+        pop = popStack(invokeDynamicInsn.arity(), instruction.opcode);
+        programState = invokeDynamicInsn.hasReturnValue() ? pop.state.stackValue(new SymbolicValue()) : pop.state;
```
benzonico: Matter of style  : 

``` 
programState = pop.state;
if(instr.hasReturnValue() {
programState = programState.stackValue(new Sv());
 }
```

java-frontend/src/main/java/org/sonar/java/bytecode/cfg/Instruction.java
```diff
@@ -174,4 +174,45 @@ public int hashCode() {
...
+      return methodType.getArgumentTypes().length;
+    }
+
+    @Override
+    public boolean hasReturnValue() {
+      return Type.getMethodType(desc).getReturnType() != Type.VOID_TYPE;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
benzonico: There is an overall lack of tests regarding equals/hashcode of Instruction. Worth a dedicated commit or even PR.

java-frontend/src/test/java/org/sonar/java/bytecode/se/BytecodeEGWalkerExecuteTest.java
saberduck: target is functional interface to which lambda expression is converted, and this in fact can never be void. I decided to remove this part, as this is indeed impossible to happen due to lambda semantic. I also added `NOT_NULL` constraint on the target functional interface, as this is in fact mandated by the spec, see runtime exceptions paragraph in https://docs.oracle.com/javase/specs/jvms/se9/html/jvms-6.html#jvms-6.5.invokedynamic

java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
saberduck: condition removed altogether, because as mentioned above the false branch was in fact impossible

java-frontend/src/main/java/org/sonar/java/bytecode/cfg/Instruction.java
saberduck: tests added

PR #1732, opened 2017-10-11
SONARJAVA-2508 handle long and double values

saberduck:


java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
```diff
@@ -431,6 +497,34 @@ void executeInstruction(Instruction instruction) {
...
 
+  private static ProgramState stackValues(ProgramState.Pop pop, int... values) {
+    ProgramState ps = pop.state;
+    for (int value : values) {
+      ps = ps.stackValue(pop.values.get(value));
+    }
+    return ps;
+  }
+
+  private ProgramState setDoubleOrLong(SymbolicValue sv) {
```
m-g-sonar: I think I would prefer to have this one removed, it is only used once and add one indirection for not much gain.

java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
```diff
@@ -323,21 +376,28 @@ void executeInstruction(Instruction instruction) {
...
-      case I2F:
       case I2D:
-      case L2I:
-      case L2F:
-      case L2D:
-      case F2I:
       case F2L:
       case F2D:
+        sv = programState.peekValue();
+        Preconditions.checkNotNull(sv, "%s needs value on stack", instruction.opcode);
```
m-g-sonar: This is going to display the numerical value, and not the real code, right?

saberduck: fixed by relying on `Instruction.toString()` , also on few other places.

PR #1735, opened 2017-10-12
SONARJAVA-2521 Filter out  behaviors of non-overrideable methods

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
```diff
@@ -239,12 +233,52 @@ void setTrueBlock(Block trueBlock) {
...
+
+    public BytecodeCFGClassVisitor(BytecodeCFGMethodVisitor methodVisitor, String targetedMethodSignatures) {
+      super(Opcodes.ASM5);
+      this.methodVisitor = methodVisitor;
+      this.methodSignature = targetedMethodSignatures;
+    }
+
+    @Override
+    public void visit(int version, int access, String name, String signature, String superName, String[] interfaces) {
+      isFinalClass = Flags.isFlagged(Flags.filterAccessBytecodeFlags(access), Flags.FINAL);
```
benzonico: Why filtering ?

java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
```diff
@@ -239,12 +233,52 @@ void setTrueBlock(Block trueBlock) {
...
+        methodVisitor.isStaticMethod = Flags.isFlagged(methodFlags, Flags.STATIC);
+        methodVisitor.isVarArgs = Flags.isFlagged(methodFlags, Flags.VARARGS);
+        methodVisitor.isNonOverrideableMethod = methodCanNotBeOverriden(methodFlags);
+        return new JSRInlinerAdapter(methodVisitor, access, name, desc, signature, exceptions);
+      }
+      return null;
+    }
+
+    private boolean methodCanNotBeOverriden(int methodFlags) {
+      if (Flags.isFlagged(methodFlags, Flags.NATIVE)) {
```
benzonico: This is not aligned with the name of the method. Worth a comment or renaming the method.

java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
```diff
@@ -239,12 +233,52 @@ void setTrueBlock(Block trueBlock) {
...
+    }
+  }
+
   private static class BytecodeCFGMethodVisitor extends MethodVisitor {
     Map<Label, Block> blockByLabel = new HashMap<>();
     private Block currentBlock;
     private BytecodeCFG cfg;
     private boolean isStaticMethod;
     private boolean isVarArgs;
+    private boolean isNonOverrideableMethod;
```
benzonico: I would go for isOverrideable... double negation is hard. And in fact that would be : isOverrideableOrNative

PR #1739, opened 2017-10-13
SONARJAVA-2522 add constraint for objects created with NEW and use it…

saberduck:
… in ATHROW

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
+++ b/java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
@@ -392,7 +392,7 @@ public class BytecodeCFGBuilder {
     }
 
     @Override
-    public void visitTryCatchBlock(Label start, Label end, Label handler, String type) {
+    public void visitTryCatchBlock(Label start, Label end, Label handler, @Nullable String type) {
       String exception = type == null ? null : Type.getObjectType(type).getClassName();
       tryCatchBlocks.add(new TryCatchBlock(start, end, handler, exception));
     }
--- a/java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
+++ b/java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
@@ -393,7 +393,8 @@ public class BytecodeCFGBuilder {
 
     @Override
     public void visitTryCatchBlock(Label start, Label end, Label handler, String type) {
-      tryCatchBlocks.add(new TryCatchBlock(start, end, handler, type));
+      String exception = type == null ? null : Type.getObjectType(type).getClassName();
+      tryCatchBlocks.add(new TryCatchBlock(start, end, handler, exception));
     }
 
     private static class TryCatchBlock {
@@ -406,7 +407,7 @@ public class BytecodeCFGBuilder {
       @Nullable
       private final String type;
 
-      TryCatchBlock(Label start, Label end, Label handler, String type) {
+      TryCatchBlock(Label start, Label end, Label handler, @Nullable String type) {
         this.start = start;
         this.end = end;
         this.handler = handler;
--- a/java-frontend/src/main/java/org/sonar/java/se/SymbolicExecutionVisitor.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/SymbolicExecutionVisitor.java
@@ -32,29 +32,34 @@ import org.sonar.java.se.symbolicvalues.RelationalSymbolicValue;
 import org.sonar.java.se.xproc.BehaviorCache;
 import org.sonar.java.se.xproc.MethodBehavior;
 import org.sonar.plugins.java.api.JavaFileScanner;
+import org.sonar.plugins.java.api.JavaFileScannerContext;
 import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
-import javax.annotation.Nullable;
 import java.util.List;
 
 public class SymbolicExecutionVisitor extends SubscriptionVisitor {
   private static final Logger LOG = Loggers.get(SymbolicExecutionVisitor.class);
 
   @VisibleForTesting
-  public final BehaviorCache behaviorCache;
+  public BehaviorCache behaviorCache;
   private final ExplodedGraphWalker.ExplodedGraphWalkerFactory egwFactory;
-  private final SemanticModel semanticModel;
+  private SquidClassLoader classLoader;
 
   public SymbolicExecutionVisitor(List<JavaFileScanner> executableScanners) {
-    this(executableScanners, new SquidClassLoader(Lists.newArrayList()), null);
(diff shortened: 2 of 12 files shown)


java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
```diff
@@ -393,7 +393,8 @@ public void visitLabel(Label label) {
 
     @Override
     public void visitTryCatchBlock(Label start, Label end, Label handler, String type) {
-      tryCatchBlocks.add(new TryCatchBlock(start, end, handler, type));
+      String exception = type == null ? null : Type.getObjectType(type).getClassName();
```
SonarTech: MAJOR Change this condition so that it does not always evaluate to "false" [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS2583)

SonarTech:
* ![MAJOR][MAJOR] 1 major

java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
saberduck: I am fixing this

java-frontend/src/test/java/org/sonar/java/bytecode/se/BytecodeEGWalkerExecuteTest.java
```diff
@@ -79,6 +83,25 @@
...
 
+  private static final String TRY_CATCH_SIGNATURE = BytecodeEGWalkerExecuteTest.class.getCanonicalName() + "#tryCatch(Z)V";
+  private static final String TRY_WRONG_CATCH_SIGNATURE = BytecodeEGWalkerExecuteTest.class.getCanonicalName() + "#tryWrongCatch(Z)V";
+
+
+  SemanticModel semanticModel;
+  BytecodeEGWalker walker;
+  private SquidClassLoader squidClassLoader;
+
+  public void setUp() {
```
benzonico: I would challenge the name. please refine it:  setup in context of Junit test means something and it is confusing to have a method named liked this and not annotated with @Before

java-frontend/src/test/java/org/sonar/java/bytecode/se/BytecodeEGWalkerExecuteTest.java
```diff
@@ -966,25 +995,26 @@ public void test_enqueuing_only_happy_path() {
...
+    BytecodeCFGBuilder.Block b2 = cfg.blocks().get(2);
     walker.programState = ProgramState.EMPTY_STATE.stackValue(new SymbolicValue()).stackValue(new SymbolicValue());
     walker.programPosition = new ProgramPoint(b2).next().next();
     walker.executeInstruction(b2.elements().get(3));
-    //FIXME should be 2 as the path to the catch should not be enqueued : should be fix by tracking properly exception types
-    assertThat(walker.workList).hasSize(3);
-    assertThat(walker.workList.pop().programState.exitValue()).isNotNull().isInstanceOf(SymbolicValue.ExceptionalSymbolicValue.class);
+
+    assertThat(walker.workList).hasSize(2);
     assertThat(walker.workList.pop().programState.exitValue()).isNotNull().isInstanceOf(SymbolicValue.ExceptionalSymbolicValue.class);
```
benzonico: May be worth asserting the exception type in that case ?

java-frontend/src/main/java/org/sonar/java/se/xproc/BehaviorCache.java
```diff
@@ -38,12 +39,18 @@
 
   private final SymbolicExecutionVisitor sev;
   private final SquidClassLoader classLoader;
+  private final SemanticModel semanticModel;
```
benzonico: Given the structure of the constructor this field can be null. 
It is only used to be passed to a bytecodeEGWalker which expects it to be non null. (No check of nullity are made on EGWalker side). 
I tend to think that the overloading of constructor is not required here (or at the very least should be annotated with @VisibleForTesting but for sake of simplicity I would avoid the overloading and update tests accordingly.

java-frontend/src/main/java/org/sonar/java/se/constraint/TypedConstraint.java
```diff
@@ -19,6 +19,33 @@
...
 
+import org.sonar.plugins.java.api.semantic.Type;
+
+import javax.annotation.Nullable;
+import java.util.Objects;
+
 public class TypedConstraint implements Constraint {
-  // Empty class for now, but should store the resolved type for instanceof operator.
+  @Nullable
+  public final Type type;
```
benzonico: I am wondering about not having this field nullable and rather rely on Symbols.unknownType

java-frontend/src/main/java/org/sonar/java/se/symbolicvalues/SymbolicValue.java
```diff
@@ -280,7 +280,7 @@ public String toString() {
         if (ps.size() == 1 && ps.get(0).equals(programState)) {
           // FIXME we already know that operand is NOT NULL, so we add a different constraint to distinguish program state. Typed Constraint
           // should store the deduced type.
-          return ImmutableList.of(programState.addConstraint(this, new TypedConstraint()));
+          return ImmutableList.of(programState.addConstraint(this, new TypedConstraint(null)));
```
benzonico: Symbols.unknownType seems more appropriate than null (see comment above)

java-frontend/src/main/java/org/sonar/java/se/SymbolicExecutionVisitor.java
```diff
@@ -64,7 +68,7 @@ public void visitNode(Tree tree) {
   }
 
   public void execute(MethodTree methodTree) {
-    ExplodedGraphWalker walker = egwFactory.createWalker(behaviorCache, (SemanticModel) context.getSemanticModel());
+    ExplodedGraphWalker walker = egwFactory.createWalker(behaviorCache, semanticModel);
```
benzonico: This makes me wonder a bit if we are not creating a new way to access semantic model rather than relying on what's existing.

saberduck: I reworked it, so semantic model is acquired from scanner context

java-frontend/src/main/java/org/sonar/java/se/SymbolicExecutionVisitor.java
```diff
@@ -42,15 +43,23 @@
...
+
   public SymbolicExecutionVisitor(List<JavaFileScanner> executableScanners, SquidClassLoader classLoader) {
-    behaviorCache = new BehaviorCache(this, classLoader);
     egwFactory = new ExplodedGraphWalker.ExplodedGraphWalkerFactory(executableScanners);
+    this.classLoader = classLoader;
+  }
+
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    behaviorCache = new BehaviorCache(this, classLoader, (SemanticModel) context.getSemanticModel());
```
benzonico: getSemanticModel might return a null value (both corner case in case of semantic analysis failure (can happen) or when analyzing some jdk classes (less likely)).

PR #1740, opened 2017-10-16
Parse bytecode listing for generating cfg

benzonico:


java-frontend/src/test/java/org/sonar/java/bytecode/cfg/BytecodeListingParserTest.java
```diff
@@ -0,0 +1,97 @@
...
+    assertCFG("L0\n" +
+      "ICONST_0\n",
+      "B0(Exit)\n" +
+        "B1\n" +
+        "0: ICONST_0\n" +
+        "Jumps to: B0 \n");
+  }
+
+  @Test
+  public void name() throws Exception {
```
m-g-sonar: name?

benzonico: // TODO  : find a better name than name.

PR #1744, opened 2017-10-17
SONARJAVA-2533 Fix handling of catch handlers

benzonico:


java-frontend/src/main/java/org/sonar/java/bytecode/cfg/BytecodeCFGBuilder.java
```diff
@@ -389,6 +395,9 @@ public void visitLabel(Label label) {
           previous.successors.add(currentBlock);
         }
       }
+      currentTryCatches.addAll(handlersStartingWith(label));
+      currentTryCatches.removeAll(handlersEndingWith(label));
```
m-g-sonar: It seems to me that we are removing the handlers too soon, before wiring them. If the handler has this label as END, then it should still be wired. However, because we remove it from the list, I think that we are going to miss it.

PR #1756, opened 2017-10-23
SONARJAVA-2506 include JAXB framework to avoid dependency on JDK module

saberduck:


pom.xml
```diff
@@ -95,6 +95,7 @@
 
     <!-- Release: enable publication to Bintray -->
     <artifactsToPublish>${project.groupId}:sonar-java-plugin:jar</artifactsToPublish>
+    <version.shade.plugin>3.1.0</version.shade.plugin>
```
benzonico: I think this one deserves a comment. At the very least to know when to remove it.

sonar-java-plugin/pom.xml
```diff
@@ -156,6 +156,54 @@
                     <include>**</include>
                   </includes>
                 </filter>
+                <filter>
```
benzonico: I think this filter configuration also deserves a comment.

pom.xml
```diff
@@ -228,6 +229,23 @@
...
+        <groupId>javax.xml.bind</groupId>
+        <artifactId>jaxb-api</artifactId>
+        <version>2.3.0</version>
+      </dependency>
+      <dependency>
+        <groupId>org.glassfish.jaxb</groupId>
+        <artifactId>jaxb-runtime</artifactId>
+        <version>2.3.0</version>
+      </dependency>
+      <dependency>
```
benzonico: While I can understand easily the runtime I am unsure of why javax.activation is required. Comment ?

saberduck: jaxb depends on `javax.activation.DataSource` , I don't know why this is not transitive dependency but it will fail when not provided.

pom.xml
```diff
@@ -95,6 +95,8 @@
 
     <!-- Release: enable publication to Bintray -->
     <artifactsToPublish>${project.groupId}:sonar-java-plugin:jar</artifactsToPublish>
+    <!-- Override value from parent pom, needed to process jars compiled with Java 9 (mainly JAXB) -->
+    <version.shade.plugin>3.1.0</version.shade.plugin>
```
benzonico: I am wondering if we require this, why not upgrade the version in the parent pom ? cc @drautureau-sonarsource

drautureau-sonarsource: Indeed I just need to release the next version of the parent

benzonico: @drautureau-sonarsource any ETA ?

drautureau-sonarsource: Parent 41 is available

PR #1765, opened 2017-10-30
SONARJAVA-2466 introduce sonarpedia.json and update rules metadata

m-g-sonar:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S1212_java.html
```diff
@@ -6,6 +6,4 @@
   <li> forbid access to <code>java.util.Vector</code>, <code>java.util.Hashtable</code> and <code>java.util.Enumeration</code> from any classes </li>
   <li> forbid access to <code>java.sql.**</code> from <code>**.ui.**</code> and <code>**.web.**</code> classes </li>
 </ul>
-<h2>Deprecated</h2>
-<p>This rule is deprecated, and will eventually be removed.</p>
```
benzonico: Why remove this ? rule is actually deprecated : https://jira.sonarsource.com/browse/RSPEC-1212

sonarpedia.json
```diff
@@ -0,0 +1,7 @@
+{
+  "rules-metadata-path": "./java-checks/src/main/resources/org/sonar/l10n/java/rules/squid",
+  "languages": [
+    "JAVA"
```
benzonico: I am surprised by the case of the language in this file cc @erichirlemann

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/Sonar_security_way_profile.json
```diff
@@ -1 +0,0 @@
-{"name":"Sonar security way","ruleKeys":[]}
```
benzonico: :+1:

sonarpedia.json
benzonico: I mean that we have nowhere the java language used in uppercase. The langage key is "java" as defined per https://github.com/SonarSource/sonar-java/blob/master/sonar-java-plugin/src/main/java/org/sonar/plugins/java/Java.java#L39

erichirlemann: This key comes from rule-api instead.

PR #1768, opened 2017-11-01
Setup property 'sonar.java.xfile' to enable x-file on demand only

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/JavaSquid.java
```diff
@@ -88,12 +88,15 @@ public JavaSquid(JavaVersion javaVersion,
     //AstScanner for main files
     ActionParser<Tree> parser = JavaParser.createParser();
     astScanner = new JavaAstScanner(parser, sonarComponents);
-    boolean enableSymbolicExecution = hasASymbolicExecutionCheck(visitors);
-    astScanner.setVisitorBridge(createVisitorBridge(codeVisitors, classpath, javaVersion, sonarComponents, enableSymbolicExecution));
+    SymbolicExecutionMode mode = SymbolicExecutionMode.DISABLED;
```
benzonico: This bit of code could be part of the enum as well (if we stick to this).

PR #1789, opened 2017-11-28
SONARJAVA-2565 Unchecked exceptions should be enqueued

saberduck:


java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
```diff
@@ -790,6 +791,35 @@ private boolean handleMethodInvocation(Instruction instruction) {
...
+  private boolean isUncheckedExceptionCatchBlock(BytecodeCFG.Block b) {
+    String exceptionTypeName = b.getExceptionType();
+    if (exceptionTypeName == null) {
+      return false;
+    }
+    Type exceptionType = semanticModel.getClassType(exceptionTypeName);
+    if (exceptionType == null) {
+      return false;
+    }
+    return exceptionType.isSubtypeOf("java.lang.RuntimeException")
```
m-g-sonar: I think I would prefer to have this part shared between both `EGWalker`. In order to be sure that we are handling unchecked exceptions the same way. Maybe by putting it as a `static` method in `ExceptionalSymbolicValue` ?

```
public static boolean isUncheckedException(@Nullable Type exceptionType) {
  if (exceptionType == null) {
    return false;
  }
  return exceptionType.isSubtypeOf("java.lang.RuntimeException")
    || exceptionType.isSubtypeOf("java.lang.Error")
    || exceptionType.is("java.lang.Exception")
    || exceptionType.is("java.lang.Throwable");
}
```

PR #1811, opened 2017-12-05
SONARJAVA-2577 provide override of write bytes

benzonico:


java-checks/src/main/java/org/sonar/java/checks/OutputStreamOverrideWriteCheck.java
```diff
@@ -0,0 +1,71 @@
...
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S4349")
+public class OutputStreamOverrideWriteCheck extends IssuableSubscriptionVisitor {
+
+  private static final MethodMatcher WRITE_BYTES_INT_INT = MethodMatcher.create().typeDefinition(TypeCriteria.anyType()).name("write")
+    .addParameter("byte[]")
```
m-g-sonar: I would prefer using `parameters("byte[]", "int", "int")` instead,  which more concise.

java-checks/src/main/java/org/sonar/java/checks/OutputStreamOverrideWriteCheck.java
```diff
@@ -0,0 +1,71 @@
...
+      return;
+    }
+    Optional<MethodTree> writeByteIntInt = classTree.members()
+      .stream()
+      .filter(m -> m.is(Tree.Kind.METHOD))
+      .map(m -> (MethodTree) m)
+      .filter(WRITE_BYTES_INT_INT::matches)
+      .findFirst();
+
+    if(!writeByteIntInt.isPresent()) {
```
m-g-sonar: I think it's not enough, because in case of a abstract direct subtypes of `OutputStream`, and not implementing `write(int b)` this may raise a FP. Why asking for implementing `write(byte[], int, int)`, while `write(int)` is not even implemented?

For direct subtypes of `FilterOutputStream` it should be enough, as the method is implemented.

benzonico: Your concern is legit, given the very narrow domain of application of the rule I would suggest to try the naive approach at first and wait for feedback.
Because this would be a FP only if abstract class is not overriding and the subclasses are implementing.

m-g-sonar: Mmh. I don't like the idea of waiting for feedback in such cases. I would prefer the opposite approach and avoid the FP directly, right now. As it's a possible case, even really unlikely. We can predict it, so we can avoid it. Removing any potential noise.

java-checks/src/main/java/org/sonar/java/checks/OutputStreamOverrideWriteCheck.java
```diff
@@ -0,0 +1,81 @@
...
+        if (writeIntTree.block().body().isEmpty()) {
+          message = "Provide an empty override of \"write(byte[],int,int)\" for this class as well.";
+        }
+      }
+      reportIssue(className, message);
+    }
+
+  }
+
+  private Optional<MethodTree> findMethod(ClassTree classTree, MethodMatcher methodMatcher) {
```
SonarTech: MAJOR Make "findMethod" a "static" method. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS2325)

SonarTech:
* ![MAJOR][MAJOR] 1 major

PR #1812, opened 2017-12-05
SONARJAVA-2120 Handle boxing on same type using valueOf()

m-g-sonar:


java-checks/src/test/files/checks/ImmediateReverseBoxingCheck.java
```diff
@@ -1,16 +1,18 @@
...
+  void intBoxingAndUnboxing(int int1, Integer integer1, String string, Double double1) {
     new Integer(int1).intValue(); // Noncompliant [[sc=5;ec=22]] {{Remove the boxing of "int1".}}
     new Integer(1 + 2).intValue(); // Noncompliant [[sc=5;ec=23]] {{Remove the boxing to "Integer".}}
     Integer.valueOf(int1).intValue(); // Noncompliant
     Integer.valueOf(1 + 2).intValue(); // Noncompliant
     processInt(string, new Integer(int1)); // Noncompliant
     new Integer(int1).toString();
     integer1.intValue();
+    Integer.valueOf(integer1); // Noncompliant [[sc=5;ec=20]] {{Remove the boxing to 'Integer'; The argument is already of type Integer.}}
+    Integer.valueOf(double1); // Compliant
```
benzonico: I would add the case with Long.valueOf (to justify the generalization you made in the code).

java-checks/src/test/files/checks/ImmediateReverseBoxingCheck.java
```diff
@@ -1,16 +1,18 @@
 abstract class A {
-  
+
   public void processInt(String s, int a) {}
   public void processInteger(String s, Integer... a) {}
-  
-  void intBoxingAndUnboxing(int int1, Integer integer1, String string) {
+
+  void intBoxingAndUnboxing(int int1, Integer integer1, String string, Double double1) {
```
benzonico: A case I don't see and I am wondering about covering it or not : 
```
Integer integer1 = 1;
new Integer(integer1); // shall raise an issue or not ? 
```
I think issue should be legit most of the time the only doubt I can think of is if you want to have two instances of the same value (but that seems a corner case). If you think we should cover that one, RSPEC might require a minor update.

m-g-sonar: I would prefer not to raise an issue this. Like you said, if you want another instance with the same value, it would be the only way to have it without raising an issue, because `new Integer(integer1.intValue());` is already raising an issue related to unboxing (line 27 from test file).

benzonico: But this huge case seems highly arguable whereas detecting such issue when it is not intended will be valuable.

m-g-sonar: Okay, so let's handle it. Adding the case to RSPEC and updating code.

PR #1818, opened 2017-12-07
SONARJAVA-2569 Rule S2444 updated to not raise issue if no synchronized methods

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/StaticFieldInitializationCheck.java
```diff
@@ -46,55 +50,92 @@
...
+  private boolean isInStaticInitializer() {
+    return withinStaticInitializer.peek();
+  }
+
+  private static Boolean hasSynchronizedMethod(ClassTree tree) {
+    return tree.members().stream()
+      .filter(member -> member.is(Tree.Kind.METHOD))
+      .map(MethodTree.class::cast)
+      .map(MethodTree::modifiers)
+      .anyMatch(modifiers -> ModifiersUtils.hasModifier(modifiers, Modifier.SYNCHRONIZED));
```
benzonico: flatMap(MethodTree::modifiers).anyMatch(Modifier.SYNCHRONIZED::equals) ? avoid the call to ModifiersUtils.

java-checks/src/test/files/checks/StaticFieldInitializationCheck.java
```diff
@@ -55,6 +55,17 @@ public static int getInstance5() {
     instance5 = 12;  // compliant - primitive type
     return instance5;
   }
+
+  protected static Object instance6 = null;
```
benzonico: arguable : I don't see the testcase which will make you require the stack.

java-checks/src/main/java/org/sonar/java/checks/StaticFieldInitializationCheck.java
m-g-sonar: `Modifier.SYNCHRONIZED` is not a tree, contrarily to the modifiers from the list. the steps to get there make it less clear from my point of view.

java-checks/src/test/files/checks/StaticFieldInitializationCheck.java
m-g-sonar: Classes B,C, and D from test files are nested classes of A, which has a synchronized method.

PR #1822, opened 2017-12-08
SONARJAVA-2488 RSPEC-3824 Fix fp when value is invocation throwing a …

benzonico:
…checked exception

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/se/checks/MapComputeIfAbsentOrPresentCheck.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/checks/MapComputeIfAbsentOrPresentCheck.java
@@ -28,6 +28,7 @@ import org.sonar.java.cfg.CFG;
 import org.sonar.java.matcher.MethodMatcher;
 import org.sonar.java.matcher.TypeCriteria;
 import org.sonar.java.model.ExpressionUtils;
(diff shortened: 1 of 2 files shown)


java-frontend/src/main/java/org/sonar/java/se/checks/MapComputeIfAbsentOrPresentCheck.java
```diff
@@ -102,20 +103,28 @@ public ProgramState checkPreStatement(CheckerContext context, Tree syntaxNode) {
...
-        mapGetInvocations.get(mapSV).stream()
-          .filter(getOnSameMap -> getOnSameMap.withSameKey(keySV))
-          .findAny()
-          .ifPresent(getOnSameMap -> {
-            ObjectConstraint constraint = ps.getConstraint(getOnSameMap.value, ObjectConstraint.class);
-            if (constraint != null && isInsideIfStatementWithNullCheckWithoutElse(mit)) {
-              checkIssues.add(new CheckIssue(context.getNode(), getOnSameMap.mit, mit, getOnSameMap.value, constraint));
-            }
-          });
+        if(!isMethodInvocationThrowingCheckedException(mit.arguments().get(1))) {
```
saberduck: I would prefer to merge the `if` with the one on the line 100 , logically it's equivalent and I believe also more clear

PR #1823, opened 2017-12-08
SONARJAVA-2524 Fix fp when duplication appears only in constants

benzonico:


java-checks/src/test/files/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -23,11 +23,14 @@ default void foo() {
...
-
+class AllConstants {
+  private static final String Constant1 = "allConstant";
+  private static final String Constant2 = "allConstant";
+}
 class ConstantAlreadyDefined {
 
   private static final String A = "constant";
-  private static final String B = "constant"; // Noncompliant [[secondary=31,36]] {{Use already-defined constant 'A' instead of duplicating its value here.}}
+  private static final String B = "constant"; // Noncompliant [[secondary=34,39]] {{Use already-defined constant 'A' instead of duplicating its value here.}}
```
saberduck: Duplication should not be detected when literal is used to initialize another constant. This issue should only be raised on line 34.

java-checks/src/main/java/org/sonar/java/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -65,6 +65,16 @@ public void scanFile(JavaFileScannerContext context) {
     scan(context.getTree());
     for (String entry : occurrences.keySet()) {
       Collection<LiteralTree> literalTrees = occurrences.get(entry);
+      boolean allInConstants = context.getSemanticModel() != null && literalTrees.stream().allMatch(t -> {
```
saberduck: I think it is simpler to ignore "constants" during the visit, i.e. cut the visit in the ``visitVariable`` method, so constants are not counted in ``occurences``

java-checks/src/test/files/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -23,18 +23,22 @@ default void foo() {
...
+  private static final String B = "constant";
   private static final String C = "constant";
 
   private static final String REPORT_WITHOUT_THRESHOLD = "blabla";
 
   void test() {
+    System.out.println("constant"); // Noncompliant [[secondary=40]] {{Use already-defined constant 'A' instead of duplicating its value here.}}
     System.out.println("constant");
-    System.out.println("blabla"); // Noncompliant {{Use already-defined constant 'REPORT_WITHOUT_THRESHOLD' instead of duplicating its value here.}}
+    System.out.println("blabla");
```
saberduck: Why is this not reported anymore?  Condition on line 68 should be changed for `> 0`

PR #1824, opened 2017-12-11
SONARJAVA-2591 Fix method yields on invocation check and order of debug rules execution

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/debug/DebugMethodYieldsOnInvocationsCheck.java
```diff
@@ -50,23 +52,20 @@
   tags = "debug")
 public class DebugMethodYieldsOnInvocationsCheck extends SECheck implements DebugCheck {
 
-  private Map<IdentifierTree, MethodBehavior> methodBehaviors = new HashMap<>();
+  private Deque<Map<IdentifierTree, Symbol.MethodSymbol>> methodSymbols = new LinkedList<>();
```
saberduck: I think we don't really need a map between identifiers and method symbols, we can just keep list of method invocations and get symbols in `reportAll` method

java-frontend/src/main/java/org/sonar/java/se/checks/debug/DebugMethodYieldsOnInvocationsCheck.java
```diff
@@ -82,20 +81,25 @@ private static IdentifierTree getIdentifier(ExpressionTree methodSelect) {
...
   }
 
-  private void reportAll() {
-    methodBehaviors.entrySet().forEach(entry -> {
-      MethodBehavior mb = entry.getValue();
-      IdentifierTree methodName = entry.getKey();
-      Set<Flow> flow = flowFromYield(mb, methodName);
-      reportIssue(methodName, String.format("Method '%s' has %d method yields.", methodName.name(), mb.yields().size()), flow);
+  private void reportAll(CheckerContext context) {
+    Map<Symbol.MethodSymbol, MethodBehavior> behaviors = new HashMap<>();
```
saberduck: I think this cache is not necessary, call to `CheckerDispatcher.peekMethodBehavior` is cheap, so no need to add another cache

PR #1831, opened 2017-12-14
SONARJAVA-2545

benzonico:


java-checks/src/test/files/checks/LDAPInjectionCheck.java
```diff
@@ -2,6 +2,10 @@
 import javax.naming.directory.DirContext;
 import javax.naming.directory.InitialDirContext;
 import javax.naming.directory.SearchControls;
+import java.util.ArrayList;
```
m-g-sonar: This should have been part of the second commit.

PR #1837, opened 2017-12-18
SONARJAVA-2594 enqueue exit blocks in bytecode analysis

saberduck:
Because exit blocks are now properly enqueued when doing bytecode analysis, many more yields will have known exception types, which required change in checks how exit from the method on exception is handled. Now this decision is delegated to the individual checks.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
+++ b/java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
@@ -228,6 +228,7 @@ public class BytecodeEGWalker {
 
   @VisibleForTesting
   ExplodedGraph explodedGraph;
+  private BytecodeCFG.Block exitBlock;
 
   /**
    * Because some instructions manipulate stack differently depending on the type of the value, we need this constraint to know category of the value
@@ -323,6 +324,7 @@ public class BytecodeEGWalker {
     if (bytecodeCFG == null) {
       return;
     }
+    exitBlock = bytecodeCFG.exitBlock();
     steps = 0;
     for (ProgramState startingState : startingStates(signature, ProgramState.EMPTY_STATE, lookup.isStatic)) {
       enqueue(new ProgramPoint(bytecodeCFG.entry()), startingState);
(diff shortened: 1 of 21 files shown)


java-frontend/src/main/java/org/sonar/java/bytecode/se/BytecodeEGWalker.java
```diff
@@ -817,11 +822,26 @@ private ProgramState stateWithException(ProgramState programState, BytecodeCFG.B
...
+        .filter(b -> isExceptionHandledByBlock(exceptionType, b))
+        .collect(Collectors.toList());
+    if (!blocksCatchingException.isEmpty()) {
+      blocksCatchingException.forEach(b -> enqueue(new ProgramPoint(b), ps));
+      if (isCatchExhaustive(exceptionType, blocksCatchingException)) {
+        return;
+      }
+    }
+    // exception was not handled or was handled only partially, enqueue exit block with exceptional SV
+    ps.storeExitValue();
```
m-g-sonar: Are we sure at this point that there is a Exceptional SV on top of the stack? Maybe we should enforce it with a precondition.

java-frontend/src/test/java/org/sonar/java/bytecode/se/BytecodeEGWalkerTest.java
```diff
@@ -256,6 +257,45 @@ public void test_enqueueing_of_catch_blocks2() {
...
 
+  @Test
+  public void test_enqueueing_of_exit_block() {
+    MethodBehavior mb = getMethodBehavior(ExceptionEnqueue.class, "enqueueExitBlock()Z");
+    List<MethodYield> yields = mb.yields();
+    assertThat(yields).hasSize(1);
+    assertThat(mb.happyPathYields().findFirst().isPresent()).isFalse();
+    ExceptionalYield exceptionalYield = mb.exceptionalPathYields().findFirst().get();
+    Type exceptionType = exceptionalYield.exceptionType(semanticModel);
+    assertThat(exceptionType.is("java.io.FileNotFoundException")).isTrue();
```
m-g-sonar: :+1:

java-frontend/src/test/java/org/sonar/java/bytecode/se/testdata/ExceptionEnqueue.java
```diff
@@ -78,6 +78,37 @@ static boolean testCatchBlockEnqueue2() throws IOException {
...
+      int x = 0;
+    }
+    return false;
+  }
+
+  static boolean enqueueExitBlock4() {
+    try {
+      throwSpecificException();
+    } catch (Throwable e) {
+      return true;
```
m-g-sonar: wonder if it would not be more interesting to have `false` here to see that we are correctly jumping into the catch.

saberduck: indeed. I modified the unreachable return below to return `false`, so test remains the same, but it is more explicit.

PR #1838, opened 2017-12-18
SONARJAVA-2578 Add behaviors for methods in org.springframework.util.Assert

saberduck:


java-frontend/src/main/java/org/sonar/java/se/xproc/BehaviorCache.java
```diff
@@ -47,32 +47,44 @@
...
+    "org.apache.commons.lang3.StringUtils#isEmpty",
+    "org.apache.commons.lang3.StringUtils#isNotEmpty",
+    "org.apache.commons.lang3.StringUtils#isBlank",
+    "org.apache.commons.lang3.StringUtils#isNotBlank",
+    "org.apache.commons.lang3.Validate#notEmpty",
+    "org.apache.commons.lang3.Validate#notNull",
+
+    "org.apache.logging.log4j.core.util.Assert#requireNonNull",
+
+    "org.springframework.util.CollectionUtils#isEmpty",
```
m-g-sonar: would have been easier to spot the change if the reordering was in another commit.

saberduck: yeah, sorry for that

PR #1842, opened 2017-12-22
SONARJAVA-2599 Prevent class cast exception

benzonico:
No reproducer for that one but we faced the problem as it was reported by a user.


kingoleg:
It was nice comment :)

PR #1849, opened 2018-01-04
 SONARJAVA-2580 Fix handling of Junit 5.0 tests having wrong class names in reports

m-g-sonar:


java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java
```diff
@@ -124,23 +127,34 @@ private static void sanitize(UnitTestIndex index) {
...
       LOGGER.warn("There is {} test(s) reported with negative time by surefire, total duration may not be accurate.", negativeTimeTestNumber);
     }
   }
 
+  private Map<InputFile, UnitTestClassReport> consolidateResults(Map<String, UnitTestClassReport> indexByClassname) {
+    Map<InputFile, UnitTestClassReport> result = new HashMap<>();
+    indexByClassname.forEach((className, index) -> {
+      InputFile resource = getUnitTestResource(className, index);
+      if (resource != null) {
+        result.putIfAbsent(resource, new UnitTestClassReport());
```
saberduck: minor -  replace `putIfAbsent` with  `computeIfAbsent` and use returned value to avoid the call to `Map.get` on the next line

java-surefire/src/test/resources/org/sonar/plugins/surefire/SurefireSensorTest/ignoreSuiteAsInnerClass/TEST-org.apache.shindig.protocol.TestHandler$Input.xml
```diff
@@ -1,68 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <testsuite errors="0" skipped="1" tests="1" time="0" failures="0" name="org.apache.shindig.protocol.TestHandler$Input">
   <properties>
-    <property value="Java(TM) 2 Runtime Environment, Standard Edition" name="java.runtime.name"/>
```
benzonico: While I understand the reasoning behind this removal : 
- I don't see why we should remove it : it was not a problem as far as I understand
- It is a valid report format/part and as such was a way to test the proper parsing of those report.

I don't think we should revert this but I tend to think it is a bit of a useless change. The only thing I would like to be sure of is that somewhere we have a small list of properties to ensure we parse it without issues.

java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java
benzonico: @m-g-sonar shouldn't this be fixed ?

java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java
```diff
@@ -124,23 +127,34 @@ private static void sanitize(UnitTestIndex index) {
...
-        }
+        save(report, entry.getKey(), context);
       }
     }
     if (negativeTimeTestNumber > 0) {
       LOGGER.warn("There is {} test(s) reported with negative time by surefire, total duration may not be accurate.", negativeTimeTestNumber);
     }
   }
 
+  private Map<InputFile, UnitTestClassReport> consolidateResults(Map<String, UnitTestClassReport> indexByClassname) {
```
benzonico: I tend to think the name is a bit too generic : mapToInputFile would be more appropriate.

java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java
```diff
@@ -164,8 +178,19 @@ protected void saveResults(InputFile testFile, UnitTestClassReport report) {
...
 
-  protected InputFile getUnitTestResource(String classKey) {
-    return javaResourceLocator.findResourceByClassName(classKey);
+  @CheckForNull
+  private InputFile getUnitTestResource(String className, UnitTestClassReport unitTestClassReport) {
+    InputFile resource = javaResourceLocator.findResourceByClassName(className);
+    if (resource == null) {
+      // fall back on testSuite class name (repeated and parameterized tests from JUnit 5.0 are using test name as classname)
+      // Should be fixed with JUnit 5.1, see: https://github.com/junit-team/junit5/issues/1182
+      return unitTestClassReport.getResults().stream()
```
benzonico: two things I struggle to see through or understand are : 
- How will this behave on junit 4 parameterized test ? 
- How will this behave when fix is provided in 5.1 ?

java-surefire/src/test/resources/org/sonar/plugins/surefire/SurefireSensorTest/ignoreSuiteAsInnerClass/TEST-org.apache.shindig.protocol.TestHandler$Input.xml
m-g-sonar: Agreed, let's discard it, it's working well with it and have been working for years. I removed the properties because it's completely ignored by the parser (and include some funny stuff such as the location of jdk on Simon's dropbox), but this part and all the possible properties are indeed some valid entries that we could encounter.

java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java
m-g-sonar: These aspects are indeed missing.
* I'll add a test for junit 4 parameterized tests.
* I'll add a "fake" test case in the report simulating fix of 5.1 and a correct class name.

PR #1855, opened 2018-01-10
SONARJAVA-2538 set nullable constraints on first read of parameter SV

benzonico:


java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -594,7 +585,9 @@ private void visit(Tree tree, @Nullable Tree terminator) {
         executeUnaryExpression(tree);
         break;
       case IDENTIFIER:
-        executeIdentifier((IdentifierTree) tree);
+        if(executeIdentifier((IdentifierTree) tree)) {
+          return;
```
m-g-sonar: I don't like the fact that the method is now interrupting visit. I would appreciate at least a documentation of what means the returned boolean value (it's not at all explicit from the method name)

its/ruling/src/test/resources/sonar-server/squid-S2259.json
```diff
@@ -1,4 +1,7 @@
 {
+'org.sonarsource.sonarqube:sonar-server:src/main/java/org/sonar/server/computation/task/projectanalysis/measure/BestValueOptimization.java':[
+63,
```
m-g-sonar: I don't understand this issue. looks like FP to me.

java-frontend/src/test/files/se/MethodInvocationLeadingToNPE.java
```diff
@@ -62,8 +62,8 @@ static void npeIfArg0IsTrueAndArg1IsNull(boolean arg0, Object arg1, boolean arg2
     }
   }
 
-  static void nullable(@Nullable Object o) {
-    o.toString(); // Noncompliant {{A "NullPointerException" could be thrown; "o" is nullable here.}} flow@foo5 [[order=3]] {{'NullPointerException' is thrown.}}
+  static void nullable(@Nullable Object o) { // flow@foo5 [[order=3]] {{Implies 'o' can be null.}}
```
m-g-sonar: This one is bothering me. The message is not coherent with what we just said in two previous flow messages:

- `implies 'o' is null.`
- `'o' is passed to nullable().`
- `implies 'o' can be null.` (it IS null in this flow)

PR #1864, opened 2018-01-12
Test external contribution

Wohops:


Wohops:
Did not trigger QA, as expected.

PR #1874, opened 2018-01-16
SONARJAVA-2622 SONARJAVA-2621 Define hidden metric for parse error feedback on sonarcloud

benzonico:


java-frontend/src/main/java/org/sonar/java/AnalysisError.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+public class AnalysisError {
+
+  private final String message;
+  private final String cause;
+  private final String filename;
+  private final String type;
```
m-g-sonar: I would expect something a bit more robust/testable than a `String` for the type of analysis error. It would allow to be exhaustive, adding new type when we want to handle them.

if you plan to modify that for the next error, feel free to ignore it.

java-frontend/src/main/java/org/sonar/java/AnalysisError.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public AnalysisError(Exception exception, String filename) {
+    this.message = exception.getMessage();
+    StringWriter sw = new StringWriter();
+    exception.printStackTrace(new PrintWriter(sw));
+    this.cause = sw.toString();
```
m-g-sonar: I would extract the part getting the stack trace in a dedicated method.

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -264,4 +269,18 @@ private Charset getCharset(File file) {
...
     return context.isCancelled();
   }
+
+  public void addAnalysisError(AnalysisError analysisError) {
+    analysisErrors.add(analysisError);
+  }
+
+  public void saveAnalysisErrors() {
+    if (!analysisErrors.isEmpty() && "https://sonarcloud.io".equals(context.config().get("sonar.host.url").orElse(""))) {
+      Metric.Builder metricBuilder = new Metric.Builder("sonarjava_feedback", "SonarJava feedback", Metric.ValueType.DATA);
```
m-g-sonar: I would prefer to have the key of the metric extracted to a constant somewhere, maybe in JavaPlugin class itself.

java-frontend/src/main/java/org/sonar/java/AnalysisError.java
benzonico: Why ?

java-frontend/src/main/java/org/sonar/java/AnalysisError.java
benzonico: For the time being it is hardcoded string but I have in mind to probably make it an enum.

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
benzonico: This was changed to use Server API thanks to @simonbrandhof feedback.

PR #1878, opened 2018-01-19
Add support for @Nested test classes (JUnit5)

olcbean:
This PR verifies that if there is a `@Nested` class with tests, SonarQube will not trigger a FP `S2187` : `Add some tests to this class.`

Also if there are no tests in the `@Nested` class, `S2187` still will be trigegred.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
@@ -164,10 +164,8 @@ public class NoTestInTestClassCheck extends IssuableSubscriptionVisitor {
       .flatMap(i -> getAllMembers(i.symbol()))
       .filter(m -> ((JavaSymbol.MethodJavaSymbol) m).isDefault());
     for (Symbol s : symbol.memberSymbols()) {
-      if (s instanceof Symbol.TypeSymbol) {
-        if (s.metadata().isAnnotatedWith("org.junit.jupiter.api.Nested")) {
-          members = Stream.concat(members, getAllMembers((Symbol.TypeSymbol) s));
-        }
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
```diff
@@ -163,6 +163,13 @@ private boolean isTestMethod(Symbol method) {
     Stream<Symbol.MethodSymbol> defaultMethodsFromInterfaces = symbol.interfaces().stream()
       .flatMap(i -> getAllMembers(i.symbol()))
       .filter(m -> ((JavaSymbol.MethodJavaSymbol) m).isDefault());
+    for (Symbol s : symbol.memberSymbols()) {
+      if (s instanceof Symbol.TypeSymbol) {
```
m-g-sonar: I would prefer if you could use `s.isTypeSymbol()` instead. It would allow to remove the `instanceof` and stay with simpler usage of the semantic API for semantic checking.

You can also combine the two `if` statements in a single one with the conditions combined with `&&`.

java-checks/src/test/files/checks/NoTestInTestClassCheck.java
```diff
@@ -209,3 +210,22 @@ void foo() {}
...
+}
+
+class NoTestsInNestedTest { // Noncompliant {{Add some tests to this class.}}
+  @Nested
+  class NestedClass {
+    public void foo() {
+      Assert.assertTrue(true);
+    }
+  }
+}
```
m-g-sonar: missing new line.

olcbean:
@m-g-sonar thanks for the review and the hints :) 
I just rebased and pushed the changes.

m-g-sonar:
Thanks for the update @olcbean , everything looks perfect now. 

PR #1879, opened 2018-01-21
Removing CatchOfThrowableOrErrorCheck as there are good use cases for doing so

eperez:
The rule CatchOfThrowableOrErrorCheck should be removed as there are good use cases for catching Throwable.
Please see some examples:
https://github.com/spring-projects/spring-framework/blob/master/spring-jdbc/src/main/java/org/springframework/jdbc/support/JdbcUtils.java#L74
https://github.com/apache/tomcat/blob/trunk/java/org/apache/catalina/core/StandardWrapperValve.java#L243

This rule should be removed or people will break correct code.


m-g-sonar:
Nop, we won't drop the rule.

There may be indeed some pretty valid use cases to catch runtime exceptions, but it does not invalidate the rule neither. Their is tons of cases where catching `Throwable` or `Error` is completely overkill and impacting a project negatively. See [CEW-396](http://cwe.mitre.org/data/definitions/396.html) for more details.

For cases where you judge that the rule wrong, closing the issues as `Won't Fix` in SonarQube UI is an appropriate solution (this option is there for that), or mark it as False Positive. If you see particular patterns or situations which it is systematically a good practice to catch `Throwable` (undocumented behaviors, bad API design), then report the issue on our mailing list and we will try to address it by **fixing the rule**.

Cheers,
Michael

eperez:
If you are interested in disabling this rule and other Sonar bogus rules, take a look at the Spring Framework default Sonar rules, they have more development experience and they already disabled this rule and other bogus rules that will reduce your code quality if enabled.
https://github.com/joshlong/spring-framework-1/wiki/Sonar-integration

benzonico:
@eperez First of all, please be kind. Your are entitled to your opinion but snarky comments and passive aggressive tone is simply not acceptable. 

As mentioned in the previous comment closing this PR : if you want to reach us to discuss things, there is a space for it : https://discuss.sonarsource.com/ where you are welcome with a constructive remark or critic of rules with clear separation between opinion and facts.

I am locking this conversation as there is no way we can carry on in a constructive manner here.

PR #1880, opened 2018-01-22
SONARJAVA-2478 Swallow and log exceptions from checks

benzonico:
Include fix of NPE introduced with the previous related tickets.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
+++ b/java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
@@ -20,9 +20,11 @@
 package org.sonar.java.model;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Throwables;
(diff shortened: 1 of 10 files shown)


java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
```diff
@@ -119,14 +124,39 @@ public void visitFile(@Nullable Tree parsedTree) {
       new SymbolicExecutionVisitor(executableScanners, behaviorCache).scanFile(javaFileScannerContext);
       behaviorCache.cleanup();
     }
-    for (JavaFileScanner scanner : executableScanners) {
-      scanner.scanFile(javaFileScannerContext);
-    }
+    runChecks(javaFileScannerContext);
```
m-g-sonar: From how I see it, we are not covering here exception occurring during SE (a few lines aboves).
We should handle them as well.

PR #1881, opened 2018-01-22
SONARJAVA-2284 Raise an issue when using null on method annotated with @MethodParametersAreNonNullByDefault

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/NullableAnnotationUtils.java
```diff
@@ -71,22 +83,32 @@ private static boolean isAnnotatedWithOneOf(Symbol symbol, Set<String> annotatio
   }
 
   public static boolean isGloballyAnnotatedParameterNonNull(Symbol.MethodSymbol method) {
-    return isGloballyAnnotatedWith(method, "javax.annotation.ParametersAreNonnullByDefault")
-      || usesEclipseNonNullByDefault(method, "PARAMETER");
+    return parameterNonNullGlobalAnnotation(method).isPresent();
+  }
+
+  public static Optional<String> parameterNonNullGlobalAnnotation(Symbol.MethodSymbol method) {
```
benzonico: as far as I understand the usage of this method it seems you don't get anything out of Optional: this method should just return null

PR #1882, opened 2018-01-22
Fix CatchOfThrowableOrErrorCheck as there are good use cases for catching Throwable

eperez:
The rule CatchOfThrowableOrErrorCheck should be fixed as there are good use cases for catching Throwable.
Please see some examples:
https://github.com/spring-projects/spring-framework/blob/master/spring-jdbc/src/main/java/org/springframework/jdbc/support/JdbcUtils.java#L74
https://github.com/apache/tomcat/blob/trunk/java/org/apache/catalina/core/StandardWrapperValve.java#L243

This rule should be fixed or people will break correct code.

One way to fix this rule is to only warn when catching Error as I do not know a valid use case for catching Error.

This pull request is not complete, I just submitted it as I did not found a way to create an issue.

### DIFF
--- a/java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S1181_java.json
+++ b/java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S1181_java.json
@@ -1,5 +1,5 @@
 {
-  "title": "Throwable and Error should not be caught",
+  "title": "Error should not be caught",
   "type": "CODE_SMELL",
   "status": "ready",
   "remediation": {
--- a/java-checks/src/test/files/checks/CatchOfThrowableOrErrorCheck.java
+++ b/java-checks/src/test/files/checks/CatchOfThrowableOrErrorCheck.java
@@ -5,7 +5,7 @@ class A extends Exception {
     Closer closer = Closer.create();
     try {
     } catch (RuntimeException e) {    // Compliant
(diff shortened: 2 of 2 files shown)


benzonico:
Hi, 

As mentioned on the readme : https://github.com/SonarSource/sonar-java#topic-in-sonarqube-google-group and on the contribution guideline : https://github.com/SonarSource/sonar-java/blob/master/CONTRIBUTING.md#contributing-guidelines

Please open a thread on our google group to discuss how and why we should improve a rule. 

This PR is simply not usable nor it is of any value if not discussed with the community.
As a consequence I am closing it.

eperez:
If you are interested in disabling this rule and other Sonar bogus rules, take a look at the Spring Framework default Sonar rules, they have more development experience and they already disabled this rule and other bogus rules that will reduce your code quality if enabled.
https://github.com/joshlong/spring-framework-1/wiki/Sonar-integration

benzonico:
@eperez First of all, please be kind. Your are entitled to your opinion but snarky comments and passive aggressive tone is simply not acceptable.

As mentioned in the previous comment closing this PR : if you want to reach us to discuss things, there is a space for it : https://discuss.sonarsource.com/ where you are welcome with a constructive remark or critic of rules with clear separation between opinion and facts.

I am locking this conversation as there is no way we can carry on in a constructive manner here.

see https://github.com/SonarSource/sonar-java/pull/1879

PR #1883, opened 2018-01-22
Removing TryWithResourcesCheck as try-with-resources is not equivalent to try-finally

eperez:
try-with-resources is not equivalent to try-finally as try-with-resources will always throw any exception thrown from close but that is not usually wanted.
There are important projects that are not migrating to try-with-resources because of this issue:
https://github.com/spring-projects/spring-framework/pull/1098

Migrating try-finally to try-with-resources may introduce bugs


benzonico:
Hi, as mentioned by @m-g-sonar on https://github.com/SonarSource/sonar-java/pull/1879 removing the rule (and especially as a PR) is not the appropriate way to mark issues as won't fix or false positive nor it is the appropriate way to make feedback on the analyzer and discuss how to improve a rule. 

To report any issue with the analyzer please use community feedback following instructions at : https://www.sonarqube.org/community/

So please reach out via the appropriate channel and please state the **problem** before suggesting a solution (and especially an aggressive one that is dropping the rule altogether).

As a conseque I am closing this pull request. Thanks for your understanding.

PR #1886, opened 2018-01-24
SONARJAVA-2624 Add hidden property to allow fail on SonarSource's tes…

benzonico:
…t environment

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/SonarComponents.java
+++ b/java-frontend/src/main/java/org/sonar/java/SonarComponents.java
@@ -302,4 +302,8 @@ public class SonarComponents {
       context.<String>newMeasure().forMetric(metricBuilder.create()).on(context.module()).withValue(metricValue).save();
     }
   }
+
+  public boolean shouldFailAnalysisOnException() {
+    return context.config().getBoolean("sonar.java.failOnException").orElse(false);
+  }
 }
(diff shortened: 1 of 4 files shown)


java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -302,4 +302,8 @@ public void saveAnalysisErrors() {
       context.<String>newMeasure().forMetric(metricBuilder.create()).on(context.module()).withValue(metricValue).save();
     }
   }
+
+  public boolean shouldFailAnalysisOnCheckRaisingException() {
```
m-g-sonar: I would rename it to be more widely `shouldFailAnalysisOnException`

java-frontend/src/test/java/org/sonar/java/model/VisitorsBridgeTest.java
```diff
@@ -117,4 +121,31 @@ private static CompilationUnitTree parse(String code) {
     }
   }
 
+  @Test
+  public void rethrow_exception_when_hidden_property_set_to_true() {
+    NullPointerException npe = new NullPointerException("BimBadaboum");
+    JavaFileScanner visitor = c -> {throw npe;};
```
m-g-sonar: :+1:

PR #1895, opened 2018-02-02
SONARJAVA-2597 Do not interrupt execution when S3655 is raising an issue

m-g-sonar:


its/ruling/src/test/resources/sonar-server/squid-S3655.json
```diff
@@ -12,12 +12,17 @@
 ],
 'org.sonarsource.sonarqube:sonar-server:src/main/java/org/sonar/server/es/EsClientProvider.java':[
 52,
+64,
```
m-g-sonar: All the issues are legit to me. Previously only the first access on an optional was reported, now all the ones which are following in the code. I do NOT consider them as noise.

benzonico: That seems arguable but ok.

PR #1897, opened 2018-02-04
Fixing the documentation as it implies that there are no uses cases for doing so

eperez:


benzonico:
Hi, as stated in the [readme.md]( https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/README.md ) of the directory containing the files modified in your pr,
those files are generated : as such let's [continue the conversation on the mailing list ](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/sonarqube/t_YPf5Qfenw/XyIMcYxZAgAJ)and we will modify the rule referential if required.

Thanks for the suggestion.

PR #1900, opened 2018-02-06
SONARJAVA-2640 Fix fp on RSPEC-3398 with unknown methods

benzonico:


java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
```diff
@@ -79,15 +100,29 @@ private boolean isPrivateMethodOfOuterClass(Symbol symbol) {
       return symbol.isPrivate() && symbol.owner().equals(classSymbol.owner()) && !"<init>".equals(symbol.name());
     }
 
-    public void checkUsages() {
-      for (Symbol methodUsed : usages.elementSet()) {
-        if (methodUsed.usages().size() == usages.count(methodUsed)) {
-          reportIssueOnMethod((MethodTree) methodUsed.declaration());
+    void checkUsages() {
+      for (Map.Entry<Symbol.TypeSymbol, Multiset<Symbol>> usageByInnerClassEntry : usagesByInnerClass.entrySet()) {
+        final Multiset<Symbol> innerClassUsages = usageByInnerClassEntry.getValue();
```
m-g-sonar: do you really need the final modifier on all the variables? we usually don't use them

PR #1901, opened 2018-02-07
SONARJAVA-2641 S2221 should treat constructor like methods for the check

benzonico:


its/ruling/src/test/resources/jdk6/squid-S2221.json
```diff
@@ -194,9 +194,6 @@
 'jdk6:java/text/AttributedString.java':[
 513,
 ],
-'jdk6:java/text/BreakIterator.java':[
-595,
```
m-g-sonar: I don't see why this one disappeared. Did you investigated? None of the constructors called in the body of the `try` are explicitly throwing a `java.lang.Exception`. Unresolved method?

benzonico: Did not go to full extent but indeed : unresolved constructor.

benzonico: To be more precise : constructor on line 584 is unresolved.

PR #1902, opened 2018-02-07
SONARJAVA-2585 Fix issue on completing annotation on java 9

benzonico:


java-frontend/src/main/java/org/sonar/java/resolve/BytecodeAnnotationVisitor.java
```diff
@@ -52,7 +52,16 @@ public AnnotationVisitor visitAnnotation(String name, String desc) {
 
   @Override
   public void visitEnum(String name, String desc, String value) {
-    getSymbol(desc).members().lookup(value).stream()
+    JavaSymbol.TypeJavaSymbol sym = getSymbol(desc);
+    if(sym.completing) {
+      sym.callbackOnceComplete(() -> addSymbolAsValue(name, value, sym));
+      return;
```
m-g-sonar: why not simply an `else` statement here?

java-frontend/src/main/java/org/sonar/java/resolve/JavaSymbol.java
```diff
@@ -272,6 +274,10 @@ public Tree declaration() {
     return null;
   }
 
+  public void callbackOnceComplete(Runnable callback) {
```
m-g-sonar: Not convinced by usage of `Runnable` here. We just want to execute an action. While to me the `Runnable` interface refers closely to thread handling, which are not relevant in this context.

benzonico: Runnable is the recommended functional interface for lambdas with no parameter nor return : see functional interface section in the documentation of runnable https://docs.oracle.com/javase/8/docs/api/java/lang/Runnable.html

java-frontend/src/main/java/org/sonar/java/resolve/BytecodeAnnotationVisitor.java
benzonico: reminder after refactoring and as style was going I did not feel the need to use a else... not really sure what's best here.

java-frontend/src/main/java/org/sonar/java/resolve/JavaSymbol.java
benzonico: https://stackoverflow.com/a/40153253/1263543

PR #1904, opened 2018-02-09
Add "includeSwitchBlocks" rule property to rule "S1142"

armsnyder:
https://groups.google.com/forum/#!topic/sonarqube/_EOsh0bVA5Q

The change is backwards-compatible since the default setting for the new property enforces the current rule behavior.

The rationale for the added property is that some teams may want to add an exception for switch blocks if they find that

```
case
  return
case
  return
case
  return
```

is more readable than

```
init var
case
  set var
  break
case
  set var
  break
case
  set var
  break
return
```

screen shot 2018-02-09 at 2 26 54 pm

screen shot 2018-02-09 at 2 36 48 pm


radixj:
Love the flexibility this provides, thanks for doing this!

agigleux:
I'm rejecting this PR. Please see https://groups.google.com/d/topic/sonarqube/_EOsh0bVA5Q/discussion to get the reason.

armsnyder:
Thank you for the response @agigleux . No problem. I can do it as a custom rule.

PR #1912, opened 2018-02-13
SONARJAVA-2636 Fix corner case of resolution of conditional operator …

benzonico:
…type

### DIFF
--- a/java-frontend/src/test/files/sym/ConditionalOperator.java
+++ b/java-frontend/src/test/files/sym/ConditionalOperator.java
@@ -1,0 +1,10 @@
+import java.util.List;
+public class App {
+  public class Foo<F extends List<Object>> {
+  }
+  public class Bar<G extends List<Object>> extends Foo<G> {
+  }
+  private <H extends List<Object>> Foo fun() {
+    return true ? new Bar() : new Foo<List<Object>>(); // triggers computation of least upper bound
+  }
+}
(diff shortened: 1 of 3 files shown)


java-frontend/src/test/files/sym/ConditionalOperator.java
```diff
@@ -0,0 +1,10 @@
+import java.util.List;
+public class App {
+  public class Foo<F extends List<Object>> {
+  }
+  public class Bar<G extends List<Object>> extends Foo<G> {
+  }
+  private <H extends List<Object>> Foo fun() {
+    return true ? new Bar() : new Foo<List<Object>>(); // resolving type of this expression leads to a failing invariant in LeastUpperBound computation
```
m-g-sonar: comment should be updated to reflect how it will behave, or at least not tell that it fails.

java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
```diff
@@ -1602,4 +1601,13 @@ public void fieldType_in_parameterized_hierarchy() throws Exception {
...
     softly.assertAll();
   }
+
+  @Test
+  public void conditional_operator_expression_type() {
+    Result res = Result.createFor("ConditionalOperator");
+    ExpressionTree conditional = ((ReturnStatementTree) ((MethodTree) res.symbol("fun").declaration()).block().body().get(0)).expression();
+    Type conditionalExpressionType = conditional.symbolType();
+    assertThat(conditionalExpressionType.is("App$Foo")).isTrue();
+    assertThat(((JavaType) conditionalExpressionType).isParameterized()).isTrue();
```
m-g-sonar: I would maybe add a test on the parameter type to be complete about the expected type.

PR #1918, opened 2018-02-15
SONARJAVA-2647 Clean dependency over sslr-squid-bridge

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/AnalysisException.java
```diff
@@ -0,0 +1,32 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java;
+
+public class AnalysisException extends RuntimeException {
```
benzonico: Worth documenting IMO.

java-frontend/src/main/java/org/sonar/java/ExceptionHandler.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java;
+
+import com.sonar.sslr.api.RecognitionException;
+
+public interface ExceptionHandler {
```
benzonico: Lack documentation.

java-frontend/src/main/java/org/sonar/plugins/java/api/JavaVisitor.java
```diff
@@ -0,0 +1,27 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.java.api;
+
+/**
+ * Root of all visitor classes from SonarJava analyzers
```
benzonico: I would get the javadoc consistent with the one on JavaCheck : https://github.com/SonarSource/sonar-java/pull/1918/files#diff-7c14b6cca5558b9a05292852b3d9b6a2L25 

I am also wondering about the relevance of JavaCheck interface with the introduction of JavaVisitor.  (and maybe those two classes should be renamed to SonarJavaCheck (given it is also used for pom and xml) and SonarJavaVisitor ?

m-g-sonar: I don't want to rename the class, as it is part of the API. However, I'm going to drop JavaVisitor. Xml and Maven checks are already extending JavaCheck. Issue filters as well. No need to add a new layer.

java-frontend/src/main/java/org/sonar/java/ExceptionHandler.java
m-g-sonar: It is currently only used by the ParseError check and some tests. I wonder if it would not be simpler to simply drop it and simplify the handling of exception in context.

PR #1921, opened 2018-03-02
SONARJAVA-2618 RSPEC-2761 Doubled prefix operators

stylianos-agapiou-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/DoublePrefixOperatorCheck.java
```diff
@@ -0,0 +1,72 @@
...
+      reportIssue(temp.operatorToken(), ((UnaryExpressionTree) expr).operatorToken(), "Remove multiple operator prefixes.");
+    }
+  }
+
+  private static boolean usedForDoublePrefix(UnaryExpressionTree tree) {
+    if (prefixSet.contains(tree)) {
+      return true;
+    }
+    Tree parent = tree;
+    while (parent != null) {
```
m-g-sonar: I wonder if we are going to be able to reach that condition now. I have the feeling that it's useless. Probably we can replace the stop condition by the one you added below, and continue looping while the parent **IS** a parenthesis, or any of the targeted node.

java-checks/src/test/files/checks/DoublePrefixOperatorCheck/DoublePrefixOperatorCheck.java
```diff
@@ -0,0 +1,31 @@
...
+  
+  int a5 =  + + +a2;  // Noncompliant [[sc=13;ec=16]] {{Remove multiple operator prefixes.}}
+  
+  int a6 =  + +a2;  // Noncompliant [[sc=13;ec=16]] {{Remove multiple operator prefixes.}}
+  
+  int a7 = --a2;  //Compliant
+  int a8 = ++a2;  //Compliant
+  boolean flag1 = !flag4;  // Compliant
+  int b = ~a1;  //Compliant
+ }
```
m-g-sonar: missing empty line.

java-checks/src/test/files/checks/DoublePrefixOperatorCheck/DoublePrefixOperatorCheck.java
```diff
@@ -0,0 +1,31 @@
...
+  
+  boolean flag6 = !(!(!flag4)); // Noncompliant [[sc=19;ec=22]] {{Remove multiple operator prefixes.}}
+  
+  int a3 =  - - -a2;  // Noncompliant [[sc=13;ec=16]] {{Remove multiple operator prefixes.}} 
+  
+  int a4 =  - -a2;  // Noncompliant [[sc=13;ec=16]] {{Remove multiple operator prefixes.}}
+  
+  int a5 =  + + +a2;  // Noncompliant [[sc=13;ec=16]] {{Remove multiple operator prefixes.}}
+  
+  int a6 =  + +a2;  // Noncompliant [[sc=13;ec=16]] {{Remove multiple operator prefixes.}}
```
m-g-sonar: You can probably remove the message part `{{...}}` almost everywhere, except a few places, to get it checked.

java-checks/src/test/files/checks/DoublePrefixOperatorCheck/DoublePrefixOperatorCheck.java
```diff
@@ -0,0 +1,31 @@
...
+  int a = 1;
+  boolean flag = true;
+ 
+  boolean flag2 = !!flag;  // Noncompliant  {{Remove multiple operator prefixes.}}
+  int a1 = ~~~a; // Noncompliant {{Remove multiple operator prefixes.}}
+
+  boolean flag3 = !!!flag; // Noncompliant {{Remove multiple operator prefixes.}}
+  int a2 = ~~a; // Noncompliant {{Remove multiple operator prefixes.}}
+
+  boolean flag4 = !!!flag2; // Noncompliant [[sc=19;ec=21]] {{Remove multiple operator prefixes.}}
```
m-g-sonar: this one is the same as line 8, with a different identifier. I would remove it, or replace it by something else, such as a method invocation for instance:
`boolean flag4 = !!!foo();`

java-checks/src/main/java/org/sonar/java/checks/DoublePrefixOperatorCheck.java
```diff
@@ -38,35 +41,52 @@
...
-    UnaryExpressionTree temp = (UnaryExpressionTree) tree;
-    if (usedForDoublePrefix(temp)) {
+    UnaryExpressionTree exprTree = (UnaryExpressionTree) tree;
+    if (alreadyReported(exprTree)) {
       return;
     }
-    ExpressionTree expr = ExpressionUtils.skipParentheses(temp.expression());
-    if (temp.is(expr.kind())) {
+    ExpressionTree expr = ExpressionUtils.skipParentheses(exprTree.expression());
+    if (caseHandlingDoubleTilde(expr)) {
```
m-g-sonar: This should be called from within the `exprTree.is(expr.kind())` condition (so only starting line 63), or you can not guarantee that both parent and child node are bitwise operator. You **need** to be sure that both parent and child are the same kind. On line 59, you don't know yet.

java-checks/src/main/java/org/sonar/java/checks/DoublePrefixOperatorCheck.java
```diff
@@ -38,35 +41,52 @@
...
         return true;
       }
     }
     return false;
   }
+
+  private static boolean caseHandlingDoubleTilde(ExpressionTree expr) {
+    if (expr.is(Tree.Kind.BITWISE_COMPLEMENT)) {
+      ExpressionTree exprIdentifier = ExpressionUtils.skipParentheses(((UnaryExpressionTree) expr).expression());
+      if (exprIdentifier.is(Tree.Kind.IDENTIFIER) || exprIdentifier.getClass().equals(BinaryExpressionTreeImpl.class)) {
```
m-g-sonar: Mmmh, if I agree with the first part, I don't with the second. 
`exprIdentifier.getClass().equals(BinaryExpressionTreeImpl.class)` is way too generic and does not cover the correct cases to me.

We want to spot the repetition of more than 2 bitwise complements in a row, and say nothing in case of only 2. So let's simplify this by rewriting the method:
(assuming you call from within the `if`)

```
private static boolean caseHandlingDoubleTilde(ExpressionTree expr) {
  if (expr.is(Tree.Kind.BITWISE_COMPLEMENT)) {
    if (ExpressionUtils.skipParentheses(((UnaryExpressionTree) expr).expression().is(Tree.Kind.BITWISE_COMPLEMENT) {
      // third '~' in a row, we can raise an issue
      return true;
    }
  }
  // only two, may be used for hash control
  return false;
}
```

java-checks/src/test/files/checks/DoublePrefixOperatorCheck/DoublePrefixOperatorCheck.java
```diff
@@ -1,31 +1,43 @@
...
+  boolean flag3 = !!!flag; // Noncompliant
 
-  boolean flag4 = !!!flag2; // Noncompliant [[sc=19;ec=21]] {{Remove multiple operator prefixes.}}
+  boolean flag4 = !!!foo(); // Noncompliant [[sc=19;ec=21]]
 
   boolean flag5 = !(!flag4); // Noncompliant [[sc=19;ec=22]] {{Remove multiple operator prefixes.}}
 
   int c = ~(~(~a3));    // Noncompliant [[sc=11;ec=14]] {{Remove multiple operator prefixes.}}
   
+  int d = ~~~hash;   // Noncompliant [[sc=11;ec=13]]{{Remove multiple operator prefixes.}}
```
m-g-sonar: this is a similar case to line 8.  can be dropped

java-checks/src/main/java/org/sonar/java/checks/DoublePrefixOperatorCheck.java
```diff
@@ -56,10 +55,10 @@ public void visitNode(Tree tree) {
       return;
     }
     ExpressionTree expr = ExpressionUtils.skipParentheses(exprTree.expression());
-    if (caseHandlingDoubleTilde(expr)) {
-      return;
-    }
     if (exprTree.is(expr.kind())) {
+      if (caseHandlingDoubleTilde(expr) && expr.is(Tree.Kind.BITWISE_COMPLEMENT)) {
```
m-g-sonar: Ok now I see something that we can do even better:

```
if (exprTree.is(expr.kind())) {
  UnaryExpressionTree child = (UnaryExpressionTree) expr;
  if (child.is(Tree.Kind.BITWISE_COMPLEMENT) && caseHandlingDoubleTilde(child)) {
    return;
  }
  prefixSet.add(child);
  reportIssue(exprTree.operatorToken(), child.operatorToken(), "Remove multiple operator prefixes.");
}
```

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S2761_java.html
```diff
@@ -0,0 +1,27 @@
+<p>The needless repetition of an operator is usually a typo. There is no reason to write <code>!!!i</code> when <code>!i</code> will do.</p>
+<p>On the other hand, the repetition of increment and decrement operators may have been done on purpose, but doing so obfuscates the meaning, and
+should be simplified.</p>
+<p>This rule raises an issue for sequences of: <code>!</code>, <code>~</code>, <code>-</code>, and <code>+</code>.</p>
+<h2>Noncompliant Code Example</h2>
+<pre>
+int i = 1;
+
+int j = - - -i;  // Noncompliant; just use -i
+int k = ~~i;     // Noncompliant; same as i
```
m-g-sonar: You need to update the description file again (accordingly to what we modified).

m-g-sonar:
Please also *rebase* the branch on top of master (`git rebase master`).

PR #1925, opened 2018-03-06
SONARJAVA-2634 Fix reporting to reduce noise and clarify issues

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/ArrayForVarArgCheck.java
```diff
@@ -66,7 +65,7 @@ public void visitNode(Tree tree) {
       MethodInvocationTree mit = (MethodInvocationTree) tree;
       sym = mit.symbol();
       args = mit.arguments();
-      methodName = mit.methodSelect();
+      methodName = ExpressionUtils.methodName(mit);
```
benzonico: This is a change of behavior : this should imply another refactoring or a test modification.

m-g-sonar: Agreed. I was hesitating for this one. I'm going to discard it.

PR #1929, opened 2018-03-07
SONARJAVA-2635 RSPEC-4454 "equals" method parameters should not be marked "@Nonnull"

stylianos-agapiou-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/EqualsParametersMarkedNonNullCheck.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  public List<Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.METHOD);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodTree methodTree = (MethodTree) tree;
+    if (hasSemantic() && "equals".equals(methodTree.symbol().name()) && methodTree.parameters().size() == 1 &&
```
m-g-sonar: You need to check the type of the parameter as well, otherwise you are going to match with methods called "equals()" which are NOT the override from `Object.equals(Object o)`.

java-checks/src/test/files/checks/EqualsParametersMarkedNonNullCheck.java
```diff
@@ -0,0 +1,21 @@
+class test{
```
m-g-sonar: you can remove the parent class `test`, it is not required. Just put all the classes at the same level.

java-checks/src/main/java/org/sonar/java/checks/EqualsParametersMarkedNonNullCheck.java
```diff
@@ -0,0 +1,48 @@
...
+      NullableAnnotationUtils.isAnnotatedNonNull(methodTree.parameters().get(0).symbol())) {
+      reportIssue(methodTree.parameters().get(0), "\"equals\" method parameters should not be marked \"@Nonnull\".");
```
m-g-sonar: Use a dedicated variable instead of calling twice `methodTree.parameters().get(0)`

PR #1931, opened 2018-03-08
SONARJAVA-2649 RSPEC-1821 "switch" statements should not be nested

stylianos-agapiou-sonarsource:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S1821_java.html
```diff
@@ -0,0 +1,5 @@
+<p>Nested <code>switch</code> structures are difficult to understand because you can easily confuse the cases of an inner <code>switch</code> as
+belonging to an outer statement. Therefore nested <code>switch</code> statements should be avoided.</p>
+<p>Specifically, you should structure your code to avoid the need for nested <code>switch</code> statements, but if you cannot, then consider moving
+the inner <code>switch</code> to another function.</p>
+
```
m-g-sonar: This is not enough. We need a `Compliant` and `Noncompliant` example for java. To do so, create a subtask of [RSPEC-1821](https://jira.sonarsource.com/browse/RSPEC-1821) called "Java" and put some simple examples of code. Then regenerate the documentation.

java-checks/src/main/java/org/sonar/java/checks/SwitchStatementNestedCheck.java
```diff
@@ -0,0 +1,48 @@
...
+import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.tree.CaseGroupTree;
+import org.sonar.plugins.java.api.tree.StatementTree;
+import org.sonar.plugins.java.api.tree.SwitchStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.Tree.Kind;
+
+@Rule(key = "S1821")
+public class SwitchStatementNestedCheck extends IssuableSubscriptionVisitor {
```
m-g-sonar: I would prefer `NestedSwitchStatementCheck`, which sounds definitely more correct in English.

java-checks/src/main/java/org/sonar/java/checks/SwitchStatementNestedCheck.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  public List<Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.SWITCH_STATEMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    SwitchStatementTree switchStatementTree = (SwitchStatementTree) tree;
+    List<CaseGroupTree> groups = switchStatementTree.cases();
+    List<StatementTree> nestedSwitchStatements = groups.stream().flatMap(x -> x.body().stream().filter(t -> t.is(Tree.Kind.SWITCH_STATEMENT))).collect(Collectors.toList());
```
m-g-sonar: This will only catch the `switch` statement which are at NOT nested in other structures, and only detect the ones which are in direct child of the `body()`. For instance, the following code won't be detected while **it should**:

```
switch(myVar) {
  case A: {
    switch(myOtherVar) { // switch nested in the block
      case STUFF:
        return 42;
      default:
        break;
    }
  }
  default:
    break;
}
```

In order to catch these, you will need to use a child `BaseTreeVisitor` to look for `switch` within the body.
Feel free to ask me how to do it.

java-checks/src/test/files/checks/SwitchStatementNestedCheck.java
```diff
@@ -0,0 +1,59 @@
...
+  void foo2() {
+    int i = 0;
+    switch(i) {  // Compliant
+      case 1:
+        break;
+    }
+  }
+  
+}
+
```
m-g-sonar: remove the space on last line (and clean useless spaces on other empty lines).

java-checks/src/main/java/org/sonar/java/checks/NestedSwitchStatementCheck.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private static final String REPORT = "Refactor the code to eliminate this nested \"switch\".";
+
+  @Override
+  public void visitNode(Tree tree) {
+    SwitchStatementTree switchStatementTree = (SwitchStatementTree) tree;
+    List<CaseGroupTree> cases = switchStatementTree.cases();
+    List<StatementTree> nestedSwitchStatements = cases.stream().flatMap(c -> c.body().stream().filter(stmt -> {
+      if (stmt.is(Tree.Kind.BLOCK)) {
+        stmt.accept(new NestedSwitchBlockVisitor());
```
m-g-sonar: Nop, that's not the way to use a second visitor. By this you only covers block which was only an exemple in my previous comment. You are still missing all the other switch which may be nested in other control structures (if, for, while, etc. ).

java-checks/src/test/files/checks/NestedSwitchStatementCheck.java
```diff
@@ -0,0 +1,63 @@
...
+class B{
+  void foo2() {
+    int i = 0;
+    switch(i) {  // Compliant
+      case 1:
+        break;
+    }
+  }
+}
+
```
m-g-sonar: remove the last space.

java-checks/src/main/java/org/sonar/java/checks/NestedSwitchStatementCheck.java
```diff
@@ -0,0 +1,67 @@
...
+import org.sonar.plugins.java.api.tree.LambdaExpressionTree;
+import org.sonar.plugins.java.api.tree.SwitchStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.Tree.Kind;
+
+@Rule(key = "S1821")
+public class NestedSwitchStatementCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Kind> nodesToVisit() {
```
m-g-sonar: Use `Tree.Kind` instead, to remove the import.

java-checks/src/main/java/org/sonar/java/checks/NestedSwitchStatementCheck.java
```diff
@@ -0,0 +1,67 @@
...
+    return ImmutableList.of(Tree.Kind.SWITCH_STATEMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    SwitchStatementTree switchStatementTree = (SwitchStatementTree) tree;
+    List<CaseGroupTree> cases = switchStatementTree.cases();
+    cases.stream().forEach(c -> c.accept(new NestedSwitchVisitor()));
```
m-g-sonar: you can inline everything here: `((SwitchStatementTree) tree).cases().stream()...`

java-checks/src/main/java/org/sonar/java/checks/NestedSwitchStatementCheck.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Override
+    public void visitLambdaExpression(LambdaExpressionTree lambdaExpressionTree) {
+      // skip Lambdas
+    }
+
+    @Override
+    public void visitSwitchStatement(SwitchStatementTree tree) {
+      reportIssue((tree).switchKeyword(),
```
m-g-sonar: remove the useless parentheses, and no need for 2 different lines.

java-checks/src/test/files/checks/NestedSwitchStatementCheck.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+}
+
+class B{
+  List<Integer> list = new ArrayList<>();
```
m-g-sonar: no need for the constructor to `ArrayList` here, you can keep the field list alone.  Note that at the moment `ArrayList` won't be resolved by the engine, because there is no `import` of `java.util.ArrayList`.

java-checks/src/main/java/org/sonar/java/checks/NestedSwitchStatementCheck.java
```diff
@@ -24,33 +24,29 @@
...
-    cases.stream().forEach(c -> c.accept(new NestedSwitchVisitor()));
+    ((SwitchStatementTree) tree).cases().stream().forEach(c -> c.accept(new NestedSwitchVisitor()));
   }
 
   private class NestedSwitchVisitor extends BaseTreeVisitor {
 
     @Override
-    public void visitClass(ClassTree tree) {
-      // skip Class
+    public void visitNewClass(NewClassTree tree) {
```
m-g-sonar: Sorry, I asked you to change the comment and **not the tree** you visit. Unfortunately, my comment was not complete... It should have been `skip nested classes and anonymous classes`.

You should revert that and continue skipping classes (ClassTree), otherwise you are going to have False Positives (FPs). The following (ugly) code is **now** raising FPs, while the previous one was not:
```java
  void gul(int i) {
    switch (i) {
      case 0:
      case 42:
        class B { // nested class
          void bar(int j) {
            switch (j) { // FP - this switch is not nested
              case 42:
                System.out.println("yolo");
                break;
              case 0:
                break;
              default:
                break;
            }
          }
        }
        new B().bar(i);
        break;

      default:
        break;
    }
  }
```

java-checks/src/main/java/org/sonar/java/checks/NestedSwitchStatementCheck.java
```diff
@@ -24,33 +24,29 @@
...
+  public List<Tree.Kind> nodesToVisit() {
     return ImmutableList.of(Tree.Kind.SWITCH_STATEMENT);
   }
 
   @Override
   public void visitNode(Tree tree) {
-    SwitchStatementTree switchStatementTree = (SwitchStatementTree) tree;
-    List<CaseGroupTree> cases = switchStatementTree.cases();
-    cases.stream().forEach(c -> c.accept(new NestedSwitchVisitor()));
+    ((SwitchStatementTree) tree).cases().stream().forEach(c -> c.accept(new NestedSwitchVisitor()));
```
m-g-sonar: No need to recreate all the time a new instance of the visitor. The same visitor can be reused for all the case group:

```java
    NestedSwitchVisitor visitor = new NestedSwitchVisitor();
    ((SwitchStatementTree) tree).cases().stream().forEach(c -> c.accept(visitor));
```

java-checks/src/test/files/checks/NestedSwitchStatementCheck.java
```diff
@@ -1,5 +1,3 @@
-import java.util.List;
```
m-g-sonar: Do not remove that. You still need `java.util.List`, you are declaring it line 56! That's `ArrayList` which was not imported, and also useless.

PR #1934, opened 2018-03-13
SONARJAVA-2486 RSPEC-3366 "this" should not be exposed from constructors

stylianos-agapiou-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree tree) {
+      if (this.owner == tree.symbol().owner()) {
+        return;
+      }
+      Arguments args = tree.arguments();
+      if (args.isEmpty()) {
+        return;
+      }
+      args.stream().filter(arg -> is(arg, "this")).forEach(expr -> reportIssue(expr, REPORT));
```
m-g-sonar: See my comment about useless "this". You will probably then be able to use a method reference instead.

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,107 @@
...
+
+  private class ConstructorBodyVisitor extends BaseTreeVisitor {
+    private Symbol owner;
+
+    public ConstructorBodyVisitor(Symbol owner) {
+      this.owner = owner;
+    }
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree tree) {
```
m-g-sonar: you need to call `super.VisitMethodInvocation(tree)` at the end of this method, otherwise, you are going to cut the exploration of child, and therefore miss any nested method invocation in the parameters:

`foo(myObject.bar(this))`

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,107 @@
...
+      if (variable.is(Tree.Kind.IDENTIFIER) && isStaticField((IdentifierTree) variable)) {
+        return;
+      }
+      reportIssue(tree, REPORT);
+    }
+
+    private boolean is(ExpressionTree expression, String match) {
+      if (expression.is(Tree.Kind.IDENTIFIER)) {
+        String targetName = ((IdentifierTree) expression).name();
+        return match.equals(targetName);
```
m-g-sonar: you can inline `targetName`.
I'm also not sure it's enough. I would check that the symbol type of `this` is the same as `owner` as well. See example below (I'm not sure the following example compile, maybe just skipping annonymous/inner class would do the trick)

```java
class A {
  B b;
  public A() {
    this.b = new B() {
      @Override
      void bar() {
        B.foo(this); // Here is the "this" from B, not from A;
      }
    }
  }
}
```

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,107 @@
...
+      }
+      ExpressionTree variable = tree.variable();
+      MemberSelectExpressionTree memberSelect;
+      if (variable.is(Tree.Kind.MEMBER_SELECT)) {
+        memberSelect = (MemberSelectExpressionTree) variable;
+        if (memberSelect.expression().symbolType().symbol().equals(this.owner) && memberSelect.identifier().symbol().isStatic()) {
+          return;
+        }
+      }
+      if (variable.is(Tree.Kind.IDENTIFIER) && isStaticField((IdentifierTree) variable)) {
```
m-g-sonar: I would use `else if` here

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private boolean is(ExpressionTree expression, String match) {
+      if (expression.is(Tree.Kind.IDENTIFIER)) {
+        String targetName = ((IdentifierTree) expression).name();
+        return match.equals(targetName);
+      }
+      return false;
+    }
+
+    private boolean isStaticField(IdentifierTree identifier) {
```
m-g-sonar: not sure it worth having a dedicated method here, especially if you use it only once.

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
benzonico: @stylianos-agapiou-sonarsource This comment was not addressed. You can add a failing testcase using the example provided in the previous comment. And then fix it following the suggestion of Michael.

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Override
+    public void visitClass(ClassTree tree) {
+      // skip nested and anonymous classes
+    }
+
+    private boolean isThis(ExpressionTree expression) {
+      if (expression.is(Tree.Kind.IDENTIFIER)) {
+        String targetName = ((IdentifierTree) (ExpressionUtils.skipParentheses(expression))).name();
```
benzonico: Please add a case where the skipParentheses call is necessary. (in other words : with your current unit test, I can remove the call to ExpressionUtils.skipParentheses and the tests would not fail).

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,104 @@
...
+
+@Rule(key = "S3366")
+public class ThisExposedFromConstructorCheck extends IssuableSubscriptionVisitor {
+
+  private static final String REPORT = "Make sure the use of \"this\" doesn't" +
+    " expose partially-constructed instances of this class in multi-threaded environments.";
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.CONSTRUCTOR);
```
benzonico: java.util.Collections.singletonList() should be prefered for singleton list. (always a better practice to rely on JDK classes rather than libs).

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+      if (!isThis(tree.expression())) {
+        return;
+      }
+      ExpressionTree variable = tree.variable();
+      if (variable.is(Tree.Kind.MEMBER_SELECT)) {
+        MemberSelectExpressionTree memberSelect = (MemberSelectExpressionTree) variable;
+        if (memberSelect.expression().symbolType().symbol().equals(this.owner) && memberSelect.identifier().symbol().isStatic()) {
```
benzonico: I struggle to understand that line, probably worth a comment explaining WHY we don't raise an issue in this case.

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,108 @@
...
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S3366")
+public class ThisExposedFromConstructorCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList((Tree.Kind.CONSTRUCTOR));
```
benzonico: Unecessary parentheses.

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,108 @@
...
+      report(tree);
+    }
+
+    @Override
+    public void visitClass(ClassTree tree) {
+      // skip nested and anonymous classes
+    }
+
+    private boolean isThis(ExpressionTree expression) {
+      expression = (ExpressionUtils.skipParentheses(expression));
```
benzonico: uncessary parenthesis and it is usually not a nice practice to reassign a parameter. Please declare a local var to work with.

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+      if (!isThis(tree.expression())) {
+        return;
+      }
+      ExpressionTree variable = tree.variable();
+      if (variable.is(Tree.Kind.MEMBER_SELECT)) {
+        MemberSelectExpressionTree memberSelect = (MemberSelectExpressionTree) variable;
+        // the symbol of "this" is the same as the owner of the lhs and and at the same time the variable must be defined static
```
benzonico: This comment describes WHAT the condition of the if is doing. Not the WHY. 
The comment should answer to the question : WHY should we return if the condition is true.

java-checks/src/main/java/org/sonar/java/checks/ThisExposedFromConstructorCheck.java
```diff
@@ -92,9 +92,9 @@ public void visitClass(ClassTree tree) {
     }
 
     private boolean isThis(ExpressionTree expression) {
-      expression = (ExpressionUtils.skipParentheses(expression));
-      if (expression.is(Tree.Kind.IDENTIFIER)) {
-        String targetName = ((IdentifierTree) expression).name();
+      ExpressionTree expressionNoParenthesis = (ExpressionUtils.skipParentheses(expression));
```
benzonico: useless parentheses

PR #1939, opened 2018-03-15
SONARJAVA-2642 RSPEC-126 "if ... else if" constructs should end with "else" clauses

stylianos-agapiou-sonarsource:


java-checks/src/test/files/checks/IfElseIfStatementEndsWithElseCheck.java
```diff
@@ -0,0 +1,42 @@
...
+  private static final boolean FALSE = false;
+  private int x = 2;
+
+  void foo() {
+    if (x == 0) {       // Compliant
+    } else if (x == 1) {
+        if(x == 2) {  // Noncompliant
+
+        }
+        else if(x == 3) {
```
m-g-sonar: can you please reformat this test file to have something a bit more compact and coherent in term of style? You can also remove empty lines.

java-checks/src/main/java/org/sonar/java/checks/IfElseIfStatementEndsWithElseCheck.java
```diff
@@ -0,0 +1,64 @@
...
+import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.JavaFileScanner;
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.IfStatementTree;
+import org.sonar.plugins.java.api.tree.StatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S126")
+public class IfElseIfStatementEndsWithElseCheck extends BaseTreeVisitor implements JavaFileScanner {
```
m-g-sonar: Why a `BaseTreeVisitor` ? It seems to me that this approach does not work for all the cases. It makes things harder to understand and add complexity to the tree exploration.

From my point of view, it should be a lot easier to simply subscribe to to `IF_STATEMENT`, verify that the `else` statement exists and is also a `IF_STATEMENT`, and finally if that child `IF` is missing a `else`.

java-checks/src/main/java/org/sonar/java/checks/IfElseIfStatementEndsWithElseCheck.java
```diff
@@ -0,0 +1,64 @@
...
+        context.reportIssue(this, tree, "\"if ... else if\" constructs should end with \"else\" clauses.");
+      } else if (!statement.is(Tree.Kind.IF_STATEMENT)) {
+        return;
+      }
+    }
+  }
+
+  private void visit(IfStatementTree tree) {
+    scan(tree.condition());
+    scan(tree.thenStatement());
```
m-g-sonar: because you are not visiting the `else`, it means that you are going to fail finding these cases:
```java
    if (true) {
      // do something
    } else {
      if (true) {  //  not detected
      } else if (true) {
      }
    }
```

java-checks/src/main/java/org/sonar/java/checks/IfElseIfStatementEndsWithElseCheck.java
```diff
@@ -0,0 +1,64 @@
...
+    if (tree.elseKeyword() == null) {
+      return;
+    }
+    IfStatementTree ifStmt;
+    while (statement != null) {
+      ifStmt = (IfStatementTree) statement;
+      visit(ifStmt);
+      statement = ifStmt.elseStatement();
+      if (statement == null) {
+        context.reportIssue(this, tree, "\"if ... else if\" constructs should end with \"else\" clauses.");
```
m-g-sonar: I think reporting on the whole tree (meaning from the first token to the last of the `if`, which can be MANY lines...) is going to do too much noise. Try reporting only on the last `else if`.

java-checks/src/main/java/org/sonar/java/checks/IfElseIfStatementEndsWithElseCheck.java
```diff
@@ -0,0 +1,46 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.IF_STATEMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    StatementTree elseStmt = ((IfStatementTree) tree).elseStatement();
+    if (elseStmt != null
+      && elseStmt.is(Tree.Kind.IF_STATEMENT) && ((IfStatementTree) elseStmt).elseStatement() == null) {
+      reportIssue(elseStmt, "\"if ... else if\" constructs should end with \"else\" clauses.");
```
m-g-sonar: You should **not** report on the whole `else` statement. It is going to do way too much noise in the SonarQube UI, and put the issue on the last line, while we want it on the last `else if` line.

Please report only on `else if` (the `else` keyword is available as a token in the parent `IfStatementTree`, and the `if` keyword will be available in the child `ifStatementTree`)
 
Currently, in the following code for instance, it would highlight in red the 8 lines following the `else`!
```java
        if (c == '\'')
          inQuote = true;
        else if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z')) {
          int ci = from.indexOf(c);
          if (ci == -1)
            throw new IllegalArgumentException("Illegal pattern " +
              " character '" +
              c + "'");
          c = to.charAt(ci);
        }
      }
```

image

java-checks/src/main/java/org/sonar/java/checks/IfElseIfStatementEndsWithElseCheck.java
```diff
@@ -0,0 +1,47 @@
...
+    return Collections.singletonList(Tree.Kind.IF_STATEMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    StatementTree elseStmt = ((IfStatementTree) tree).elseStatement();
+    if (elseStmt != null
+      && elseStmt.is(Tree.Kind.IF_STATEMENT) && ((IfStatementTree) elseStmt).elseStatement() == null) {
+
+      reportIssue(((IfStatementTree) elseStmt.parent()).elseKeyword(), ((IfStatementTree) elseStmt).ifKeyword(), "\"if ... else if\" constructs should end with \"else\" clauses.");
```
m-g-sonar: No need to call `parent()` here, you already have this tree, its the `tree` parameter from `visitNode` method.

java-checks/src/main/java/org/sonar/java/checks/IfElseIfStatementEndsWithElseCheck.java
```diff
@@ -0,0 +1,47 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.IF_STATEMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    StatementTree elseStmt = ((IfStatementTree) tree).elseStatement();
+    if (elseStmt != null
+      && elseStmt.is(Tree.Kind.IF_STATEMENT) && ((IfStatementTree) elseStmt).elseStatement() == null) {
+
```
m-g-sonar: remove this empty line

java-checks/src/main/java/org/sonar/java/checks/IfElseIfStatementEndsWithElseCheck.java
```diff
@@ -0,0 +1,47 @@
...
+public class IfElseIfStatementEndsWithElseCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.IF_STATEMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    StatementTree elseStmt = ((IfStatementTree) tree).elseStatement();
```
m-g-sonar: In order to avoid to have to cast the tree 2 times, declare a `ifStmt` variable first, you will then be able to use it directly when reporting.

java-checks/src/main/java/org/sonar/java/checks/IfElseIfStatementEndsWithElseCheck.java
```diff
@@ -37,11 +37,14 @@
...
-    StatementTree elseStmt = ((IfStatementTree) tree).elseStatement();
-    if (elseStmt != null
-      && elseStmt.is(Tree.Kind.IF_STATEMENT) && ((IfStatementTree) elseStmt).elseStatement() == null) {
-
-      reportIssue(((IfStatementTree) elseStmt.parent()).elseKeyword(), ((IfStatementTree) elseStmt).ifKeyword(), "\"if ... else if\" constructs should end with \"else\" clauses.");
+    IfStatementTree treeIfStmt = (IfStatementTree) tree;
+    StatementTree elseStmt = treeIfStmt.elseStatement();
+    IfStatementTree ifStmt = null;
+    if (elseStmt != null && elseStmt.is(Tree.Kind.IF_STATEMENT)) {
+      ifStmt = (IfStatementTree) elseStmt;
```
m-g-sonar: I think you are over-complicating things here. Try to keep it simple and "logical" to read, introducing the variable you need and minimizing the number of validation you need to do.

In this case, simply nest the 2 `if` statements, and put the definition of the `ifStmt` variable within the body of the first `if`. you can then remove the `ifStmt != null` test.

PR #1940, opened 2018-03-15
SONARJAVA-2664 Upgrade to sslr-squid-bridge 2.7.0

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/JavaSquid.java
```diff
@@ -109,18 +109,18 @@ private static VisitorsBridge createVisitorBridge(
   }
 
 
-  public void scan(Iterable<File> sourceFiles, Iterable<File> testFiles) {
+  public void scan(Collection<File> sourceFiles, Collection<File> testFiles) {
```
benzonico: Why this change ?

java-frontend/src/main/java/org/sonar/java/xml/XmlAnalyzer.java
```diff
@@ -63,15 +64,15 @@ public XmlAnalyzer(SonarComponents sonarComponents, JavaCheck... visitors) {
     this.xPath = XPathFactory.newInstance().newXPath();
   }
 
-  public void scan(Iterable<File> files) {
+  public void scan(Collection<File> files) {
```
benzonico: Iterables.isEmpty if we make this move of iterable to collection

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaRulesDefinitionTest.java
```diff
@@ -68,6 +68,11 @@ public void test_creation_of_rules() {
     RulesDefinition.Rule unusedPrivateMethodRule = repository.rule("UnusedPrivateMethod");
     assertThat(unusedPrivateMethodRule.activatedByDefault()).isEqualTo(true);
 
+    // rule templates are manually defined
+    assertThat(repository.rules().stream()
```
benzonico: This lacks the opposite testing : if I add a new rule template how will I know I actually properly declared it as a rule template ?

sonar-java-plugin/src/main/java/org/sonar/plugins/java/XmlFileSensor.java
```diff
@@ -69,7 +70,7 @@ private boolean hasXmlFiles() {
     return fs.inputFiles(xmlFilePredicate);
   }
 
-  private static Iterable<File> toFile(Iterable<InputFile> inputFiles) {
+  private static Collection<File> toFile(Iterable<InputFile> inputFiles) {
     return StreamSupport.stream(inputFiles.spliterator(), false).map(InputFile::file).collect(Collectors.toList());
```
benzonico: This can be simplified with the move to Collection.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaSquidSensor.java
```diff
@@ -92,15 +93,15 @@ public void execute(SensorContext context) {
...
   }
 
-  private Iterable<File> getTestFiles() {
+  private Collection<File> getTestFiles() {
     return toFile(fs.inputFiles(fs.predicates().and(fs.predicates().hasLanguage(Java.KEY), fs.predicates().hasType(InputFile.Type.TEST))));
   }
 
-  private static Iterable<File> toFile(Iterable<InputFile> inputFiles) {
+  private static Collection<File> toFile(Iterable<InputFile> inputFiles) {
     return StreamSupport.stream(inputFiles.spliterator(), false).map(InputFile::file).collect(Collectors.toList());
```
benzonico: this can be simplified with the move to Collection.

java-frontend/src/main/java/org/sonar/java/JavaSquid.java
m-g-sonar: In order to translate more easily the collections of `File` into collections of `String`, using streams, at the bottom of the chain.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaSquidSensor.java
m-g-sonar: Not sure to see how

benzonico: Sorry indeed, was confused by the return type. input type has not changed so comment is irrelevant.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/XmlFileSensor.java
benzonico: sorry, confusion on my side, forget this comment.

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaRulesDefinitionTest.java
m-g-sonar: I agree. We were not testing template rules before, but I don't know how to put in place one without hardcoding the number of non-template rule that we currently have... Which will requires to edit the test with each new rule. I'm not sure we want that.
Otherwise I will simply go for:
```
assertThat(rules.stream().filter(rule -> !rule.template())).hasSize(445 - 6);
```

benzonico: This is a convuleted test of the number of rules.

PR #1947, opened 2018-03-21
SONARJAVA-2650 RSPEC-2234 Parameters should be passed in the correct order

stylianos-agapiou-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.semantic.Symbol;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S2234")
+public class MethodParametersOrderCheck extends IssuableSubscriptionVisitor {
+
+  private Map<Symbol, LinkedHashSet<String>> methodNameKeyAndParameteresValue = new HashMap<>();
```
m-g-sonar: Should be `Map<Symbol, Set<String>>`.

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree.is(Tree.Kind.METHOD)) {
+      MethodTree methodTree = (MethodTree) tree;
+      methodNameKeyAndParameteresValue.put(methodTree.symbol(), new LinkedHashSet<>(methodTree.parameters().stream().map(param -> param.simpleName().name()
+        .toLowerCase(Locale.ENGLISH)).collect(Collectors.toList())));
+      return;
+    } else {
+      MethodInvocationTree methodInvTree = (MethodInvocationTree) tree;
+      LinkedHashSet<String> definedFunctionParameters = methodNameKeyAndParameteresValue.get(methodInvTree.symbol());
```
m-g-sonar: This won't work. It means that you need to declare the method **before** any invocation in the compilation unit. This is not what we want, that's only a special case and not the general case.

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    methodNameKeyAndParameteresValue.clear();
+    super.scanFile(context);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree.is(Tree.Kind.METHOD)) {
```
m-g-sonar: This won't work. It implicitly implies that order of exploration is important, which should **not**. 

Only relying on symbol should be enough, the only constraint being that the method is declared in the same file.

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
m-g-sonar: Name is too complex (and wrong, you are not storing method name, but method symbol). Why not `ParameterNamesByMethod`?

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+  private Map<Symbol, LinkedHashSet<String>> methodNameKeyAndParameteresValue = new HashMap<>();
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.METHOD, Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    methodNameKeyAndParameteresValue.clear();
```
m-g-sonar: :+1:

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S2234")
+public class MethodParametersOrderCheck extends IssuableSubscriptionVisitor {
+
+  private Map<Symbol, LinkedHashSet<String>> methodNameKeyAndParameteresValue = new HashMap<>();
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.METHOD, Tree.Kind.METHOD_INVOCATION);
```
m-g-sonar: Subscribing to `METHOD` is only introducing confusion, and cause problem with the rule.
Subscribing to `METHOD_INVOCATION` should be enough. When hitting a method invocation, you can then look into the map if something exist for that symbol. If it's not the case, retrieve the parameters names and add a new entry on the map. Then you check for parameter names.

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+    methodNameKeyAndParameteresValue.clear();
+    super.scanFile(context);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree.is(Tree.Kind.METHOD)) {
+      MethodTree methodTree = (MethodTree) tree;
+      methodNameKeyAndParameteresValue.put(methodTree.symbol(), new LinkedHashSet<>(methodTree.parameters().stream().map(param -> param.simpleName().name()
+        .toLowerCase(Locale.ENGLISH)).collect(Collectors.toList())));
```
m-g-sonar: the `new LinkedHashSet()` and collectors can be simplified into `.collect(Collectors.toCollection(LinkedHashSet::new))`

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+      return;
+    } else {
+      MethodInvocationTree methodInvTree = (MethodInvocationTree) tree;
+      LinkedHashSet<String> definedFunctionParameters = methodNameKeyAndParameteresValue.get(methodInvTree.symbol());
+      LinkedHashSet<String> invocationParameters = new LinkedHashSet<>(
+        methodInvTree.arguments().stream().map(param -> param.toString().toLowerCase(Locale.ENGLISH)).collect(Collectors.toList()));
```
m-g-sonar: Nop, no way. Don't use `.toString()` on trees, never.

You should iterate on parameters, and transform the ones which are identifiers to take their name, for all the others, use a placeholder, such as `null`.

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+  public void visitNode(Tree tree) {
+    if (tree.is(Tree.Kind.METHOD)) {
+      MethodTree methodTree = (MethodTree) tree;
+      methodNameKeyAndParameteresValue.put(methodTree.symbol(), new LinkedHashSet<>(methodTree.parameters().stream().map(param -> param.simpleName().name()
+        .toLowerCase(Locale.ENGLISH)).collect(Collectors.toList())));
+      return;
+    } else {
+      MethodInvocationTree methodInvTree = (MethodInvocationTree) tree;
+      LinkedHashSet<String> definedFunctionParameters = methodNameKeyAndParameteresValue.get(methodInvTree.symbol());
+      LinkedHashSet<String> invocationParameters = new LinkedHashSet<>(
```
m-g-sonar: You can probably use `Set` directly here, and previous variable as well.

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+        methodInvTree.arguments().stream().map(param -> param.toString().toLowerCase(Locale.ENGLISH)).collect(Collectors.toList()));
+      if (definedFunctionParameters == null || invocationParameters.isEmpty()) {
+        return;
+      }
+      if (matchParameters(definedFunctionParameters, invocationParameters)) {
```
m-g-sonar: why not `if(!matchParameters(definedFunctionParameters, invocationParameters))` instead? then you can drop the useless "return;"

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+        return;
+      } else {
+        reportIssue(methodInvTree, "Parameters to " + methodInvTree.symbol().name() + " have the same names but not the same order as the method arguments.");
```
m-g-sonar: This is not enough regarding reporting.
The specification [RSPEC-2234](https://jira.sonarsource.com/browse/RSPEC-2234) is explicitly saying the following:

> Highlighting:
> * Primary: actual parameters clause
> * Secondary: formal parameters clause

You are reporting on the totality of the method invocations, while we should only highlight arguments, and we should have a secondary location with the formal parameters.

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+  }
+
+  private static boolean matchParameters(LinkedHashSet<String> functionParameters, LinkedHashSet<String> invocationParameters) {
```
m-g-sonar: You are answering `true` when you **don't** have a match of parameters... to me the name of the method is then wrong, that's not what it is saying it is doing!

java-checks/src/test/files/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,49 @@
...
+  
+  public double divide(int divisor, int dividend) {
+    return divisor/dividend;
+  }
+
+  public void doTheThing() {
+    int divisor = 15;
+    int dividend = 5;
+
+    double result = divide(dividend, divisor);  // Noncompliant [[sc=21;ec=46]] {{Parameters to divide have the same names but not the same order as the method arguments.}}
```
m-g-sonar: This won't work if I place the `divide` method under `doTheThing`.

java-checks/src/test/files/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,49 @@
...
+    int result = divide(dividend, divisOr);  // Noncompliant {{Parameters to divide have the same names but not the same order as the method arguments.}}
+  
+    result = divide(divisOr, dividend);      // Compliant
+    result = divide(parameter1, parameter2);  // Compliant
+    result = divide(fun(), dividend);         //Compliant
+  
+    // coverage
+    fun();
+    
+    java.util.Objects.equals(divisOr, dividend);
```
m-g-sonar: this is useless. You are not trying to make the rule fail, nor making it clear.

You should try with `equals(b, a)` to invert the one from the method definition, and explicitly mention it is compliant, maybe even with a comment **why**, to show the person reviewing that the case is known, meaning : We are **only** focusing on method declared in the same file.

java-checks/src/test/files/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,49 @@
...
+    int dividend = 5;
+  
+    int result = divide(dividend, divisOr);  // Noncompliant {{Parameters to divide have the same names but not the same order as the method arguments.}}
+  
+    result = divide(divisOr, dividend);      // Compliant
+    result = divide(parameter1, parameter2);  // Compliant
+    result = divide(fun(), dividend);         //Compliant
+  
+    // coverage
+    fun();
```
m-g-sonar: Ok, so you tested with zero, and two parameters... What about 1 or 3 ? What about variadic methods ? Don't focus on special cases.

```
void foo(int a, int ... bs) { /* ... */ }
```

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,83 @@
...
+    if (invocationParameters.containsAll(functionParameters)) {
+      return new ArrayList<>(functionParameters).equals(new ArrayList<>(invocationParameters));
```
m-g-sonar: I don't like that we have to create new lists all the time. I wonder if there is another alternative...

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
m-g-sonar: Yes, but we try to limit as much as possible use of implementation class in variable declaration in Java.
Using interfaces rather than implementation classes reduces coupling and allow more flexibility. In this particular case, using a `LinkedHashSet` is an implementation details (maintaining the order in your `set`), but you are not using any particular method which would be defined only in `LinkedHashSet` and not in `Set`. Refer to this article: https://www.javaworld.com/article/2073649/core-java/why-extends-is-evil.html 

It should be:
```Set<String> invocationParameters = new LinkedHashSet<>(... ``` 
instead of 
```LinkedHashSet<String> invocationParameters = new LinkedHashSet<>(... ```

Comment (m-g-sonar) [reply java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java]: And we should use Lists.

File: java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java [hunk 16]
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private static boolean argumentsTypesMatchParametersTypesAndNamesNotOrdered(List<Symbol> parameterList, List<IdentifierTree> argumentList) {
+    Iterator<IdentifierTree> argumentsIterator = argumentList.stream().filter(Objects::nonNull).iterator();
+    int countArgumentsNotOrdered = 0;
+    while (argumentsIterator.hasNext()) {
+      IdentifierTree argument = argumentsIterator.next();
+      int index = (parameterList.stream().map(param -> param.name().toLowerCase(Locale.ENGLISH)).collect(Collectors.toList()))
+        .indexOf(argument.name().toLowerCase(Locale.ENGLISH));
+      if (parameterList.get(index).type().equals(argument.symbolType())) {
```
m-g-sonar: not `.equals(...)`, `.is(argument.symbolType().fullyQualifiedName())`

m-g-sonar: we also need to deal with unknown types. If the formal type is unknown or the argument type is unknown, we should return false.

m-g-sonar: invert the logic to not require the `continue` statement and simplify the code.
```java
if (formalType.isUnknown() || argType.isUnknown() || !formalType.is(argType))) {
  return false;
}
if (argumentList.indexOf(argument) != index) {
  countArgumentsNotOrdered++;
}
```

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,144 @@
...
+  private static class ParametersList {
+
+    private List<String> parameterNames;
+    private List<Type> parameterTypes;
+
+    public ParametersList(MethodTree methodTree) {
+      parameterNames = new ArrayList<>();
+      parameterTypes = new ArrayList<>();
+      List<Symbol> symbolListOfParameters = methodTree.parameters().stream().map(VariableTree::symbol).collect(Collectors.toList());
+      parameterNames.addAll(symbolListOfParameters.stream().map(parameter -> parameter.name().toLowerCase(Locale.ENGLISH)).collect(Collectors.toList()));
```
m-g-sonar: You can iterate only once on the lit of symbols and collect everything:
```java
methodTree.parameters().stream().map(VariableTree::symbol).forEach(symbol -> {
  parameterNames.add(symbol.name().toLowerCase(Locale.ENGLISH));
  parameterTypes.add(symbol.type())
});
```

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,144 @@
...
+      parameterTypes.addAll(symbolListOfParameters.stream().map(Symbol::type).collect(Collectors.toList()));
+    }
+
+    public boolean contains(String argument) {
```
m-g-sonar: `hasArgumentWithName`

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -0,0 +1,144 @@
...
+    if (methodDeclaration == null) {
+      return;
+    }
+    ParametersList formalParameterList = parametersByMethod.computeIfAbsent(methodInvTree.symbol(), m -> new ParametersList(methodDeclaration));
+    List<IdentifierTree> argumentsList = methodInvTree.arguments().stream().map(this::argumentToIdentifier).collect(Collectors.toList());
+    if (matchingNames(formalParameterList, argumentsList)
+      && matchingTypesWrongOrder(formalParameterList, argumentsList)) {
+      List<JavaFileScannerContext.Location> flow = methodDeclaration.parameters().stream().map(param -> new JavaFileScannerContext.Location("Formal Parameters", param))
+        .collect(Collectors.toList());
+      reportIssue(methodInvTree.arguments(), "Parameters to " + methodInvTree.symbol().name() + " have the same names but not the same order as the method arguments.",
```
m-g-sonar: should not include parenthesis, as you are not including them in the "flow"

PR #1954, opened 2018-03-26
SONARJAVA-2513 RSPEC-4274 Asserts should not be used to check the par…

stylianos-agapiou-sonarsource:


its/ruling/src/test/resources/jdk6/squid-S4274.json
```diff
@@ -0,0 +1,31 @@
+{
+'jdk6:java/awt/AWTKeyStroke.java':[
+857,
```
m-g-sonar: I would prefer to have also the 2 next statements raising an issue. It would make more sense to me to raise issues as soon as there is an usage, even if it's as parameter of method invocation.

java-checks/src/main/java/org/sonar/java/checks/AssertsOnParametersOfPublicMethodCheck.java
```diff
@@ -0,0 +1,82 @@
...
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.METHOD);
+  }
+
+  private Set<AssertStatementTree> assertReported = new HashSet<>();
+
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    assertReported.clear();
```
m-g-sonar: :+1:

java-checks/src/main/java/org/sonar/java/checks/AssertsOnParametersOfPublicMethodCheck.java
```diff
@@ -0,0 +1,82 @@
...
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = "S4274")
+public class AssertsOnParametersOfPublicMethodCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.METHOD);
```
m-g-sonar: We should also register to `CONSTRUCTOR`, as they are just "special" methods. It would make more sense to me to target both kinds of methods.

java-checks/src/main/java/org/sonar/java/checks/AssertsOnParametersOfPublicMethodCheck.java
```diff
@@ -0,0 +1,82 @@
...
+      return;
+    }
+    MethodTree methodTree = (MethodTree) tree;
+    if (!methodTree.symbol().isPublic()) {
+      return;
+    }
+    for (VariableTree parameter : methodTree.parameters()) {
+      for (IdentifierTree paramUsage : parameter.symbol().usages()) {
+        Tree paramParent = paramUsage.identifierToken().parent();
+        while (!paramParent.equals(methodTree) && !assertReported.contains(paramParent)) {
```
m-g-sonar: extract this to a dedicated method.

java-checks/src/main/java/org/sonar/java/checks/AssertsOnParametersOfPublicMethodCheck.java
```diff
@@ -0,0 +1,82 @@
...
+          if (parentType(paramParent)) {
```
m-g-sonar: I don't like this method doing 2 things, checking for kind (not `type`, it has another meaning in the context of the SonarJava plugin) and reporting an issue. Consequently, naming of the method is not explicit.

You can probably simply check if it's an `assert`, and then in the body of the `if`, add the node to the set and report directly the issue

java-checks/src/main/java/org/sonar/java/checks/AssertsOnParametersOfPublicMethodCheck.java
```diff
@@ -0,0 +1,82 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
+      return;
+    }
+    MethodTree methodTree = (MethodTree) tree;
+    if (!methodTree.symbol().isPublic()) {
+      return;
+    }
+    for (VariableTree parameter : methodTree.parameters()) {
```
m-g-sonar:
```java
methodTree.parameters().stream()
  .map(VariableTree::symbol)
  .map(Symbol::usages)
  .flatMap(List::stream)
  .forEach(this::checkUsageInAssert);

// ...

void checkUsageInAssert(IdentifierTree usage) { ... }
```

java-checks/src/test/files/checks/AssertsOnParametersOfPublicMethodCheck.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public int getPrice(int a) {
+    int price2 = 10;
+    assert price2 >= 0 && price2 <= 10000; // Compliant
+    return 0;
+  }
+
+  public void setPrice1(int price3) {
+    if (price3 < 0 || price3 > 10000) {
+      throw new IllegalArgumentException("Invalid price: " + price3); // Compliant
```
m-g-sonar: what is this part for? you are not checking throw statements.

java-checks/src/test/files/checks/AssertsOnParametersOfPublicMethodCheck.java
```diff
@@ -0,0 +1,71 @@
...
+
+class C {
+
+  public synchronized void put(String a, Integer b) {
+    assert ((a != null) && (b != null)); // Noncompliant
+    assert (find2(b) == null);
+    assert (find1(a) == null);
+  }
+
+  public synchronized Integer find1(String a) {
```
m-g-sonar: do you really need to keep both cases?

java-checks/src/main/java/org/sonar/java/checks/AssertsOnParametersOfPublicMethodCheck.java
```diff
@@ -54,29 +55,22 @@ public void visitNode(Tree tree) {
       return;
     }
     MethodTree methodTree = (MethodTree) tree;
-    if (!methodTree.symbol().isPublic()) {
+    if (!methodTree.symbol().isPublic() || methodTree.symbol().owner().isPrivate()) {
```
m-g-sonar: I would drop that part about private owner. Let's look at the result once deployed on PEACH.

java-checks/src/main/java/org/sonar/java/checks/AssertsOnParametersOfPublicMethodCheck.java
```diff
@@ -54,29 +55,22 @@ public void visitNode(Tree tree) {
...
-        Tree paramParent = paramUsage.identifierToken().parent();
-        while (!paramParent.equals(methodTree) && !assertReported.contains(paramParent)) {
-          if (parentType(paramParent)) {
-            break;
-          }
-          paramParent = paramParent.parent();
-        }
-      }
-    }
+    methodTree.parameters().stream().map(VariableTree::symbol).map(Symbol::usages).flatMap(List::stream)
```
m-g-sonar: It's easier to read when all stream operations are on dedicated lines. Could you rewrite this?

PR #1956, opened 2018-03-27
SONARJAVA-2702 Fix ClassCastException when building UCFG for static method call

vilchik-elena:


java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -236,8 +237,9 @@ private void buildMethodInvocation(UCFGBuilder.BlockBuilder blockBuilder, Identi
     if (isString(mit.symbol().owner().type())) {
       // myStr.myMethod(args) -> myMethod(myStr, args)
       arguments = new ArrayList<>();
-      // safe cast, only case where methodSelect cannot be a MSE is when analysing java.lang.String source and SonarJava don't build semantic for java.lang package.
-      arguments.add(idGenerator.lookupIdFor(((MemberSelectExpressionTree) mit.methodSelect()).expression()));
+      if (mit.methodSelect().is(MEMBER_SELECT)) {
```
benzonico: That would teach me to make assumptions. Nice catch and nice fix.

PR #1958, opened 2018-03-27
SONARJAVA-2704 NullPointerException when building UCFG for unknown method

vilchik-elena:


java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -151,6 +151,26 @@ public void build_parameter_annotations() {
...
+    assertCodeToUCfg("import java.util.Set;\n" +
+      "import java.util.Collection;\n" +
+      "import java.util.stream.Collectors;\n" +
+      "public class A {\n" +
+      "  void method(Set<String> arg) { \n" +
+      "    arg.stream()\n" +
+      "      .flatMap(Collection::stream)\n" +
+      "      .collect(Collectors.toCollection(LinkedHashSet::new)); \n" +
+      "  }\n" +
+      "}", expectedUCFG);
```
benzonico: This is unreliable test : that might be fix by improvement of semantic and so make this unrelated test fail. 
A nicest way to test that is just to invoke a method named "__unknown" which is not defined in your source.

PR #1960, opened 2018-03-27
SONARJAVA-2700 Adjust the behavior of S131 to not raise an issue when "default" is not at the correct position

stylianos-agapiou-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/SwitchLastCaseIsDefaultCheck.java
```diff
@@ -51,12 +48,7 @@ public void visitNode(Tree tree) {
     }
     SwitchStatementTree switchStatementTree = (SwitchStatementTree) tree;
     Optional<CaseLabelTree> defaultLabel = getDefaultLabel(switchStatementTree);
```
m-g-sonar: You should go one step more in simplifying the rule: Change this method to return a boolean: `hasNoDefault(switchStatementTree)`

```
return allLabels(switchStatementTree).noneMatch(SwitchLastCaseIsDefaultCheck::isDefault);
```

PR #1961, opened 2018-03-27
SONARJAVA-2626 Rule S4423: Weak SSLContexts should not be used

pynicolas:


pynicolas:
I'm not sure whether we also want to support cases like:
```
public static final String PROTOCOL = "SSL";
...
context = SSLContext.getInstance(PROTOCOL);
```

java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
```diff
@@ -0,0 +1,66 @@
...
+import org.sonar.plugins.java.api.tree.Arguments;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S4423")
+public class WeakSSLContextCheck extends IssuableSubscriptionVisitor {
+
+  private static final Set<String> STRONG_PROTOCOLS = new HashSet<>(Arrays.asList("\"TLSv1.2\"", "\"DTLSv1.2\""));
```
m-g-sonar: I would prefer if you remove the quotes here, and use `org.sonar.java.model.LiteralUtils.trimQuotes(String)` on line 58. It would make it simple, and the "implementation details" of having the quotes in the string literal is handled when doing the matching.

java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (hasSemantic()) {
+      MethodInvocationTree mit = (MethodInvocationTree) tree;
+      Arguments arguments = mit.arguments();
+      if (SSLCONTEXT_GETINSTANCE_MATCHER.matches(mit)) {
+        ExpressionTree firstArgument = arguments.get(0);
+        if (firstArgument.is(Tree.Kind.STRING_LITERAL) && !STRONG_PROTOCOLS.contains(((LiteralTree) firstArgument).value())) {
+          reportIssue(mit, "Change this code to use a stronger protocol.");
```
m-g-sonar: I know it's specified that way, but I think it would be more interesting to only highlight the parameter.

java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
pynicolas: Done in 4c6ce73.

java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
pynicolas: Changed in b5571bb

PR #1964, opened 2018-03-28
SONARJAVA-2698 Update rules "All/Two branches ... with same implementation"

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/IdenticalCasesInSwitchCheck.java
```diff
@@ -142,6 +147,22 @@ private static boolean isTrivialIfStatement(StatementTree node) {
     return !node.is(Tree.Kind.BLOCK) || ((BlockTree) node).body().size() <= 1;
   }
 
+  protected static boolean hasDefaultClause(SwitchStatementTree switchStatement) {
+    return switchStatement.cases().stream()
+      .anyMatch(caseGroupTree -> caseGroupTree.labels().stream()
```
benzonico: This should be a flat map followed by a anymatch

java-checks/src/main/java/org/sonar/java/checks/IdenticalCasesInSwitchCheck.java
```diff
@@ -142,6 +147,22 @@ private static boolean isTrivialIfStatement(StatementTree node) {
...
+        .anyMatch(caseLabelTree -> caseLabelTree.caseOrDefaultKeyword().text().equals("default")));
+  }
+
+  protected static boolean hasElseClause(IfStatementTree ifStatement) {
+    StatementTree elseStatement = ifStatement.elseStatement();
+
+    while (elseStatement != null && elseStatement.is(Tree.Kind.IF_STATEMENT)) {
+      elseStatement = ((IfStatementTree) elseStatement).elseStatement();
+    }
+
```
benzonico: unecessary empty lines in that method.

java-checks/src/test/files/checks/AllBranchesAreIdentical.java
```diff
@@ -46,6 +66,11 @@ void ifStatement() {
     if (true) f(); // Noncompliant
     else f();
 
+    if(b == 0) { // Compliant: this is the exception case as defined in RSPEC
```
benzonico: Comment is a bit dodgy : maybe explain WHY it is compliant (because it has no else, and this case is handled by another rule)

PR #1967, opened 2018-03-29
SONARJAVA-2691 FP on S2637: static field should not be reported

christophe-zurn-sonarsource:


java-frontend/src/test/files/se/NonNullSetToNullCheck.java
```diff
@@ -12,6 +12,9 @@
   private String primary;
   private String secondary;
 
+  @Nonnull
+  private static final Integer STATIC_FIELD; // Compliant as static fields are not reported
```
benzonico: maybe add a case with a non annotated static field.

PR #1969, opened 2018-03-29
SONARJAVA-2667 FP on S1905 when using lambdas and default methods

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -139,9 +141,37 @@ private static Tree skipParentheses(Tree parent) {
...
+    return expressionType.isSubtypeOf(target) && !isLambdaCastNeeded(expressionType, typeCastTree, target);
   }
+
+  private static boolean isLambdaCastNeeded(Type expressionType, TypeCastTree typeCastTree, Type target) {
+    ExpressionTree expression = ExpressionUtils.skipParentheses(typeCastTree.expression());
+    if (expression.is(Tree.Kind.LAMBDA_EXPRESSION)) {
+      List<MethodJavaSymbol> childMethods = getMethodSymbolsOf(expressionType);
+      List<MethodJavaSymbol> parentMethods = getMethodSymbolsOf(target);
+
+      Predicate<JavaSymbol> isDefaultMethod = JavaSymbol::isDefault;
```
benzonico: Maybe make this as a field (even a static final one) ?

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -139,9 +141,37 @@ private static Tree skipParentheses(Tree parent) {
...
+      Set<MethodJavaSymbol> childDefaultMethods = childMethods.stream().filter(isDefaultMethod).collect(Collectors.toSet());
+      Optional<MethodJavaSymbol> childAbstractMethod = childMethods.stream().filter(isDefaultMethod.negate()).findAny();
+      Set<MethodJavaSymbol> parentAbstractMethods = parentMethods.stream().filter(isDefaultMethod.negate()).collect(Collectors.toSet());
```
benzonico: I may be wrong but I tend to think that you don't actually need to collect that one but rather anyMatch it with a predicate you would compute from the result of the previous optional. 

so you should end up with a 
```

return parentAbstractMethods.stream.filter().anyMatch(predicate)

File: java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java [hunk 3]
```diff
@@ -139,9 +141,37 @@ private static Tree skipParentheses(Tree parent) {
...
   }
+
+  private static boolean isLambdaCastNeeded(Type expressionType, TypeCastTree typeCastTree, Type target) {
+    ExpressionTree expression = ExpressionUtils.skipParentheses(typeCastTree.expression());
+    if (expression.is(Tree.Kind.LAMBDA_EXPRESSION)) {
+      List<MethodJavaSymbol> childMethods = getMethodSymbolsOf(expressionType);
+      List<MethodJavaSymbol> parentMethods = getMethodSymbolsOf(target);
+
+      Predicate<JavaSymbol> isDefaultMethod = JavaSymbol::isDefault;
+      Set<MethodJavaSymbol> childDefaultMethods = childMethods.stream().filter(isDefaultMethod).collect(Collectors.toSet());
```
benzonico: This needs to be computed only when the optional is not present and should be done as such in the correct if branch.

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
christophe-zurn-sonarsource: yep, done

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
christophe-zurn-sonarsource: It doesn't need to be collected all the time, but it needs to be collected in the following `else` branch, as the `parentAbstractMethodsSet::contains` may be called multiple times, so it cannot be replaced by a `anyMatch` in that case. I've updated the code accordingly, so that streams are not collected when not needed

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
christophe-zurn-sonarsource: True, I updated code accordingly

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -39,14 +46,11 @@
...
 
-import javax.annotation.CheckForNull;
-
-import java.util.List;
-import java.util.Set;
-
 @Rule(key = "S1905")
 public class RedundantTypeCastCheck extends IssuableSubscriptionVisitor {
 
+  private static final Predicate<JavaSymbol> ABSTRACT_METHOD_PREDICATE = ((Predicate<JavaSymbol> )JavaSymbol::isDefault).negate();
```
SonarTech: Code Smell Code Smell: Remove this unnecessary cast to "Predicate". ([squid:S1905](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS1905))

[See it in SonarQube](https://next.sonarqube.com/sonarqube/project/issues?id=org.sonarsource.java%3Ajava&pullRequest=1969&issues=AWKLTadyAnwv6XEVDnDp&open=AWKLTadyAnwv6XEVDnDp)

[ISSUE_KEY]: AWKLTadyAnwv6XEVDnDp

christophe-zurn-sonarsource:
I inversed the logic to simplify as much as possible the code.
While doing so, I found another FP on this rule. Cf. code smell above

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -139,9 +144,33 @@ private static Tree skipParentheses(Tree parent) {
...
-  private static boolean isSubtype(Type expression, Type target) {
-    return expression.isSubtypeOf(target);
+  private static boolean isUnnecessarySubtypeCast(Type childType, TypeCastTree typeCastTree, Type parentType) {
+    return childType.isSubtypeOf(parentType)
+      && (!ExpressionUtils.skipParentheses(typeCastTree.expression()).is(Tree.Kind.LAMBDA_EXPRESSION)
+        || isUnnecessaryLambdaCast(childType, parentType));
+  }
+
+  private static boolean isUnnecessaryLambdaCast(Type childType, Type parentType) {
+    if (parentType.is(childType.fullyQualifiedName())) {
```
benzonico: I wonder about this : this can become really tricky in case of parameterization with wildcard and raw type...

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -39,14 +46,11 @@
...
 
-import javax.annotation.CheckForNull;
-
-import java.util.List;
-import java.util.Set;
-
 @Rule(key = "S1905")
 public class RedundantTypeCastCheck extends IssuableSubscriptionVisitor {
 
+  private static final Predicate<JavaSymbol> ABSTRACT_METHOD_PREDICATE = symbol -> !symbol.isDefault();
```
benzonico: Dodgy name : this is just about default not only about abstract. You can have a method with a body which is not default in class that are not interfaces.

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
christophe-zurn-sonarsource: I updated the condition to `parentType.isSubtypeOf(childType)`, this avoid FP in cases cast is actually needed to access specific method(s) of a parametrized type

PR #1970, opened 2018-03-29
SONARJAVA-2713 FP on S2757: when '=!' is not used in the context of a multiple assignment

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/WrongAssignmentOperatorCheck.java
```diff
@@ -47,12 +47,17 @@ public void visitNode(Tree tree) {
...
       reportIssue(operatorToken, expressionFirstToken, "Was \"" + expressionFirstToken.text() + "=\" meant instead?");
     }
   }
 
+  private static boolean isSingleNegationAssignment(SyntaxToken firstToken, AssignmentExpressionTree aeTree) {
+    return "!".equals(firstToken.text()) && (aeTree.parent() == null || !aeTree.parent().is(Tree.Kind.ASSIGNMENT));
+  }
+
   private static boolean noSpacingBetween(SyntaxToken firstToken, SyntaxToken secondToken) {
     return firstToken.column() + firstToken.text().length() == secondToken.column();
```
m-g-sonar: I just realized that this could generate FPs when placed on different lines:
```java
target =
        +num;
```
We should check that both token are sharing the same line.

christophe-zurn-sonarsource: oh nice. good catch. I'll update accordingly

PR #1972, opened 2018-03-29
SONARJAVA-2453 RSPEC-4248 Regex patterns should not be created needlessly

stylianos-agapiou-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    ExpressionTree argument = mit.arguments().get(0);
+    if (mit.symbolType().is(PATTERN)) {
+      if (argument.is(Tree.Kind.NULL_LITERAL) || (checkArgs(argument) && invokedOnlyOnce(mit))) {
+        return;
+      }
+    } else {
+      if (!argument.is(Tree.Kind.STRING_LITERAL) ||
```
benzonico: This is not enough : what if I call one of the string method with a constant : 
```
static final String constant = "Foo";

"myString".matches(constant); // false negative
```

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,104 @@
...
+      if (parentTree.is(Tree.Kind.VARIABLE)) {
+        return ((VariableTree) parentTree).symbol().isFinal();
+      }
+      parentTree = parentTree.parent();
+    }
+    return false;
+  }
+
+  private static boolean specialCaseNotReported(ExpressionTree argument) {
+    String argString = ((LiteralTree) argument).value().replace("\"", "");
```
benzonico: This cast is a bit unsafe.

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,104 @@
...
+    int stringLength = argString.length();
+    char arg1 = argString.charAt(0);
```
benzonico: What if argString is empty ?

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,104 @@
...
+    while (!parentTree.is(Tree.Kind.METHOD, Tree.Kind.CLASS)) {
+      if (parentTree.is(Tree.Kind.VARIABLE)) {
+        return ((VariableTree) parentTree).symbol().isFinal();
+      }
+      parentTree = parentTree.parent();
+    }
+    return false;
+  }
+
+  private static boolean specialCaseNotReported(ExpressionTree argument) {
```
benzonico: name is dodgy, I struggle to understand the code of this method and what it is supposed to do. The fact that this method should only apply to split might be mentioned in its name...

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,101 @@
...
+      MethodMatcher.create().typeDefinition(STRING).name("replaceFirst").withAnyParameters());
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    ExpressionTree argument = mit.arguments().get(0);
+    if (mit.symbol().name().equals("split") && argument.is(Tree.Kind.STRING_LITERAL)) {
+      String argValue = LiteralUtils.trimQuotes(((LiteralTree) argument).value());
+      int strLength = argValue.length();
+      if (((strLength == 1 || (strLength == 2 && argValue.charAt(0) == '\\')) &&
```
m-g-sonar: I would extract this part into a dedicated method, because it's quite difficult to understand:
```
(strLength == 1 || (strLength == 2 && argValue.charAt(0) == '\\'))
```

That's the only place where you need the length, so you can directly provide `argValue` as parameter and deal with length in the metod

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,101 @@
...
+      MethodMatcher.create().typeDefinition(PATTERN).name("compile").addParameter(STRING),
+      MethodMatcher.create().typeDefinition(STRING).name("matches").withAnyParameters(),
+      MethodMatcher.create().typeDefinition(STRING).name("split").withAnyParameters(),
+      MethodMatcher.create().typeDefinition(STRING).name("replaceAll").withAnyParameters(),
+      MethodMatcher.create().typeDefinition(STRING).name("replaceFirst").withAnyParameters());
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    ExpressionTree argument = mit.arguments().get(0);
```
m-g-sonar: `firstArgument`, and I would skip parenthesis as well. `Pattern.compile(("hello"));`

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,101 @@
...
+    if (mit.symbol().name().equals("split") && argument.is(Tree.Kind.STRING_LITERAL)) {
```
m-g-sonar: move `mit.symbol().name().equals("split")` into a `isSplitMethod(mit)`

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  private static boolean isConstant(ExpressionTree expr) {
+    IdentifierTree identifierTree = null;
+    if (expr.is(Tree.Kind.IDENTIFIER)) {
+      identifierTree = (IdentifierTree) expr;
+    } else if (expr.is(Tree.Kind.MEMBER_SELECT)) {
+      identifierTree = ((MemberSelectExpressionTree) expr).identifier();
+    }
+    return identifierTree != null && identifierTree.symbol().isFinal() && identifierTree.symbol().isStatic();
```
m-g-sonar: Wondering if it would not be nicer to write it that way:
```
IdentifierTree identifierTree;
switch(expr.kind()) {
  case IDENTIFIER:
    identifierTree = (IdentifierTree) expr;
  case MEMBER_SELECT:
    identifierTree = ((MemberSelectExpressionTree) expr).identifier();
  default:
    return false;
}
Symbol symbol = identifierTree.symbol();
return symbol.isFinal() && symnol.isStatic();
```

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+    if (!storedInStaticFinal(mit) && (argument.is(Tree.Kind.STRING_LITERAL) || isConstant(argument))) {
+      reportIssue(mit, "Refactor this code to use a \"static final\" Pattern.");
+    }
+  }
+
+  private static boolean storedInStaticFinal(MethodInvocationTree mit) {
+    Tree tree = mit.parent();
+    while (!tree.is(Tree.Kind.VARIABLE, Tree.Kind.CLASS, Tree.Kind.ASSIGNMENT)) {
```
m-g-sonar: You can probably stop at `METHOD` level as well. 
If you reach a `CLASS` or a `METHOD`, before an `ASSIGNMENT` or a `VARIABLE`, it can not be stored in a static final field.

Now, you may also want to group the `KIND`s you target by category for consistency (`CLASS` and `METHOD` are failing stop conditions, `ASSIGNMENT` and `VARIABLE` allows you to investigate further)

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,101 @@
...
+      tree = tree.parent();
+    }
+    if (tree.is(Tree.Kind.CLASS)) {
+      return false;
+    } else if (tree.is(Tree.Kind.ASSIGNMENT)) {
+      return isConstant(((AssignmentExpressionTree) tree).variable());
+    }
+    VariableTree variableTree = (VariableTree) tree;
+    return variableTree.symbol().isStatic() && variableTree.symbol().isFinal();
```
m-g-sonar: Change the `isConstant` method to accept a `Tree` as argument instead of an `ExpressionTree`, and you can then use it for the variables of both `ASSIGNMENT` and `VARIABLE` trees.

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,101 @@
...
+      reportIssue(mit, "Refactor this code to use a \"static final\" Pattern.");
+    }
+  }
+
+  private static boolean storedInStaticFinal(MethodInvocationTree mit) {
+    Tree tree = mit.parent();
+    while (!tree.is(Tree.Kind.VARIABLE, Tree.Kind.CLASS, Tree.Kind.ASSIGNMENT)) {
+      tree = tree.parent();
+    }
+    if (tree.is(Tree.Kind.CLASS)) {
```
m-g-sonar: `|| tree.is(Tree.Kind.METHOD)`

its/ruling/src/test/resources/sonar-server/squid-S4248.json
```diff
@@ -0,0 +1,43 @@
...
+118,
+],
+'org.sonarsource.sonarqube:sonar-server:src/main/java/org/sonar/server/es/textsearch/ComponentTextSearchQueryFactory.java':[
+95,
+],
+'org.sonarsource.sonarqube:sonar-server:src/main/java/org/sonar/server/measure/index/ProjectsTextSearchQueryFactory.java':[
+104,
+],
+'org.sonarsource.sonarqube:sonar-server:src/main/java/org/sonar/server/platform/UrlSettings.java':[
+45,
```
m-g-sonar: Looking at this issue, I think we should change where to report issue.

java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -0,0 +1,101 @@
...
+    if (mit.symbol().name().equals("split") && argument.is(Tree.Kind.STRING_LITERAL)) {
+      String argValue = LiteralUtils.trimQuotes(((LiteralTree) argument).value());
+      int strLength = argValue.length();
+      if (((strLength == 1 || (strLength == 2 && argValue.charAt(0) == '\\')) &&
+        !SPLIT_EXCLUSION.matcher(argValue).matches()) || metaCharactersInSplit(argValue)) {
+        return;
+      }
+    }
+    if (!storedInStaticFinal(mit) && (argument.is(Tree.Kind.STRING_LITERAL) || isConstant(argument))) {
+      reportIssue(mit, "Refactor this code to use a \"static final\" Pattern.");
```
m-g-sonar: Reporting directly on the `mit` will potentially highlight way too much things if methods are chained. In the following case, 4 lines are going to be highlighted.

```
myObj
  .foo()
  .bar()
  .replaceAll(MY_REGEX, "yolo");
```

What about reporting from `ExpressionUtils.methodName(mit)` to `mit.arguments()` ? It will then only highlight the method names and include all the arguments (till the last parenthesis).

PR #1974, opened 2018-04-04
SONARJAVA-2586 S1130 should not raise issue when exception is documented but not thrown

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -55,19 +57,26 @@
...
 
+  private static final Pattern THROWS_JAVADOC_PATTERN = Pattern.compile(".*(?:@throws|@exception)\\s++(?<clazz>\\S*)(\\s++)?(?<descr>.+)?");
+
   @Override
   public List<Tree.Kind> nodesToVisit() {
     return ImmutableList.of(Tree.Kind.METHOD, Tree.Kind.CONSTRUCTOR);
   }
 
   @Override
   public void visitNode(Tree tree) {
```
SonarTech: Code Smell Code Smell: Refactor this method to reduce its Cognitive Complexity from 16 to the 15 allowed. ([squid:S3776](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS3776))

[See it in SonarQube](https://next.sonarqube.com/sonarqube/project/issues?id=org.sonarsource.java%3Ajava&pullRequest=1974&issues=AWKPucFeCiQRpXGouqtT&open=AWKPucFeCiQRpXGouqtT)

[ISSUE_KEY]: AWKPucFeCiQRpXGouqtT

java-frontend/src/main/java/org/sonar/java/resolve/JavaSymbol.java
```diff
@@ -729,6 +729,11 @@ public MethodTree declaration() {
       return declaration;
     }
 
+    @Override
+    public boolean isOverridable() {
```
m-g-sonar: This could be used in [`ConstructorCallingOverridableCheck`](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/java/org/sonar/java/checks/ConstructorCallingOverridableCheck.java#L135) as well.

There is also another implementation (testing it the other way), which is more complete, as checking for the `native` flag: https://github.com/SonarSource/sonar-java/blob/master/java-frontend/src/main/java/org/sonar/java/se/SymbolicExecutionVisitor.java#L94

You may want to have a look to it.

java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
```diff
@@ -143,14 +130,14 @@ private void visitNode(Tree tree, Tree reportTree, SymbolMetadata symbolMetadata
...
-        List<String> undocumentedParameters = javadoc.undocumentedParameters(tree);
+        Set<String> undocumentedParameters = javadoc.undocumentedParameters();
         if (!undocumentedParameters.isEmpty()) {
           context.reportIssue(this, reportTree, "Document the parameter(s): " + undocumentedParameters.stream().collect(Collectors.joining(", ")));
         }
         if (hasNonVoidReturnType(tree) && javadoc.noReturnDescription()) {
           context.reportIssue(this, reportTree, "Document this method return value.");
         }
-        List<String> undocumentedExceptions = javadoc.undocumentedThrownExceptions(tree);
+        Set<String> undocumentedExceptions = javadoc.undocumentedThrownExceptions();
```
m-g-sonar: I think it should remains a `List`, and not a `Set`. I have the feeling that with a set we could not guarantee order of the results, and therefore have some impacts on tests (flickering fail/pass depending of orders). I would stay with lists to guarantee that.

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -0,0 +1,266 @@
...
+  public Set<String> undocumentedThrownExceptions() {
+    Map<String, List<String>> thrownExceptionsMap = getJavadocExceptions();
+    List<String> exceptionNames = elementExceptionNames;
+    if (exceptionNames.size() == 1 && GENERIC_EXCEPTIONS.contains(toSimpleName(exceptionNames.get(0))) && !thrownExceptionsMap.isEmpty()) {
+      // check for documented exceptions without description when only "Exception" is declared as being thrown
+      return thrownExceptionsMap.entrySet().stream()
+        .filter(e -> isEmptyDescription(e.getValue()))
+        .map(Map.Entry::getKey)
+        .map(Javadoc::toSimpleName)
+        .collect(Collectors.toSet());
```
m-g-sonar: I would collect to a `List`.

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -0,0 +1,266 @@
...
+    }
+    return exceptionNames.stream()
+      .map(Javadoc::toSimpleName)
+      .filter(simpleName -> noDescriptionForException(thrownExceptionsMap, simpleName))
+      .collect(Collectors.toSet());
```
m-g-sonar: I would collect to a `List`.

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -0,0 +1,266 @@
...
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.TypeParameterTree;
+import org.sonar.plugins.java.api.tree.TypeTree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+public class Javadoc {
```
m-g-sonar: - `final` maybe?
- some members of the class are `public`, some are package-default. If no particular reason, I would go for all public if used externally, or private if not. If it's package-default for test, then annotated with `@VisibleForTesting`.

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -0,0 +1,266 @@
...
+  public boolean noMainDescription() {
+    return isEmptyDescription(mainDescription);
+  }
+
+  public boolean noReturnDescription() {
+    return isEmptyDescription(blockTagDescriptions.get(BlockTagKey.of(BlockTag.RETURN, null)));
+  }
+
+  public Set<String> undocumentedParameters() {
+    return elementParameters.stream()
```
m-g-sonar: Probably worth computing it only once, and store it in local private variable (to prevent being called multiple time in a row and trigger computation.)

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -0,0 +1,266 @@
...
+      .filter(name -> isEmptyDescription(blockTagDescriptions.get(BlockTagKey.of(BlockTag.PARAM, name))))
+      .collect(Collectors.toSet());
+  }
+
+  public Set<String> undocumentedThrownExceptions() {
+    Map<String, List<String>> thrownExceptionsMap = getJavadocExceptions();
```
m-g-sonar: Probably worth computing it only once, and store it in local private variable (to prevent being called multiple time in a row and trigger computation.)

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -0,0 +1,266 @@
...
+        .filter(e -> toSimpleName(e.getKey()).equals(exceptionSimpleName))
+        .map(Map.Entry::getValue)
+        .flatMap(List::stream)
+        .collect(Collectors.toList());
+    }
+    return isEmptyDescription(descriptions);
+  }
+
+  private Map<String, List<String>> getJavadocExceptions() {
+    return blockTagDescriptions.entrySet().stream()
```
m-g-sonar: probably worth being computed directly when creating the `JavaDoc` object, or at least only once if it needs to be called multiple times.

java-checks/src/test/files/checks/unused/UnusedMethodParameterCheck.java
```diff
@@ -203,30 +203,45 @@ void bar(ActionMapping mapping, ActionForm form, HttpServletRequest request, Htt
 
 class DocumentedMethod {
   /**
-   * @param firstArg
-   * @param secondArg
-   * @param fourthArg
+   * @param firstArg proper javadoc description
```
m-g-sonar: This change of behavior of `UnusedMethodParameterCheck` needs to be documented in ticket title/summary as well.

christophe-zurn-sonarsource: Updated ticket title and description to describe this behaviour change

PR #1975, opened 2018-04-04
SONARJAVA-2707 Introduce new VarTypeTree and handle highlighting

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/ast/parser/grammar/statements/TryStatementTest.java
```diff
@@ -61,4 +61,10 @@ public void java9() {
       .matches("try (super.resource; Resource resource = new Resource();) {}")
     ;
   }
+
+  @Test
+  public void java10() {
```
benzonico: dodgy name of test, I would detail it a bit more.

PR #1978, opened 2018-04-06
SONARJAVA-2541 RSPEC-4276 Functional Interfaces should be as specialised as possible

stylianos-agapiou-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,265 @@
...
+    ParameterTypeNameAndTreeType firstArgument = new ParameterTypeNameAndTreeType(args.get(0));
+    ParameterTypeNameAndTreeType secondArgument = new ParameterTypeNameAndTreeType(args.get(1));
+    if (firstArgument.paramTypeName != null && secondArgument.paramTypeName != null) {
+      return firstArgument.paramTypeName + "To" + secondArgument.paramTypeName + "Function";
+    } else if (firstArgument.paramTypeName == null && secondArgument.paramTypeName != null) {
+      return "To" + secondArgument.paramTypeName + "Function<" + firstArgument.originalName + ">";
+    } else if (firstArgument.paramTypeName != null) {
+      return firstArgument.paramTypeName + "Function<" + secondArgument.originalName + ">";
+    } else {
+      return !args.get(0).is(Tree.Kind.PARAMETERIZED_TYPE) ? ("UnaryOperator<" + firstArgument.originalName + ">") : "UnaryOperator<" + firstArgument.originalName;
```
SonarTech: Code Smell Code Smell: Add parentheses to make the operator precedence explicit. ([squid:S864](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS864))

[See it in SonarQube](https://next.sonarqube.com/sonarqube/project/issues?id=org.sonarsource.java%3Ajava&pullRequest=1978&issues=AWKZ5Bfg-RUO4oGVXasB&open=AWKZ5Bfg-RUO4oGVXasB)

[ISSUE_KEY]: AWKZ5Bfg-RUO4oGVXasB

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+import org.sonar.plugins.java.api.tree.TypeTree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+import org.sonar.plugins.java.api.tree.WildcardTree;
+
+@Rule(key = "S4276")
+public class FunctionalInterfacesSpecialisedCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.CLASS, Tree.Kind.NEW_CLASS, Tree.Kind.VARIABLE);
```
m-g-sonar: Why subscribing to `NEW_CLASS` if we are not doing anything with it in `visitNode(...)` method ?
If it is useless, remove it from the `nodesToVisit`.

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  private static final String REPORT_STRING = "Refactor this code to use the more specialised Functional Interface '";
+  private static final String UNARY_OPERATOR = "UnaryOperator<";
+
+  @Override
+  public void visitNode(Tree tree) {
+    String reportString = null;
```
m-g-sonar: only used in the case of `VARIABLE`. should be moved there.

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+    if (tree.is(Tree.Kind.CLASS)) {
+      ClassTree interfaceTree = (ClassTree) tree;
+      interfaceTree.superInterfaces().stream().map(FunctionalInterfacesSpecialisedCheck::matchInterface).filter(Objects::nonNull)
+        .forEach(reportMessage -> report(interfaceTree, reportMessage));
+    } else if (tree.is(Tree.Kind.VARIABLE)) {
+      VariableTree variableTree = (VariableTree) tree;
+      TypeTree interfaceType = variableTree.type();
+      if (variableTree.initializer() != null && variableTree.initializer().is(Tree.Kind.NEW_CLASS, Tree.Kind.LAMBDA_EXPRESSION)) {
```
m-g-sonar: This will work for lambdas, but **not** for anonymous class.

The following code will raise 2 issues:
```
class MySupplier implements Supplier<Integer> { // Issue OK
  Supplier<Integer> mySupplier = new MySupplier(); // Issue NOK! FP!! 
  // I can not fix this with 'IntSupplier mySupplier = new MySupplier();'

  @Override
  public Integer get() {
    return null;
  }
}
```

Please that the it's a proper anonymous class! (an anonymous class has a body!)
Moreover, I can not see any test with something which is not a an anonymous class or lambda.

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+
+  private static final String REPORT_STRING = "Refactor this code to use the more specialised Functional Interface '";
+  private static final String UNARY_OPERATOR = "UnaryOperator<";
+
+  @Override
+  public void visitNode(Tree tree) {
+    String reportString = null;
+    if (tree.is(Tree.Kind.CLASS)) {
+      ClassTree interfaceTree = (ClassTree) tree;
+      interfaceTree.superInterfaces().stream().map(FunctionalInterfacesSpecialisedCheck::matchInterface).filter(Objects::nonNull)
```
m-g-sonar: I think it's a bit dangerous to report issues when there is more than one interfaces, but I'm not sure.
Do you think it's valuable to report 2 issues in the following example?

```
class MySupplier implements Supplier<Integer>, Function<Integer, Integer> {

  @Override
  public Integer get() {
    return null;
  }

  @Override
  public Integer apply(Integer t) {
    return null;
  }
}
```

If it's OK for you, fine, but add an example. I can't see classes with more than on interface.

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  @CheckForNull
+  private static String matchInterface(TypeTree interfaceType) {
+    TypeArguments args = interfaceParameters(interfaceType);
+    if (args == null) {
+      return null;
+    }
+    String reportString = null;
+    switch (interfaceType.symbolType().fullyQualifiedName()) {
```
m-g-sonar: if you need the symbol type, it means the check requires semantic. Please add the `hasSemantic()` control at the beginning of the check, and a proper test.

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+      return null;
+    }
+  }
+
+  @CheckForNull
+  private static TypeArguments interfaceParameters(TypeTree interfaceType) {
+    TypeArguments args = null;
+    if (interfaceType.is(Tree.Kind.PARAMETERIZED_TYPE)) {
+      ParameterizedTypeTree parameters = (ParameterizedTypeTree) interfaceType;
+      args = parameters.typeArguments();
```
m-g-sonar: Reduce the code efficiently.
```
  return `((ParameterizedTypeTree) interfaceType).typeArguments();
}
return null;
```

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+      switch (tree.kind()) {
+        case IDENTIFIER:
+          IdentifierTree idTree = (IdentifierTree) tree;
+          this.paramSymbol = idTree.symbolType();
+          this.paramTypeName = returnStringFromJavaObject(this.paramSymbol);
+          this.originalName = idTree.name();
+          break;
+        case MEMBER_SELECT:
+          MemberSelectExpressionTree memberTree = (MemberSelectExpressionTree) tree;
+          this.paramSymbol = memberTree.identifier().symbolType();
```
m-g-sonar: why not using directly use the `symbolType()` from the `memberTree` ?

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+          this.paramTypeName = returnStringFromJavaObject(this.paramSymbol);
+          this.originalName = ((IdentifierTree) memberTree.expression()).name() + "." + memberTree.identifier().name();
```
m-g-sonar: This won't work and crash in the following case: `a.b.c.Foo`. Just use the last identifier.

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  @CheckForNull
+  private static String handleBiConsumerInterface(TypeArguments args) {
+    return "Obj" + returnStringFromJavaObject(((IdentifierTree) args.get(1)).symbolType()) + "Consumer";
+  }
+
+  @CheckForNull
+  private static String returnStringFromJavaObject(Type argType) {
+    if (argType.is("java.lang.Integer")) {
```
m-g-sonar: As a good practice, I would prefer to have it the other way (same for others): https://next.sonarqube.com/sonarqube/coding_rules?open=squid%3AS1132&rule_key=squid%3AS1132

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
```diff
@@ -0,0 +1,257 @@
...
+  private void report(Tree tree, String reportString) {
+    if (tree.is(Tree.Kind.CLASS)) {
+      reportIssue(((ClassTree) tree).superInterfaces(), REPORT_STRING + reportString + "'");
+    } else {
+      reportIssue(tree, REPORT_STRING + reportString + "'");
+    }
+  }
+
+  private static class ParameterTypeNameAndTreeType {
+    Type paramSymbol;
```
m-g-sonar: It's not a symbol, it's a type.

java-checks/src/main/java/org/sonar/java/checks/FunctionalInterfacesSpecialisedCheck.java
stylianos-agapiou-sonarsource: In my opinion, in case of two or more interfaces implemented by a class we should report only one issue on the interfaces, e.g. in the above case the message should be: "Refactor this code to use the more specialised Functional Interfaces 'IntSupplier, DoubleConsumer'".

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  public void checkClassInterfaces(ListTree<TypeTree> classInterfaces) {
+    List<InterfaceTreeAndStringPairReport> reportTreeAndStringInterfaces = classInterfaces.stream().map(SpecializedFunctionalInterfacesCheck::matchFunctionalInterface)
+      .filter(Optional::isPresent)
+      .map(Optional::get)
+      .collect(Collectors.toList());
+    if (!reportTreeAndStringInterfaces.isEmpty()) {
+      List<JavaFileScannerContext.Location> flow = reportTreeAndStringInterfaces.stream()
+        .map(interf -> new JavaFileScannerContext.Location("Interface Not Specialized", interf.classInterface))
```
m-g-sonar: I wonder about this message. Probably we should just highlight it, but without any message. I fear that it may be too much text with `"Interface Not Specialized"`.

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+        .collect(Collectors.toList());
+      if (reportTreeAndStringInterfaces.size() > 1) {
+        reportIssue(((ClassTree) classInterfaces.parent()).simpleName(),
```
m-g-sonar: rather than using `((ClassTree) classInterfaces.parent()`, simply pass the classTree to the method `checkClassInterfaces`, and then as a first step, collect the interfaces. You already had the parent ClassTree before, so it does not look efficient to me to have to get it back.

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+          REPORT_STRING + "s '" + reportTreeAndStringInterfaces.stream().map(x -> x.reportString).collect(Collectors.joining(", ")) + "'", flow, null);
```
m-g-sonar: shouldn't the separator be `', '` instead?

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+      } else {
+        reportIssue(((ClassTree) classInterfaces.parent()).simpleName(), REPORT_STRING + " '" + reportTreeAndStringInterfaces.get(0).reportString + "'", flow, null);
```
m-g-sonar: you should extract a method to get the report message, it would simplify your life:`reportMessage(String ... interfaces)`

```
private static String reportMessage(String interface) {
  reportMessage(Collections.singletonList(interface));
}

private static String reportMessage(List<String> interfaces) {
    String functionalInterfaces = interfaces.stream().collect(Collectors.joining(
      "', '",
      (interfaces.size() > 1 ? "s" : "") + " '",
      "'."));
    return String.format("Refactor this code to use the more specialised Functional Interface%s", functionalInterfaces);
  }

```

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+          break;
+        case PARAMETERIZED_TYPE:
+          ParameterizedTypeTree p = (ParameterizedTypeTree) tree;
+          paramTypeName = null;
+          paramType = p.symbolType();
+          break;
+        case EXTENDS_WILDCARD:
+        case UNBOUNDED_WILDCARD:
+        case SUPER_WILDCARD:
+          WildcardTree w = (WildcardTree) tree;
```
m-g-sonar: single usage, skip the variable. make field order consistent.

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+
+  private static Optional<TypeArguments> interfaceParameters(TypeTree interfaceType) {
+    if (interfaceType.is(Tree.Kind.PARAMETERIZED_TYPE)) {
+      return Optional.of(((ParameterizedTypeTree) interfaceType).typeArguments());
+    }
+    return Optional.empty();
+  }
+
+  private static class InterfaceTreeAndStringPairReport {
+    String reportString;
```
m-g-sonar: both fields should be final.

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+        case "java.util.function.Consumer":
+        case "java.util.function.Predicate":
+        case "java.util.function.UnaryOperator":
+        case "java.util.function.BinaryOperator":
+          reportString = handleOneParameterInterface(arguments).map(s -> s + interfaceType.symbolType().name());
+          break;
+        default:
+          return Optional.empty();
+      }
+      if (reportString.isPresent()) {
```
m-g-sonar: you don't need this test, you can directly map:
```
return reportString.map(rs -> new InterfaceAndStringParReport(rs, interfaceType));
```

If it's empty, then nothing will be done, and empty will be returned.

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+        return Optional.of(new InterfaceTreeAndStringPairReport(reportString.get(), interfaceType));
+      }
+    }
+    return Optional.empty();
+  }
+
+  private static Optional<String> handleOneParameterInterface(TypeArguments args) {
+    String reportString = new ParameterTypeNameAndTreeType(args.get(0)).paramTypeName;
+    if (reportString != null) {
```
m-g-sonar: This whole test can be replaced by `return Optional.ofNullable(reportString);`.
It will automatically create an empty optional if its null.

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  private static Optional<String> handleFunctionInterface(TypeArguments args) {
+    ParameterTypeNameAndTreeType firstArgument = new ParameterTypeNameAndTreeType(args.get(0));
+    ParameterTypeNameAndTreeType secondArgument = new ParameterTypeNameAndTreeType(args.get(1));
+    String reportString = null;
+    if (firstArgument.paramType.equals(secondArgument.paramType)) {
+      reportString = "UnaryOperator<" + firstArgument.paramType.name() + ">";
+    } else if (firstArgument.paramTypeName != null && secondArgument.paramTypeName != null) {
+      reportString = firstArgument.paramTypeName + "To" + secondArgument.paramTypeName + "Function";
```
m-g-sonar: `String.format("%sTo%sFunction", firstArgument.paramTypeName, secondArgument.paramTypeName)`

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+    } else if (firstArgument.paramTypeName == null && secondArgument.paramTypeName != null) {
+      reportString = "To" + secondArgument.paramTypeName + "Function<" + firstArgument.paramType.name() + ">";
```
m-g-sonar: `String.format("To%sFunction<%s>", secondArgument.paramTypeName, firstArgument.paramTypeName)`

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+    return Optional.empty();
+  }
+
+  private static Optional<String> handleFunctionInterface(TypeArguments args) {
+    ParameterTypeNameAndTreeType firstArgument = new ParameterTypeNameAndTreeType(args.get(0));
+    ParameterTypeNameAndTreeType secondArgument = new ParameterTypeNameAndTreeType(args.get(1));
+    String reportString = null;
+    if (firstArgument.paramType.equals(secondArgument.paramType)) {
+      reportString = "UnaryOperator<" + firstArgument.paramType.name() + ">";
```
m-g-sonar: `String.format("UnaryOperator<%s>", firstArgument.paramType.name())`

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+    } else if (firstArgument.paramTypeName != null && secondArgument.paramTypeName != null) {
+      reportString = firstArgument.paramTypeName + "To" + secondArgument.paramTypeName + "Function";
+    } else if (firstArgument.paramTypeName == null && secondArgument.paramTypeName != null) {
+      reportString = "To" + secondArgument.paramTypeName + "Function<" + firstArgument.paramType.name() + ">";
+    } else if (firstArgument.paramTypeName != null) {
+      reportString = firstArgument.paramTypeName + "Function<" + secondArgument.paramType.name() + ">";
```
m-g-sonar: `String.format("%sFunction<%s>", firstArgument.paramTypeName, secondArgument.paramType.name())`

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+    ParameterTypeNameAndTreeType thirdArgument = new ParameterTypeNameAndTreeType(args.get(2));
+    if (firstArgument.paramType.equals(secondArgument.paramType) && firstArgument.paramType.equals(thirdArgument.paramType)) {
+      return Optional.of("BinaryOperator<" + firstArgument.paramType.name() + ">");
+    }
+    return Optional.empty();
+  }
+
+  private static Optional<String> handleBiConsumerInterface(TypeArguments args) {
+    String reportString = returnStringFromJavaObject(((IdentifierTree) args.get(1)).symbolType());
+    if (reportString != null) {
```
m-g-sonar: `return Optional.ofNullable(...)`

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+        reportIssue(((ClassTree) classInterfaces.parent()).simpleName(), REPORT_STRING + " '" + reportTreeAndStringInterfaces.get(0).reportString + "'", flow, null);
+      }
+    }
+  }
+
+  public void checkVariableTypeAndInitializer(VariableTree variableTree) {
+    TypeTree interfaceType = variableTree.type();
+    ExpressionTree initializer = variableTree.initializer();
+    if (initializer != null && (initializer.is(Tree.Kind.LAMBDA_EXPRESSION) || isAnonymousClass(initializer))) {
+      matchFunctionalInterface(interfaceType)
```
m-g-sonar: inline `variableTree.type()`

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+    if (firstArgument.paramType.equals(secondArgument.paramType) && firstArgument.paramType.equals(thirdArgument.paramType)) {
+      return Optional.of("BinaryOperator<" + firstArgument.paramType.name() + ">");
+    }
+    return Optional.empty();
+  }
+
+  private static Optional<String> handleBiConsumerInterface(TypeArguments args) {
+    String reportString = returnStringFromJavaObject(((IdentifierTree) args.get(1)).symbolType());
+    if (reportString != null) {
+      return Optional.of("Obj" + reportString + "Consumer");
```
m-g-sonar: And what about the first argument ?!

From what I can see here, `Obj...Consumer` interfaces are paremeterized! https://docs.oracle.com/javase/8/docs/api/java/util/function/ObjIntConsumer.html

It should be something like `ObjIntConsumer<T>`, with T being the other parameter.
Please update the rule as well. (You should have seen that!)

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,239 @@
...
+      return Optional.of(reportString);
+    }
+    return Optional.empty();
+  }
+
+  private static Optional<String> handleBiFunctionInterface(TypeArguments args) {
+    ParameterTypeNameAndTreeType firstArgument = new ParameterTypeNameAndTreeType(args.get(0));
+    ParameterTypeNameAndTreeType secondArgument = new ParameterTypeNameAndTreeType(args.get(1));
+    ParameterTypeNameAndTreeType thirdArgument = new ParameterTypeNameAndTreeType(args.get(2));
+    if (firstArgument.paramType.equals(secondArgument.paramType) && firstArgument.paramType.equals(thirdArgument.paramType)) {
```
m-g-sonar: How do you know that `firstArgument.paramType` is not going to be null ?! I see a NPE hidden here.

stylianos-agapiou-sonarsource: I think that this won't happen, so no reason to be nullable.

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,230 @@
...
+          paramTypeName = null;
+          break;
+        case EXTENDS_WILDCARD:
+        case UNBOUNDED_WILDCARD:
+        case SUPER_WILDCARD:
+          paramType = ((WildcardTree) tree).symbolType();
+          paramTypeName = null;
+          break;
+        default:
+          break;
```
m-g-sonar: if you reach default, then paramType will be null.

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,230 @@
...
+
+    public InterfaceTreeAndStringPairReport(String report, TypeTree interf) {
+      reportString = report;
+      classInterface = interf;
+    }
+  }
+
+  private static class ParameterTypeNameAndTreeType {
+
+    Type paramType;
```
m-g-sonar: should be final to guarantee initialization by constructor

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,230 @@
...
+
+    @Nullable
+    String paramTypeName;
```
m-g-sonar: should be final to guarantee initialization by constructor

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,230 @@
...
+
+    public ParameterTypeNameAndTreeType(Tree tree) {
+      switch (tree.kind()) {
```
m-g-sonar: the default is not covered. Did you tried? (it is) If it's not possible, then throw an IllegalStateException in such case, with an adequate message (look in other checks).

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
m-g-sonar:
```
BiFunction<Integer[][][], Integer[][][], Integer[][][]> myBiFunc = new BiFunction<Integer[][][], Integer[][][], Integer[][][]>() {
      @Override
      public Integer[][][] apply(Integer[][][] t, Integer[][][] u) {
        return null;
      }
    };
```

java-checks/src/main/java/org/sonar/java/checks/SpecializedFunctionalInterfacesCheck.java
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  private static class ParameterTypeNameAndTreeType {
+
+    Type paramType;
+
+    @Nullable
+    String paramTypeName;
+
+    public ParameterTypeNameAndTreeType(Tree tree) {
```
m-g-sonar: I'm more and more convinced that this should work directly and only on semantic, not on syntax. You have the type of the variable/class, so you should be able to access directly the arguments type of the parameterized type (See `org.sonar.java.resolve.ParametrizedTypeJavaType` and the way it is used in other checks).

PR #1979, opened 2018-04-06
SONARJAVA-2231 FP on S128 (Switch cases without break) with nested switches

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/SwitchCaseWithoutBreakCheck.java
```diff
@@ -21,66 +21,65 @@
...
-        CaseLabelTree caseLabel = caseGroup.labels().get(caseGroup.labels().size() - 1);
+  public void visitNode(Tree tree) {
+    SwitchStatementTree switchStatementTree = (SwitchStatementTree) tree;
+    List<CaseGroupTree> caseGroupTrees = switchStatementTree.cases();
+    CFG cfg = CFG.buildCFG(Collections.singletonList(tree), true);
+    CFG.Block entry = cfg.entry();
+    List<CFG.Block> cfgCases = new ArrayList<>(entry.successors());
+
+    if (!hasDefaultClause(switchStatementTree)) {
+      cfgCases.remove(cfgCases.size() - 1);
```
pynicolas: `CFG.Block.successors` returns a `Set<Block>` and I don't think that's any guarantee on the order of the `case` blocks.

java-checks/src/main/java/org/sonar/java/checks/SwitchCaseWithoutBreakCheck.java
```diff
@@ -21,66 +21,65 @@
...
+      .map(caseGroupLabels -> caseGroupLabels.get(caseGroupLabels.size() - 1))
+      .forEach(label -> reportIssue(label, "End this switch case with an unconditional break, return or throw statement."));
   }
 
-  private static boolean intentionalFallThrough(SwitchStatementTree switchStatement, CaseGroupTree caseGroup) {
-    FallThroughCommentVisitor visitor = new FallThroughCommentVisitor();
+  private static boolean hasDefaultClause(SwitchStatementTree switchStatement) {
+    return switchStatement.cases().stream()
+      .flatMap(caseGroupTree -> caseGroupTree.labels().stream())
+      .anyMatch(caseLabelTree -> caseLabelTree.caseOrDefaultKeyword().text().equals("default"));
```
pynicolas: It would be slightly better to use `JavaKeyword.DEFAULT.getValue()`.

java-checks/src/main/java/org/sonar/java/checks/SwitchCaseWithoutBreakCheck.java
```diff
@@ -21,66 +21,65 @@
...
+    FallThroughCommentVisitor visitor = new FallThroughCommentVisitor();
     List<Tree> treesToScan = ImmutableList.<Tree>builder().addAll(caseGroup.body()).add(nextCaseGroup.firstToken()).build();
     visitor.scan(treesToScan);
     return visitor.hasComment;
   }
 
   private static class FallThroughCommentVisitor extends SubscriptionVisitor {
-
-    private static final Pattern FALL_THROUGH_PATTERN = Pattern.compile("falls?\\-?thro?u[gh]?", Pattern.CASE_INSENSITIVE);
+    private static final Pattern FALL_THROUGH_PATTERN = Pattern.compile("falls?[\\-\\s]?thro?u[gh]?", Pattern.CASE_INSENSITIVE);
```
pynicolas: This change should be documented in a separate JIRA ticket.

java-checks/src/main/java/org/sonar/java/checks/SwitchCaseWithoutBreakCheck.java
```diff
@@ -21,66 +21,65 @@
...
 
-          // Reverse the body as commonly the unconditional exit will be at the end of the body.
-        if (Lists.reverse(caseGroup.body()).stream().noneMatch(SwitchCaseWithoutBreakCheck::isUnconditionalExit) && !intentionalFallThrough(switchStatement, caseGroup)) {
-          context.reportIssue(this, caseLabel, "End this switch case with an unconditional break, return or throw statement.");
-        }
-      });
+    List<CFG.Block> cases = Lists.reverse(cfgCases);
 
-    super.visitSwitchStatement(switchStatement);
+    IntStream.range(1, cases.size())
```
pynicolas: I would prefer to iterate over instances of `Block` or `CaseGroupTree` rather than on integers.

java-checks/src/main/java/org/sonar/java/checks/SwitchCaseWithoutBreakCheck.java
christophe-zurn-sonarsource: Created SONARJAVA-2717 for this

java-checks/src/main/java/org/sonar/java/checks/SwitchCaseWithoutBreakCheck.java
```diff
@@ -20,66 +20,92 @@
...
+    switchSuccessors.forEach(cfgBlock -> cfgBlock.elements().stream()
+      .filter(element -> element.is(Tree.Kind.CASE_GROUP))
+      .filter(caseGroupTrees::contains)
+      .findAny()
+      .ifPresent(caseGroup -> mapping.put(cfgBlock, (CaseGroupTree) caseGroup)));
+    return mapping;
+  }
 
-    super.visitSwitchStatement(switchStatement);
+  private static Stream<CaseGroupTree> getCaseGroupPredecessorStream(CFG.Block cfgBlock, Map<CFG.Block, CaseGroupTree> cfgBlockToCaseGroupMap) {
```
pynicolas: Suggested name: `getForbiddenCaseGroupPredecessors`.

java-checks/src/main/java/org/sonar/java/checks/SwitchCaseWithoutBreakCheck.java
```diff
@@ -20,66 +20,92 @@
...
+  @Nullable
+  private static CaseGroupTree getCaseGroupPredecessor(CFG.Block predecessor, Map<CFG.Block, CaseGroupTree> cfgBlockToCaseGroupMap) {
+    if (cfgBlockToCaseGroupMap.get(predecessor) != null) {
+      return cfgBlockToCaseGroupMap.get(predecessor);
+    }
+
+    Set<CFG.Block> previousPredecessors = predecessor.predecessors();
+    for (CFG.Block previousPredecessor : previousPredecessors) {
+      CaseGroupTree caseGroupTree = getCaseGroupPredecessor(previousPredecessor, cfgBlockToCaseGroupMap);
+      if (caseGroupTree != null) {
```
pynicolas: Removing this condition does not make unit tests fail.

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -705,6 +705,7 @@ private void buildSwitchStatement(SwitchStatementTree switchStatementTree) {
         if (!hasDefaultCase) {
           hasDefaultCase = containsDefaultCase(caseGroupTree.labels());
         }
+        currentBlock.elements.add(caseGroupTree);
```
pynicolas: @m-g-sonar Are you ok to add `caseGroupTree` as a block element in the CFG? (See also the tests below) Is it the right order of the elements in the block?

m-g-sonar: Worth having a comment to mention why we need it. If it's simplify implementation of the rule, let's document it.

java-frontend/src/test/java/org/sonar/java/cfg/CFGTest.java
```diff
@@ -622,18 +627,21 @@ public void switch_statement_with_piledUpCases_againstDefault() {
...
       block(
+        element(Kind.CASE_GROUP),
         element(INT_LITERAL, "2"),
         element(Tree.Kind.IDENTIFIER, "System"),
         element(Tree.Kind.MEMBER_SELECT),
         element(Tree.Kind.IDENTIFIER, "qix"),
         element(Tree.Kind.METHOD_INVOCATION)).terminator(Tree.Kind.BREAK_STATEMENT).successors(0),
       block(
+        element(Kind.CASE_GROUP),
         element(INT_LITERAL, "4"),
```
m-g-sonar: I find this a bit suspicious (i would have expected the inverted order), but it's obviously out of scope of the current ticket.

java-frontend/src/test/java/org/sonar/java/cfg/CFGTest.java
```diff
@@ -679,12 +689,14 @@ public void switch_statement_with_expression_in_case() {
       "void fun() { int a; switch(b) { case c : System.out.println(1);break; case d || e: System.out.println(2);break;} }");
```
m-g-sonar: even if syntactically valid, this won't compile. the boolean OR operator can not be used here, as the switch can not be used with a boolean as parameter. (also out of scope of the ticket)

christophe-zurn-sonarsource:
Related to the out-of-scope comments, I created these two Jira tickets:
https://jira.sonarsource.com/browse/SONARJAVA-2720
https://jira.sonarsource.com/browse/SONARJAVA-2721

PR #1989, opened 2018-04-11
SONARJAVA-2603 FP on S1872 when the class name is not obtained from "xxx.getClass()"

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ClassComparedByNameCheck.java
```diff
@@ -46,57 +42,43 @@
...
+
+    // Only check comparison for string literals and use of class#getName methods to avoid FP. Ref: SONARJAVA-2603
+    boolean firstOperandNonCompliant = useClassGetNameMethod(firstOperand);
+    boolean secondOperandNonCompliant = useClassGetNameMethod(secondOperand);
+
+    if (firstOperandNonCompliant && secondOperandNonCompliant) {
+      reportIssue(mit, "Use \"isAssignableFrom\" instead.");
+    } else if (firstOperandNonCompliant && secondOperand.is(Tree.Kind.STRING_LITERAL)) {
+      reportIssue(mit, "Use an \"instanceof\" comparison instead.");
+    } else if (secondOperandNonCompliant && firstOperand.is(Tree.Kind.STRING_LITERAL)) {
```
m-g-sonar: could be a OR in previous if

java-checks/src/main/java/org/sonar/java/checks/ClassComparedByNameCheck.java
```diff
@@ -46,57 +42,43 @@
...
+    expression.accept(visitor);
+    return visitor.useClassGetName;
   }
 
   private static class ClassGetNameDetector extends BaseTreeVisitor {
-
     private boolean useClassGetName = false;
-    private boolean useStackTraceElementGetClassName = false;
 
     private final MethodMatcherCollection methodMatchers = MethodMatcherCollection.create(
```
m-g-sonar: should probably be static, we don't need to recompute it for every new instance

java-checks/src/main/java/org/sonar/java/checks/ClassComparedByNameCheck.java
```diff
@@ -46,57 +42,43 @@
...
-      String message = "Use an \"instanceof\" comparison instead.";
-      if(useAssignableMessage) {
-        message = "Use \"isAssignableFrom\" instead.";
-      }
-      reportIssue(mit, message);
+    ExpressionTree firstOperand = ExpressionUtils.skipParentheses(((MemberSelectExpressionTree) mit.methodSelect()).expression());
+    ExpressionTree secondOperand = ExpressionUtils.skipParentheses(mit.arguments().get(0));
+
+    // Only check comparison for string literals and use of class#getName methods to avoid FP. Ref: SONARJAVA-2603
+    boolean firstOperandNonCompliant = useClassGetNameMethod(firstOperand);
```
m-g-sonar: I would prefer something more explicit: `firstOpIsGetName`, `secondOpIsGetName` ?

christophe-zurn-sonarsource:
Yes, it's the better solution for now, without having a symbolic execution engine.

PR #1997, opened 2018-04-12
SONARJAVA-2247 RSPEC-3973 Conditionally executed code should be denoted by either indentation or curly braces

stylianos-agapiou-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/IndentationAfterConditionalCheck.java
```diff
@@ -0,0 +1,172 @@
...
+import org.sonar.plugins.java.api.tree.ForStatementTree;
+import org.sonar.plugins.java.api.tree.IfStatementTree;
+import org.sonar.plugins.java.api.tree.StatementTree;
+import org.sonar.plugins.java.api.tree.SyntaxTrivia;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.WhileStatementTree;
+
+@Rule(key = "S3973")
+public class IndentationAfterConditionalCheck extends BaseTreeVisitor implements JavaFileScanner {
+
```
christophe-zurn-sonarsource: I think it would be better to inherit `IssuableSubscriptionVisitor` and subscribe to the relevant `Tree.Kind` instead, it would avoid having duplicated/similar code in the `visit[If|While|For|ForEach]Statement` methods.
That way you only have to implement the `visitNode` method and use `tree.firstToken()` to retrieve the keyword

java-checks/src/main/java/org/sonar/java/checks/IndentationAfterConditionalCheck.java
```diff
@@ -0,0 +1,172 @@
...
+  @Override
+  public void visitForEachStatement(ForEachStatement tree) {
+    if (thereIsNoIssue(tree)) {
+      return;
+    }
+    report(tree.forKeyword(), tree.closeParenToken(), tree.forKeyword().text(), findSecondariesToReport(tree));
+    super.visitForEachStatement(tree);
+  }
+
+  private static boolean thereIsNoIssue(Tree tree) {
```
christophe-zurn-sonarsource: Ok. I tend to prefer concise naming so that code stays clear to read, so I would have simply named the method `noIssue`, but it is personal preferences

java-checks/src/main/java/org/sonar/java/checks/IndentationAfterConditionalCheck.java
```diff
@@ -0,0 +1,172 @@
...
+    return (blockExists(tree) || checkIdentation(tree));
```
christophe-zurn-sonarsource: There is a typo in `checkIdentation`, and the naming is not entirely clear. I would expect a method named check* to actually do something (like raising an issue or throwing an error), if it simply returns a boolean I'd rather name it something like `isIndentationCorrect`

java-checks/src/main/java/org/sonar/java/checks/IndentationAfterConditionalCheck.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  public void report(Tree firstTree, Tree secondTree, String operationName, List<JavaFileScannerContext.Location> flow) {
+    context.reportIssue(this, firstTree, secondTree,
+      "Use curly braces or indentation to denote the code conditionally executed by this \"" + operationName + "\".", flow, null);
+  }
+
+  private static boolean checkElseStatementReports(IfStatementTree tree) {
```
christophe-zurn-sonarsource: why report**s**?

java-checks/src/main/java/org/sonar/java/checks/IndentationAfterConditionalCheck.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+    return false;
+  }
+
+  private static List<JavaFileScannerContext.Location> findSecondariesToReport(Tree tree) {
+    Tree parentTree = tree.parent();
+    List<Tree> secondaryLinesToReport = new ArrayList<>();
+    StatementTree statement = returnStatement(tree);
+    secondaryLinesToReport.add(statement);
+    if (parentTree.is(Tree.Kind.BLOCK)) {
```
christophe-zurn-sonarsource: What about the case where the parent is an if/for/while/switch-case ?
A more general solution could be to implement a `BaseTreeVisitor` to scan the parent tree and check until what line the following statements/expressions go to

java-checks/src/main/java/org/sonar/java/checks/IndentationAfterConditionalCheck.java
```diff
@@ -0,0 +1,172 @@
...
+  private static boolean linesBeforeCurrentAreNotCommented(List<Integer> syntaxTrivia, int prevLine) {
+    return !syntaxTrivia.contains(prevLine);
+  }
+
+  private static int columnStatement(Tree tree) {
+    return tree.firstToken().column();
+  }
+
+  private static boolean checkIdentation(Tree tree) {
+    return returnStatement(tree).firstToken().column() > columnStatement(tree);
```
christophe-zurn-sonarsource: You can also use invocation to `columnStatement` in the first part of the condition

java-checks/src/main/java/org/sonar/java/checks/IndentationAfterConditionalCheck.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private JavaFileScannerContext context;
+
+  private Set<IfStatementTree> statementSet = new HashSet<>();
+
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    this.context = context;
+    statementSet.clear();
+    scan(context.getTree());
```
christophe-zurn-sonarsource: You should clear the set after the scan to, to be sure that references to the `IfStatementTree` are not kept for the whole scanning process.
Actually, I think I'd prefer to avoid using a set altogether by using the parent to retrieve the `else` keyword (i.e. if the parent is a `IfStatementTree` and current if is the else statement of the parent). What do you think?

stylianos-agapiou-sonarsource: Nice idea!

java-checks/src/main/java/org/sonar/java/checks/IndentationAfterConditionalCheck.java
```diff
@@ -0,0 +1,86 @@
...
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    this.context = context;
+    scan(context.getTree());
+  }
+
+  @Override
+  public void visitIfStatement(IfStatementTree tree) {
+    Tree parentTree = tree.parent();
+    if (parentTree.is(Tree.Kind.IF_STATEMENT) && ((IfStatementTree) parentTree).elseKeyword() != null) {
```
christophe-zurn-sonarsource: You should replace condition `((IfStatementTree) parentTree).elseKeyword() != null` with `tree.equals(((IfStatementTree) parentTree).elseStatement())` because current `if` can be a child of an if-statement without being the `else if` statement (i.e. the current if-statement is the `thenStatement` of the parent).
Ex:
```
    if (i > 0)
      if (i >1) // Noncompliant but issue not caught by current implementation 
      doTheThing();
      else
        doTheOtherThing();
    else
      somethingElseEntirely();
```

christophe-zurn-sonarsource: You can add this test case in the unit test too :)

PR #2000, opened 2018-04-29
Update README.md : increase SonarQube version

olcbean:
The minimum version of SonarQube Server is `6.7`


benzonico:
Thanks for the fix and congrats for opening the 2000th PR on this project ! :tada:

PR #2001, opened 2018-04-30
SONARJAVA-2742 Prevent ucfg serialization for method with unkown args…

benzonico:
… and return types

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
+++ b/java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
@@ -19,6 +19,7 @@
  */
 package org.sonar.java;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
 import java.io.File;
@@ -64,6 +65,7 @@ import org.sonar.ucfg.UCFGBuilder.BlockBuilder;
 import org.sonar.ucfg.UCFGtoProtobuf;
 
 import static org.sonar.plugins.java.api.tree.Tree.Kind.ASSIGNMENT;
+import static org.sonar.plugins.java.api.tree.Tree.Kind.CONSTRUCTOR;
 import static org.sonar.plugins.java.api.tree.Tree.Kind.MEMBER_SELECT;
 import static org.sonar.plugins.java.api.tree.Tree.Kind.METHOD_INVOCATION;
 import static org.sonar.plugins.java.api.tree.Tree.Kind.PLUS;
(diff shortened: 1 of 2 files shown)


java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -93,14 +94,25 @@ public void scanFile(JavaFileScannerContext context) {
   @Override
   public void visitMethod(MethodTree tree) {
     super.visitMethod(tree);
-    if (tree.block() != null) {
+    Stream<Type> types;
+    if("<init>".equals(tree.symbol().name())) {
+      types = tree.symbol().parameterTypes().stream();
+    } else {
+      types = Stream.concat(Stream.of(tree.symbol().returnType().type()), tree.symbol().parameterTypes().stream());
```
m-g-sonar: `tree.symbol()` reused 3 times, make it a variable.

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -93,14 +94,25 @@ public void scanFile(JavaFileScannerContext context) {
   @Override
   public void visitMethod(MethodTree tree) {
     super.visitMethod(tree);
-    if (tree.block() != null) {
+    Stream<Type> types;
```
m-g-sonar: What about using a simple `List<Type>` first, being directly `tree.symbol().parameterTypes();`, and then only if it's a constructor, simply add the return type to a (new) list including parameters. Then you can call `types.stream()` only once before calling `noneMatch`.

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -325,21 +345,26 @@ private UCFG assertCodeToUCfg(String source, UCFG expectedUCFG, boolean testLoca
...
-    if(testLocations) {
-      Stream<LocationInFile> locStream = toLocationStream(actualUCFG);
-      assertThat(locStream).containsExactlyElementsOf(toLocationStream(expectedUCFG).collect(Collectors.toList()));
-    }
     return actualUCFG;
   }
 
+  @Test
+  public void method_with_unknown_symbol_should_not_produce_ucfg() {
+    CompilationUnitTree cut = getCompilationUnitTreeWithSemantics("class A {void foo(UnkownType t) {return t;}}");
```
m-g-sonar: move this line next to the `UCFGJavaVisitor.visitCompilationUnit(cut);` line, in order to have both test cases close to each other. Difference is not explicit at first glance. Maybe add a comment on top of each case.

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -325,21 +345,26 @@ private UCFG assertCodeToUCfg(String source, UCFG expectedUCFG, boolean testLoca
...
+      @Override
+      protected void serializeUCFG(MethodTree tree, CFG cfg) {
+       fail("should not serialize a UCFG of a method with unknown parameters");
+      }
+    };
+    UCFGJavaVisitor.fileKey = FILE_KEY;
+    UCFGJavaVisitor.visitCompilationUnit(cut);
+    cut = getCompilationUnitTreeWithSemantics("class A {UnkownType foo(String t) {return t;}}");
+    UCFGJavaVisitor.visitCompilationUnit(cut);
+
```
m-g-sonar: useless empty lines

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -93,14 +94,25 @@ public void scanFile(JavaFileScannerContext context) {
...
+    if("<init>".equals(tree.symbol().name())) {
```
m-g-sonar: `tree.is(Kind.CONSTRUCTOR)` ?

PR #2004, opened 2018-05-04
SONARJAVA-2728 : Improve secondary issue reporting for S2234

olcbean:
Report secondary location only for the misplace parameters for S2234
( Parameters should be passed in the correct order )

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
@@ -68,13 +68,14 @@ public class MethodParametersOrderCheck extends IssuableSubscriptionVisitor {
     }
     ParametersList formalParameterList = parametersByMethod.computeIfAbsent(methodInvTree.symbol(), m -> new ParametersList(methodDeclaration));
     List<IdentifierTree> argumentsList = methodInvTree.arguments().stream().map(this::argumentToIdentifier).collect(Collectors.toList());
-    if (matchingNames(formalParameterList, argumentsList)
-      && matchingTypesWrongOrder(formalParameterList, argumentsList)) {
-      List<JavaFileScannerContext.Location> flow = methodDeclaration.parameters().stream().map(param -> new JavaFileScannerContext.Location("Misplaced Parameters", param))
-        .collect(Collectors.toList());
-      reportIssue(methodInvTree.arguments(), "Parameters to " + methodInvTree.symbol().name() + " have the same names but not the same order as the method arguments.",
-        flow, null);
-
+    if (matchingNames(formalParameterList, argumentsList)) {
+      List<VariableTree> matchingTypesWrongOrder = matchingTypesWrongOrder(formalParameterList, argumentsList);
+      if (!matchingTypesWrongOrder.isEmpty()) {
+        List<JavaFileScannerContext.Location> flow = matchingTypesWrongOrder.stream().map(param -> new JavaFileScannerContext.Location("Misplaced Parameters", param))
+          .collect(Collectors.toList());
+        reportIssue(methodInvTree.arguments(), "Parameters to " + methodInvTree.symbol().name() + " have the same names but not the same order as the method arguments.", flow,
+          null);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


m-g-sonar:
hey @olcbean,

I think you are only halfway the fix we were expecting.  
From what I can see, we are adding a "Formal/misplaced parameters" to all of the parameters in the method declaration, while in fact we would expect to have secondaries only on the one which are actually wrongly placed, accordingly to the method invocation. So for instance in the following case:

```java
void foo(
  int a,        //  Misplaced parameter
  int b,        //  parameter ok -> no message
  int c         //  Misplaced parameter       
) { /* ... */}

void bar(int a, int b, int c) {
  foo(c, b, a);
}
```

Regarding testing, there is currently no way in our `JavaCheckVerifier` to test the messages associated to each secondaries... However you can test the **lines** associated to these secondaries. If you spread definition of parameters in multiple lines, you will then be able to test it by adding the `secondaries` part of the issue classifier. Probably using the something similar as the following (assuming `int a` and `int b` would have been defined on line 4 and 6):

```
  foo(c, b, a); // Noncompliant [[secondaries=4,6]] {{bla bla bla}}
```

m-g-sonar:
Forgot to say it, but of course, thanks for your contribution as well! :)

olcbean:
:facepalm: I totally misread this one, sorry.

@m-g-sonar thanks for the detailed explanation and the example!

I just pushed another commit to address the task at hand and corrected the description accordingly.

Please let me know what you think!

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -68,13 +68,14 @@ public void visitNode(Tree tree) {
...
-      && matchingTypesWrongOrder(formalParameterList, argumentsList)) {
-      List<JavaFileScannerContext.Location> flow = methodDeclaration.parameters().stream().map(param -> new JavaFileScannerContext.Location("Formal Parameters", param))
-        .collect(Collectors.toList());
-      reportIssue(methodInvTree.arguments(), "Parameters to " + methodInvTree.symbol().name() + " have the same names but not the same order as the method arguments.",
-        flow, null);
-
+    if (matchingNames(formalParameterList, argumentsList)) {
+      List<VariableTree> matchingTypesWrongOrder = matchingTypesWrongOrder(formalParameterList, argumentsList);
+      if (!matchingTypesWrongOrder.isEmpty()) {
+        List<JavaFileScannerContext.Location> flow = matchingTypesWrongOrder.stream().map(param -> new JavaFileScannerContext.Location("Formal Parameters", param))
```
m-g-sonar: As this was also causing some trouble, I would still go for the change you previously did: replacing the strange `Formal parameters` by `Misplaced parameter`.

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -98,28 +99,32 @@ public static boolean allUnique(List<String> argListNames) {
...
-    int countArgumentsNotOrdered = 0;
+    List<VariableTree> misplacedParameters = new ArrayList<>();
     while (argumentsIterator.hasNext()) {
       IdentifierTree argument = argumentsIterator.next();
       int index = formalParameterList.indexOf(argument.name().toLowerCase(Locale.ENGLISH));
       Type formalType = formalParameterList.typeOfIndex(index);
       Type argType = argument.symbolType();
       if (!formalType.is(argType.fullyQualifiedName()) || formalType.isUnknown() || argType.isUnknown()) {
-        return false;
+        return new ArrayList<>();
```
m-g-sonar: I would prefer to use `Collections.emptyList()` here to have the returned collection unmodifiable by other methods.

java-checks/src/main/java/org/sonar/java/checks/MethodParametersOrderCheck.java
```diff
@@ -98,28 +99,32 @@ public static boolean allUnique(List<String> argListNames) {
...
       if (argumentList.indexOf(argument) != index) {
-        countArgumentsNotOrdered++;
+        misplacedParameters.add(formalParameterList.parameterAt(index));
       }
     }
-    return countArgumentsNotOrdered >= 2;
+    if (misplacedParameters.size() >= 2) {
+      return misplacedParameters;
+    }
+    return new ArrayList<>();
```
m-g-sonar: I would prefer to use `Collections.emptyList()` as well.

olcbean:
@m-g-sonar thanks for the review! I just pushed a commit addressing your comments. 

Can you let me know if you prefer the commits be squashed before you open the 'verify' PR.

m-g-sonar:
Thx for the update @olcbean . No need to squash everything, I'll do it from here when merging, and I prefer to have clear distinct commits history for review.

m-g-sonar:
Merged! Thanks for your contribution!

PR #2008, opened 2018-05-08
SONARJAVA-2625 Rule S4424: TrustManagers should not blindly accept any certificates

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/security/TrustManagerCertificateCheck.java
```diff
@@ -0,0 +1,112 @@
...
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
+      return;
+    }
+    MethodTree methodTree = (MethodTree) tree;
+    BlockTree blockTree = methodTree.block();
+    if (blockTree == null) {
+      return;
+    }
+    if (!CHECK_CLIENT_TRUSTED_MATCHER.matches(methodTree) && !CHECK_SERVER_TRUSTED_MATCHER.matches(methodTree)) {
```
andrei-epure-sonarsource: Think it would be easier to read without negations and return. 
```
if (CHECK_CLIENT_TRUSTED_MATCHER.matches(methodTree) || CHECK_SERVER_TRUSTED_MATCHER.matches(methodTree)) {
  if (!ThrowExceptionVisitor.throwsException(blockTree)) { 
    //...
  }
}
```

andrei-epure-sonarsource: Think it would be easier to read without negations and return. 
```
if (CHECK_CLIENT_TRUSTED_MATCHER.matches(methodTree) || CHECK_SERVER_TRUSTED_MATCHER.matches(methodTree)) {
  if (!ThrowExceptionVisitor.throwsException(blockTree)) { 
    //...
  }
}
```

java-checks/src/test/files/checks/security/TrustManagerCertificateCheck.java
```diff
@@ -0,0 +1,105 @@
...
+      @Override
+      public X509Certificate[] getAcceptedIssuers() {
+        return null;
+      }
+    };
+
+    trustManager = new X509TrustManager() {
+      @Override
+      public void checkClientTrusted(X509Certificate[] x509Certificates, String s) throws CertificateException {
+        throw new CertificateException();
```
andrei-epure-sonarsource: Would it be worth to make sure it throws a CertificateException and not other types of exceptions?
So to add a test where it throws say IllegalArgumentException and make it NonCompliant?

alban-auzeill: I was thinking about checking that the thrown exception is a CertificateException. But if you throw something else like:
throw new UnsupportedOperationException();
Then, there's no security issue.
An improvement could be to check that at least one of checkClientTrusted or checkServerTrusted throws CertificateException, but I think it does not worth the effort.

PR #2013, opened 2018-05-11
SONARJAVA-2744 S2226 should not raise issue for field annotated with @Resource

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/ServletInstanceFieldCheck.java
```diff
@@ -73,7 +79,7 @@ public void visitNode(Tree tree) {
 
   private static boolean isExcluded(VariableTree variable) {
     SymbolMetadata varMetadata = variable.symbol().metadata();
-    return isStaticOrFinal(variable) || varMetadata.isAnnotatedWith("javax.inject.Inject") || varMetadata.isAnnotatedWith("javax.ejb.EJB");
+    return isStaticOrFinal(variable) || ANNOTATIONS_EXCLUDING_FIELDS.stream().anyMatch(varMetadata::isAnnotatedWith);
```
andrei-epure-sonarsource: j'adore ce concept!

PR #2017, opened 2018-05-14
SONARJAVA-2676 Rule S4426: Cryptographic keys should not be too short

stylianos-agapiou-sonarsource:


java-checks/src/test/files/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,130 @@
+import java.security.KeyPairGenerator;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import javax.crypto.KeyGenerator;
+
+class A {
```
andrei-epure-sonarsource: nitpick: there's lots of tests here. I think that grouping the tests in 3 methods would increase readability 

- `public void nonCompliant()`
- `public void falseNegatives()`
- `public void compliant()`

Also, why do you need `class A` and `class B` ?

andrei-epure-sonarsource: for the false negatives, you can add a comment with the number of the ticket PY is working on (SONARJAVA-2491) and mention it can solve the problem

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,135 @@
...
+      if (expr.is(Tree.Kind.IDENTIFIER)) {
+        isValidKeyGenSymbol = getInstanceInfo.keyGenMethodSymbol.equals(((IdentifierTree) expr).symbol());
+      } else if (expr.is(Tree.Kind.MEMBER_SELECT)) {
+        isValidKeyGenSymbol = getInstanceInfo.keyGenMethodSymbol.equals(((MemberSelectExpressionTree) expr).identifier().symbol());
+      }
+    }
+    return isValidKeyGenSymbol;
+  }
+
+  private static boolean isGetInstanceMethod(MethodInvocationTree mit) {
```
andrei-epure-sonarsource: it's not ideal to have side effects in methods, even if private.
This function basically does 2 things:

- checks if it's a get instance method of the known algorithms
- sets the global minKey size

Based on the [Single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle), this method does two much.

I suggest that this method should just return the algorithm type and a pass that type to `checkAlgorithmParameterToReport`. Then the `checkAlgorithmParameterToReport` method can use a dictionary declared top-level that maps the algorithm type to the minKey size value. Doing this you can get rid of the global state inside `GetInstanceMethodInfo.minKeySizeValue`

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,135 @@
...
+  private static Optional<String> checkAlgorithmParameterToReport(ExpressionTree argument) {
+    String reportString = null;
+    if (argument.is(Tree.Kind.INT_LITERAL) && (Integer.parseInt(((LiteralTree) argument).value()) < getInstanceInfo.minKeySizeValue)) {
+      reportString = "Use a key length of at least " + getInstanceInfo.minKeySizeValue + " bits.";
+    }
+    getInstanceInfo.nullifyAttributes();
+    return Optional.ofNullable(reportString);
+  }
+
+  private static void storeSymbolOfAssignedInstance(MethodInvocationTree mit) {
```
andrei-epure-sonarsource: Why not returning a `Symbol` and rename the method `extractSymbolOfAssignedInstance`? This way you can get rid of the global state inside the `GetInstanceMethodInfo.keyGenMethodSymbol`

java-checks/src/test/files/checks/security/CryptographicKeySizeCheck.java
stylianos-agapiou-sonarsource: In the test files we don't use names for variables, functions and classes that would provide information to a reader.
From the comments: ` Compliant ` and ` Noncompliant ` someone can easily understand which test cases are supposed to raise an issue or not.

andrei-epure-sonarsource: I know what the Noncompliant, FN and Compliant are for. What is hard for me is to understand the overall picture, what's the difference between each test, what each test does differently. There are 21 tests here. Are there 21 different situations that this rule is tested for? If yes, I cannot see clearly what these differences are.

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
stylianos-agapiou-sonarsource: I agree with the single responsibility principle and that a dictionary could show clearer the relationship between the algorithm and the key size.

But regarding the algorithm, when ` getInstance ` function is found, I need to store the name of the algorithm(here I store the minKeySize) and the `Symbol` of the variable that the ` KeyPairGenetaror ` / ` KeyGenerator ` is assigned to, because the issue is raised when an `init`/`initialize` function is invoked.
When, `init`/`initialize` is invoked I need to check if the `Symbol` of the variable that calls the function is the same as the one stored and then according to the algorithm that I have found from `geInstance` and the argument of `init`/`initialize` raise an issue or not.
IMO I prefer having the Object `GetInstanceMethodInfo`, than having variables defined globally.

java-checks/src/test/files/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,130 @@
...
+    keyG = KeyPairGenerator.getInstance("RSA");
+    keyG.initialize(64);  // Noncompliant {{Use a key length of at least 2048 bits.}}
+  }
+
+  public void foo6() throws NoSuchAlgorithmException {
+    keyG = KeyPairGenerator.getInstance("RSA");
+    keyG.initialize(2048);  // Compliant
+  }
+
+  public void foo7() {
```
andrei-epure-sonarsource: foo7 and foo8 refer to illegalArgument. This is where naming helps. It's hard for a tired dev to see the one-digit difference between 1024 and 1028 :). Could better use 999 or something like this to point out the idea of the test

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,133 @@
...
+    if (GET_INSTANCE_METHOD.equals(mit.symbol().name())) {
+      ExpressionTree argument = mit.arguments().get(0);
+      if (argument.is(Tree.Kind.STRING_LITERAL)) {
+        String algorithm = LiteralUtils.trimQuotes(((LiteralTree) argument).value());
+        if (algorithmKeySizeMap.containsKey(algorithm)) {
+          storeAlgorithmKeySize(algorithm);
+          storeSymbolOfAssignedInstance(mit);
+        }
+      }
+    } else {
```
andrei-epure-sonarsource: nitpick: no need for 2 nested levels, can use  `} else if (isValidKeyGenSymbolOfInitializeMethod(mit)) {`

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
andrei-epure-sonarsource: as discussed offline, the rule needs the class-level state because must be invoked twice to find an issue: first for the getInstance() call, second for the initialization call

java-checks/src/test/files/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,124 @@
...
+
+  public void foo1() throws NoSuchAlgorithmException {
+    KeyGenerator keyGen = KeyGenerator.getInstance("Blowfish");
+    keyGen.init(64);  // Noncompliant [[sc=5;ec=20]] {{Use a key length of at least 128 bits.}}
+  }
+
+  public void foo2() throws NoSuchAlgorithmException {
+    KeyGenerator keyGen = KeyGenerator.getInstance("Blowfish");
+    int x = 64;
+    keyGen.init(x); // FN
```
andrei-epure-sonarsource: for the false negatives, you can add a comment with the number of the ticket PY is working on (SONARJAVA-2491) and mention it can solve the problem (if my understanding is correct)

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,133 @@
...
+      if (isValidKeyGenSymbolOfInitializeMethod(mit)) {
+        checkAlgorithmParameterToReport(mit.arguments().get(0)).ifPresent(reportString -> reportIssue(mit, reportString));
+      }
+    }
+  }
+
+  private static boolean isValidKeyGenSymbolOfInitializeMethod(MethodInvocationTree mit) {
```
andrei-epure-sonarsource: I find the name a little confusing because `a.keyPairG.initialize(1024);` is called on a valid symbol from a Java perspective, but not from the rule perspective...

maybe `initializeIsCalledOnPreviouslyDefinedKeyGen` or `keyGenSymbolIsInClassScope` (or something without _valid_ in it)?

java-checks/src/test/files/checks/security/CryptographicKeySizeCheck.java
m-g-sonar: Good point. Will make it explicit when solved that this is expected. And if no issue is raised when fixing the ticket, may raise some questions as well.

stylianos-agapiou-sonarsource: Actually, SONARJAVA-2491 will enable us to read only the values of constant variables.
So, it will not solve the problem and I suspect that we need Symbolic Execution to fix the FNs of this rule.

m-g-sonar: Cool, so it's an excellent reason to add a comment to be very explicit on why we are not detecting it, even if the case is trivial. For instance "requires following variables values with SE-based engine".

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,131 @@
...
+
+@Rule(key = "S4426")
+public class CryptographicKeySizeCheck extends AbstractMethodDetection {
+
+  private static final String KEY_PAIR_GENERATOR = "java.security.KeyPairGenerator";
+  private static final String KEY_GENERATOR = "javax.crypto.KeyGenerator";
+  private static final String GET_INSTANCE_METHOD = "getInstance";
+  private static final String STRING = "java.lang.String";
+
+  private static final GetInstanceMethodInfo getInstanceInfo = new GetInstanceMethodInfo();
```
pynicolas: This static field is modified. That makes the class thread hostile, which is not acceptable.

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
pynicolas: No, that doesn't require class-level state, it can be handled by instance-level state.

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,121 @@
...
+import org.sonar.plugins.java.api.tree.LambdaExpressionTree;
+import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.NewClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S4426")
+public class CryptographicKeySizeCheck extends IssuableSubscriptionVisitor {
+
+  private static final String KEY_PAIR_GENERATOR = "java.security.KeyPairGenerator";
```
andrei-epure-sonarsource: these constants should be inside the `MethodVisitor`

stylianos-agapiou-sonarsource: I am working on it still. I just uploaded some changes, but did not request for a review

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,151 @@
...
+        if ("getInstance".equals(methodName)) {
+          if (arg.is(Tree.Kind.STRING_LITERAL)) {
+            extractSymbolOfAssignedInstance(tree).ifPresent(keyGenSymbol -> {
+              algorithm = LiteralUtils.trimQuotes(((LiteralTree) arg).value());
+              geInstanceKeyGenSymbol = keyGenSymbol;
+            });
+          }
+        } else if (isInitMethod(methodName) && initializeIsCalledOnPreviouslyDefinedKeyGen(tree)) {
+          Integer minKeySize = findMinimumKeySizeFromAlgorithm(algorithm);
+          if (minKeySize != -1) {
```
pynicolas: The approach requires that getInstance is called only once in a method. Otherwise, we get some false negatives: that can be fine but would deserve some test cases to highlight the limits of the approach.

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    private boolean initializeIsCalledOnPreviouslyDefinedKeyGen(MethodInvocationTree mit) {
+      boolean isValidKeyGenSymbol = false;
+      if (geInstanceKeyGenSymbol != null) {
+        ExpressionTree expr = ((MemberSelectExpressionTree) mit.methodSelect()).expression();
+        if (expr.is(Tree.Kind.IDENTIFIER)) {
+          isValidKeyGenSymbol = geInstanceKeyGenSymbol.equals(((IdentifierTree) expr).symbol());
+        } else if (expr.is(Tree.Kind.MEMBER_SELECT)) {
+          isValidKeyGenSymbol = geInstanceKeyGenSymbol.equals(((MemberSelectExpressionTree) expr).identifier().symbol());
```
pynicolas: This logic duplicates a part of the previous method: in both cases you to get a `Symbol` from an `ExpressionTree`.

java-checks/src/test/files/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  public void foo10() throws NoSuchAlgorithmException {
+    KeyGenerator keyGen = KeyGenerator.getInstance("AES");
+    KeyGenerator keyGen2 = KeyGenerator.getInstance("Blowfish");
+    keyGen.init(64); // Compliant
+  }
+}
+
+class B {
```
pynicolas: What's the difference between class A and class B? Can we find better names? There are a lot of test cases, probably too many, and it's really hard to follow how each one is specific.

java-checks/src/test/files/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  public void foo3() throws NoSuchAlgorithmException {
+    String algorithm = "Blowfish";
+    KeyGenerator keyGen = KeyGenerator.getInstance(algorithm);
+    keyGen.init(64); // FN requires following variable's values with SE-based engine
+    keyGen.getAlgorithm();
+  }
+
+  public void foo4() throws NoSuchAlgorithmException {
```
pynicolas: The names of the methods could be used to indicate what's unique to each test case: for example, this method could be named "key_assignment".

java-checks/src/test/files/checks/security/CryptographicKeySizeCheck.java
andrei-epure-sonarsource: one thing I have seen with other rules is splitting test cases in multiple files.
Here class B uses the inner variable declared in class A. Class B could be moved in a different file, with a smaller class A (that wouldn't contain test cases, but only the inner variable)

java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -0,0 +1,115 @@
...
+    return (MethodTree) tree;
+  }
+
+  private static class MethodVisitor extends BaseTreeVisitor {
+
+    private String algorithm = null;
+    private String reportString = null;
+    private MethodInvocationTree treeToReport = null;
+
+    public MethodVisitor(ExpressionTree getInstanceArg) {
```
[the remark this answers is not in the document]
stylianos-agapiou-sonarsource: I don't think it should call `super()` explicitly.

java-checks/src/test/files/checks/security/CryptographicKeySizeCheck/RSA.java
```diff
@@ -0,0 +1,74 @@
+import java.security.KeyPairGenerator;
+import java.security.NoSuchAlgorithmException;
+import javax.crypto.KeyGenerator;
+
+class RSA {
```
pynicolas: I think that there are too many tests here which duplicate what's tested for Blowfish.

stylianos-agapiou-sonarsource: You are right no need to keep everything.
I just left a Compliant and a Noncompliant Case for RSA.

PR #2019, opened 2018-05-14
SONARJAVA-2491 Read constant values from bytecode

pynicolas:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/helpers/ConstantUtils.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/helpers/ConstantUtils.java
@@ -1,0 +1,72 @@
+/*
+ * SonarQube Java
+ * Copyright (C) 2012-2018 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java.checks.helpers;
+
+import javax.annotation.CheckForNull;
+import org.sonar.java.model.LiteralUtils;
+import org.sonar.java.resolve.JavaSymbol;
+import org.sonar.plugins.java.api.semantic.Symbol;
+import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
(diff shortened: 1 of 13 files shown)


java-frontend/src/main/java/org/sonar/java/resolve/BytecodeVisitor.java
```diff
@@ -198,8 +199,12 @@ public FieldVisitor visitField(int flags, String name, String desc, @Nullable St
     Preconditions.checkNotNull(desc);
     if (isNotSynthetic(flags)) {
       //Flags from asm lib are defined in Opcodes class and map to flags defined in Flags class
-      final JavaSymbol.VariableJavaSymbol symbol = new JavaSymbol.VariableJavaSymbol(Flags.filterAccessBytecodeFlags(flags),
-          name, convertAsmType(org.objectweb.asm.Type.getType(desc)), classSymbol);
+      int filteredFlags = Flags.filterAccessBytecodeFlags(flags);
+      JavaType type = convertAsmType(Type.getType(desc));
+      JavaSymbol.VariableJavaSymbol symbol = new JavaSymbol.VariableJavaSymbol(filteredFlags, name, type, classSymbol);
```
benzonico: it should be dropped and GC anyway, but I would avoid creating a variable if it is of no use.

java-frontend/src/main/java/org/sonar/java/resolve/BytecodeCompleter.java
```diff
@@ -83,6 +89,35 @@ public void complete(JavaSymbol symbol) {
     }
   }
 
+  @CheckForNull
+  public Object constantValue(Symbol owner, String constantName) {
```
benzonico: I think this method deserves quite some documentation to mention the fact that it is intended to be called from semantic resolution from source and that is why we need to read class bytes.

java-checks/src/main/java/org/sonar/java/checks/helpers/ConstantUtils.java
```diff
@@ -0,0 +1,72 @@
...
+      expression = ((ParenthesizedTree) expression).expression();
+    }
+    if (expression.is(Tree.Kind.MEMBER_SELECT)) {
+      expression = ((MemberSelectExpressionTree) expression).identifier();
+    }
+    if (expression.is(Tree.Kind.IDENTIFIER)) {
+      IdentifierTree id = (IdentifierTree) expression;
+      Symbol symbol = id.symbol();
+      if (symbol instanceof JavaSymbol.ConstantJavaSymbol) {
+        Object constantValue = ((JavaSymbol.ConstantJavaSymbol) symbol).value();
```
benzonico: For this I would probably have the value() method returning an optional on variable symbol : always empty on varibable and wrapping a value on constants.

java-frontend/src/main/java/org/sonar/java/resolve/BytecodeVisitor.java
pynicolas: Fixed in f7ea69796727befbad0ae1ec1e09982ec507b652

java-frontend/src/main/java/org/sonar/java/resolve/BytecodeCompleter.java
pynicolas: Added some documentation in f7ea69796727befbad0ae1ec1e09982ec507b652

java-checks/src/main/java/org/sonar/java/checks/helpers/ConstantUtils.java
pynicolas: Changed in f7ea69796727befbad0ae1ec1e09982ec507b652

PR #2020, opened 2018-05-15
SONARJAVA-2743 FN on S3688 when disallowed class is used in import declaration or method invocation

alban-auzeill:


java-checks/src/test/files/checks/DisallowedClassCheck.java
```diff
@@ -1,3 +1,9 @@
+import unknown;
+import java.lang.Object;
+import java.lang.String; // Noncompliant
+import java.lang.String.*; // Noncompliant
+import static java.lang.String.valueOf; // Noncompliant
+
 class A {
   private String disallowed; // Noncompliant [[sc=11;ec=17]]
   private Integer disallowed2;
```
andrei-epure-sonarsource: now that you are editing the file, could you please rename this `disallowed2` Integer to something else? because the Integer class is not disallowed for this test, only  ` visitor.disallowedClass = "java.lang.String";`

java-checks/src/test/files/checks/DisallowedClassCheck.java
```diff
@@ -1,3 +1,9 @@
+import unknown;
```
andrei-epure-sonarsource: cannot comment unchanged line inside github, but could you please add a comment why line 36 is Compliant? ... is has String as well and it's not intuitive for me
`((java.util.Optional<String>) o).map(value -> 1);`

java-checks/src/main/java/org/sonar/java/checks/DisallowedClassCheck.java
```diff
@@ -23,11 +23,15 @@
 import org.sonar.check.Rule;
```
andrei-epure-sonarsource: why didn't you add some tests in DisallowedClassCheckRegex as well?

java-checks/src/test/files/checks/DisallowedClassCheck.java
alban-auzeill: I added the comment: // Compliant, ignore inferred type
This explain why "value" is compliant, but I didn't challenge why type cast is compliant.

PR #2022, opened 2018-05-16
SONARJAVA-2735 S1604 FP when instance methods of the anonymous class are invoked

alban-auzeill:


java-checks/src/test/files/checks/AnonymousClassShouldBeLambdaCheck.java
```diff
@@ -251,3 +251,63 @@ void foo() {
...
+  }
+
+  interface InvokeStatic {
+
+    int func();
+
+    static int staticFunc() {
+      InvokeStatic f = new InvokeStatic() {  // Noncompliant
+        @Override
+        public String func(int n) {
```
andrei-epure-sonarsource: int func();

java-checks/src/test/files/checks/AnonymousClassShouldBeLambdaCheck.java
```diff
@@ -251,3 +251,63 @@ void foo() {
...
+  }
+
+  interface Math {
+    int powerOfTwo(int n);
+  }
+
+  void testRecursion() {
+    Math f = new Math() {  // Compliant, recursion
+      @Override
+      public String powerOfTwo(int n) {
```
andrei-epure-sonarsource: int powerOfTwo(int n);

PR #2024, opened 2018-05-17
SONARJAVA-2672 Rule S3329: Cypher Block Chaining IV's should be random and unique

stylianos-agapiou-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,98 @@
...
+    Symbol symbolArgument = null;
+    if (argument.is(Tree.Kind.IDENTIFIER)) {
+      symbolArgument = ((IdentifierTree) argument).symbol();
+    } else if (argument.is(Tree.Kind.MEMBER_SELECT)) {
+      symbolArgument = ((MemberSelectExpressionTree) argument).identifier().symbol();
+    }
+    return symbolArgument;
+  }
+
+  private static Tree findParentMethod(Tree tree) {
```
pynicolas: This method could instead return a `MethodTree` or `null`. And it could be named `findEnclosingMethod`.

java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,98 @@
...
+    while (!((tree).is(Tree.Kind.METHOD)) && !tree.is(Tree.Kind.CLASS)) {
```
pynicolas: Why parentheses around `(tree)` and `((tree).is(Tree.Kind.METHOD)`?

java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,98 @@
...
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return ImmutableList.of(
+      MethodMatcher.create().typeDefinition("javax.crypto.spec.IvParameterSpec").name("<init>").withAnyParameters());
+  }
+
+  @Override
+  protected void onConstructorFound(NewClassTree newClassTree) {
+    Tree mTree = findParentMethod(newClassTree);
+    if (mTree.is(Tree.Kind.METHOD)) {
+      Optional<StatementTree> secureRandomOptional = findSecureRandomInMethod((MethodTree) mTree, returnSymbolOfArgument(newClassTree.arguments().get(0)));
```
pynicolas: It seems we don't care about the StatementTree, only about its existence. `findSecureRandomInMethod` could therefore return a boolean.

java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,98 @@
...
+      if (!secureRandomOptional.isPresent()) {
+        reportIssue(newClassTree, "Use a dynamically-generated, random IV.");
+      }
+    }
+  }
+
+  private static Optional<StatementTree> findSecureRandomInMethod(MethodTree methodTree,
+                                                                  @javax.annotation.Nullable Symbol argumentSymbol) {
```
pynicolas: `argumentSymbol` is a strange name here. In the context of this method, this parameter represents the symbol of the "iv" ("initialization vector", see the doc for `IvParameterSpec`).

java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,98 @@
...
+    if (methodTree.block() != null && argumentSymbol != null) {
+      return methodTree.block().body().stream().filter(statementTree -> {
```
pynicolas: That way of looking for method invocations works only in the simplest cases. Example of FP:
```
  void foo() {
    if (true) {
      byte[] bytes = "111".getBytes("UTF-8");
      random.nextBytes(bytes);
      IvParameterSpec iv = new IvParameterSpec(bytes);
    }
  }
```

java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private static final MethodMatcher SECURE_RANDOM_NEXT_BYTES = MethodMatcher.create()
+      .typeDefinition("java.security.SecureRandom")
+      .name("nextBytes")
+      .withAnyParameters();
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree methodInvocation) {
+      if (SECURE_RANDOM_NEXT_BYTES.matches(methodInvocation)) {
+        secureRandomFound = true;
```
pynicolas: For this rule, I think it would be better to also check the symbol (the variable containing bytes), as I described in the comment on [SONARJAVA-2672](https://jira.sonarsource.com/browse/SONARJAVA-2672).

java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,81 @@
...
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.NewClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S3329")
+public class CipherBlockChainingCheck extends AbstractMethodDetection {
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return ImmutableList.of(
```
pynicolas: Let's use `Collections.singletonList()` to avoid adding a dependency on guava.

java-checks/src/test/files/checks/security/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,89 @@
...
+
+  SecureRandom random = new SecureRandom();
+  private byte[] bytes2 = "111".getBytes();
+  {
+    random.nextBytes(bytes2);
+  }
+  IvParameterSpec iv = new IvParameterSpec(bytes2); // Compliant FP, will be fixed with SE engine
+
+  void foo1()
+    throws InvalidKeyException, InvalidAlgorithmParameterException, NoSuchAlgorithmException, NoSuchPaddingException, UnsupportedEncodingException, IllegalBlockSizeException,
```
pynicolas: What does this list of exceptions bring to the test? Can't we replace it with `throws Exception`?

java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
stylianos-agapiou-sonarsource: Ok, my misunderstanding. I will use the code from the previous commit, the function `returnSymbolOfInitializerVectorBuffer`

PR #2026, opened 2018-05-22
SONARJAVA-2680 Rule S4435: XML transformers should be secured

pynicolas:


java-checks/src/test/files/checks/security/SecureXmlTransformerCheck.java
```diff
@@ -0,0 +1,88 @@
...
+    return factory;
+  }
+
+  TransformerFactory secure_processing_true_with_literal() {
+    TransformerFactory factory = TransformerFactory.newInstance();
+    factory.setFeature("http://javax.xml.XMLConstants/feature/secure-processing", true);
+    return factory;
+  }
+
+  TransformerFactory secure_processing_true() {
```
andrei-epure-sonarsource: `setattribute_dtd_only` - actually this duplicates line 63 ... no, this is setFeature, so should be removed?

java-checks/src/test/files/checks/security/SecureXmlTransformerCheck.java
```diff
@@ -0,0 +1,88 @@
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.XMLConstants;
+
+class A {
```
andrei-epure-sonarsource: would it be useful to add some `This implementation will not check` FN cases, just to be clear from the tests what the rule does not cover?

java-checks/src/test/files/checks/security/SecureXmlTransformerCheck.java
```diff
@@ -0,0 +1,88 @@
...
+  TransformerFactory no_call_to_securing_method() {
+    TransformerFactory factory = TransformerFactory.newInstance(); // Noncompliant
+    return factory;
+  }
+
+  // setFeature
+
+  TransformerFactory secure_processing_true() {
+    TransformerFactory factory = TransformerFactory.newInstance();
+    factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
```
andrei-epure-sonarsource: would be useful to have a test where `true` is in a constant

java-checks/src/main/java/org/sonar/java/checks/security/SecureXmlTransformerCheck.java
```diff
@@ -0,0 +1,128 @@
...
+      if (SET_FEATURE.matches(methodInvocation)
+        && XMLConstants.FEATURE_SECURE_PROCESSING.equals(resolveAsStringConstant(arguments.get(0)))
+        && LiteralUtils.isTrue(arguments.get(1))) {
+
+        hasSecureProcessingFeature = true;
+      }
+
+      if (SET_ATTRIBUTE.matches(methodInvocation)) {
+        String attributeName = resolveAsStringConstant(arguments.get(0));
+        String attributeValue = resolveAsStringConstant(arguments.get(1));
```
andrei-epure-sonarsource: could also use `JavaPropertiesHelper.retrievedPropertyDefaultValue` ?

java-checks/src/test/files/checks/security/SecureXmlTransformerCheck.java
pynicolas: Why should it be removed? I'm going to change the name of the method which is not helpful at all.

java-checks/src/main/java/org/sonar/java/checks/security/SecureXmlTransformerCheck.java
pynicolas: I don't know which behavior we would want for this rule when properties are used and I don't think it would add value to the implementation.

java-checks/src/test/files/checks/security/SecureXmlTransformerCheck.java
pynicolas: Added in eb6a439ed.

java-checks/src/test/files/checks/security/SecureXmlTransformerCheck.java
pynicolas: Added in eb6a439ed

PR #2027, opened 2018-05-22
FP on S1118 with Lombok constructors

bannmann:
This fixes false positives of S1118 ("Utility classes should not have public constructors") regarding lombok-generated constructors; see my post at https://groups.google.com/forum/#!topic/sonarqube/hu2sITqPyQs

    - **_My local ruling test failed, but had exactly the same results as with `master`._**
    - **_There was no reaction to my post, so there is no Jira ticket._**

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
+++ b/java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
@@ -36,7 +36,6 @@ import org.sonar.plugins.java.api.tree.Tree;
 
 import javax.annotation.Nullable;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 
 public class LombokFilter extends BaseTreeVisitorIssueFilter {
@@ -122,41 +121,30 @@ public class LombokFilter extends BaseTreeVisitorIssueFilter {
     return false;
   }
 
-  private boolean generatesPrivateConstructor(ClassTree classTree) {
+  private static boolean generatesPrivateConstructor(ClassTree classTree) {
     if (usesAnnotation(classTree, GENERATE_PRIVATE_CONSTRUCTOR)) {
       return true;
     }
     SymbolMetadata metadata = classTree.symbol().metadata();
-    for (String annotation : GENERATE_CONSTRUCTOR) {
-      if (metadata.isAnnotatedWith(annotation) && generatesPrivateAccess(metadata, annotation)) {
-        return true;
-      }
-    }
-    return false;
+    return GENERATE_CONSTRUCTOR.stream().anyMatch(annotation -> generatesPrivateAccess(metadata, annotation));
   }
 
   private static boolean generatesPrivateAccess(SymbolMetadata metadata, String annotation) {
-    for (SymbolMetadata.AnnotationValue annotationValue : metadata.valuesForAnnotation(annotation)) {
-      if (Objects.equals(annotationValue.name(), "access") && Objects.equals(getAccessLevelValue(annotationValue.value()), "PRIVATE")) {
-        return true;
-      }
-    }
-    return false;
+    return metadata.isAnnotatedWith(annotation) && metadata.valuesForAnnotation(annotation).stream().anyMatch(av -> "access".equals(av.name()) && "PRIVATE".equals(getAccessLevelValue(av.value())));
   }
 
   @Nullable
   private static String getAccessLevelValue(Object value) {
-    String retentionValue = null;
     if (value instanceof Tree) {
       Tree tree = (Tree) value;
       if (tree.is(Tree.Kind.MEMBER_SELECT)) {
-        retentionValue = ((MemberSelectExpressionTree) tree).identifier().name();
+        return ((MemberSelectExpressionTree) tree).identifier().name();
       } else if (tree.is(Tree.Kind.IDENTIFIER)) {
-        retentionValue = ((IdentifierTree) tree).name();
+        return ((IdentifierTree) tree).name();
       }
     } else if (value instanceof Symbol.VariableSymbol) {
-      retentionValue = ((Symbol.VariableSymbol) value).name();
+      return ((Symbol.VariableSymbol) value).name();
     }
-    return retentionValue;
+    return null;
   }
 }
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -116,4 +121,42 @@ private static boolean usesAnnotation(ClassTree classTree, List<String> annotati
...
+    for (String annotation : GENERATE_CONSTRUCTOR) {
+      if (metadata.isAnnotatedWith(annotation) && generatesPrivateAccess(metadata, annotation)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private static boolean generatesPrivateAccess(SymbolMetadata metadata, String annotation) {
+    for (SymbolMetadata.AnnotationValue annotationValue : metadata.valuesForAnnotation(annotation)) {
```
m-g-sonar: You can use a Stream directly here:
```
return metadata.valuesForAnnotation(annotation).anyMatch(av -> ... );
```

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -116,4 +121,42 @@ private static boolean usesAnnotation(ClassTree classTree, List<String> annotati
...
+      if (Objects.equals(annotationValue.name(), "access") && Objects.equals(getAccessLevelValue(annotationValue.value()), "PRIVATE")) {
```
m-g-sonar: I think we could simplify this a bit by removing usage of `Objects`:
```
if ("access".equals(annotationValue.name())) && "PRIVATE".equals(getAccessLevelValue(annotationValue.value())) {
```

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -116,4 +121,42 @@ private static boolean usesAnnotation(ClassTree classTree, List<String> annotati
     }
     return false;
   }
+
+  private boolean generatesPrivateConstructor(ClassTree classTree) {
+    if (usesAnnotation(classTree, GENERATE_PRIVATE_CONSTRUCTOR)) {
+      return true;
+    }
+    SymbolMetadata metadata = classTree.symbol().metadata();
+    for (String annotation : GENERATE_CONSTRUCTOR) {
```
m-g-sonar: You can replace this part by :
```
return GENERATE_CONSTRUCTOR.stream().anyMatch(metadata -> metadata.isAnnotatedWith(annotation) && generatesPrivateAccess(metadata, annotation));
```

I would also move the `isAnnotatedWith(annotation)`directly in the `gernatePrivateAccess` method, so it makes the stream simpler.

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -116,4 +121,42 @@ private static boolean usesAnnotation(ClassTree classTree, List<String> annotati
...
+      if (Objects.equals(annotationValue.name(), "access") && Objects.equals(getAccessLevelValue(annotationValue.value()), "PRIVATE")) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  @Nullable
+  private static String getAccessLevelValue(Object value) {
+    String retentionValue = null;
```
m-g-sonar: This is not a `retentionValue`... that's the Access level. Rename the variable or simply drop it to return immediately the value when getting it.

`RetentionValue` is variable name from `java.lang.annotation.Retention` (copy-pasted from ReflectionOnNonRuntimeAnnotationCheck` ?).

m-g-sonar:
Thanks for the update @bannmann. Now, everything looks good to me regarding your contribution, good job!

It's perfect to me to keep commits separated during review phase. It's indeed when opening a PR that we usually appreciate to have everything squashed together (if it makes sense).

In a few minutes I'm going to open a new PR with your changes, which will trigger our internal QA. If everything goes green, I'll then close mine and merge your PR. I'll let you know if there is anything to change according to QA (no ideas why ITs were not working on your machine).

Cheers,
Michael

m-g-sonar:
Okay, 2 issues spotted by SonarQube analysis: https://next.sonarqube.com/sonarqube/project/issues?id=org.sonarsource.java%3Ajava&pullRequest=2052&resolved=false

Sorry I missed that. Anyway, from my point of view, the NPE can not occur, because you are explicitly testing if `isAnnotatedWith(...)` before trying to access annotations values... Nevertheless, it made me realize we can be a bit more efficient/explicit.

Because `valuesForAnnotation()` is actually returning null if not annotated, I would rewrite the following
```java
private static boolean generatesPrivateAccess(SymbolMetadata metadata, String annotation) {
  return metadata.isAnnotatedWith(annotation) && metadata.valuesForAnnotation(annotation).stream().anyMatch(av -> "access".equals(av.name()) && "PRIVATE".equals(getAccessLevelValue(av.value())));
}
```
to :
```java
private static boolean generatesPrivateAccess(SymbolMetadata metadata, String annotation) {
  List<AnnotationValue> values = metadata.valuesForAnnotation(annotation);
  return values != null && values.stream().anyMatch(av -> "access".equals(av.name()) && "PRIVATE".equals(getAccessLevelValue(av.value())));
}
```

And now,  we could simplify a bit more by also changing the caller method to:
```java
private static boolean generatesPrivateConstructor(ClassTree classTree) {
  if (usesAnnotation(classTree, GENERATE_PRIVATE_CONSTRUCTOR)) {
    return true;
  }
  SymbolMetadata metadata = classTree.symbol().metadata();
  return GENERATE_CONSTRUCTOR.stream()
    .map(metadata::valuesForAnnotation)
    .filter(Objects::nonNull)
    .anyMatch(LombokFilter::generatesPrivateAccess);
}

private static boolean generatesPrivateAccess(List<AnnotationValue> values) {
  return values.stream().anyMatch(av -> "access".equals(av.name()) && "PRIVATE".equals(getAccessLevelValue(av.value())));
}
```

I let you choose which approach you prefer, between the null check or the full rewriting, and I'll update the duplicate PR once updated.

Cheers,
Michael

bannmann:
I went with the rewrite. The null safety created by the `isAnnotatedWith()` check is way too subtle for my taste, too, and keeping it plus adding a redundant null check to satisfy SonarQube is just clumsy.

By the way, I also played with using `flatMap()` instead of `map()` so that the `generatesPrivateAccess()` would not need to operate on `List`, but in the end that did make the code clearer or more concise in any way.

Added commit and rebased on new master.

m-g-sonar:
Hey @bannmann ,

Thanks for the update and for your contribution. Merged into master!

The fix will be part of release 5.5 of SonarJava, to be delivered in 2-3 weeks.

Cheers,
Michael

PR #2034, opened 2018-05-23
SONARSEC-62 Generate constructor calls regarding strings in ucfg

benzonico:


java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -224,6 +227,9 @@ private void buildCall(Tree element, UCFGBuilder.BlockBuilder blockBuilder, Iden
     if (element.is(METHOD_INVOCATION)) {
       MethodInvocationTree methodInvocationTree = (MethodInvocationTree) element;
       buildMethodInvocation(blockBuilder, idGenerator, methodInvocationTree);
+    } else if (element.is(NEW_CLASS)) {
```
m-g-sonar: What about method references on constructors, such as `MyStringsList.map(MySink::new).forEach(...)` ?
(I know this is also probably not handled for simple method references)

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -248,6 +254,21 @@ private void buildCall(Tree element, UCFGBuilder.BlockBuilder blockBuilder, Iden
     }
   }
 
+  private void buildConstructorInvocation(BlockBuilder blockBuilder, IdentifierGenerator idGenerator, NewClassTree newClassTree) {
+    List<Expression> arguments = null;
+    if(isString(newClassTree.constructorSymbol().owner().type()) || newClassTree.arguments().stream().map(ExpressionTree::symbolType).anyMatch(UCFGJavaVisitor::isString)) {
+      arguments = argumentIds(idGenerator, newClassTree.arguments());
+    }
+
+    if (arguments != null) {
```
m-g-sonar: I think we can refactor these part in order to share the logic for method invocations and constructors call.

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -224,6 +227,9 @@ private void buildCall(Tree element, UCFGBuilder.BlockBuilder blockBuilder, Iden
     if (element.is(METHOD_INVOCATION)) {
       MethodInvocationTree methodInvocationTree = (MethodInvocationTree) element;
       buildMethodInvocation(blockBuilder, idGenerator, methodInvocationTree);
+    } else if (element.is(NEW_CLASS)) {
+      NewClassTree newClassTree = (NewClassTree) element;
+      buildConstructorInvocation(blockBuilder, idGenerator, newClassTree);
```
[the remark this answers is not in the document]
benzonico: And you could imagine the opposite case. As per soundness I don't think that anonymous classes should be handled any differently than other constructor calls.

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
benzonico: As we don't handle collections I don't think this is relevant and linked to this PR.

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -248,6 +254,21 @@ private void buildCall(Tree element, UCFGBuilder.BlockBuilder blockBuilder, Iden
...
+  private void buildConstructorInvocation(BlockBuilder blockBuilder, IdentifierGenerator idGenerator, NewClassTree newClassTree) {
+    List<Expression> arguments = null;
+    if(isString(newClassTree.constructorSymbol().owner().type()) || newClassTree.arguments().stream().map(ExpressionTree::symbolType).anyMatch(UCFGJavaVisitor::isString)) {
+      arguments = argumentIds(idGenerator, newClassTree.arguments());
+    }
+
+    if (arguments != null) {
+      String destination = idGenerator.newIdFor(newClassTree);
+      blockBuilder.assignTo(variableWithId(destination),
+        UCFGBuilder.call(signatureFor((Symbol.MethodSymbol) newClassTree.constructorSymbol()))
```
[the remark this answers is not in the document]
benzonico: If constructor is not resolved we will have a classCastException : i'm fixing this. Thanks.

PR #2036, opened 2018-05-24
SONARJAVA-2718 Method from parent interface accessed using 'I.super.myMethod()' not resolved

alban-auzeill:


java-frontend/src/main/java/org/sonar/java/resolve/SecondPass.java
```diff
@@ -109,6 +109,12 @@ private void complete(JavaSymbol.TypeJavaSymbol symbol) {
 
     if ((symbol.flags() & Flags.INTERFACE) == 0) {
       symbol.members.enter(new JavaSymbol.VariableJavaSymbol(Flags.FINAL, "super", type.supertype, symbol));
+    } else {
+      // JLS9 - 15.12.1 : To support invocation of default methods in superinterfaces, the TypeName may also refer
```
m-g-sonar: Wording is not really correct, and a bit ambiguous. `TypeName` is not defined. I would go for

> Used in form 'TypeName.super.foo()', where 'TypeName' is an interface. To support invocation of default methods **from** super-interfaces, 'TypeName' may also refer to a direct super-interface of the current class or interface. **The method being invoked ('foo()') has to be searched in that super-interface**.

java-frontend/src/main/java/org/sonar/java/resolve/SecondPass.java
```diff
@@ -109,6 +109,12 @@ private void complete(JavaSymbol.TypeJavaSymbol symbol) {
...
+      // to a direct superinterface of the current class or interface, and the target is that superinterface.
+      symbol.members.enter(new JavaSymbol.VariableJavaSymbol(Flags.FINAL, "super", type, symbol));
+      // Note: The above "super" symbol will always be qualified when referenced. e.g. A.super.hashCode()
```
m-g-sonar: mentioning the super-interfaces explicitly can also come from the fact that an interface/class can extend/implement multiple interfaces. Having it mentioned explicitly remove any ambiguity.

java-checks/src/test/files/checks/AssertionsInTestsCheck/Junit4.java
```diff
@@ -51,7 +55,7 @@ public void junit_assert_true() {
 
   @Test
   public void junit_assert_that() {
-    org.junit.Assert.assertThat("aaa", org.junit.matchers.JUnitMatchers.containsString("a"));
+    assertThat("aaa", org.junit.matchers.JUnitMatchers.containsString("a"));
```
m-g-sonar: Do not change that. Keep it explicit to show that it's a junit assert.

java-checks/src/test/files/checks/AssertionsInTestsCheck/Junit4.java
```diff
@@ -88,7 +92,7 @@ public void junit_rule_expected_exception_message() {
 
   @Test
   public void junit_rule_error_collector() { // Compliant
-    errorCollector.checkThat("123", org.hamcrest.CoreMatchers.equalTo("123"));
+    errorCollector.checkThat("123", equalTo("123"));
```
m-g-sonar: unwanted change.

java-checks/src/test/files/checks/AssertionsInTestsCheck/Junit4.java
```diff
@@ -144,4 +148,18 @@ public void test_fail() {
...
+  interface I {
+    default String returnFoo() {
+      return "Foo";
+    }
+  }
+
+  public class ITest implements I {
+    @Test
+    public void testSuper() {
+      assertThat(I.super.returnFoo(), equalTo("Foo"));
```
m-g-sonar: Please use fully qualified names of these static methods if you want to keep this new case. 

However, I'm not sure this is wanted. You have a dedicated unit test related to the fixed semantic. So this is just an extra case which have no reasons NOT to be compliant. It was non-compliant because of a bug in the semantic... which is somehow not directly related to this rule. Some other rule could have been impacted, and we don't want to add test cases for all of them.

java-checks/src/test/files/checks/AssertionsInTestsCheck/Junit4.java
alban-auzeill: I removed all changes on AssertionsInTestsCheck/Junit4.java

PR #2037, opened 2018-05-24
SONARJAVA-2766 FN S2142: no issue are raised when using union type

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
```diff
@@ -64,17 +70,31 @@ public void visitNode(Tree tree) {
...
         if(!blockVisitor.threadInterrupted && !isWithinInterruptingFinally()) {
-          reportIssue(catchTree.parameter(), "Either re-interrupt this method or rethrow the \""+catchType.name()+"\".");
+          reportIssue(catchTree.parameter(), "Either re-interrupt this method" +
+            " or rethrow the \""+interruptType.get().name()+"\".");
         }
       }
     }
   }
 
+  private static Optional<Type> findInterruptingType(VariableTree parameter) {
```
benzonico: No need to use optional here, just use anyMatch and return a boolean

alban-auzeill: The method return the type and not a boolean to be able to add the type name in the issue message (line 79).

PR #2039, opened 2018-05-24
SONARJAVA-2681 Rule S4499: SMTP SSL connection should check server identity

stylianos-agapiou-sonarsource:


java-checks/pom.xml
```diff
@@ -134,12 +134,24 @@
                   <version>4.3.7.RELEASE</version>
                   <type>jar</type>
                 </artifactItem>
+                <artifactItem>
+                    <groupId>org.apache.commons</groupId>
+                    <artifactId>commons-email</artifactId>
+                    <version>1.3.2</version>
+                    <type>jar</type>
+                </artifactItem>
```
pynicolas: Can we fix the indentation to be consistent with the rest of the file?

java-checks/pom.xml
```diff
@@ -134,12 +134,24 @@
...
                   <artifactId>spring-beans</artifactId>
                   <version>4.3.7.RELEASE</version>
                   <type>jar</type>
                 </artifactItem>
+                <artifactItem>
+                  <groupId>javax.mail</groupId>
+                  <artifactId>mail</artifactId>
+                  <version>1.4.7</version>
+                  <type>jar</type>
+                </artifactItem>
```
pynicolas: Why do we need this dependency?

java-checks/src/main/java/org/sonar/java/checks/security/SMTPSSLServerIdentityCheck.java
```diff
@@ -0,0 +1,150 @@
...
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S4499")
+public class SMTPSSLServerIdentityCheck extends AbstractMethodDetection {
+
+  private static final String EMAIL_CLASS_NAME = "org.apache.commons.mail.Email";
```
pynicolas: This constant name should contain either "APACHE" or "COMMONS".

java-checks/src/main/java/org/sonar/java/checks/security/SMTPSSLServerIdentityCheck.java
```diff
@@ -0,0 +1,150 @@
...
+      .typeDefinition(TypeCriteria.is(EMAIL_CLASS_NAME))
+      .name("setSSLOnConnect")
+      .addParameter(BOOLEAN);
+
+    private static final MethodMatcher SET_SSL_CHECK_SERVER_ID = MethodMatcher.create()
+      .typeDefinition(EMAIL_CLASS_NAME)
+      .name("setSSLCheckServerIdentity")
+      .addParameter(BOOLEAN);
+
+    private static final MethodMatcher MAIL_SMTP_SSL = MethodMatcher.create()
```
pynicolas: Why this name?

java-checks/src/main/java/org/sonar/java/checks/security/SMTPSSLServerIdentityCheck.java
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  private static class MethodBodyVisitor extends BaseTreeVisitor {
+
+    private Integer bothFunctionsAreChecked = 0;
+    /*
+     * MethodVisitor constructor's parameter is 0 when "send()" from org.apache.commons.mail.Email is invoked
+     * and 1 for "getDefaultInstance" from javax.mail.Session
+     */
+    private final Integer differentiateEmailAndSessionInvocations;
```
pynicolas: This field is used to trigger 2 different behaviors. We should have 2 different classes instead.

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/JavaMailSession.java
```diff
@@ -0,0 +1,58 @@
...
+        return new PasswordAuthentication("username@gmail.com", "password");
+      }
+    });
+  }
+
+  public void false_positive(boolean flag) {
+    Properties props = new Properties();
+    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory");
+    props.put("mail.smtp.ssl.checkserveridentity", flag);
+    Session s = Session.getDefaultInstance(props, new javax.mail.Authenticator() { // Noncompliant {{Enable server identity validation, set "mail.smtp.ssl.checkserveridentity" to true}}
```
pynicolas: If we consider this as a FP, then can't we change the implementation to not raise an issue when `mail.smtp.ssl.checkserveridentity` is set to an unknown value?

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/JavaMailSession.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class JavaMailSession {
+
+  public void foo() {
+    Properties props = new Properties();
+    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory");
+    Session s = Session.getDefaultInstance(props, new javax.mail.Authenticator() { // Noncompliant [[sc=17;ec=43]] {{Enable server identity validation, set "mail.smtp.ssl.checkserveridentity" to true}}
+      protected PasswordAuthentication getPasswordAuthentication() {
+        return new PasswordAuthentication("username@gmail.com", "password");
+      }
```
pynicolas: This piece of code to create an `Authenticator` is duplicated 4 times, but it's not relevant to the test. Can we make the test cases shorter so that we can see easily the differences between them?

java-checks/src/main/java/org/sonar/java/checks/security/SMTPSSLServerIdentityCheck.java
```diff
@@ -0,0 +1,150 @@
...
+        && (SET_SSL_CHECK_SERVER_ID.matches(mit) || SET_SSL_ON_CONNECT.matches(mit))
+        && LiteralUtils.isTrue(args.get(0))) {
+        bothFunctionsAreChecked++;
+      }
+
+      if (differentiateEmailAndSessionInvocations == 1 && MAIL_SMTP_SSL.matches(mit)) {
+        ExpressionTree arg1 = args.get(0);
+        ExpressionTree arg2 = args.get(1);
+        if (mailSessionIsChecked(arg1, arg2) || parametersSocketFactoryMatch(arg1, arg2)) {
+          bothFunctionsAreChecked++;
```
pynicolas: No issue is raised for the following code:
```
    Properties props = new Properties();
    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory");
    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory");
    Session s = Session.getDefaultInstance(props, new javax.mail.Authenticator() {...
```

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/JavaMailSession.java
stylianos-agapiou-sonarsource: I will fix that

java-checks/pom.xml
stylianos-agapiou-sonarsource: Now, that I don't need to highlight on `Session` instance, I got rid of it.

java-checks/src/main/java/org/sonar/java/checks/security/SMTPSSLServerIdentityCheck.java
```diff
@@ -0,0 +1,139 @@
...
+    while (!tree.is(Tree.Kind.CLASS, Tree.Kind.METHOD)) {
+      tree = tree.parent();
+    }
+    if (tree.is(Tree.Kind.CLASS)) {
+      return null;
+    }
+    return (MethodTree) tree;
+  }
+
+  private static boolean parametersSocketFactoryMatch(ExpressionTree arg1, ExpressionTree arg2) {
```
pynicolas: Having a separate method for that does not seem really useful and the method signature does not convey its intent.

java-checks/src/main/java/org/sonar/java/checks/security/SMTPSSLServerIdentityCheck.java
```diff
@@ -0,0 +1,139 @@
...
+    public void visitMethodInvocation(MethodInvocationTree mit) {
+      Arguments args = mit.arguments();
+      if (HASHTABLE_PUT.matches(mit)
+        && mailSessionIsChecked(args.get(0), args.get(1))) {
+        this.isSecured = true;
+      }
+      super.visitMethodInvocation(mit);
+    }
+
+    private static boolean mailSessionIsChecked(ExpressionTree arg1, ExpressionTree arg2) {
```
pynicolas: Having a separate method for that does not seem really useful and the method signature does not convey its intent.

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/ApacheEmail.java
```diff
@@ -0,0 +1,45 @@
...
+
+  public void foo4() {
+    Email email = new SimpleEmail();
+    email.setSSLOnConnect(true);  // Noncompliant
+    email.setSSLCheckServerIdentity(false);
+  }
+
+  public void false_negative1(boolean flag) {
+    Email email = new SimpleEmail();
+    email.setSSLOnConnect(flag);     // Compliant FN flag value will be read with SE engine
```
pynicolas: Is it really a FN? I think we can't tell without more context.

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/JavaMailSession.java
```diff
@@ -0,0 +1,27 @@
...
+  public void foo4() {
+    Properties props = new Properties();
+    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory"); // Noncompliant {{Enable server identity validation, set "mail.smtp.ssl.checkserveridentity" to true}}
+    props.put("mail.smtp.ssl.checkserveridentity", false);
+  }
+
+  public void false_positive(boolean flag) {
+    Properties props = new Properties();
+    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory");
+    props.put("mail.smtp.ssl.checkserveridentity", flag);  // Compliant FN
```
pynicolas: Is it really a FN?

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/JavaMailSession.java
```diff
@@ -0,0 +1,27 @@
...
+    props.put("mail.smtp.ssl.checkserveridentity", true);
+  }
+
+  public void foo4() {
+    Properties props = new Properties();
+    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory"); // Noncompliant {{Enable server identity validation, set "mail.smtp.ssl.checkserveridentity" to true}}
+    props.put("mail.smtp.ssl.checkserveridentity", false);
+  }
+
+  public void false_positive(boolean flag) {
```
pynicolas: The method name is not consistent with the content of the method.

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/JavaMailSession.java
```diff
@@ -0,0 +1,27 @@
...
+    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory"); // Compliant
+    props.put("mail.smtp.ssl.checkserveridentity", true);
+  }
+
+  public void foo4() {
+    Properties props = new Properties();
+    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory"); // Noncompliant {{Enable server identity validation, set "mail.smtp.ssl.checkserveridentity" to true}}
+    props.put("mail.smtp.ssl.checkserveridentity", false);
+  }
+
```
pynicolas: I would add a test where we call the `put` method with a different property name.

java-checks/src/main/java/org/sonar/java/checks/security/SMTPSSLServerIdentityCheck.java
```diff
@@ -0,0 +1,139 @@
...
+    MethodTree method = findEnclosingMethod(mit);
+    if (method != null) {
+      Arguments args = mit.arguments();
+      if (SET_SSL_ON_CONNECT.matches(mit) && LiteralUtils.isTrue(args.get(0))) {
+        MethodBodyApacheVisitor apacheVisitor = new MethodBodyApacheVisitor();
+        method.accept(apacheVisitor);
+        if (!apacheVisitor.isSecured) {
+          reportIssue(mit, "Enable server identity validation on this SMTP SSL connection.");
+        }
+      } else if (HASHTABLE_PUT.matches(mit) && parametersSocketFactoryMatch(args.get(0), args.get(1))) {
```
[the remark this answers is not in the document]
stylianos-agapiou-sonarsource: In the MethodMatcher `HASHTABLE_PUT` I use the function `withAnyParameters`,
because I use it twice.
In the function `onMethodInvocationFound` we proceed only if both parameters are Strings.
On the other hand, in the function `visitMethodInvocation` of the `MethodBodyHashtableVisitor`
we want the first parameter to be a String and the second boolean.
I could create a second MethodMatcher for the `MethodBodyHashtableVisitor`.

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/JavaMailSession.java
stylianos-agapiou-sonarsource: My mistake it is a FN, if the parameter's value is true, which we don't know.

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/JavaMailSession.java
```diff
@@ -0,0 +1,43 @@
...
+  public void foo4(boolean flag) {
+    Properties props = new Properties();
+    props.put("mail.smtp.socketFactory.class", "javax.net.ssl.SSLSocketFactory");
+    props.put("mail.smtp.ssl.checkserveridentity", flag);  // Compliant
+  }
+
+  public void foo5() {
+    Properties props = new Properties();
+    props.put("mail.smtp.socketFactory.class", ""); // coverage
```
pynicolas: I don't understand this "coverage" comments. It sounds like "I had to please a tool".

java-checks/src/test/files/checks/security/SMTPSSLServerIdentityCheck/ApacheEmail.java
```diff
@@ -0,0 +1,50 @@
+import org.apache.commons.mail.Email;
+import org.apache.commons.mail.SimpleEmail;
+
+public class ApacheEmail {
+
+  static {
+    Email email = new SimpleEmail();
+    email.setSSLOnConnect(true);  // for coverage
```
pynicolas: Rather than "for coverage", I would write "invocation outside a method".

java-checks/src/main/java/org/sonar/java/checks/security/SMTPSSLServerIdentityCheck.java
```diff
@@ -0,0 +1,129 @@
...
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S4499")
+public class SMTPSSLServerIdentityCheck extends AbstractMethodDetection {
+
+  private static final String APACHE = "org.apache.commons.mail.Email";
```
pynicolas: Let's rename to `APACHE_EMAIL`.

PR #2043, opened 2018-05-25
SONARJAVA-2769 S2255 should not raise issue for null value stored in a Cookie

andrei-epure-sonarsource:


java-frontend/src/main/java/org/sonar/java/model/LiteralUtils.java
```diff
@@ -92,7 +92,11 @@ private static Integer minus(@Nullable Integer nullableInteger) {
   }
 
   public static boolean isEmptyString(Tree tree) {
-    return tree.is(Tree.Kind.STRING_LITERAL) && trimQuotes(((LiteralTree) tree).value()).isEmpty();
+    return tree.is(Kind.STRING_LITERAL) && trimQuotes(((LiteralTree) tree).value()).isEmpty();
+  }
+
+  public static boolean isNullOrWhitespace(Tree tree) {
+    return tree.is(Kind.NULL_LITERAL) || (tree.is(Kind.STRING_LITERAL) && StringUtils.isBlank(trimQuotes(((LiteralTree) tree).value())));
```
pynicolas: I wouldn't put this method here. It seems quite specific to the rule we want to fix.

java-frontend/src/test/java/org/sonar/java/model/LiteralUtilsTest.java
```diff
@@ -184,11 +184,13 @@ public void is_0xff() {
     assertThat(LiteralUtils.is0xff(tree)).isFalse();
   }
 
+  @Test
```
pynicolas: Ouch! Can we think of a new rule to catch that?

m-g-sonar: Public non-static method in test class, without invocations from within the file, and having assertions but no test annotations?

pynicolas: @m-g-sonar Let's evaluate that pattern on our code base :)

PR #2050, opened 2018-05-31
SONARSEC-77 use projectDefinition to get unique working dir for entir…

benzonico:
…e analysis

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/SonarComponents.java
+++ b/java-frontend/src/main/java/org/sonar/java/SonarComponents.java
@@ -81,6 +81,7 @@ public class SonarComponents {
   private final FileLinesContextFactory fileLinesContextFactory;
   private final JavaTestClasspath javaTestClasspath;
   private final CheckFactory checkFactory;
+  @Nullable
   private final ProjectDefinition projectDefinition;
   private final FileSystem fs;
   private final JavaClasspath javaClasspath;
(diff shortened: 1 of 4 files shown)


java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -90,19 +92,28 @@
   @VisibleForTesting
   public List<AnalysisError> analysisErrors;
   private int errorsSize = 0;
+
   public SonarComponents(FileLinesContextFactory fileLinesContextFactory, FileSystem fs,
                          JavaClasspath javaClasspath, JavaTestClasspath javaTestClasspath,
                          CheckFactory checkFactory) {
-    this(fileLinesContextFactory, fs, javaClasspath, javaTestClasspath, checkFactory, null);
+    this(fileLinesContextFactory, fs, javaClasspath, javaTestClasspath, checkFactory, null, null);
```
m-g-sonar: This constructor is going to make `workDir()` trigger a NPE... but only used in tests. Should we still mark it as `@Nullable` and null-checked? I don't really know what would be the best. In production we are never going to have it null.

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -87,6 +87,8 @@ public UCFGJavaVisitor(File workdir) {
     this.protobufDirectory = new File(new File(workdir, "ucfg"), "java");
     if(!protobufDirectory.exists()) {
       protobufDirectory.mkdirs();
+    } else {
+      index = protobufDirectory.list().length;
```
m-g-sonar: I would appreciate a comment here to explicitly mention that we restart from the number of UCFG already created.

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -351,6 +356,28 @@ private UCFG createUCFG(String source) {
...
+    CompilationUnitTree cut = getCompilationUnitTreeWithSemantics("class A {String fun() {return \"\";}} ");
+    UCFGJavaVisitor UCFGJavaVisitor = new UCFGJavaVisitor(tmp.getRoot());
+    UCFGJavaVisitor.fileKey = FILE_KEY;
+    UCFGJavaVisitor.visitCompilationUnit(cut);
+
+    UCFGJavaVisitor = new UCFGJavaVisitor(tmp.getRoot());
+    UCFGJavaVisitor.fileKey = FILE_KEY;
+    UCFGJavaVisitor.visitCompilationUnit(cut);
+
+    String[] list = new File(new File(tmp.getRoot(), "ucfg"), "java").list();
```
m-g-sonar: isn't it `java_ucfg_dir` here ?

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
benzonico: As it can be null in sonarlint context i am going to be defensive about this even though that should not be triggered.

PR #2051, opened 2018-06-08
SONARJAVA-2751 RSPEC-4635 String offset rule

benzonico:


java-checks/src/test/files/checks/StringOffsetMethodsCheck.java
```diff
@@ -0,0 +1,18 @@
+class A {
+  void test(String str, char char1, int beginIndex, String str2) {
+    str.substring(beginIndex).indexOf(char1); // Noncompliant
```
andrei-epure-sonarsource: message + location

PR #2054, opened 2018-06-11
SONARJAVA-2757 Rule S4507: Spring Security's debug mode should not be used

alban-auzeill:


java-checks/src/test/files/checks/spring/SpringSecurityDebugModeCheck.java
```diff
@@ -0,0 +1,12 @@
+package files.checks.spring;
+
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+
+@EnableWebSecurity(debug = true) // Noncompliant [[sc=20;ec=32]] {{Deactivate Spring Security's debug mode.}}
+@EnableWebSecurity(unknown = true, debug = true) // Noncompliant [[sc=36;ec=48]]
+@EnableWebSecurity(debug = false)
+@EnableWebSecurity(debug)
+@EnableWebSecurity(true)
```
benzonico: Why is that case compliant given the rule ? it should be noncompliant as per : https://docs.spring.io/spring-security/site/docs/4.2.5.RELEASE/apidocs/org/springframework/security/config/annotation/web/configuration/EnableWebSecurity.html debug is the only attribute.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringSecurityDebugModeCheck.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  private static boolean isDebugArgument(ExpressionTree expression) {
+    if (expression.is(Tree.Kind.ASSIGNMENT)) {
+      AssignmentExpressionTree assignment = (AssignmentExpressionTree) expression;
+      return assignment.variable().is(Tree.Kind.IDENTIFIER) &&
+        "debug".equals(((IdentifierTree) assignment.variable()).name());
+    }
+    return false;
```
benzonico: In this method you should most probably use `ConstantUtils.resolveAsBooleanConstant` to handle eventual constant passed as arguments in order to cover more (and trickier) cases.

java-checks/src/test/files/checks/spring/SpringSecurityDebugModeCheck.java
alban-auzeill: Attribute name can be omitted only when it is called "value". Here, it is semantically not valid java code, because "EnableWebSecurity" does not have a "value" attribute. I added this case only for coverage reason.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringSecurityDebugModeCheck.java
alban-auzeill: Like I said in my previous comment, the attribute name is "debug" and not "value". So attribute name will not be omitted. It will always be an ASSIGNMENT. So `ConstantUtils.resolveAsBooleanConstant` used at line 48 should be enough.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringSecurityDebugModeCheck.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    AnnotationTree annotation = (AnnotationTree) tree;
+    if (annotation.symbolType().is("org.springframework.security.config.annotation.web.configuration.EnableWebSecurity")) {
+      annotation.arguments().stream()
+        .filter(SpringSecurityDebugModeCheck::isDebugArgument)
+        .map(AssignmentExpressionTree.class::cast)
+        .filter(assignment -> Boolean.TRUE.equals(ConstantUtils.resolveAsBooleanConstant(assignment.expression())))
```
benzonico: Can you please add a testcase where this is used with a constant ?

benzonico: This filter should be part of the "isDebugArgument" function to avoid an implicit coupling between the "isDebugArgument" and the cast to AssignmentExpressionTree.

I would make the "isDebugArgument" to a "getDebugArgument" possibly returning null and then filtering on non-null values.

alban-auzeill: I added Boolean.TRUE

PR #2058, opened 2018-06-11
SONARJAVA-2747 Most specific method should apply correct substitution

benzonico:


java-frontend/src/test/files/sym/MostSpecificMethod.java
```diff
@@ -0,0 +1,14 @@
+public class A implements I {
+  public A() {
+    foo(String.class); // resolved to I.foo(...)
```
m-g-sonar: Please remove or fix the comments, they describes the wrong behavior which was observed before.

java-frontend/src/test/java/org/sonar/java/resolve/SymbolTableTest.java
```diff
@@ -873,8 +873,9 @@ public void constructorWithTypeArguments() {
     JavaSymbol.TypeJavaSymbol classSymbol = (JavaSymbol.TypeJavaSymbol) result.symbol("MyClass");
     List<JavaSymbol> constructors = classSymbol.members.lookup("<init>");
 
-    assertThat(constructors.get(0).usages()).hasSize(1);
-    assertThat(constructors.get(1).usages()).hasSize(1);
+    // FIXME : there should be one reference to each constructor : this is a bug in how method inference specificity is applied see JLS8 18.5.4
```
m-g-sonar: This is bothering me. By fixing this ticket we may fix a case (and obviously fix an issue which was there for a while), but we break another one which was working... And this test was existing solely for the case which is now not properly resolved. Should we reoppen [SONARJAVA-1667](https://jira.sonarsource.com/browse/SONARJAVA-1667)?

java-frontend/src/main/java/org/sonar/java/resolve/Resolve.java
```diff
@@ -742,15 +742,22 @@ private JavaSymbol selectMostSpecific(JavaSymbol m1, JavaSymbol m2, List<JavaTyp
...
       m2Substitution = typeSubstitutionSolver.getTypeSubstitution((JavaSymbol.MethodJavaSymbol) m2, callSite, ImmutableList.of(), argTypes);
     }
     if (m2Substitution == null) {
       m2Substitution = new TypeSubstitution();
     }
     boolean m1SignatureMoreSpecific = isSignatureMoreSpecific(m1, m2, argTypes, m1Substitution, m2Substitution);
-    boolean m2SignatureMoreSpecific = isSignatureMoreSpecific(m2, m1, argTypes, m1Substitution, m2Substitution);
+    boolean m2SignatureMoreSpecific = isSignatureMoreSpecific(m2, m1, argTypes, m2Substitution, m1Substitution);
     if (m1SignatureMoreSpecific && m2SignatureMoreSpecific) {
+      if(!m1IsGeneric && m2IsGeneric) {
```
m-g-sonar: This deserves a reference to JLS (I'm not sure it fits https://docs.oracle.com/javase/specs/jls/se10/html/jls-18.html#jls-18.5.4 ?)

From my point of view, this is also maybe incomplete, as the JLS speaks about type inference, which is not necessarily the case with all generics.

PR #2060, opened 2018-06-12
SONARJAVA-2759 Rule S4602: "@SpringBootApplication" and "@ComponentScan" should not be used in the default package

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
```diff
@@ -0,0 +1,171 @@
...
+
+  private void checkIfAnnotationIsInDefaultPackage(AnnotationTree annotation) {
+    if (isNodeInDefaultPackage(annotation)) {
+      reportIssue(annotation.annotationType(), "Remove the annotation \"@" + annotation.symbolType().name() +
+        "\" or move the annotated class out of the default package.");
+    }
+  }
+
+  private void checkAnnotationPackageAttributes(AnnotationTree annotation, List<ExpressionTree> scanPackageAttributes) {
+    List<ExpressionTree> values = scanPackageAttributes.stream()
```
benzonico: This can be part of the previous collect and the list of values should be passed to this method.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
```diff
@@ -0,0 +1,171 @@
...
+    values.stream()
+      .flatMap(SpringScanDefaultPackageCheck::findClassInDefaultPackage)
+      .forEach(identifier -> reportIssue(identifier, "Remove the annotation \"@" + annotation.symbolType().name() +
+        "\" or move the \"" + identifier.name() + "\" class out of the default package."));
+  }
+
+  private static String attributeName(ExpressionTree expression) {
+    if (expression.is(Tree.Kind.ASSIGNMENT)) {
+      AssignmentExpressionTree assignment = (AssignmentExpressionTree) expression;
+      if (assignment.variable().is(Tree.Kind.IDENTIFIER)) {
```
benzonico: can it be something else ? is it covered ? if not and it is defensive programming : worth a comment to explain lack of coverage.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
```diff
@@ -0,0 +1,171 @@
...
+    Set<String> scanPackageAttributeNames = SCAN_PACKAGE_ATTRIBUTES.get(annotation.symbolType().fullyQualifiedName());
+    if (scanPackageAttributeNames != null) {
+      List<ExpressionTree> scanPackageAttributes = annotation.arguments().stream()
+        .filter(argument -> scanPackageAttributeNames.contains(attributeName(argument)))
+        .collect(Collectors.toList());
+
+      if (scanPackageAttributes.isEmpty()) {
+        checkIfAnnotationIsInDefaultPackage(annotation);
+      } else {
+        checkAnnotationPackageAttributes(annotation, scanPackageAttributes);
```
benzonico: IMO you should always pass the scanPackageValues to one method and if it is empty check the default package.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
```diff
@@ -0,0 +1,171 @@
...
+      .collect(Collectors.toList());
+
+    values.stream()
+      .flatMap(SpringScanDefaultPackageCheck::findEmptyString)
+      .forEach(expression -> reportIssue(expression, "Define packages to scan. Don't rely on the default package."));
+
+    values.stream()
+      .flatMap(SpringScanDefaultPackageCheck::findClassInDefaultPackage)
+      .forEach(identifier -> reportIssue(identifier, "Remove the annotation \"@" + annotation.symbolType().name() +
+        "\" or move the \"" + identifier.name() + "\" class out of the default package."));
```
benzonico: With this approach of reusing the list you can end up with a case where two issues are raised : I don't see this testcase. 
I am not sure this is actually relevant : if you have one attribute that could raise an issue and the other not, should we raise an issue in this case ?

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+  }
+
+  private void checkAnnotationPackageAttributes(AnnotationTree annotation, List<ExpressionTree> scanPackageAttributes) {
+    List<ExpressionTree> values = scanPackageAttributes.stream()
+      .flatMap(SpringScanDefaultPackageCheck::extractValues)
+      .collect(Collectors.toList());
+
+    values.stream()
+      .flatMap(SpringScanDefaultPackageCheck::findEmptyString)
```
benzonico: This should not use flatMap but just map to an optional of expression and the lambda in foreach should make use of isPresent.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
```diff
@@ -0,0 +1,171 @@
...
+
+  private static Stream<ExpressionTree> findEmptyString(ExpressionTree expression) {
+    String stringValue = ConstantUtils.resolveAsStringConstant(expression);
+    if (stringValue != null && stringValue.isEmpty()) {
+      return Stream.of(expression);
+    }
+    return Stream.empty();
+  }
+
+  private static Stream<IdentifierTree> findClassInDefaultPackage(ExpressionTree expression) {
```
benzonico: Method that returns a stream of one element OR an empty stream should make use of optional rather than stream.

benzonico: Moreover another issue here is that we could imagine to have mutiple issues on the same annotation and that does not seem to be covered in testcases. (and is it something expected).

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
alban-auzeill: ok, moved to the caller of this method

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
alban-auzeill: the only possible kind for "variable()" in annotation is IDENTIFIER ([JavaGrammar.java:L589](https://github.com/SonarSource/sonar-java/blob/01ec449f34ed81b46a85c30a1df89131d2432ac2/java-frontend/src/main/java/org/sonar/java/ast/parser/JavaGrammar.java#L589)) , I removed the test.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
alban-auzeill: I moved the ".isEmpty()" test inside "checkAnnotationPackageAttributes"

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
alban-auzeill: Issues are raised on each invalid values of "@ComponentScan", it's intentional. I added a test case to highlight those cases: several empty strings "" and several classes in the default package.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
alban-auzeill: Now it is covered.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheck.java
```diff
@@ -0,0 +1,164 @@
...
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.ANNOTATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    AnnotationTree annotation = (AnnotationTree) tree;
+    Set<String> scanPackageAttributeNames = SCAN_PACKAGE_ATTRIBUTES.get(annotation.symbolType().fullyQualifiedName());
```
andrei-epure-sonarsource: do actions only `if (hasSemantic())` ?

java-checks/src/test/files/checks/spring/SpringScanDefaultPackageCheckInAPackage.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.boot.web.servlet.ServletComponentScan;
+import org.springframework.context.annotation.ComponentScan;
+
+@ComponentScan
+@SpringBootApplication
+@ServletComponentScan
+
+@ComponentScan(basePackages = "") // Noncompliant
+@ComponentScan(basePackageClasses = App.class)
+
```
andrei-epure-sonarsource: nitpick: also add `@SpringBootApplication(scanBasePackages = "") // Noncompliant` and `@ServletComponentScan("") // Noncompliant`

java-checks/src/test/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheckTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.Test;
+import org.sonar.java.checks.verifier.JavaCheckVerifier;
+
+public class SpringScanDefaultPackageCheckTest {
+
+  @Test
+  public void test() {
+    JavaCheckVerifier.verify("src/test/files/checks/spring/SpringScanDefaultPackageCheck.java", new SpringScanDefaultPackageCheck());
+    JavaCheckVerifier.verify("src/test/files/checks/spring/SpringScanDefaultPackageCheckInAPackage.java", new SpringScanDefaultPackageCheck());
+    JavaCheckVerifier.verifyNoIssueWithoutSemantic("src/test/files/checks/spring/SpringScanDefaultPackageCheck.java", new SpringScanDefaultPackageCheck());
```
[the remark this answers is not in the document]
alban-auzeill: I disagree, now there's a "if (hasSemantic())" check, so one test is enough to cover this case.

java-checks/pom.xml
```diff
@@ -177,6 +177,21 @@
                   <artifactId>spring-security-config</artifactId>
                   <version>5.0.5.RELEASE</version>
                 </dependency>
+                <dependency>
```
andrei-epure-sonarsource: artifactItem

PR #2063, opened 2018-06-12
Upgrade to Orchestrator 3.19 to fix support of ITs with SQ 7.2

simonbrandhof:


its/plugin/tests/src/test/java/com/sonar/it/java/suite/JavaTest.java
```diff
@@ -103,7 +102,7 @@ public void shouldFailIfInvalidJavaPackage() {
 
   @Test
   public void should_create_issue_pom_xml() {
-    assumeTrue(orchestrator.getServer().version().isGreaterThanOrEquals("6.3"));
+    assumeTrue(orchestrator.getServer().version().isGreaterThanOrEquals(6, 3));
```
benzonico: we can probably drop that line altogether in fact...

PR #2065, opened 2018-06-12
SONARJAVA-2762 Rule S4605: Spring beans seen by "@ComponentScan"

andrei-epure-sonarsource:


java-frontend/src/test/java/org/sonar/java/model/DefaultJavaFileScannerContextTest.java
```diff
@@ -72,6 +161,26 @@ public void report_issue_on_tree() {
...
+    assertThat(reportedMessage.flows).isEmpty();
+
+    assertMessagePosition(reportedMessage, 1, 0, 4, 1);
+  }
+
+  @Test
+  public void report_issue_on_tree_with_cross_file_scanner_throws() {
+    assertThatExceptionOfType(UnsupportedOperationException.class)
+        .isThrownBy(() -> context.reportIssue(CROSS_FILE_SCANNER, compilationUnitTree, "msg"))
+        .withMessage("CrossFileScanner show only call reportIssue with AnalyzerMessage and should never pass a Tree reference.");
```
andrei-epure-sonarsource: s/show/should

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,157 @@
...
+    } else if (packageNames.is(Tree.Kind.NEW_ARRAY)) {
+      for (ExpressionTree p : ((NewArrayTree) packageNames).initializers()) {
+        String name = ConstantUtils.resolveAsStringConstant(p);
+        scannedPackages.add(name);
+      }
+    }
+  }
+
+  private static String packageNameOf(ClassTree classTree) {
+    Tree classTreeParent = classTree.parent();
```
m-g-sonar: You could also rely on the `TypeSymbol` of the classTree and simply remove everything before the className.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,157 @@
...
+    if (classTreeParent != null && classTreeParent.is(Tree.Kind.COMPILATION_UNIT)) {
+      return PackageUtils.packageName(((CompilationUnitTree) classTreeParent).packageDeclaration(), ".");
+    }
+    return "";
```
m-g-sonar: So we implicitly ignore inner classes which would be beans? Why? (Not sure it's possible, from a spring perspective). Worth adding a comment if it's not possible, or handle the inner classes.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  private static boolean hasAnnotation(ClassTree classTree, String... annotationName) {
+    return Arrays.stream(annotationName).anyMatch(annotation -> classTree.symbol().metadata().isAnnotatedWith(annotation));
```
m-g-sonar: extract `classTree.symbol().metadata()` to not recompute it for each annotation.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,157 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    if (!hasSemantic() || !(context instanceof AnalyzerMessageReporter)) {
+      return;
+    }
+
+    ClassTree classTree = (ClassTree) tree;
+    String classPackageName = packageNameOf(classTree);
+
+    Optional<AnnotationTree> componentScanAnnotation = classTree.modifiers().annotations().stream()
```
m-g-sonar: I think it would be simpler to rewrite all this part relying on `valuesForAnnotation(String)` from the symbolMetdata, and then check for values.

If the method returns null, it is not annotated with the annotation, if it return a list, then you have directly the values (which could be both symbols or trees, but we probably only needs trees)

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,157 @@
...
+      scannedPackages.add(classPackageName);
+    } else if (hasAnnotation(classTree, SPRING_BEAN_ANNOTATIONS)) {
+      addMessageToMap(classPackageName, classTree);
+    }
+  }
+
+  private void addMessageToMap(String classPackageName, ClassTree classTree) {
+    AnalyzerMessageReporter reporter = (AnalyzerMessageReporter) context;
+    IdentifierTree className = classTree.simpleName();
+    if (className != null) {
```
m-g-sonar: you should drop anonymous class before, preferably directly in the visitNode method.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,157 @@
...
+      AnalyzerMessage analyzerMessage = reporter.createAnalyzerMessage(this, className, String.format(MESSAGE_FORMAT, className));
+      messagesPerPackage.computeIfAbsent(classPackageName, k -> new ArrayList<>());
```
m-g-sonar: `computeIfAbsent` returns the value if present, or the new value. So you already have the list and does not need to get it again on the next line. Just chain the `.add(analyzerMessage)`

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,157 @@
...
+      ExpressionTree argumentAssignmentVar = argumentAssignment.variable();
+      if (argumentAssignmentVar.is(Tree.Kind.IDENTIFIER) && COMPONENT_SCAN_ARGUMENTS.contains(((IdentifierTree) argumentAssignmentVar).name())) {
+        addLiteralsToScannedPackages(argumentAssignment.expression());
+      }
+    } else {
+      addLiteralsToScannedPackages(annotationArgument);
+    }
+  }
+
+  private void addLiteralsToScannedPackages(ExpressionTree packageNames) {
```
m-g-sonar: You can drop all of this when using method from metadata.

java-frontend/src/main/java/org/sonar/java/CrossFileScanner.java
```diff
@@ -0,0 +1,34 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java;
+
+import org.sonar.plugins.java.api.JavaCheck;
+
+/**
+ * Common interface for all checks that do cross-file analysis.
+ */
+public interface CrossFileScanner extends JavaCheck {
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: Thanks for raising the issue. Naming is always difficult, especially for high-level abstractions.

java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
```diff
@@ -231,6 +232,10 @@ public void endOfAnalysis() {
       }
       LOG.warn("Classes not found during the analysis : [{}{}]", classesNotFound.stream().limit(50).collect(Collectors.joining(", ")), message);
     }
+    scanners.stream()
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: I'd rather open a task to make that change. `executableScanners` gets set twice, once in the constructor and once in `setJavaVersion` using `scanners`. It is out of scope for this PR.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
andrei-epure-sonarsource: good catch!

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
andrei-epure-sonarsource: good idea!

java-frontend/src/main/java/org/sonar/java/model/DefaultJavaFileScannerContext.java
```diff
@@ -109,29 +112,39 @@ public void addIssue(File file, JavaCheck check, int line, String message) {
 
   @Override
   public void reportIssue(JavaCheck javaCheck, Tree tree, String message) {
+    throwIfCrossFileScanner(javaCheck);
     reportIssue(javaCheck, tree, message, ImmutableList.of(), null);
   }
 
   @Override
   public void reportIssue(JavaCheck javaCheck, Tree syntaxNode, String message, List<Location> secondary, @Nullable Integer cost) {
+    throwIfCrossFileScanner(javaCheck);
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: `reportIssue(AnalyzerMessage message)` is called twice in this class, so I have two options:

* I throw only in those two methods - line 127 and line 139 in this file
* I throw in all five public methods

I choose to throw in all, because the intent is more clear like this. Moreover, other public classes might change and call `reportIssue(AnalyzerMessage message)` directly, and would then need to do the throw-check

java-frontend/src/main/java/org/sonar/java/model/DefaultJavaFileScannerContext.java
```diff
@@ -109,29 +112,44 @@ public void addIssue(File file, JavaCheck check, int line, String message) {
 
   @Override
   public void reportIssue(JavaCheck javaCheck, Tree tree, String message) {
+    throwIfEndOfAnalysisCheck(javaCheck);
```
benzonico: Let's use this method on overloading.

java-frontend/src/main/java/org/sonar/java/model/DefaultJavaFileScannerContext.java
```diff
@@ -38,7 +41,7 @@
 import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
-public class DefaultJavaFileScannerContext implements JavaFileScannerContext {
+public class DefaultJavaFileScannerContext implements JavaFileScannerContext, AnalyzerMessageReporter {
```
benzonico: I don't think we should use a dedicated interface, let's downcast in check implementation for this one case for now, we'll clean it up if required.

java-frontend/src/main/java/org/sonar/java/EndOfAnalysisCheck.java
```diff
@@ -0,0 +1,36 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java;
+
+import com.google.common.annotations.Beta;
+import org.sonar.plugins.java.api.JavaCheck;
+
+/**
+ * Common interface for checks that are triggered at the end of the analysis, after all files have been scanned
```
benzonico: Please add a warning that this should be used with care to avoid memory leak.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+
+  /**
+   * Returns the prefix of the class name, which:
+   * - in general, is the package name (e.g. 'foo.bar')
+   * - for inner classes, is the package name + outer class name (e.g. 'foo.bar.Outer')
+   * - for classes in the default package, is the empty string
+   */
+  private static String extractClassPrefix(String fullyQualifiedClassName) {
```
benzonico: Please use the `Convert` class for all those operations as I am really unsure your implementation would work with a class with a $ in its name.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,145 @@
...
+    } else if (hasAnnotation(classSymbolMetadata, SPRING_BOOT_APP_ANNOTATION)) {
+      scannedPackages.add(classPackageName);
+    } else if (hasAnnotation(classSymbolMetadata, SPRING_BEAN_ANNOTATIONS)) {
+      addMessageToMap(classPackageName, classTree.simpleName());
+    }
+  }
+
+  private void addMessageToMap(String classPackageName, IdentifierTree classNameTree) {
+    AnalyzerMessageReporter reporter = (AnalyzerMessageReporter) context;
+    AnalyzerMessage analyzerMessage = reporter.createAnalyzerMessage(this, classNameTree, String.format(MESSAGE_FORMAT, classNameTree));
```
benzonico: The message formatting should be relying on classnameTree.name  not on the toString of the tree.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
andrei-epure-sonarsource: It does work with inner classes. I have tests for this in `../ComponentScan/A.java`.

Nevertheless, good point, I was not aware of the `Convert` class.  I added noncompliant tests as well in `../ComponentScan/C.java`.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -0,0 +1,132 @@
...
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+
+    if (!hasSemantic() || classTree.simpleName() == null) {
+      return;
+    }
+
+    String classPackageName = Convert.packagePart(classTree.symbol().type().fullyQualifiedName());
```
benzonico: https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/java/org/sonar/java/checks/StaticMethodCheck.java#L149

PR #2066, opened 2018-06-12
SONARJAVA-2181 Rule S3752: "@RequestMapping" methods should specify HTTP method

alban-auzeill:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S3752_java.html
```diff
@@ -0,0 +1,26 @@
+<p>A <code>@RequestMapping</code> method handles all matching requests by default. That means that a method you intended only to be
```
andrei-epure-sonarsource: If you talk with @agigleux and decide that we should decide to recommend usage of `@PostMapping`, don't forget to update the rule description as well as the message

alban-auzeill: Alex's rule description is compatible with older version of Spring. About @PostMapping we are going to enforce it using another rule: RSPEC-4488

PR #2067, opened 2018-06-13
SONARJAVA-2753 Rule S3305: No autowired fields in "@Configuration" class

andrei-epure-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/spring/SpringConfigurationWithAutowiredFieldsCheck.java
```diff
@@ -0,0 +1,117 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+    if (hasSemantic() && classTree.symbol().metadata().isAnnotatedWith(CONFIGURATION_ANNOTATION)) {
+      AutowiredFieldVisitor autowiredFieldVisitor = new AutowiredFieldVisitor();
+      classTree.accept(autowiredFieldVisitor);
+      BeanMethodVisitor beanMethodVisitor = new BeanMethodVisitor(autowiredFieldVisitor.autowiredFields.keySet());
+      classTree.accept(beanMethodVisitor);
+      beanMethodVisitor.beanMethodsThatUseAutowiredField.entrySet().stream()
+        .filter(fieldSymbolAndBeanMethods -> fieldSymbolAndBeanMethods.getValue().size() == 1)
```
andrei-epure-sonarsource: better rename the Entry to _beanMethodsThatUseAutowiredField_ and the map to _beanMethodsThatUseAutowiredField_**s**

java-checks/src/main/java/org/sonar/java/checks/spring/SpringConfigurationWithAutowiredFieldsCheck.java
```diff
@@ -0,0 +1,117 @@
...
+    return Collections.singletonList(Tree.Kind.CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+    if (hasSemantic() && classTree.symbol().metadata().isAnnotatedWith(CONFIGURATION_ANNOTATION)) {
+      AutowiredFieldVisitor autowiredFieldVisitor = new AutowiredFieldVisitor();
+      classTree.accept(autowiredFieldVisitor);
+      BeanMethodVisitor beanMethodVisitor = new BeanMethodVisitor(autowiredFieldVisitor.autowiredFields.keySet());
```
alban-auzeill: Note: Yes, here it's perhaps better to use a visitor that also check nested class methods. (but I'm not 100% sure)

java-checks/src/main/java/org/sonar/java/checks/spring/SpringConfigurationWithAutowiredFieldsCheck.java
```diff
@@ -0,0 +1,117 @@
...
+    private static class IdentifiersVisitor extends BaseTreeVisitor {
+      private final Map<Symbol, Boolean> isFieldReferenced = new HashMap<>();
+
+      IdentifiersVisitor(Set<Symbol> autowiredFields) {
+        autowiredFields.forEach(f -> isFieldReferenced.put(f, false));
+      }
+
+      @Override
+      public void visitIdentifier(IdentifierTree identifierTree) {
+        isFieldReferenced.computeIfPresent(identifierTree.symbol(), (fieldSym, isPresent) -> isPresent = true);
```
alban-auzeill: Replace "isPresent = true" by "true"

java-checks/src/main/java/org/sonar/java/checks/spring/SpringConfigurationWithAutowiredFieldsCheck.java
```diff
@@ -0,0 +1,117 @@
...
+  private static class BeanMethodVisitor extends BaseTreeVisitor {
+    private final Map<Symbol, List<String>> beanMethodsThatUseAutowiredFields = new HashMap<>();
+
+    BeanMethodVisitor(Set<Symbol> autowiredFields) {
+      autowiredFields.forEach(f -> beanMethodsThatUseAutowiredFields.put(f, new ArrayList<>()));
+    }
+
+    @Override
+    public void visitMethod(MethodTree methodTree) {
+      if (methodTree.symbol().metadata().isAnnotatedWith(BEAN_ANNOTATION)) {
```
alban-auzeill: I think we should also visit methods without annotation and do some logic to avoid the following FP:
```
@Configuration
class A {

  @Autowired private Bar multipleUsage; // false-positive

  @Bean
  public Foo method2() {
    return extractedMethod();
  }

  private Foo extractedMethod() {
    return new Foo(this.multipleUsage);
  }

  @Bean
  public Foo method3() {
    return new Foo(this.multipleUsage);
  }

}
```

java-checks/src/main/java/org/sonar/java/checks/spring/SpringConfigurationWithAutowiredFieldsCheck.java
```diff
@@ -0,0 +1,117 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+    if (hasSemantic() && classTree.symbol().metadata().isAnnotatedWith(CONFIGURATION_ANNOTATION)) {
+      AutowiredFieldVisitor autowiredFieldVisitor = new AutowiredFieldVisitor();
+      classTree.accept(autowiredFieldVisitor);
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: Good point. I now filter the members to avoid double scanning (because an inner static class is a member, too).

java-checks/src/main/java/org/sonar/java/checks/spring/SpringConfigurationWithAutowiredFieldsCheck.java
```diff
@@ -0,0 +1,114 @@
...
+  private static final String MESSAGE_FORMAT = "Inject this field value directly into \"%s\", the only method that uses it.";
+
+  private static final String CONFIGURATION_ANNOTATION = "org.springframework.context.annotation.Configuration";
+  private static final String BEAN_ANNOTATION = "org.springframework.context.annotation.Bean";
+  private static final List<String> AUTOWIRED_ANNOTATIONS = Arrays.asList(
+    "org.springframework.beans.factory.annotation.Autowired",
+    "javax.inject.Inject");
+
+  private final Map<Symbol, VariableTree> autowiredFields = new HashMap<>();
+  private final Map<Symbol, List<MethodTree>> methodsThatUseAutowiredFields = new HashMap<>();
```
alban-auzeill: You don't need to be "state full at project level" for this rule.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringConfigurationWithAutowiredFieldsCheck.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+    if (classTree.symbol().metadata().isAnnotatedWith(CONFIGURATION_ANNOTATION) && isNotStatic(classTree)) {
```
alban-auzeill: IMO we should remove "isNotStatic(classTree)".

PR #2068, opened 2018-06-13
SONARJAVA-2666 Rule S4488: Composed "@RequestMapping" variants should be preferred

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/spring/SpringComposedRequestMappingCheck.java
```diff
@@ -0,0 +1,117 @@
...
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.NewArrayTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S4488")
+public class SpringComposedRequestMappingCheck extends IssuableSubscriptionVisitor {
+
+  private static final Map<String, String> PREFERRED_METHOD_MAP = buildPreferredMethodMap();
```
andrei-epure-sonarsource: You could skip the helper method, I find it more readable like below

```
  PREFERRED_METHOD_MAP = Collections.unmodifiableMap(Stream.of(
         entry("GET", "@GetMapping"),
         entry("POST", "@PostMapping"))
        .collect(Collectors.toMap((e) -> e.getKey(), (e) -> e.getValue())));

  // end of file

  private static AbstractMap.SimpleEntry<String, String> entry(String k, String v) { return new AbstractMap.SimpleEntry<>(k, v); }
```

PR #2069, opened 2018-06-15
SONARJAVA-2750 S2070 consider Spring weak PasswordEncoders

andrei-epure-sonarsource:


java-checks/pom.xml
```diff
@@ -192,6 +192,16 @@
...
                 </artifactItem>
+                <artifactItem>
+                  <groupId>org.springframework.security</groupId>
+                  <artifactId>spring-security-crypto</artifactId>
+                  <version>5.0.6.RELEASE</version>
+                </artifactItem>
+                <artifactItem>
+                  <groupId>org.springframework.security</groupId>
+                  <artifactId>spring-security-core</artifactId>
+                  <version>4.2.6.RELEASE</version>
```
alban-auzeill: Why not using the latest version is 5.0.6.RELEASE ?

andrei-epure-sonarsource: ShaPasswordEncoder and Md5PasswordEncoder have been removed from Spring Security 5.x

PR #2071, opened 2018-06-18
SONARJAVA-2767 S2092: support more Cookie types

andrei-epure-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -19,96 +19,199 @@
...
+  /**
+   * Some constructors have the 'secure' parameter and do not need a 'setSecure' call afterwards.
+   */
+  private static final String CONSTRUCTOR = "<init>";
+  private static final String JAVA_LANG_STRING = "java.lang.String";
+  private static final String INT = "int";
+  private static final String BOOLEAN = "boolean";
+
+  private static final List<MethodMatcher> CONSTRUCTORS_WITH_SECURE_PARAM_LAST = Arrays.asList(
+    MethodMatcher.create().typeDefinition(TypeCriteria.subtypeOf(JAX_RS_NEW_COOKIE)).name(CONSTRUCTOR)
```
m-g-sonar: I see many repetitions of `MethodMatcher.create().typeDefinition(TypeCriteria.subtypeOf(...)).name(CONSTRUCTOR).parameters(...)`

Please extract a dedicated static method:

```java
private static MethodMatcher constructorOf(String type, String ... parameters) {
  return MethodMatcher.create()
    .typeDefinition(TypeCriteria.subtypeOf(type))
    .name(CONSTRUCTOR)
    .parameters(parameters);
}
```

It will simplify all these invocations and hopefully make it easier to read.

java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -19,96 +19,199 @@
...
+      .parameters(JAX_RS_COOKIE, JAVA_LANG_STRING, INT, BOOLEAN),
+    MethodMatcher.create().typeDefinition(TypeCriteria.subtypeOf(JAX_RS_NEW_COOKIE)).name(CONSTRUCTOR)
```
m-g-sonar: Seems strange to me to have a `subtypeOf()` here. I think we want an exact match... Except if all the other subtypes have constructors having the exact match of parameters. Is it the case? Is there some subtypes of all these cookies?

java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -19,96 +19,199 @@
...
   private void addToUnsecuredCookies(VariableTree variableTree) {
-    Type type = variableTree.type().symbolType();
-    if (type.is("javax.servlet.http.Cookie") && isConstructorInitialized(variableTree)) {
-      Symbol variableSymbol = variableTree.symbol();
-      //Ignore field variables
-      if (variableSymbol.isVariableSymbol() && variableSymbol.owner().isMethodSymbol()) {
-        unsecuredCookies.add((Symbol.VariableSymbol) variableSymbol);
+    ExpressionTree initializer = variableTree.initializer();
+    // Ignore field variables
+    if (initializer != null && variableTree.symbol().isVariableSymbol() && variableTree.symbol().owner().isMethodSymbol()) {
```
m-g-sonar: Extract `variableTree.symbol()` instead of calling it 3 times.

java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -19,96 +19,199 @@
...
+      if (isInitializedWithConstructor && isMatchedType && isSecureParamFalse((NewClassTree) initializer)) {
+        unsecuredCookies.add((Symbol.VariableSymbol) variableTree.symbol());
+      }
+    }
+  }
+
+  private void addToUnsecuredCookies(AssignmentExpressionTree assignment) {
+    if (assignment.expression().is(Tree.Kind.NEW_CLASS) && assignment.variable().is(Tree.Kind.IDENTIFIER)) {
+      IdentifierTree identifier = (IdentifierTree) assignment.variable();
+      boolean isMethodVariable = identifier.symbol().isVariableSymbol() && identifier.symbol().owner().isMethodSymbol();
```
m-g-sonar: extract `identifier.symbol()` into a dedicated variable in order to improve readability.

java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -19,96 +19,199 @@
...
+      boolean isMatchedType = isCookieClass(identifier.symbolType()) || isCookieClass(assignment.expression().symbolType());
+      if (isMethodVariable && isMatchedType && isSecureParamFalse((NewClassTree) assignment.expression())) {
+        unsecuredCookies.add((Symbol.VariableSymbol) ((IdentifierTree) assignment.variable()).symbol());
```
m-g-sonar: This can trigger NPE (even if unlickely...) if the variable is not defined in the same file and visible from this context:

```java
class A { // in A.java
  public UnsecureCookie myCookie;
}

class B extends A { // in B.java

  void foo() {
    myCookie = new UnsecureCookie(...); // will trigger a NPE, the declaration is not available
  }
}

File: java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java [hunk 6]
```diff
@@ -19,96 +19,199 @@
...
+      boolean isInitializedWithConstructor = initializer.is(Tree.Kind.NEW_CLASS);
+      boolean isMatchedType = isCookieClass(variableTree.symbol().type()) || isCookieClass(initializer.symbolType());
+      if (isInitializedWithConstructor && isMatchedType && isSecureParamFalse((NewClassTree) initializer)) {
+        unsecuredCookies.add((Symbol.VariableSymbol) variableTree.symbol());
+      }
+    }
+  }
+
+  private void addToUnsecuredCookies(AssignmentExpressionTree assignment) {
+    if (assignment.expression().is(Tree.Kind.NEW_CLASS) && assignment.variable().is(Tree.Kind.IDENTIFIER)) {
```
m-g-sonar: Restricting to identifiers won't detect case such as `this.myCookie`. Is it expected?

andrei-epure-sonarsource: Yes, we ignore fields.

java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -138,23 +129,29 @@ public void visitNode(Tree tree) {
...
   private void addToUnsecuredCookies(AssignmentExpressionTree assignment) {
     if (assignment.expression().is(Tree.Kind.NEW_CLASS) && assignment.variable().is(Tree.Kind.IDENTIFIER)) {
       IdentifierTree identifier = (IdentifierTree) assignment.variable();
-      boolean isMethodVariable = identifier.symbol().isVariableSymbol() && identifier.symbol().owner().isMethodSymbol();
+      Symbol identifierSymbol = identifier.symbol();
+      boolean isMethodVariable = identifierSymbol.isVariableSymbol() && identifierSymbol.owner().isMethodSymbol();
       boolean isMatchedType = isCookieClass(identifier.symbolType()) || isCookieClass(assignment.expression().symbolType());
-      if (isMethodVariable && isMatchedType && isSecureParamFalse((NewClassTree) assignment.expression())) {
-        unsecuredCookies.add((Symbol.VariableSymbol) ((IdentifierTree) assignment.variable()).symbol());
+      Symbol assignmentVariableSymbol = ((IdentifierTree) assignment.variable()).symbol();
```
m-g-sonar: that's `identifierSymbol` that you already have.

java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -138,23 +129,29 @@ public void visitNode(Tree tree) {
...
+      if (isMethodVariable
+          && isMatchedType
+          && assignmentVariableSymbol.isVariableSymbol()
```
m-g-sonar: You already tested that with `isMethodVariable` .

PR #2072, opened 2018-06-19
SONARJAVA-2800 Fix semantic + SONARJAVA-2758 Rule S4601: "HttpSecurity" URL patterns should be correctly ordered

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/spring/SpringAntMatcherOrderCheck.java
```diff
@@ -0,0 +1,185 @@
...
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S4601")
+public class SpringAntMatcherOrderCheck extends AbstractMethodDetection {
+
+  private static final Pattern MATCHER_SPECIAL_CHAR = Pattern.compile("[?*{]");
+
+  private static final MethodMatcher AUTHORIZE_REQUESTS_MATCHER = MethodMatcher.create()
+    .typeDefinition(TypeCriteria.anyType())
```
m-g-sonar: The typeDefinition should be `"org.springframework.security.config.annotation.web.builders.HttpSecurity"`.

The class is `final`, so it can not be a subtype, and only defined in that type.
Note that you will probably need to add a dependency over spring-security:

```
    <artifactItem>
      <groupId>org.springframework.security</groupId>
      <artifactId>spring-security-config</artifactId>
      <version>5.0.6.RELEASE</version>
    </artifactItem>
```

java-checks/src/main/java/org/sonar/java/checks/spring/SpringAntMatcherOrderCheck.java
```diff
@@ -0,0 +1,185 @@
...
+    .name("authorizeRequests")
+    .withAnyParameters();
+
+  private static final List<MethodMatcher> METHOD_MATCHERS = Collections.singletonList(
```
m-g-sonar: I would call this `ANT_MATCHERS`

java-checks/src/main/java/org/sonar/java/checks/spring/SpringAntMatcherOrderCheck.java
```diff
@@ -0,0 +1,185 @@
...
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return METHOD_MATCHERS;
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree method) {
+    List<StringConstant> patterns = antMatchersPatterns(method);
+    if (!patterns.isEmpty()) {
+      List<StringConstant> previousPatterns = new PreviousMethodDetection().scan(method).previousPatterns;
```
m-g-sonar: I would make the `scan` method static, in order to not have to create a new instance of `PreviousMethodDetection`. This can be hidden in the method. You could also directly return the `previousPatterns` rather than the instance.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringAntMatcherOrderCheck.java
```diff
@@ -0,0 +1,185 @@
...
+      return null;
+    }
+
+  }
+
+  private static List<StringConstant> antMatchersPatterns(MethodInvocationTree mit) {
+    List<StringConstant> argumentsAsString = mit.arguments().stream()
+      .map(StringConstant::of)
+      .collect(Collectors.toList());
+    return argumentsAsString.stream().allMatch(Objects::nonNull) ? argumentsAsString : Collections.emptyList();
```
m-g-sonar: Why only considering the cases having only literals?

I think we should also consider the following cases, as we already know they are wrong, even without considering `myVar`:
```java
  http.authorizeRequests().antMatchers("/admin/**", myVar, "/admin/login"); // Noncompliant
```

Therefore, I would simply skip the null values.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringAntMatcherOrderCheck.java
```diff
@@ -0,0 +1,185 @@
...
+    return pattern.replaceAll("([.(){}+|^$\\[\\]\\\\])", "\\\\$1");
+  }
+
+  private static class PreviousMethodDetection {
+
+    List<StringConstant> previousPatterns = new ArrayList<>();
+
+    private PreviousMethodDetection scan(MethodInvocationTree parent) {
+      MethodInvocationTree childMethod = childMethodInvocation(parent);
+      while (childMethod != null) {
```
m-g-sonar: I don't like the fact that for each method invocations, we are going to scan again and again all the previous (child) chained invocations. This could and should be computed only once.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringAntMatcherOrderCheck.java
alban-auzeill: It was a shortcut to exclude this method signature: .antMatchers(HttpMethod.GET,"/home")
And yes, in the case of antMatchers("/a", myVar, "/b"); we have some false-negative.
I'm going to check the method signature (methodSymbol.parameterTypes().get(0).is("java.lang.String")) instead, to not have false-positive in this case.

java-checks/src/test/files/checks/spring/SpringAntMatcherOrderCheck.java
```diff
@@ -0,0 +1,69 @@
...
+        .authorizeRequests()
+          .antMatchers("/level1").permitAll();
+
+    http
+      .antMatcher("/root1")
+        .authorizeRequests()
+          .antMatchers("/level1").permitAll()
+          .antMatchers("/level1").permitAll(); // Noncompliant
+
+    http
```
m-g-sonar: As discussed, I would appreciate to see a test case with more than one issue on a sequence of `antMatchers`, like in the example from the rule documentation.

alban-auzeill: I forgot this case in my "Fix from review" commit. I'm going to add a small new commit.

PR #2074, opened 2018-06-19
SONARJAVA-2592 Handle spring jdbc creators

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -473,13 +479,23 @@ public void visitMethodInvocation(MethodInvocationTree syntaxNode) {
     }
 
     private boolean methodOpeningResource(MethodInvocationTree mit) {
-      return !isWithinTryHeader(mit)
-        && !excludedByRuleOption(mit.symbolType())
+      return !excludedByRuleOption(mit.symbolType())
+        && !isWithinTryHeader(mit)
```
benzonico: Why switching the order of the short circuit ?

m-g-sonar: No reason. I'll missed that unnecessary change when opening the PR. I'll revert that

PR #2076, opened 2018-06-20
SONARJAVA-2712 Enum should be compared with ==

benzonico:


java-checks/src/test/files/checks/EnumEqualCheck.java
```diff
@@ -0,0 +1,31 @@
...
+    return candidateFruit.equals(Fruit.GRAPE); // Noncompliant {{Use "==" to perform this enum comparison instead of using "equals"}}
+  }
+  public boolean isFruitGrape(Fruit candidateFruit) {
+    return candidateFruit == Fruit.GRAPE; // Compliant; there is only one instance of Fruit.GRAPE - if candidateFruit is a GRAPE it will have the same reference as Fruit.GRAPE
+  }
+
+  public boolean isFruitGrape(Cake candidateFruit) {
+    return candidateFruit == Fruit.GRAPE; // Compliant: compilation time failure
+  }
+}
```
andrei-epure-sonarsource: nitpick: could have more tests cases

```
Cake cakeField;

public Cake getCake() {}

public boolean checkCake() {
  return getCake().equals(Fruit.GRAPE); // Noncompliant
}

public boolean checkField() {
  return this.cakeField.equals("");  // Noncompliant
}
```

benzonico: This would be testing expression type not the rule. We already have some tests of semantic.

PR #2078, opened 2018-06-20
SONARJAVA-2752 Rule S2229: Methods should not call same-class methods with incompatible "@Transactional" values

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -0,0 +1,196 @@
...
+      if (values != null) {
+        return getAnnotationAttributeAsString(values, "value", defaultValue);
+      } else {
+        return null;
+      }
+    }
+  }
+
+  private static String getAnnotationAttributeAsString(List<AnnotationValue> values, String attributeName, String defaultValue) {
+    Optional<Object> propagation = values.stream()
```
benzonico: This can be rewritten to have only a reduction of Optional and a orElse

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -0,0 +1,196 @@
...
+
+  private static boolean hasSameValues(Map<Symbol, String> methodsPropagationMap) {
+    if (methodsPropagationMap.size() < 2) {
+      return true;
+    }
+    String firstValue = methodsPropagationMap.values().iterator().next();
+    return methodsPropagationMap.values().stream().allMatch(value -> Objects.equals(value, firstValue));
+  }
+
+  private static boolean hasMethodsAnnotatedWithTransactional(ClassTree classTree) {
```
benzonico: This will also match fields annotated with the two annotation specified, no idea if you did that on purpose or not, but there is a mismatch between what the method is doing and its name.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -0,0 +1,196 @@
...
+      if (member.is(Tree.Kind.METHOD)) {
+        MethodTree method = (MethodTree) member;
+        methodPropagationMap.put(method.symbol(), getPropagation(method.symbol(), classPropagation));
+      }
+    }
+    return methodPropagationMap;
+  }
+
+  private static boolean hasSameValues(Map<Symbol, String> methodsPropagationMap) {
+    if (methodsPropagationMap.size() < 2) {
```
benzonico: This is a bit useless and premature optimization.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -0,0 +1,196 @@
...
+    return Collections.singletonList(Tree.Kind.CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
+      return;
+    }
+    ClassTree classTree = (ClassTree) tree;
+    if (!hasMethodsAnnotatedWithTransactional(classTree)) {
```
benzonico: This check is not really useful as the loop over the methods of the class is going to be done afterwards. So this can be dropped and the method can be deleted.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -0,0 +1,196 @@
...
+
+  private static boolean hasMethodsAnnotatedWithTransactional(ClassTree classTree) {
+    return classTree.symbol().memberSymbols().stream()
+      .map(Symbol::metadata)
+      .anyMatch(metadata -> metadata.isAnnotatedWith(SPRING_TRANSACTIONAL_ANNOTATION) ||
+        metadata.isAnnotatedWith(JAVAX_TRANSACTIONAL_ANNOTATION));
+  }
+
+  @CheckForNull
+  private static String getPropagation(Symbol symbol, @Nullable String inheritedPropagation) {
```
benzonico: What about inheritance of annotation through class hierarchy ? how does it work in spring and do we support this or not ? please add a testcase to show the support of it or not.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -0,0 +1,196 @@
...
+    for (Tree member : classTree.members()) {
+      if (member.is(Tree.Kind.METHOD)) {
+        MethodTree method = (MethodTree) member;
+        methodPropagationMap.put(method.symbol(), getPropagation(method.symbol(), classPropagation));
+      }
+    }
+    return methodPropagationMap;
+  }
+
+  private static boolean hasSameValues(Map<Symbol, String> methodsPropagationMap) {
```
benzonico: IMO this should be done this way : https://stackoverflow.com/questions/22412078/check-if-all-values-in-a-map-are-equal
I think this method should accept a list of String and not the whole map.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -0,0 +1,196 @@
...
+    });
+  }
+
+  private static boolean methodInvocationOnThisInstance(MethodInvocationTree methodInvocation) {
+    ExpressionTree expression = methodInvocation.methodSelect();
+    if (expression.is(Tree.Kind.MEMBER_SELECT)) {
+      expression = ((MemberSelectExpressionTree) expression).expression();
+      return expression.is(Tree.Kind.IDENTIFIER) && ((IdentifierTree) expression).name().equals("this");
+    }
+    return expression.is(Tree.Kind.IDENTIFIER);
```
benzonico: If the method is static and/or statically imported this will return true. This is a discrepancy between name and implementation.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -0,0 +1,196 @@
...
+
+  private void checkMethodInvocation(MethodTree method, @Nullable String callerPropagation, Map<Symbol, String> methodsPropagationMap) {
+    BlockTree methodBody = method.block();
+    if (methodBody == null) {
+      return;
+    }
+    methodBody.accept(new BaseTreeVisitor() {
+      @Override
+      public void visitMethodInvocation(MethodInvocationTree methodInvocation) {
+        super.visitMethodInvocation(methodInvocation);
```
benzonico: If i'm correct you have no testcase that requires this line (which should be here). In other words, if I remove that line, no tests will fail : please add a case (a method invocation with another method invocation as argument for instance).

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+    Map<Symbol, String> methodsPropagationMap = collectMethodsPropagation(classTree);
+    if (hasSameValues(methodsPropagationMap)) {
+      return;
+    }
+    methodsPropagationMap
+      .forEach((symbol, propagation) -> checkMethodInvocation((MethodTree) symbol.declaration(), propagation, methodsPropagationMap));
+  }
+
+  private void checkMethodInvocation(MethodTree method, @Nullable String callerPropagation, Map<Symbol, String> methodsPropagationMap) {
```
benzonico: naming : checkMethodInvocation**s**

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
alban-auzeill: The method was just optimization, I removed it.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
alban-auzeill: removed

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
alban-auzeill: now it accepts a list of string

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
alban-auzeill: I added a check about "symbol().isStatic()"

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
alban-auzeill: I added a test case

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
alban-auzeill: I added a test case to show that we don't support class hierarchy

PR #2080, opened 2018-06-21
SONARJAVA-2760 Rule S4603: "@Import"s should be preferred to "@ComponentScan"s

alban-auzeill:


java-checks/src/test/files/checks/spring/SpringComponentScanCheck.java
```diff
@@ -0,0 +1,28 @@
...
+
+@ComponentScan(basePackages = "com.myapp") // Noncompliant [[sc=2;ec=15]] {{Consider replacing "@ComponentScan" by a list of beans imported with @Import to speed-up the start-up of the application.}}
+public class MyFirstApp {
+}
+
+@SpringBootApplication(scanBasePackages = "com.myapp") // Noncompliant
+class MySecondApp {
+}
+
+@SpringBootApplication // Noncompliant
```
andrei-epure-sonarsource: should add a message with `Consider replacing "@SpringBootApplication" ...` to test it gets constructed correctly

PR #2081, opened 2018-06-22
SONARJAVA-2754 RSPEC-3439 "DefaultMessageListenerContainer" instances should not drop messages during restarts

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/xml/spring/DefaultMessageListenerContainerCheck.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Override
+  public void scanFileWithXPathExpressions(XmlCheckContext context) {
+    StreamSupport.stream(context.evaluateOnDocument(defaultMessageListenerContainerBeanExpression).spliterator(), false)
+      .filter(bean -> !hasAcceptMessagePropertyEnabled(context, bean) && hasSessionTransactedDisabled(context, bean))
+      .forEach(bean -> reportIssue(bean, "Enable \"acceptMessagesWhileStopping\"."));
+  }
+
+  private boolean hasAcceptMessagePropertyEnabled(XmlCheckContext context, Node bean) {
+    return hasAttributeValue(bean, "p:acceptMessagesWhileStopping")
```
alban-auzeill: Namespace prefix 'p:' should not be hard coded. This will generate some false positives if the namespace definition does not use 'p:', for example:
```xml
<beans xmlns:x="http://www.springframework.org/schema/p">
    <bean class="org.springframework.jms.listener.DefaultMessageListenerContainer"
        x:acceptMessagesWhileStopping="true"></bean>
</beans>
```

java-checks/src/main/java/org/sonar/java/checks/xml/spring/DefaultMessageListenerContainerCheck.java
```diff
@@ -0,0 +1,77 @@
...
+    return !hasAttributeValue(bean, "sessionTransacted")
+      && !hasPropertyAsChild(context, bean, sessionTransactedPropertyExpression);
+  }
+
+  private static boolean hasAttributeValue(Node bean, String attributeName) {
+    NamedNodeMap attributes = bean.getAttributes();
+    return IntStream.range(0, attributes.getLength())
+      .mapToObj(attributes::item)
+      // ignore namespace
+      .anyMatch(attribute -> attribute.getNodeName().endsWith(attributeName) && "true".equals(attribute.getNodeValue()));
```
alban-auzeill: My understanding of org.w3c.dom.Node API is for the following attribute 'acceptMessagesWhileStopping':
```xml
<beans xmlns:x="http://www.springframework.org/schema/p">
    <bean class="org.springframework.jms.listener.DefaultMessageListenerContainer"
        x:acceptMessagesWhileStopping="true"></bean>
</beans>
```
```
attribute.getPrefix()       -> "x"
attribute.getLocalName()    -> "acceptMessagesWhileStopping"
attribute.getNodeName()     -> "x:acceptMessagesWhileStopping"
attribute.getNamespaceURI() -> "http://www.springframework.org/schema/p"
```
So you should replace `attribute.getNodeName().endsWith(attributeName)` by `attributeName.equals(attribute.getLocalName())`

alban-auzeill: Ok, let's keep the "endsWith" workaround, the risk to have FP/FN is near zero.

PR #2082, opened 2018-06-26
 SONARJAVA-2755 Rule S3753: cleanup "@SessionAttributes" at end of session

andrei-epure-sonarsource:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S3753_java.html
```diff
@@ -0,0 +1,39 @@
+<p>A Spring <code>@Controller</code> that uses <code>@SessionAttributes</code> is designed to handle a stateful / multi-post form. Such
+<code>@Controller</code>s use the specified <code>@SessionAttributes</code> to store data on the server between requests. That data should be cleaned
+up when the session is over, but unless <code>setComplete()</code> is called on the <code>SessionStatus</code> object from a
+<code>@RequestMapping(method = POST)</code> method, neither Spring nor the JVM will know it's time to do that. Note that the
```
alban-auzeill: I have removed "(method = POST)" from the RSPEC, because SessionStatus can also be used in a GET request. I added a comment to the RSPEC to explain my view about this rule implementation.

java-checks/src/main/java/org/sonar/java/checks/spring/ControllerWithSessionAttributesCheck.java
```diff
@@ -0,0 +1,105 @@
...
+
+    ClassTree classTree = (ClassTree) tree;
+    SymbolMetadata classMetadata = classTree.symbol().metadata();
+    Optional<AnnotationTree> sessionAttributesAnnotation = classTree.modifiers().annotations()
+      .stream()
+      .filter(a -> a.annotationType().symbolType().fullyQualifiedName().equals("org.springframework.web.bind.annotation.SessionAttributes"))
+      .findFirst();
+    if (classMetadata.isAnnotatedWith("org.springframework.stereotype.Controller")
+        && sessionAttributesAnnotation.isPresent()
+        && classTree.members().stream().noneMatch(ControllerWithSessionAttributesCheck::methodCompletesSessionStatus)) {
```
alban-auzeill: Applying directly MethodInvocationVisitor to the "classTree" itself, will not only visit methods, but also field initializers (and field can be a lambda calling "setComplete()")

java-checks/src/main/java/org/sonar/java/checks/spring/ControllerWithSessionAttributesCheck.java
```diff
@@ -0,0 +1,105 @@
...
+      return annotation.symbolType().is("org.springframework.web.bind.annotation.PostMapping");
+    }
+  }
+
+  private static class MethodInvocationVisitor extends BaseTreeVisitor {
+    boolean found;
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree tree) {
+      if (tree.symbol().toString().equals("SessionStatus#setComplete()")) {
```
alban-auzeill: It's a fragile way to test "setComplete()", you should use:
SET_COMPLETE_MATCHER = MethodMatcher.create().typeDefinition("org.springframework.web.bind.support.SessionStatus").name("setComplete").withoutParameter());
and then:
SET_COMPLETE_MATCHER.matches(tree)

PR #2095, opened 2018-07-02
SONARJAVA-2815 Bump SQ API to 7.3 and fix removed APIs

saberduck:
*DO_NOT_MERGE* needs to be rebased on master once SONARJAVA-2816 is merged


saberduck:
I decided to integrate changes from this PR in feature/security_hotspots PR , because they don't make much sense to be separate

PR #2097, opened 2018-07-03
SONARSEC-103 Add target object as first argument to method call in uCFG

andrei-epure-sonarsource:


java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -335,19 +337,23 @@ private LocationInFile location(SyntaxToken firstToken, SyntaxToken lastToken) {
...
-    return isString(var.type().symbolType());
+    return isObject(var.type().symbolType());
   }
 
   private static boolean isString(Type type) {
     return type.is("java.lang.String");
   }
 
+  private static boolean isObject(Type type) {
+    return type.isSubtypeOf("java.lang.Object");
```
benzonico: probably better in this case to use the negation of the "isPrimitive" method : more efficient and closer to the definition of what we are trying to achieve.

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -362,6 +362,10 @@ private static boolean isObject(Type type) {
     return type.isSubtypeOf("java.lang.Object");
   }
 
+  private static Expression fullyQualifiedNameExpression(Type type) {
```
benzonico: Given the usage, I would simply inline this one, the method does not bring any added value.

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -245,6 +247,24 @@ public void implicit_usage_of_this() {
         "}", expectedUCFG);
   }
 
+  @Test
+  public void explicit_static_method_call() {
```
benzonico: I tend to think that testing with something else than "java.lang.String" could be better. You can define your own static method or use one of java.util.Objects maybe. Especially given the modification of the other test.

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -303,6 +281,30 @@ private void buildMethodInvocation(UCFGBuilder.BlockBuilder blockBuilder, Identi
...
+    if (element.is(PLUS) && elementIsString) {
+      BinaryExpressionTree binaryExpressionTree = (BinaryExpressionTree) element;
+      Expression lhs = idGenerator.lookupExpressionFor(binaryExpressionTree.leftOperand());
+      Expression rhs = idGenerator.lookupExpressionFor(binaryExpressionTree.rightOperand());
+      Expression.Variable var = variableWithId(idGenerator.newIdFor(binaryExpressionTree));
+      blockBuilder.assignTo(var, call("__concat").withArgs(lhs, rhs), location(element));
+    } else if (element.is(PLUS_ASSIGNMENT) && elementIsString) {
+      Expression var = idGenerator.lookupExpressionFor(((AssignmentExpressionTree) element).variable());
+      Expression expr = idGenerator.lookupExpressionFor(((AssignmentExpressionTree) element).expression());
+      if (!var.isConstant()) {
```
benzonico: this condition was used to safely cast to Variable, but with the upgrade to latest version of protobuf this is not enough to guarantee that var will be an Expression.Variable.

andrei-epure-sonarsource: I don't know how to cover that with tests, though

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -356,6 +356,10 @@ private static boolean isObject(Type type) {
     return type.isSubtypeOf("java.lang.Object");
   }
 
+  private static boolean isNonConstantVariable(Expression var) {
+    return var instanceof Expression.Variable && !var.isConstant();
```
benzonico: right operand is useless in this case.

PR #2099, opened 2018-07-04
Community issue 731: Improve documentation for rule RSPEC-2168

Serik0:
Problem is described here:
   https://community.sonarsource.com/t/volatile-keyword-solves-rspec-2168/731


benzonico:
Hi, 

Thanks for the suggestion. As stated in the readme of that directory, those files are generated. Therefore this PR can't be accepted and we need to take your modification back to our repository ( https://jira.sonarsource.com/browse/RSPEC-2168 ). Let's have the conversation in your thread in discuss and i'll link your commit to it so your suggestion is not lost.

PR #2101, opened 2018-07-04
Add more UTs for UCFGJavaVisitor

andrei-epure-sonarsource:


andrei-epure-sonarsource:
I closed the PR because the tests either became irrelevant, or were covered by other changes

PR #2103, opened 2018-07-05
SONARSEC-123 Translate method arguments to simple expressions

andrei-epure-sonarsource:
Add UT for already handled case.

### DIFF
--- a/java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
+++ b/java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
@@ -584,6 +584,34 @@ public class UCFGJavaVisitorTest {
         "}", expectedUCFG);
   }
 
+  @Test
+  public void method_argument_expressions_get_stored_in_auxiliary_local_variables() {
+    Expression.Variable arg1 = UCFGBuilder.variableWithId("arg1");
+    Expression.Variable arg2 = UCFGBuilder.variableWithId("arg2");
+    Expression.Variable arg3 = UCFGBuilder.variableWithId("arg3");
+    Expression.Variable aux0 = UCFGBuilder.variableWithId("%0");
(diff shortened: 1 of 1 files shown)


java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -584,6 +584,34 @@ public void constructor_call_returned() {
...
+    Expression.Variable aux2 = UCFGBuilder.variableWithId("%2");
+    Expression.Variable aux3 = UCFGBuilder.variableWithId("%3");
+
+    UCFG expectedUCFG = UCFGBuilder.createUCFGForMethod("A#method(Ljava/lang/Object;[Ljava/lang/String;LA$Foo;)Ljava/lang/String;").addMethodParam(arg1).addMethodParam(arg2).addMethodParam(arg3)
+        .addStartingBlock(newBasicBlock("1")
+            .newObject(aux0, "java.lang.Integer", new LocationInFile(FILE_KEY, 4, 35, 4, 42))
+            .assignTo(aux1, call("java.lang.Integer#<init>(I)V").withArgs(aux0, constant("\"\"")), new LocationInFile(FILE_KEY, 4,31,4,45))
+            .assignTo(aux2, call("java.lang.Object#toString()Ljava/lang/String;").withArgs(arg1), new LocationInFile(FILE_KEY, 4,47,4,62))
+            .assignTo(aux3, call("java.lang.String#format(Ljava/lang/String;[Ljava/lang/Object;)Ljava/lang/String;")
+                    .withArgs(new Expression.ClassName("java.lang.String"), constant("%s"), aux0, aux2, constant("\"\""), constant("\"\"")),
```
m-g-sonar: I think it worth adding comments for the field access saying that it is expected to change once SONARSEC-121 is handled. Maybe the same for array access., to a least mention that it is expected to change a some point

PR #2104, opened 2018-07-05
SONARJAVA-2817 Set security standards on rules when supported

saberduck:


saberduck:
QA failure is expected, because API is not yet available on DEV version of SQ

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
```diff
@@ -190,12 +202,18 @@ private static String readResource(URL resource) {
     String type;
     String[] tags;
     String defaultSeverity;
+    SecurityStandards securityStandards = new SecurityStandards();
```
vilchik-elena: do we need this initializer?

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
```diff
@@ -169,6 +170,17 @@ private void addMetadata(NewRule rule, @Nullable RuleMetadata metadata) {
...
+    if (supportsSecurityHotspots) {
+      addSecurityStandards(rule, metadata.securityStandards);
+    }
+  }
+
+  private static void addSecurityStandards(NewRule rule, SecurityStandards securityStandards) {
+    for (String s : securityStandards.OWASP) {
+      rule.addOwaspTop10(OwaspTop10.valueOf(s));
+    }
+    int[] cweNumbers = Arrays.stream(securityStandards.CWE).map(cwe -> cwe.substring(4)).mapToInt(Integer::valueOf).toArray();
```
vilchik-elena: hm... do we want change API? to accept full names?

saberduck: this was discussed and decision was that they prefer to have int because it's simpler for them

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
saberduck: yes, otherwise if the field is not present it will fail on NPE

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
vilchik-elena: should rule api then produce numbers?

saberduck: done. rule-api was updated to use numbers

PR #2106, opened 2018-07-05
SONARJAVA-2685 Rule S4510 Xml deserialization from untrusted source s…

saberduck:
…hould be avoided


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S4510_java.json
```diff
@@ -0,0 +1,31 @@
+{
+  "title": "XML deserialization from an untrusted source should be avoided",
+  "type": "VULNERABILITY",
```
vilchik-elena: shouldn't it be hotspot? cc @agigleux

java-checks/src/main/java/org/sonar/java/checks/XmlDeserializationCheck.java
```diff
@@ -0,0 +1,82 @@
...
+@Rule(key = "S4510")
+public class XmlDeserializationCheck extends AbstractMethodDetection {
+
+  private static final MethodMatcher READ_OBJECT = MethodMatcher.create().typeDefinition("java.beans.XMLDecoder")
+    .name("readObject").withAnyParameters();
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return Collections.singletonList(
+      MethodMatcher.create().typeDefinition("java.beans.XMLDecoder").name("<init>").withAnyParameters()
```
vilchik-elena: why we extracted into constant filed one `MethodMatcher` but not this one?

java-checks/src/main/java/org/sonar/java/checks/XmlDeserializationCheck.java
```diff
@@ -0,0 +1,82 @@
...
+    List<JavaFileScannerContext.Location> secondaries = collectSecondaryLocations(newClassTree);
+    reportIssue(newClassTree.identifier(),
+      "Verify the InputStream of XMLDecoder in not corrupted because it is coming from untrusted user input.",
+      secondaries, null);
+  }
+
+  private static List<JavaFileScannerContext.Location> collectSecondaryLocations(NewClassTree newClassTree) {
+    Tree parentMethodOrClass = parentMethod(newClassTree);
+    if (parentMethodOrClass == null) {
+      return Collections.emptyList();
```
vilchik-elena: can we add test for this line?

java-checks/src/main/java/org/sonar/java/checks/XmlDeserializationCheck.java
```diff
@@ -0,0 +1,82 @@
...
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return Collections.singletonList(
+      MethodMatcher.create().typeDefinition("java.beans.XMLDecoder").name("<init>").withAnyParameters()
+    );
+  }
+
+  @Override
+  protected void onConstructorFound(NewClassTree newClassTree) {
+    List<JavaFileScannerContext.Location> secondaries = collectSecondaryLocations(newClassTree);
```
vilchik-elena: IMO it would be nice to show in UT limitations of this approach (i.e. having 2 decoders in same scope, or any other way to have `readObject` called not on the reported object)

saberduck: test case added

java-checks/src/main/java/org/sonar/java/checks/XmlDeserializationCheck.java
vilchik-elena:
```  
enum A {
    foo(new XMLDecoder(new ByteArrayInputStream("foo".getBytes(StandardCharsets.UTF_8))));


    A(XMLDecoder xmlDecoder) {
      
    }
  }

```

saberduck: ok, done

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/Sonar_way_profile.json
```diff
@@ -320,6 +320,7 @@
     "S4499",
     "S4502",
     "S4507",
+    "S4510",
```
vilchik-elena: as discussed should be deactivated until hotspots are supported by plugin

PR #2107, opened 2018-07-09
SONARJAVA-2729 Rule S4544 Unsafe Jackson deserialization

saberduck:


java-checks/src/test/files/checks/JacksonDeserialization.java
```diff
@@ -0,0 +1,58 @@
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import static com.fasterxml.jackson.annotation.JsonTypeInfo.Id.MINIMAL_CLASS;
+
+class JacksonDeserialization {
+
+  public void enableDefaultTyping() {
+    ObjectMapper mapper = new ObjectMapper();
+    mapper.enableDefaultTyping(); // Noncompliant {{Consider using @JsonTypeInfo instead of enabling polymorphic type handling globally}}
```
andrei-epure-sonarsource: nitpick: add [[sc=..;ec=..]]

java-checks/src/test/files/checks/JacksonDeserialization.java
```diff
@@ -0,0 +1,58 @@
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import static com.fasterxml.jackson.annotation.JsonTypeInfo.Id.MINIMAL_CLASS;
+
+class JacksonDeserialization {
```
andrei-epure-sonarsource: you only have annotated classes in this test - please add tests for annotated fields as well

java-checks/src/main/java/org/sonar/java/checks/JacksonDeserializationCheck.java
```diff
@@ -0,0 +1,124 @@
...
+    if (!isJsonTypeId(tree)) {
+      return false;
+    }
+    String valueName;
+    if (tree.is(Tree.Kind.MEMBER_SELECT)) {
+      valueName = ((MemberSelectExpressionTree) tree).identifier().name();
+    } else {
+      valueName = ((IdentifierTree) tree).name();
+    }
+    for (String enumValue : value) {
```
andrei-epure-sonarsource: nitpick: if you pass `value` as a Set, you can just `return VALUE_SET.contains(valueName);` instead of for + if + two return statements

saberduck: nice suggestion, thanks!

PR #2108, opened 2018-07-09
SONARJAVA-2686 Rule S4508 Object deserialization from untrusted source

saberduck:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S4508_java.html
```diff
@@ -0,0 +1,19 @@
+<p>Object deserialization from an untrusted source can lead to unexpected code execution. ObjectInputStream doesn't provide a way to apply rules on
+its InputStream argument. Knowing that all serializable classes in the scope of the classloader will be deserialized, there is a possibility to
+execute malicious code during the deserialization phase even if, at the end, a ClassCastException will be raised.</p>
+<p>This rule raises an issue when one of the methods "readArray", "readObject", "readUnshared" is called on a "java.io.ObjectInputStream" or any
```
andrei-epure-sonarsource: please update rule to not contain "readArray"

java-checks/src/main/java/org/sonar/java/checks/ObjectDeserializationCheck.java
```diff
@@ -0,0 +1,46 @@
...
+
+@Rule(key = "S4508")
+public class ObjectDeserializationCheck extends AbstractMethodDetection {
+
+  private static final String OBJECT_INPUT_STREAM = "java.io.ObjectInputStream";
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return ImmutableList.of(
+      MethodMatcher.create().typeDefinition(OBJECT_INPUT_STREAM).name("readObject").withoutParameter(),
```
andrei-epure-sonarsource: does it make sense to check for subclasses of `ObjectInputStream`, and also look for `readObjectOverride​`?

java-checks/src/test/files/checks/ObjectDeserialization.java
```diff
@@ -0,0 +1,14 @@
+package test;
+
+import java.io.ObjectInputStream;
+
+public class ObjectDeserialization {
+
+  public void foo() {
+    InputStream untrusted = new WhatEverInputStream();
+    ObjectInputStream ois = new ObjectInputStream(untrusted);
+    Foo deserialized = (Foo) ois.readObject();  // Noncompliant {{Verify there is no risk to deserialize from this source}}
```
andrei-epure-sonarsource: nitpick: specify the highlighting coordinates for the first message

java-checks/src/main/java/org/sonar/java/checks/ObjectDeserializationCheck.java
saberduck: Subclasses will be handled by MethodMatcher automatically. I don't see why we should check for `readObjectOverride`, can you provide some reasoning?

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S4508_java.html
saberduck: Done. Thanks!

java-checks/src/main/java/org/sonar/java/checks/ObjectDeserializationCheck.java
saberduck: I added test case with subclass

PR #2110, opened 2018-07-09
SONARJAVA-2730 Rule S4529 Spring endpoints

saberduck:


java-checks/src/test/files/checks/spring/SpringEndpoints.java
```diff
@@ -0,0 +1,31 @@
+package test;
+
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.*;
+
+@Controller
+@RequestMapping(value = "/ex")
+public class SpringEndpoints {
```
benzonico: For sake of coverage shouldn't we have a method without an annotation that triggers the issue ?

java-checks/src/main/java/org/sonar/java/checks/spring/SpringEndpointsCheck.java
```diff
@@ -0,0 +1,64 @@
...
+  private static final String REQUEST_MAPPING_ANNOTATION = "org.springframework.web.bind.annotation.RequestMapping";
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.ANNOTATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    AnnotationTree annotationTree = (AnnotationTree) tree;
```
benzonico: Matter of taste, so up to you to fix it : I tend to prefer to have the explicit semantic test before relying on it. I know that in this case it works fine but making it explicit allows nicer refactor experiences in the future IMO.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringEndpointsCheck.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  private static boolean isSpringWebHandler(AnnotationTree annotationTree) {
+    Type annotationType = annotationTree.annotationType().symbolType();
+    return annotationType.is(REQUEST_MAPPING_ANNOTATION)
+        || annotationType.symbol().metadata().isAnnotatedWith(REQUEST_MAPPING_ANNOTATION);
+  }
+
+  private static Optional<MethodTree> findParentMethod(AnnotationTree annotationTree) {
```
benzonico: nitpicking : in this case I tend to think optional does not bring anything compared to a simple null check but an unecessary wrapping layer.

benzonico: And for sake of thinking : approaching this by visiting methods and their annotations would avoid to go up the tree and make the (already simple enough) algorithm a bit simpler to read.

java-checks/src/test/files/checks/spring/SpringEndpoints.java
saberduck: Did you mean "doesn't trigger the issue", or I don't understand?

java-checks/src/main/java/org/sonar/java/checks/spring/SpringEndpointsCheck.java
saberduck: Sure, added.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringEndpointsCheck.java
saberduck: I like this, visiting methods is indeed simpler

java-checks/src/test/files/checks/spring/SpringEndpoints.java
benzonico: You're absolutely right, I forgot the negation.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringEndpointsCheck.java
```diff
@@ -36,29 +35,26 @@
...
-    if (isSpringWebHandler(annotationTree)) {
-      findParentMethod(annotationTree).ifPresent(annotatedMethod ->
-          reportIssue(annotatedMethod.simpleName(), "Review this Spring request handler"));
+    if (!hasSemantic()) {
+      return;
     }
+    MethodTree methodTree = (MethodTree) tree;
+    List<AnnotationTree> annotations = methodTree.modifiers().annotations();
+    annotations.forEach(annotationTree -> {
+      if (isSpringWebHandler(annotationTree)) {
```
benzonico: filter ?

benzonico: and to be a bit clearer : I would have a stream and filter and then for each remaining report an issue.

saberduck: yeah :)

PR #2115, opened 2018-07-10
SONARSEC-158 Array variable declaration should be translated to uCFG

andrei-epure-sonarsource:


java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -255,6 +260,17 @@ private void buildCall(Tree element, UCFGBuilder.BlockBuilder blockBuilder, Iden
...
   }
 
+  private void buildNewArrayInvocation(BlockBuilder blockBuilder, IdentifierGenerator idGenerator, NewArrayTree tree) {
+    Expression.Variable newArray = variableWithId(idGenerator.newIdFor(tree));
+    blockBuilder.newObject(newArray, tree.symbolType().fullyQualifiedName(), location(tree));
+    idGenerator.varForExpression(tree, newArray.id());
+
+    for (ExpressionTree initializer : tree.initializers()) {
+      Expression argument = idGenerator.lookupExpressionFor(initializer);
+      blockBuilder.assignTo(variableWithId(idGenerator.newId()), arraySet(newArray, argument), location(tree));
```
m-g-sonar: I think this should be the location of the `initializer`, rather than the tree. WDYT?

andrei-epure-sonarsource: The `initializer` gets assigned to the array, though...

PR #2118, opened 2018-07-12
SONARSEC-165 Fix uCFG array access inconsistency with primitives

andrei-epure-sonarsource:
Instead of translating only arrayGet on primitive arrays (and not arraySet), ignore both arrayGet and arraySet.
Keep references to arrays in the uCFG.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
+++ b/java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
@@ -36,7 +36,6 @@ import org.sonar.api.utils.log.Loggers;
 import org.sonar.java.cfg.CFG;
 import org.sonar.java.cfg.VariableReadExtractor;
 import org.sonar.java.model.LiteralUtils;
-import org.sonar.java.resolve.ArrayJavaType;
 import org.sonar.java.resolve.JavaSymbol;
 import org.sonar.plugins.java.api.JavaFileScanner;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
@@ -288,16 +287,15 @@ public class UCFGJavaVisitor extends BaseTreeVisitor implements JavaFileScanner 
   }
 
   private void buildNewArrayInvocation(BlockBuilder blockBuilder, IdentifierGenerator idGenerator, NewArrayTree tree) {
-    if (tree.type() != null && !isObject(tree.type().symbolType())) {
-      return;
-    }
     Expression.Variable newArray = variableWithId(idGenerator.newIdFor(tree));
     blockBuilder.newObject(newArray, tree.symbolType().fullyQualifiedName(), location(tree));
     idGenerator.varForExpression(tree, newArray.id());
-
-    for (ExpressionTree initializer : tree.initializers()) {
-      Expression argument = idGenerator.lookupExpressionFor(initializer);
-      blockBuilder.assignTo(variableWithId(idGenerator.newId()), arraySet(newArray, argument), location(tree));
+    List<ExpressionTree> initializers = tree.initializers();
+    if (!initializers.isEmpty() && isObject(initializers.get(0).symbolType())) {
+      initializers.forEach(i -> {
+        Expression argument = idGenerator.lookupExpressionFor(i);
+        blockBuilder.assignTo(variableWithId(idGenerator.newId()), arraySet(newArray, argument), location(tree));
+      });
     }
   }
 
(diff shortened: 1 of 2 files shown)


java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -276,21 +276,26 @@ private void buildCall(Tree element, UCFGBuilder.BlockBuilder blockBuilder, Iden
...
     } else if (element.is(ARRAY_ACCESS_EXPRESSION) && !element.parent().is(PLUS_ASSIGNMENT, ASSIGNMENT)) {
       // PLUS_ASSIGNMENT and ASSIGNMENT might imply an array set, otherwise an array access is always a get
-      Expression.Variable getValue = variableWithId(idGenerator.newId());
-      Expression array = idGenerator.lookupExpressionFor(((ArrayAccessExpressionTree)element).expression());
-      blockBuilder.assignTo(getValue, arrayGet(array), location(element));
-      idGenerator.varForExpression(element, getValue.id());
+      ArrayAccessExpressionTree arrayAccessExpressionTree = (ArrayAccessExpressionTree) element;
+      if (isObject(arrayAccessExpressionTree.symbolType())) {
+        Expression.Variable getValue = variableWithId(idGenerator.newId());
+        Expression array = idGenerator.lookupExpressionFor((arrayAccessExpressionTree).expression());
```
m-g-sonar: unnecessary parentheses

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -276,21 +276,26 @@ private void buildCall(Tree element, UCFGBuilder.BlockBuilder blockBuilder, Iden
...
   private void buildNewArrayInvocation(BlockBuilder blockBuilder, IdentifierGenerator idGenerator, NewArrayTree tree) {
     Expression.Variable newArray = variableWithId(idGenerator.newIdFor(tree));
     blockBuilder.newObject(newArray, tree.symbolType().fullyQualifiedName(), location(tree));
     idGenerator.varForExpression(tree, newArray.id());
-
-    for (ExpressionTree initializer : tree.initializers()) {
-      Expression argument = idGenerator.lookupExpressionFor(initializer);
-      blockBuilder.assignTo(variableWithId(idGenerator.newId()), arraySet(newArray, argument), location(tree));
+    List<ExpressionTree> initializers = tree.initializers();
+    if (!initializers.isEmpty() && isObject(initializers.get(0).symbolType())) {
```
m-g-sonar: why checking only the first element?
```java
  void foo(String tainted) {
    Object[] array = {0, tainted};
    String query = String.format("SELECT * FROM %1 LIMIT %0", array);
    // ...
  }
```
For mixed type arrays, you may want to check every initializers

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -276,21 +276,26 @@ private void buildCall(Tree element, UCFGBuilder.BlockBuilder blockBuilder, Iden
       buildPlusAssignmentInvocation(blockBuilder, idGenerator, assignmentExpressionTree);
     } else if (element.is(ARRAY_ACCESS_EXPRESSION) && !element.parent().is(PLUS_ASSIGNMENT, ASSIGNMENT)) {
       // PLUS_ASSIGNMENT and ASSIGNMENT might imply an array set, otherwise an array access is always a get
-      Expression.Variable getValue = variableWithId(idGenerator.newId());
-      Expression array = idGenerator.lookupExpressionFor(((ArrayAccessExpressionTree)element).expression());
-      blockBuilder.assignTo(getValue, arrayGet(array), location(element));
-      idGenerator.varForExpression(element, getValue.id());
+      ArrayAccessExpressionTree arrayAccessExpressionTree = (ArrayAccessExpressionTree) element;
+      if (isObject(arrayAccessExpressionTree.symbolType())) {
```
m-g-sonar: I'm not 100% sure that this will not always return true. All arrays are not primitives. I believe that you should look at the type of the `ArrayType.elementType()` (which can be another array).

andrei-epure-sonarsource: In the current implementation, the type of an `ArrayAccessExpressionTree` is the type of the element that is accessed. It is false for the tests with `int` and `byte`

m-g-sonar: So if for instance its a 2-dimensional integer array `int[][] myArray`, the type of `myArray[0]` is going to be `int[]`, translated into `$Array`, which we still don't want.

PR #2121, opened 2018-07-13
SONARJAVA-2826 Support the import of PMD issues reports

pynicolas:


external-reports/src/main/java/org/sonar/java/externalreport/pmd/PmdXmlReportReader.java
```diff
@@ -0,0 +1,161 @@
...
+  private static String getAttributeValue(StartElement startElement, String attributeName) {
+    Attribute attribute = startElement.getAttributeByName(new QName(attributeName));
+    return attribute == null ? "" : attribute.getValue();
+  }
+
+  private static Integer getAttributeAsInt(StartElement startElement, String attributeName) {
+    return Integer.parseInt(getAttributeValue(startElement, attributeName));
+  }
+
+  private static Map<Integer, Severity> severities() {
```
vilchik-elena: isn't it better to put this code into static initializer block

external-reports/src/main/resources/org/sonar/l10n/java/rules/pmd/rules.json
```diff
@@ -0,0 +1,2306 @@
+[
+  {
+    "key": "AbstractClassWithoutAbstractMethod",
+    "name": "AbstractClassWithoutAbstractMethod",
+    "type": "CODE_SMELL",
+    "severity": "MAJOR",
```
vilchik-elena: you provide 2 severities: for rule and issue, you should either drop this one or ignore one from report

external-reports/src/main/resources/org/sonar/l10n/java/rules/pmd/rules.json
```diff
@@ -0,0 +1,2306 @@
+[
+  {
+    "key": "AbstractClassWithoutAbstractMethod",
+    "name": "AbstractClassWithoutAbstractMethod",
```
vilchik-elena: it's a not a nice name. For some other linter we transformed key, e.g. `Abstract class without abstract method`

external-reports/src/test/java/org/sonar/java/externalreport/pmd/PmdSensorTest.java
```diff
@@ -0,0 +1,226 @@
...
+  @Rule
+  public LogTester logTester = new LogTester();
+
+  @Test
+  public void test_descriptor() {
+    DefaultSensorDescriptor sensorDescriptor = new DefaultSensorDescriptor();
+    sensor.describe(sensorDescriptor);
+    assertThat(sensorDescriptor.name()).isEqualTo("Import of PMD issues");
+    MapSettings settings = new MapSettings();
+    assertThat(sensorDescriptor.configurationPredicate().test(new ConfigurationBridge(settings))).isFalse();
```
vilchik-elena: there is method `MapSettings#asConfig` to convert to config

external-reports/src/main/java/org/sonar/java/externalreport/pmd/PmdXmlReportReader.java
pynicolas: Why would it be better?

external-reports/src/test/java/org/sonar/java/externalreport/pmd/PmdSensorTest.java
pynicolas: I didn't know that, thanks!

external-reports/src/main/java/org/sonar/java/externalreport/pmd/PmdXmlReportReader.java
vilchik-elena: then it's explicit that this method is intended to be used only for initialization

pynicolas: It's not intended to be used only for initialization.

external-reports/src/main/java/org/sonar/java/externalreport/pmd/PmdXmlReportReader.java
```diff
@@ -0,0 +1,161 @@
...
+import org.sonar.api.rules.RuleType;
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
+import org.sonarsource.analyzer.commons.ExternalRuleLoader;
+
+public class PmdXmlReportReader {
+
+  private static final Logger LOG = Loggers.get(PmdXmlReportReader.class);
+
+  private static final Map<Integer,Severity> SEVERITIES = severities();
```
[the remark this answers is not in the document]
pynicolas: Changed in b56ab44

external-reports/src/main/resources/org/sonar/l10n/java/rules/pmd/rules.json
pynicolas: Changed in b56ab44

external-reports/src/main/resources/org/sonar/l10n/java/rules/pmd/rules.json
pynicolas: Changed in b56ab44. And it seems impossible to fully automate because of a few special cases...

PR #2123, opened 2018-07-13
SONARSEC-121 Translate field access to uCFG

andrei-epure-sonarsource:


java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -387,13 +396,80 @@ private void buildAssignCall(BlockBuilder blockBuilder, IdentifierGenerator idGe
...
+  private Expression.FieldAccess buildFieldAccess(BlockBuilder blockBuilder, IdentifierGenerator idGenerator, MemberSelectExpressionTree memberSelectTree) {
+    ExpressionTree lhsTree = memberSelectTree.expression();
+    Symbol rhsTreeSymbol = memberSelectTree.identifier().symbol();
+    if (!rhsTreeSymbol.isVariableSymbol()) {
+      return null;
+    }
+    Expression.Variable rightSide = variableWithId(memberSelectTree.identifier().name());
+    Expression.FieldAccess fieldAccess = null;
+    if (!rhsTreeSymbol.isStatic()) {
+      buildFieldReadAccessInvocation(blockBuilder, idGenerator, lhsTree);
```
andrei-epure-sonarsource: `leftSide` can be directly `lookupExpression(lhsTree)` instead of lines 431-432

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -86,38 +86,222 @@ public void visit_java_file() {
...
+        .addBasicBlock(newBasicBlock("1")
+            .assignTo(aux0, call("A#getA()LA;").withArgs(Expression.THIS), new LocationInFile(FILE_KEY, 4,24,4,30))
+            .assignTo(aux1, call("__id").withArgs(new Expression.FieldAccess(aux0, variableWithId("field"))), new LocationInFile(FILE_KEY, 4,24,4,36))
+            .ret(aux1, new LocationInFile(FILE_KEY, 4,17,4,37)))
+        .build();
+    assertCodeToUCfg("class A {\n" +
+        "  String field;\n" +
+        "  A getA() { return new A(); }\n" +
+        "  String foo() { return getA().field; }\n" +
+        "  String bar() { return (new A()).field; }\n" +
```
andrei-epure-sonarsource: remove bar and baz as not used

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -393,13 +402,82 @@ private void buildAssignCall(BlockBuilder blockBuilder, IdentifierGenerator idGe
...
+    Expression.Variable rightSide = variableWithId(memberSelectTree.identifier().name());
+    Expression.FieldAccess fieldAccess = null;
+    if (!rhsTreeSymbol.isStatic()) {
+      // this is a recursive call that will translate any field accesses on the left
+      // we do not call lookupExpression() because that translates array access, too
+      buildFieldReadAccessInvocation(blockBuilder, idGenerator, lhsTree);
+      Expression leftSide = idGenerator.lookupExpressionFor(lhsTree);
+      if (leftSide.equals(Expression.THIS)) {
+        fieldAccess = new Expression.FieldAccess(rightSide);
+      } else if (leftSide instanceof Expression.Variable) {
```
m-g-sonar: With latest version of SonarUCFG, you should have access to `isVariable()` on any `Expression`.

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -393,13 +402,82 @@ private void buildAssignCall(BlockBuilder blockBuilder, IdentifierGenerator idGe
...
+          new Expression.ClassName(lhsIdentifierTree.symbol().type().fullyQualifiedName()),
+          rightSide);
+    }
+    return fieldAccess;
+  }
+
+  private static Expression.FieldAccess buildFieldAccess(IdentifierTree identifierTree) {
+    if (!identifierTree.symbol().isVariableSymbol()
+        || identifierTree.symbol().owner().isMethodSymbol()
+        || identifierTree.name().equals("this")) {
```
m-g-sonar: Not sure why we are excluding this case, is it going to be caught by `lookupExpressionFor` already ?

And what about this case:
```java
class A {

  abstract class B {
    void foo() {
      A.this.myField = taint();
    }

    abstract String taint();
  }

  String myField = "foo";
}
```

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -86,38 +86,222 @@ public void visit_java_file() {
...
-      .addBasicBlock(newBasicBlock("1").ret(arg, new LocationInFile(FILE_KEY, 1,69,1,80)))
+      .addBasicBlock(newBasicBlock("1")
+          .assignTo(new Expression.FieldAccess(variableWithId("field")), call("__id").withArgs(arg), new LocationInFile(FILE_KEY, 1,51,1,67))
+          .ret(arg, new LocationInFile(FILE_KEY, 1,69,1,80)))
       .build();
     assertCodeToUCfg("class A { String field; String method(String arg) {this.field = arg; return arg;}}", expectedUCFG);
   }
 
+  @Test
+  public void method_call() {
```
m-g-sonar: way too general test name.

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -86,38 +86,222 @@ public void visit_java_file() {
...
+                call("A#method(Ljava/lang/String;)Ljava/lang/String;").withArgs(Expression.THIS, arg),
+                new LocationInFile(FILE_KEY, 1,45,1,61))
+            .ret(aux0, new LocationInFile(FILE_KEY, 1,38,1,62)))
+        .build();
+    assertCodeToUCfg("class A { String method(String arg) { return this.method(arg);}}", expectedUCFG);
+  }
+
+  @Test
+  public void field_access_on_this() {
+    Expression.Variable x = UCFGBuilder.variableWithId("x");
```
m-g-sonar: There is a static import for `variableWithId()`, not need to use `UCFGBuilder.variableWithId`.

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -86,38 +86,222 @@ public void visit_java_file() {
...
+            .assignTo(aux3,
+                call("__arrayGet").withArgs(aux2),
+                new LocationInFile(FILE_KEY, 7,4,7,18))
+            .assignTo(aux4,
+                call("__id").withArgs(new Expression.FieldAccess(a, variableWithId("foo"))),
+                new LocationInFile(FILE_KEY, 7,25,7,30))
+            .assignTo(aux5,
+                call("__id").withArgs(new Expression.FieldAccess(aux4, variableWithId("baw"))),
+                new LocationInFile(FILE_KEY, 7,25,7,34))
+            .assignTo(new Expression.FieldAccess(aux3, variableWithId("baz")),
```
m-g-sonar: This order looks strange to me. Why is this occurring after evaluating the right hand side (`RHS`)? I would have expected to have the `LHS` fully completed before evaluating the RHS.

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -748,11 +935,12 @@ public void method_argument_expressions_get_stored_in_auxiliary_local_variables(
             .assignTo(aux1, call("java.lang.Integer#<init>(I)V").withArgs(aux0, constant("\"\"")), new LocationInFile(FILE_KEY, 4,31,4,45))
             .assignTo(aux2, call("java.lang.Object#toString()Ljava/lang/String;").withArgs(argument1), new LocationInFile(FILE_KEY, 4,47,4,62))
             .assignTo(aux3, call("__arrayGet").withArgs(argument2), new LocationInFile(FILE_KEY, 4,64,4,71))
+            .assignTo(aux4, call("__id").withArgs(new Expression.FieldAccess(argument3, variableWithId("foo"))), new LocationInFile(FILE_KEY, 4,73,4,81))
             // field access will change with SONARSEC-121
```
m-g-sonar: removing the comment maybe?

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -244,11 +428,13 @@ public void build_primitive_types_to_constants() {
...
         "    float var7 = 1.0;\n" +
         "    double var8 = 1.0;\n" +
-        "    System.out.print(var5);\n" +
-        "    System.out.print(arg1);\n" +
+        "    foo(var5);\n" +
+        "    bar(arg1);\n" +
         "    String s = \"\" + var5 + var6 + var7 + var8;\n" +
         "    return s + arg1 + arg2 + arg3 + arg4;\n" +
         "  }\n" +
+        "  void foo(int i) {}\n" +
```
m-g-sonar: or simply `void foo(Object o) {}\n` to be used by both.
But I would have kept one of the chained field access.

andrei-epure-sonarsource: There's a special test with `field_access_chain`, this is on ignoring primitives. Initially it wasn't a problem because field access was ignored, but now it would pollute the test.

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
andrei-epure-sonarsource: will change on `method_call_on_this`

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -511,7 +590,11 @@ public Expression lookupExpressionFor(@Nullable Tree tree) {
     public String lookupIdFor(@Nullable Tree tree) {
       if (tree == null) {
         return CONST;
-      } else if (tree.is(IDENTIFIER)) {
+      }
+      while (tree.is(Tree.Kind.PARENTHESIZED_EXPRESSION)) {
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: I agree about parameter reassignment, but it's a very small method so it's easy to follow

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -387,13 +396,82 @@ private void buildAssignCall(BlockBuilder blockBuilder, IdentifierGenerator idGe
...
+  }
+
   private static List<Expression> argumentIds(IdentifierGenerator idGenerator, Arguments arguments) {
     return arguments.stream().map(idGenerator::lookupExpressionFor).collect(Collectors.toList());
   }
 
   /**
-   * An array access expression depends on context - it might be a get or a set.
-   * This method should be used for the contexts where it is clear that the array access (if present) is a get.
+   * This method should be used for the contexts where
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: The method itself would not fail, but the calling method would not work. For example, if this is called for `this.field = x`, `this.field` would be stored in a `%0 = __id(FieldAccess(THIS,field))`, and would not work for field write access.

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -387,13 +396,82 @@ private void buildAssignCall(BlockBuilder blockBuilder, IdentifierGenerator idGe
...
+      Expression.Variable aux = variableWithId(idGenerator.newId());
+      blockBuilder.assignTo(aux, call("__id").withArgs(fieldAccess), location(tree));
+      idGenerator.varForExpression(tree, aux.id());
+    }
+  }
+
+  private Expression.FieldAccess buildFieldAccess(BlockBuilder blockBuilder, IdentifierGenerator identifierGenerator, Tree tree) {
+    Expression.FieldAccess fieldAccess = null;
+    if (tree.is(IDENTIFIER)) {
+      fieldAccess = buildFieldAccess((IdentifierTree) tree);
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: My only problem with early return is that it masks branching complexity

java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -387,13 +396,82 @@ private void buildAssignCall(BlockBuilder blockBuilder, IdentifierGenerator idGe
...
+    } else if (tree.is(MEMBER_SELECT)) {
+      fieldAccess = buildFieldAccess(blockBuilder, identifierGenerator, (MemberSelectExpressionTree) tree);
+    }
+    return fieldAccess;
```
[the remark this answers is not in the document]
andrei-epure-sonarsource: will add a test

java-frontend/src/test/java/org/sonar/java/UCFGJavaVisitorTest.java
```diff
@@ -86,38 +86,269 @@ public void visit_java_file() {
...
+  }
+
+  @Test
+  public void field_access_on_this_from_inner_class() {
+    Expression.Variable aux0 = variableWithId("%0");
+    Expression.Variable aux1 = variableWithId("%1");
+    UCFG expectedUCFG = UCFGBuilder.createUCFGForMethod("A$B#innerFoo(Ljava/lang/String;)Ljava/lang/String;")
+        .addBasicBlock(newBasicBlock("1")
+            .assignTo(aux0,
+                // FIXME what do we expect here?
```
m-g-sonar: I think we are gonna live with this for the time being. This is quite a corner case.

PR #2126, opened 2018-07-18
 SONARJAVA-2834 Update messages in security hotspot rules

saberduck:


agigleux:
I can't accept this PR. The .htlm files for Hotspot rules are not compliant with what is inside RSPEC. A WARN was raised during the generation of the json/html files, we thought it was just a WARN but actually it's not generating the content of RSPEC.

saberduck:
@agigleux I updated the metadata with new rule-api

PR #2127, opened 2018-07-18
SONARSEC-175 UCFGJavaVisitor throws NPE for multi-dim array initialization

andrei-epure-sonarsource:


java-frontend/src/main/java/org/sonar/java/UCFGJavaVisitor.java
```diff
@@ -463,7 +463,11 @@ private static boolean isObjectVarDeclaration(Tree tree) {
...
   }
 
-  private static boolean isObject(Type type) {
+  private static boolean isObject(@Nullable Type type) {
+    if (type == null) {
+      // typeless arrays will have a null type
+      return true;
+    }
     if (type.isArray()) {
       return isObject(((ArrayJavaType)type).elementType());
```
andrei-epure-sonarsource: cast to ArrayType

PR #2131, opened 2018-07-19
SONARJAVA-2836 Handle empty javadoc

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -244,6 +244,10 @@ private static String exceptionName(TypeTree typeTree) {
     }
     String trimmedJavadoc = javadoc.trim();
     // remove start and end of Javadoc as well as stars
```
andrei-epure-sonarsource: nitpick: move this comment above line 251

PR #2133, opened 2018-07-20
MMF-1300 External rule definitions should be updated with the last versions of linters and the descriptions should only contain links

pynicolas:


agigleux:
LGTM

PR #2143, opened 2018-08-08
SONARJAVA-2683 RSPEC-3011 Accessibility should not be changed

benzonico:


java-checks/src/test/files/checks/AccessibilityChangeCheck.java
```diff
@@ -0,0 +1,8 @@
+class A {
+    void makeItPublic(String methodName) throws NoSuchMethodException {
+      this.getClass().getMethod(methodName).setAccessible(true); // Noncompliant {{Remove this accessibility update.}}
```
andrea-guarino-sonarsource: Could you add a test for `public static void setAccessible​(AccessibleObject[] array, boolean flag)`?

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S3011_java.json
```diff
@@ -0,0 +1,21 @@
+{
+  "title": "Accessibility should not be changed",
+  "type": "CODE_SMELL",
```
andrea-guarino-sonarsource: It should be a security hotspot

PR #2144, opened 2018-08-08
SONARJAVA-2820 Rule S2119: "Random" objects should be reused

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/ReuseRandomCheck.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return Collections.singletonList(MethodMatcher.create().typeDefinition("java.util.Random").name("<init>").withoutParameter());
+  }
+
+  @Override
+  protected void onConstructorFound(NewClassTree newClassTree) {
+    Optional<Symbol> variable = lookupInitializedSymbol(newClassTree);
+    if (!variable.isPresent()) {
```
benzonico: This should be replaced by : 
```
 lookupInitializedSymbol(newClassTree).map(Optional::of).orElseGet(() -> lookupAssignedSymbol(newClassTree) )
```

java-checks/src/main/java/org/sonar/java/checks/ReuseRandomCheck.java
```diff
@@ -0,0 +1,83 @@
...
+      .filter(ReuseRandomCheck::isNotConstructorOrStaticMain)
+      .ifPresent(owner -> reportIssue(newClassTree.identifier(), "Save and re-use this \"Random\"."));
+  }
+
+  private static Optional<Symbol> lookupInitializedSymbol(ExpressionTree expression) {
+    return Optional.of(expression)
+      .map(Tree::parent)
+      .filter(tree -> tree.is(Kind.VARIABLE))
+      .map(VariableTree.class::cast)
+      .map(VariableTree::simpleName)
```
benzonico: No need to map to simpleName : there is a symbol method on the variable tree. Remove this line.

java-checks/src/test/files/checks/ReuseRandomCheck.java
```diff
@@ -0,0 +1,29 @@
+import java.util.Random;
+
+public class A {
```
benzonico: You lack a case where the instantiated object is assigned to a field and/or a static field.
You also lack a case where the left expression of an assignement is not a variable : `this.a`
Moreover : what about such cases  : `a = b = c = new Random();`

java-checks/src/main/java/org/sonar/java/checks/ReuseRandomCheck.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Override
+  protected void onConstructorFound(NewClassTree newClassTree) {
+    Optional<Symbol> variable = lookupInitializedSymbol(newClassTree);
+    if (!variable.isPresent()) {
+      variable = lookupAssignedSymbol(newClassTree);
+    }
+    variable
+      .filter(Symbol::isVariableSymbol)
```
benzonico: I don't see a testcase where this can be false.

java-checks/src/test/files/checks/ReuseRandomCheck.java
```diff
@@ -0,0 +1,39 @@
...
+    Random a = new Random(); // Noncompliant [[sc=20;ec=26]] {{Save and re-use this "Random".}}
+    Random b = new Random(seed); // Compliant for Random(long seed)
+    Object c = new Object();
+
+    s = new Random();
+    f = new Random();
+    thid.f = new Random();
+
+    f = a = new Random();
+    a = b = new Random(); // false-negative, a and b are local variables, corner case limitation
```
benzonico: I dont' see why we should accept that false negative. Please modify your code so we can handle it.

java-checks/src/test/files/checks/ReuseRandomCheck.java
```diff
@@ -0,0 +1,39 @@
...
+public class A {
+
+  static Random s = new Random(); // Compliant for static field
+  Random f = new Random(); // Compliant for field
+
+  A() {
+    Random v = new Random(); // Compliant in constructor
+  }
+
+  void func(long seed, Random p) {
```
benzonico: could you please name your variable in your testcase in a way that it is understandble : field could be named field, s -> staticField, p -> param, and a b or c localVar1, localVar2, etc.

sonar-java-plugin/pom.xml
```diff
@@ -112,7 +112,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>9000000</maxsize>
+                  <maxsize>9030000</maxsize>
```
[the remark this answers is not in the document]
alban-auzeill: after a rebase on master, this new small rule exceed the maven-enforcer-plugin limit

java-checks/src/test/files/checks/ReuseRandomCheck.java
```diff
@@ -2,37 +2,43 @@
...
+    Random localVar1 = new Random(); // Noncompliant [[sc=28;ec=34]] {{Save and re-use this "Random".}}
+    Random localVar2 = new Random(seed); // Compliant for Random(long seed)
+    Object localVar3 = new Object();
 
-    s = new Random();
-    f = new Random();
-    thid.f = new Random();
+    staticField = new Random();
+    field = new Random();
+    thid.field = new Random();
```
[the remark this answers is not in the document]
alban-auzeill: yes, of course

PR #2149, opened 2018-08-08
SONARJAVA-2824 Rule S2115: Databases should be password-protected

andrea-guarino-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/EmptyDatabasePasswordCheck.java
```diff
@@ -0,0 +1,85 @@
...
+
+  private static final String MESSAGE = "Add password protection to this database.";
+  private static final int PASSWORD_ARGUMENT = 2;
+  private static final int URL_ARGUMENT = 0;
+  private static final Pattern EMPTY_PASSWORD_PATTERN = Pattern.compile(".*password\\s*=\\s*([&;].*|$)");
+
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return ImmutableList.of(
```
benzonico: Collections.singletonList is better (at least for meaning) in this case.

java-checks/src/main/java/org/sonar/java/checks/security/EmptyDatabasePasswordCheck.java
```diff
@@ -0,0 +1,85 @@
...
+    if (expression.is(Tree.Kind.IDENTIFIER)) {
+      ExpressionTree last = ReassignmentFinder.getClosestReassignmentOrDeclarationExpression(expression, ((IdentifierTree) expression).symbol());
+      return last == null ? null : getStringValue(last);
+    }
+    return ConstantUtils.resolveAsStringConstant(expression);
+  }
+
+  private void checkForUrlConnection(ExpressionTree expression) {
+    String url = getStringValue(expression);
+    if (url != null && (!url.contains("password") || EMPTY_PASSWORD_PATTERN.matcher(url).matches())) {
```
benzonico: If i'm correct you have no cases where the url is actually null.

java-checks/src/main/java/org/sonar/java/checks/security/EmptyDatabasePasswordCheck.java
```diff
@@ -0,0 +1,85 @@
...
+    return literal != null && literal.trim().isEmpty();
+  }
+
+  @Nullable
+  private static String getStringValue(ExpressionTree expression) {
+    if (expression.is(Tree.Kind.IDENTIFIER)) {
+      ExpressionTree last = ReassignmentFinder.getClosestReassignmentOrDeclarationExpression(expression, ((IdentifierTree) expression).symbol());
+      return last == null ? null : getStringValue(last);
+    }
+    return ConstantUtils.resolveAsStringConstant(expression);
```
benzonico: There should be a testcase where the expression will lead this expression to return null

java-checks/src/main/java/org/sonar/java/checks/security/EmptyDatabasePasswordCheck.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  private static boolean hasEmptyValue(ExpressionTree expression) {
+    String literal = getStringValue(expression);
+    return literal != null && literal.trim().isEmpty();
+  }
+
+  @Nullable
+  private static String getStringValue(ExpressionTree expression) {
+    if (expression.is(Tree.Kind.IDENTIFIER)) {
```
benzonico: This method won't work if a constant outside of the compilation unit is used as an identifier (with a static import for instance).
No reassignement will be found but the resolveAsStringConstant could still work.
The method should first try to resolve as constant and only then use the reassignement finder. This will be a bit tough to test.

java-checks/src/main/java/org/sonar/java/checks/security/EmptyDatabasePasswordCheck.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  private static boolean hasEmptyValue(ExpressionTree expression) {
+    String literal = getStringValue(expression);
+    return literal != null && literal.trim().isEmpty();
+  }
+
+  @Nullable
+  private static String getStringValue(ExpressionTree expression) {
+    String value = ConstantUtils.resolveAsStringConstant(expression);
```
benzonico: There is no test case where the expression can't be resolved as a constant and is NOT an identifier.

java-checks/src/main/java/org/sonar/java/checks/security/EmptyDatabasePasswordCheck.java
```diff
@@ -0,0 +1,87 @@
...
+
+  private static final String MESSAGE = "Add password protection to this database.";
+  private static final int PASSWORD_ARGUMENT = 2;
+  private static final int URL_ARGUMENT = 0;
+  private static final Pattern EMPTY_PASSWORD_PATTERN = Pattern.compile(".*password\\s*=\\s*([&;].*|$)");
+
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return Collections.singletonList(
```
benzonico: This will be recreated for every file, store it in a constant.

java-checks/src/test/java/org/sonar/java/checks/security/EmptyDatabasePasswordCheckTest.java
```diff
@@ -0,0 +1,32 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java.checks.security;
+
+import org.junit.Test;
+import org.sonar.java.checks.verifier.JavaCheckVerifier;
+
+public class EmptyDatabasePasswordCheckTest {
+  public final static String EMPTY_PASSWORD = "";
+  public final static String NON_EMPTY_PASSWORD = "foo";
```
benzonico: IMO worth documenting how those are being used with javadoc.

java-checks/src/test/files/checks/security/EmptyDatabasePasswordCheck.java
```diff
@@ -0,0 +1,48 @@
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import static org.sonar.java.checks.security.EmptyDatabasePasswordCheckTest.EMPTY_PASSWORD;
+import static org.sonar.java.checks.security.EmptyDatabasePasswordCheckTest.NON_EMPTY_PASSWORD;
+
+class S2115 {
+  void foo(Properties connectionProps) throws SQLException {
```
benzonico: There is no testing of the highlighting of the error. I think we should add one.

PR #2154, opened 2018-08-09
SONARJAVA-2856 S2092: setSecure=false detected when using identifiers

andrea-guarino-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -157,20 +160,32 @@ private void addToUnsecuredCookies(AssignmentExpressionTree assignment) {
...
         // builder method
         unsecuredSetters.add(mit);
       }
     }
   }
 
+  @Nullable
+  private static Boolean getBooleanValue(ExpressionTree expression) {
+    Boolean value = ConstantUtils.resolveAsBooleanConstant(expression);
+    if (value == null && expression.is(Tree.Kind.IDENTIFIER)) {
```
alban-auzeill: expression.is(Tree.Kind.IDENTIFIER) never fails, you could add the following test: cookie.setSecure(!value);

PR #2158, opened 2018-08-10
SONARJAVA-2243 Change type of symbol to generic type rather than raw …

benzonico:
…type

Fix type inference accordingly.


java-frontend/src/main/java/org/sonar/java/resolve/TypeSubstitution.java
```diff
@@ -28,6 +28,7 @@
 import java.util.Map;
 
 public class TypeSubstitution {
+  public boolean unchecked = false;
```
m-g-sonar: Worth adding a few word to document what it means to have a unchecked "TypeSubstitution". Your example of `List` is good enough and would be good to document this change in code.

m-g-sonar: I also agree with that issue raised by SonarQube. I would prefer to have this private, and possibly final... Only visible though a getter `isUnchecked()`. 

The only case where you need to set it to "true" is when identifying raw type or null type in `TypeInferenceSolver`. It is then never modified. I would prefer to have it explicitly in `TypeSubstitution`. 

I would propose two approaches:

* Having a dedicated method doing only that, allowing to have only `TypeSubstitution` knowing how to build a unchecked `TypeSubstitution`. Which is probably a bit more elegant.
```java
public static SubstitutionType uncheckedTypeSubstitution() {
  TypeSubstitution typeSubstitution = new TypeSubstitution();
  typeSubstitution.unchecked = true;
  return typeSubstitution;
}
```

* Or going further and reuse a constant everytime we need a unchecked TypeSubstitution:
```java
class TypeSubstitution {

  private static final UNCHECKED = new TypeSubstitution() {
    @Override
    public boolean isUnchecked() {
      return true;
    }
  }

  public boolean isUnchecked() {
    return false;
  }

  public static SubstitutionType uncheckedTypeSubstitution() {
    return UNCHECKED;
  }
}
```

java-frontend/src/main/java/org/sonar/java/resolve/FirstPass.java
```diff
@@ -248,6 +250,9 @@ public void visitClass(ClassTree tree) {
       symbol.addTypeParameter((TypeVariableJavaType) typeVariableSymbol.type);
       enterSymbol(typeParameterTree, typeVariableSymbol);
     }
+    if(!tree.typeParameters().isEmpty()) {
```
m-g-sonar: what about extracting `tree.typeParameters()` now that it is called 3 times? (lines 245, 248, 253)

java-frontend/src/main/java/org/sonar/java/resolve/TypeInferenceSolver.java
```diff
@@ -69,7 +67,7 @@ private TypeSubstitution inferTypeSubstitutionRec(MethodJavaSymbol method, List<
 
       substitution = inferTypeSubstitution(method, substitution, formalType, argType, variableArity, remainingArgTypes);
 
-      if (!method.isConstructor() && substitution.typeVariables().containsAll(method.typeVariableTypes)) {
+      if (substitution.unchecked || (!method.isConstructor() && substitution.typeVariables().containsAll(method.typeVariableTypes))) {
```
m-g-sonar: `substitution.isUnchecked()`

java-frontend/src/main/java/org/sonar/java/resolve/TypeVariableJavaType.java
```diff
@@ -19,13 +19,14 @@
...
 
+import java.util.ArrayList;
 import org.sonar.plugins.java.api.semantic.Type;
 
 import java.util.List;
 
 public class TypeVariableJavaType extends JavaType {
 
-  List<JavaType> bounds;
+  List<JavaType> bounds = new ArrayList<>();
```
m-g-sonar: Why do you need this? It seems to me that bounds are ALWAYS set beforehand. Isn't it going to make the computation of erasure fail in some case?! (`bounds.get(0)`)

its/ruling/src/test/resources/guava/squid-S2175.json
```diff
@@ -0,0 +1,10 @@
+{
+'com.google.guava:guava:src/com/google/common/cache/LocalCache.java':[
+3669,
```
m-g-sonar: aren't these issues FPs?

its/ruling/src/test/resources/guava/squid-S2325.json
```diff
@@ -1,72 +1,20 @@
 {
-'com.google.guava:guava:src/com/google/common/base/Absent.java':[
-101,
```
m-g-sonar: This should be part of another commit... In fact I don't understand how you spotted them.

I know that the `readResolve` method from serialization can not be static, and raising issues on them should be removed, what triggered the fix? Was `INSTANCE` not resolved ?!

Which make me think that we do not have a rule for unknown identifiers in the Semantic Test. We should most probably add one... WDYT?

its/ruling/src/test/resources/guava/squid-S2388.json
```diff
@@ -4,6 +4,11 @@
 1618,
 1665,
 ],
+'com.google.guava:guava:src/com/google/common/collect/AbstractBiMap.java':[
+424,
```
m-g-sonar: These are FPs to me, these calls are legit.

java-checks/src/main/java/org/sonar/java/checks/StaticMethodCheck.java
```diff
@@ -48,7 +48,10 @@
     MethodMatcher.create()
       .typeDefinition(TypeCriteria.subtypeOf(JAVA_IO_SERIALIZABLE)).name("writeObject").addParameter(TypeCriteria.subtypeOf("java.io.ObjectOutputStream")),
     MethodMatcher.create()
-      .typeDefinition(TypeCriteria.subtypeOf(JAVA_IO_SERIALIZABLE)).name("readObjectNoData").withoutParameter());
+      .typeDefinition(TypeCriteria.subtypeOf(JAVA_IO_SERIALIZABLE)).name("readObjectNoData").withoutParameter(),
+    MethodMatcher.create()
+      .typeDefinition(TypeCriteria.subtypeOf(JAVA_IO_SERIALIZABLE)).name("readResolve").withoutParameter()
```
m-g-sonar: According to Serializable javadoc, maybe we should add `writeReplace()` as well (https://docs.oracle.com/javase/10/docs/api/java/io/Serializable.html). But this really needs a proper ticket and PR, and should not be hidden in that change.

its/ruling/src/test/resources/guava/squid-S2388.json
benzonico: These calls are calling super class methods from an inner class which makes them relevant to be highlighted by the rule.

its/ruling/src/test/resources/guava/squid-S2175.json
benzonico: indeed : https://jira.sonarsource.com/browse/SONARJAVA-2863

java-frontend/src/main/java/org/sonar/java/resolve/TypeVariableJavaType.java
```diff
@@ -19,9 +19,8 @@
  */
 package org.sonar.java.resolve;
 
-import org.sonar.plugins.java.api.semantic.Type;
-
 import java.util.List;
+import org.sonar.plugins.java.api.semantic.Type;
```
m-g-sonar: useless change.

java-frontend/src/main/java/org/sonar/java/resolve/TypeSubstitutionSolver.java
```diff
@@ -230,7 +231,8 @@ private JavaType substituteInParametrizedType(ParametrizedTypeJavaType type, Typ
   }
 
   private JavaType substituteInTypeVar(TypeVariableJavaType typevar, TypeSubstitution substitution) {
-    if(typevarExplored.contains(typevar.symbol)) {
+    typevar.symbol.owner().complete();
```
m-g-sonar: worth having a comment why completion could be necessary.

PR #2160, opened 2018-08-13
SONARJAVA-2739 Exclude simple cases of loggers with resource bundles

benzonico:


java-checks/src/test/files/checks/PrintfMisuseCheck.java
```diff
@@ -177,6 +178,21 @@ void foo(Calendar c){
     slf4jLog.warn("message ", 1); // Noncompliant {{String contains no format specifiers.}}
     slf4jLog.warn("message {}", 1);
     slf4jLog.warn("message");
+
+    java.util.logging.Logger logger2 = java.util.logging.Logger.getLogger("som.foo", "som.foo.resources.i18n.LogMessages");
```
andrea-guarino-sonarsource: If I'm not wrong, this case would still raise an FP:
```
java.util.logging.Logger logger2;
logger2 = java.util.logging.Logger.getLogger("som.foo", "som.foo.resources.i18n.LogMessages");
logger2.log(java.util.logging.Level.WARNING, "som.foo.errorcode", 404);
```

java-checks/src/main/java/org/sonar/java/checks/PrintfMisuseCheck.java
```diff
@@ -67,6 +76,30 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
...
+      return false;
+    }
+    if (id.is(Tree.Kind.MEMBER_SELECT)) {
+      id = ((MemberSelectExpressionTree) id).identifier();
+    }
+    if (id.is(Tree.Kind.IDENTIFIER)) {
+      Tree decl = ((IdentifierTree) id).symbol().declaration();
+      if (decl != null && decl.is(Tree.Kind.VARIABLE)) {
+        VariableTree var = ((VariableTree) decl);
+        ExpressionTree init = var.initializer();
```
andrea-guarino-sonarsource: Should we check for reassignments as well? See my comments on tests below.

java-checks/src/test/files/checks/PrintfMisuseCheck.java
benzonico: You are right but I would not care too much. We are talking about loggers here, reassigning them seems to me a corner case.

andrea-guarino-sonarsource: Agreed, LGTM then.

PR #2163, opened 2018-08-13
SONARJAVA-2772 S3330 should not raise an issue when Cookie is named 'XSRF-TOKEN'

andrea-guarino-sonarsource:


java-checks/src/test/files/checks/security/CookieHttpOnlyCheck.java
```diff
@@ -145,10 +146,10 @@ SimpleCookie getC4() {
   void playFw() {
     play.mvc.Http.Cookie c1 = new play.mvc.Http.Cookie("1", "2", 3, "4", "5", true, false); // Noncompliant
     play.mvc.Http.Cookie c2 = new play.mvc.Http.Cookie("1", "2", 3, "4", "5", true, true);
-    CookieBuilder cb1 = Cookie.builder("1", "2");
+    play.mvc.Http.CookieBuilder cb1 = play.mvc.Http.Cookie.builder("1", "2");
     cb1.withHttpOnly(false); // Noncompliant
     cb1.withHttpOnly(true); // is ignored, so above is a FN
-    CookieBuilder cb2 = Cookie.builder("1", "2");
+    play.mvc.Http.CookieBuilder cb2 = play.mvc.Http.Cookie.builder("1", "2");
```
alban-auzeill: `Cookie.builder` is the only possible method invocation found in a variable or assignment initialiser, that why `PLAY_COOKIE_BUILDER.matches(mit)` is always true according to the coverage.
Add another method invocation in the unit test to improve the coverage.

java-checks/src/main/java/org/sonar/java/checks/security/CookieHttpOnlyCheck.java
```diff
@@ -171,20 +172,42 @@ private void checkVariableDeclaration(VariableTree declaration) {
...
       }
     }
   }
 
+  @CheckForNull
+  private static VariableSymbol getVariableSymbol(AssignmentExpressionTree assignment) {
+    VariableSymbol variableSymbol = null;
+    if (assignment.variable().is(Tree.Kind.IDENTIFIER)) {
+      Symbol reference = ((IdentifierTree) assignment.variable()).symbol();
+      if (reference instanceof VariableSymbol) {
```
alban-auzeill: replace everywhere "symbol instanceof VariableSymbol" by "symbol.isVariableSymbol()"

java-checks/src/main/java/org/sonar/java/checks/security/CookieHttpOnlyCheck.java
```diff
@@ -171,20 +172,42 @@ private void checkVariableDeclaration(VariableTree declaration) {
...
+        variableSymbol = (VariableSymbol) ((IdentifierTree) assignment.variable()).symbol();
```
alban-auzeill: Why not just:  variableSymbol = (VariableSymbol) reference;

PR #2167, opened 2018-08-14
Fix test which was sensitive to CRLF line endings

pynicolas:


external-reports/src/test/java/org/sonar/java/externalreport/PmdSensorTest.java
```diff
@@ -132,8 +133,10 @@ public void skip_issue_on_invalid_priority() throws IOException {
   public void invalid_text_range() throws IOException {
     List<ExternalIssue> externalIssues = execute(SQ72, "invalid-text-range.xml");
     assertThat(externalIssues).hasSize(2);
-    assertThat(externalIssues.get(1).primaryLocation().textRange()).isEqualTo(
-      new DefaultTextRange(new DefaultTextPointer(4, 0), new DefaultTextPointer(4, 14)));
```
andrei-epure-sonarsource: I would suggest to replace the other usages of `DefaultTextPointer` from lines 161 and 171, as well

pynicolas: Why?

andrei-epure-sonarsource:
Ok, I tested locally and it's fixed.

PR #2168, opened 2018-08-14
SONARJAVA-2864 Fix nullness annotations from Spring when jsr305 binary is missing

alban-auzeill:


java-frontend/src/test/java/org/sonar/java/se/JavaCheckVerifier.java
```diff
@@ -177,16 +177,20 @@ public static void verifyIssueOnFile(String filename, String message, JavaFileSc
     javaCheckVerifier.scanFile(filename, new JavaFileScanner[] {check});
   }
 
-  public void scanFile(String filename, JavaFileScanner[] checks) {
+  public static Collection<File> classpathFromDirectory(String jarsDirectory) {
```
benzonico: there's no need to change this class : your test should include only the required dependency.

java-frontend/src/test/java/org/sonar/java/se/checks/ConditionAlwaysTrueOrFalseCheckTest.java
```diff
@@ -34,6 +36,14 @@ public void test() {
     JavaCheckVerifier.verify("src/test/files/se/ConditionAlwaysTrueOrFalseCheck.java", new ConditionalUnreachableCodeCheck(), new BooleanGratuitousExpressionsCheck());
   }
 
+  @Test
+  public void test_without_jsr305() {
+    Collection<File> classpath = JavaCheckVerifier.classpathFromDirectory("target/test-jars").stream()
+      .filter(file -> !(file.getName().startsWith("jsr305-") || file.getName().startsWith("annotations-")))
+      .collect(Collectors.toList());
```
benzonico: This test is not very robust : you should only include the class you need to have your test case happening.

java-frontend/src/test/files/se/ConditionAlwaysTrueOrFalseWithoutJSR305.java
```diff
@@ -0,0 +1,25 @@
+package my.app;
+
+import org.springframework.web.context.request.RequestAttributes;
```
benzonico: Name of the file is a bit dodgy : this is only testing spring annotations via the rule and not the rule : I would not use the ConditionAlwaysTrueOrFalse in the name of the file.

java-frontend/src/test/files/se/ConditionAlwaysTrueOrFalseCheck.java
```diff
@@ -1325,6 +1325,18 @@ private Object nullableMethod() {
...
 
+  private void annotationWithNullableAndNonnullReturn() {
+    Object foo = annotationWithNullableAndNonnull();
+    if(foo == null) { // compliant, nullable wins
+    }
+  }
+
+  @javax.annotation.Nullable
+  @javax.annotation.Nonnull
+  private Object annotationWithNullableAndNonnull() {
```
benzonico: With the other testcase is that one still relevant ?

alban-auzeill: Yes. It shows explicitly our decision when we face a Nullable/Nonnull conflict.

java-frontend/src/test/files/se/ConditionAlwaysTrueOrFalseWithoutJSR305.java
alban-auzeill: renamed

PR #2169, opened 2018-08-14
SONARJAVA-2841 S2755 should consider "javax.xml.validation.Validator" and "javax.xml.validation.SchemaFactory"

andrea-guarino-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/security/XmlExternalEntityProcessingCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/security/XmlExternalEntityProcessingCheck.java
@@ -28,6 +28,8 @@ import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.SAXParserFactory;
 import javax.xml.stream.XMLInputFactory;
+import javax.xml.validation.SchemaFactory;
+import javax.xml.validation.Validator;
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/security/XmlExternalEntityProcessingCheck.java
```diff
@@ -183,7 +196,35 @@ private static MethodMatcher setFeatureMethodMatcher(String className) {
...
+    private AccessExternalDTDOrSchemaPredicate(String className) {
+      this.methodMatcher = setPropertyMethodMatcher(className);
+    }
+
+    @Override
+    public boolean test(MethodInvocationTree methodInvocation) {
+      if (methodMatcher.matches(methodInvocation)) {
+        Arguments arguments = methodInvocation.arguments();
+        String propertyName = resolveAsStringConstant(arguments.get(0));
+        String propertyValue = resolveAsStringConstant(arguments.get(1));
```
benzonico: If you use this resolveStringAsConstant I tend to think there should be testcase with the inlined value of the property (ie not using the constant).

andrea-guarino-sonarsource: good point, I will add a test case for that.

PR #2175, opened 2018-08-15
SONARJAVA-2858 FN on S1607 using assumeTrue(false)

benzonico:


java-checks/src/main/java/org/sonar/java/checks/IgnoredTestsCheck.java
```diff
@@ -43,5 +60,22 @@ public void visitNode(Tree tree) {
     if (ignoreAnnotationValues != null && ignoreAnnotationValues.isEmpty()) {
       reportIssue(methodTree.simpleName(), "Fix or remove this skipped unit test");
     }
+    BlockTree block = methodTree.block();
+    if(block != null) {
```
alban-auzeill: Here we only visit Tree.Kind.METHOD and not Tree.Kind.CONSTRUCTOR, so we have this false-negative:
```
class MyTest {
  MyTest() {
    Assume.assumeTrue(false); // false-negative
  }
  // ... tests
}
```

java-checks/src/test/files/checks/IgnoredTestsCheck.java
```diff
@@ -12,4 +12,28 @@ void qix() {}
...
+
+  void assume1() {
+    Assume.assumeTrue(false); // Noncompliant
+  }
+  void assume2() {
+    Assume.assumeFalse(true); // Noncompliant
+  }
+  void assume3() {
+    if(something) {
+      Assume.assumeFalse(true); // compliant, we only detect basic cases
```
alban-auzeill: Why not subscribing to all METHOD_INVOCATION to cover all cases? Is there a valid use case where a logic containing "assumeFalse(true)" makes sense?

java-checks/src/main/java/org/sonar/java/checks/IgnoredTestsCheck.java
```diff
@@ -20,17 +20,34 @@
...
 public class IgnoredTestsCheck extends IssuableSubscriptionVisitor {
 
+  private static final String ORG_JUNIT_ASSUME = "org.junit.Assume";
+  private static final String BOOLEAN_TYPE = "boolean";
+
+  private static final MethodMatcherCollection ASSUME_METHODS = MethodMatcherCollection.create(
+    MethodMatcher.create().typeDefinition(ORG_JUNIT_ASSUME).name("assumeTrue").parameters(BOOLEAN_TYPE),
+    MethodMatcher.create().typeDefinition(ORG_JUNIT_ASSUME).name("assumeTrue").parameters("java.lang.String", BOOLEAN_TYPE),
+    MethodMatcher.create().typeDefinition(ORG_JUNIT_ASSUME).name("assumeFalse").parameters(BOOLEAN_TYPE),
+    MethodMatcher.create().typeDefinition(ORG_JUNIT_ASSUME).name("assumeFalse").parameters("java.lang.String", BOOLEAN_TYPE)
```
alban-auzeill: 1) there's no test of assumeTrue/assumeFalse with a description
2) consistency: If `@Ignore` with a description is compliant, assumeTrue/assumeFalse with a description should be compliant, so I would remove the method matchers with 2 arguments

java-checks/src/test/files/checks/IgnoredTestsCheck.java
benzonico: Being conservative here : I want to highlight "simple case" and not raise issue if there is convoluted logic leading to not executing the test.

java-checks/src/main/java/org/sonar/java/checks/IgnoredTestsCheck.java
benzonico: Agreed with second point, i remove those 2 methods.

java-checks/src/main/java/org/sonar/java/checks/IgnoredTestsCheck.java
benzonico: Wow... really a corner case IMO but indeed. Case covered.

PR #2179, opened 2018-08-15
SONARJAVA-2798 S1313 (hardcoded IP) should not raise an issue for Strings used as ASN.1 OIDs

alban-auzeill:


pynicolas:
Excluding strings starting with "2.5." seems a better approach: see #2187.

PR #2180, opened 2018-08-15
SONARJAVA-2861: message from S1598 should be explicit when folder name contains dots

andrea-guarino-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/MismatchPackageDirectoryCheck.java
```diff
@@ -27,12 +29,12 @@
 import org.sonar.plugins.java.api.tree.CompilationUnitTree;
 import org.sonar.plugins.java.api.tree.PackageDeclarationTree;
 
-import java.io.File;
-
 @Rule(key = "S1598")
 public class MismatchPackageDirectoryCheck extends BaseTreeVisitor implements JavaFileScanner {
 
   private JavaFileScannerContext context;
+  private static final String MESSAGE = "This file \"${0}\" should be located in \"${1}\" directory, not in \"${2}\"";
```
alban-auzeill: I know that you dream to be a millionaire, but it's not a reason to add *$$$* everywhere.
MessageFormat.format arguments should be `{0}`, `{1}`, `{2}`  and not  `${0}`, `${1}`, `${2}`

java-checks/src/main/java/org/sonar/java/checks/MismatchPackageDirectoryCheck.java
```diff
@@ -49,8 +51,14 @@ public void visitCompilationUnit(CompilationUnitTree tree) {
       File javaFile = context.getFile();
       String dir = javaFile.getParent();
       if (!dir.endsWith(packageName)) {
-        context.reportIssue(
-          this, packageDeclaration.packageName(), "This file \"" + javaFile.getName() + "\" should be located in \"" + packageName + "\" directory, not in \"" + dir + "\".");
+        String dirWithoutDots = dir.replace(".", "/");
```
alban-auzeill: Replace "/" by File.separator

java-checks/src/test/files/checks/mismatchPackage/with.dots/PackageWithDots.java
```diff
@@ -0,0 +1,3 @@
+package mismatchPackage.with.dots; // Noncompliant {{This file "$PackageWithDots.java" should be located in "$mismatchPackage/with/dots" directory, not in "$src/test/files/checks/mismatchPackage/with.dots"(Do not use dots in directory names).}}
```
alban-auzeill: This test should fail on windows, I don't see a simple solution to test the message. We could probably accept that the message is not tested.

andrea-guarino-sonarsource: Good point, I'll remove the message then

java-checks/src/main/java/org/sonar/java/checks/MismatchPackageDirectoryCheck.java
andrea-guarino-sonarsource: :sweat_smile: thanks, I mixed two syntaxes

PR #2183, opened 2018-08-16
 SONARJAVA-2824 S2115: Fix FP on PlusAssignement and on Oracle and MySql connection url

andrea-guarino-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/helpers/IdentifierUtils.java
```diff
@@ -35,8 +36,15 @@ private IdentifierUtils() {
     T value = resolver.apply(expression);
     if (value == null && expression.is(Tree.Kind.IDENTIFIER)) {
       ExpressionTree last = ReassignmentFinder.getClosestReassignmentOrDeclarationExpression(expression, ((IdentifierTree) expression).symbol());
-      value = last == null || last == expression ? null : getValue(last, resolver);
+      value = last == null || !isStrictAssignmentOrDeclaration(last) || last == expression ? null : getValue(last, resolver);
```
pynicolas: This ternary operator seems barely readable to me. I think I would prefer a plain if/else.

java-checks/src/main/java/org/sonar/java/checks/security/EmptyDatabasePasswordCheck.java
```diff
@@ -63,13 +66,18 @@ private static boolean hasEmptyValue(ExpressionTree expression) {
...
-    if (url != null && (!url.contains("password") || EMPTY_PASSWORD_PATTERN.matcher(url).matches())) {
-      reportIssue(mit, MESSAGE);
-    }
+    return url != null &&
+      (patternContainsEmptyPassword(url, MYSQL_URL_PATTERN) ||
+        patternContainsEmptyPassword(url, ORACLE_URL_PATTERN) ||
+        EMPTY_PASSWORD_PATTERN.matcher(url).matches());
+  }
 
+  private static boolean patternContainsEmptyPassword(String url, Pattern mysqlUrlPattern) {
```
pynicolas: Naming:
* Why not name the method `urlContainsEmptyPassword`?
* Why does the parameter name start with `mysql`?

andrea-guarino-sonarsource: - Agreed, going to rename the method name
- Forgot to rename the parameter name proposed by IntelliJ, thanks for pointing that out

PR #2195, opened 2018-08-22
SONARJAVA-2874: Exclude Spring @RequestMapping shortcut annotations from too many parameters check rule

hathazylaszlojr:
Updated the white list for `@GetMapping` and other shortcut annotations for `@RequestMapping`


hathazylaszlojr:
Great! Thank you very much!

PR #2205, opened 2018-09-17
SONARJAVA-2888 Register warning when missing bytecode detected

janos-ss:
This work depends on new features in the SonarQube API. It can be merged after the API changes are part of a milestone / stable release (ETA Sep 28).

**Update:** the new ETA is Oct 10.

### DIFF
--- a/external-reports/src/test/java/org/sonar/java/externalreport/SpotBugsSensorTest.java
+++ b/external-reports/src/test/java/org/sonar/java/externalreport/SpotBugsSensorTest.java
@@ -108,7 +108,7 @@ public class SpotBugsSensorTest {
     ExternalIssue first = externalIssues.get(0);
     assertThat(first.primaryLocation().inputComponent().key()).isEqualTo("spotbugs-project:src/main/java/org/myapp/Main.java");
     assertThat(first.ruleKey().rule()).isEqualTo("HE_EQUALS_USE_HASHCODE");
-    assertThat(first.ruleKey().repository()).isEqualTo("spotbugs");
+    assertThat(first.ruleKey().repository()).isEqualTo("external_spotbugs");
     assertThat(first.type()).isEqualTo(RuleType.CODE_SMELL);
     assertThat(first.severity()).isEqualTo(Severity.MAJOR);
     assertThat(first.remediationEffort().longValue()).isEqualTo(5L);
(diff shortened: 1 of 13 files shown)


sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaPlugin.java
```diff
@@ -82,8 +87,19 @@ public void define(Context context) {
...
+      builder.add(AnalysisWarningsWrapper.class);
+    }
+
+    context.addExtensions(builder.build());
   }
 
+  /**
+   * Drop this and related when the minimum supported version of SonarJava reaches 7.4.
+   */
+  private static boolean isSupported(SonarRuntime runtime) {
```
benzonico: the name of the method is a bit too generic, maybe this should be isAnalysisWarningsSupported ?

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaPluginTest.java
```diff
@@ -66,4 +69,29 @@ public void sonarqube_7_2_extensions() {
...
+  }
+
+  @Test
+  public void use_AnalysisWarningsWrapper_before_SQ_7_4_and_sonarlint() {
+    SonarRuntime runtime = SonarRuntimeImpl.forSonarQube(Version.create(7, 3), SonarQubeSide.SCANNER);
+    Plugin.Context context = new Plugin.Context(runtime);
+    javaPlugin.define(context);
+    assertThat(context.getExtensions()).doesNotContain(AnalysisWarningsWrapper.class);
+
+    runtime = SonarRuntimeImpl.forSonarQube(ANALYSIS_WARNINGS_MIN_SUPPORTED_SQ_VERSION, SonarQubeSide.SCANNER);
```
benzonico: I don't think for testing we should use this constant but rather configure the version using VERSION_7_4 and make the constant ANALYSIS_...  a private static final variable in java plugin to better encapsulate things.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaPlugin.java
```diff
@@ -37,6 +40,8 @@
 
 public class JavaPlugin implements Plugin {
 
+  public static final Version ANALYSIS_WARNINGS_MIN_SUPPORTED_SQ_VERSION = Version.create(7, 4);
```
benzonico: I think this should be private.

external-reports/src/test/java/org/sonar/java/externalreport/SpotBugsSensorTest.java
```diff
@@ -108,7 +108,7 @@ public void issues_with_sonarqube_72() throws IOException {
     ExternalIssue first = externalIssues.get(0);
     assertThat(first.primaryLocation().inputComponent().key()).isEqualTo("spotbugs-project:src/main/java/org/myapp/Main.java");
     assertThat(first.ruleKey().rule()).isEqualTo("HE_EQUALS_USE_HASHCODE");
-    assertThat(first.ruleKey().repository()).isEqualTo("spotbugs");
+    assertThat(first.ruleKey().repository()).isEqualTo("external_spotbugs");
```
benzonico: FYI: this change is ok because of a change on SensorContextTester side : so only the way to test things has changed with upgrade of API.

benzonico:
LGTM

PR #2225, opened 2018-10-12
SONARJAVA-2711 Rule S3078: "volatile" variables should not be incremented using increment operator

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -0,0 +1,97 @@
...
+    return ImmutableList.of(Tree.Kind.PREFIX_INCREMENT, Tree.Kind.POSTFIX_INCREMENT, Tree.Kind.PREFIX_DECREMENT, Tree.Kind.POSTFIX_DECREMENT, Tree.Kind.LOGICAL_COMPLEMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    UnaryExpressionTree unaryExpression = (UnaryExpressionTree) tree;
+    ExpressionTree expression = ExpressionUtils.skipParentheses(unaryExpression.expression());
+    IdentifierTree identifier = getVariableIdentifier(expression);
+
+    if (identifier == null || !identifier.symbol().isVolatile()) {
```
m-g-sonar: because you use semantic here, you should be sure that we have access to it, and start the `visitNode()` method with semantic check:
```
if (!hasSemantic()) {
  return;
}
```

java-checks/src/main/java/org/sonar/java/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -0,0 +1,97 @@
...
+import org.sonar.plugins.java.api.tree.UnaryExpressionTree;
+
+@Rule(key = "S3078")
+public class VolatileVariablesIncrementDecrementCheck extends IssuableSubscriptionVisitor {
+
+  private static final String MESSAGE_TEMPLATE = "Use an \"%s\" for this field; its %s are atomic.";
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return ImmutableList.of(Tree.Kind.PREFIX_INCREMENT, Tree.Kind.POSTFIX_INCREMENT, Tree.Kind.PREFIX_DECREMENT, Tree.Kind.POSTFIX_DECREMENT, Tree.Kind.LOGICAL_COMPLEMENT);
```
m-g-sonar: According to the [CERT](https://wiki.sei.cmu.edu/confluence/display/java/VNA02-J.+Ensure+that+compound+operations+on+shared+variables+are+atomic) this rule is based on, it seems to me that by restraining to only these operators (`++`, `--`, `!`), we are missing quite a lot of cases: 

> Expressions that include postfix or prefix increment (++), postfix or prefix decrement (--), or compound assignment operators always result in compound operations. Compound assignment expressions use operators such as ***=, /=, %=, +=, -=, <<=, >>=, >>>=, ^= and |=**.

Do we willingly restrain ourselves to these cases? (would be fine by me, but i would expect another rule covering the others then)

java-checks/src/main/java/org/sonar/java/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  private void checkBooleanToggling(@Nullable Tree tree, IdentifierTree identifier) {
+    Tree parent = tree != null ? tree.parent() : null;
+    if (parent != null) {
+      if (parent.is(Tree.Kind.PARENTHESIZED_EXPRESSION)) {
+        checkBooleanToggling(parent, identifier);
+      } else if (parent.is(Tree.Kind.ASSIGNMENT)) {
+        ExpressionTree variable = ((AssignmentExpressionTree) parent).variable();
+        if ((variable.is(Tree.Kind.IDENTIFIER) && identifier.name().equals(((IdentifierTree) variable).name()))
```
m-g-sonar: You should check for symbol equality here. If it's a field and a parameter (or worst, the field of another totally unrelated object having the same name), variables will have the same name without being the same object underneath!

java-checks/src/main/java/org/sonar/java/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -0,0 +1,97 @@
...
+      return (IdentifierTree) expressionTree;
+    } else if (expressionTree.is(Tree.Kind.MEMBER_SELECT)) {
+      return ((MemberSelectExpressionTree) expressionTree).identifier();
+    }
+    return null;
+  }
+
+  private void checkBooleanToggling(@Nullable Tree tree, IdentifierTree identifier) {
+    Tree parent = tree != null ? tree.parent() : null;
+    if (parent != null) {
```
m-g-sonar: instead of nesting operators, i usually prefer the early return:
```
if (parent == null) {
  return;
}
```

java-checks/src/main/java/org/sonar/java/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -0,0 +1,97 @@
...
+  private static IdentifierTree getVariableIdentifier(ExpressionTree expressionTree) {
+    if (expressionTree.is(Tree.Kind.IDENTIFIER)) {
+      return (IdentifierTree) expressionTree;
+    } else if (expressionTree.is(Tree.Kind.MEMBER_SELECT)) {
+      return ((MemberSelectExpressionTree) expressionTree).identifier();
+    }
+    return null;
+  }
+
+  private void checkBooleanToggling(@Nullable Tree tree, IdentifierTree identifier) {
```
m-g-sonar: To be honest, I'm not sure the `@Nullable` annotation is required here. By construction of the language, I don't see how it would be possible, starting from a logical compliment and looping through parenthesis, to end up with the compilation unit has parent. I would simply drop the case and mark the SQ issue as won't fix.

java-checks/src/main/java/org/sonar/java/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -0,0 +1,97 @@
...
+        if ((variable.is(Tree.Kind.IDENTIFIER) && identifier.name().equals(((IdentifierTree) variable).name()))
+          || (variable.is(Tree.Kind.MEMBER_SELECT) && identifier.name().equals(((MemberSelectExpressionTree) variable).identifier().name()))) {
+          reportIssue(parent, "Use an \"AtomicBoolean\" for this field");
+        }
+      }
+    }
+  }
+
+  private void checkIncrementDecrement(Tree tree, IdentifierTree identifier, String type) {
+    if (identifier.symbol().type().is("int") || identifier.symbol().type().is("java.lang.Integer")) {
```
m-g-sonar: Use a dedicated variable instead of calling 4 times `identifier.symbol().type()`

java-checks/src/main/java/org/sonar/java/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -0,0 +1,97 @@
...
+    IdentifierTree identifier = getVariableIdentifier(expression);
+
+    if (identifier == null || !identifier.symbol().isVolatile()) {
+      return;
+    }
+
+    if (tree.is(Tree.Kind.LOGICAL_COMPLEMENT)) {
+      checkBooleanToggling(tree, identifier);
+    } else if (tree.is(Tree.Kind.PREFIX_INCREMENT) || tree.is(Tree.Kind.POSTFIX_INCREMENT)) {
+      checkIncrementDecrement(tree, identifier, "increments");
```
m-g-sonar: Why doing the test of kind here (and provide the consequent String)? You could have this logic moved into the `chekIncrementDecrement()` method, which would simplify the handling of the various cases.

java-checks/src/test/files/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -57,4 +57,17 @@ public boolean toggleBooleans(){
     bool1 = boo2 = !false;
```
m-g-sonar: should probably by `boo1` here, otherwise it does not compile.

java-checks/src/test/files/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -28,9 +28,9 @@ public void incrementCounts() {
   }
 
   public void decrementCounts() {
-    count1--; // Noncompliant {{Use an "AtomicInteger" for this field; its decrements are atomic.}}
+    count1--; // Noncompliant {{Use an "AtomicInteger" for this field; its operations are atomic.}}
     --count1; // Noncompliant
-    (count2)--; // Noncompliant {{Use an "AtomicLong" for this field; its decrements are atomic.}}
+    (count2)--; // Noncompliant {{Use an "AtomicLong" for this field; its operations are atomic.}}
     (--count2); // Noncompliant
```
m-g-sonar: this does not compile

java-checks/src/test/files/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -11,13 +11,13 @@
...
   public void incrementCounts() {
-    count1++; // Noncompliant {{Use an "AtomicInteger" for this field; its increments are atomic.}}
-    ++this.count1; // Noncompliant {{Use an "AtomicInteger" for this field; its increments are atomic.}}
-    (count2)++; // Noncompliant {{Use an "AtomicLong" for this field; its increments are atomic.}}
-    (++count2); // Noncompliant {{Use an "AtomicLong" for this field; its increments are atomic.}}
-    count3++; // Noncompliant {{Use an "AtomicInteger" for this field; its increments are atomic.}}
+    count1++; // Noncompliant {{Use an "AtomicInteger" for this field; its operations are atomic.}}
+    ++this.count1; // Noncompliant {{Use an "AtomicInteger" for this field; its operations are atomic.}}
+    (count2)++; // Noncompliant {{Use an "AtomicLong" for this field; its operations are atomic.}}
+    (++count2); // Noncompliant {{Use an "AtomicLong" for this field; its operations are atomic.}}
```
m-g-sonar: this does not compile

java-checks/src/main/java/org/sonar/java/checks/VolatileVariablesIncrementDecrementCheck.java
```diff
@@ -71,26 +92,24 @@ private static IdentifierTree getVariableIdentifier(ExpressionTree expressionTre
...
-          reportIssue(parent, "Use an \"AtomicBoolean\" for this field");
-        }
+  private void checkBooleanToggling(Tree tree, Symbol identifierSymbol) {
+    Tree parent = tree.parent();
+    if (parent.is(Tree.Kind.PARENTHESIZED_EXPRESSION)) {
+      checkBooleanToggling(parent, identifierSymbol);
+    } else if (parent.is(Tree.Kind.ASSIGNMENT)) {
+      IdentifierTree variableIdentifier = getVariableIdentifier(((AssignmentExpressionTree) parent).variable());
+      if (variableIdentifier != null && identifierSymbol.equals(variableIdentifier.symbol())) {
+        reportIssue(parent, "Use an \"AtomicBoolean\" for this field");
```
m-g-sonar: we are missing the synchronized cases which are valid.
The following cases are synchronized, so there should not raise any issues.
```java
  public synchronized void toggleBoo() {
    boo1 = !boo1;
  }

  public void toggleBoo2() {
    synchronized (stuff) {
      boo1 = !boo1;
    }
  }
```

PR #2227, opened 2018-10-12
SONARJAVA-2882 Rule S3864: "Stream.peek" should not be used

mpaladin:


java-checks/src/main/java/org/sonar/java/checks/StreamPeekCheck.java
```diff
@@ -0,0 +1,43 @@
...
+import org.sonar.java.matcher.MethodMatcher;
+import org.sonar.java.model.ExpressionUtils;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+
+@Rule(key = "S3864")
+public class StreamPeekCheck extends AbstractMethodDetection {
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return ImmutableList.of(MethodMatcher.create().typeDefinition("java.util.stream.Stream").name("peek").withAnyParameters());
```
m-g-sonar: I think we should also include primitive-based streams `IntStream.peek(IntConsumer)`, `LongStream.peek(LongConsumer)` and `DoubleStream.peek(DoubleConsumer)`.

PR #2236, opened 2018-10-16
SONARJAVA-2919 S1319 Fix false-positive when method is inherited: signature cannot be changed

benzonico:


java-checks/src/test/files/checks/CollectionImplementationReferencedCheck.java
```diff
@@ -40,3 +40,14 @@ public method(LinkedList<Employee> employees) {} // Noncompliant {{The type of t
     return employees;
   }
 }
+
+class A {
+  public void foo(HashMap<String> map) { // Noncompliant
```
mpaladin: @benzonico `HashMap<String>` makes me wonder if rule should be implemented on semantic + subscription. Are you applying any strategy to rewrite non semantic or visitors rules?
If not, at least let's modify test with `HashMap<String, String>`, same few lines below.

benzonico: There is no such strategy in place, no need to go for semantic if not required. 
Let's fix the test anyway.

PR #2238, opened 2018-10-16
SONARJAVA-2688 RSPEC-4517 InputSteam.read() implementation should not return a signed byte

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/InputStreamReadCheck.java
```diff
@@ -0,0 +1,78 @@
...
+    return Collections.singletonList(Tree.Kind.METHOD);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
+      return;
+    }
+    MethodTree method = (MethodTree) tree;
+    if (!INPUT_STREAM_READ.matches(method) || method.block() == null) {
```
benzonico: I would switch those two conditions.

PR #2239, opened 2018-10-16
SONARJAVA-2809 S2189: (Loops should not be infinite) should benefit from SE "always true" knowledge

mpaladin:


java-frontend/src/main/java/org/sonar/java/se/checks/NoWayOutLoopCheck.java
```diff
@@ -239,6 +247,16 @@ public void visitUnaryExpression(UnaryExpressionTree expression) {
     }
   }
 
+  private static Tree firstStatementParent(Tree node) {
+    Tree current = node;
+    while (true) {
```
benzonico: probably safer to have a current != null

PR #2240, opened 2018-10-16
SONARJAVA-2933 Change approach of Subscription visitor

benzonico:
Only one visit with callback to checks instead of one visit by check

Rough estimation of improvement on ruling jdk6 on my machine : 
no checks : ~1'43"
all checks : ~2'44"
new Approach all checks ~2'10"

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/UselessImportCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/UselessImportCheck.java
@@ -21,10 +21,15 @@ package org.sonar.java.checks;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Sets;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import org.sonar.check.Rule;
 import org.sonar.java.RspecKey;
+import org.sonar.java.ast.visitors.SubscriptionVisitor;
 import org.sonar.java.checks.helpers.ExpressionsHelper;
-import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.JavaFileScanner;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
 import org.sonar.plugins.java.api.tree.ArrayTypeTree;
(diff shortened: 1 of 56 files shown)


java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
```diff
@@ -124,12 +132,74 @@ public void visitFile(@Nullable Tree parsedTree) {
...
+        checks.get(k).add(s);
+      })
+      );
+    }
+
+
+    public void run(JavaFileScannerContext javaFileScannerContext) {
+      subscriptionVisitors.forEach(s -> s.setContext(javaFileScannerContext));
+      visit(javaFileScannerContext.getTree());
+      subscriptionVisitors.forEach(s -> s.leaveFile(javaFileScannerContext));
```
m-g-sonar: There is something I really don't like here... But I have no argument strictly against this approach. I'm however just wondering about the memory impact.

 We have numerous checks which are keeping in memory (references to) some nodes (for hierarchy checking for instance, reporting, or whatever other cases). Previously, this memory was (theoretically) systematically cleaned when leaving each file, and filled up again with a new check. With this new approach, the memory usage of each node is going to grow as long as we didn't reach the end of file. Do we want to live with this? On the long run it may means asking for more memory for an analysis.

benzonico: As discussed this is not a lot different from what we have today when keeping a context within a file : most of the checks are only clearing their references to AST when entering the next file anyway.

java-checks/src/main/java/org/sonar/java/checks/UselessImportCheck.java
```diff
@@ -223,13 +222,7 @@ public void visitTrivia(SyntaxTrivia syntaxTrivia) {
...
 
     private void updatePendingImportsForComments(String comment) {
-      Iterator<String> it = pendingImports.iterator();
-      while (it.hasNext()) {
-        String pendingImport = it.next();
-        if (comment.contains(extractLastClassName(pendingImport))) {
-          it.remove();
-        }
-      }
+      pendingImports.removeIf(pendingImport -> comment.contains(extractLastClassName(pendingImport)));
```
m-g-sonar: This is nice, could almost be a rule!

java-frontend/src/main/java/org/sonar/plugins/java/api/IssuableSubscriptionVisitor.java
```diff
@@ -31,6 +31,11 @@
  */
 public abstract class IssuableSubscriptionVisitor extends SubscriptionVisitor {
 
+  @Override
+  protected void scanTree(Tree tree) {
+    throw new UnsupportedOperationException("IssuableSubscriptionVisitor should not drive visit of AST.");
```
m-g-sonar: This need to be documented as public API change for next release (what about adding it to the ticket?)
Some (badly written) custom rules may start failing after the update because of it.

java-checks/src/main/java/org/sonar/java/checks/UselessImportCheck.java
benzonico: made from intellij check but indeed @agigleux can you have a look ?

PR #2241, opened 2018-10-17
SONARJAVA-2819 S2658 Don't raise issue for constants arguments

benzonico:


java-checks/src/main/java/org/sonar/java/checks/DynamicClassLoadCheck.java
```diff
@@ -41,7 +42,10 @@
 
   @Override
   protected void onMethodInvocationFound(MethodInvocationTree mit) {
-    reportIssue(ExpressionUtils.methodName(mit), "Remove this use of dynamic class loading.");
+    String stringConstant = ConstantUtils.resolveAsStringConstant(mit.arguments().get(0));
+    if(stringConstant == null) {
```
mpaladin: @benzonico formatting, I will fix it for you.

PR #2244, opened 2018-10-17
SONARJAVA-2822 Fix resolution of inner class constructor in method re…

benzonico:
…ference


java-frontend/src/test/files/sym/ConstructorInMethodRef.java
```diff
@@ -20,5 +20,18 @@ Type erasure() {
...
+  public class Bar {
+    public void print() {
+      java.util.List<String> list = java.util.Arrays.asList("x", "y", "z");
+      java.util.List<Foo> foos = list.stream().map(Foo::new).collect(Collectors.toList());
+      System.out.println(foos.get(0).foo);
+    }
 
+    public class Foo {
+      private String foo;
+      private Foo(String foo) { // FP
```
m-g-sonar: this is not a FP anymore.

PR #2246, opened 2018-10-17
SONARJAVA-2710 Rule S3077: Non-primitive fields should not be "volatile"

benzonico:


java-checks/src/main/java/org/sonar/java/checks/VolatileNonPrimitiveFieldCheck.java
```diff
@@ -54,12 +54,12 @@ public void visitNode(Tree tree) {
...
     Type varType = variableTree.type().symbolType();
-    if(varType.isArray()) {
+    if (varType.isArray()) {
       String atomicType = "Reference";
       Type elementType = ((Type.ArrayType) varType).elementType();
-      if(elementType.isPrimitive(Type.Primitives.LONG)) {
+      if (elementType.isPrimitive(Type.Primitives.LONG)) {
         atomicType = "Long";
-      } else if(elementType.isPrimitive(Type.Primitives.INT)) {
+      } else if (elementType.isPrimitive(Type.Primitives.INT)) {
```
benzonico: ok i'll try to pay attention to this...

mpaladin: 😉

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S3077_java.json
```diff
@@ -0,0 +1,20 @@
...
+    "constantCost": "20min"
+  },
+  "tags": [
+    "multi-threading",
+    "cert"
+  ],
+  "standards": [
+    "CERT"
+  ],
+  "defaultSeverity": "Major",
```
mpaladin: @benzonico any thought about being major bug?

benzonico: The "bug" might be arguable given the amount of issues in ruling (and in jdk/guava especially)

PR #2251, opened 2018-10-18
SONARJAVA-2924 Fix infinite recursion because of inheritance of neste…

benzonico:
…d types

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
@@ -49,29 +49,6 @@ public class NoTestInTestClassCheck extends IssuableSubscriptionVisitor {
   private final Set<String> testAnnotations = new HashSet<>();
   private final Set<String> seenAnnotations = new HashSet<>();
 
-  private boolean isTestAnnotation(Type type) {
-    return testAnnotations.contains(type.fullyQualifiedName()) || isJUnitTestableMetaAnnotated(type);
-  }
-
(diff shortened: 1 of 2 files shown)


java-checks/src/test/files/checks/ArrayStackTest.java
```diff
@@ -0,0 +1,122 @@
+/*
```
christophe-zurn-sonarsource: I don't think this file is related to this PR.

java-checks/src/main/java/org/sonar/java/checks/NoTestInTestClassCheck.java
```diff
@@ -167,8 +136,35 @@ private boolean isTestMethod(Symbol method) {
...
+    return false;
+  }
+
   private static Stream<Symbol.MethodSymbol> getAllMembers(Symbol.TypeSymbol symbol, boolean isEnclosed) {
-    if ("java.lang.Object".equals(symbol.type().fullyQualifiedName())) {
+    return getAllMembers(symbol, isEnclosed, new HashSet<>());
+  }
+
+  private static Stream<Symbol.MethodSymbol> getAllMembers(Symbol.TypeSymbol symbol, boolean isEnclosed, Set<Symbol> visitedSymbols) {
+    if (!visitedSymbols.add(symbol) || "java.lang.Object".equals(symbol.type().fullyQualifiedName())) {
```
christophe-zurn-sonarsource: Minor improvement: you could use
```suggestion
    if (!visitedSymbols.add(symbol) || symbol.type().is("java.lang.Object")) {
```
instead

benzonico:
agreed, and made it cleaner.

PR #2254, opened 2018-10-18
SONARJAVA-2925 Add mechanism to detect method references invocation

benzonico:


java-checks/src/main/java/org/sonar/java/checks/methods/AbstractMethodDetection.java
```diff
@@ -70,6 +76,10 @@ protected void onConstructorFound(NewClassTree newClassTree) {
     // Do nothing by default
   }
 
+  protected void onMethodReferenceFound(MethodReferenceTree methodReferenceTree) {
```
m-g-sonar: From my point of view, this is not yet complete to add support of method references. MethodMatchers are relying on the same mechanism and should support them as well.

I would expect to modify the `MethodMatcher` and `MethodMatcherCollection` classes to add some  `matches(MethodReferenceTree)` / `anyMatch(MethodReferenceTree)` methods there.

java-checks/src/main/java/org/sonar/java/checks/methods/AbstractMethodDetection.java
```diff
@@ -57,6 +58,11 @@ private void checkInvocation(Tree tree, MethodMatcher invocationMatcher) {
       if (invocationMatcher.matches(newClassTree)) {
         onConstructorFound(newClassTree);
       }
+    } else if (tree.is(Tree.Kind.METHOD_REFERENCE)) {
+      MethodReferenceTree methodReferenceTree = (MethodReferenceTree) tree;
+      if (invocationMatcher.matches(methodReferenceTree.method().symbol())) {
```
m-g-sonar: having dedicated method in method matchers would simplify this

java-frontend/src/main/java/org/sonar/java/matcher/MethodMatcher.java
```diff
@@ -158,6 +162,14 @@ private boolean matches(Symbol symbol, @Nullable Type callSiteType) {
     return symbol.isMethodSymbol() && isSearchedMethod((MethodSymbol) symbol, callSiteType);
   }
 
+  private static Type getCallSiteType(MethodReferenceTree referenceTree) {
```
m-g-sonar: Should be annotated with `@CheckForNull`

m-g-sonar: Should be annotated with `@CheckForNull`

m-g-sonar: should be annotated with `@CheckForNull`.

m-g-sonar: This method should be annotated with `@CheckForNull`.

m-g-sonar: This method should be annotated with `@CheckForNull`.

PR #2263, opened 2018-10-22
SONARJAVA-2898  RSPEC-4784 Using regex is a hotspot

benzonico:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S4784_java.html
```diff
@@ -0,0 +1,90 @@
...
+<p>Avoid executing a user input string as a regular expression. If this is required, restrict the allowed regular expressions.</p>
+<p>Check whether your regular expression engine (the algorithm executing your regular expression) has any known vulnerabilities. Search for
+vulnerability reports mentioning the one engine you're are using.</p>
+<p>Test your regular expressions with techniques such as equivalence partitioning, and boundary value analysis, and test for robustness. Try not to
+make complex regular expressions as they are difficult to understand and test. Note that some regular expression engines will match only part of the
+input if no anchors are used. In PHP for example <code>preg_match("/[A-Za-z0-9]+/", $text)</code> will accept any string containing at least one
+alphanumeric character because it has no anchors.</p>
+<h2>Questionable Code Example</h2>
+<pre>
+  import java.util.regex.Pattern;
```
m-g-sonar: I really don't like this formatting. It mixes imports and content which should be in the body of a method. I think we should improve it and split the two examples. The second one preceded by one sentence explicitly mentioning the bean validation context, without having it being a simple comment before new imports.

PR #2264, opened 2018-10-22
SONARJAVA-2909 RSPEC-4834 Controlling permissions is security-sensitive

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/security/ControllingPermissionsCheck.java
```diff
@@ -0,0 +1,141 @@
...
+    "org.springframework.security.access.AccessDecisionVoter",
+    "org.springframework.security.access.AccessDecisionManager",
+    "org.springframework.security.access.AfterInvocationProvider",
+    "org.springframework.security.access.PermissionEvaluator",
+    "org.springframework.security.access.expression.SecurityExpressionOperations",
+    "org.springframework.security.access.expression.method.MethodSecurityExpressionHandler",
+    "org.springframework.security.core.GrantedAuthority",
+    "org.springframework.security.acls.model.PermissionGrantingStrategy"));
+
+  private static final List<String> CLASSES = Collections.singletonList("org.springframework.security.config.annotation.method.configuration.GlobalMethodSecurityConfiguration");
```
benzonico: Naming of this constant can probably be improved to designate which CLASSES/INTERFACES/ANNOTATIONS it targets.

java-checks/src/main/java/org/sonar/java/checks/security/ControllingPermissionsCheck.java
```diff
@@ -0,0 +1,141 @@
...
+    if (context.getSemanticModel() == null) {
+      return;
+    }
+    this.context = context;
+    scan(context.getTree());
+  }
+
+  @Override
+  public void visitClass(ClassTree tree) {
+    if (!tree.symbol().isInterface()) {
```
benzonico: I did not spot a testcase where we would expect an issue but it is not raised because it is an interface (with a proper comment about this).

java-checks/src/main/java/org/sonar/java/checks/security/ControllingPermissionsCheck.java
```diff
@@ -0,0 +1,141 @@
...
+      tree.superInterfaces().forEach(superInterface -> {
+        Type interfaceType = superInterface.symbolType();
+        if (INTERFACES.stream().anyMatch(interfaceType::is)) {
```
benzonico: why not `tree.superInterfaces.stream().filter(Interfaces.... ).forEach(reportIssue)` ?

java-checks/src/main/java/org/sonar/java/checks/security/ControllingPermissionsCheck.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+      });
+      TypeTree superClass = tree.superClass();
+      if (superClass != null && CLASSES.stream().anyMatch(superClass.symbolType()::is)) {
+        context.reportIssue(this, superClass, MESSAGE);
+      }
+    }
+    tree.modifiers().annotations().forEach(annotationTree -> {
+      Type annotationType = annotationTree.symbolType();
+      if (JSR_250_ANNOTATIONS.stream().anyMatch(annotationType::is)) {
```
benzonico: Same remark : filter + for each ?

java-checks/src/main/java/org/sonar/java/checks/security/ControllingPermissionsCheck.java
```diff
@@ -0,0 +1,141 @@
...
+      }
+    }
+    super.visitNewClass(tree);
+  }
+
+  @Override
+  public void visitMethod(MethodTree tree) {
+    tree.modifiers().annotations().forEach(annotationTree -> {
+      Type annotationType = annotationTree.symbolType();
+      if (Stream.concat(ANNOTATIONS.stream(), JSR_250_ANNOTATIONS.stream()).anyMatch(annotationType::is)) {
```
benzonico: I would just go for a simple || here, but this is mainly styling matter, I find the concat confusing.

java-checks/src/test/files/checks/security/ControllingPermissionsCheck.java
```diff
@@ -0,0 +1,112 @@
...
+    GlobalMethodSecurityConfiguration myGlobalMethodSecurityConfiguration1 = new GlobalMethodSecurityConfiguration() { }; // Noncompliant
+    GlobalMethodSecurityConfiguration myGlobalMethodSecurityConfiguration2 = new GlobalMethodSecurityConfiguration();
+
+    MyClass myClass = new MyClass() { };
+  }
+}
+
+@RolesAllowed("jsr250") // Noncompliant
+class JSR_250 {
+  @RolesAllowed(value = "foo") // Noncompliant
```
benzonico: Maybe worth adding at least one test for annotation highlighting.

java-checks/src/main/java/org/sonar/java/checks/security/ControllingPermissionsCheck.java
```diff
@@ -0,0 +1,141 @@
...
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.NewClassTree;
+import org.sonar.plugins.java.api.tree.TypeTree;
+
+@Rule(key = "S4834")
+public class ControllingPermissionsCheck extends BaseTreeVisitor implements JavaFileScanner {
```
benzonico: I have no big issue with this (so no pblm to keep it like that), but i'm wondering about the reason to choose a BaseTreeVisitor ? I don't see a strong justification for it.

java-checks/src/main/java/org/sonar/java/checks/security/ControllingPermissionsCheck.java
benzonico: Given the usage of those list I'm wondering if it would not be better to have the predicates as constants (so you can eventually compose them when needed) rather than storing the lists directly.

PR #2265, opened 2018-10-23
SONARJAVA-2907 Rule S4829: Reading the Standard Input is security-sensitive

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/StandardInputReadCheck.java
```diff
@@ -0,0 +1,94 @@
...
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodReferenceTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S4829")
+public class StandardInputReadCheck extends IssuableSubscriptionVisitor {
+
+  private static final List<MethodMatcher> METHOD_MATCHERS = Arrays.asList(
```
m-g-sonar: You should use a `MethodMatcherCollection`, which already have a `anyMatch` method. It will simplify your rule and drop the need of your predicate.

java-checks/src/main/java/org/sonar/java/checks/security/StandardInputReadCheck.java
```diff
@@ -0,0 +1,94 @@
...
+    if (!hasSemantic()) {
+      return;
+    }
+
+    if (tree.is(Tree.Kind.METHOD_INVOCATION)) {
+      checkMethod(tree, methodMatcher -> methodMatcher.matches((MethodInvocationTree) tree));
+    } else if (tree.is(Tree.Kind.METHOD_REFERENCE)) {
+      checkMethod(tree, methodMatcher -> methodMatcher.matches((MethodReferenceTree) tree));
+    } else {
+      checkMemberSelect((MemberSelectExpressionTree) tree);
```
m-g-sonar: Checking member select is not enough, we should focus on identifier. Its probably also possible to only write the test on subscribing to identifiers

```
import static java.lang.System.in; // this will probably raise an issue

class A {
  void foo() throws IOException {
    in.read(); // but this should raise an issue as well, it's an usage of "in"
  }
}
```

java-checks/src/main/java/org/sonar/java/checks/security/StandardInputReadCheck.java
christophe-zurn-sonarsource: Yep, nice :+1:  Forgot about that one

java-checks/src/main/java/org/sonar/java/checks/security/StandardInputReadCheck.java
christophe-zurn-sonarsource: Indeed, focusing on identifiers will enable us to catch all cases

java-checks/src/main/java/org/sonar/java/checks/security/StandardInputReadCheck.java
```diff
@@ -0,0 +1,88 @@
...
+      reportIssue(tree);
+    } else if (tree.is(Tree.Kind.METHOD_REFERENCE) && METHOD_MATCHERS.anyMatch((MethodReferenceTree) tree)) {
+      reportIssue(tree);
+    } else if (tree.is(Tree.Kind.IDENTIFIER)) {
+      checkIdentifier((IdentifierTree) tree);
+    }
+  }
+
+  private void checkIdentifier(IdentifierTree identifier) {
+    if (identifier.symbol().enclosingClass() != null
```
m-g-sonar: I would store the `enclosingClass` in a variable, so we don't recompute it twice.

PR #2267, opened 2018-10-23
SONARJAVA-2901 Rule S4787: Encrypting data is security-sensitive

mpaladin:


java-checks/src/main/java/org/sonar/java/checks/security/DataEncryptionCheck.java
```diff
@@ -0,0 +1,51 @@
...
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+
+@Rule(key = "S4787")
+public class DataEncryptionCheck extends AbstractMethodDetection {
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return ImmutableList.of(
+      MethodMatcher.create()
+        .typeDefinition(TypeCriteria.is("javax.crypto.Cipher"))
```
benzonico: You can use the string overload when TypeCriteria is exact match.

mpaladin: 👍

PR #2269, opened 2018-10-24
SONARJAVA-2900 RSPEC-4797 File creation is a hotspot

benzonico:


benzonico:
indeed, updated the matchers accordingly.

PR #2278, opened 2018-10-25
SONARJAVA-2903 Rule S4825: Sending HTTP requests is security-sensitive

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/HttpRequestsHotspotCheck.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    if (tree.is(Tree.Kind.METHOD_INVOCATION)) {
+      MethodInvocationTree mit = (MethodInvocationTree) tree;
+      if (METHOD_MATCHER.anyMatch(mit)) {
+        reportIssue(ExpressionUtils.methodName(mit), "Make sure that this http request is sent safely.");
+      }
+    } else {
+      TypeCastTree castTree = (TypeCastTree) tree;
+      if (castTree.type().symbolType().is("java.net.HttpURLConnection")) {
```
benzonico: I am wondering about this choice of raising issue on casting : I think this deserves a testcase with a variable casted without method invocation which will trigger the issue. 
I tend to think we should raise issue on methodInvocation url.openconnection that are casted to HttpURLConnection. Maybe to be refined with @nicolas-harraudeau-sonarsource  ?

nicolas-harraudeau-sonarsource: 354K+ hits on Github for `(HttpURLConnection) url.openConnection()`
I think @benzonico's idea is good. Let's go for it.

java-checks/src/main/java/org/sonar/java/checks/security/HttpRequestsHotspotCheck.java
```diff
@@ -0,0 +1,71 @@
...
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    if (!"openConnection".equals(mit.symbol().name()) || isCastToHttpUrlConnection(mit.parent())) {
+      reportIssue(ExpressionUtils.methodName(mit), "Make sure that this http request is sent safely.");
+    }
+  }
+
+  private static boolean isCastToHttpUrlConnection(Tree parent) {
+    if (parent.is(Tree.Kind.PARENTHESIZED_EXPRESSION)) {
+      return isCastToHttpUrlConnection(parent.parent());
```
benzonico: This case is not tested and you can use `ExpressionUtils.skipParentheses`

PR #2286, opened 2018-10-26
SONARJAVA-2900 add missing methods

benzonico:


java-checks/src/main/java/org/sonar/java/checks/security/FileHandlingCheck.java
```diff
@@ -49,6 +49,7 @@
       MethodMatcher.create().typeDefinition(JAVA_IO_FILE).name(INIT).parameters(JAVA_LANG_STRING),
       MethodMatcher.create().typeDefinition(JAVA_IO_FILE).name(INIT).parameters(JAVA_LANG_STRING, JAVA_LANG_STRING),
       MethodMatcher.create().typeDefinition(JAVA_IO_FILE).name(INIT).parameters("java.net.URI"),
+      MethodMatcher.create().typeDefinition(JAVA_IO_FILE).name("createTempFile").parameters(JAVA_LANG_STRING, JAVA_LANG_STRING),
```
m-g-sonar: I guess we don't consider the other `createTempFile()` having a file as parameter for the same reason as file constructor: because the file would already have been flagged?

java-checks/src/test/files/checks/security/FileHandlingCheck.java
```diff
@@ -28,7 +28,7 @@ void fun(String strPath, String StrParent, String StrChild, String prefix, Strin
     new File(parent, StrChild); // compliant we rely on the fact that parent should be highlighted
 
 
-      File.createTempFile(prefix, suffix);
+    File.createTempFile(prefix, suffix); // Noncompliant
```
m-g-sonar: I would add the other createTempFile method to enforce the fact that we explicitly target this one only.

PR #2288, opened 2018-10-26
SONARJAVA-2926 "serialVersionUID" should not be declared blindly

jblievremont:


java-checks/src/test/java/org/sonar/java/checks/serialization/BlindSerialVersionUidCheckTest.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.sonar.java.checks.serialization;
+
+import org.junit.Test;
+import org.sonar.java.checks.verifier.JavaCheckVerifier;
+
+public class BlindSerialVersionUidCheckTest {
+  @Test
+  public void test() {
+    JavaCheckVerifier.verify("src/test/files/checks/serialization/BlindSerialVersionUidCheck.java", new BlindSerialVersionUidCheck());
```
benzonico: You miss the test without semantic, see the method `verifyNoIssueWithoutSemantic`

PR #2289, opened 2018-10-29
Fix quality flaws : improve coverage

benzonico:


java-checks/src/test/java/org/sonar/java/checks/naming/BadTypeParameterNameCheckTest.java
```diff
@@ -26,7 +26,11 @@
 
   @Test
   public void test() {
-    JavaCheckVerifier.verify("src/test/files/checks/naming/BadGenericNameNoncompliant.java", new BadTypeParameterNameCheck());
+    BadTypeParameterNameCheck check = new BadTypeParameterNameCheck();
+    JavaCheckVerifier.verify("src/test/files/checks/naming/BadGenericNameNoncompliant.java", check);
+    // verify that pattern was compiled once for the instance.
```
christophe-zurn-sonarsource: If you want to ensure it is only compiled once, shouldn't the string `format` of the check be changed between the calls? to ensure it has no effect on the second all to `verify`

PR #2291, opened 2018-10-29
SONARJAVA-2918 "read(byte[],int,int)" should be overridden

jblievremont:


java-checks/src/main/java/org/sonar/java/checks/InputStreamOverrideReadCheck.java
```diff
@@ -0,0 +1,79 @@
...
+
+    ClassTree classTree = (ClassTree) tree;
+    Type superType = classTree.symbol().superClass();
+    IdentifierTree className = classTree.simpleName();
+    if (className == null || classTree.symbol().isAbstract() || superType == null || !(superType.is("java.io.InputStream") || superType.is("java.io.FilterInputStream"))) {
+      return;
+    }
+
+    Optional<MethodTree> readByteIntInt = findMethod(classTree, READ_BYTES_INT_INT);
+    Optional<MethodTree> readInt = findMethod(classTree, READ_INT);
```
benzonico: tiny optim, this method and the resulting variable should be called only inside the if.

PR #2297, opened 2018-10-29
SONARJAVA-2536 Strenghten acceptable types for lambda and method refe…

benzonico:
…rences

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/resolve/Resolve.java
+++ b/java-frontend/src/main/java/org/sonar/java/resolve/Resolve.java
@@ -1027,31 +1027,36 @@ public class Resolve {
   }
 
   public Optional<JavaSymbol.MethodJavaSymbol> getSamMethod(JavaType lambdaType) {
-    Optional<JavaSymbol.MethodJavaSymbol> result = Optional.empty();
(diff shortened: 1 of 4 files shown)


java-frontend/src/main/java/org/sonar/java/resolve/Resolve.java
```diff
@@ -1027,22 +1027,36 @@ private static JavaType getPrimitive(JavaType primitiveOrWrapper) {
...
-      Optional<JavaSymbol.MethodJavaSymbol> samMethod = getSamMethod(type);
-      if (samMethod.isPresent()) {
-        return samMethod;
+    if (methodSymbols.size() == 1) {
+      JavaSymbol.MethodJavaSymbol override = methodSymbols.get(0).overriddenSymbol();
+      Optional<JavaSymbol.MethodJavaSymbol> otherAbstractMethod = lambdaType.symbol.superTypes()
+        .stream()
+        .filter(t -> t.symbol().isInterface())
+        .flatMap(t -> abstractMethodsOfType(t).stream())
+        .filter(m -> override == null || !override.equals(m))
```
christophe-zurn-sonarsource: We could add cases to the `Lambdas.java` test to cover these cases already, it seems it is not entirely covered by current tests

PR #2301, opened 2018-10-30
SONARJAVA-2894 Iteration on Collection should use the Collection's item type

jblievremont:


its/ruling/src/test/resources/guava/squid-S4838.json
```diff
@@ -0,0 +1,14 @@
+{
+'com.google.guava:guava:src/com/google/common/collect/MapMakerInternalMap.java':[
+3517,
```
christophe-zurn-sonarsource: This seems to be a FP

jblievremont: Agreed. As discussed, I updated the rule to only raise an issue when the iteration variable is actually downcast in the loop statement.

java-checks/src/main/java/org/sonar/java/checks/ForLoopVariableTypeCheck.java
```diff
@@ -0,0 +1,118 @@
...
+      this.collectionItemType = collectionItemType;
+    }
+
+    @Override
+    public List<Tree.Kind> nodesToVisit() {
+      return Collections.singletonList(Tree.Kind.TYPE_CAST);
+    }
+
+    @Override
+    public void visitNode(Tree tree) {
```
christophe-zurn-sonarsource: You should check that the cast is related to the variable. Otherwise any unrelated cast would raise the issue.
Ex:
```
J obj;
    for(I i: collectionOfK) { // Noncompliant
      (L) obj;
    }
```

java-checks/src/main/java/org/sonar/java/checks/ForLoopVariableTypeCheck.java
```diff
@@ -0,0 +1,118 @@
...
+      ForLoopVariableTypeCheck.this.reportIssue(forEachStatement.variable().type(), String.format(PRIMARY_MESSAGE, variableType.name()),
```
christophe-zurn-sonarsource: We can stop the visit once we've found a relevant cast, as we only want to raise the issue once.

java-checks/src/main/java/org/sonar/java/checks/ForLoopVariableTypeCheck.java
```diff
@@ -0,0 +1,130 @@
...
+        ForLoopVariableTypeCheck.this.reportIssue(forEachStatement.variable().type(), String.format(PRIMARY_MESSAGE, variable.type().symbolType().name()),
+          getSecondaryLocations(), 0);
+        blockAlreadyFlagged = true;
+      }
+    }
+
+    @Override
+    public void scanTree(Tree tree) {
+      super.scanTree(tree);
+    }
```
christophe-zurn-sonarsource: You could remove this override as it does not do anything

jblievremont: It does something very subtle: it makes the method ``public`` instead of ``protected`` :)

PR #2304, opened 2018-10-31
SONARJAVA-2878 S2115: add secondary location

andrea-guarino-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/helpers/IdentifierUtils.java
```diff
@@ -51,4 +56,15 @@ private static boolean isStrictAssignmentOrDeclaration(ExpressionTree expression
     }
     return true;
   }
+
+  public static class ValueAndExpression <T> {
```
andrea-guarino-sonarsource: @benzonico Don't really like this name but couldn't think of a better one. Moreover, probably it would be better to have a List of ExpressionTree representing the "resolution path" but I prefer to wait to work on the ticket about refactoring of `IdentifierUtils`.

PR #2313, opened 2018-11-01
SONARJAVA-2943 Rule S2139: Exceptions should be either logged or rethrown but not both

andrea-guarino-sonarsource:


java-checks/src/test/files/checks/LoggedRethrownExceptionsCheck.java
```diff
@@ -0,0 +1,66 @@
+import java.util.logging.Logger;
+import java.util.logging.Level;
+import java.sql.SQLException;
+
+class A {
+
+  public foo() {
```
christophe-zurn-sonarsource: Method is missing return type

java-checks/src/main/java/org/sonar/java/checks/LoggedRethrownExceptionsCheck.java
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    CatchTree catchTree = (CatchTree) tree;
+    boolean isRethrowing = false;
+    boolean isLogging = false;
+    List<Location> secondaryLocations = new ArrayList<>();
+    for (StatementTree statementTree : catchTree.block().body()) {
+      if (statementTree.is(Tree.Kind.THROW_STATEMENT)) {
+        secondaryLocations.add(new Location("", ((ThrowStatementTree) statementTree).expression()));
+        isRethrowing = true;
```
christophe-zurn-sonarsource: I think you could just report issue here if error was logged and return, as having statements after a throw is not valid java code

java-checks/src/main/java/org/sonar/java/checks/LoggedRethrownExceptionsCheck.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private static boolean isLoggingMethod(StatementTree statementTree, String exceptionIdentifier) {
+    if (!statementTree.is(Tree.Kind.EXPRESSION_STATEMENT)) {
+      return false;
+    }
+    ExpressionTree expression = ((ExpressionStatementTree) statementTree).expression();
+    if (expression.is(Tree.Kind.METHOD_INVOCATION)) {
+      MethodInvocationTree mit = (MethodInvocationTree) expression;
+      return JAVA_UTIL_LOGGER.matches(mit) && mit.arguments().stream().anyMatch(
+        param -> param.is(Tree.Kind.IDENTIFIER) && ((IdentifierTree)param).name().equals(exceptionIdentifier));
```
christophe-zurn-sonarsource: I would look for any use of the exception symbol inside the arguments, otherwise you will miss the following cases:
```
      logger.log(Level.ALL, "MyError: " + e);
      logger.log(Level.ALL, e.getMessage());
      throw new MySQLException(contextInfo, e);
```

java-checks/src/main/java/org/sonar/java/checks/LoggedRethrownExceptionsCheck.java
```diff
@@ -0,0 +1,122 @@
...
+import org.sonar.plugins.java.api.tree.ThrowStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+import static org.sonar.plugins.java.api.JavaFileScannerContext.Location;
+
+@Rule(key = "S2139")
+public class LoggedRethrownExceptionsCheck extends IssuableSubscriptionVisitor {
+  private static final String JAVA_UTIL_LOGGING_LOGGER = "java.util.logging.Logger";
+  private static final String SLF4J_LOGGER = "org.slf4j.Logger";
+  private static final List<MethodMatcher> LOGGING_METHODS = Collections.unmodifiableList(Arrays.asList(
```
christophe-zurn-sonarsource: You could use a `MethodMatcherCollection` instead

java-checks/src/main/java/org/sonar/java/checks/LoggedRethrownExceptionsCheck.java
```diff
@@ -0,0 +1,122 @@
...
+    boolean isExceptionIdentifierUsed = false;
+
+
+    ExceptionUsageVisitor(IdentifierTree exceptionIdentifier) {
+      this.exceptionIdentifier = exceptionIdentifier;
+    }
+
+    @Override
+    public void visitIdentifier(IdentifierTree tree) {
+      if (tree.name().equals(exceptionIdentifier.name())) {
```
christophe-zurn-sonarsource: There is no need to recheck equality if `isExceptionIdentifierUsed` is already true

java-checks/src/test/files/checks/LoggedRethrownExceptionsCheck.java
```diff
@@ -0,0 +1,182 @@
+import java.util.logging.Logger;
+import java.util.logging.Level;
+import java.sql.SQLException;
+import org.slf4j.Logger;
```
christophe-zurn-sonarsource: This import conflicts with the previous one `java.util.logging.Logger`

PR #2319, opened 2018-11-06
SONARJAVA-2955 Fix implementation of S4838

jblievremont:
* Add support for iteration over arrays
* Fix computation of collection type
* Fix handling of raw collection types


java-checks/src/main/java/org/sonar/java/checks/ForLoopVariableTypeCheck.java
```diff
@@ -60,20 +64,39 @@ public void visitNode(Tree tree) {
 
     if (collectionItemType != null && !isMostPreciseType(variableType, collectionItemType)) {
       // Second pass: check if the variable is down-cast in the statement block
-      DownCastVisitor downCastVisitor = new DownCastVisitor(actualStatement, actualStatement.variable(), collectionItemType);
-      downCastVisitor.setContext(context);
-      downCastVisitor.scanTree(((ForEachStatement) tree).statement());
+      DownCastVisitor downCastVisitor = new DownCastVisitor(actualStatement.variable().symbol());
+      actualStatement.accept(downCastVisitor);
```
benzonico: we only need to visit the statement of the foreachstatement. not the expression or variable.

PR #2325, opened 2018-11-19
SONARJAVA-2969 RSPEC-5042 Expanding archive files is security-sensitive

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/security/ZipEntryCheck.java
```diff
@@ -0,0 +1,73 @@
...
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = "S5042")
+public class ZipEntryCheck extends BaseTreeVisitor implements JavaFileScanner {
+
+  private static final Set<String> ZIP_ENTRY_TYPES = Stream.of(
```
benzonico: keep it simple : `Arrays.asList` ?

java-checks/src/test/files/checks/security/ZipEntryCheck.java
```diff
@@ -0,0 +1,52 @@
+package org.foo;
+
+import java.util.Enumeration;
+import java.util.jar.JarEntry;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import org.apache.commons.compress.archivers.ArchiveEntry;
+import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
+
+abstract class A {
```
benzonico: minor detail but for sake of documentation : having one non compliant and a compliant cases with fields (not only method parameters and local vars).

java-checks/src/main/java/org/sonar/java/checks/security/ZipEntryCheck.java
benzonico: (and just for fun and the thought experiment, you could imagine to have a field which is a Predicate<Type> and have those two strings encoded in the predicate).

PR #2331, opened 2018-11-26
SONARJAVA-2966 Fix NPE on S4248 on pattern used in enums

vilchik-elena:


java-checks/src/main/java/org/sonar/java/checks/RegexPatternsNeedlesslyCheck.java
```diff
@@ -71,14 +72,10 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
 
   private static boolean storedInStaticFinal(MethodInvocationTree mit) {
     Tree tree = mit.parent();
-    while (!tree.is(Tree.Kind.VARIABLE, Tree.Kind.CLASS, Tree.Kind.ASSIGNMENT, Tree.Kind.METHOD)) {
+    while (!tree.is(Kind.VARIABLE, Kind.CLASS, Kind.ASSIGNMENT, Kind.METHOD, Kind.ENUM)) {
```
m-g-sonar: I have the feeling that we should be careful and also check for interfaces and annotations (`Tree.Kind.INTERFACE`, `Tree.Kind.ANNOTATION_TYPE`, as both of them are ClassTrees). It would however be pure defensive programming, because I can not manage to write a compiling example.

PR #2334, opened 2018-11-27
SONARJAVA-2976 Adjust rule S5042

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/security/ZipEntryCheck.java
```diff
@@ -46,16 +46,24 @@ public void scanFile(JavaFileScannerContext context) {
...
+    if (!isField(tree) && tree.initializer() == null && IS_ZIP_ENTRY.test(tree.symbol().type())) {
       // skip the visit in case of issue already reported on the variable
       context.reportIssue(this, tree, ISSUE_MESSAGE);
     } else {
       super.visitVariable(tree);
     }
   }
 
+  private boolean isField(VariableTree tree) {
+    return false;
```
m-g-sonar: this is so wrong!

PR #2335, opened 2018-11-27
Fix xml-parser related issues

vilchik-elena:


m-g-sonar:
LGTM! thanks a lot for the fix!

PR #2341, opened 2018-12-05
return false as soon as a final equals method is found

troosan:
This PR fixes issue reported here: https://community.sonarsource.com/t/rule-s2160-false-positive-when-class-has-one-parent-with-non-final-equals-method/4885


troosan:
tests seem to have failed. I had the same locally, after relaunching they passed

troosan:
issue created here: https://community.sonarsource.com/t/rule-s2160-false-positive-when-class-has-one-parent-with-non-final-equals-method/4885/2

troosan:
@m-g-sonar never mind it seems the dependency exists for Jenkins.

m-g-sonar:
@troosan Yeah, sorry for that. We are currently reworking the XML handling of the SonarJava plugin... And I did not realized we are relying on a dependency which is not yet on maven central. So you won't be able to build locally. That's pretty bad.

Currently, the new artifact is only visible from our own infrastructure. This should get back to normal in a few days. In the meantime, travis-ci will pass, as we configured it to rely on our own base.

If you want to check your build locally, a workaround could be to:
1.  build the master branch of the following project locally: https://github.com/SonarSource/sonar-analyzer-commons
2. change the associated version of the dependency in the parent pom of SonarJava, in order to rely on a snapshot version

m-g-sonar:
@troosan thanks again for your contribution! I merged the pull request (#2375) which included your fix, rebased. I slightly reworked the rule at the same time. Your commit 9a33d05319b230a00b2386a3b595db8892bcb9a7 is now part of our master. I'm consequently closing this PR.

Thanks again for helping us making SonarJava every time a bit better.

Cheers,
Michael

PR #2344, opened 2018-12-18
Start migration to XML parser from analyzer-commons

vilchik-elena:


java-checks/src/main/java/org/sonar/java/checks/xml/AbstractXPathBasedCheck.java
```diff
@@ -0,0 +1,46 @@
...
+    try {
+      return XPATH.compile(expression);
+    } catch (XPathExpressionException e) {
+      // fixme
+    }
+    return null;
+  }
+
+  @Override
+  protected void scanFile(XmlFile file) {
```
m-g-sonar: We don't need to override this method. It will be done by checks.

java-checks/src/main/java/org/sonar/java/checks/xml/AbstractXPathBasedCheck.java
```diff
@@ -0,0 +1,46 @@
...
+
+public abstract class AbstractXPathBasedCheck extends SonarXmlCheck {
+
+  private static final XPath XPATH = XPathFactory.newInstance().newXPath();
+
+  protected static XPathExpression getXPathExpression(String expression){
+    try {
+      return XPATH.compile(expression);
+    } catch (XPathExpressionException e) {
+      // fixme
```
m-g-sonar: I would prefer to throw an exception directly rather than returning null. Even if we are going to rework it, it's cleaner for the time being.

PR #2347, opened 2018-12-18
SONARJAVA-2992 rewrite rule S3419

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/xml/maven/GroupIdNamingConventionCheck.java
```diff
@@ -35,19 +43,36 @@
...
   }
 
-  @Override
-  protected NamedLocatedAttribute getTargetedLocatedAttribute(MavenProject mavenProject) {
-    return new NamedLocatedAttribute("groupId", mavenProject.getGroupId());
+  private Pattern getPattern() {
+    if (pattern == null) {
+      try {
+        pattern = Pattern.compile(regex, Pattern.DOTALL);
+      } catch (IllegalArgumentException e) {
```
vilchik-elena: this looks weird, but I see it's a copy paste so ok:)

PR #2354, opened 2018-12-19
SONARJAVA-2987 Rewrite rule S3369

amelie-renard-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/xml/web/AbstractWebXmlXPathBasedCheck.java
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java.checks.xml.web;
+
+import org.sonar.java.checks.xml.AbstractXPathBasedCheck;
+import org.sonarsource.analyzer.commons.xml.XmlFile;
+
+public abstract class AbstractWebXmlXPathBasedCheck extends AbstractXPathBasedCheck {
```
m-g-sonar: This is a first step. I think we could have this logic a bit more wider, to run some checks only when matching a given file name, or filename predicate. WDYT?

This would allow to simplify checks targeting `pom.xml` files, and maybe others.

In the meantime, I think it's OK to merge this already.

java-checks/src/main/java/org/sonar/java/checks/xml/web/SecurityConstraintsInWebXmlCheck.java
```diff
@@ -19,30 +19,28 @@
...
 
-  private XPathExpression securityConstraintExpression;
+  private XPathExpression securityConstraintExpression = getXPathExpression(WEB_XML_ROOT + "/security-constraint");
 
-  @Override
-  public void precompileXPathExpressions(XmlCheckContext context) {
-    this.securityConstraintExpression = context.compile(WEB_XML_ROOT + "/security-constraint");
+  private boolean hasNoSecurityConstraint(XmlFile file) {
+    NodeList nodeList = evaluate(securityConstraintExpression, file.getNamespaceUnawareDocument());
+    return (nodeList.getLength() == 0);
```
m-g-sonar: no need for the parenthesis

java-checks/src/main/java/org/sonar/java/checks/xml/web/SecurityConstraintsInWebXmlCheck.java
```diff
@@ -19,30 +19,28 @@
...
   }
 
   @Override
-  public void scanWebXml(XmlCheckContext context) {
-    if (hasNoSecurityConstraint(context)) {
-      reportIssueOnFile("Add \"security-constraint\" elements to this descriptor.");
+  public void scanWebXml(XmlFile file) {
+    if (hasNoSecurityConstraint(file)) {
+      List<Integer> secondaryLocationLines = new ArrayList<>();
+      reportIssueOnFile("Add \"security-constraint\" elements to this descriptor.", secondaryLocationLines);
```
m-g-sonar: instead of an empty ArrayList, I would use `Collections.emptyList()`.

java-checks/src/main/java/org/sonar/java/checks/xml/web/AbstractWebXmlXPathBasedCheck.java
amelie-renard-sonarsource: Yes, I totally agree with that.

PR #2355, opened 2018-12-19
SONARJAVA-2994 rewrite rule S3421

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/xml/maven/DeprecatedPomPropertiesCheck.java
```diff
@@ -19,35 +19,43 @@
...
-      for (Node textNode : context.evaluateOnDocument(textsExpression)) {
-        String text = textNode.getFirstChild().getNodeValue();
-        while (StringUtils.contains(text, POM_PROPERTY_PREFIX)) {
-          String property = extractPropertyName(text);
-          reportIssue(textNode, "Replace \"pom." + property + "\" with \"project." + property + "\".");
-          text = skipFirstProperty(text);
-        }
+  private void checkText(Node textNode) {
+    NodeList childNodes = textNode.getChildNodes();
+    for (int j = 0; j < childNodes.getLength(); j++) {
```
amelie-renard-sonarsource: In the previous version, we checked only the first child of the node.
In this version, we iterate on all the children of the node -> is it normal?

m-g-sonar: Yes, this is something new.

With the document generated by the new parser, I was not able to catch the modified Lorem Ipsum text node. This comes from the fact that the first text node of the element is now only the spaces and new line preceding the `<!-- Noncompliant ...` comment. I then realized that there were potentially multiple text nodes within the parent node.

PR #2356, opened 2018-12-19
SONARJAVA-2990 Rewrite maven rule S3422 (DependencyWithSystemScopeCheck)

vilchik-elena:


java-checks/src/main/java/org/sonar/java/checks/xml/maven/DependencyWithSystemScopeCheck.java
```diff
@@ -19,43 +19,59 @@
...
-          message = "Update this scope and remove the \"systemPath\".";
-        }
-        context.reportIssue(this, scope.startLocation().line(), message, secondaries);
-      }
+  protected void scanFile(XmlFile xmlFile) {
+    if (!"pom.xml".equalsIgnoreCase(xmlFile.getInputFile().filename())) {
+      return;
+    }
+
+    NodeList dependencies = evaluate(dependencyExpression, xmlFile.getNamespaceUnawareDocument());
```
m-g-sonar: you can now rely on the new methods from `AbstracXPathBasedCheck` to directly iterate over list of nodes

PR #2357, opened 2018-12-20
SONARJAVA-2991 Rewrite maven rule S3423 (PomElementOrderCheck)

vilchik-elena:


java-checks/src/main/java/org/sonar/java/checks/xml/maven/PomElementOrderCheck.java
```diff
@@ -19,68 +19,82 @@
...
+import org.w3c.dom.Node;
 
 @Rule(key = "S3423")
-public class PomElementOrderCheck implements PomCheck {
+public class PomElementOrderCheck extends SonarXmlCheck {
 
-  private static final Comparator<LocatedTree> LINE_COMPARATOR = (l1, l2) -> Integer.compare(l1.startLocation().line(), l2.startLocation().line());
+  private static final Comparator<Node> LINE_COMPARATOR = Comparator.comparingInt(n -> XmlFile.nodeLocation(n).getStartLine());
+
+  private static final List<String> REQUIRED_ORDER = ImmutableList.of(
```
m-g-sonar: You can use `Arrays.asList()`, it's also `unmodifiable` and native from jdk.

java-checks/src/main/java/org/sonar/java/checks/xml/maven/PomElementOrderCheck.java
```diff
@@ -19,68 +19,82 @@
...
+      return;
+    }
+
+    checkPositions(xmlFile.getDocument().getDocumentElement());
+  }
 
-    if (!issues.isEmpty()) {
-      context.reportIssue(this, project.startLocation().line(), "Reorder the elements of this pom to match the recommended order.", issues);
+  private static Optional<Node> getChildElementByName(String name, List<Node> children) {
+    for (Node child : children) {
```
m-g-sonar: can't you write this with a stream?
```java
return children.stream()
  .filter(child -> child.getNodeType() == Node.ELEMENT_NODE)
  .map(Element.class::cast)
  .filter(element -> element.getTagName().equals(name))
  .findFirst();
```

PR #2360, opened 2018-12-20
SONARJAVA-2988 Rewrite rule S3355

amelie-renard-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/xml/web/ValidationFiltersCheck.java
```diff
@@ -19,46 +19,30 @@
...
+    evaluateAsList(filterNamesFromFilterExpression, file.getNamespaceUnawareDocument())
+      .forEach(node -> {
+        String filterName = getStringValue(node);
+        if (!filtersInMapping.contains(filterName)) {
+          reportIssue(node, "\"" + filterName + "\" should have a mapping.");
+        }
+      });
   }
 
   private static String getStringValue(Node node) {
```
vilchik-elena: If tag is empty we will have NPE here. Could we make this impl safer?

java-checks/src/main/java/org/sonar/java/checks/xml/web/ValidationFiltersCheck.java
```diff
@@ -39,14 +39,15 @@ public void scanWebXml(XmlFile file) {
...
+        if (!filterName.isEmpty() && !filtersInMapping.contains(filterName)) {
           reportIssue(node, "\"" + filterName + "\" should have a mapping.");
         }
       });
   }
 
   private static String getStringValue(Node node) {
-    return node.getFirstChild().getNodeValue();
+    Node firstChild = node.getFirstChild();
+    return (firstChild == null) ? "" : firstChild.getNodeValue();
```
vilchik-elena: you didn't add test for this case

And also it's better to use Optional instead of empty string

PR #2363, opened 2018-12-21
SONARJAVA-2985 Rewrite rule S3373

amelie-renard-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/xml/struts/ActionNumberCheck.java
```diff
@@ -41,30 +38,25 @@
...
-    for (Node action : context.evaluateOnDocument(actionsExpression)) {
-      Iterable<Node> extraForwards = Iterables.skip(context.evaluate(forwardsFromActionExpression, action), maximumForwards);
-      if (!Iterables.isEmpty(extraForwards)) {
-        List<XmlCheckContext.XmlDocumentLocation> secondaries = new LinkedList<>();
-        for (Node forward : extraForwards) {
-          secondaries.add(new XmlCheckContext.XmlDocumentLocation("Move this forward to another action.", forward));
+  protected void scanFile(XmlFile xmlFile) {
+    evaluateAsList(actionsExpression, xmlFile.getNamespaceUnawareDocument())
+      .forEach(node -> {
+        List<Secondary> secondaries = new ArrayList<>();
```
m-g-sonar: instead of adding element to the list, you can use a map directly to build this list:
```java
List<Secondary> secondaries = evaluateAsList(forwardsFromActionExpression, node).stream()
  .skip(maximumForwards)
  .map(forward -> new Secondary(forward, "Move this forward to another action."))
```

java-checks/src/main/java/org/sonar/java/checks/xml/struts/ActionNumberCheck.java
```diff
@@ -41,30 +38,25 @@
...
-  public void scanFileWithXPathExpressions(XmlCheckContext context) {
-    for (Node action : context.evaluateOnDocument(actionsExpression)) {
-      Iterable<Node> extraForwards = Iterables.skip(context.evaluate(forwardsFromActionExpression, action), maximumForwards);
-      if (!Iterables.isEmpty(extraForwards)) {
-        List<XmlCheckContext.XmlDocumentLocation> secondaries = new LinkedList<>();
-        for (Node forward : extraForwards) {
-          secondaries.add(new XmlCheckContext.XmlDocumentLocation("Move this forward to another action.", forward));
+  protected void scanFile(XmlFile xmlFile) {
+    evaluateAsList(actionsExpression, xmlFile.getNamespaceUnawareDocument())
+      .forEach(node -> {
```
m-g-sonar: I'm not a big fan of nested lambdas. Can you extract this part into a dedicated method?
It will then be simply something such as:
```
 evaluateAsList(actionsExpression, xmlFile.getNamespaceUnawareDocument()).forEach(this::checkAction);
```

PR #2366, opened 2018-12-24
SONARJAVA-2999 Add progress report to XML analyzer and drop sslr-squid-bridge

m-g-sonar:


sonar-java-plugin/src/main/java/org/sonar/plugins/java/XmlFileSensor.java
```diff
@@ -56,8 +61,32 @@ public XmlFileSensor(CheckFactory checkFactory) {
...
+
+    if (inputFiles.isEmpty()) {
+      return;
+    }
+
+    ProgressReport progressReport = new ProgressReport("Report about progress of Java XML analyzer", TimeUnit.SECONDS.toMillis(10));
+    progressReport.start(inputFiles.stream().map(InputFile::toString).collect(Collectors.toList()));
+
+    boolean successfullyCompleted = false;
+    boolean cancelled = false;
```
vilchik-elena: do we really need this variable?

m-g-sonar: Good catch, I forgot to handle analysis cancellation.

PR #2370, opened 2018-12-28
SONARJAVA-3000 handle cases when bytecode is not in sync

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/MinMaxRangeCheck.java
```diff
@@ -161,14 +161,19 @@ public ProgramState checkPostStatement(CheckerContext context, Tree syntaxNode)
...
+    SymbolicValue constant = programState.getValue(identifier);
+    if (constant == null) {
+      return programState;
+    }
+    NumericalConstraint numericalConstraint = programState.getConstraint(constant, NumericalConstraint.class);
+    Optional<Object> constantValue = ((JavaSymbol.VariableJavaSymbol) identifier).constantValue();
+    if (numericalConstraint == null && constantValue.isPresent()) {
+      Object value = constantValue.get();
+      if (value instanceof Number) {
+        return programState.addConstraint(constant, new NumericalConstraint((Number) value));
```
benzonico: let's make this functional style if w e have an optional : 
```
if (numericalConstraint == null) {
return constantValue.filter(v -> v instance of Number)
.map( v -> programState.addConstraint(constant, new NumericalConstraint((Number) v)))
.orElse(programState);
}

PR #2373, opened 2019-01-02
SONARJAVA-2950 CastArithmeticOperandCheck

amelie-renard-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/CastArithmeticOperandCheck.java
```diff
@@ -103,6 +105,14 @@ public void visitBinaryExpression(BinaryExpressionTree tree) {
     }
   }
 
+  @Override
+  public void visitNewClass(NewClassTree tree) {
+    if (tree.is(Tree.Kind.NEW_CLASS)) {
```
m-g-sonar: the `tree` is necessarily a `NewClassTree` here, so you already know it's going to be of the good kind. You can drop the `if`.

java-checks/src/test/files/checks/CastArithmeticOperandCheck.java
```diff
@@ -58,3 +58,13 @@ void test_int_division() {
     double c = ((1 + 0) / 2 + 0) / 5.0; // Noncompliant [[sc=17;ec=28]] {{Cast one of the operands of this integer division to a "double".}}
   }
 }
+
+class Date {
```
m-g-sonar: I think it would be simpler to use the java native date: `java.util.Date`, it has a constructor which use a `long`.
Furthermore, this java code does not compile (you are not defining the body of the constructor).

just add the following method to the previous class:
```java
void test_constructors() {
  java.io.Date date1 = new java.io.Date(2 + 1); // Noncompliant {{Cast one of the operands of this addition operation to a "long".}}
  java.io.Date date2 = new java.io.Date("today"); // Compliant
}
```

PR #2374, opened 2019-01-02
SONARJAVA-2947 CompareObjectWithEqualsCheck

amelie-renard-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/CompareObjectWithEqualsCheck.java
```diff
@@ -66,16 +66,24 @@ private static boolean isEquals(MethodTree tree) {
...
         && bothImplementsEqualsMethod(leftOpType, rightOpType)
-        && neitherIsPublicStaticFinal(tree.leftOperand(), tree.rightOperand())) {
+        && neitherIsPublicStaticFinal(leftExpression, rightExpression)) {
         context.reportIssue(this, tree.operatorToken(), "Use the \"equals\" method if value comparison was intended.");
       }
     }
   }
 
+  private static boolean neitherIsThis(ExpressionTree leftExpression, ExpressionTree rightExpression) {
+    return (!(leftExpression.is(Tree.Kind.IDENTIFIER) && "this".equals(((IdentifierTree) leftExpression).name()))
```
m-g-sonar: You should skip parenthesis, otherwise we are going to miss the cases such as: `myObj == (this)`.
I would also inline into a simpler method checking an expression, called for both operands:

```java
  private static boolean neitherIsThis(ExpressionTree leftExpression, ExpressionTree rightExpression) {
     return !isThis(leftExpression) && !isThis(rightExpression);
  }

  private static boolean isThis(ExpressionTree expression) {
    ExpressionTree newExpression = ExpressionUtils.skipParentheses(expression);
    return newExpression.is(Tree.Kind.IDENTIFIER) && "this".equals(((IdentifierTree) newExpression).name();
  }
```

I'm quite surprised we don't have this in the API already. Could be valuable! It seems that we have already half a dozen of similar method invocations in the checks.

PR #2376, opened 2019-01-03
SONARJAVA-3003 S4973 CompareStringsBoxedTypesWithEqualsCheck

amelie-renard-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/CompareWithEqualsVisitor.java
```diff
@@ -0,0 +1,65 @@
...
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    this.context = context;
+    if (context.getSemanticModel() != null) {
+      scan(context.getTree());
+    }
+  }
+
+  @Override
+  public void visitMethod(MethodTree tree) {
```
m-g-sonar: make it final. You don't want classes extending it to override this, as it's driving the analysis.

java-checks/src/main/java/org/sonar/java/checks/CompareWithEqualsVisitor.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private static boolean isEquals(MethodTree tree) {
+    return ((MethodTreeImpl) tree).isEqualsMethod();
+  }
+
+  protected static boolean isNullComparison(Type leftOpType, Type rightOpType) {
+    return isBot(leftOpType) || isBot(rightOpType);
+  }
+
+  protected static boolean isBot(Type type) {
```
m-g-sonar: why not private?

java-checks/src/main/java/org/sonar/java/checks/CompareObjectWithEqualsCheck.java
```diff
@@ -75,7 +49,7 @@ public void visitBinaryExpression(BinaryExpressionTree tree) {
         && neitherIsThis(leftExpression, rightExpression)
         && bothImplementsEqualsMethod(leftOpType, rightOpType)
         && neitherIsPublicStaticFinal(leftExpression, rightExpression)) {
-        context.reportIssue(this, tree.operatorToken(), "Use the \"equals\" method if value comparison was intended.");
+        reportIssue(this, tree.operatorToken());
```
m-g-sonar: you don't need to provide `this` here, just use it in the abstract class.

java-checks/src/main/java/org/sonar/java/checks/CompareWithEqualsVisitor.java
```diff
@@ -0,0 +1,65 @@
...
+import org.sonar.java.model.declaration.MethodTreeImpl;
+import org.sonar.java.resolve.JavaType;
+import org.sonar.plugins.java.api.JavaFileScanner;
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.SyntaxToken;
+
+public class CompareWithEqualsVisitor extends BaseTreeVisitor implements JavaFileScanner {
```
m-g-sonar: this class should be abstract

java-checks/src/main/java/org/sonar/java/checks/CompareWithEqualsVisitor.java
```diff
@@ -0,0 +1,65 @@
...
+  protected static boolean isNullComparison(Type leftOpType, Type rightOpType) {
+    return isBot(leftOpType) || isBot(rightOpType);
+  }
+
+  protected static boolean isBot(Type type) {
+    return ((JavaType) type).isTagged(JavaType.BOT);
+  }
+
+  protected void reportIssue(JavaFileScanner scanner, SyntaxToken opToken) {
+    context.reportIssue(scanner, opToken, "Use the \"equals\" method if value comparison was intended.");
```
m-g-sonar: Use `this` instead of scanner here.

its/ruling/src/test/resources/jdk6/squid-S4973.json
```diff
@@ -0,0 +1,151 @@
+{
+'jdk6:java/awt/BorderLayout.java':[
+836,
```
m-g-sonar: I guess all these issues are going to disappear from other rule once updated to not overlap? Did you check?

java-checks/src/main/java/org/sonar/java/checks/CheckList.java
```diff
@@ -660,7 +660,8 @@ private CheckList() {
       PrimitivesMarkedNullableCheck.class,
       ForLoopVariableTypeCheck.class,
       ReplaceGuavaWithJava8Check.class,
-      LoggedRethrownExceptionsCheck.class);
+      LoggedRethrownExceptionsCheck.class,
+      CompareStringsBoxedTypesWithEqualsCheck.class);
```
m-g-sonar: maybe put it next to the other one in the list.

java-checks/src/main/java/org/sonar/java/checks/CompareWithEqualsVisitor.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+
+  @Override
+  public void visitMethod(MethodTree tree) {
+    if (!isEquals(tree)) {
+      super.visitMethod(tree);
+    }
+  }
+
```
m-g-sonar: I would also override `visitBooleanExpression`, make it final to guarantee no change in subclasses, and only call an abstract method when its `==`or `!=`:
```java
  @Override
  public void visitBinaryExpression(BinaryExpressionTree tree) {
    super.visitBinaryExpression(tree);
    if (tree.is(Tree.Kind.EQUAL_TO, Tree.Kind.NOT_EQUAL_TO)) {
      checkEquality(tree);
    }
  }

  public abstract void checkEqualityExpression(BinaryExpressionTree tree);
```

You can then simply implement `checkEqualityExpression` in each of the sub classes.

PR #2377, opened 2019-01-04
SONARJAVA-2960 S3551 SynchronizedOverrideCheck

amelie-renard-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/SynchronizedOverrideCheck.java
```diff
@@ -0,0 +1,62 @@
...
+    if (!hasSemantic()) {
+      return;
+    }
+
+    MethodTree methodTree = (MethodTree) tree;
+    Symbol.MethodSymbol overriddenSymbol =  methodTree.symbol().overriddenSymbol();
+    if (overriddenSymbol == null || overriddenSymbol.isUnknown()) {
+      return;
+    }
+    MethodTree overridenMethodTree = overriddenSymbol.declaration();
```
m-g-sonar: this won't work. in the immense majority of cases, the declaration of the parent method won't be in the same file, and so you won't have access to the modifier.

java-checks/src/main/java/org/sonar/java/checks/SynchronizedOverrideCheck.java
```diff
@@ -0,0 +1,62 @@
...
+    if (null != overridenMethodTree && isSynchronized(overridenMethodTree) && !isSynchronized(methodTree)) {
```
m-g-sonar: Because there is nothing in the semantic public API, you will need to rely on deeper logic (please remove the fully qualified name and use the imports):
```java
private static boolean isSynchronized(Symbol methodSymbol) {
  return org.sonar.java.resolve.Flags.isFlagged(
    ((org.sonar.java.resolve.JavaSymbol) methodSymbol).flags(), 
    org.sonar.java.resolve.Flags.SYNCHRONIZED);
}
```

java-checks/src/test/files/checks/SynchronizedOverrideCheck.java
```diff
@@ -0,0 +1,21 @@
...
+
+  @java.lang.Override
+  public void f2(){} // Compliant
+
+  public void f3(){} // Noncompliant {{Make this method "synchronized" to match the parent class implementation.}}
+}
+
+class C extends B{
+  @java.lang.Override
+  public void f1(){} // Compliant
```
m-g-sonar: I don't know about this one. I think it deserve an issue as well... because if we don't then the issue will appear only after fixing the first one. Which is OK, but could be painful sometime. Up to you to decide, I let you judge.

java-checks/src/main/java/org/sonar/java/checks/SynchronizedOverrideCheck.java
```diff
@@ -0,0 +1,62 @@
...
+      reportIssue(methodTree.simpleName(), "Make this method \"synchronized\" to match the parent class implementation.");
```
m-g-sonar: what about adding a secondary location on the overridden, **if it is present in the file**?

java-checks/src/test/files/checks/SynchronizedOverrideCheck.java
amelie-renard-sonarsource: I agree with you but on another side you may not have access or you may do not see if `synchronized` is written far away in the hierarchy...
Both choices have drawbacks, I think I will keep this one for now.

java-checks/src/main/java/org/sonar/java/checks/SynchronizedOverrideCheck.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+    if (isSynchronized(overriddenSymbol) && !isSynchronized(methodSymbol)) {
+      MethodTree overridenMethodTree = overriddenSymbol.declaration();
+      if (null != overridenMethodTree) {
+        List<JavaFileScannerContext.Location> secondaryLocations = Collections.singletonList(
+          new JavaFileScannerContext.Location("", overridenMethodTree.simpleName()));
+        reportIssue(methodTree.simpleName(), MESSAGE, secondaryLocations, null);
+      }
+      else {
+        reportIssue(methodTree.simpleName(), MESSAGE);
```
m-g-sonar: what about calling `reportIssue()` only once, and only set the secondary list in the `if-then-else` ? I think it would be a bit nicer to read.
```java
List<JavaFileScannerContext.Location> secondaries = Collections.emptyList();
MethodTree overridenMethodTree = overriddenSymbol.declaration();
if (overridenMethodTree != null) {
  secondaries = Collections.singletonList(new JavaFileScannerContext.Location("", overridenMethodTree.simpleName()));
}
reportIssue(methodTree.simpleName(), MESSAGE, secondaries, null);

File: java-checks/src/main/java/org/sonar/java/checks/SynchronizedOverrideCheck.java [hunk 6]
```diff
@@ -0,0 +1,75 @@
...
+
+    MethodTree methodTree = (MethodTree) tree;
+    Symbol.MethodSymbol methodSymbol = methodTree.symbol();
+    Symbol.MethodSymbol overriddenSymbol =  methodSymbol.overriddenSymbol();
+    if (overriddenSymbol == null || overriddenSymbol.isUnknown()) {
+      return;
+    }
+    if (isSynchronized(overriddenSymbol) && !isSynchronized(methodSymbol)) {
+      MethodTree overridenMethodTree = overriddenSymbol.declaration();
+      if (null != overridenMethodTree) {
```
m-g-sonar: It seems to me that the case where we don't have the declaration tree of the overriden symbol is not covered in the unit test.

By looking at the result of the ruling, it should be easy to reproduce with any class extending `Throwable` (or `Exception`) and overriding `getCause()` without adding the `synchronized` modifier.

PR #2384, opened 2019-01-10
[DO-NOT-MERGE] Proof of Concept with early work on handling 'Tests as First Citizens'

m-g-sonar:


vilchik-elena:
LGTM

m-g-sonar:
Rules which need to be updated for tests according to first results:
* (S100) Method name convention -> should ignore method annotated with `@Test` [junit 4 & 5 & testng]
* (S112) method throwing generic exception -> should ignore methods annotated with `@Test`
* (S2259) NPE rule -> should consider `assertThat(obj).isNotNull();`
* (S1228) package-info rule -> should not raise issue on test package, this file is useless there

PR #2390, opened 2019-01-17
Update S1126_java.html

flm1972:
In the compliant solution bar is the function to call, not foo (which would be recurrent)


benzonico:
Hi, 

Thanks for the feedback. As those files are generated I fixed the origin of this trouble in : https://jira.sonarsource.com/browse/RSPEC-4948 

This should be fixed in next version of SonarJava.
As a consequence I am closing this PR.

PR #2397, opened 2019-01-18
SONARJAVA-3011 Improve S2259 (NPE check) to handle Assertions from unit tests

ghost:


java-frontend/src/main/java/org/sonar/java/se/xproc/BehaviorCache.java
```diff
@@ -94,6 +94,8 @@
     "com.google.common.base.Strings#isNullOrEmpty",
     "com.google.common.base.Platform#stringIsNullOrEmpty",
 
+    "org.junit.Assert#assertNotNull",
```
vilchik-elena: junit5

java-frontend/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -100,6 +103,12 @@ public ProgramState checkPreStatement(CheckerContext context, Tree syntaxNode) {
             return ps;
           }
         }
+        if (ASSERTJ_IS_NOT_NULL.matches((MethodInvocationTree) syntaxNode)) {
+          ps = handleAssertJIsNotNull(ps, context);
```
vilchik-elena: `checkAssertJ...`

PR #2404, opened 2019-02-01
SONARJAVA-3011 Support fest assertions and improve S2259

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -128,19 +134,41 @@ public ProgramState checkPreStatement(CheckerContext context, Tree syntaxNode) {
...
     }
-    SymbolicValue assertThatParameter = parentNode.programState.peekValue();
+    // Retrieved the MethodInvocationTree corresponding to the 'assertThat()' coupled with 'isNotNull()'.
+    // Now searching for corresponding Node in Exploded Graph
+    ExplodedGraph.Node node = context.getNode();
+    Tree previousSyntaxNode;
+    do {
+      node = node.parent();
+      previousSyntaxNode = node.programPoint.syntaxTree();
+    } while (!previousSyntaxNode.equals(mit));
```
vilchik-elena: I'm scared of having infinite loop here in some corner case. Can we be more defensive here?

m-g-sonar: Okay adding a defensive way to exit the loop.

PR #2406, opened 2019-02-04
SONARJAVA-3014 CORS hotspot

benzonico:


java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S5122_java.html
```diff
@@ -0,0 +1,68 @@
...
+
+    @CrossOrigin(origins = "http://domain2.com") // Questionable
+    @RequestMapping(value = "/test1")
+    public ResponseEntity&lt;String&gt; test1() {
+        return ResponseEntity.ok().body("ok");
+    }
+}
+</pre>
+<h2>Exceptions</h2>
+<p>If more than one CORS related header is set per method/function, only one issue will be raised.</p>
```
christophe-zurn-sonarsource: This exception does not reflect current implementation. The exception should either be removed from RSPEC or implementation updated accordingly

java-checks/src/test/files/checks/CORSCheck.java
```diff
@@ -0,0 +1,41 @@
...
+  // === Spring MVC Controller annotation ===
+  @CrossOrigin(origins = "http://domain1.com") // Noncompliant [[sc=4;ec=15]] {{Make sure that enabling CORS is safe here.}}
+  @RequestMapping("")
+  public class TestController {
+    public String home(ModelMap model) {
+      model.addAttribute("message", "ok ");
+      return "view";
+    }
+
+    @CrossOrigin(origins = "http://domain2.com") // Noncompliant
```
christophe-zurn-sonarsource: If RSPEC exception is implemented, I would like to see a method with 2 annotations to ensure only one raises an issue

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S5122_java.html
benzonico: RSPEC has been modified thanks to your suggestion

PR #2412, opened 2019-02-05
SONARJAVA-2927 S2699 supports a parameter to configure custom Assertion Frameworks

christophe-zurn-sonarsource:


toniedzwiedz:
Thanks for introducing this. Looks very useful.

How about adding support for JUnit 5 Jupiter assertions by default? I'd expect them to be used very commonly.

```
org.junit.jupiter.api.Assertions.assertEquals;
org.junit.jupiter.api.Assertions.assertFalse;
org.junit.jupiter.api.Assertions.assertTrue;
```

Not sure this is the right place to do so. I got here via https://jira.sonarsource.com/browse/SONARJAVA-2927 and I've searched for other tickets pertaining to squid:S2699 and JUnit 5 / Jupiter but found none.

What do you think?

christophe-zurn-sonarsource:
Hello @toniedzwiedz ,

Thanks a lot for your suggestion, it is indeed a good idea and I've created a corresponding [Jira ticket](https://jira.sonarsource.com/browse/SONARJAVA-3034)!

Generally, if you have any suggestion, remark or feedback, it is better to make it on [our community forum](https://community.sonarsource.com/), as there are more people involved there :)

Cheers,
-Chris

PR #2418, opened 2019-02-07
SONARJAVA-3025 Do not use basic auth

benzonico:


java-checks/src/main/java/org/sonar/java/checks/BasicAuthCheck.java
```diff
@@ -0,0 +1,63 @@
...
+      String authentication = resolveAsStringConstant(arg);
+      if (authentication != null && authentication.startsWith("Basic")) {
+        reportIssue(arg, "Use a more secure method than basic authentication.");
+      }
+    }
+  }
+
+  private static ExpressionTree mostLeft(ExpressionTree arg) {
+    ExpressionTree res = arg;
+    while (res.is(Tree.Kind.PLUS)) {
```
christophe-zurn-sonarsource: What about skipping parenthesis, so that we still raise issue for the following:
`conn.setRequestProperty("Authorization", ("Basic " + encoding + encoding))`

java-checks/src/main/java/org/sonar/java/checks/BasicAuthCheck.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.sonar.java.checks.helpers.ConstantUtils.resolveAsStringConstant;
+
+@Rule(key = "S2647")
+public class BasicAuthCheck extends AbstractMethodDetection {
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return Arrays.asList(
+      MethodMatcher.create().typeDefinition("org.apache.http.message.AbstractHttpMessage").name("setHeader").withAnyParameters(),
+      MethodMatcher.create().typeDefinition("java.net.URLConnection").name("setRequestProperty").withAnyParameters()
```
christophe-zurn-sonarsource: We could also consider these 2 methods for the check:
```
org.apache.http.message.AbstractHttpMessage#addHeader(String,String)
java.net.URLConnection#addRequestProperty(String,String)
```

Eventually, this constructor as well `new org.apache.http.message.BasicHeader("Authorization", "Basic abc")`

PR #2419, opened 2019-02-07
SONARJAVA-3024 Cookie domain should have at least 2 levels

benzonico:


java-checks/src/main/java/org/sonar/java/checks/CookieDomainCheck.java
```diff
@@ -0,0 +1,50 @@
...
+      MethodMatcher.create().typeDefinition("javax.servlet.http.Cookie").name("setDomain").addParameter("java.lang.String"),
+      MethodMatcher.create().typeDefinition("java.net.HttpCookie").name("setDomain").addParameter("java.lang.String")
+    );
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    ExpressionTree arg = mit.arguments().get(0);
+    String domain = ConstantUtils.resolveAsStringConstant(arg);
+    if (domain != null && !domain.substring(1).contains(".")) {
```
christophe-zurn-sonarsource: Check will crash if code contains a call to `setDomain` method with empty string `myCookie.setDomain("");`

java-checks/src/test/files/checks/CookieDomainCheck.java
```diff
@@ -0,0 +1,15 @@
+import javax.servlet.http.Cookie;
+import java.net.HttpCookie;
+class A {
+ void foo() {
+   Cookie myCookie = new Cookie("name", "val");
+   myCookie.setDomain(".com"); // Noncompliant [[sc=23;ec=29]] {{Specify at least a second-level cookie domain.}}
+   myCookie.setDomain(".myDomain.com"); // Compliant
+ }
+ void HttpCookie(String domain) {
+   HttpCookie myCookie = new Cookie("name", "val");
```
christophe-zurn-sonarsource: Should it not be `new HttpCookie` instead?

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S3331_java.html
```diff
@@ -0,0 +1,25 @@
+<p>A cookie's domain specifies which websites should be able to read it. Left blank, browsers are supposed to only send the cookie to sites that
```
christophe-zurn-sonarsource: Not directly related to implementation, but I think it's a shame that `HttpCookie` is mentioned nowhere in RSPEC description, not even as an example

java-checks/src/main/java/org/sonar/java/checks/CookieDomainCheck.java
benzonico: nice catch !

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S3331_java.html
benzonico: I'll update the example in RSPEC.

PR #2421, opened 2019-02-11
SONARJAVA-2192 S2699 Fix for indirect assertions in a chain of method calls

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/AssertionsInTestsCheck.java
```diff
@@ -93,18 +95,21 @@
...
     this.context = context;
     assertionInMethod.clear();
+    visitedMethods.clear();
     inUnitTest.push(false);
     methodContainsAssertion.push(false);
     scan(context.getTree());
     methodContainsAssertion.pop();
     inUnitTest.pop();
     assertionInMethod.clear();
+    visitedMethods.clear();
```
benzonico: As discussed, you can probably rely on the keySet of assertionInMethod map rather than using an other set.

PR #2432, opened 2019-02-13
SONARJAVA-2286 Exclude arrays and refine inner class constants

benzonico:


java-checks/src/main/java/org/sonar/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -73,7 +72,8 @@ private boolean isObjectInInnerClass(VariableTree variableTree) {
       if (variableTree.type().is(Tree.Kind.IDENTIFIER)) {
         return !"String".equals(((IdentifierTree) variableTree.type()).name());
       } else {
-        return !variableTree.type().is(Tree.Kind.PRIMITIVE_TYPE);
+        ExpressionTree initializer = variableTree.initializer();
+        return !(variableTree.type().is(Tree.Kind.PRIMITIVE_TYPE) && initializer != null && ConstantUtils.resolveAsConstant(initializer) != null);
```
christophe-zurn-sonarsource: checking `initializer != null` is not necessary here since this method is only called when `hasConstantInitializer` is true, which already ensure the initializer is not null

PR #2437, opened 2019-02-15
SONARJAVA-2965 Detect useless static imports

benzonico:


java-checks/src/main/java/org/sonar/java/checks/UselessImportCheck.java
```diff
@@ -104,13 +106,33 @@ private void reportIssue(ImportTree importTree) {
...
+  private void checkSymbolUsage(ImportTree importTree, String importName) {
+    IdentifierTree id = null;
+    if (importTree.qualifiedIdentifier().is(Tree.Kind.IDENTIFIER)) {
+      id = (IdentifierTree) importTree.qualifiedIdentifier();
+    } else if (importTree.qualifiedIdentifier().is(Tree.Kind.MEMBER_SELECT)) {
+      id = ((MemberSelectExpressionTree) importTree.qualifiedIdentifier()).identifier();
+    }
+    if (id != null) {
+      SemanticModel semanticModel = (SemanticModel) context.getSemanticModel();
+      Symbol symbol = semanticModel.getSymbol(importTree);
```
christophe-zurn-sonarsource: we should ensure `semanticModel`and `symbol` are not null before using it, or alternatively only run the rule when semantic is available and ensure `verifyNoIssueWithoutSemantic`

PR #2450, opened 2019-03-13
SONARJAVA-3048 Parse Java 12 new switch expression

alban-auzeill:


java-frontend/src/main/java/org/sonar/java/model/statement/CaseLabelTreeImpl.java
```diff
@@ -34,20 +35,25 @@
...
   private final InternalSyntaxToken caseOrDefaultKeyword;
-  @Nullable
-  private final ExpressionTree expression;
-  private final InternalSyntaxToken colonToken;
+  private final List<ExpressionTree> expressions;
+  private final Kind kind;
+  private final InternalSyntaxToken colonOrArrowToken;
+
+  public CaseLabelTreeImpl(InternalSyntaxToken caseOrDefaultKeyword, @Nullable ExpressionTree expression, InternalSyntaxToken colonOrArrowToken) {
+    this(caseOrDefaultKeyword, Collections.singletonList(expression), Kind.CASE_LABEL_COLON, colonOrArrowToken);
```
m-g-sonar: I don't like the fact of having two distinct Kinds associated with the same Tree. Moreover, it introduces a breaking change in API, which I REALLY don't like.

What I propose is to introduce a new method, for instance `isFallThrough()` which would be equivalent in term of functionality, and the proper javadoc detailing the difference between "colon" and "arrow" cases. The CASE_LABEL then stays the same, and we only enrich the API with some new methods, and a deprecated one.

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/BreakStatementTree.java
```diff
@@ -43,6 +43,12 @@
   @Nullable
   IdentifierTree label();
 
+  /**
+   * @since 5.12
```
m-g-sonar: I would also add a reference to the Java 12 new feature in the javadoc, to properly explain what it is.

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/CaseLabelTree.java
```diff
@@ -40,9 +40,27 @@
 
   SyntaxToken caseOrDefaultKeyword();
 
+  /**
+   * @deprecated use the {@link #expressions()} method.
+   */
+  @Deprecated
```
m-g-sonar: Add `@since 5.12` tag for deprecated methods, in order to know from when we introduce the change.

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/SwitchStatementTree.java
```diff
@@ -37,7 +37,7 @@
  * @since Java 1.3
  */
 @Beta
-public interface SwitchStatementTree extends StatementTree {
+public interface SwitchStatementTree extends StatementTree, ExpressionTree {
```
m-g-sonar: :+1:

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/Tree.java
```diff
@@ -144,9 +144,17 @@
     CASE_GROUP(CaseGroupTree.class),
 
     /**
+     * case with colon: case <expression-list> :
      * {@link CaseLabelTree}
      */
-    CASE_LABEL(CaseLabelTree.class),
+    CASE_LABEL_COLON(CaseLabelTree.class),
```
m-g-sonar: Like mentioned earlier, I would keep only one CASE_LABEL kind and rely on a boolean.

java-frontend/src/test/java/org/sonar/java/ast/parser/grammar/statements/SwitchStatementTest.java
```diff
@@ -165,27 +155,23 @@ public void variable_declaration_in_a_case() {
...
   }
 
-  // TODO to support java 12
-  @Ignore
   @Test
   public void switch_should_be_a_statement_with_or_without_a_semicolon() {
     assertThat(JavaLexer.STATEMENT)
-      .matches("switch (i) { case 1 -> print('A'); }")
-      .matches("switch (i) { case 1 -> print('A'); };")
+      .matches("{ switch (i) { case 1 -> print('A'); } }")
```
m-g-sonar: this is NOT testing that it's a statement anymore, now for the next two lines we endup testing that the block is a statement. This is not was the test was doing.

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/CaseLabelTree.java
alban-auzeill: Ok, I will add: `/** @deprecated (since 5.12) use ...`
To not confuse with `@since 5.12` that indicate the creation date of the method.

PR #2455, opened 2019-03-19
SONARJAVA-3060 IndexOutOfBoundsException in S2275 when the string contains wrong reference to parameter

alban-auzeill:


java-checks/src/test/files/checks/PrintfFailCheck.java
```diff
@@ -17,6 +17,7 @@ void foo(Calendar c){
     String.format("Display %3$d and then %d", 1, 2, 3);
     String.format("Too many arguments %d and %d", 1, 2, 3);
     String.format("Not enough arguments %d and %d", 1);  // Noncompliant {{Not enough arguments.}}
+    String.format("%1$d %2$d %9$-3.3s", 1, 2, "hello");  // Noncompliant {{Argument at index 9 is not provided.}}
```
m-g-sonar: I would prefer to have a message more explicit. Why not something such as:
```
"Not enough arguments to feed formater at index 9: '%9$-3.3s'."
```

alban-auzeill: Unfortunately, in the above case `rawParam` does not contain `%9$-3.3s` but a cleaned version `%9$-s`, so I decided to only show the `%9$` part.

PR #2462, opened 2019-03-25
SONARJAVA-3054 Remove calls to SQ APIs which were deprecated in version 6.7 or before

alban-auzeill:


m-g-sonar:
LGTM! Nice work!

PR #772, opened 2016-04-06
SONARJAVA-1628 Uses LUB when infering vararg types

m-g-sonar:

PR #2466, opened 2019-03-26
SONARJAVA-3061 CFG for switch statement without falltrough

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -764,22 +785,27 @@ private void buildSwitchStatement(SwitchStatementTree switchStatementTree) {
     currentBlock = conditionBlock;
   }
 
-  private void buildSwitchExpression(SwitchExpressionTree tree) {
-    // FIXME When used as expression, switches are considered as a simple element
-    currentBlock.elements.add(tree);
+  /**
+   * A switch expression can use the traditional cases with 'column' (with fall-through) or,
```
saberduck: typo column -> colon

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -725,27 +725,48 @@ private void buildLabeledStatement(LabeledStatementTree labeledStatement) {
...
     breakTargets.addLast(switchSuccessor);
     boolean hasDefaultCase = false;
-    if (!switchStatementTree.cases().isEmpty()) {
-      CaseGroupTree firstCase = switchStatementTree.cases().get(0);
-      for (CaseGroupTree caseGroupTree : Lists.reverse(switchStatementTree.cases())) {
+    if (!switchExpressionTree.cases().isEmpty()) {
+      boolean withoutFallTrough = switchWithoutFallThrough(switchExpressionTree);
+      CaseGroupTree firstCase = switchExpressionTree.cases().get(0);
+      for (CaseGroupTree caseGroupTree : Lists.reverse(switchExpressionTree.cases())) {
+        if (withoutFallTrough) {
```
saberduck: I think we could move this `if` to the place where the successor is being set (line #772)

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -764,22 +785,27 @@ private void buildSwitchStatement(SwitchStatementTree switchStatementTree) {
...
   }
 
   private static boolean containsDefaultCase(List<CaseLabelTree> labels) {
-    for (CaseLabelTree caseLabel : labels) {
-      if ("default".equals(caseLabel.caseOrDefaultKeyword().text())) {
-        return true;
-      }
-    }
-    return false;
+    return labels.stream().anyMatch(caseLabel -> "default".equals(caseLabel.caseOrDefaultKeyword().text()));
```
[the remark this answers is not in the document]
m-g-sonar: I agree, it could be nice in the API, but I don't feel it's related to the CFG directly. I would prefer to do that in a second time.

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
m-g-sonar: Nop, it does not work, depending of what is happening when building the body of the case, it result in different orders.

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -764,22 +785,27 @@ private void buildSwitchStatement(SwitchStatementTree switchStatementTree) {
...
+   * starting with java 12, the 'arrow' cases (without fall-through). Cases can not be mixed.
+   *
+   * @param switchExpressionTree the switch to evaluate
+   * @return true if the switch uses fall-through
+   */
+  private static boolean switchWithoutFallThrough(SwitchExpressionTree switchExpressionTree) {
+    return switchExpressionTree.cases().stream()
+      .map(CaseGroupTree::labels)
+      .flatMap(List::stream)
+      .noneMatch(CaseLabelTree::isFallThrough);
```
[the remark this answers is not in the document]
m-g-sonar: I'm not sure to follow you. The thing is that `break` statements are going to be naturally connected, by construction, to adequate next block with the current implementation.

The new switch, however, is necessary without fall-through, while there is no `break`, and consequently we should force the connection to the end of the switch, and not the next case.

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
saberduck: ok, the code is bit messy, but it would require bigger restructuring to clean-up.

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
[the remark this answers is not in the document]
saberduck: problem is that `CaseLabelTree.isFallThrough` will return `true` for old style case with colon even if it is not falling through, because body contains `break`.  I think it would be better to use `isArrow` as a name for the function, as there is no ambiguity.

PR #2475, opened 2019-04-01
SONARJAVA-3051 Implement rule S5194

saberduck:


java-checks/src/main/java/org/sonar/java/checks/UseSwitchExpressionCheck.java
```diff
@@ -0,0 +1,127 @@
...
+      if (hasSingleStatement(caseGroupTree.body(), THROW_STATEMENT)) {
+        continue;
+      }
+      Symbol variable = assigningVariable(caseGroupTree.body());
+      if (variable == null) {
+        return null;
+      }
+      if (switchAssigningVariable == null) {
+        switchAssigningVariable = variable;
+      } else if (switchAssigningVariable != variable) {
```
m-g-sonar: what about unknown symbols? It seems to me that if symbols are always unknown, then we will have FPs:

```java
switch(expr) {
  case 1:
    unknown1 = 42;
    break;
  case 2:
    unkown2 = 666;
    break;
}
```

java-checks/src/test/files/checks/UseSwitchExpressionCheck.java
```diff
@@ -0,0 +1,181 @@
+
+class A {
+
+  int field;
+
+  enum DoW {
+    MONDAY, TUESDAY, WEDNESDAY, THURSDAY, FRIDAY, SATURDAY, SUNDAY
+  }
+
+  void foo(DoW day) {
```
m-g-sonar: I would add an example with the new switch as well:

```java
    int numLetters;
    switch (day) {  // Noncompliant {{Use expression switch to set value of 'numLetters'.}}
      case MONDAY, FRIDAY, SUNDAY -> {
        numLetters = 6;
      }
      case TUESDAY -> {
        numLetters = 7;
      }
      case THURSDAY, SATURDAY -> {
        numLetters = 8;
      }
      case WEDNESDAY -> {
        numLetters = 9;
      }
      default -> throw new IllegalStateException("Wat: " + day);
    }
```

java-checks/src/main/java/org/sonar/java/checks/UseSwitchExpressionCheck.java
```diff
@@ -0,0 +1,127 @@
...
+        return null;
+      }
+    }
+    return switchAssigningVariable;
+  }
+
+  private static boolean isSwitchReturningValue(SwitchStatementTree switchStatementTree) {
+    for (CaseGroupTree caseGroupTree : switchStatementTree.cases()) {
```
m-g-sonar: Why not relying on streams?
```java
return switchStatementTree.cases().stream()
  .map(CaseGroupTree::body)
  .allMatch(body -> hasSingleStatement(body, THROW_STATEMENT, RETURN_STATEMENT));
```

java-checks/src/main/java/org/sonar/java/checks/UseSwitchExpressionCheck.java
```diff
@@ -0,0 +1,127 @@
...
+    if (!body.get(1).is(Tree.Kind.BREAK_STATEMENT)) {
+      return null;
+    }
+    ExpressionTree expression = ((ExpressionStatementTree) body.get(0)).expression();
+    if (!expression.is(Tree.Kind.ASSIGNMENT)) {
+      return null;
+    }
+    ExpressionTree variable = ((AssignmentExpressionTree) expression).variable();
+    if (variable.is(Tree.Kind.IDENTIFIER)) {
+      return ((IdentifierTree) variable).symbol();
```
m-g-sonar: before returning, what about checking for unknown symbol? It would prevent FPs.

java-checks/src/test/files/checks/UseSwitchExpressionCheck.java
```diff
@@ -0,0 +1,181 @@
...
+      case SATURDAY:
+        throw new IllegalStateException();
+      default:
+        throw new IllegalStateException();
+    }
+  }
+
+  void set_field(int x) {
+    switch (x) { // Noncompliant {{Use expression switch to set value of 'field'.}}
+      case 1:
```
m-g-sonar: what if the body is a block?
```java
case 1: {
  this.field = 1;
  break;
}
```

java-checks/src/main/java/org/sonar/java/checks/UseSwitchExpressionCheck.java
```diff
@@ -0,0 +1,127 @@
...
+    for (CaseGroupTree caseGroupTree : switchStatementTree.cases()) {
+      if (hasSingleStatement(caseGroupTree.body(), THROW_STATEMENT, RETURN_STATEMENT)) {
+        continue;
+      }
+      return false;
+    }
+    return true;
+  }
+
+  private static Symbol assigningVariable(List<StatementTree> body) {
```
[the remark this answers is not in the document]
saberduck: Added @CheckForNull, however I do not think that using `Optional` would add to clarity here

java-checks/src/main/java/org/sonar/java/checks/UseSwitchExpressionCheck.java
```diff
@@ -0,0 +1,127 @@
...
+    SwitchStatementTree switchStatementTree = (SwitchStatementTree) tree;
+    Symbol switchAssigningVariable = isSwitchAssigningVariable(switchStatementTree);
+    if (switchAssigningVariable != null) {
+      reportIssue(switchStatementTree.switchKeyword(), "Use expression switch to set value of '" + switchAssigningVariable.name() + "'.");
+    } else if (isSwitchReturningValue(switchStatementTree)) {
+      reportIssue(switchStatementTree.switchKeyword(), "Use expression switch to return the value.");
+    }
+  }
+
+  private static Symbol isSwitchAssigningVariable(SwitchStatementTree switchStatementTree) {
```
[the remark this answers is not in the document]
saberduck: Added the annotation and I rewrote the logic using streams to avoid null checks

java-checks/src/test/files/checks/UseSwitchExpressionCheck.java
saberduck: I have not considered the expression switch for the rule before, but good idea! However, this required bigger rework of the rule itself to handle differences in syntax

java-checks/src/main/java/org/sonar/java/checks/UseSwitchExpressionCheck.java
```diff
@@ -0,0 +1,155 @@
...
+import static org.sonar.plugins.java.api.tree.Tree.Kind.RETURN_STATEMENT;
+import static org.sonar.plugins.java.api.tree.Tree.Kind.SWITCH_EXPRESSION;
+import static org.sonar.plugins.java.api.tree.Tree.Kind.THROW_STATEMENT;
+
+@Rule(key = "S5194")
+public class UseSwitchExpressionCheck extends IssuableSubscriptionVisitor implements JavaVersionAwareVisitor {
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(SWITCH_EXPRESSION);
```
m-g-sonar: I'm not sure why you decided to use a `SWITCH_EXPRESSION` here.
Every `SWITCH_STATEMENT` will contain a `SWITCH_EXPRESSION` as first and unique child, and from my point of view the rule only apply to switch statement. A switch expression which does not return always a value would be wrong.

saberduck: fixed

PR #2483, opened 2019-04-04
SONARJAVA-3072 Revert change on 'JavaFileScannerContext#getFileKey()' behavior

m-g-sonar:


java-checks/src/test/java/org/sonar/java/filters/PostAnalysisIssueFilterTest.java
```diff
@@ -64,7 +64,7 @@ public void setUp() {
 
     context = mock(JavaFileScannerContext.class);
     when(context.getFile()).thenReturn(inputFile.file());
-    when(context.getFileKey()).thenReturn(inputFile.key());
+    when(context.getFileKey()).thenReturn(inputFile.file().getAbsolutePath());
```
saberduck: I think instead of this change, we can simplify code in `org.sonar.java.filters.PostAnalysisIssueFilter#scanFile` to drop the usage of file and rely solely on `context.getInputFile` 
That way there will be no usage of `getFileKey` in the plugin

PR #2496, opened 2019-04-14
SONARJAVA-3087 Update S3655 to consider Optional#isEmpty

antmordel:
Implementation of RSPEC-3655 should be updated to handle Optional#isEmpty method introduced in Java 11.

### DIFF
--- a/java-frontend/src/test/files/se/OptionalGetBeforeIsPresentCheck.java
+++ b/java-frontend/src/test/files/se/OptionalGetBeforeIsPresentCheck.java
@@ -56,6 +56,10 @@ abstract class A {
     return s.isPresent() ? null : s.get(); // Noncompliant
   }
 
+  String isEmpty(Optional<String> s) {
+    return s.isEmpty() ? s.get() : null; // Noncompliant
+  }
+
   private void usingEmpty() {
     Optional<String> op = Optional.empty();
     op.get(); // Noncompliant
@@ -80,6 +84,30 @@ abstract class A {
     if (o != null) {
       op.get(); // Compliant - if o is not null, then the optional is necessarily present
     }
+  }
+
+  private void usingIsEmpty() {
+    Optional<Object> op = Optional.emtpy();
+    if (!op.isEmpty()) {
+      op.get(); // Compliant - dead code
+    }
+  }
+
+  private void usingIsEmpty2() {
+    Optional<Object> op = Optional.emtpy();
+    if (op.isEmpty()) {
+      // Noop
+    } else {
+      op.get(); // Compliant - dead code
+    }
+  }
+
+  void usingIsEmpty3() {
+    Optional<String> s = optional;
+    if (!s.isEmpty()) {
+      s.get(); // Compliant
+    }
+    s.get(); // Noncompliant
   }
 
   private void usingFilter1(Optional<String> op) {
--- a/java-frontend/src/test/files/se/OptionalGetBeforeIsPresentCheck.java
+++ b/java-frontend/src/test/files/se/OptionalGetBeforeIsPresentCheck.java
@@ -87,14 +87,14 @@ abstract class A {
   }
 
   private void usingIsEmpty() {
-    Optional<Object> op = Optional.emtpy();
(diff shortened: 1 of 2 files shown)


java-frontend/src/test/files/se/OptionalGetBeforeIsPresentCheck.java
```diff
@@ -82,6 +86,30 @@ private void usingOfNullableWithTest(@Nullable Object o) {
     }
   }
 
+  private void usingIsEmpty() {
+    Optional<Object> op = Optional.emtpy();
```
saberduck: typo in the name of the invoked method, should be `Optional.empty()`

java-frontend/src/test/files/se/OptionalGetBeforeIsPresentCheck.java
```diff
@@ -82,6 +86,30 @@ private void usingOfNullableWithTest(@Nullable Object o) {
...
+    if (!op.isEmpty()) {
+      op.get(); // Compliant - dead code
+    }
+  }
+
+  private void usingIsEmpty2() {
+    Optional<Object> op = Optional.emtpy();
```
saberduck: same as above, should be `Optional.empty`

antmordel: Thank you. Just fixed.

java-frontend/src/test/files/se/OptionalGetBeforeIsPresentCheck.java
antmordel: Thank you. Just fixed.

m-g-sonar:
Hey @antmordel,

I had a look at your PR, and it looks OK to me. Nice work, thanks for your contribution!

It duplicates a bit the part regarding Symbolic Values owned by the check... but we are going to live with it for the time being. The new one being dual to the 1st one, I have the feeling it could be simplified, but I'm not convinced the ROI rewriting it worth the effort.

The only thing which bothers me is the fact that if we build the project with jdk 8, the unit test will fail (we won't be able to resolve the isEmpty method in bytecode, introduced only in jdk 11). So I'll extract it to another file. See PR #2498.

Once the PR#2498 will be green according to our QA and pass the quality gate (a few conditions missing in coverage), I'll merge it.

Cheers,
Michael

antmordel:
> Hey @antmordel,
> 
> I had a look at your PR, and it looks OK to me. Nice work, thanks for your contribution!
> 
> It duplicates a bit the part regarding Symbolic Values owned by the check... but we are going to live with it for the time being. The new one being dual to the 1st one, I have the feeling it could be simplified, but I'm not convinced the ROI rewriting it worth the effort.
> 
> The only thing which bothers me is the fact that if we build the project with jdk 8, the unit test will fail (we won't be able to resolve the isEmpty method in bytecode, introduced only in jdk 11). So I'll extract it to another file. See PR #2498.
> 
> Once the PR#2498 will be green according to our QA and pass the quality gate (a few conditions missing in coverage), I'll merge it.
> 
> Cheers,
> Michael

When I was writing this code for `isEmpty()` I didn't like the fact that it was pretty much the same as it is in `isPresent` but negated for the symbolic engine. A nice refactor can be done.

On another note, sorry for not realizing about the fact that the unit tests are run using `jdk8`. In case all the QA goes OK, do you have an ETA for 5.13 to be released?

Cheers.

m-g-sonar:
I'm gonna see what I can do in the other PR to try to reduce the duplication.

No clear ETA yet for 5.13. We just released 5.12, and I do expect another sprint to start soon to bring some more value (most probably a batch of new rules targeting hotspots). But for the time being there is not enough to do a release. Probably early may.

m-g-sonar:
I merged the other PR, including your (slightly reworked) changes. The fix will be part of 5.13.
I'm consequently closing this PR.

Cheers,
Michael

antmordel:
Thank you so much @m-g-sonar. It was really nice to contribute to this. Thank you for your help.

Cheers,
Antonio

PR #2497, opened 2019-04-15
Move some utility methods from MethodTreeImpl into dedicated class

Godin:


m-g-sonar:
Merged duplicate which went through QA.

PR #2515, opened 2019-04-24
Remove dead code

Godin:


m-g-sonar:
I don't know why, but it seems that Quality Gates failed, but the corresponding Github check didn't failed. There is two issues: https://next.sonarqube.com/sonarqube/project/issues?pullRequest=2515&id=org.sonarsource.java%3Ajava&resolved=false

m-g-sonar:
Burgr is however correctly showing the QGATE block as being red.

PR #2520, opened 2019-05-01
Update REF_MESSAGE of VolatileNonPrimitiveFieldCheck

pivovarit:
The original message is a bit too drastic and can confuse some developers and lead them straight into introducing concurrency issues:

> "Remove the \"volatile\" keyword from this field."

Replaced it with:

> "Remember that declaring a reference \"volatile\" does not guarantee safe publication of the members of the referenced object"


m-g-sonar:
Hello @pivovarit ,

Thanks for your contribution. 

We like our issue messages to be actionable for the end users. The new message you are suggesting is however not that much actionable, nor clearer (too long and complex).

I would be more in favor of a better description within the rule, with explicit cases on which it would be perfectly valid to use the `volatile` keyword, rather than radically modifying the message. Ideally, helping identifying the case where it's legit would help, and prevent us raising issues... Or we could modify the message with something like:

> "Remove the "volatile" keyword from this field, or make sure that  xxxxx."

In order to explicitly say that it may be important to look for something else, but we need the `xxxxx` part to be clear and concise.

PR #2522, opened 2019-05-02
Fix syntax errors in test files

m-g-sonar:
According to Java Language Specification expression statement can
be only assignment, method invocation, class instance creation and
pre/post increment/decrement.


java-checks/src/test/files/checks/StringLiteralInsideEqualsCheck.java
```diff
@@ -2,14 +2,13 @@
   void foo() {
     "foo".equals("bar");        // Noncompliant {{Move the "bar" string literal on the left side of this string comparison.}}
     foo.equals("qux");          // Noncompliant [[sc=16;ec=21]] {{Move the "qux" string literal on the left side of this string comparison.}}
-    "foo".equals(bar);          
-    "foo".foo("bar");           
-    "foo".equals();             
-    foo.equals("bar".length()); 
-    "foo".equals[0];
```
m-g-sonar: dropping only this line... but also removing trailing spaces on other lines. sorry for the noise.

Godin: no pb

PR #2533, opened 2019-05-15
SONARJAVA-3093 Rule S2255: update implementation to NOT include Cookie reading

andrea-guarino-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/CookieShouldNotContainSensitiveDataCheck.java
```diff
@@ -59,21 +58,13 @@
...
   private static final String SET_VALUE_METHOD = "setValue";
-  private static final String GET_VALUE_METHOD = "getValue";
   private static final String WITH_VALUE_METHOD = "withValue";
   private static final String BUILDER_METHOD = "builder";
   private static final String JAVA_LANG_STRING = "java.lang.String";
 
   @Override
   protected List<MethodMatcher> getMethodInvocationMatchers() {
     return Arrays.asList(
-      // setters
```
saberduck: so much for useful comments, right? :)

PR #2534, opened 2019-05-15
SONARJAVA-3104 Rule S5344: Authentication should not rely on insecure…

saberduck:
… "PasswordEncoder"


java-checks/src/main/java/org/sonar/java/checks/security/PasswordEncoderCheck.java
```diff
@@ -0,0 +1,102 @@
...
+  );
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Arrays.asList(Tree.Kind.METHOD, Tree.Kind.NEW_CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree.is(Tree.Kind.NEW_CLASS) && UNSAFE_PASSWORD_ENCODERS.anyMatch(((NewClassTree) tree))) {
```
andrea-guarino-sonarsource: I think we should check if we have semantic information
```
    if (!hasSemantic()) {
      return;
    }
```

java-checks/src/test/java/org/sonar/java/checks/security/PasswordEncoderCheckTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.junit.Test;
+import org.sonar.java.checks.security.PasswordEncoderCheck;
+import org.sonar.java.checks.verifier.JavaCheckVerifier;
+
+public class PasswordEncoderCheckTest {
+
+  @Test
+  public void test() {
+    JavaCheckVerifier.verify("src/test/files/checks/security/PasswordEncoder.java", new PasswordEncoderCheck());
```
andrea-guarino-sonarsource: please, add a call to `JavaCheckVerifier.verifyNoIssueWithoutSemantic`

java-checks/src/main/java/org/sonar/java/checks/security/PasswordEncoderCheck.java
saberduck: fixed

java-checks/src/test/java/org/sonar/java/checks/security/PasswordEncoderCheckTest.java
saberduck: fixed

PR #2536, opened 2019-05-15
Integrate custom rules tutorial

m-g-sonar:
Integrating as MD file the tutorial written as introduction to custom rules, previously available in SonarSource's documentation.


docs/CUSTOM_RULES.md
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * @deprecated since 4.11
+ */
+@Deprecated
+public static final String SUREFIRE_REPORTS_PATH_PROPERTY = "sonar.junit.reportsPath";
+```
+
+### NEWLY DEPRECATED
+
+* `org.sonar.plugins.java.api.JavaFileScannerContext`: Deprecate usage of File-based methods from API, which will be removed in future release. Starting from this version, methods relying on InputFile has to be preferred.
```
saberduck: "has to be preferred" => "are preferred" or alternatively "should be used"

docs/CUSTOM_RULES.md
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * @deprecated (since 5.12) use the {@link #colonOrArrowToken()} method.
+ */
+@Deprecated
+SyntaxToken colonToken();
+```
+
+### NEW API
+
+* `org.sonar.plugins.java.api.JavaFileScannerContext`: Following methods have been added in order to provide help reporting issues at project level, and access data through SonarQube's `InputFile` API, which won't be possible anymore through files:
```
saberduck: drop "which won't be possible...", the sentence is too long already

docs/CUSTOM_RULES.md
```diff
@@ -0,0 +1,228 @@
...
+InputFile getInputFile();
+ 
+/**
+ * InputComponent representing the project being analyzed
+ * @return the project component
+ */
+InputComponent getProject();
+```
+
+* `org.sonar.plugins.java.api.tree.SwitchExpressionTree`: In order to cover the Java 12 new switch expression, introduce a new Tree in the SonarJava Syntax Tree API  (Corresponding `Tree.Kind`: `SWITCH_EXPRESSION` ). New methods have also been added to fluently integrate the new switch expression into the SonarJava API.
```
saberduck: "new switch expression",  new subclass of Tree was introduced

docs/CUSTOM_RULES_101.md
```diff
@@ -0,0 +1,529 @@
...
+
+When creating the rule class, we chose to implement the `IssuableSubscriptionVisitor` class from the API. This class, on top of providing a bunch of useful methods to raise issues, also **defines the strategy which will be used when analyzing a file**. As its name is telling us, it is based on a subscription mechanism, allowing to specify on what kind of tree the rule should react. The list of node types to cover is specified through the `nodesToVisit()` method. In the previous steps, we modified the implementation of the method to return an empty list, therefore not subscribing to any node of the syntax tree.
+
+Now its finally time to jump in to the implementation of our first rule! Go back to the `MyFirstCustomCheck` class, and modify the list of Kinds returned by the nodesToVisit() method. Since our rule targets method declarations, we only need to visit methods. To do so, simply add `Kind.METHOD` as a parameter of the returned immutable list, as shown in the following code snippet.
+
+```java
+@Override
+public List<Kind> nodesToVisit() {
+  return ImmutableList.of(Kind.METHOD);
```
saberduck: I think we should avoid using Guava in the tutorial, let's use `Collections.singletonList` here

docs/CUSTOM_RULES_101.md
```diff
@@ -0,0 +1,529 @@
...
+  // ...
+}
+```
+
+### Rule Activation
+The second things to to is to activate the rule within the plugin. To do so, open class `RulesList` (`org.sonar.samples.java.RulesList`). In this class, you will notice methods `getJavaChecks()` and `getJavaTestChecks()`. These methods are used to register our rules with alongside the rule of the Java plugin. Note that rules registered in `getJavaChecks()` will only be played against source files, while rules registered in `getJavaTestChecks()` will only be played against test files. To register the rule, simply add the rule class to the list builder, as in the following code snippet:
+
+```java
+public static List<Class<? extends JavaCheck>> getJavaChecks() {
+  return ImmutableList.<Class<? extends JavaCheck>>builder()
```
saberduck: same, let's not use Guava APIs

m-g-sonar: But for this I also need to modify the custom plugin template on the other repository.

PR #2541, opened 2019-05-17
SONARJAVA-3076 Rule S5320: Broadcasting intents is security-sensitive

saberduck:


saberduck:
Do not merge, I created dogfooding branch

PR #2551, opened 2019-05-23
SONARJAVA-3089 Handle lombok.val for S2175 and S2159

m-g-sonar:


java-checks/src/test/files/filters/LombokFilter.java
```diff
@@ -244,6 +244,31 @@ private UtilityClassWithPublicConstructorCheck() {
   }
 }
 
+class LombokVal {
```
saberduck: would be nice to add test which is using `import lombok.val` and just `val` as a type

m-g-sonar: I truly don't know how the front-end will react on this... knowing that val is supported with java 11. I fear to open the Pandora Box. But I'm going to have a look.

PR #2555, opened 2019-05-23
Add paragraph about allowed api

m-g-sonar:


docs/CUSTOM_RULES_101.md
```diff
@@ -346,11 +347,21 @@ Test passed? If not, then check if you somehow missed a step.
...
+>
 > :tada: **Congratulations!** :confetti_ball:
 >
 > [*You implemented your first custom rule for the SonarQube Java Analyzer!*](resources/success.jpg)
-> ***
+>
+
+### What you can use, and what you can't
+
+As a rule of thumb, you should keep in mind when writing custom rules to **always**, and **only**, use classes from the [`org.sonar.plugins.java.api`](https://github.com/SonarSource/sonar-java/tree/5.12.1.17771/java-frontend/src/main/java/org/sonar/plugins/java/api) packages from the SonarJava plugin.
```
saberduck: Remove "As a rule of thumb, you should keep in mind", [rule of thumb](https://en.wikipedia.org/wiki/Rule_of_thumb) means that it is not a precise rule, only some guidance, which is not the case here, as this is a very strict requirement. 
Suggestion: "When writing custom rules your plugin can only use classes from package `org.sonar.plugin.api` "

docs/CUSTOM_RULES_101.md
```diff
@@ -346,11 +347,21 @@ Test passed? If not, then check if you somehow missed a step.
...
+
+These classes define the public API which will be available at runtime for your custom rules. Exploring it, you will notice that you already have a few classes which will help you writing enhanced and powerfull rules (for instance triggered only on a specific version of the Java Language, or by relying on the corresponding `CFG` of a method). **Using any other class from the SonarJava Plugin is strictly forbidden**.
+
+When browsing the existing 500+ rules from the SonarJava plugin, you will sometime notice use of some utility classes (`ExpressionUtils`, `PackageUtils`, `LiteralUtils`, `MethodMatcher`, implementation classes of Trees, etc...). While these classes could be sometime extremely useful in your context, **these classes are not available at runtime** for custom rule plugins. It means that, while your unit tests are still going to pass when building the plugin, your rules will most likely make analysis **crash at anaylsis time**.
```
saberduck: typo: "anaylsis" -> "analysis"

docs/CUSTOM_RULES_101.md
```diff
@@ -346,11 +347,21 @@ Test passed? If not, then check if you somehow missed a step.
...
+
+Now, we are always open to discussion, so don't hesitate to reach us and participate to threads, through our [community forum](https://community.sonarsource.com/), to suggest features and API improvement!
```
saberduck: I find the above three paragraphs to be bit redundant, they could be summarized in a single sentence - "Use only classes in the api package, other classes are not available at runtime". I understand that this is a matter of style, but personally, I prefer technical documentation to be succinct and to the point.

docs/CUSTOM_RULES_101.md
saberduck: I am not sure it is a good idea to give examples of useful classes we are not exposing. people could get angry :)

docs/CUSTOM_RULES_101.md
m-g-sonar: I don't know. I appreciate both. In the meantime, we are asking people to first look at implementation of other rules. So I really prefer to warn that they can not use blindly everything.

PR #2556, opened 2019-05-23
SONARJAVA-3096 Rule S2068, S2092, S2115, S3330: fix FN on variable re…

andrea-guarino-sonarsource:
…solution with single assignment

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/AccessibilityChangeCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/AccessibilityChangeCheck.java
@@ -22,7 +22,7 @@ package org.sonar.java.checks;
 import java.util.Arrays;
 import java.util.List;
 import org.sonar.check.Rule;
-import org.sonar.java.checks.helpers.ConstantUtils;
+import org.sonar.java.checks.helpers.ExpressionsHelper;
 import org.sonar.java.checks.methods.AbstractMethodDetection;
 import org.sonar.java.matcher.MethodMatcher;
 import org.sonar.java.matcher.TypeCriteria;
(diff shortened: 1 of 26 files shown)


java-checks/src/main/java/org/sonar/java/checks/helpers/IdentifierUtils.java
```diff
@@ -0,0 +1,97 @@
...
+
+import static org.sonar.java.checks.helpers.ReassignmentFinder.getInitializerOrExpression;
+import static org.sonar.java.checks.helpers.ReassignmentFinder.getReassignments;
+
+public class IdentifierUtils {
+  private IdentifierUtils() {
+    // This class only contains static methods
+  }
+
+  public static <T> ValueResolution<T> getUniqueAssignmentValueResolution(ExpressionTree expression, Function<ExpressionTree,T> resolver) {
```
saberduck: This function would work not only for identifiers, I think it should go into  `ExpressionUtils`, also this name is very long, perhaps something like `getConstantValue` would be sufficient

java-checks/src/main/java/org/sonar/java/checks/helpers/IdentifierUtils.java
```diff
@@ -0,0 +1,97 @@
...
+    return valueResolution(expression, resolver, new ValueResolution<>());
+  }
+
+  private static <T> ValueResolution<T> valueResolution(ExpressionTree expression, Function<ExpressionTree,T> resolver, ValueResolution<T> valueResolution) {
```
saberduck: I think we could use `ConstantUtils.resolveAsConstant` as universal resolver and remove the need for `resolver` parameter. Users can cast the return value on the call site if needed.

saberduck:
Usage search for `Constant.resolveAsString` shows more rules which could benefit from this mechanism like S3331, S5122, S4432, S4925, S4719 . I think especially recent rules in `checks.security` package should be updated to use this.

andrea-guarino-sonarsource:
As discussed, I will create a separate ticket to update other rules using `Constant.resolveAsString`

PR #2564, opened 2019-06-24
SONARJAVA-3118 Use Java 11 and Maven 3.6 in ITs

saberduck:


its/plugin/projects/pom.xml
```diff
@@ -0,0 +1,28 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.sonarsource.it.projects</groupId>
```
m-g-sonar: Being on the open source side of SonarSource, I think we should rather use something such as `org.sonarsource.java.it.projects` here, rather than an `com.sonarsource`

m-g-sonar:
One small comment about the group ID of the parent pom used for our sample projects in ITs. Otherwise, LGTM!

PR #2565, opened 2019-06-24
SONARJAVA-3118 Update Apache Beanutils ruling to prepare migration of ITs to Java 11

saberduck:
Note that this PR is not changing the version of JDK to use for ruling. Changes here are only due to change of the version of Apache BeanUtils we analyze


saberduck:
Replaced by #2564

PR #2568, opened 2019-06-25
Fix flaky test: Change expectations for analysis error size

saberduck:


saberduck:
Because we serialize stacktrace of the exceptions, the number of exceptions which are serialized can vary depending on the size of the stacktrace. This can depend on the platfrom and JDK version used.

PR #2573, opened 2019-07-15
SONARJAVA-3158 Remove usage of deprecated APIs in ITs

saberduck:
See https://jira.sonarsource.com/browse/SONAR-8776

### DIFF
--- a/its/plugin/tests/pom.xml
+++ b/its/plugin/tests/pom.xml
@@ -49,7 +49,11 @@
     <dependency>
       <groupId>org.sonarsource.sonarqube</groupId>
       <artifactId>sonar-ws</artifactId>
-      <version>6.7</version>
+      <version>7.9</version>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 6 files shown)


its/plugin/tests/pom.xml
```diff
@@ -49,7 +49,11 @@
     <dependency>
       <groupId>org.sonarsource.sonarqube</groupId>
       <artifactId>sonar-ws</artifactId>
-      <version>6.7</version>
+      <version>7.9</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
```
pynicolas: Why that new dependency?

saberduck: Version 6.7 of sonar-ws was transitively providing this dependency and we depend on it. Version 7.9 doesnt export this dependency anymore, so we have to add it

pynicolas: Ok.

PR #2574, opened 2019-07-16
SONARJAVA-3128 Rule S5261: "else" statements should be clearly matched with an "if"

yassin-kammoun-sonarsource:


java-checks/src/test/files/checks/DanglingElseStatementsCheck.java
```diff
@@ -0,0 +1,34 @@
...
+        e++;
+      }
+
+    if (a)
+      if (b)
+        d++;
+    else        // Noncompliant {{Add explicit curly braces to avoid dangling else.}}
+      e++;
+  }
+}
```
saberduck: You should configure your editor to insert empty line at the end of file, see here https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

PR #2575, opened 2019-07-16
Update version of sonar-plugin-api to 7.9

saberduck:


saberduck:
Build is failing, because we would need to migrate and release analyzers-commons to use 7.9 api

saberduck:
As we don't want to migrate analyzers-commons now, and there is no strong incentive yet, I am closing this PR

PR #2576, opened 2019-07-17
Add EditorConfig file

saberduck:
EditorConfig is standard for editors to synchronize some basic editing settings, so it is easier to use different IDEs to collaborate on the project. See http://EditorConfig.org  for details.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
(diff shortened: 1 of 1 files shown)


m-g-sonar:
do we reaaaaaally need this in our repository @saberduck ?
I feel it should be rather here: https://github.com/SonarSource/sonar-developer-toolset

saberduck:
1. we have it already on quite some repositories already at SonarSource https://github.com/search?q=org%3ASonarSource+.editorconfig&type=Code 

2. having it in the repository itself is the point exactly, you don't need to setup something externally

m-g-sonar:
LGTM!

PR #2577, opened 2019-07-18
SONARJAVA-3133 Rule S1067 should be ignored in equals methods

quentin-jaquier-sonarsource:


java-checks/src/test/files/checks/ExpressionComplexityCheck.java
```diff
@@ -58,4 +58,37 @@ Runnable lambda() {
...
+      @Override
+      public boolean equals(Object o) {
+        return true && true && true && false && false; // Compliant, the expression is inside equals method
+      }
+    }
+    return true && true && true && false && false; // Compliant, the expression is inside equals method
+  }
+}
+
+enum AbbreviationOfDays{
```
m-g-sonar: haha, nice pick! :)
Could have been an interface as well!

PR #2578, opened 2019-07-18
SONARJAVA-2971 Rule S3012: Arrays should not be copied using loops

yassin-kammoun-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ArrayCopyLoopCheck.java
```diff
@@ -0,0 +1,309 @@
...
+      ForEachStatement loop = (ForEachStatement) tree;
+      return getBody(loop.statement(), 1);
+    } else {
+      WhileStatementTree loop = (WhileStatementTree) tree;
+      return getBody(loop.statement(), 2);
+    }
+  }
+
+  private static StatementTree getBody(StatementTree tree, int expectedSize) {
+    StatementTree statement = tree;
```
saberduck: Do we need this variable?

java-checks/src/main/java/org/sonar/java/checks/ArrayCopyLoopCheck.java
```diff
@@ -0,0 +1,309 @@
...
+
+  private static boolean isCollection(ExpressionTree tree) {
+    return tree.symbolType().isSubtypeOf("java.util.Collection");
+  }
+
+  private static boolean isIdentifier(ExpressionTree tree) {
+    return tree.is(Kind.IDENTIFIER);
+  }
+
+  private static IdentifierTree getIdentifier(ExpressionTree tree) {
```
saberduck: When  you use this method, you subsequently always need `Symbol` from the identifier, it would be more usefull to return the `Symbol` directly. Also annotate this method with `@CheckForNull`

java-checks/src/main/java/org/sonar/java/checks/ArrayCopyLoopCheck.java
```diff
@@ -0,0 +1,309 @@
...
+      ExpressionTree lhs = assignment.variable();
+      ExpressionTree rhs = assignment.expression();
+      if (isIdentifier(lhs) && isInteger(lhs)) {
+        IdentifierTree identifier = (IdentifierTree) lhs;
+        ctx.setCounter(identifier.symbol());
+        if (rhs.is(Kind.PLUS)) {
+          BinaryExpressionTree addition = (BinaryExpressionTree) rhs;
+          lhs = addition.leftOperand();
+          rhs = addition.rightOperand();
+          return (isCounter(lhs, ctx) && isOne(rhs)) ^ (isOne(lhs) && isCounter(rhs, ctx));
```
saberduck: Please use `||` , its more idiomatic for boolean types

java-checks/src/main/java/org/sonar/java/checks/ArrayCopyLoopCheck.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+    return null;
+  }
+
+  private static boolean isCounter(ExpressionTree tree, Context ctx) {
+    IdentifierTree identifier = getIdentifier(tree);
+    return identifier != null && identifier.symbol().equals(ctx.getCounter());
+  }
+
+  private static boolean isInteger(ExpressionTree tree) {
```
saberduck: I would suggest renaming this to `isIntegerType`, to make it less ambiguous (from the name I was thinking that you are testing the expression to be integer literal)

java-checks/src/main/java/org/sonar/java/checks/ArrayCopyLoopCheck.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+    return false;
+  }
+
+  private static boolean checkCondition(ExpressionTree tree, Context ctx) {
+    if (tree.is(Kind.LESS_THAN, Kind.LESS_THAN_OR_EQUAL_TO, Kind.GREATER_THAN, Kind.GREATER_THAN_OR_EQUAL_TO, Kind.NOT_EQUAL_TO)) {
+      BinaryExpressionTree comparison = (BinaryExpressionTree) tree;
+      ExpressionTree lhs = comparison.leftOperand();
+      ExpressionTree rhs = comparison.rightOperand();
+      return isInteger(lhs) && isInteger(rhs) && (isIdentifier(lhs) && isCounter(lhs, ctx)) ^ (isIdentifier(rhs) && isCounter(rhs, ctx));
```
saberduck: calling `isInteger` or `isIdentifier` with `isCounter` is redundant - if it's the same symbol, it means it has the same type and that it is identifier.

java-checks/src/main/java/org/sonar/java/checks/ArrayCopyLoopCheck.java
```diff
@@ -0,0 +1,309 @@
...
+
+  private static boolean isArrayToArrayCopy(ExpressionTree expression, Context ctx) {
+    if (expression.is(Kind.ASSIGNMENT)) {
+      AssignmentExpressionTree assignment = (AssignmentExpressionTree) expression;
+      ExpressionTree lhs = assignment.variable();
+      ExpressionTree rhs = assignment.expression();
+      if (lhs.is(Kind.ARRAY_ACCESS_EXPRESSION) && rhs.is(Kind.ARRAY_ACCESS_EXPRESSION)) {
+        ArrayAccessExpressionTree src = (ArrayAccessExpressionTree) rhs;
+        ArrayAccessExpressionTree dst = (ArrayAccessExpressionTree) lhs;
+        return isIdentifier(src.expression()) && useCounter(src, ctx) && isIdentifier(dst.expression()) && useCounter(dst, ctx);
```
saberduck: can't we just use `isCounter` for both `src.expression` and `dst.expression` ?

java-checks/src/main/java/org/sonar/java/checks/ArrayCopyLoopCheck.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+    return false;
+  }
+
+  private static boolean checkCondition(ExpressionTree tree, Symbol counter) {
+    if (tree.is(Kind.LESS_THAN, Kind.LESS_THAN_OR_EQUAL_TO, Kind.GREATER_THAN, Kind.GREATER_THAN_OR_EQUAL_TO, Kind.NOT_EQUAL_TO)) {
+      BinaryExpressionTree comparison = (BinaryExpressionTree) tree;
+      ExpressionTree lhs = comparison.leftOperand();
+      ExpressionTree rhs = comparison.rightOperand();
+      return isCounter(lhs, counter) ^ isCounter(rhs, counter);
```
saberduck: Use `||` instead of xor

java-checks/src/main/java/org/sonar/java/checks/ArrayCopyLoopCheck.java
```diff
@@ -0,0 +1,301 @@
...
+    return null;
+  }
+
+  @CheckForNull
+  private static Symbol isPlusAssignment(ExpressionStatementTree update) {
+    if (update.expression().is(Kind.PLUS_ASSIGNMENT)) {
+      AssignmentExpressionTree assignment = (AssignmentExpressionTree) update.expression();
+      ExpressionTree lhs = assignment.variable();
+      ExpressionTree rhs = assignment.expression();
+      IdentifierTree identifier = (IdentifierTree) lhs;
```
saberduck: This cast is not safe, `lhs` can be `MemberSelectExpressionTree` when you use field like
```
    for (this.x = 0; this.x < 10; this.x += 1) {

    }
```

java-checks/src/main/java/org/sonar/java/checks/ArrayCopyLoopCheck.java
```diff
@@ -0,0 +1,301 @@
...
+            Symbol identifier = getIdentifier(argument);
+            return identifier != null && identifier.equals(iterated.symbol());
+          }
+        }
+      }
+    }
+    return false;
+  }
+
+  private static boolean isArray(ExpressionTree tree) {
```
saberduck: this function is used only once, we can inline it

PR #2580, opened 2019-07-18
SONARJAVA-3151 Rule S4684 should not raise issue when entity argument is annotated with @PathVariable

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/spring/PersistentEntityUsedAsRequestParameterCheck.java
```diff
@@ -73,4 +76,8 @@ private static boolean isRequestMappingAnnotated(Symbol.MethodSymbol methodSymbo
   private static boolean isPersistentEntity(VariableTree variableTree) {
     return ENTITY_ANNOTATIONS.stream().anyMatch(variableTree.type().symbolType().symbol().metadata()::isAnnotatedWith);
   }
+
+  private static boolean hasNoPathVariableAnnotation(VariableTree variableTree) {
+    return variableTree.modifiers().annotations().stream().noneMatch(a -> a.symbolType().is(PATH_VARIABLE_ANNOTATION));
```
saberduck: We have another API which is used in the methods above - `variableTree.symbol.metadata().isAnnotatedWith` , I think it would better to use that one for consistency

PR #2582, opened 2019-07-19
SONARJAVA-3149 S1948 FP on serializable collection initialized to null

alban-auzeill:


its/ruling/src/test/resources/jdk6/squid-S1948.json
```diff
@@ -123,7 +123,6 @@
 81,
 ],
 'jdk6:java/security/SecureRandom.java':[
-78,
```
alban-auzeill: :heavy_check_mark: `private Provider provider = null;` it was a FP `Provider` implements `Serializable` and `null` is also serializable.

PR #2583, opened 2019-07-19
SONARJAVA-3150 S2276 FP when Tread.sleep call is nested in a lambda

quentin-jaquier-sonarsource:
Rule S2276, S2444, S2696, S2273 are impacted by the change.

### DIFF
--- a/java-checks/src/test/files/checks/StaticFieldInitializationCheck.java
+++ b/java-checks/src/test/files/checks/StaticFieldInitializationCheck.java
@@ -149,14 +149,14 @@ class A {
 
   }
 
-  class E {
-    static Config CONFIG = null;
+  static class E {
+    static Config CONFIG;
 
-    synchronized Object foo() {
+    synchronized void foo() {
       CONFIG = null; // Compliant
     }
 
-    synchronized Object foo() {
+    synchronized Consumer<String> bar() {
       return s -> {
         CONFIG = null; // Noncompliant
       };
--- a/java-checks/src/test/files/checks/StaticFieldInitializationCheck.java
+++ b/java-checks/src/test/files/checks/StaticFieldInitializationCheck.java
@@ -148,4 +148,18 @@ class A {
     }
 
   }
+
+  class E {
+    static Config CONFIG = null;
+
+    synchronized Object foo() {
+      CONFIG = null; // Compliant
+    }
+
+    synchronized Object foo() {
+      return s -> {
+        CONFIG = null; // Noncompliant
+      };
+    }
+  }
 }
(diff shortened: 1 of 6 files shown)


java-checks/src/test/files/checks/StaticFieldInitializationCheck.java
```diff
@@ -148,4 +148,18 @@ public static void resetConfiguration2() {
     }
 
   }
+
+  class E {
```
m-g-sonar: Try to keep the code you introduce as close as possible from something which compiles. It helps and is usually closer from real use cases. In particular, I modified your code as follow:

:warning: WARNING :warning: : I didn't checked if this code is still getting a 100% in term of code coverage.

```
static class E {
    static Config CONFIG;

    synchronized void foo() {
      CONFIG = null; // Compliant
    }

    synchronized Consumer<String> bar() {
      return s -> {
        CONFIG = null; // Noncompliant
      };
    }
  }
```

java-checks/src/test/files/checks/StaticFieldUpateCheck.java
```diff
@@ -71,6 +71,22 @@ public int compareTo(Object o) {
     };
   }
 
+  private synchronized Comparable<Object> foo() {
+    return new Comparable<Object>() {
+      @Override
+      public int compareTo(Object o) {
+        value = 0; // FN: "value" may not be assigned inside a static method
```
m-g-sonar: To be verified that it was NOT covered before. Otherwise it's a regression.

java-checks/src/test/files/checks/ThreadSleepCheck.java
```diff
@@ -17,3 +21,54 @@ synchronized void foo() {
     Thread.sleep(200, 12); // Noncompliant
   }
 }
+
+public class Testcase {
+  private static final Logger LOGGER = Logger.getLogger("com.reftel.magnus.sqsynchtest");
```
m-g-sonar: logger is useless

java-checks/src/test/files/checks/ThreadSleepCheck.java
```diff
@@ -17,3 +21,54 @@ synchronized void foo() {
...
+  private static synchronized Comparable<Object> foo() {
+    return new Comparable<Object>() {
+      @Override
+      public int compareTo(Object o) {
+        Thread.sleep(200);    // Compliant, this won't be invoked when the lock is held
+      }
+    };
+  }
+
+  public synchronized void print() {
```
m-g-sonar: I don't see the interest of this method.

java-checks/src/test/files/checks/StaticFieldUpateCheck.java
quentin-jaquier-sonarsource: I verified: this one (and the next one added in this file) were already FN before the change.

PR #2587, opened 2019-07-20
SONARJAVA-3162 Rule S3067: NPE synchronize on getClass in initializer

hajush:
Fix for issue at https://jira.sonarsource.com/browse/SONARJAVA-3162

Relevant details are in the ticket. Includes a unit test (through adding non-compliant code to the SynchronizationOnGetClassCheck.java)

This is an exercise to ensure my proposed fix can be documented and considered, as it is currently being used in a rather large project, or the NullPointerException effectively disabled the sonar-java scans. I hope this helps!


java-checks/src/main/java/org/sonar/java/checks/synchronization/SynchronizationOnGetClassCheck.java
```diff
@@ -65,10 +67,15 @@ public void visitNode(Tree tree) {
   private static boolean isEnclosingClassFinal(ExpressionTree expressionTree) {
     if (expressionTree.is(IDENTIFIER)) {
       Tree parent = expressionTree.parent();
-      while (!parent.is(METHOD, CONSTRUCTOR)) {
+      while (!parent.is(METHOD, CONSTRUCTOR, INITIALIZER)) {
```
quentin-jaquier-sonarsource: This parent dereference is the root of the NPE, the loop never stopped and eventually reached the top of the tree.
Are you sure that it will not happen again, that you covered all the cases? What about `STATIC_INITIALIZER`?
What about synchronized statement inside a lambda:
```
  Consumer<String> c = s -> {
      synchronized (getClass()) {  }
  };
```

It is tricky to get all of them, therefore you may want to check parent for null here!

quentin-jaquier-sonarsource:
Hello @hajush,

We plan to be ready for the next release this Wednesday, it would be great if this fix is part of it!

I understand that the schedule is tight, please let me know if you plan to work on it soon, otherwise I will handle the fixes to merge it before the release.

hajush:
Hi! Very sorry, I’ve been on a vacation with spouse & family since your
pull request note. I will be back at my home office tomorrow evening. I’d
love to get this in as well, and I’ll work on it if that’s soon enough. But
I won’t be offended if you do the fix yourself. My PR was mostly to
document what I did for my client.

Harold

hajush:
I overestimated what I could do that weekend, as was off to conferences since then. I’m working on it again now.

java-checks/src/main/java/org/sonar/java/checks/synchronization/SynchronizationOnGetClassCheck.java
hajush: The Java language grammar does not allow anything outside of a class, so we should be able to determine the enclosing class. The method isEnclosingClassFinal should not be able to climb up the ExpressionTree without coming to an enclosing class, or there's a bug in the code or in the parser.

I'm uncomfortable "bullet proofing" the code as it will be potentially (and silently) incorrect. Perhaps there is a way to issue an intelligent error so a bug can be filed. I'm not familiar enough with the philosophy and standards here.

I've added a test case for the STATIC_INITIALIZER and for LAMBDA_EXPRESSION. But I'm not finding a way to determine the enclosing class of the LAMBDA_EXPRESS like we can with MethodTree. Any ideas?

java-checks/src/main/java/org/sonar/java/checks/synchronization/SynchronizationOnGetClassCheck.java
```diff
@@ -65,10 +67,15 @@ public void visitNode(Tree tree) {
...
         parent = parent.parent();
       }
-      return ((MethodTree) parent).symbol().owner().isFinal();
+      if (parent instanceof MethodTree) {
+        return ((MethodTree) parent).symbol().owner().isFinal();
+      } else {
+        // hypothesis: the parent is type class for initializers
```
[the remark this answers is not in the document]
hajush: Agreed. Though to be fair, the issue was a specific and relevant one that no longer NPE's. So the submitted code isn't just a trade. In fact, with the STATIC_INITIALIZER and LAMBDA_EXPRESSION, it still falls off and NPE's. At this point, I think I've proven to myself that adding INITIALIZER and STATIC_INITIALIZER to the while loop check means we are getting METHOD or initializers that we can turn into classtree, so it's no longer a hypothesis.

java-checks/src/main/java/org/sonar/java/checks/synchronization/SynchronizationOnGetClassCheck.java
hajush: On additional study, the STATIC_INITIALIZER, INITIALIZER, and LAMBDA_EXPRESSION cases are best covered simply by climbing up the tree until finding the enclosing class, which valid java syntax will require.

java-checks/src/main/java/org/sonar/java/checks/synchronization/SynchronizationOnGetClassCheck.java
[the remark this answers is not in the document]
hajush: FYI - I believe every case is handled by covering METHOD and CONSTRUCTOR as before, and all other cases by climbing up the tree until the enclosing CLASS is encountered. I've also rebased two times so it should be up to date.

hajush:
@quentin-jaquier-sonarsource as far as I can tell I've addressed all your concerns. Will you let me know if I missed anything?

m-g-sonar:
Hey @hajush , we are currently in the middle of a huge refactor of the SonarJava Frontend. I'll handle your PR once we are done with it, as we can't take the time to look at it at the moment.

And thanks for your contribution!

m-g-sonar:
Hello @hajush ,

Sorry for the delay handling your PR. I duplicated your contribution in another PR (to trigger our Quality Check and QA), and amended it to also cover the case of synchronize block within enums. 

Once everything is green, I will merge it and close this one.
Your contribution will be part of SonarJava 6.0, the next version to be released.

Cheers,
Michael

PR #2590, opened 2019-07-23
SONARJAVA-3124 S3077: ignore some standard immutable types

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/VolatileNonPrimitiveFieldCheck.java
```diff
@@ -36,6 +36,44 @@
...
+    "java.time.OffsetDateTime",
+    "java.time.OffsetTime",
+    "java.time.Year",
+    "java.time.YearMonth",
+    "java.time.ZoneId",
+    "java.time.ZoneOffset",
+    "java.time.ZonedDateTime",
+    "java.time.Duration",
+    "java.time.Period",
+    "java.util.Date",
```
saberduck: `java.util.Date` is not immutable, see https://docs.oracle.com/javase/7/docs/api/java/util/Date.html#setDate(int)

java-checks/src/main/java/org/sonar/java/checks/VolatileNonPrimitiveFieldCheck.java
```diff
@@ -36,6 +36,44 @@
   private static final String REF_MESSAGE = "Remove the \"volatile\" keyword from this field.";
   private static final String ARRAY_MESSAGE = "Use an \"Atomic%sArray\" instead.";
 
+  private static final List<String> STANDARD_IMMUTABLE_TYPES = Arrays.asList(
```
saberduck: can you add `java.net.Inet4Address`, `java.net.Inet6Address`

java-checks/src/test/files/checks/VolatileNonPrimitiveFieldCheck.java
```diff
@@ -1,3 +1,41 @@
+import java.util.concurrent.atomic.AtomicIntegerArray;
+import java.awt.Color;
+import java.awt.Cursor;
+import java.awt.Font;
+import java.io.File;
+import java.lang.Boolean;
```
saberduck: no need to import `java.lang` , they are in scope by default

java-checks/src/main/java/org/sonar/java/checks/VolatileNonPrimitiveFieldCheck.java
quentin-jaquier-sonarsource: Indeed, nice catch, I have updated the ticket aswell!

PR #2592, opened 2019-07-24
SONARJAVA-3148 S1130: take implicit constructor call into account

quentin-jaquier-sonarsource:


java-checks/src/test/files/checks/RedundantThrowsDeclaration.java
```diff
@@ -287,3 +287,47 @@ public void missingJavadocForException() throws MyException, MyException2, java.
...
+public class Child extends Parent {
+
+  public Child(Integer a) throws IllegalAccessException { // Compliant, implicit call can throw IllegalAccessException
+    // implicit call to parent constructor
+    System.out.println("a:" + a);
+  }
+
+  public Child(Long a) throws IllegalAccessException { // Compliant
+    super();
+    // no implicit call to parent constructor
```
m-g-sonar: being completely explicit, I think you can remove the comments here and in the following methods

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -236,10 +272,21 @@ private static boolean emptyBody(MethodTree methodTree) {
...
     public void visitMethodInvocation(MethodInvocationTree tree) {
+      if (isSuperInvocation(tree)) {
+        visitedSuper = true;
+      }
       addThrownTypes(tree.symbol());
       super.visitMethodInvocation(tree);
     }
 
+    private static boolean isSuperInvocation(MethodInvocationTree tree) {
+      if (tree.methodSelect().is(Tree.Kind.IDENTIFIER)) {
```
m-g-sonar: this is not enough. It won't cover cases of nested classes, called for instance with something such as `A.this.super()`.

```java
class A {

  class Parent {
    public Parent() throws IllegalAccessException { // Compliant
      throw new IllegalAccessException();
    }
  }

  class Child extends Parent {

    public Child(Long a) throws IllegalAccessException { // Compliant
      A.this.super();
      // no implicit call to parent constructor
      System.out.println("a:" + a);
    }
  }
}
```

However... You are lucky. It seems that we are not resolving properly this particular call. And it is resolved as `unknownSymbol`, which makes the rule say nothing.

To me, the proper way should however be to check that the method symbol name is `"<init>"` and that the owner of the super constructor is the parent class of the present constructor.

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -216,14 +219,47 @@ private static boolean emptyBody(MethodTree methodTree) {
     if (block != null) {
       MethodInvocationVisitor visitor = new MethodInvocationVisitor();
       block.accept(visitor);
-      return visitor.thrownExceptions();
+      Set<Type> thrownExceptions = visitor.thrownExceptions();
+
+      if (thrownExceptions != null && methodTree.is(Tree.Kind.CONSTRUCTOR) && !visitor.visitedSuper) {
```
m-g-sonar: I'm not sure we need this. II would prefer to let the logic of collecting exceptions hidden in the visitor. Why not directly handling this case in the `MethodInvocationVisitor` then?  You can perfectly give the symbol of the current methodTree as parameter of the `MethodInvocationVisitor` constructor, so you don't lose the info from where you are calling.

At the end of the visit, you could then simply add thrown exception from the default super constructor, if `super` has not been called and it's really a constructor.

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -216,14 +219,47 @@ private static boolean emptyBody(MethodTree methodTree) {
...
+          return Optional.of(candidates.get(0));
+        }
+      }
+    }
+    return Optional.empty();
+  }
+
+  private static boolean isDefaultConstructor(Symbol symbol) {
+    if (symbol.isMethodSymbol()) {
+      MethodTree methodTree = ((Symbol.MethodSymbol)symbol).declaration();
```
m-g-sonar: when the class is not in the same file, you don't have access to the constructor. Check for a symbol called `"<init>"` without parameters. Don't restrain yourself to what we can do with the syntax only.

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
quentin-jaquier-sonarsource: I'm not so lucky:
```
class A {
  class Parent {
    public Parent() throws IllegalAccessException { // Compliant
      throw new IllegalAccessException();
    }
    public Parent(String a) { // Compliant
    }
  }

  class Child extends Parent {
    public Child(Long a) throws IllegalAccessException { // Noncompliant
      A.this.super("a:" + a);
    }
  }
}
```

This is a FN, since it does not detect that super is already called!
I will use your way

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
quentin-jaquier-sonarsource: Yes, let's put this logic inside the visitor

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
quentin-jaquier-sonarsource: I'm not so lucky:
```
class A {
  class Parent {
    public Parent() throws IllegalAccessException { // Compliant
      throw new IllegalAccessException();
    }
    public Parent(String a) { // Compliant
    }
  }

  class Child extends Parent {
    public Child(Long a) throws IllegalAccessException { // Noncompliant
      A.this.super("a:" + a);
    }
  }
}
```

This is a FN, since it does not detect that super is already called!
I will use your way

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -224,18 +226,31 @@ private static boolean emptyBody(MethodTree methodTree) {
...
+      if (methodTree.is(Tree.Kind.CONSTRUCTOR) && !visitedOtherConstructor) {
+        Optional<Symbol.MethodSymbol> constructor = getImplicitlyCalledConstructor(methodTree);
+        constructor.ifPresent(c -> thrownExceptions.addAll(c.thrownTypes()));
+      }
       return thrownExceptions;
     }
 
     @Override
     public void visitMethodInvocation(MethodInvocationTree tree) {
+      if ("<init>".equals(tree.symbol().name())) {
```
quentin-jaquier-sonarsource: In my opinion, this is enough, if any `<init>` is called inside the body (this or super), the implicit constructor will not be present!

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -224,18 +225,31 @@ private static boolean emptyBody(MethodTree methodTree) {
...
+    }
 
     @Nullable
     public Set<Type> thrownExceptions() {
       if (visitedUnknown || thrownExceptions.stream().anyMatch(Type::isUnknown)) {
         // as soon as there is an unknown type, we discard any attempt to find an issue
         return null;
       }
+      if (methodTree.is(Tree.Kind.CONSTRUCTOR) && !visitedOtherConstructor) {
+        Optional<Symbol.MethodSymbol> constructor = getImplicitlyCalledConstructor(methodTree);
```
m-g-sonar: I think it would be even cleaner if you do not use the variable:
```java
getImplicitlyCalledConstructor(methodTree)
  .map(Symbol.MethodSymbol::thrownTypes)
  .ifPresent(thrownExceptions::addAll);
```

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -272,6 +286,31 @@ public void visitClass(ClassTree tree) {
...
+      Type superType = ((Symbol.TypeSymbol)methodTree.symbol().owner()).superClass();
+      return superType.symbol().memberSymbols().stream()
+        .filter(MethodInvocationVisitor::isDefaultConstructor)
+        .map(Symbol.MethodSymbol.class::cast)
+        .findFirst();
+    }
+
+    private static boolean isDefaultConstructor(Symbol symbol) {
+      if (symbol.isMethodSymbol()) {
+        Symbol.MethodSymbol methodSymbol = ((Symbol.MethodSymbol)symbol);
```
m-g-sonar: useless parenthesis

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
```diff
@@ -272,6 +286,31 @@ public void visitClass(ClassTree tree) {
...
+        if ("<init>".equals(methodSymbol.name())) {
```
m-g-sonar: extract `"<init>"` into a constant, such as `CONSTRUCTOR_NAME`, to make it explicit and help future reviewer/reader to get the meaning.

java-checks/src/main/java/org/sonar/java/checks/RedundantThrowsDeclarationCheck.java
quentin-jaquier-sonarsource: Yes, nice proposition!

PR #2594, opened 2019-07-24
SONARJAVA-2972 Rule S3039: "String" calls should not go beyond their …

yassin-kammoun-sonarsource:


java-checks/src/test/files/checks/StringCallsBeyondBoundsCheck.java
```diff
@@ -0,0 +1,73 @@
+class A {
+  void f() {
+    String str = "Hello, world!";
+    char[] array = new char[str.length()];
+    int zero = 0, minusOne = -1, any = System.currentTimeMillis();
+
```
saberduck: I think following pattern should be detected 
```
    s.charAt(s.length());
```
even if we don't statically know the lenght of `s` . All your tests assume that we statically know the length of the string (which is fine), but we should also cover the cases where `length()` is a parameter to the call on the same symbol

java-checks/src/main/java/org/sonar/java/checks/StringCallsBeyondBoundsCheck.java
```diff
@@ -0,0 +1,154 @@
...
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree invocation) {
+    boolean issue = false;
+    String method = invocation.symbol().name();
+    switch (method) {
+      case "charAt":
+      case "codePointAt":
+        // we cannot use `!lt` instead of `gte` because it would evaluate to true on an empty optional
+        issue = check(invocation, (str, args) -> lt(val(args.get(0)), opt(0)) || gte(val(args.get(0)), len(str)));
```
saberduck: these conditions are a bit hard to reason about, I believe it's because they work on too low abstraction level. It would be more readable if you defined functions on higher abstraction. You can have a function like `isValidIndex(ExpressionTree indexParam, int lowerBound, int upperBound)`,  then for example `isValidStringIndex(ExpressionTree str, ExpressionTree index)` which calls the previous with `0, len(str)`.... Once you have such functions you will be able to simplify many of the expressions, and more importantly, it will be easier to reason about conditions.

java-checks/src/main/java/org/sonar/java/checks/StringCallsBeyondBoundsCheck.java
```diff
@@ -0,0 +1,154 @@
...
+import org.sonar.plugins.java.api.tree.Tree.Kind;
+
+@Rule(key = "S3039")
+public class StringCallsBeyondBoundsCheck extends AbstractMethodDetection {
+
+  private static final String STRING = "java.lang.String";
+  private static final MethodMatcher STRING_LENGTH =
+    MethodMatcher.create().typeDefinition(STRING).name("length").withoutParameter();
+
+  private static interface Condition {
```
saberduck: I think you can just use `java.util.function.BiFunction`

java-checks/src/main/java/org/sonar/java/checks/helpers/ConstantUtils.java
```diff
@@ -84,7 +84,7 @@ public static Object resolveAsConstant(ExpressionTree tree) {
     if (expression.is(Tree.Kind.STRING_LITERAL)) {
       return LiteralUtils.trimQuotes(((LiteralTree) expression).value());
     }
-    if (tree.is(Tree.Kind.INT_LITERAL)) {
+    if (tree.is(Tree.Kind.INT_LITERAL, Tree.Kind.UNARY_MINUS, Tree.Kind.UNARY_PLUS)) {
```
saberduck: This will only work for Integer literals and not for Long literals. You can either make it work for Long literals, or you can add negative test, to document that it doesn't work for Long literals

PR #2595, opened 2019-07-24
Fix QA tests not being executed

saberduck:


m-g-sonar:
LGTM! Thanks for the fix!

PR #2596, opened 2019-07-25
SONARJAVA-3122 Support more @VisibleForTesting annotations

alban-auzeill:
Let's support all `VisibleForTesting` annotation whatever their package names

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/ProtectedMemberInFinalClassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ProtectedMemberInFinalClassCheck.java
@@ -37,7 +37,6 @@ import org.sonar.plugins.java.api.tree.VariableTree;
 @Rule(key = "S2156")
 public class ProtectedMemberInFinalClassCheck extends IssuableSubscriptionVisitor {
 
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/ProtectedMemberInFinalClassCheck.java
```diff
@@ -84,7 +83,8 @@ private void checkComplianceOnModifiersAndSymbol(ModifiersTree modifiers, Symbol
   }
 
   private static boolean isVisibleForTesting(Symbol symbol) {
-    return symbol.metadata().isAnnotatedWith(GUAVA_FQCN);
+    return symbol.metadata().annotations().stream()
+      .anyMatch(annotation -> "VisibleForTesting".equals(annotation.symbol().type().name()));
```
m-g-sonar: I'm a bit afraid here of cases where the annotation is called "VisibleForTesting" but not resolved by our semantic engine. In the following example, their is no change we can resolve the type, and still, it's a VisibleForTesting annotation, so I would expect no issue.
```
@org.foo.bar.VisibleForTesting
final int var2;
```

PR #2601, opened 2019-07-25
SONARJAVA-3113 Handle big integers in binary format

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/LiteralUtils.java
```diff
@@ -47,11 +47,15 @@ public static Integer intLiteralValue(ExpressionTree expression) {
...
   private static Integer intLiteralValue(LiteralTree literal) {
     String literalValue = literal.value().replaceAll("_", "");
     if (literalValue.startsWith("0b") || literalValue.startsWith("0B")) {
-      return Integer.valueOf(literalValue.substring(2), 2);
+      String binaryWithoutPrefix = literalValue.substring(2);
+      if (binaryWithoutPrefix.length() >= 32 && binaryWithoutPrefix.charAt(0) == '1') {
+        // assume it is used as bit mask
+        return Integer.parseUnsignedInt(binaryWithoutPrefix, 2);
+      }
+      return Integer.valueOf(binaryWithoutPrefix, 2);
```
alban-auzeill: It does not work with this valid java integer literal `0b0000_1111_1111_1111_1111_0000_0000_0000_0000`
I would use `Integer.parseUnsignedInt` for all `literalValue` stating with `0b`, the test `length() >= 32 && charAt(0) == '1'` seems useless.

PR #2602, opened 2019-07-25
SONARJAVA-3154 Ignore issues on imports for lombok

m-g-sonar:


java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -93,56 +99,44 @@
...
+    excludeLinesIfTrue(generatesPrivateConstructor(tree), tree, UtilityClassWithPublicConstructorCheck.class);
+    excludeLinesIfTrue(usesAnnotation(tree, UTILITY_CLASS), tree, BadFieldNameCheck.class, ConstantsShouldBeStaticFinalCheck.class);
+
+    super.visitClass(tree);
+  }
 
-    if (usesAnnotation(tree, UTILITY_CLASS)) {
-      excludeLines(tree, BadFieldNameCheck.class);
-      excludeLines(tree, ConstantsShouldBeStaticFinalCheck.class);
+  private final void excludeLinesIfTrue(boolean shouldExclude, Tree tree, Class<? extends JavaCheck> rule) {
```
saberduck: `final` doesn't make sense for private method, we should have a rule for that

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -93,56 +99,44 @@
...
-      acceptLines(tree, BadFieldNameCheck.class);
-      acceptLines(tree, ConstantsShouldBeStaticFinalCheck.class);
+      acceptLines(tree, rule);
     }
+  }
 
-    super.visitClass(tree);
+  @SafeVarargs
+  private final void excludeLinesIfTrue(boolean shouldExclude, Tree tree, Class<? extends JavaCheck>... rules) {
+    BiConsumer<Tree, Class<? extends JavaCheck>> consumer = shouldExclude ? this::excludeLines : this::acceptLines;
```
saberduck: I find it a bit strange to declare lambda variable, which is used on the next line. I don't think using Stream API gives a lot in terms of clarity here, I would just use old fashioned
```
for (Class<? extends JavaCheck> rule : rules) {
      if (shouldExclude) {
        excludeLines(tree, rule);
      } else {
        acceptLines(tree, rule);
      }
    }```

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
m-g-sonar: this one is indeed a mistake. the `final` was forced for the other one because of the `@SafeVarargs` annotation

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
m-g-sonar: I wanted to avoid to repeat the boolean test for the potentially N checks from the varargs. But as there is only max 2 of them at the moment, it's maybe overkill.

PR #2605, opened 2019-07-26
SONARJAVA-3131 Rule S5164 Call remove on ThreadLocal

saberduck:


java-checks/src/main/java/org/sonar/java/checks/ThreadLocalCleanupCheck.java
```diff
@@ -0,0 +1,89 @@
...
+      Symbol.TypeSymbol clazz = ((ClassTree) tree).symbol();
+      clazz.memberSymbols().stream()
+        .filter(Symbol::isVariableSymbol)
+        .filter(s -> s.type().is(THREAD_LOCAL))
+        .filter(ThreadLocalCleanupCheck::isNotSuper)
+        .forEach(this::checkThreadLocalField);
+    } else {
+      MethodInvocationTree mit = (MethodInvocationTree) tree;
+      if (THREADLOCAL_SET.matches(mit) && mit.arguments().get(0).is(Tree.Kind.NULL_LITERAL)) {
+        reportIssue(mit, "Use \"remove()\" instead of \"set(null)\".");
```
m-g-sonar: I don't see why, in the description, this would not work. If it's a common mistake, I would add a small text to make it explicit.

java-checks/src/main/java/org/sonar/java/checks/ThreadLocalCleanupCheck.java
```diff
@@ -0,0 +1,89 @@
...
+      }
+    }
+  }
+
+  private static boolean isNotSuper(Symbol s) {
+    return !"super".equals(s.name());
+  }
+
+  private void checkThreadLocalField(Symbol field) {
+    if (field.usages().stream().noneMatch(ThreadLocalCleanupCheck::usageIsRemove)) {
```
m-g-sonar: Is it enough to judge? The issue message specifically mention "in a finally block", what if the method is NOT called from inside a finally block? If it's not related, then there is maybe some information we need to drop.

java-checks/src/main/java/org/sonar/java/checks/ThreadLocalCleanupCheck.java
```diff
@@ -0,0 +1,89 @@
...
+      reportIssue(((VariableTree) field.declaration()).simpleName(), "Call the \"remove()\" on \"" + field.name() + "\" in a \"finally\" block.");
+    }
+  }
+
+  private static boolean usageIsRemove(IdentifierTree usage) {
+    Tree parent = usage.parent();
+    if (parent.is(Tree.Kind.MEMBER_SELECT)) {
+      Tree mseParent = parent.parent();
+      return mseParent != null && mseParent.is(Tree.Kind.METHOD_INVOCATION)
```
m-g-sonar: not sure about the null check here (I know that you are getting it because of the annotation on the `parent()` method), but it will never be possible here. Maybe `Objects.requireNonNull(parent.parent()).is(Tree.Kind.METHOD_INVOCATION)`?

saberduck: tbh, I find null check more readable that the alternative you are suggesting. I prefer to be bit more defensive than to risk NPE if API contract changes

java-checks/src/main/java/org/sonar/java/checks/CheckList.java
```diff
@@ -674,7 +674,8 @@ private CheckList() {
       DanglingElseStatementsCheck.class,
       ArrayCopyLoopCheck.class,
       StringCallsBeyondBoundsCheck.class,
-      SuspiciousListRemoveCheck.class
+      SuspiciousListRemoveCheck.class,
+      ThreadLocalCleanupCheck.class
```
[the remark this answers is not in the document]
saberduck: yeah, we should sort the list alphabetically, I will perhaps do it in another PR

its/ruling/src/test/resources/guava/squid-S5164.json
```diff
@@ -0,0 +1,11 @@
+{
+'com.google.guava:guava:src/com/google/common/cache/Striped64.java':[
```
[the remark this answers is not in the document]
saberduck: indeed, I excluded non-private fields

java-checks/src/main/java/org/sonar/java/checks/ThreadLocalCleanupCheck.java
```diff
@@ -0,0 +1,89 @@
...
+        .forEach(this::checkThreadLocalField);
+    } else {
+      MethodInvocationTree mit = (MethodInvocationTree) tree;
+      if (THREADLOCAL_SET.matches(mit) && mit.arguments().get(0).is(Tree.Kind.NULL_LITERAL)) {
+        reportIssue(mit, "Use \"remove()\" instead of \"set(null)\".");
+      }
+    }
+  }
+
+  private static boolean isNotSuper(Symbol s) {
```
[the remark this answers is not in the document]
saberduck: I removed subclasses of `ThreadLocal`

java-checks/src/main/java/org/sonar/java/checks/ThreadLocalCleanupCheck.java
saberduck: I added some explanation to the RSPEC

java-checks/src/main/java/org/sonar/java/checks/ThreadLocalCleanupCheck.java
saberduck: Removed the "finally" part from the message, it was a recomendation from linked article, but it is not possible in every scenario and detecting when it is possible would be too complicated with little roi

PR #2608, opened 2019-07-26
SONARJAVA-3101 FP on S1168 for implementation of Spring Batch ItemProcessor

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ReturnEmptyArrayNotNullCheck.java
```diff
@@ -170,6 +180,14 @@ private static boolean isAllowingNull(MethodTree methodTree) {
         return true;
       }
     }
-    return false;
+    return requiresReturnNull(methodTree);
+  }
+
+  private static boolean requiresReturnNull(MethodTree methodTree) {
+    return isOverriding(methodTree) && REQUIRES_RETURN_NULL.stream().anyMatch(methodTree.symbol().owner().type()::isSubtypeOf);
```
alban-auzeill: `methodTree` could be a `foo()` method overriding another `foo()` and the owner of this `foo()` method could implements `ItemProcessor`, this does not imply that the `foo()` method override `ItemProcessor#process`.

quentin-jaquier-sonarsource: Nice catch!

PR #2612, opened 2019-07-29
SONARJAVA-3134 Rule S2121: Silly String operations should not be made

yassin-kammoun-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/SillyStringOperationsCheck.java
```diff
@@ -0,0 +1,183 @@
...
+  private static boolean isZero(ExpressionTree tree) {
+    return Optional.ofNullable(ConstantUtils.resolveAsIntConstant(tree)).map(n -> n == 0).orElse(false);
+  }
+
+  private static boolean isStringLength(ExpressionTree str, ExpressionTree tree) {
+    Symbol symbol = symbol(str);
+    if (symbol != null && tree.is(Kind.METHOD_INVOCATION)) {
+      MethodInvocationTree invocation = (MethodInvocationTree) tree;
+      if (STRING_LENGTH.matches(invocation) && invocation.methodSelect().is(Kind.MEMBER_SELECT)) {
+        return symbol.equals(symbol(((MemberSelectExpressionTree) invocation.methodSelect()).expression()));
```
saberduck: You can reuse `isSameSymbol` here and simplify this method a bit

java-checks/src/main/java/org/sonar/java/checks/SillyStringOperationsCheck.java
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+    return false;
+  }
+
+  @CheckForNull
+  private static Symbol symbol(ExpressionTree tree) {
```
saberduck: You should annotate parameter with `@Nullable` , or if you don't need to handle `null` argument, you don't need `ofNullable`

java-checks/src/main/java/org/sonar/java/checks/SillyStringOperationsCheck.java
```diff
@@ -0,0 +1,183 @@
...
+      return isStringLength(str, args.get(0)) || (isZero(args.get(0)) && isStringLength(str, args.get(1)));
+    }
+  }
+
+  private static boolean isSameString(ExpressionTree str, ExpressionTree tree) {
+    return isSameSymbol(str, tree) || isSameStringLiteral(str, tree);
+  }
+
+  private static boolean isSameSymbol(ExpressionTree tree, ExpressionTree other) {
+    return Optional.ofNullable(symbol(tree)).map(s -> s.equals(symbol(other))).orElse(false);
```
saberduck: I think this usage of `Optional` is not adding much in terms of clarity. Using usual control flow constructs is more straightforward in Java

PR #2613, opened 2019-07-29
Fix syntax error in test file

Godin:
According to Java Language Specification 12 expression statement can
be only assignment, method invocation, class instance creation and
pre/post increment/decrement.


m-g-sonar:
LGTM!

PR #2621, opened 2019-07-30
SONARJAVA-3129 Rule S5411: Boxed Boolean should be avoided in boolean…

yassin-kammoun-sonarsource:


its/ruling/src/test/resources/jdk6/squid-S5411.json
```diff
@@ -0,0 +1,12 @@
+{
+'jdk6:java/awt/Component.java':[
+5968,
+],
+'jdk6:java/awt/GraphicsEnvironment.java':[
+114,
```
saberduck: we should avoid having duplicate issue, you can use `BaseTreeVisitor` to cut the visit

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -0,0 +1,119 @@
...
+        if (isBooleanIdentifier(expression.rightOperand())) {
+          return expression.rightOperand();
+        }
+      }
+    }
+    return null;
+  }
+
+  private static boolean isBooleanIdentifier(ExpressionTree tree) {
+    return tree.symbolType().is(BOOLEAN) && tree.is(Kind.IDENTIFIER);
```
saberduck: It's sufficient to test for `java.lang.Boolean` type, that way we can also support usage of functions with `java.lang.Boolean` return value (don't forget to add test)

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  @CheckForNull
+  private static ExpressionTree findBoxedBoolean(ExpressionTree tree) {
+    if (isBooleanIdentifier(tree)) {
+      return tree;
+    }
+    if (tree instanceof BinaryExpressionTree) {
+      BinaryExpressionTree expression = (BinaryExpressionTree) tree;
```
saberduck: In case of binary expression we should just check that either lhs or rhs is `java.lang.Boolean` and `boolean` respectively, no need to check for null checks.

java-checks/src/test/files/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -0,0 +1,132 @@
...
+  boolean boxedEqual1(Boolean B, boolean b) {
+    return B == b; // Noncompliant
+  }
+
+  boolean boxedEqual2(Boolean B, boolean b) {
+    return b == B; // Noncompliant
+  }
+
+  boolean boxedEqual3(Boolean B, Boolean V) {
+    return B == V; // Noncompliant
```
saberduck: why is this Noncompliant?

java-checks/src/test/files/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -0,0 +1,132 @@
...
+  boolean boxedNotEqual1(Boolean B, boolean b) {
+    return B != b; // Noncompliant
+  }
+
+  boolean boxedNotEqual2(Boolean B, boolean b) {
+    return b != B; // Noncompliant
+  }
+
+  boolean boxedNotEqual3(Boolean B, Boolean V) {
+    return B != V; // Noncompliant
```
saberduck: why is this noncompliant?

PR #2628, opened 2019-07-31
SONARJAVA-2970 Rule S3064 Assignment should be the last step with dou…

saberduck:
…ble-checked locking


java-checks/src/main/java/org/sonar/java/checks/DoubleCheckedLockingAssignmentCheck.java
```diff
@@ -0,0 +1,184 @@
...
+    if (expr.is(Tree.Kind.ASSIGNMENT)) {
+      AssignmentExpressionTree assignment = (AssignmentExpressionTree) expr;
+      if (symbol.equals(symbol(assignment.variable()))) {
+        return assignment;
+      }
+    }
+    return null;
+  }
+
+  static class UsageVisitor extends BaseTreeVisitor {
```
m-g-sonar: I think you should skip lambdas and classes or methods. You may have inner classes, or anonymous classes using this symbol.

java-checks/src/test/files/checks/DoubleCheckedLockingAssignment.java
```diff
@@ -0,0 +1,123 @@
...
+          strings = new ArrayList<>();  // Noncompliant [[sc=11;ec=38;secondary=12,13]] {{Fully initialize "strings" before assigning it.}}
+          strings.add("Hello");
+          strings.add("World");
+        }
+      }
+    }
+    return strings;
+  }
+
+  public List<String> getStrings() {
```
m-g-sonar: please use a different name for the method here, this should not compile...

java-checks/src/main/java/org/sonar/java/checks/DoubleCheckedLockingAssignmentCheck.java
```diff
@@ -0,0 +1,184 @@
...
+        IfStatementTree secondIf = (IfStatementTree) statementTree;
+        Symbol secondIfSymbol = nullCheck(secondIf.condition());
+        if (symbol.equals(secondIfSymbol)) {
+          checkUsageAfterAssignment(symbol, secondIf);
+        }
+      }
+    }
+  }
+
+  private void checkUsageAfterAssignment(Symbol symbol, IfStatementTree secondIf) {
```
m-g-sonar: I would provide the then statement immediately here, rather than the second if.

java-checks/src/test/files/checks/DoubleCheckedLockingAssignment.java
```diff
@@ -0,0 +1,123 @@
+package test;
+
+public class DoubleCheckedLockingAssignment {
+
+  private volatile List<String> strings;
```
m-g-sonar: I know we don't really care about the type of the symbol itself, but could you import `java.util.List` and `java.util.ArrayList` to be sure that unresolved types does not play a role?

java-checks/src/main/resources/org/sonar/l10n/java/rules/squid/S3064_java.html
```diff
@@ -0,0 +1,50 @@
...
+      }
+    }
+    return strings;
+  }
+}
+
+</pre>
+<h2>See</h2>
+<ul>
+  <li> <a href="https://www.securecoding.cert.org/confluence/x/IgAZAg">CERT, LCK10-J.</a> - Use a correct form of the double-checked locking idiom
```
m-g-sonar: In the RSPEC there is a link to another rule. This is missing here.

PR #2636, opened 2019-08-06
Fix assertions on expected exception

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/ast/JavaAstScannerTest.java
```diff
@@ -65,6 +66,7 @@
 import org.sonar.sslr.grammar.LexerlessGrammarBuilder;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.fail;
```
bannmann: You're mixing JUnit assertions with assertJ here. For consistency, you should switch to `org.assertj.core.api.Assertions.fail`.

(I'm not the reviewer, but I hope you don't mind this comment.)

m-g-sonar: Haha, So you are going to hate what is coming! :D

I had a discussion yesterday about something similar with @Godin ... And it fact, if you look around in the plugin, there is almost no consistency. Yet.

Gonna fix this one, thx! :)

PR #2638, opened 2019-08-10
Make reports of Junit5 unit tests in @Nested inner classes show up in scan

kwuerl:
This PR fixes the Problem mentioned here: https://community.sonarsource.com/t/number-of-unit-tests-is-not-measured-with-nested-junit-5-tests/4255/3

I added a fix so that testcases in inner class reports are new igored if they where already present in the parent class without completely ignoring test reports from inner classes.

This is neccessary due to the fact that the reports created by the gradle junit5 test runner creates seperate reports for @Nested inner class test cases.

As the name property in UnitTestResult seems to be only ever empty in the unit tests i added a default test name (as UUID) to be able to identify all UnitTestResults by name properly.


m-g-sonar:
Hey @kwuerl ,

First, thanks a lot for the high quality of your contribution, and sorry for the time it took us to get back to you. Indeed, this case has never been properly covered. Your contribution fixes this!

I created the corresponding ticket to document the change and have it part of our release note: [SONARJAVA-3222](https://jira.sonarsource.com/browse/SONARJAVA-3222)

I also rebased your PR on top of our master branch, and open another PR containing your duplicated (and squashed) changes, renamed according to our ticket.

If everything goes fine on QA/Quality Gate side, I'll merge it.
The fix will then be part of the next release of SonarJava: 6.0.

Cheers,
Michael

m-g-sonar:
Merged the squashed commits. Thanks again!

PR #2640, opened 2019-08-13
Fix test file

m-g-sonar:
* 'enum' used as identifier was only possible till java 1.4
* '_' used as identifier was only possible till java 1.8,
  and has never been possible with lambdas

### DIFF
--- a/java-checks/src/test/java/org/sonar/java/checks/naming/KeywordAsIdentifierCheckTest.java
+++ b/java-checks/src/test/java/org/sonar/java/checks/naming/KeywordAsIdentifierCheckTest.java
@@ -26,7 +26,7 @@ public class KeywordAsIdentifierCheckTest {
 
   @Test
   public void test() {
-    JavaCheckVerifier.verify("src/test/files/checks/naming/KeywordAsIdentifierCheck.java", new KeywordAsIdentifierCheck());
+    JavaCheckVerifier.verify("src/test/files/checks/naming/KeywordAsIdentifierCheck.java", new KeywordAsIdentifierCheck(), 4);
   }
 
 }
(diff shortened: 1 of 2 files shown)


java-checks/src/test/files/checks/naming/KeywordAsIdentifierCheck_enum.java
```diff
@@ -0,0 +1,21 @@
+class A {
+  int foo = 0;
+  int enum = 0; // Noncompliant [[sc=7;ec=11]] {{Use a different name than "enum".}}
+
+  enum mynum {RED, GREEN, YELLOW};
```
Godin: Unfortunately this is still not valid Java code of any version:
* not Java 5 or greater, because `enum` is used as identifier at line 3
* not Java 4 or lower, because of attempt to declare `enum` at lines 5 and 19

m-g-sonar: Good point. Sorry for missing this.

java-checks/src/test/java/org/sonar/java/checks/naming/KeywordAsIdentifierCheckTest.java
```diff
@@ -25,8 +25,13 @@
...
   @Test
-  public void test() {
-    JavaCheckVerifier.verify("src/test/files/checks/naming/KeywordAsIdentifierCheck.java", new KeywordAsIdentifierCheck());
+  public void test_enum() {
+    JavaCheckVerifier.verify("src/test/files/checks/naming/KeywordAsIdentifierCheck_enum.java", new KeywordAsIdentifierCheck(), 4);
+  }
+
+  @Test
+  public void test_underscore() {
+    JavaCheckVerifier.verify("src/test/files/checks/naming/KeywordAsIdentifierCheck_underscore.java", new KeywordAsIdentifierCheck(), 8);
```
Godin: This file doesn't use anything that requires Java greater than 4, thus actually no need to split :wink: Sorry that I did not notice this before.

PR #2643, opened 2019-08-13
S1610 Autovalue builder support

coutin:
* Fix S1610 Autovalue support not taking @AutoValue.Builder
annotation into account and throws S1610 where it shouldn't

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/AbstractClassNoFieldShouldBeInterfaceCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/AbstractClassNoFieldShouldBeInterfaceCheck.java
@@ -102,6 +102,6 @@ public class AbstractClassNoFieldShouldBeInterfaceCheck extends IssuableSubscrip
     return tree.modifiers().annotations().stream()
       .map(AnnotationTree::annotationType)
       .map(TypeTree::symbolType)
-      .noneMatch(type -> type.is("com.google.auto.value.AutoValue") || type.is("org.immutables.value.Value$Immutable"));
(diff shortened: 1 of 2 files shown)


zhirsch:
I am also encountering this false positive, so it would be great to have this PR merged!

zhirsch:
> I am also encountering this false positive, so it would be great to have this PR merged!

Also, a similar problem exists for [@AutoOneOf](https://github.com/google/auto/blob/master/value/src/main/java/com/google/auto/value/AutoOneOf.java) where an abstract class is annotated with @AutoOneOf and sonar recommends that an interface is used instead.

Is it possible to extend this PR to suppress squid:S1610 for abstract classes that are annotated with `com.google.auto.value.AutoOneOf`, too?

m-g-sonar:
Hey @coutin , thanks for your contribution and sorry for the delay it took us to work on it.

According to the [Javadoc of `AutoValue.Builder` annotation](https://github.com/google/auto/blob/master/value/src/main/java/com/google/auto/value/AutoValue.java#L49), I somehow disagree that this would be a FP, as having annotated interface is clearly mentioned as being something possible. However, it also mention classes, which explicitly allow this case. In such case, I can understand that someone would consider such issues as noise.

Thanks @zhirsch for also mentionning `@AutoOneOf`. I agree it should also be handled.

I created ticket [SONARJAVA-3216](https://jira.sonarsource.com/browse/SONARJAVA-3216) to document the issue and track it.

I'm now going to include your fix in another PR and add handling of the other annotations. Once our QA and Quality Gates are green, I'll merge it.

Cheers,
Michael

PR #2645, opened 2019-08-14
Rule S1188: Add separate parameter for lambdas

alefhar:
The length of a body of a lambda expression and an anonymous class should be
treated differently: Anonymous classes may contain more than a single method
and should be permitted to have a longer body. Lambdas on the other hand
should be short and concise and thus require a different setting.

Post in Sonar Source Community: [S1188 should differentiate between lambdas and anonymous classes](https://community.sonarsource.com/t/java-s1188-should-differentiate-between-lambdas-and-anonymous-classes/7437)

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/AnonymousClassesTooBigCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/AnonymousClassesTooBigCheck.java
@@ -37,10 +37,15 @@ public class AnonymousClassesTooBigCheck extends BaseTreeVisitor implements Java
 
   private static final int DEFAULT_MAX = 20;
 
-  @RuleProperty(key = "Max",
-    description = "Maximum allowed lines in an anonymous class/lambda",
+  @RuleProperty(key = "Max (Anonymous Class)",
+    description = "Maximum allowed lines in an anonymous class",
     defaultValue = "" + DEFAULT_MAX)
-  public int max = DEFAULT_MAX;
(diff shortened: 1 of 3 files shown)


alefhar:
I changed the key for the parameter from "Max" to "Max (Anonymous Class)" and introduced a new key "Max (Lambda)". Could this lead to problems when parameter was set to something other than the default value? I.e., the rule would suddenly check against 20 lines instead of, say, 10 lines previously configured.

m-g-sonar:
Hey @alefhar,

First, let me apologies for the time it took to get back to you. This should have been addressed a long time ago...

We are usually quite strongly **against** adding new parameters to rules, as we really try to make each rule very simple to understand, and configure. With such approach, we avoid troubles when making rules evolve on the long run (your question precisely reflects this).

Following your idea, we decided to split rule S1188 in two. The original S1188 will continue to target anonymous classes, while a new rule, S5612, will target exclusively lambdas.

The new rule will be implemented at the same time as we will do the change of S1188. Here are the corresponding tickets in our JIRA: [SONARJAVA-3218](https://jira.sonarsource.com/browse/SONARJAVA-3218) & [SONARJAVA-3219](https://jira.sonarsource.com/browse/SONARJAVA-3219)

I'm consequently closing this PR. Feel free to open a new PR introducing the new rule and changing the old one, if you want. Otherwise, we will implement it in the upcoming days, as it is scheduled tp be included in our next release, SonarJava 6.0.

Cheers,
Michael

PR #2657, opened 2019-09-04
Adding unit tests for JSymbol methods

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/JParser.java
```diff
@@ -844,6 +844,8 @@ private EnumConstantTreeImpl processEnumConstantDeclaration(EnumConstantDeclarat
       ),
       separatorToken
     );
+    t.variableBinding = e.resolveVariable();
```
Godin: @m-g-sonar thank you for spotting this 👍however `declaration(t.variableBinding, t)` is missing in this change :wink: I added both into master in 482a3f933fac6dc3c1034763d4476089fb16a69b

m-g-sonar: I have seen the fix on master :+1: .
I'll rebase the branch.

PR #2658, opened 2019-09-06
Use declaration bindings to save and retrieve usages and declarations

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/model/JParserSemanticTest.java
```diff
@@ -349,6 +349,17 @@ void declaration_type() {
...
 
+  @Test
+  void declaration_type_parameterized() {
+    JavaTree.CompilationUnitTreeImpl cu = test("class C<T> { C<String> field; }");
+    ClassTreeImpl c = (ClassTreeImpl) cu.types().get(0);
+    AbstractTypedTree fieldType = (AbstractTypedTree) ((VariableTreeImpl) c.members().get(0)).type();
+
+    assertThat(c.typeBinding).isNotNull();
+    assertThat(cu.sema.declarations.get(c.typeBinding)).isSameAs(c);
+    assertThat(cu.sema.usages.get(c.typeBinding)).containsOnly((IdentifierTreeImpl) ((ParameterizedTypeTree) fieldType).type());
```
Godin: Modification in `JParser#usage` is enough to pass this test, so I'm wondering if other modifications are really needed.

To use modifications in `JSymbol` one need to call `declaration` or `usages` on `cu.sema.typeSymbol(fieldType.typeBinding)`, which corresponds to `fieldType.symbolType().symbol()` in old implementation. And I'm wondering about existence of such calls.

java-frontend/src/main/java/org/sonar/java/model/JParser.java
```diff
@@ -346,14 +346,14 @@ private void declaration(@Nullable IBinding binding, Tree node) {
     if (binding == null) {
       return;
     }
-    sema.declarations.put(binding, node);
+    sema.declarations.put(sema.declarationBinding(binding), node);
```
Godin: IMO for declarations `binding == sema.declarationBinding(binding)` already.

java-frontend/src/test/java/org/sonar/java/model/JParserSemanticTest.java
```diff
@@ -379,6 +390,18 @@ void declaration_method() {
...
+  @Test
+  void declaration_method_parameterized() {
+    JavaTree.CompilationUnitTreeImpl cu = test("class C { <T> void m(T t) { m(\"hello\"); } }");
+    ClassTreeImpl c = (ClassTreeImpl) cu.types().get(0);
+    MethodTreeImpl m = (MethodTreeImpl) c.members().get(0);
+    MethodInvocationTreeImpl m1Invoc = (MethodInvocationTreeImpl) ((ExpressionStatementTree) m.block().body().get(0)).expression();
+
+    assertThat(m.methodBinding).isNotNull();
+    assertThat(cu.sema.declarations.get(m.methodBinding)).isEqualTo(m);
+    assertThat(cu.sema.usages.get(m.methodBinding)).containsOnly((IdentifierTreeImpl) m1Invoc.methodSelect());
```
Godin: Similarly to use modifications in JSymbol one need to call `declaration` or `usages` on `cu.sema.methodSymbol((IMethodBinding) identifier.binding)`, which corresponds to `identifier.symbol()` in old implementation. And I'm wondering about existence of such calls.

java-frontend/src/main/java/org/sonar/java/model/JParser.java
m-g-sonar: MMh indeed. For declaration maybe it does not make sense.

java-frontend/src/test/java/org/sonar/java/model/JParserSemanticTest.java
m-g-sonar: I can see quite a lot of `symbolType().symbol()` in checks (14 matches, 13 checks relying on this).

m-g-sonar: However modifications in `JSymbol` are mandatory, otherwise we simply never get back to the declaration bindings from parameterized symbol. I didn't tested it because calling `symbol().usages()` in `JParserSemanticTest` returns the old symbol.

Godin: Call `cu.sema.typeSymbol(fieldType.typeBinding).usages()` :wink:

PR #2664, opened 2019-09-19
Avoid ClassCastException in RedundantTypeCastCheck

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -92,37 +90,51 @@ private static boolean requiredForMemberAccess(TypeCastTree typeCastTree) {
...
+          if (type instanceof MethodJavaType) {
+            return ((MethodJavaType) type).resultType();
+          }
+          return methodTree.symbol().returnType().type();
+        }
+        return null;
+      case VARIABLE:
+        return ((VariableTree) parent).symbol().type();
+      case METHOD_INVOCATION:
+        // FIXME this can never happen. Dead code.
```
Godin:
```suggestion
        // FIXME(SONARJAVA-3191) dead code
```

PR #2665, opened 2019-09-20
Drop usage of MethodJavaSymbol#checkOverridingParameters() in checks

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/LeastSpecificTypeCheck.java
```diff
@@ -204,12 +201,30 @@ private Type leastSpecificType() {
...
+    private static boolean isOverriding(Symbol s1, Symbol s2) {
+      return s1.isMethodSymbol()
+        && s2.isMethodSymbol()
+        && s1.name().equals(s2.name())
+        && isPotentialOverride((Symbol.MethodSymbol) s1, (Symbol.MethodSymbol) s2);
     }
 
-    private static boolean isOverridingMethod(JavaSymbol.MethodJavaSymbol s1, JavaSymbol.MethodJavaSymbol s2, ClassJavaType superClass) {
-      return s1.name().equals(s2.name()) && BooleanUtils.isTrue(s1.checkOverridingParameters(s2, superClass));
+    private static boolean isPotentialOverride(Symbol.MethodSymbol method, Symbol.MethodSymbol overrideeCandidate) {
```
Godin: This is a pure duplicate of method in another class. If we don't want to  add method to some helper class, then let's at least reuse the same method, since it is static.

java-checks/src/main/java/org/sonar/java/checks/LeastSpecificTypeCheck.java
```diff
@@ -164,21 +161,21 @@ private void computeChainsForSuperType(List<List<Type>> result, Symbol m, Type t
...
-    private static boolean definesOrInheritsSymbol(Symbol symbol, JavaSymbol.TypeJavaSymbol typeSymbol) {
+    private static boolean definesOrInheritsSymbol(Symbol symbol, Symbol.TypeSymbol typeSymbol) {
       return definesSymbol(symbol, typeSymbol)
-        || typeSymbol.superTypes().stream().anyMatch(superType -> definesSymbol(symbol, superType.symbol()));
+        || JUtils.superTypes(typeSymbol).stream().anyMatch(superType -> definesSymbol(symbol, superType.symbol()));
     }
 
     private static boolean definesSymbol(Symbol m, Symbol.TypeSymbol typeSymbol) {
-      return typeSymbol.memberSymbols().stream().anyMatch(s -> isOverriding(m, s, ((ClassJavaType) typeSymbol.type())));
+      return typeSymbol.memberSymbols().stream().anyMatch(s -> isOverriding(m, s));
```
Godin: It was already like this before your changes, but must admit that implementation of this check looks to me either extremely inefficient or extremely redundant: `isOverriding` checks `m.isMethodSymbol` and `m` doesn't depend from elements of stream.

PR #2668, opened 2019-09-23
Fix CollectionInappropriateCallsCheck

m-g-sonar:


m-g-sonar:
Changing approach after master fix.

PR #2682, opened 2019-11-11
SONARJAVA-3219 Remove Lambda specificity for S1188

guillaume-dequenne-sonarsource:


java-checks/src/test/files/checks/LambdaTooBigCheck.java
```diff
@@ -0,0 +1,26 @@
+class A {
+
+  Callable<Integer> c1 = () -> { // Noncompliant [[sc=26;ec=31;secondary=13]] {{Reduce this lambda expression number of lines from 11 to at most 10.}}
```
m-g-sonar: I know you copy-pasted this code from old implementation, but this does not compile.
Please use the adequate import, or define a `Callable` class to be used in such context.

java-checks/src/test/files/checks/LambdaTooBigCheck.java
```diff
@@ -0,0 +1,26 @@
...
+    System.out.println();
+    System.out.println();
+    System.out.println();
+    System.out.println();
+    System.out.println();
+    return 1;
+  };
+
+  Callable<Integer> c2 = () -> {
+    System.out.printtln();
```
m-g-sonar: typo. does not compile, and Callable probably needs a return statement.

java-checks/src/test/files/checks/LambdaTooBigCheckCustom.java
```diff
@@ -0,0 +1,41 @@
+class A {
+
+  private void f() {
+    Callable<Integer> c0 = () -> {
```
m-g-sonar: Same comments as earlier. Does not compile. `Callable`, `Predicate`, `List` types are unknown and need to be imported.

PR #2685, opened 2019-11-11
SONARJAVA-3177 FP on S1128 when import used in class annotations

amelie-renard-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/UselessImportCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/UselessImportCheck.java
@@ -26,6 +26,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import javax.annotation.CheckForNull;
 import org.sonar.check.Rule;
 import org.sonar.java.RspecKey;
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/UselessImportCheck.java
```diff
@@ -129,13 +130,29 @@ private void checkSymbolUsage(ImportTree importTree, String importName) {
...
         // Exclude method symbols : they could be ambiguous or unresolved and lead to FP.
-        if (symbol.isVariableSymbol() && symbol.usages().stream().allMatch(identifierTree -> JUtils.enclosingClass(identifierTree) == owner)) {
+        if (symbol.isVariableSymbol() && symbol.usages().stream().allMatch(identifierTree -> enclosingClass(identifierTree) == owner)) {
           context.reportIssue(this, importTree, "Remove this unused import '" + importName + "'.");
         }
       }
     }
   }
 
+  private static Symbol enclosingClass(Tree t) {
```
m-g-sonar: As it can return null (line 143), please annotated the method with `@CheckForNull`.

PR #2690, opened 2019-11-12
SONARJAVA-3189 S1989 should report only once per method invocation

guillaume-dequenne-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ServletMethodsExceptionsThrownCheck.java
```diff
@@ -104,20 +104,43 @@ private void checkMethodInvocation(MethodInvocationTree node) {
...
-        addIssueIfNotCatched(types, ExpressionUtils.methodName(node), "Add a \"try/catch\" block for \"" + symbol.name() + "\".");
+        addIssueIfNotCaught(types, ExpressionUtils.methodName(node), symbol.name());
       }
     }
   }
 
-  private void addIssueIfNotCatched(Iterable<Type> thrown, Tree node, String message) {
+  private void addIssueIfNotCaught(Type thrown, Tree node) {
+    if (isNotCaught(thrown)) {
+      String message = "Handle the " + "\"" + thrown + "\"" + " thrown here in a \"try/catch\" block.";
```
m-g-sonar: Never rely on the ToString() method of a type. What if it change in implementation class for debugging purpose?
Use `thrown.name()` instead.

java-checks/src/main/java/org/sonar/java/checks/ServletMethodsExceptionsThrownCheck.java
```diff
@@ -104,20 +104,43 @@ private void checkMethodInvocation(MethodInvocationTree node) {
...
+      reportIssue(node, message);
+    }
+  }
+
+  private void addIssueIfNotCaught(Iterable<Type> thrown, Tree node, String symbolName) {
```
m-g-sonar: please rename parameter `symbolName` in `methodName`, it will make the code easier to read.

java-checks/src/main/java/org/sonar/java/checks/ServletMethodsExceptionsThrownCheck.java
```diff
@@ -104,20 +104,43 @@ private void checkMethodInvocation(MethodInvocationTree node) {
...
+    }
+    if (!uncaughtTypes.isEmpty()) {
+      reportIssue(node, buildMessage(symbolName, uncaughtTypes));
+    }
+  }
+
+  private static String buildMessage(String symbolName, List<Type> uncaughtTypes) {
+    StringBuilder stringBuilder = uncaughtTypes.size() == 1 ? new StringBuilder("Handle the following exception that could be thrown by " + "\"" + symbolName + "\": ") :
+      new StringBuilder("Handle the following exceptions that could be thrown by " + "\"" + symbolName + "\": ");
+    for (int i = 0; i < uncaughtTypes.size(); i++) {
```
m-g-sonar: Instead of relying on a string builder, you can directly stream it from the initial list:
```
uncaughtTypes.stream().map(Type::name).collect(Collectors.join(", "));
```

PR #2693, opened 2019-11-12
SONARJAVA-3191 S1905 should also raise issues in method invocation and constructor call arguments

amelie-renard-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
@@ -58,14 +58,16 @@ public class RedundantTypeCastCheck extends IssuableSubscriptionVisitor {
       return;
     }
     TypeCastTree typeCastTree = (TypeCastTree) tree;
-    Type cast = typeCastTree.type().symbolType();
-    Type target = targetType(typeCastTree);
     Type expressionType = typeCastTree.expression().symbolType();
     if (isPrimitiveWrapperInConditional(expressionType, typeCastTree) || requiredForMemberAccess(typeCastTree)) {
       // Primitive wrappers excluded because covered by S2154
       return;
     }
-    if (target != null && (isRedundantNumericalCast(cast, expressionType) || isUnnecessarySubtypeCast(expressionType, typeCastTree, target))) {
+    Type cast = typeCastTree.type().symbolType();
+    Type target = targetType(typeCastTree);
+    boolean isArgument = skipParentheses(typeCastTree.parent()).is(Tree.Kind.ARGUMENTS);
+    if (target != null &&
+      (isRedundantNumericalCast(cast, expressionType) || isUnnecessarySubtypeCast(expressionType, typeCastTree, target, isArgument))) {
       reportIssue(typeCastTree.type(), "Remove this unnecessary cast to \"" + cast.erasure() + "\".");
     }
   }
@@ -140,9 +142,10 @@ public class RedundantTypeCastCheck extends IssuableSubscriptionVisitor {
     return skip;
   }
 
-  private static boolean isUnnecessarySubtypeCast(Type childType, TypeCastTree typeCastTree, Type parentType) {
+  private static boolean isUnnecessarySubtypeCast(Type childType, TypeCastTree typeCastTree, Type parentType, boolean isArgument) {
     return !childType.isPrimitive()
-      && childType.isSubtypeOf(parentType)
+      // Exception: subtype cast are tolerated in method or constructor call arguments
+      && ((isArgument && childType.equals(parentType)) || (!isArgument && childType.isSubtypeOf(parentType)))
       && (!ExpressionUtils.skipParentheses(typeCastTree.expression()).is(Tree.Kind.LAMBDA_EXPRESSION)
         || isUnnecessaryLambdaCast(childType, parentType));
   }
--- a/java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
@@ -33,6 +33,7 @@ import org.sonar.java.resolve.TypeVariableJavaType;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.Arguments;
 import org.sonar.plugins.java.api.tree.ArrayAccessExpressionTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodInvocationTree;
@@ -114,16 +115,10 @@ public class RedundantTypeCastCheck extends IssuableSubscriptionVisitor {
         return null;
       case VARIABLE:
         return ((VariableTree) parent).symbol().type();
-      case METHOD_INVOCATION:
-        // FIXME(SONARJAVA-3191) dead code
-        // If placed in invocation arguments, parent of typecast will be ARGUMENTS, not METHOD_INVOCATION
-        MethodInvocationTree mit = (MethodInvocationTree) parent;
-        if (mit.symbol().isMethodSymbol()) {
-          Symbol.MethodSymbol sym = (Symbol.MethodSymbol) mit.symbol();
-          int castArgIndex = mit.arguments().indexOf(typeCastTree);
-          return sym.parameterTypes().get(castArgIndex);
-        }
-        return null;
+      case ARGUMENTS:
+        Arguments arguments = (Arguments)parent;
+        int castArgIndex = arguments.indexOf(typeCastTree);
+        return arguments.get(castArgIndex).symbolType();
       case MEMBER_SELECT:
       case CONDITIONAL_EXPRESSION:
         return typeCastTree.type().symbolType();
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -57,14 +58,16 @@ public void visitNode(Tree tree) {
...
-    Type target = targetType(typeCastTree);
     Type expressionType = typeCastTree.expression().symbolType();
     if (isPrimitiveWrapperInConditional(expressionType, typeCastTree) || requiredForMemberAccess(typeCastTree)) {
       // Primitive wrappers excluded because covered by S2154
       return;
     }
-    if (target != null && (isRedundantNumericalCast(cast, expressionType) || isUnnecessarySubtypeCast(expressionType, typeCastTree, target))) {
+    Type cast = typeCastTree.type().symbolType();
+    Type target = targetType(typeCastTree);
+    boolean isArgument = skipParentheses(typeCastTree.parent()).is(Tree.Kind.ARGUMENTS);
```
m-g-sonar: You only need `isArgument` within method `isUnnecessarySubtypeCast()`, and you can computed it from there, as you are anyway using passing `typeCastTree` as parameter.

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -114,16 +117,10 @@ private static Type targetType(TypeCastTree typeCastTree) {
...
-        // FIXME(SONARJAVA-3191) dead code
-        // If placed in invocation arguments, parent of typecast will be ARGUMENTS, not METHOD_INVOCATION
-        MethodInvocationTree mit = (MethodInvocationTree) parent;
-        if (mit.symbol().isMethodSymbol()) {
-          Symbol.MethodSymbol sym = (Symbol.MethodSymbol) mit.symbol();
-          int castArgIndex = mit.arguments().indexOf(typeCastTree);
-          return sym.parameterTypes().get(castArgIndex);
-        }
-        return null;
+      case ARGUMENTS:
```
m-g-sonar: nice move, I didn't thought about this when creating the ticket! :+1:

java-checks/src/test/files/checks/RedundantTypeCastCheck.java
```diff
@@ -18,16 +18,16 @@
...
     A a = (A) new B(); // Noncompliant {{Remove this unnecessary cast to "A".}}
     A[][] as = (A[][]) new B[1][1]; // Noncompliant {{Remove this unnecessary cast to "A[][]".}}
     B b;
     fun(b);
-    fun((A) b);
+    fun((A) b); // Compliant - exception to distinguish the method to call
     List<B> bees = new java.util.ArrayList<B>();
     java.util.List<A> aaas = (java.util.List) bees;
-    C c = new C((A)null);
+    C c = new C((A)null); // Compliant - exception to distinguish the constructor to call
```
m-g-sonar: I would add a new constructor call to check that we are new raising issues in newClassTree as well.

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -114,16 +115,10 @@ private static Type targetType(TypeCastTree typeCastTree) {
...
+        Arguments arguments = (Arguments)parent;
+        int castArgIndex = arguments.indexOf(typeCastTree);
+        return arguments.get(castArgIndex).symbolType();
```
m-g-sonar: Sorry I didn't catch it earlier... this won't work. What you wrote here is not equivalent to what we had earlier.
What is here now is this equivalent to do : `typeCastTree.symbolType()`, so the type of the resulting cast. We already have this information and that's not what we are searching for.

Previously, we were looking at the symbol of the method which was called, and looking at the EXPECTED type, as defined in his declaration.

In the following example: previously we would have retrieved `Collection`, today we are getting `ArrayList`

```
void foo(Collection param) { ... }
//       ^^^^^^^^^^ type from declaration = Collection

void bar(List param) {
  foo((ArrayList) param);
//     ^^^^^^^^^ type from cast = ArrayList
}
```

So in fact probably you will still need to get to the NewClassTree or MethodInvocationTree, but the symbol corresponding is only one parent ahead.

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -114,16 +115,10 @@ private static Type targetType(TypeCastTree typeCastTree) {
...
-        // If placed in invocation arguments, parent of typecast will be ARGUMENTS, not METHOD_INVOCATION
-        MethodInvocationTree mit = (MethodInvocationTree) parent;
-        if (mit.symbol().isMethodSymbol()) {
-          Symbol.MethodSymbol sym = (Symbol.MethodSymbol) mit.symbol();
-          int castArgIndex = mit.arguments().indexOf(typeCastTree);
-          return sym.parameterTypes().get(castArgIndex);
-        }
-        return null;
+      case ARGUMENTS:
+        Arguments arguments = (Arguments)parent;
```
m-g-sonar: missing parenthesis after the cast.

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -57,14 +58,14 @@ public void visitNode(Tree tree) {
...
-    Type target = targetType(typeCastTree);
     Type expressionType = typeCastTree.expression().symbolType();
     if (isPrimitiveWrapperInConditional(expressionType, typeCastTree) || requiredForMemberAccess(typeCastTree)) {
       // Primitive wrappers excluded because covered by S2154
       return;
     }
-    if (target != null && (isRedundantNumericalCast(cast, expressionType) || isUnnecessarySubtypeCast(expressionType, typeCastTree, target))) {
+    Type cast = typeCastTree.type().symbolType();
+    Type target = targetType(typeCastTree);
+    if (target != null &&(isRedundantNumericalCast(cast, expressionType) || isUnnecessarySubtypeCast(expressionType, typeCastTree, target))) {
```
m-g-sonar: missing parenthesis after `&&` . this line is unchanged.

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
amelie-renard-sonarsource: You're totally right. Should be fixed

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -114,14 +116,14 @@ private static Type targetType(TypeCastTree typeCastTree) {
...
-          return sym.parameterTypes().get(castArgIndex);
+      case ARGUMENTS:
+        Arguments arguments = (Arguments) parent;
+        if (arguments.parent().is(Tree.Kind.METHOD_INVOCATION)) {
+          MethodInvocationTree mit = (MethodInvocationTree) arguments.parent();
+          return targetTypeFromMethodSymbol(mit.symbol(), mit.arguments(), typeCastTree);
+        } else if (arguments.parent().is(Tree.Kind.NEW_CLASS)) {
+          NewClassTree nct = (NewClassTree) arguments.parent();
+          return targetTypeFromMethodSymbol(nct.constructorSymbol(), nct.arguments(), typeCastTree);
         }
```
m-g-sonar: It can be even more simplified in order to avoid calling 4 times `arguments.parent()`, and reusing `arguments`. It also makes code a bit easier to read:
```java
Arguments arguments = (Arguments) parent;
Tree invocation = arguments.parent();
if (invocation.is(Tree.Kind.METHOD_INVOCATION)) {
  return targetTypeFromMethodSymbol(((MethodInvocationTree) invocation).symbol(), arguments, typeCastTree);
}
// no need for else, early return
if (invocation.is(Tree.Kind.NEW_CLASS)) {
  return targetTypeFromMethodSymbol(((NewClassTree) invocation).constructorSymbol(), arguments, typeCastTree);
}
return null;
```
In term of code coverage, the only 3rd case of Arguments tree being used is when using annotations (see `AnnotationTree`). It worth adding a test case (if not already present). Note that for me it's OK if we don't consider these cases, as annotations parameters are always constant expression... I don't see many cases where wrong/redundant type cast would occur

```java
public class A {

  @MyAnnotation((int) (0L + 42))
             // ^^^^^^^^^^^^^^^ type cast in arguments from AnnotationTree
  Object field;

  @interface MyAnnotation {
    int value() default 42;
  }
}
```

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -53,17 +55,17 @@
 
   @Override
   public void visitNode(Tree tree) {
-    if(!hasSemantic()) {
+    if (!hasSemantic()) {
```
m-g-sonar: unnecessary change :smile:

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -53,17 +55,17 @@
...
     }
     TypeCastTree typeCastTree = (TypeCastTree) tree;
-    Type cast = typeCastTree.type().symbolType();
-    Type target = targetType(typeCastTree);
     Type expressionType = typeCastTree.expression().symbolType();
     if (isPrimitiveWrapperInConditional(expressionType, typeCastTree) || requiredForMemberAccess(typeCastTree)) {
       // Primitive wrappers excluded because covered by S2154
       return;
     }
+    Type cast = typeCastTree.type().symbolType();
```
m-g-sonar: :+1:

m-g-sonar:
LGTM!

PR #2706, opened 2019-11-14
SONARJAVA-2439 Assert correct type inference and no FP on S2699

guillaume-dequenne-sonarsource:


amelie-renard-sonarsource:
LGTM

PR #2712, opened 2019-11-15
SONARJAVA-3178 Fix S2200 does not handle properly unary post/pre-fix increment/decrement

amelie-renard-sonarsource:


m-g-sonar:
LGTM!

PR #2715, opened 2019-11-15
SONARJAVA-3228 Handle NPE when resolving super type in interface

m-g-sonar:


m-g-sonar:
@amelie-renard-sonarsource @guillaume-dequenne-sonarsource please do NOT squash the PR when merging.

java-checks/src/test/java/org/sonar/java/filters/FilterVerifier.java
```diff
@@ -143,4 +150,23 @@ public void visitTrivia(SyntaxTrivia syntaxTrivia) {
...
+    context.setSettings(new MapSettings().setProperty("sonar.java.failOnException", true));
+    SonarComponents sonarComponents = new SonarComponents(null, context.fileSystem(), null, null, null) {
+      @Override
+      public boolean reportAnalysisError(RecognitionException re, InputFile inputFile) {
+        return false;
+      }
+
+      @Override
+      public void addAnalysisError(AnalysisError analysisError) {
+        Fail.fail(String.format("Should not fail analysis (%s)", analysisError.getKind().name()));
```
guillaume-dequenne-sonarsource: Maybe there is something I'm missing but I feel like there should be a test case triggering this. Currently I can just comment this change and all tests are passing anyway. Even when leaving out your fix on the other commit, I've been unable to reach this (I'm pretty sure I'm doing something wrong here - but I still think we should have an easier way to make sure this does its intended job).

m-g-sonar: Added a test case.

amelie-renard-sonarsource:
LGTM

PR #2720, opened 2019-11-18
SONARJAVA-2795 Fix FP in S1612 when method reference is ambiguous

saberduck:


java-checks/src/main/java/org/sonar/java/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -46,30 +45,31 @@
 import org.sonar.plugins.java.api.tree.VariableTree;
 
 @Rule(key = "S1612")
-public class ReplaceLambdaByMethodRefCheck extends BaseTreeVisitor implements JavaFileScanner {
+public class ReplaceLambdaByMethodRefCheck extends SubscriptionVisitor {
```
m-g-sonar: While I do appreciate the change, it has nothing to do with handling the FP in itself. 

For the sack of consistency and clarity when looking at changes, please split it in two commits:
1. Changing to a SubscriptionVisitor + all the style changes (missing spaces, changes of tabs), which are unrelated
2. Update the rule to handle the FP

java-checks/src/test/files/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -168,3 +169,41 @@ void test(Object param) {
     bar(new D(param)::foo);
   }
 }
+
+class SONARJAVA_2795 {
```
m-g-sonar: I'm not a huge fan of using ticket name in test cases. Can you explicitly name this class with what it tests? The ticket can be retrieved by blaming the test file, while when reading the test, it's easier to read the "why".

PR #2722, opened 2019-11-18
SONARJAVA-1706 investigation

saberduck:
It looks that both invocations are resolved, but the type of the argument is different.

Nothing to merge, just investigating the behavior. You can close the PR unless there is something I missed.


amelie-renard-sonarsource:
LGTM

m-g-sonar:
LGTM!

PR #2728, opened 2019-11-19
SONARJAVA-2952 S2175 detects issues in types extending standard collections

amelie-renard-sonarsource:


java-checks/src/test/files/checks/CollectionInappropriateCallsCheck.java
```diff
@@ -254,3 +254,18 @@ public void test(Set<Integer> used, List<List<Integer>> allInts) {
...
+class Color {}
+abstract class ColorList implements java.util.List<Color> {}
+abstract class SpecificColorList extends ColorList {}
+
+class TestColorList {
+  void foo(ColorList colors, SpecificColorList specificColors, Color c, Animal a) {
+    colors.remove(c);
+    specificColors.remove(c);
+
+    colors.remove(a); // Noncompliant
```
m-g-sonar: I feel this is not right. If you look at the resulting message of the issue, we are getting the following:
```java
colors.remove(a); // Noncompliant {{A "ColorList<Color>" cannot contain a "Animal"}}
specificColors.remove(a); // Noncompliant {{A "SpecificColorList<Color>" cannot contain a "Animal"}}
```
More particularly, `ColorList` is not a generic type, and it can not expressed as `ColorList<Color>`. Consequently, the message, although correct on semantic basis, is not corresponding to reality. Same comment for the next one, which is even a level lower.

I would suggest to use a different approach in these cases, to report something such as:
 `"ColorList" is a "List<Color>", which cannot contain a "Animal"`

PR #2732, opened 2019-11-20
SONARJAVA-2939 an issue is raised when an "AutoCloseable" resource is initialized outside of a try-catch block

amelie-renard-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/TryWithResourcesCheck.java
```diff
@@ -55,10 +57,64 @@ public void leaveFile(JavaFileScannerContext context) {
...
+    }
+
+    if (blockParent != null) {
+      CFG cfg = CFG.buildCFG(Collections.singletonList(blockParent));
+      if (!cfg.blocks().isEmpty()) {
+        return newFollowedByTryStatement(cfg.blocks().get(0));
+      }
+    }
+
+    return false;
```
m-g-sonar: The 2 "return false" are unreachable by construction, so I would add a comment on top of each to make it explicit that it's defensive programming.

PR #2742, opened 2019-11-25
SONARJAVA-3082 Restore deprecated method JavaFileScannerContext.getFileKey()

m-g-sonar:


saberduck:
@m-g-sonar lgtm, you will need to accept code smells on the PR

PR #2399, opened 2019-01-24
SONARJAVA-3059: S3553 also check Guava Optional

bannmann:
Guava's `Optional` was the inspiration for the JDK `Optional`. Although it is different in some aspects (serialization, being recommended for use as collection elements), using it as a parameter type causes exactly the same problems as for JDK `Optional`.

- [x] Use the following formatting style: [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)
- [x] Unit tests are passing and you provided a unit test for your fix
- [x] ITs should pass : To run ITs locally, checkout the [README](https://github.com/SonarSource/sonar-java/blob/master/README.md) of the project.
- [ ] If there is a [Jira](http://jira.sonarsource.com/browse/SONARJAVA) ticket available, please make your commits and pull request start with the ticket number (SONARJAVA-XXXX)

PR #2763, opened 2020-01-02
SONARJAVA-3272 - S2789: also cover Guava Optional

bannmann:
PR #2399 changed `S3553` ("Optional" should not be used for parameters) to also cover Guava. `S2789` ("null" should not be used with "Optional") should do likewise.

----


m-g-sonar:
Hey Jens,

Thanks also for the contribution on S2789. Like for the other rules, I created the following ticket to track the improvement of the rule: [SONARJAVA-3272](https://jira.sonarsource.com/browse/SONARJAVA-3272).

Now, regarding your fix, I'm really not sure it works in production (I'll test it right now and let you know).

Because you are using different constructors for the rule to be used, I fear it might break the normal behavior, if not SonarQube. I especially don't see any default constructor anymore, which I fear is the one used by SonarQube to instantiate the rules. Also, from my point of view, both `Optional` should be always considered by the rule, which I fear is not the case with this approach. What if you combine both in a file?

Could you update your fix to have both Optionals always considered?

Cheers,
Michael

bannmann:
> Could you update your fix to have both Optionals always considered?

Oh right - in hindsight, it's obvious that it could never work this way. I'll rework the code accordingly.

m-g-sonar:
LGTM! Thanks for the contribution and its rework.
As usual, I'm going to open a new PR to trigger QA and merge eventually.

Cheers,
Michael

m-g-sonar:
Merged & squashed the other PR. Thanks again!

PR #2772, opened 2020-01-27
SONARJAVA-3240 Fixing handling of exceptions in rules

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
```diff
@@ -175,22 +175,20 @@ private void runScanner(Runnable action, JavaFileScanner scanner, AnalysisError.
...
-          scanner.getClass(), ruleKey, currentFile),
-        e);
+      String message = String.format(
+        "Unable to run check %s - %s on file '%s', To help improve SonarJava, please report this problem to SonarSource : see https://www.sonarqube.org/community/",
+        scanner.getClass(), ruleKey, currentFile);
+
+      LOG.error(message, e);
       addAnalysisError(e, currentFile, kind);
 
-      // for IssuableSubscriptionVisitors as well as for SymbolicExecution, this should lead to changing file
```
quentin-jaquier-sonarsource: In the current situation, if you want to implement this behavior, you can remove the two try/catch in `visitFile()`, the `CheckFailureException` will be catched in `JavaAstScanner::simpleScan` and fail fast if needed.

However:

- in the case of IssuableSubscriptionVisitors, it will not change anything as it is the last step of the visit.
- do we really want to report no issue on a file who fails SymbolicExecution checks?

PR #2775, opened 2020-01-30
SONARJAVA-3279 S1309 (Track uses of "@SuppressWarnings" annotations) does not support correctly changes of repository name

alban-auzeill:


m-g-sonar:
LGTM! Nice work!

PR #2781, opened 2020-02-03
SONARJAVA-3209 Rule S5542: Encryption algorithms should be used with secure mode and padding scheme

quentin-jaquier-sonarsource:


java-checks/src/test/files/checks/security/EncryptionAlgorithmCheck.java
```diff
@@ -0,0 +1,67 @@
...
+    - everytime ECB mode is used whatever the encryption algorithm
+       - By default without specifying operation mode ECB is chosen
+    - when CBC mode is used with PKCS5Padding or PKCS7Padding
+    - when RSA is used without OAEPWithSHA-1AndMGF1Padding or OAEPWITHSHA-256ANDMGF1PADDING padding scheme
+    */
+
+    try
+    {
+      // First case
+      Cipher c0 = Cipher.getInstance("AES"); // Noncompliant [[sc=38;ec=43]] {{Use secure mode and padding scheme.}}
```
m-g-sonar: Why complicating stuff with variable assignment?
Just drop all the `Cipher cX =` statements, they are useless and simply add some noise in test file.

java-checks/src/test/files/checks/security/EncryptionAlgorithmCheck.java
```diff
@@ -0,0 +1,67 @@
...
+  public void foo(java.util.Properties props) {
+    /*
+    should complain:
+    - everytime ECB mode is used whatever the encryption algorithm
+       - By default without specifying operation mode ECB is chosen
+    - when CBC mode is used with PKCS5Padding or PKCS7Padding
+    - when RSA is used without OAEPWithSHA-1AndMGF1Padding or OAEPWITHSHA-256ANDMGF1PADDING padding scheme
+    */
+
+    try
```
m-g-sonar: usually, in java, brackets are on the same line as the keyword. Can you fix this?

java-checks/src/test/files/checks/security/EncryptionAlgorithmCheck.java
```diff
@@ -0,0 +1,67 @@
...
+      Cipher c16 = Cipher.getInstance("RSA/NONE/OAEPWITHSHA-256ANDMGF1PADDING"); // Compliant
+
+      // Other
+      Cipher.getInstance(null);
+      Cipher.getInstance(""); // Noncompliant
+      String algo = props.getProperty("myAlgo", "AES/ECB/PKCS5Padding");
+      Cipher c17 = Cipher.getInstance(algo); // Noncompliant
+      String s = "RSA/NONE/NoPadding";
+    }
+    catch(NoSuchAlgorithmException|NoSuchPaddingException e)
```
m-g-sonar: Why not replacing by a single `catch (Exception e) {}`, or simply define `foo` as throwing `Exception` ? It would get rid of the try-catch block.

java-checks/src/main/java/org/sonar/java/checks/security/EncryptionAlgorithmCheck.java
```diff
@@ -0,0 +1,84 @@
...
+@Rule(key = "S5542")
+public class EncryptionAlgorithmCheck extends AbstractMethodDetection {
+
+  private static final Pattern ALGORITHM_PATTERN = Pattern.compile("(.+)/(.+)/(.+)");
+
+  @Override
+  protected List<MethodMatcher> getMethodInvocationMatchers() {
+    return Collections.singletonList(
+      MethodMatcher.create()
+        .typeDefinition(TypeCriteria.is("javax.crypto.Cipher"))
```
m-g-sonar: You can directly use the string is you are using `TypeCriteria.is()`

java-checks/src/main/java/org/sonar/java/checks/security/EncryptionAlgorithmCheck.java
```diff
@@ -0,0 +1,84 @@
...
+    ExpressionTree algorithmTree = defaultPropertyValue == null ? firstArgument : defaultPropertyValue;
+    String algorithmName = ExpressionsHelper.getConstantValueAsString(algorithmTree).value();
+    if (algorithmName != null && isInsecureAlgorithm(algorithmName)) {
+      reportIssue(firstArgument, "Use secure mode and padding scheme.");
+    }
+  }
+
+  private static boolean isInsecureAlgorithm(String algorithmName) {
+    Matcher matcher = ALGORITHM_PATTERN.matcher(algorithmName);
+    if (matcher.find()) {
```
m-g-sonar: I'm not quite sure about usage of `find()` here. If we are searching for the entire string, I would recommend using `matches()` instead. I don't see any test case which would justify using finding for subsequences rather than on everything.

java-checks/src/main/java/org/sonar/java/checks/security/EncryptionAlgorithmCheck.java
```diff
@@ -0,0 +1,84 @@
...
+      String algorithm = matcher.group(1);
+      String mode = matcher.group(2);
+      String padding = matcher.group(3);
+
+      if (mode.equals("ECB")) {
```
m-g-sonar: for all these string comparison, please invert the string constant with the variable. See rule [S1132](https://next.sonarqube.com/sonarqube/coding_rules?open=java%3AS1132&rule_key=java%3AS1132).

java-checks/src/main/java/org/sonar/java/checks/security/EncryptionAlgorithmCheck.java
quentin-jaquier-sonarsource: Good point

PR #2782, opened 2020-02-03
SONARJAVA-3207 Rule S5547: Cipher algorithms should be robust

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/StrongCipherAlgorithmCheck.java
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  protected void onConstructorFound(NewClassTree newClassTree) {
+    reportIssue(newClassTree.identifier(), MESSAGE);
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    ExpressionTree firstArg = mit.arguments().get(0);
+    ExpressionTree value = getValueOrPropertyDefaultValue(firstArg);
+    if (value.is(Tree.Kind.STRING_LITERAL)) {
```
quentin-jaquier-sonarsource: In [this check](https://github.com/SonarSource/sonar-java/blob/2a8504690070195c4084ed97206afd10d7e05a1b/java-checks/src/main/java/org/sonar/java/checks/security/EncryptionAlgorithmCheck.java#L56), we did something similar and we used `ExpressionsHelper.getConstantValueAsString`.  We may want to use it to be consistent.
It would also avoid to trim the quotes at line 74.

java-checks/src/test/files/checks/StrongCipherAlgorithmCheck.java
```diff
@@ -0,0 +1,99 @@
...
+
+class MyCipher extends Cipher {
+  public MyCipher() {
+    super(null, null, "");
+  }
+}
+
+class A {
+  void foo() throws NoSuchAlgorithmException, NoSuchPaddingException {
+    Cipher c;
```
quentin-jaquier-sonarsource: I think it's better to remove the assignments for clarity.

java-checks/src/test/files/checks/StrongCipherAlgorithmCheck.java
```diff
@@ -0,0 +1,99 @@
...
+    c = Cipher.getInstance("Blowfish"); // Noncompliant
+    c = Cipher.getInstance("Blowfish/ECB"); // Noncompliant
+    c = Cipher.getInstance("Blowfish/ECB/PKCS5Padding"); // Noncompliant
+    c = Cipher.getInstance("Blowfish/GCM"); // Noncompliant
+    c = Cipher.getInstance("Blowfish/GCM/NoPadding"); // Noncompliant
+    c = Cipher.getInstance("Blowfish/GCM/PKCS5Padding"); // Noncompliant
+
+    c = Cipher.getInstance("AES/GCM/NoPadding"); // Compliant
+
+    c.getBlockSize();
```
quentin-jaquier-sonarsource: What is this call?

alban-auzeill: it was to remove:
`WARNING StrongCipherAlgorithmCheck.java:70: The value of the local variable c is not used`
But I will apply your above proposal, remove `Cipher c;`

PR #2783, opened 2020-02-03
SONARJAVA-3202 Introduce rules S5527

m-g-sonar:


java-checks/src/test/files/checks/security/VerifiedServerHostnamesCheck/HostnameVerifier.java
```diff
@@ -0,0 +1,102 @@
+import java.util.function.BiFunction;
```
alban-auzeill: `WARNING HostnameVerifier.java:1: The import java.util.function.BiFunction is never used`

m-g-sonar: good catch, I dropped the line which was using it, for no reason.

PR #2785, opened 2020-02-04
SONARJAVA-3253 filter out queries from S2068

m-g-sonar:


java-checks/src/test/files/checks/HardCodedCredentialsCheck.java
```diff
@@ -17,6 +17,12 @@ private void a(char[] pwd, String var) {
     String variable5 = "login=a&password=";
     String variable6 = "login=a&password= ";
 
+    String query1 = "password=?"; // Compliant
+    String query4 = "password='" + pwd + "'"; // Compliant
+    String query2 = "password=:password"; // Compliant
+    String query3 = "password=:param"; // Compliant
```
pierre-loup-tristant-sonarsource: This use case is not limited to `:param` but to `:something`. This is to avoid FP when literal string is an Hibernate query parameters.
https://docs.jboss.org/hibernate/orm/3.2/api/org/hibernate/Query.html

m-g-sonar: Thanks for the feedback. fix on his way. It was not obvious to me from the description.

PR #2787, opened 2020-02-04
SONARJAVA-3183 S3329 should consider the call to SecureRandom.generat…

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -56,6 +70,24 @@ protected void onConstructorFound(NewClassTree newClassTree) {
     }
   }
 
+  private static boolean isDynamicallyGenerated(ExpressionTree tree) {
+    if (tree.is(Tree.Kind.IDENTIFIER)) {
+      Symbol symbol = ((IdentifierTree) tree).symbol();
+      VariableTree declaration = ((Symbol.VariableSymbol) symbol).declaration();
+      return declaration != null &&
+        (isSecureRandomGenerateSeed(declaration.initializer()) ||
+          getReassignments(symbol.declaration(), symbol.usages()).stream()
```
m-g-sonar: you should reuse `declaration` instead of `symbol.declaration()` here.

java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -56,6 +70,24 @@ protected void onConstructorFound(NewClassTree newClassTree) {
     }
   }
 
+  private static boolean isDynamicallyGenerated(ExpressionTree tree) {
+    if (tree.is(Tree.Kind.IDENTIFIER)) {
+      Symbol symbol = ((IdentifierTree) tree).symbol();
+      VariableTree declaration = ((Symbol.VariableSymbol) symbol).declaration();
```
m-g-sonar: `ECJ` can be quite tricky here. In case of strange resolution/recovery from its engine, it might assign an unknown symbol rather than a `VariableSymbol`. Before casting, I would check that the symbol is not unknown.

quentin-jaquier-sonarsource: Good to know!

PR #2795, opened 2020-02-06
SONARJAVA-3197 Rule S3330 should report on "setHttpOnly" method when set to "false" and not when initialized

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/security/CookieHttpOnlyCheck.java
```diff
@@ -339,11 +339,8 @@ private void updateIssuesToReport(MethodInvocationTree mit) {
       // ignore XSRF-TOKEN cookies
       return;
     }
-    if (setterArgumentHasCompliantValue(mit.arguments())) {
-      variablesToReport.remove(reference);
-    } else if (compliantConstructorInitializations.contains(reference)) {
```
quentin-jaquier-sonarsource: By removing this check, we can completely get rid of `compliantConstructorInitializations`, that was used only here!

PR #2797, opened 2020-02-07
SONARJAVA-3250 S2245 should consider methods from ThreadLocalRandom, JVMRandom, RandomUtils, RandomStringUtils

quentin-jaquier-sonarsource:


java-checks/src/test/files/checks/PseudoRandomCheck.java
```diff
@@ -1,10 +1,52 @@
...
+    // java.lang.Math. Report only Math.random()
+    double rand1 = Math.random(); // Noncompliant [[sc=25;ec=31]]
+    double abs = Math.abs(12); // Compliant
+
+    // java.util.concurrent.ThreadLocalRandom
+    int rand2 = ThreadLocalRandom.current().nextInt();  // Noncompliant
+
+    // org.apache.commons.lang.math.JVMRandom
+    JVMRandom jvmRandom = new JVMRandom(); // Noncompliant
+    double rand3 = jvmRandom.nextDouble(); // Noncompliant
```
alban-auzeill: We should report only 1 issue and not 2.
For class used statically "on the method invocation" (ex: `Math.random()`, `RandomUtils.nextFloat()`)
Otherwise, "on the new class" (ex: `new Random()`, `new JVMRandom()`)

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -44,15 +59,35 @@
...
     }
   }
 
+  private static boolean isChainedMethodInvocation(MethodInvocationTree mit) {
+    Tree parent = mit.parent();
+    return parent != null && parent.is(Tree.Kind.MEMBER_SELECT);
+  }
+
+  private void checkSymbolType(String fullyQualifiedName, Tree reportLocation) {
+    if (RANDOM_TYPES.stream().anyMatch(ty -> ty.equals(fullyQualifiedName))) {
```
alban-auzeill: `RANDOM_TYPES.contains(fullyQualifiedName)`

PR #2799, opened 2020-02-07
SONARJAVA-3261 S5122 should consider "HttpServletResponse.addHeader" in addition to "HttpServletResponse.setHeader"

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/CORSCheck.java
```diff
@@ -94,9 +100,9 @@ private void reportTree(Tree tree) {
 
     @Override
     public void visitMethodInvocation(MethodInvocationTree mit) {
-      if (SET_HEADER_MATCHER.matches(mit)) {
+      if (SET_ADD_HEADER_MATCHER.anyMatch(mit)) {
         String constantCORS = ExpressionsHelper.getConstantValueAsString(mit.arguments().get(0)).value();
-        if (HTTP_HEADERS.contains(constantCORS)) {
+        if (constantCORS != null && HTTP_HEADERS.contains(constantCORS.toLowerCase(Locale.ENGLISH))) {
```
alban-auzeill: We should add a test with `constantCORS == null`

PR #2805, opened 2020-02-11
SONARJAVA-3223 S5122 should support CorsRegistry from Spring

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/CORSCheck.java
```diff
@@ -106,6 +110,15 @@ private static boolean setSpecificOrigins(Tree tree) {
     return !isStar((LiteralTree) tree);
   }
 
+  private void reportTree(MethodInvocationTree mit) {
```
m-g-sonar: You can use `ExpressionUtils.methodName(mit)` for that.

java-checks/src/test/files/checks/CORSCheck.java
```diff
@@ -133,20 +136,40 @@ public CorsFilter corsFilter4() {
...
-    config.addAllowedOrigin("*"); // Noncompliant [[secondary=146,147]]
+    config.addAllowedOrigin("*"); // Noncompliant [[secondary=149,150]]
     config.applyPermitDefaultValues();
     config.applyPermitDefaultValues();
-    config.addAllowedOrigin("*"); // Noncompliant [[secondary=146,147]]
+    config.addAllowedOrigin("*"); // Noncompliant [[secondary=149,150]]
     return new CorsFilter(source);
   }
 
+  class S5122_Insecure implements WebMvcConfigurer {
```
m-g-sonar: I don't see the interest of such complicated methods to tests this. Keep the test case minimal to what we really tests.  For instance, invocations of `allowedMethods`, `allowCredentials`, etc. are not related to the behavior of the rule.

PR #2806, opened 2020-02-11
SONARJAVA-3292 Expose as API utility methods resolving expression as constant

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/ExpressionUtils.java
```diff
@@ -151,4 +153,72 @@ public static boolean isThis(ExpressionTree expression) {
     return newExpression.is(Tree.Kind.IDENTIFIER) && "this".equals(((IdentifierTree) newExpression).name());
   }
 
+  public static Object resolveAsConstant(ExpressionTree tree) {
```
benzonico: I think this lacks the @checkForNull annotation

PR #2809, opened 2020-02-12
SONARJAVA-3281 Fix FPs on S1854 and S1481 when variable is used inside a lambda and bytecode is incomplete

m-g-sonar:


quentin-jaquier-sonarsource:
I saw that `UnusedMethodParameterCheck` is also implementing [a similar visitor](https://github.com/SonarSource/sonar-java/blob/ab10b6ecfaa502b96d314c66d0862714acbe400a/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java#L190), with a different implementation.

Using this visitor instead would avoid a few FN such as:

```java
int a = 42; // FN, it's unused
lambdas_not_resolved(y -> a() + y); // method a() is unresolved
```
```java
int a = 42; // FN
lambdas_not_resolved(y -> A.a + y);
```

We should consider making one visitor common for the three checks.

m-g-sonar:
Pretty good catch. I'm gonna use the same visitor for the 3 rules.

PR #2810, opened 2020-02-12
Fix quality flaw: drop ConstantUtils from checks and rely on new API

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/helpers/ExpressionsHelper.java
```diff
@@ -85,32 +86,32 @@ public static Tree reportOnClassTree(ClassTree classTree) {
 
 
   public static ValueResolution getConstantValue(ExpressionTree expression) {
```
quentin-jaquier-sonarsource: `getConstantValue` and `getConstantValueAsInteger` are not used, and not tested, we might want to remove them.

PR #2813, opened 2020-02-12
SONARJAVA-3294 S4426: duplicates and FP when multiple declaration of KeyPairGenerator in the same method

quentin-jaquier-sonarsource:


java-checks/src/test/files/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -26,8 +32,51 @@ public void report_twice() throws NoSuchAlgorithmException {
...
+  public void key_variable_this() throws NoSuchAlgorithmException {
+    KeyPairGenerator keyGen = KeyPairGenerator.getInstance("RSA");
+    this.keyGen.initialize(1024); // Compliant, only support when declared in the same method
+  }
+
+  public void key_variable_instance() throws NoSuchAlgorithmException {
+    RSA2 rsa2 = new RSA2();
+    rsa2.keyGen = KeyPairGenerator.getInstance("RSA");
+    rsa2.setKeyGen();
+    rsa2.keyGen.initialize(1024); // Compliant, only support when declared in the same method
```
quentin-jaquier-sonarsource: I think it's better to not report member select, I expect this pattern to be super unusual, and FP prone if we don't dive deeply into the subject...

We will have FN in the case where `setKeyGen()` is not called, but I think it's acceptable.

Let me know what you think!

m-g-sonar: Fair point, LGTM!

PR #2815, opened 2020-02-13
SONARJAVA-3290 S3457 should allow to use 'toString()' on exceptions to force usage of sl4j Logger(String, Object...)

alban-auzeill:


its/ruling/src/test/resources/guava/java-S3457.json
```diff
@@ -0,0 +1,25 @@
+{
+'com.google.guava:guava:src/com/google/common/io/Closer.java':[
```
m-g-sonar: All these 9 issues are FPs. The issue message says: `Format specifiers should be used instead of string concatenation.`, but the method which is called is: `java.util.logging.Logger.log(Level, String, Throwable)`.

Consequently, you HAVE to use string concatenation

java-checks/src/test/files/checks/PrintfMisuseCheck.java
```diff
@@ -182,13 +185,18 @@ void foo(Calendar c){
     slf4jLog.debug("message {} {}", new Object[]{1, 2, 3}); // Noncompliant
     slf4jLog.debug("message {} {} {}", new Object[]{1, 2, 3}); // compliant
     slf4jLog.debug("message ", new Exception());
-    slf4jLog.debug("message {}", new Exception());
+    slf4jLog.debug("message {}", new Exception()); // Noncompliant {{Missing argument for the first parameter.}}
 
-    slf4jLog.error("message {}");
+    slf4jLog.error("message {}"); // Noncompliant {{Missing argument for the first parameter.}}
+    slf4jLog.error("message {}", new Exception()); // Noncompliant {{Missing argument for the first parameter.}}
```
m-g-sonar: to me the error message is wrong here. It should be something like `Remove this argument formatter because it won't be consumed`.

java-checks/src/test/files/checks/PrintfMisuseCheck.java
```diff
@@ -224,6 +232,9 @@ void foo(Calendar c){
     java.util.logging.Logger logger4 = getLog();
     logger4.log(java.util.logging.Level.WARNING, "som.foo.errorcode", 404); // Noncompliant
     this.loggerField.log(java.util.logging.Level.WARNING, "som.foo.errorcode", 404);
+    String additionalMessage = "...";
+    logger4.log(java.util.logging.Level.WARNING, "message " + additionalMessage, new Exception());  // Noncompliant {{Format specifiers should be used instead of string concatenation.}}
```
m-g-sonar: this is a FP to me. You don't have the choice to use concatenation if you want to use the "Throwable" method.

its/ruling/src/test/resources/guava/java-S3457.json
alban-auzeill: :tired_face: My understanding of `java.util.logging.Logger` was wrong.
I thought that:
```java
LOG.log(Level.WARNING, "Message: " + param, ex);
```
Is equivalent to:
```java
LOG.log(Level.WARNING, "Message: {0}", new Object[] {param, ex});
```
But, I tested it and it's not like slf4j. The last parameter of `log(Level level, String msg, Object params[])` is not handled like `thrown` in `log(Level level, String msg, Throwable thrown)`.

In conclusion, I also think these are FPs.

alban-auzeill: So it was 9 FPs, because of the wrong message: `Format specifiers should be used instead of string concatenation`
For those 9 issues, I changed the message to: `Lambda should be used to differ string concatenation.`
For example, in case of `src/com/google/common/io/Closer.java:252`:
```
      Closeables.logger.log(Level.WARNING,
          "Suppressing exception thrown when closing " + closeable, suppressed);
```
Can be changed into:
```
      Closeables.logger.log(Level.WARNING,
          suppressed, () -> "Suppressing exception thrown when closing " + closeable);
```

java-checks/src/main/java/org/sonar/java/checks/AbstractPrintfChecker.java
```diff
@@ -120,25 +133,29 @@ protected final void checkFormatting(MethodInvocationTree mit, boolean isMessage
...
       }
     } else {
-      handleOtherFormatTree(mit, formatTree);
+      handleOtherFormatTree(mit, formatTree, args);
     }
   }
 
+  private static boolean isProbablyLog4jFormatterLogger(MethodInvocationTree mit, String formatString) {
+    return mit.symbol().owner().type().is(ORG_APACHE_LOGGING_LOG4J_LOGGER) &&
+          (!formatString.contains("{}") && formatString.contains("%"));
```
m-g-sonar: useless parenthesis

its/ruling/src/test/resources/guava/java-S3457.json
FredWRobinson: I have a question about your new message, "Lambda should be used to differ string concatenation."  It makes no sense as written.

"Differ" means "is different from," as in "This thing differs from that thing because it is orange."

"Defer" means delay or postpone, as in "Let us defer that decision until next week."

Do you mean "Lambda should be used to **postpone/delay** string concatenation"?

m-g-sonar: Indeed @FredWRobinson, that's the second option which is correct. It's a typo. I created the following ticket to track the change: [SONARJAVA-3587](https://jira.sonarsource.com/browse/SONARJAVA-3587)

Note that we are not really following messages on closed pull requests. If you want to be sure to reach us, don't hesitate to use https://community.sonarsource.com/

Cheers,
Michael

PR #2816, opened 2020-02-13
SONARJAVA-3295 Handle null-case when binary is missing for package-info.java

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/JSema.java
```diff
@@ -148,7 +148,11 @@ ITypeBinding resolveType(String name) {
       }
 
       IBinaryType type = answer.getBinaryType();
-      // FIXME can be null in case of wrong package resolution?
+      if (type == null) {
+        // Can happen for instance with ant, as ant only generates 'package-info.class'
+        // when there is annotations in the package-info.java file.
+        return new IAnnotationBinding[0];
+      }
```
quentin-jaquier-sonarsource: LGTM, well done, a lot of investigation behind these two line of code!!

PR #2819, opened 2020-02-17
SONARJAVA-3298 S2755 FN when two factories are used in the same method

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/CryptographicKeySizeCheck.java
```diff
@@ -143,12 +123,9 @@ public void visitMethodInvocation(MethodInvocationTree mit) {
     }
 
     private boolean isSameVariableSymbol(MethodInvocationTree mit) {
-      Tree methodSelect = mit.methodSelect();
+      ExpressionTree methodSelect = mit.methodSelect();
       if (methodSelect.is(Tree.Kind.MEMBER_SELECT)) {
```
m-g-sonar: this can be simplified to:
```
return methodSelect.is(Tree.Kind.MEMBER_SELECT) 
  && extractIdentifierSymbol(((MemberSelectExpressionTree) methodSelect).expression()).filter(s -> s.equals(variable)).isPresent();
```

java-checks/src/main/java/org/sonar/java/checks/security/TriggeringSecuringHelper.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree methodInvocation) {
+      if (isSameVariableSymbol(methodInvocation)) {
+        processSecuringMethodInvocation(methodInvocation);
+      }
+      super.visitMethodInvocation(methodInvocation);
+    }
+
+    private boolean isSameVariableSymbol(MethodInvocationTree mit) {
```
m-g-sonar: Isn't this method the same as in: 
https://github.com/SonarSource/sonar-java/pull/2819/files#diff-ad75e35ac7cb0aad007e6d9e495bbae8R127

Maybe you can make it static and public, outside of the inner class, to be reused from the outsilde-

java-checks/src/main/java/org/sonar/java/checks/security/TriggeringSecuringHelper.java
```diff
@@ -0,0 +1,110 @@
...
+      ExpressionTree methodSelect = mit.methodSelect();
+      if (methodSelect.is(Tree.Kind.MEMBER_SELECT)) {
```
m-g-sonar: This can also be simplified.

java-checks/src/main/java/org/sonar/java/checks/security/TriggeringSecuringHelper.java
```diff
@@ -0,0 +1,110 @@
...
+ * 4. At the end of the visit, return true if the variable was secured inside the body of the enclosing method.
+ *
+ * It also makes sure that the triggering and securing method call is made on the same symbol.
+ *
+ */
+public abstract class TriggeringSecuringHelper implements Predicate<MethodInvocationTree> {
+
+  private final MethodMatcher triggeringInvocationMatcher;
+
+  public abstract void resetState();
```
m-g-sonar: javadoc is missing for this part. If it's required to call it at some point, then document why.

java-checks/src/test/files/checks/security/XmlExternalEntityProcessingCheck.java
```diff
@@ -1,5 +1,9 @@
+import javax.xml.parsers.ParserConfigurationException;
```
m-g-sonar: Could you take the opportunity to move the test file into the new module?

PR #2822, opened 2020-02-17
Change version to 6.1.1 for bugfix and sonarpedia date

alban-auzeill:


m-g-sonar:
We won't do a bugfix release for sonarcloud, but rely directly on latest artifact

PR #2836, opened 2020-02-19
Add java-check-test-sources as new test in ruling

m-g-sonar:


quentin-jaquier-sonarsource:
After discussion, we concluded that this was not a nice idea,
For every modification in a test file, you will change the expected issues lines of the current rule in the ruling, and potentially many others unrelated ones. 
There are not enough benefits to justify this additional noise, we decided to close this PR.

PR #2837, opened 2020-02-19
SONARJAVA-3212 FN on S2755 when XML parser is initialized inside initializer block

quentin-jaquier-sonarsource:


m-g-sonar:
Cancelled.

PR #2840, opened 2020-02-24
SONARJAVA-3302 S4042 Fix message when using File.delete

quentin-jaquier-sonarsource:
Internal PR of the external PR ##2839 of @dpursehouse.

Note: I amended an update of the rule description to match the correct message.


quentin-jaquier-sonarsource:
Thanks @dpursehouse for your contribution!

PR #2842, opened 2020-02-24
SONARJAVA-3260 RSPEC-2612 Setting loose file permissions is security-sensitive

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/security/FilePermissionsCheck.java
```diff
@@ -0,0 +1,121 @@
...
+  public void visitNode(Tree tree) {
+    if (tree.is(Tree.Kind.IDENTIFIER)) {
+      check((IdentifierTree) tree);
+    } else {
+      check((MethodInvocationTree) tree);
+    }
+  }
+
+  private void check(IdentifierTree identifier) {
+    if (isPosixPermission(identifier) && !isBeingRemoved(identifier)) {
```
alban-auzeill: I would prefer `is being added` instead of `is being removed`, we could have some FN but not this kind of FP:
```
    Set<PosixFilePermission> posixFilePermissions = Files.getPosixFilePermissions(filePath);
    if (posixFilePermissions.contains(PosixFilePermission.OTHERS_READ)) { // false-positive
      // ...
    }
```

java-checks/src/main/java/org/sonar/java/checks/security/FilePermissionsCheck.java
```diff
@@ -0,0 +1,121 @@
...
+      if (arg0.symbolType().is(JAVA_LANG_STRING) && sensitiveChmodCommand(arg0)) {
+        reportIssue(arg0, ISSUE_MESSAGE);
+      }
+    }
+  }
+
+  private static boolean sensitivePermissionsAsString(ExpressionTree arg0) {
+    return arg0.asConstant(String.class)
+      .filter(chmod -> chmod.length() == 9)
+      .filter(chmod -> !chmod.endsWith("---"))
```
alban-auzeill: Nice and easy way to test: `chmod.endsWith("---")` (you see, from time to time I could agree...:wink:)

java-checks/src/main/java/org/sonar/java/checks/security/FilePermissionsCheck.java
```diff
@@ -112,10 +122,43 @@ private static boolean sensitivePermissionsAsString(ExpressionTree arg0) {
...
+    List<ExpressionTree> initializers = newArrayTree.initializers();
+    if (initializers.size() < 3) {
+      // malformed
+      return;
+    }
+    if (!chmodCommand(initializers.get(0)).isPresent()) {
+      return;
+    }
+    ExpressionTree modeArg = initializers.get(1);
+    if (initializers.size() > 3 && isRecursiveArgument(modeArg)) {
```
alban-auzeill: `--recursive` is just one of the possible options that can be pass to `chmod`, but there are others and multiple options can be passed.
No need to know at which index is the `mode` argument, we can safely test `isSensisitiveChmodMode` on all the `initializers` list (skipping first).

PR #2848, opened 2020-02-27
Fix test

Godin:
There were zero interactions with visitor,
because the mock of SonarComponents was causing NPE.


m-g-sonar:
LGTM! Thanks for fixing this!

PR #2850, opened 2020-02-28
SONARJAVA-3247 S2755 should consider "org.dom4j.io.SAXReader"

quentin-jaquier-sonarsource:


java-frontend/src/main/java/org/sonar/java/se/checks/XxeProcessingCheck.java
```diff
@@ -145,57 +152,45 @@
...
-      .typeDefinition(TypeCriteria.subtypeOf(XML_INPUT_FACTORY))
-      .name(SET_PROPERTY)
-      .parameters(JAVA_LANG_STRING, JAVA_LANG_OBJECT),
-    MethodMatcher.create()
-      .typeDefinition(TypeCriteria.subtypeOf(DOCUMENT_BUILDER_FACTORY))
-      .name(SET_FEATURE)
-      .parameters(JAVA_LANG_STRING, BOOLEAN),
+    setPropertyMatcher(XML_INPUT_FACTORY),
+    setFeatureMatcher(DOCUMENT_BUILDER_FACTORY),
     MethodMatcher.create()
```
m-g-sonar: why not create as well a `setAttributeMatcher`  then? It seems to me that it would simplify the whole list

quentin-jaquier-sonarsource: True, I though it was used only once

PR #2854, opened 2020-03-03
SONARJAVA-3312 FN for S2755: improve detection of XXE attacks

quentin-jaquier-sonarsource:
S2755 implementation failed to detect two cases:

- When "transferring" method call are used with any number of argument.
- When the property is set from a variable.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/se/checks/XxeProcessingCheck.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/checks/XxeProcessingCheck.java
@@ -24,6 +24,7 @@ import com.google.common.collect.ImmutableMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Predicate;
 import javax.annotation.Nullable;
 import org.sonar.check.Rule;
@@ -55,6 +56,7 @@ import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodInvocationTree;
 import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
 
 @Rule(key = "S2755")
 public class XxeProcessingCheck extends SECheck {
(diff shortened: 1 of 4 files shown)


java-frontend/src/main/java/org/sonar/java/se/checks/XxeProcessingCheck.java
```diff
@@ -339,8 +346,13 @@ private ProgramState checkArguments(ProgramState state, Arguments arguments, Xxe
       return state;
     }
 
-    boolean isSettingProperty(ExpressionTree arg0, String propertyName) {
-      return arg0.asConstant(String.class).filter(propertyName::equals).isPresent();
+    boolean isSettingProperty(ProgramState state, ExpressionTree arg0, XxeProperty property) {
+      if (arg0.asConstant(String.class).filter(property.propertyName()::equals).isPresent()) {
```
m-g-sonar: two options here:
* you can replace `property.propertyName()::equals` by `property::isNamed` 
* or you can drop `isNamed(String)`, as I have to feeling that it does the same at the end  (see line below)

java-frontend/src/main/java/org/sonar/java/se/checks/XxeProcessingCheck.java
```diff
@@ -351,6 +363,23 @@ public ProgramState checkPostStatement(CheckerContext context, Tree syntaxNode)
...
   }
 
+  private static ProgramState addNamedConstraint(@Nullable ExpressionTree expressionTree, ProgramState state) {
+    if (expressionTree != null) {
+      SymbolicValue sv = state.peekValue();
+      if (sv != null) {
+        Optional<String> value = expressionTree.asConstant(String.class);
+        if (value.isPresent()) {
+          for (XxeProperty property : PROPERTIES_TO_CHECK) {
+            if (property.isNamed(value.get())) {
```
m-g-sonar: what about `property.propertyName().equals(value.get())` and drop `isNamed`? or maybe we don't need anymore `propertyName()`

java-frontend/src/main/java/org/sonar/java/se/checks/XxeProperty.java
```diff
@@ -81,24 +89,28 @@ static boolean isSetToTrue(@Nullable SymbolicValue sv1, ExpressionTree arg1) {
     private final BiPredicate<SymbolicValue, ExpressionTree> unsecuring;
     private final Constraint unsecured;
 
+    private final Constraint named;
+
     public XxePropertyHolder(String propertyName,
       BiPredicate<SymbolicValue, ExpressionTree> securing, Constraint secured,
-      BiPredicate<SymbolicValue, ExpressionTree> unsecuring, Constraint unsecured) {
+      BiPredicate<SymbolicValue, ExpressionTree> unsecuring, Constraint unsecured, Constraint named) {
```
m-g-sonar: I would put `named` on the first line, next to `propertyName` to group them semantically.

PR #2855, opened 2020-03-03
SONARJAVA-3304 Handle Java 13 preview feature Switch Expressions

m-g-sonar:


java-checks/src/test/files/checks/IgnoredReturnValueCheckJava13.java
```diff
@@ -0,0 +1,11 @@
+class Foo {
+
+  void bar(String s, String cond) {
+    String v = switch (cond) {
+     case "1" -> "aValue";
+     case "2" -> s.toString();
+     case "3" -> s.substring(1);
+     default -> s.replace('a', 'b');
+    };
+  }
```
quentin-jaquier-sonarsource: We could add a test when the switch uses `yield`.

```java
    String v = switch (cond) {
      case "1": yield "aValue";
      case "2": yield s.toString();
      case "3": {
        if (cond.equals("")) {
          yield s.substring(1);
        } else {
          yield s.substring(2);
        }
      }
      default: yield s.replace('a', 'b');
    };
```

It works as expected, but it's not obvious it does at first glance.

java-checks/src/main/java/org/sonar/java/checks/IgnoredReturnValueCheck.java
```diff
@@ -90,6 +90,9 @@
   public void visitNode(Tree tree) {
     ExpressionTree expr = ((ExpressionStatementTree) tree).expression();
     if (expr.is(Tree.Kind.METHOD_INVOCATION)) {
+      if (isSwitchExpressionYieldedExpression(tree)) {
```
quentin-jaquier-sonarsource: Can we do this check conditionally depending on the current java version? To avoid FN such as:
```java
switch("cond") {
  default:
    s.replace("a", "b");
}
```

m-g-sonar:
Cancelled, let's wait for support of Java 14.

PR #2856, opened 2020-03-03
SONARJAVA-3296 FN for security rules using ExpressionUtils.getEnclosingMethod.

quentin-jaquier-sonarsource:
I decided to **not** use TriggeringSecuringHelper (and to remove it), it was suiting well the needs for S2755 old implementation but was not used anymore with the new implementation based on SE.

We decided to keep this code since it could help to solve this ticket, but it turns out that it is not trivial, every rules that could use TriggeringSecuringHelper  has a specific behavior:

- S5301 the triggering tree is a new class, and it has extra-logic in the nested method visitor. 
- S5527 the triggering method is a simple method call (not an assignment). 
- S4426 could use it as it was, but was already implemented without it.

It is possible to adapt this helper to suit the needs of everyone, but it is not more concise nor clearer, I believe 3 different implementations (still sharing parts of the code), is better.


java-checks/src/main/java/org/sonar/java/checks/security/VerifiedServerHostnamesCheck.java
```diff
@@ -179,15 +192,20 @@ private static boolean isNotFalse(ExpressionTree expression) {
...
       .addParameter("boolean");
 
+    MethodBodyApacheVisitor(Symbol variable) {
+      this.variable = variable;
+    }
+
     @Override
     public void visitMethodInvocation(MethodInvocationTree mit) {
-      if (SET_SSL_CHECK_SERVER_ID.matches(mit) && isNotFalse(mit.arguments().get(0))) {
+      if (isInvocationOnVariable(mit, variable) && SET_SSL_CHECK_SERVER_ID.matches(mit) && isNotFalse(mit.arguments().get(0))) {
```
alban-auzeill: This test condition leads to FP if we are bad at resolving the variable symbol. I would prefer to replace `isInvocationOnVariable` by something like `isInvocationOnTheGivenVariableButReturnTrueIfYourAreUnsureAboutOnWhatThisMethodIsInvoked` :wink:

java-frontend/src/main/java/org/sonar/java/model/ExpressionUtils.java
```diff
@@ -114,10 +115,21 @@ public static IdentifierTree extractIdentifier(AssignmentExpressionTree tree) {
...
-    return methodSelect.is(Tree.Kind.MEMBER_SELECT)
-      && extractIdentifierSymbol(((MemberSelectExpressionTree) methodSelect).expression()).filter(s -> s.equals(variable)).isPresent();
+    if (variable != null && methodSelect.is(Tree.Kind.MEMBER_SELECT)) {
+      Optional<Symbol> extractedSymbol = extractIdentifierSymbol(((MemberSelectExpressionTree) methodSelect).expression());
+      if (extractedSymbol.isPresent()) {
+        return extractedSymbol.get().equals(variable);
+      }
+    }
+    return defaultReturn;
   }
```
alban-auzeill: We could reduce the cognitive complexity by 1 using:
```
  public static boolean isInvocationOnVariable(MethodInvocationTree mit, @Nullable Symbol variable, boolean defaultReturn) {
    ExpressionTree methodSelect = mit.methodSelect();
    if (variable != null && methodSelect.is(Tree.Kind.MEMBER_SELECT)) {
      return extractIdentifierSymbol(((MemberSelectExpressionTree) methodSelect).expression())
        .map(variable::equals)
        .orElse(defaultReturn);
    }
    return defaultReturn;
  }
```
or
```
  public static boolean isInvocationOnVariable(MethodInvocationTree mit, @Nullable Symbol variable, boolean defaultReturn) {
    ExpressionTree methodSelect = mit.methodSelect();
    if (variable == null || !methodSelect.is(Tree.Kind.MEMBER_SELECT)) {
      return defaultReturn;
    }
    return extractIdentifierSymbol(((MemberSelectExpressionTree) methodSelect).expression())
      .map(variable::equals)
      .orElse(defaultReturn);
  }
```

quentin-jaquier-sonarsource:
Thank you @alban-auzeill for the interesting feedback, it helped to end up with a nice implementation!

PR #2864, opened 2020-03-06
SONARJAVA-3309 Compile JSP pages to Java and provide AST to checks

saberduck:


saberduck:
I created PR on parent-oss to update version of license plugin https://github.com/SonarSource/parent-oss/pull/39 . Once there is new version of parent, we can remove override of version in the sonar-java

benzonico:
LGTM @m-g-sonar can you also have a look to merge that one  ? thanks.

java-frontend/src/main/java/org/sonar/java/JavaSquid.java
```diff
@@ -91,19 +94,28 @@ public JavaSquid(JavaVersion javaVersion, boolean xFileEnabled,
...
+  private static VisitorsBridge createVisitorBridge(Iterable<? extends JavaCheck> codeVisitors, List<File> classpath,
+                                                    JavaVersion javaVersion, @Nullable SonarComponents sonarComponents,
+                                                    SymbolicExecutionMode symbolicExecutionMode) {
     VisitorsBridge visitorsBridge = new VisitorsBridge(codeVisitors, classpath, sonarComponents, symbolicExecutionMode);
     visitorsBridge.setJavaVersion(javaVersion);
     return visitorsBridge;
   }
 
 
   public void scan(Iterable<InputFile> sourceFiles, Iterable<InputFile> testFiles) {
```
m-g-sonar: if this method is now supposed to be called only by tests, I would annotated it by `@VisibleForTests` and potentially change its visibility.

java-frontend/src/main/java/org/sonar/java/JavaSquid.java
```diff
@@ -91,19 +94,28 @@ public JavaSquid(JavaVersion javaVersion, boolean xFileEnabled,
...
     astScannerForTests.setVisitorBridge(createVisitorBridge(testCodeVisitors, testClasspath, javaVersion, sonarComponents, SymbolicExecutionMode.DISABLED));
 
+    //AstScanner for generated files
+    astScannerForGeneratedFiles = new JavaAstScanner(sonarComponents);
+    astScannerForGeneratedFiles.setVisitorBridge(createVisitorBridge(generatedCodeVisitors, classpath, javaVersion, sonarComponents, SymbolicExecutionMode.DISABLED));
   }
 
-  private static VisitorsBridge createVisitorBridge(
-    Iterable<JavaCheck> codeVisitors, List<File> classpath, JavaVersion javaVersion, @Nullable SonarComponents sonarComponents, SymbolicExecutionMode symbolicExecutionMode) {
+  private static VisitorsBridge createVisitorBridge(Iterable<? extends JavaCheck> codeVisitors, List<File> classpath,
```
m-g-sonar: I don't understand this change, It seems to me that `generatedCodeVisitors` are `JavaCheck`, the change of method signature should not be required.

java-frontend/src/main/java/org/sonar/java/model/GeneratedFile.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  @Override
+  public TextRange selectLine(int line) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public Charset charset() {
+    return Charset.defaultCharset();
```
m-g-sonar: How can you be sure of that? If you transpile these files, while not using directly a predefined charset such as UTF-8?

java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
```diff
@@ -85,11 +86,13 @@ public VisitorsBridge(JavaFileScanner visitor) {
   }
 
   @VisibleForTesting
-  public VisitorsBridge(Iterable visitors, List<File> projectClasspath, @Nullable SonarComponents sonarComponents) {
+  public VisitorsBridge(Iterable<? extends JavaCheck> visitors, List<File> projectClasspath,
```
m-g-sonar: I don't see why this type needs to be specified. it's seems unrelated to your change,

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -193,6 +195,13 @@ public void registerTestCheckClasses(String repositoryKey, Iterable<Class<? exte
     return visitors;
   }
 
+  public List<JavaCheck> generatedCodeVisitors() {
+    return allChecks.stream()
+      .flatMap(c -> c.all().stream())
```
m-g-sonar: c? explicit name maybe?

m-g-sonar: c? please use a more explicit variable name. as `c.all()` is really not helping.

java-frontend/src/main/java/org/sonar/plugins/java/api/GeneratedCodeVisitor.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.sonar.plugins.java.api;
+
+import com.google.common.annotations.Beta;
+
+/**
+ * Marker interface for code visitors which are invoked on generated code
+ */
+@Beta
+public interface GeneratedCodeVisitor extends JavaCheck {
```
m-g-sonar: I don't see any mechanism which would prevent people writing custom rules to use this as well, expecting their code to be executed on "generated code". (code annotated with `@Generated`?).
While I understand that we don't necessarily want to expose everything, I would appreciate to have it a bit more explicit. note any kind of generated code is going to be analyzed by such rule.

java-jsp/src/main/java/org/sonar/java/jsp/Jasper.java
```diff
@@ -0,0 +1,104 @@
...
+        return generatedFiles;
+      }
+    } catch (Exception e) {
+      LOG.warn("Failed to transpile JSP files.", e);
+      return Collections.emptyList();
+    }
+  }
+
+  private static List<Path> jspFiles(FileSystem fs) {
+    Iterable<InputFile> inputFiles = fs.inputFiles(fs.predicates().hasLanguage("jsp"));
```
m-g-sonar: What if there is no JSP language registered? is it always the case?

java-frontend/src/main/java/org/sonar/plugins/java/api/GeneratedCodeVisitor.java
saberduck: I am not sure I understand the comment, do you just want to improve javadoc?

java-frontend/src/main/java/org/sonar/java/JavaSquid.java
saberduck: I removed this overload, and instead added `emptyList()` parameter to the callers in tests.

java-frontend/src/main/java/org/sonar/java/JavaSquid.java
saberduck: removed, this was leftover from some refactoring

java-frontend/src/main/java/org/sonar/java/model/GeneratedFile.java
saberduck: ok, I found the option how to explicitly set encoding for Jasper, so we can use UTF-8 explicitly.

java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
saberduck: it's not strictly related to this PR, but I wanted to remove usage of raw type to benefit from compiler checking the correctness of types.

java-jsp/src/main/java/org/sonar/java/jsp/Jasper.java
saberduck: html plugin is part of default distribution, so I believe it is safe to assume that it is available. If it's not available we will not find any files and nothing is executed.

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -193,6 +195,13 @@ public void registerTestCheckClasses(String repositoryKey, Iterable<Class<? exte
     return visitors;
   }
 
+  public List<JavaCheck> generatedCodeVisitors() {
```
m-g-sonar: should be called `jspCodeVisitors()`

java-frontend/src/main/java/org/sonar/java/model/GeneratedFile.java
```diff
@@ -0,0 +1,155 @@
...
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import javax.annotation.CheckForNull;
+import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.batch.fs.TextPointer;
+import org.sonar.api.batch.fs.TextRange;
+
+public class GeneratedFile implements InputFile {
```
m-g-sonar: Should be called `JspFile`

PR #2873, opened 2020-03-11
Fix false positives for Lombok's field modifier annotations

goober:
This change will ignore reported ExceptionsShouldBeImmutable issues on classes that extends Throwable and also is annoted with Lombok's @Value annotation.

The `@Value` annotation will add `final` on the generated output, hence S1165 could safely be ignored for those classes.

Also reported at the [support forum](https://community.sonarsource.com/t/squid-s1165-exception-classes-should-be-immutable-when-annotating-an-exception-with-loboks-value/21457?u=goober)

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
+++ b/java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
@@ -50,6 +50,7 @@ import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.IdentifierTree;
 import org.sonar.plugins.java.api.tree.ImportTree;
 import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
 
 public class LombokFilter extends BaseTreeVisitorIssueFilter {
 
@@ -69,6 +70,8 @@ public class LombokFilter extends BaseTreeVisitorIssueFilter {
     ExceptionsShouldBeImmutableCheck.class);
 
   private static final String LOMBOK_VAL = "lombok.val";
+  private static final String LOMBOK_VALUE = "lombok.Value";
+  private static final String LOMBOK_FIELD_DEFAULTS = "lombok.experimental.FieldDefaults";
 
   private static final List<String> GENERATE_UNUSED_FIELD_RELATED_METHODS = ImmutableList.<String>builder()
     .add("lombok.Getter")
@@ -89,11 +92,20 @@ public class LombokFilter extends BaseTreeVisitorIssueFilter {
   private static final List<String> GENERATE_EQUALS = ImmutableList.<String>builder()
     .add("lombok.EqualsAndHashCode")
     .add("lombok.Data")
-    .add("lombok.Value")
+    .add(LOMBOK_VALUE)
     .build();
 
   private static final List<String> UTILITY_CLASS = ImmutableList.<String>builder()
     .add("lombok.experimental.UtilityClass")
+    .build();
+
+  private static final List<String> NON_FINAL = ImmutableList.<String>builder()
+    .add("lombok.experimental.NonFinal")
+    .build();
+
+  private static final List<String> GENERATES_MODIFIERS = ImmutableList.<String>builder()
+    .add(LOMBOK_VALUE)
+    .add(LOMBOK_FIELD_DEFAULTS)
     .build();
 
   @Override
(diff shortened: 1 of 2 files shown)


quentin-jaquier-sonarsource:
@goober I'm not really familiar with lombok, do you think that other annotations could prevent this rule to behave correctly? What about `@FieldDefaults` for example?

It could be a good idea to take care of all them at the same time.

goober:
@quentin-jaquier-sonarsource That is true, I have never used `@FieldDefaults` so I missed that. But I can see if I can cover those cases as well.

goober:
@quentin-jaquier-sonarsource I have updated the PR to take `@FieldDefaults` into account.
In addition I have covered some additional cases where `@NonFinal` and `@PackagePrivate` are used on class variable level.

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -117,11 +124,16 @@ public void visitClass(ClassTree tree) {
     excludeLinesIfTrue(generatesEquals, tree, EqualsNotOverriddenInSubclassCheck.class, EqualsNotOverridenWithCompareToCheck.class);
     excludeLinesIfTrue(generatesPrivateConstructor(tree), tree, UtilityClassWithPublicConstructorCheck.class);
     excludeLinesIfTrue(usesAnnotation(tree, UTILITY_CLASS), tree, BadFieldNameCheck.class, ConstantsShouldBeStaticFinalCheck.class);
-    excludeLinesIfTrue(usesAnnotation(tree, Collections.singletonList("lombok.Value")), tree, FieldModifierCheck.class);
-
```
quentin-jaquier-sonarsource: If the rule targets only S1165, we should not change the behavior of other rules.
This change makes sense though, you could keep it and rename the PR to better reflect this, state that you are targetting Lombok's annotations modifying fields modifier for example.

java-checks/src/test/files/filters/LombokFilter.java
```diff
@@ -345,4 +345,54 @@ public void test() {
...
+  class A {
+    int id; // WithIssue
+  }
+  @lombok.experimental.FieldDefaults(level = lombok.AccessLevel.PRIVATE)
+  class B {
+    int id; //NoIssue
+    @lombok.experimental.PackagePrivate String name; // WithIssue
+  }
+
+  @lombok.experimental.FieldDefaults(makeFinal=true, level = lombok.AccessLevel.PRIVATE)
```
quentin-jaquier-sonarsource: We could add a test with `makeFinal=false`.

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -117,11 +124,16 @@ public void visitClass(ClassTree tree) {
...
     super.visitClass(tree);
   }
 
+  @Override
+  public void visitVariable(VariableTree tree) {
```
quentin-jaquier-sonarsource: We will go through this code for every variable declaration.
Why not going for a top-down approach: starting from a Class, and going down only when you find a relevant annotation?

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -157,6 +176,46 @@ private static boolean generatesPrivateAccess(List<SymbolMetadata.AnnotationValu
...
+    }
+    return Optional.ofNullable(tree.symbol().metadata().valuesForAnnotation("lombok.experimental.FieldDefaults"))
+      .map(List::stream)
+      .map(s -> s.filter(Objects::nonNull)
+        .anyMatch(av -> "level".equals(av.name()) && "PRIVATE".equals(getAccessLevelValue(av.value())))
+      ).orElse(false);
+  }
+
+  private static boolean generatesPrivateAccess(VariableTree tree) {
+    if (tree.parent() instanceof ClassTree && generatesPrivateAccess((ClassTree) tree.parent())) {
```
quentin-jaquier-sonarsource: Even if it would make sense that `tree.parent()` always return the same value, I believe it's safer to consider that it does not (to avoid tricky bugs).

quentin-jaquier-sonarsource:
Hello @goober , did you have time to look at it?
No stress or obligation, of course, it's just that it could be nice if we manage to merge it for version 6.3!

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
goober: Good point. I have updated the code to only scan variables for classes with relevant annotations

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -157,6 +176,46 @@ private static boolean generatesPrivateAccess(List<SymbolMetadata.AnnotationValu
     return values.stream().anyMatch(av -> "access".equals(av.name()) && "PRIVATE".equals(getAccessLevelValue(av.value())));
   }
 
+  private static boolean generatesPrivateAccess(ClassTree tree) {
+    if (usesAnnotation(tree, Collections.singletonList("lombok.Value"))) {
```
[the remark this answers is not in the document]
goober: I am but I do not have it as e pre-push hook so I missed that one. Now it is updated.

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
goober: Fixed

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
goober: PR title updated

goober:
@quentin-jaquier-sonarsource I have updated the PR to reflect your comments. I hope it is not too late

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -157,6 +188,48 @@ private static boolean generatesPrivateAccess(List<SymbolMetadata.AnnotationValu
...
+    }
+    return Optional.ofNullable(tree.symbol().metadata().valuesForAnnotation(LOMBOK_FIELD_DEFAULTS))
+      .map(List::stream)
+      .map(s -> s.filter(Objects::nonNull)
+        .anyMatch(av -> "level".equals(av.name()) && "PRIVATE".equals(getAccessLevelValue(av.value())))
+      ).orElse(false);
+  }
+
+  private static boolean generatesPrivateAccess(VariableTree tree) {
+    Tree parent = tree.parent();
```
quentin-jaquier-sonarsource: I'm still not convinced about the logic here. For all field, you will look at the value of the annotation again, you should do it only once when you visit the class!

goober:
Closing in favor of #2875

PR #2875, opened 2020-03-12
SONARJAVA-3320 Fix false positives for Lombok's field modifier annotations

quentin-jaquier-sonarsource:
Duplicates #2873 to trigger all checks.


goober:
@quentin-jaquier-sonarsource it looks good and cover the same cases as in my PR so I think we could go ahead with this PR instead!

quentin-jaquier-sonarsource:
Merged!

Thanks again for the investigation @goober !

PR #2877, opened 2020-03-17
SONARJAVA-3322 S2441: handle unknown types

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableObjectInSessionCheck.java
```diff
@@ -44,6 +44,9 @@
   protected void onMethodInvocationFound(MethodInvocationTree mit) {
     ExpressionTree argument = mit.arguments().get(1);
     Type type = argument.symbolType();
+    if (type.isUnknown()) {
+      return;
+    }
     if (ExpressionsHelper.isNotSerializable(argument)) {
```
quentin-jaquier-sonarsource: So it means that an argument with unknown type is considered as not serializable?
Should we consider making the unknown check inside `isNotSerializable` and provide the same behavior for all rules using it?

I'm saying this since the other rule (S2118) using `isNotSerializable` is also reporting an arguable issue with a similar message if we run it against an unknown type.

> Make the "!Unknown!" class "Serializable" or don't write it.

m-g-sonar: Argh, I didn't realized that we were calling ExpresisonHelper to detect this. Good catch, I'm going to move the fix in the Helper class.

PR #2881, opened 2020-03-19
SONARJAVA-2941 RSPEC-3740 Generic types shouldn't be used raw

m-g-sonar:


java-checks-test-sources/src/main/java/checks/RawTypeCheck.java
```diff
@@ -0,0 +1,30 @@
...
+
+  GenericClass bar() { return null; } // Noncompliant
+
+  static class GenericClass<T> { }
+  interface GenericInterface<T> { }
+
+  static class InnerClass1 extends GenericClass { } // Noncompliant [[sc=36;ec=48]] {{Provide the parametrized type for this generic.}}
+  static class InnerClass2 implements GenericInterface { } // Noncompliant [[sc=39;ec=55]] {{Provide the parametrized type for this generic.}}
+  enum InnerEnum implements GenericInterface { } // Noncompliant [[sc=29;ec=45]]
+}
```
alban-auzeill: It would be nice to add a test with nested generics. E.g. of FN: `Metric` is a raw generic but there's a false-negative in `implements Function<Metric, String>` here:
https://github.com/SonarSource/ruling_java/blob/7f4b47c9f013c5997f138ddc6e4f916cc7763476/sonarqube-6.5/server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/step/LoadMeasureComputersStep.java#L153

It would be great to add test cases with method parameters. And to support exclusions of issues on return type and parameters of overridden methods (`methodSymbol.overriddenSymbol() != null`, I'm not sure we need to check `@Override` annotations). Like this false-positive:
https://github.com/SonarSource/ruling_java/blob/7f4b47c9f013c5997f138ddc6e4f916cc7763476/sonarqube-6.5/server/sonar-server/src/main/java/org/sonar/server/issue/IssuesFinderSort.java#L88

PR #2883, opened 2020-03-23
SONARJAVA-3314 RSPEC-5378 Deprecated code marked for removal should be removed

m-g-sonar:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5738_java.json
```diff
@@ -0,0 +1,23 @@
+{
+  "title": "Deprecated code marked for removal should be removed",
```
quentin-jaquier-sonarsource: The name can be misleading, it is similar to [RSPEC-1133](https://jira.sonarsource.com/browse/RSPEC-1133), despite the fact that it matches the behavior of [1874](https://jira.sonarsource.com/browse/1874).
Maybe rename it to something closer to 1874?
`"@Deprecated" code marked for removal should not be used`

PR #2885, opened 2020-03-23
SONARJAVA-3328 Fix broken Jasper test on macOS

saberduck:


java-jsp/src/test/java/org/sonar/java/jsp/JasperTest.java
```diff
@@ -56,10 +57,18 @@
 
   @BeforeEach
   void setUp() throws Exception {
+    // on MacOs tmp is symbolic link which doesn't work well with Jasper, so we create tmp in 'target/tmp'
```
Godin: @saberduck official spelling - `macOS` ;)

PR #2899, opened 2020-04-01
SONARJAVA-3325 [JSP] Remove dependency on Ant from Jasper

saberduck:
Current size: 18709446
6.2: 17038520
i.e. we are at +1.7MB for JSP support

### DIFF
--- a/java-jsp/src/main/java/org/sonar/java/jsp/Jasper.java
+++ b/java-jsp/src/main/java/org/sonar/java/jsp/Jasper.java
@@ -19,7 +19,6 @@
  */
 package org.sonar.java.jsp;
 
-import com.google.common.annotations.VisibleForTesting;
 import java.io.File;
 import java.io.IOException;
 import java.net.MalformedURLException;
@@ -27,11 +26,14 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
@@ -42,6 +44,7 @@ import org.apache.jasper.compiler.Compiler;
 import org.apache.jasper.compiler.JspRuntimeContext;
 import org.apache.jasper.runtime.JspFactoryImpl;
 import org.apache.jasper.servlet.JspCServletContext;
+import org.sonar.api.batch.fs.FilePredicates;
 import org.sonar.api.batch.fs.FileSystem;
 import org.sonar.api.batch.fs.InputFile;
 import org.sonar.api.batch.sensor.SensorContext;
@@ -62,8 +65,10 @@ public class Jasper {
     if (jspFiles.isEmpty()) {
       return Collections.emptyList();
     }
+    Path uriRoot = findWebInfParentDirectory(sensorContext.fileSystem())
+      .orElse(sensorContext.fileSystem().baseDir().getAbsoluteFile().toPath());
+    LOG.debug("Context root set to {}", uriRoot);
     Path outputDir = outputDir(sensorContext);
-    File baseDir = sensorContext.fileSystem().baseDir();
     // Jasper internally calls Thread#getContextClassLoader to instantiate some classes. ContextClassLoader is set by scanner
     // and doesn't contain plugin jar, so we need to configure ContextClassLoader with the class loader of the plugin to be able
     // to run Jasper. Original classloader is restored in finally.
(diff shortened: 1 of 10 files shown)


java-jsp/src/main/java/org/sonar/java/jsp/Jasper.java
```diff
@@ -33,38 +35,71 @@
...
+      JspFactory.setDefaultFactory(new JspFactoryImpl());
+      JspCServletContext servletContext = new ServletContext(toUrl(baseDir), classLoader);
+      JasperOptions options = new JasperOptions(servletContext, outputDir);
+      JspRuntimeContext runtimeContext = new JspRuntimeContext(servletContext, options);
+
+      boolean errorTranspiling = false;
+      for (Path file : jspFiles) {
+        LOG.debug("Transpiling JSP: {}", file);
+        try {
+          String relativePath = file.toAbsolutePath().toString().substring(baseDir.getAbsolutePath().length());
```
benzonico, line 83: I think this operation you should rely on https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#relativize(java.nio.file.Path) rather than string.

java-jsp/src/main/java/org/sonar/java/jsp/Jasper.java
```diff
@@ -74,22 +109,18 @@
...
     try {
-      LOG.debug("Running JspC with args {} ", args);
-      JspC jspc = new JspC();
-      jspc.setArgs(args.toArray(new String[0]));
-      jspc.execute();
-    } catch (Exception e) {
-      throw new AnalysisException("Error while transpiling JSP pages.", e);
+      return f.toURI().toURL();
+    } catch (MalformedURLException e) {
+      // can't happen
```
benzonico, line 122: could deserve just a (tiny) why of this can't happen.

java-jsp/src/main/java/org/sonar/java/jsp/Jasper.java
```diff
@@ -33,38 +35,71 @@
...
-import static java.util.Arrays.asList;
-
 @ScannerSide
 public class Jasper {
 
   private static final Logger LOG = Loggers.get(Jasper.class);
 
-  public Collection<GeneratedFile> generateFiles(SensorContext context, List<File> javaClasspath) {
-    List<Path> jspFiles = jspFiles(context.fileSystem());
+  public Collection<GeneratedFile> generateFiles(SensorContext sensorContext, List<File> javaClasspath) {
```
benzonico, line 59: cosmetic but I tend to think that this method could be made a bit more readable by extracting the core of the loop into a compileJspFile(... ) method

java-jsp/src/main/java/org/sonar/java/jsp/JasperOptions.java
```diff
@@ -0,0 +1,238 @@
...
+import org.apache.jasper.compiler.TldCache;
+import org.apache.jasper.servlet.JspCServletContext;
+import org.apache.jasper.servlet.TldScanner;
+import org.sonar.java.AnalysisException;
+import org.xml.sax.SAXException;
+
+/**
+ * This class is replacing {@link org.apache.jasper.JspC} to avoid dependency on Ant. Most of the initialization is
+ * copied from JspC class which can be used as reference. Many of the methods are not actually used and are needed only
+ * to implement the interface.
```
benzonico, line 42: I am wondering if in order to make that statement clearer in the code, the unused method should be implemented with a `throw new UnsupportedOperationException`

java-jsp/src/test/java/org/sonar/java/jsp/JasperTest.java
```diff
@@ -133,10 +137,24 @@ void test_source_map() throws Exception {
     assertThat(generatedFile.sourceMap()).isNotNull();
   }
 
+  @Test
+  void test_failure_invalid_classpath() throws Exception {
```
benzonico, line 141: I am unsure about what this test actually verifies.

saberduck: renamed to make it a bit clear, also changed the method which is mocked so it's more realistic

java-jsp/src/main/java/org/sonar/java/jsp/JasperOptions.java
saberduck: it's quite complicated to figure out what is really used, keeping the default seems safer

java-jsp/src/main/java/org/sonar/java/jsp/Jasper.java
```diff
@@ -21,50 +21,118 @@
...
+    compiler.compile(false, true);
+  }
+
+  JasperOptions getJasperOptions(Path outputDir, JspCServletContext servletContext) {
+    return new JasperOptions(servletContext, outputDir);
+  }
+
+  private static Optional<Path> findWebInfParentDirectory(FileSystem fs) {
+    FilePredicates predicates = fs.predicates();
+    ArrayList<InputFile> inputFiles = new ArrayList<>();
```
benzonico, line 122: Why not the interface ?

PR #2900, opened 2020-04-03
unable to resolve java.lang.string reproducer

m-g-sonar:


m-g-sonar:
Won't fix. Problem of split package.

PR #2903, opened 2020-04-06
SONARJAVA-3340 [JSP] Add plugin jar to classpath for JSP checks

saberduck:


java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -152,6 +153,22 @@ public NewHighlighting highlightableFor(InputFile inputFile) {
     return javaTestClasspath.getElements();
   }
 
+  public List<File> getJspClasspath() {
+    List<File> jspClasspath = new ArrayList<>();
+    jspClasspath.add(findPluginJar());
+    jspClasspath.addAll(javaClasspath.getElements());
```
benzonico, line 159: this would probably be more robust to call for getJavaClasspath rather than relying on order of callers for javaclasspath to be initialized.

benzonico:
I think we can remove org.apache.tomcat.embed:tomcat-embed-el filter from the plugin packaging.

saberduck:
> I think we can remove org.apache.tomcat.embed:tomcat-embed-el filter from the plugin packaging.

we can't, `ExpressionFactoryImpl` is accessed via reflection, so it's not visible statically when packaging. I added a comment to explain that

PR #2906, opened 2020-04-08
Remove analysis warnings wrapper

m-g-sonar:


m-g-sonar:
@quentin-jaquier-sonarsource it's fishy that we can simpliy drop this. I might need to invesitgate a bit more to check that we are not loosing something.... but to me it's dead code already.

m-g-sonar:
Closing the PR as it can not be dropped. SonarLint does not provides an alternative to AnalysisWarnings.

PR #2914, opened 2020-04-22
SONARJAVA-3356 Rule S5779: Assertion methods should not be used within the try block of a try-catch catching an Error

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/AssertionInTryCatchCheck.java
```diff
@@ -0,0 +1,95 @@
...
+            || symbolType.is("java.lang.Throwable"));
+      }).findFirst();
+  }
+
+  private class TryBodyVisitor extends BaseTreeVisitor {
+    private final List<JavaFileScannerContext.Location> secondaryLocation;
+
+    public TryBodyVisitor(VariableTree catchTree) {
+      this.secondaryLocation = Collections.singletonList(new JavaFileScannerContext.Location(
+        "This parameter will catch the AssertionError thrown by fail()",
```
alban-auzeill, line 78: `thrown by fail()` it's not always `fail()`

PR #2916, opened 2020-04-23
SONARJAVA-3351 Update S2391 to cover JUnit4/5

m-g-sonar:
Please do not squash commits.


m-g-sonar:
I updated the PR to take into account all the JUnit5 annotations. Regarding cases where we have mixed JUnit4/5 test but code annotated with JUnit4 `@Before ` or `@After`, I believe it's simpler to simply ask to change the annotations to the JUnit5 version instead, and so I slightly changed the way it is reported. In other cases, I would prefer to rely on another rule.

alban-auzeill:
The `Fix from review` commit takes into account `@Before` and `@After` mixed with JUnit 5 tests, but only if method names match exactly "setUp" and "tearDown". Nowadays, nobody remember those names, so it would be better to support any method name like in the following case:
```
public class FooTest {
  @org.junit.Before
  public void before() { }
  @org.junit.jupiter.api.Test
  void test() { }
}
```
WDYT?

m-g-sonar:
Fair point @alban-auzeill , new fix incoming!

PR #2918, opened 2020-04-23
SONARJAVA-3350 S2701: support null literal

sebastian-hungerecker-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/BooleanOrNullLiteralInAssertionsCheck.java
```diff
@@ -94,14 +100,26 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
     }
   }
 
-  private void checkEqualityAsserts(MethodInvocationTree mit, boolean flipped) {
-    List<ExpressionTree> literals = mit.arguments().stream()
+  private static List<ExpressionTree> findLiterals(List<ExpressionTree> expressions) {
+    return expressions.stream()
       .filter(BooleanOrNullLiteralInAssertionsCheck::isBoolOrNullLiteral)
+      .collect(Collectors.toCollection(ArrayList::new));
```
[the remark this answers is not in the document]
sebastian-hungerecker-sonarsource: The list returned by `toList` isn't guaranteed to be mutable, but I want a mutable list, so I can add an additional literal to it on line 197. I'll add a comment explaining that.

its/ruling/src/test/resources/commons-beanutils/java-S2701.json
```diff
@@ -1,8 +1,12 @@
 {
+'commons-beanutils:commons-beanutils:src/test/java/org/apache/commons/beanutils2/BeanToPropertyValueTransformerTestCase.java':[
+186,
+244,
+],
 'commons-beanutils:commons-beanutils:src/test/java/org/apache/commons/beanutils2/BeanUtilsTestCase.java':[
-210,
+209,
```
quentin-jaquier-sonarsource, line 7: The issue location slightly moved (and potentially the message aswell), from the literal to the method name, with the literal being the secondary location.
@m-g-sonar I'm afraid that this will cause side effects, re-open issues for example, what do you think?

m-g-sonar: I would consider this as an improvement of the rule anyway, so I think it's perfectly fine.

PR #2920, opened 2020-04-24
Sonarjava 3362 junit5 test class method visibility

johann-beleites-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/TestClassAndMethodVisibilityCheck.java
```diff
@@ -0,0 +1,72 @@
...
+import org.sonar.plugins.java.api.tree.Modifier;
+import org.sonar.plugins.java.api.tree.ModifiersTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S5786")
+public class TestClassAndMethodVisibilityCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Arrays.asList(Tree.Kind.METHOD, Tree.Kind.CLASS);
```
alban-auzeill, line 38: The rule target properly METHOD, but for CLASS it's more complicated. A JUnit 5 test class is a class that contains at least one test method. I would probably register only on CLASS, then check all members for test methods. So I could at the same time check if one test method exists and check if each test method is compliant.

java-checks/src/main/java/org/sonar/java/checks/TestClassAndMethodVisibilityCheck.java
```diff
@@ -0,0 +1,72 @@
...
+      MethodTree methodTree = (MethodTree) tree;
+      symbol = methodTree.symbol();
+      modifiers = methodTree.modifiers();
+    } else {
+      ClassTree classTree = (ClassTree) tree;
+      symbol = classTree.symbol();
+      modifiers = classTree.modifiers();
+    }
+
+    if (symbol.metadata().isAnnotatedWith("org.junit.jupiter.api.Test") && !symbol.isPackageVisibility()) {
```
alban-auzeill, line 59: There's not only `org.junit.jupiter.api.Test` that indicate it's a JUnit 5 test method, you can use `UnitTestUtils.hasJUnit5TestAnnotation(...)`

I recently use it in `JunitNestedAnnotationCheck`:
```
  private static boolean hasJUnit5TestMethods(ClassTree classTree) {
    return classTree.members().stream()
      .filter(member -> member.is(Tree.Kind.METHOD))
      .map(MethodTree.class::cast)
      .anyMatch(UnitTestUtils::hasJUnit5TestAnnotation);
  }
```

java-checks/src/main/java/org/sonar/java/checks/TestClassAndMethodVisibilityCheck.java
```diff
@@ -0,0 +1,72 @@
...
+      Tree questionableNode = modifiers.modifiers().parallelStream()
```
alban-auzeill, line 60: why `parallelStream()` and not `stream()`?

java-checks/src/main/java/org/sonar/java/checks/TestClassAndMethodVisibilityCheck.java
```diff
@@ -0,0 +1,72 @@
...
+        .filter(keywordTree -> {
+          Modifier modifier = keywordTree.modifier();
+          return modifier.equals(Modifier.PUBLIC) || modifier.equals(Modifier.PRIVATE) || modifier.equals(Modifier.PROTECTED);
+        })
+        .map(modifierKeywordTree -> (Tree) modifierKeywordTree)
+        .findFirst()
+        .orElse(tree);
+
+      reportIssue(questionableNode, "Remove this access modifier");
```
alban-auzeill, line 69: By adding the following constant:
```
private static final List<Modifier> NON_COMPLIANT_MODIFIERS = Arrays.asList(Modifier.PUBLIC, Modifier.PRIVATE, Modifier.PROTECTED);
```
we could simplify the implementation like this:
```
      modifiers.modifiers().stream()
        .filter(modifierKeyword -> NON_COMPLIANT_MODIFIERS.contains(modifierKeyword.modifier()))
        .map(Tree.class::cast)
        .findFirst()
        .ifPresent(modifierKeyword -> reportIssue(modifierKeyword, "Remove this access modifier"));
```

java-checks-test-sources/src/main/java/checks/TestClassAndMethodVisibilityCheckTest.java
```diff
@@ -0,0 +1,22 @@
...
+
+  @Test
+  protected void testProtected() {} // Noncompliant [[sc=3;ec=12]] {{Remove this access modifier}}
+
+  @Test
+  private void testPrivate() {} // Noncompliant [[sc=3;ec=10]] {{Remove this access modifier}}
+
+  @Test
+  void testDefault() {} // compliant
+
```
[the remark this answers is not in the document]
alban-auzeill: ẁe should add a test with other modifiers (e.g. `final` or `synchronized`) to show the rule only target `public`, `private` and `protected`

java-checks-test-sources/src/main/java/checks/TestClassAndMethodVisibilityCheckTest.java
```diff
@@ -0,0 +1,22 @@
...
+  public void noTestHere() {} // compliant - this rule only applies to test methods
+
+  void noTestHereEither() {} // compliant
```
alban-auzeill, line 21: we should add some test cases about CLASS like:
```
  public static class WithoutTest { // Compliant

  }

  public static class PublicWithOneTest { // Noncompliant
    @Test
    void test() {}
  }

  static class WithOneTest { // Compliant
    @Test
    void test() {}
  }
```

java-checks-test-sources/src/main/java/checks/TestClassAndMethodVisibilityCheckTest.java
```diff
@@ -0,0 +1,22 @@
+package checks;
+
+import org.junit.jupiter.api.Test;
+
+class TestClassAndMethodVisibilityCheckTest {
+
+  @Test
+  public void testPublic() {} // Noncompliant [[sc=3;ec=9]] {{Remove this access modifier}}
```
alban-auzeill, line 8: Usually, when the message is constant, we test it only once.
So, here it's ok, you can use `// Noncompliant [[sc=3;ec=9]] {{Remove this access modifier}}`
And elsewhere it's enough to use `// Noncompliant`, without the location and the message

java-checks/src/main/java/org/sonar/java/checks/TestClassAndMethodVisibilityCheck.java
johann-beleites-sonarsource: Ah of course, I'm not filtering properly, my bad

PR #2922, opened 2020-04-27
SONARJAVA-3359 Rule S5783: Tested checked exception should be expected from one single call

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
```diff
@@ -0,0 +1,167 @@
...
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
+      return;
+    }
+
+    if (tree.is(Tree.Kind.METHOD_INVOCATION)) {
+      MethodInvocationTree mit = (MethodInvocationTree) tree;
+      if (mit.arguments().size() >= 2 && ASSERT_THROWS_MATCHER.matches(mit)) {
+        Optional<IdentifierTree> expectedException = getExpectedException(mit.arguments().get(0));
+        ExpressionTree executable = mit.arguments().get(1);
```
johann-beleites-sonarsource, line 76: JUnit4 also supports assertions with messages - unfortunately as first parameter, different to JUnit5 (e.g. https://junit.org/junit4/javadoc/latest/src-html/org/junit/Assert.html#line.998). By relying on the exception type to be in position 0 and the executable in position 1, I'd argue it could come to FNs when a developer uses that JUnit4 assertion. Do we want to support that case as well?

quentin-jaquier-sonarsource: Good point!
JUnit 5 also has a message, but as the last parameter, everything works as expected.
And indeed, it is first in JUnit 4, I believe we should also support this case.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5783_java.json
```diff
@@ -0,0 +1,13 @@
+{
+  "title": "Tested checked exception should be expected from one single call",
```
m-g-sonar, line 2: I think the title is still cumbersome.
What about "Only one method invocation is expected when testing checked exceptions"

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
```diff
@@ -0,0 +1,180 @@
...
+      if (isTryCatchIdiom(tryStatementTree)) {
+        tryStatementTree.catches().forEach(c ->
+          reportMultipleCallThrowingExceptionInTree(c.parameter().type().symbolType(), tryStatementTree.block(), c.parameter().type())
+        );
+      }
+    }
+  }
+
+  private void processAssertThrowsArguments(ExpressionTree expectedType, ExpressionTree executable) {
+    Optional<IdentifierTree> expectedException = getExpectedException(expectedType);
```
m-g-sonar, line 100: i would rewrite this part slightly differently:
```
if (!executable.is(Tree.Kind.LAMBDA_EXPRESSION)) {
  return;
}
getExpectedException(expectedType)
  .ifPresent(exception -> reportMultipleCallThrowingExceptionInTree(exception.symbolType(), ((LambdaExpressionTree) executable).body(), exception));
```

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
```diff
@@ -0,0 +1,180 @@
...
+        ExpressionTree expressionTree = ((ExpressionStatementTree) lastElement).expression();
+        if (expressionTree.is(Tree.Kind.METHOD_INVOCATION)) {
+          return JUNIT_FAIL_MATCHER.matches((MethodInvocationTree) expressionTree);
+        }
+      }
+    }
+    return false;
+  }
+
+  private void reportMultipleCallThrowingExceptionInTree(Type expectedException, Tree tree, Tree reportLocation) {
```
m-g-sonar, line 132: I don't see the point of the first parameter. It seems to me that `reportLocation` is a `TypeTree`, so you could just rely on it and extract the type within this method!
```
private void reportMultipleCallThrowingExceptionInTree(Tree tree, TypeTree exception) {
```
I would also rename "tree" by "treeToVisit" to make it clear

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
```diff
@@ -0,0 +1,180 @@
...
+    MethodInvocationThrowing visitor = new MethodInvocationThrowing(expectedException); //m.symbolType().isSubtypeOf(expectedType)
```
m-g-sonar, line 133: commented code?

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
```diff
@@ -0,0 +1,180 @@
...
+  private static class MethodInvocationThrowing extends BaseTreeVisitor {
+    List<MethodInvocationTree> methodInvocationTrees = new ArrayList<>();
+    private final Type expectedException;
+
+    MethodInvocationThrowing(Type expectedException) {
+      this.expectedException = expectedException;
+    }
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree mit) {
```
m-g-sonar, line 160: what about visiting newClasses as well? you can perfectly call constructors as well, which can define checkedExceptions:
```
new MyObjectThrowingException(methodThrowingException());
```

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
```diff
@@ -0,0 +1,180 @@
...
+      MemberSelectExpressionTree memberSelect = ((MemberSelectExpressionTree) expectedType);
+      ExpressionTree expression = memberSelect.expression();
+      if ("class".equals(memberSelect.identifier().name()) && expression.is(Tree.Kind.IDENTIFIER)) {
+        return Optional.of((IdentifierTree) expression);
+      }
+    }
+    return Optional.empty();
+  }
+
+  private static boolean isTryCatchIdiom(TryStatementTree tree) {
```
m-g-sonar, line 118: `isTryCatchFail` idiom is rather vague

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
```diff
@@ -0,0 +1,180 @@
...
+      if (JUNIT4_ASSERT_THROWS_WITH_MESSAGE.matches(mit)) {
+        processAssertThrowsArguments(arguments.get(1), arguments.get(2));
+      } else if (arguments.size() >= 2 && ALL_ASSERT_THROWS_MATCHER.matches(mit)) {
+        processAssertThrowsArguments(arguments.get(0), arguments.get(1));
+      }
+    } else {
+      TryStatementTree tryStatementTree = (TryStatementTree) tree;
+      if (isTryCatchIdiom(tryStatementTree)) {
+        tryStatementTree.catches().forEach(c ->
+          reportMultipleCallThrowingExceptionInTree(c.parameter().type().symbolType(), tryStatementTree.block(), c.parameter().type())
```
m-g-sonar, line 93: I don't really like that you are going to visit the try-block multiple times, once for each catches exception. What about collecting them first and then triggering a visit a single time?

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
quentin-jaquier-sonarsource: Definitely better.

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
quentin-jaquier-sonarsource: With the refactoring of the issue location, no need to visit the tree multiples times anymore!

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckExceptionCheck.java
```diff
@@ -0,0 +1,206 @@
...
+    .build();
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Arrays.asList(Tree.Kind.TRY_STATEMENT, Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
```
m-g-sonar, line 79: to be honest, I'm not sure this is still required in checks. From what I remember, we don't have situations anymore where semantic is not available.

PR #2924, opened 2020-04-28
SONARJAVA-3353 Rule S5776: Exception testing via JUnit ExpectedException rule should not be mixed with other assertions

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/ExpectedExceptionCheck.java
```diff
@@ -0,0 +1,98 @@
...
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.LambdaExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+
+@Rule(key = "S5776")
+public class ExpectedExceptionCheck extends AbstractMethodDetection {
+
+  private static final String MESSAGE = "Do not test exceptions via ExpectedException before other assertions.";
```
quentin-jaquier-sonarsource, line 40: As discussed, we should change the message to something actionable. 
`Consider using org.junit.Assert.assertThrows instead` for example.

PR #2925, opened 2020-04-28
SONARJAVA-3355 Improve SourceMap API to provide InputFile for source file

saberduck:


java-frontend/src/main/java/org/sonar/java/model/GeneratedFile.java
```diff
@@ -80,21 +81,26 @@ public void addSmap(SmapFile smap) {
...
-            int outputStart = lineInfo.outputStartLine + (i * lineInfo.outputLineIncrement);
-            int outputEnd = lineInfo.outputStartLine + ((i + 1) * lineInfo.outputLineIncrement) - 1;
-            // when outputLineIncrement == 0, end will be less than start (looks like bug in spec)
-            outputEnd = max(outputStart, outputEnd);
-            for (int j = outputStart; j <= outputEnd; j++) {
-              lines.merge(j, location, LocationImpl::mergeLocations);
-            }
-          }
+          Path sourcePath = sm.getUriRoot().resolve(sm.getFileSection().get(lineInfo.lineFileId).sourcePath);
+          Optional<InputFile> inputFile = fileResolver.apply(sourcePath);
```
benzonico, line 88: We can maybe avoid looking up the same lineFileId for every line by having just a check of the difference of lineInfo.lineFileId with the previous one as in most cases (if not all) we should not have a difference.

m-g-sonar:
Okay I have the feeling we are a bit missing the target here. We have the info about the InputFile in Jasper.java class when transpiling JSPs into smap files.
For it, it would make more sense to directly map GeneratedFiles with their root inputFile at this time, rather than doing it in two passes, like what it is doing in this PRs. Let's map files together when we have the info! I also feel that it would make less spaghetti code, and therefore be much more localized change.

saberduck:
@m-g-sonar indeed, I managed to get the generated file path directly from Jasper together with all the source maps, so I can avoid reading the filesystem. This simplified a lot of things, mainly thanks to assumption that 1 JSP generates 1 Java file, which seems to be always the case.

java-frontend/src/main/java/org/sonar/java/model/GeneratedFile.java
saberduck: in fact I had a look into Jasper code and it seems that 1 JSP always maps to single Java file, so this was unnecessary

java-frontend/src/main/java/org/sonar/java/model/GeneratedFile.java
```diff
@@ -48,17 +47,17 @@
...
   private final Path path;
 
   @VisibleForTesting
   final List<SmapFile> smapFiles = new ArrayList<>();
+  private final InputFile jspSourceFile;
 
   private SourceMap sourceMap;
 
-  public GeneratedFile(Path path) {
+  public GeneratedFile(Path path, @Nullable InputFile jspSourceFile) {
```
m-g-sonar: This should not be annotated with `@Nullable`. This can not happen in production, only in tests.

PR #2927, opened 2020-04-29
SONARJAVA-3357 S1452: java Collectors second parameter should be excluded

quentin-jaquier-sonarsource:


its/ruling/src/test/resources/guava/java-S1452.json
```diff
@@ -1,6 +1,7 @@
 {
 'com.google.guava:guava:src/com/google/common/base/Enums.java':[
 100,
+100,
```
quentin-jaquier-sonarsource, line 4: We detect this new issue since we are now detecting both `?` in code like:
```java
public List<? extends List<?>> getSomething()
```
while the old implementation was stopping the visit at the first WildcardTree found.

PR #2929, opened 2020-04-29
Sonarjava 3361 junit assert true should be simplified

johann-beleites-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/AssertTrueInsteadOfDedicatedAssertCheck.java
```diff
@@ -0,0 +1,205 @@
...
+    Arguments arguments = mit.arguments();
+
+    ExpressionTree argumentExpression;
+    if (hasBooleanArgumentAtPosition(arguments, 0)) {
+      argumentExpression = arguments.get(0);
+    } else if (hasBooleanArgumentAtPosition(arguments, arguments.size() - 1)) {
+      argumentExpression = arguments.get(arguments.size() - 1);
+    } else {
+      // We encountered a JUnit5 assert[True|False] method that accepts a BooleanSupplier - not supported.
+      return;
```
alban-auzeill, line 92: Let's add a test case to cover this case, and we can even use a false-negative to explicitly show the limitation of this rule:
```
Assertions.assertTrue(() -> foo == bar); // false-negative because BooleanSupplier is not supported by this rule
```

java-checks/src/main/java/org/sonar/java/checks/AssertTrueInsteadOfDedicatedAssertCheck.java
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * Returns the assertX method that should be used instead of assertTrue, if applicable.
+   *
+   * @param argumentExpression the boolean expression passed to assertTrue
+   * @return the assertion method to be used instead of assertTrue, or {@code null} if no better assertion method was determined
+   */
+  private static Optional<Assertion> getReplacementAssertion(@Nullable ExpressionTree argumentExpression) {
+    if (argumentExpression == null) {
+      return Optional.empty();
+    }
```
alban-auzeill, line 132: `argumentExpression` can not be `null`, if we pass a value to `getReplacementAssertion` coming from methods not annotated by `@Nullable` or `@CheckForNull`. It's enough to ensure that it's never null.

PR #2932, opened 2020-05-01
SONARJAVA-3360 Rule S5778: Only one method invocation is expected when testing runtime exceptions

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/AbstractOneExpectedExceptionRule.java
```diff
@@ -128,66 +127,42 @@ private static boolean isTryCatchFail(TryStatementTree tree) {
...
 
-  private static boolean isChecked(Type type) {
+  static boolean isChecked(Type type) {
     return !type.isSubtypeOf("java.lang.RuntimeException") && !type.isSubtypeOf("java.lang.Error");
   }
 
-  private static List<JavaFileScannerContext.Location> secondaryLocations(List<Tree> methodInvocationTrees) {
+  static List<JavaFileScannerContext.Location> secondaryLocations(List<Tree> methodInvocationTrees) {
     return methodInvocationTrees.stream()
       .map(expr -> new JavaFileScannerContext.Location("This call can throw an expected exception", expr))
```
m-g-sonar, line 138: I really don't like this message in fact... "expected" does not make sense. I would prefer to have `"This call can throw an exception"` instead, and ideally, even the name of the exception!

java-checks/src/main/java/org/sonar/java/checks/OneExpectedCheckedExceptionCheck.java
```diff
@@ -0,0 +1,59 @@
...
+    if (checkedTypes.isEmpty()) {
+      return;
+    }
+
+    MethodInvocationCollector visitor = new MethodInvocationCollector(symbol -> throwExpectedException(symbol, checkedTypes));
+    treeToVisit.accept(visitor);
+    List<Tree> invocationTree = visitor.invocationTree;
+    if (invocationTree.size() > 1) {
+      reportIssue(reportLocation,
+        String.format("Refactor the %s in order to have only one invocation throwing an expected exception.", placeToRefactor),
```
m-g-sonar, line 47: Here as well, "an expected" feels bad. It can happen only once, according to the way it is tested, so, to me it should be "the expected exception", or even "the exception".

java-checks/src/main/java/org/sonar/java/checks/OneExpectedRuntimeExceptionCheck.java
```diff
@@ -0,0 +1,52 @@
...
+    if (checkedTypes.isEmpty()) {
+      return;
+    }
+
+    MethodInvocationCollector visitor = new MethodInvocationCollector(symbol -> !JUNIT_FAIL_MATCHER.matches(symbol));
+    treeToVisit.accept(visitor);
+    List<Tree> invocationTree = visitor.invocationTree;
+    if (invocationTree.size() > 1) {
+      reportIssue(reportLocation,
+        String.format("Refactor the %s in order to have only one invocation throwing an expected exception.", placeToRefactor),
```
m-g-sonar, line 46: I would change it to "the excepted exception", or even "the exception". Simplicity is a key.

PR #2934, opened 2020-05-01
SONARJAVA-3334 Add support of Awaitility in S2699

m-g-sonar:


m-g-sonar:
Please do not squash commits when the review is done. I'll rework the PR if needed, but I want to keep commits separated.

PR #2935, opened 2020-05-05
SONARJAVA-3373 Rule S5810: JUnit5 test classes and methods should not have private visibility

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/AbstractJUnit5NotCompliantModifierChecker.java
```diff
@@ -55,18 +51,18 @@ public void visitNode(Tree tree) {
...
 
-  private void raiseIssueOnPresentAccessModifiers(ModifiersTree modifierTree) {
+  private void raiseIssueOnNotCompliantModifiers(ModifiersTree modifierTree) {
     modifierTree.modifiers().stream()
-      .filter(modifierKeyword -> NON_COMPLIANT_MODIFIERS.contains(modifierKeyword.modifier()))
-      .map(modifierKeywordTree -> (Tree) modifierKeywordTree)
+      .filter(modifier -> isNotCompliant(modifier.modifier()))
       .findFirst()
-      .ifPresent(modifierKeyword -> reportIssue(modifierKeyword, "Remove this access modifier"));
+      .ifPresent(modifier -> reportIssue(modifier, "Remove this '" + modifier.keyword().text() + "' modifier."));
```
johann-beleites-sonarsource, line 65: Do we have some standardized way of constructing strings? I've previously seen a lot of `String.format` as well - is that choice up to personal preference?

alban-auzeill: I didn't see any standard between `+` and `String.format` in our checks. A quick grep report 170 `reportIssue` with a `+` and 105 `String.format`. Personally, I use `String.format` when I want to share the message at several locations or because I want a specific format like `%04X`.

PR #2936, opened 2020-05-05
SONARJAVA-3258 Rule S5669: Vararg method arguments should not be confusing

m-g-sonar:


java-checks-test-sources/src/main/java/checks/ConfusingVarargCheck.java
```diff
@@ -0,0 +1,58 @@
...
+    Object nullObject = null;
+
+    new A(1, null); // Noncompliant [[sc=14;ec=18]] {{Cast this argument to 'String' to pass a single element to the vararg method.}}
+    new A(2, (String) null);
+    new A(3, args);
+    new A(4);
+
+    vararg(null); // Noncompliant [[sc=12;ec=16]] {{Cast this argument to 'Object' to pass a single element to the vararg method.}}
+    vararg((null)); // Noncompliant
+    vararg(arr); // Noncompliant [[sc=12;ec=15]] {{Cast this argument to 'Object' to pass a single element to the vararg method.}}
```
quentin-jaquier-sonarsource, line 17: I agree that there is something strange here, but is casting to Object the fix that we want to promote here?
This cast will be reported as redundant by IntelliJ, and will not change the behavior (still reports 1). It was not clear to me that this change was here to make it clear that I will have an argument of type Object[int[]], while I would expect a change that will make this call returning 3.

Am I missing something?

PR #2937, opened 2020-05-05
SONARJAVA-3367 use JUnit5 instead of JUnit4 annotations

johann-beleites-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/JUnit4AnnotationsCheck.java
```diff
@@ -0,0 +1,64 @@
...
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.ANNOTATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (!hasSemantic()) {
+      return;
+    }
```
quentin-jaquier-sonarsource, line 56: It needs to be clearly discussed and clarified at one point, but the case where the semantic is missing should never happen, we already removed this check for many rules. I believe you can also remove this one (and the corresponding test).

java-checks/src/test/java/org/sonar/java/checks/JUnit4AnnotationsCheckTest.java
```diff
@@ -0,0 +1,44 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java.checks;
+
+import org.junit.jupiter.api.Test;
+import org.sonar.java.checks.verifier.JavaCheckVerifier;
+
+import static org.sonar.java.CheckTestUtils.testSourcesPath;
+
+public class JUnit4AnnotationsCheckTest {
```
quentin-jaquier-sonarsource, line 27: 😮

java-checks/src/main/java/org/sonar/java/checks/JUnit4AnnotationsCheck.java
johann-beleites-sonarsource: Okay, good then I'll leave it out entirely in future. My latest understanding was that it can happen for some rules and not for others - far nicer if it doesn't need to be included everywhere!

johann-beleites-sonarsource: To clarify - do we generally still need the test with `.withoutSemantic()` in that case?

java-checks/src/test/java/org/sonar/java/checks/JUnit4AnnotationsCheckTest.java
johann-beleites-sonarsource: Oops, I hang my head in shame

java-checks/src/main/java/org/sonar/java/checks/JUnit4AnnotationsCheck.java
quentin-jaquier-sonarsource: No, we don't need it anymore.

PR #2939, opened 2020-05-05
SONARJAVA-2794 Rule S2479: Newline and control characters should not be used in string literals

alban-auzeill:


alban-auzeill:
:satisfied: Issues in the ruling are true-positives about the character U+00A0(no-break space) at:
```
sonar-server/src/main/java/org/sonar/server/plugins/ServerPluginRepository.java: line 272 column 67
sonar-server/src/main/java/org/sonar/server/qualityprofile/RegisterQualityProfiles.java: line 137 column 63
```

PR #2948, opened 2020-05-06
SONARJAVA-3346 Deprecate S1148 in favor of S4507

sebastian-hungerecker-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/DebugFeatureEnabledCheck.java
```diff
@@ -19,32 +19,74 @@
...
+      MemberSelectExpressionTree memberSelectExpressionTree = (MemberSelectExpressionTree) tree.methodSelect();
+      IdentifierTree identifierTree = memberSelectExpressionTree.identifier();
+      if (!enclosingClassExtendsThrowable()
+        && "printStackTrace".equals(identifierTree.name())
+        && calledOnTypeInheritedFromThrowable(memberSelectExpressionTree.expression())
+        && tree.arguments().isEmpty()) {
+        reportIssue(identifierTree, MESSAGE);
+      }
+    }
+  }
```
alban-auzeill, line 87: :scream: the previous implementation of the rule was complex and not using the MethodMatchers API.
we should use a matcher like:
```
  private static final MethodMatchers PRINT_STACK_TRACE_MATCHER = MethodMatchers.create()
    .ofSubTypes("java.lang.Throwable").names("printStackTrace").addWithoutParametersMatcher().build();
```
them `checkMethodInvocation` will become far more easier to read:
```
  private void checkMethodInvocation(MethodInvocationTree tree) {
    if (PRINT_STACK_TRACE_MATCHER.matches(tree)) {
      reportIssue(ExpressionUtils.methodName(tree), MESSAGE);
    }
  }
```

java-checks/src/main/java/org/sonar/java/checks/security/DebugFeatureEnabledCheck.java
```diff
@@ -67,4 +109,12 @@ private static AssignmentExpressionTree getDebugArgument(ExpressionTree expressi
...
   }
 
+  private boolean enclosingClassExtendsThrowable() {
+    return enclosingClass.peek() != null && enclosingClass.peek().type().isSubtypeOf("java.lang.Throwable");
+  }
+
+  private static boolean calledOnTypeInheritedFromThrowable(ExpressionTree tree) {
+    return tree.symbolType().isSubtypeOf("java.lang.Throwable");
+  }
+
```
alban-auzeill, line 119: With a MethodMatcher:

- the above two methods can be removed
- `enclosingClass` field can be removed
- nodesToVisit() on  Tree.Kind.CLASS can be removed
- leaveNode() can be removed

:sunglasses:

sebastian-hungerecker-sonarsource: The `enclosingClassExtendsThrowable` method (along witht the `enclosingClass` field) is there to prevent issues from being reported when calling `printWithStackTrace` inside classes extending `Throwable`.

As far as I can tell there's no way to say "Don't match when inside a class inheriting from X" with `MethodMatchers`, so that part of the code seems to still be necessary.

alban-auzeill: You are right, we need to keep `enclosingClass`, I didn't understand it correctly previously.

PR #2949, opened 2020-05-07
Update rules metadata

alban-auzeill:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S2129_java.html
```diff
@@ -25,5 +25,5 @@ <h2>Compliant Solution</h2>
 </pre>
 <h2>Exceptions</h2>
 <p><code>BigDecimal</code> constructor with <code>double</code> argument is ignored as using <code>valueOf</code> instead might change resulting
-value. See {rule:java:S2111}.</p>
+value. See {rule:java:S2111} .</p>
```
alban-auzeill, line 28: `S2111` is not converted into `{rule:java:S2111}` by rule-api if there's no space before the `.` :persevere:

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S3752_java.json
```diff
@@ -1,6 +1,6 @@
 {
-  "title": "\"@RequestMapping\" methods should specify HTTP method",
-  "type": "VULNERABILITY",
+  "title": "Allowing both safe and unsafe HTTP methods is security-sensitive",
```
quentin-jaquier-sonarsource, line 2: It seems that Eric is reworking this rule (SONARJAVA-3376), it requires some change in the rule, I believe we should not update the json and HTML now.

PR #2961, opened 2020-05-21
Fix typo in S2232_java.html

aikebah:
Compliant solution in documentation of S2232 is using wrong invariant for while-loop as condition-negation was apparently carried over unchanged from the non-compliant example.


quentin-jaquier-sonarsource:
Hello @aikebah 

I agree with your fix, however, we are not updating rule's descriptions directly in the `.hmtl`, but though Jira rule specification: [RSPEC-2232](https://jira.sonarsource.com/browse/RSPEC-2232).
I updated this description accordingly, it will be available for the next release.

Thanks for spotting this typo!
Best,
Quentin

PR #2966, opened 2020-05-26
SONARJAVA-3384 Rule S5831: AssertJ configuration should be applied

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
Yes, it would lead to FP. I don't think there is an easy way to support this case, one solution would be to go for symbolic execution, but at this point, I agree with you and also don't think it's worth it. We can always reconsider it in the future if we see that this rule behaves poorly, but it seems good enough for now.

PR #2969, opened 2020-05-28
Adding new Regex Parser (SONARJAVA-3425)

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
```diff
@@ -0,0 +1,106 @@
...
+ * The source code of a regular expression, made up out of 1 or more string literals
+ */
+public class RegexSource {
+
+  private final List<LiteralTree> stringLiterals;
+  private final String sourceText;
+  private final TreeMap<Integer, Integer> indices;
+
+  public RegexSource(List<LiteralTree> stringLiterals) {
+    this.stringLiterals = stringLiterals;
```
m-g-sonar, line 40: I wonder about this `List`. since we are passing the reference, it might be possible to modify it after. We probably better provide a `Collections.unmodifiableList(...)`.

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
```diff
@@ -0,0 +1,106 @@
...
+    Map.Entry<Integer, Integer> entry = indices.floorEntry(sourceIndex);
+    return new Position(entry.getKey(), entry.getValue());
+  }
+
+  private static String getString(LiteralTree literal) {
+    Optional<String> string = literal.asConstant(String.class);
+    if (string.isPresent()) {
+      return string.get();
+    } else {
+      throw new IllegalArgumentException("Only string literals allowed");
```
m-g-sonar, line 92: instead of throwing an exception here, I believe we could simply check at the very beginning of constructing the `RegexSource` if all the literals are `STRING_LITERAL`s, and throw if it's not the case.

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
```diff
@@ -0,0 +1,106 @@
...
+    return result;
+  }
+
+  private Position findPosition(int sourceIndex) {
+    Map.Entry<Integer, Integer> entry = indices.floorEntry(sourceIndex);
+    return new Position(entry.getKey(), entry.getValue());
+  }
+
+  private static String getString(LiteralTree literal) {
+    Optional<String> string = literal.asConstant(String.class);
```
m-g-sonar, line 88: can be written directly:
```
return literal.asConstant(String.class)
  .orElseThrow(() -> new IllegalArgumentException("Only string literals allowed"));
```

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
```diff
@@ -0,0 +1,106 @@
...
+    if (string.isPresent()) {
+      return string.get();
+    } else {
+      throw new IllegalArgumentException("Only string literals allowed");
+    }
+  }
+
+  private static class Position {
+    int indexOfLiteral;
```
m-g-sonar, line 97: even if used only for temporary objects when computing locations, these 2 fields can be `final`

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
```diff
@@ -0,0 +1,106 @@
...
+      int startIndex = startPosition.indexOfLiteral == i ? startPosition.indexInsideLiteral : 0;
+      int endIndex = endPosition.indexOfLiteral == i ? endPosition.indexInsideLiteral : length;
+      result.add(new Location(literal, startIndex, endIndex));
+    }
+    return result;
+  }
+
+  private Position findPosition(int sourceIndex) {
+    Map.Entry<Integer, Integer> entry = indices.floorEntry(sourceIndex);
+    return new Position(entry.getKey(), entry.getValue());
```
m-g-sonar, line 84: What about a `Position` constructor which exactly this instead of a dedicated method? Of course then the class can not be `static` anymore, but I think it makes sense.

```
Position startPosition = new Position(beginningOffset);
Position endPosition = new Position(endingOffset);
```
...
```
private class Position {
  Position(int sourceIndex) {
    Map.Entry<Integer, Integer> entry = indices.floorEntry(sourceIndex);
    this.indexOfLiteral = entry.getKey();
    this.indexInsideLiteral = entry.getValue();
  }

...
```

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
sebastian-hungerecker-sonarsource: @m-g-sonar Are you worried that `RegexSource` might modify the list? Or that someone else (with access to the original list) might modify the list after the `RegexSource` is created? Because `unmodifiableList` won't prevent the latter - we'd need to make an actual copy to protect against that. And as for the former, I don't think it's necessary to use `unmodifiableList` to prevent modifications from within the same class.

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
sebastian-hungerecker-sonarsource: That makes sense, I'll do that. Thank you.

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
sebastian-hungerecker-sonarsource: Actually you're right, we should be making it unmodifiable because the literals are accessible through a public getter. I forgot about that. Thanks.

sebastian-hungerecker-sonarsource: Though on third thought, it might make more sense to just remove the getter because I don't think we actually need it.

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
sebastian-hungerecker-sonarsource: We *are* doing this at the beginning of the constructor (that's when `getString` is called).

java-frontend/src/main/java/org/sonar/java/regex/ast/IndexRange.java
```diff
@@ -0,0 +1,67 @@
...
+  public int getBeginningOffset() {
+    return beginningOffset;
+  }
+
+  public int getEndingOffset() {
+    return endingOffset;
+  }
+
+  public IndexRange merge(IndexRange other) {
+    return new IndexRange(beginningOffset, other.endingOffset);
```
m-g-sonar, line 43: This merging can only work if the IndexRange given as argument is AFTER this IndexRange. Could it be done the other way or by construction we will always call it correctly?

sebastian-hungerecker-sonarsource: Barring bugs in the parser, it should only ever be called with the earlier index range first. It might be a good idea to verify this and throw an InvalidArgumentException otherwise though.

PR #2970, opened 2020-05-28
SONARJAVA-3390 Rule S5833: AssertJ methods setting the assertion context should come before an assertion

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/AssertJContextBeforeAssertionCheck.java
```diff
@@ -0,0 +1,86 @@
...
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S5833")
+public class AssertJContextBeforeAssertionCheck extends AbstractMethodDetection {
+
+  private static final MethodMatchers ASSERT_THAT = MethodMatchers.create()
+    .ofSubTypes("org.assertj.core.api.Assertions").names("assertThat", "assertThatObject").withAnyParameters().build();
```
quentin-jaquier-sonarsource, line 35: I realized that this is not enough, `assertThat` can come from many package:
`org.assertj.guava.api.Assertions.assertThat`, `org.assertj.core.api.AssertionsForClassTypes.assertThat`, ...

PR #2971, opened 2020-05-28
SONARJAVA-3389 S2698: improve issue reporting

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/AssertionsWithoutMessageCheck.java
```diff
@@ -38,99 +41,135 @@
 public class AssertionsWithoutMessageCheck extends AbstractMethodDetection {
 
   private static final String MESSAGE = "Add a message to this assertion.";
+  private static final String MESSAGE_FEST_LIKE = "Add a message to this assertion before calling this method.";
```
johann-beleites-sonarsource, line 44: Do you think it's helpful for a user to talk of assertion chains and predicates here? So something along the lines of `Add a message to this assertion chain before the predicate method`. I am imagining that users could be confused when they are supposed to add a message to this assertion but before 'this method', because they may think that this method *is* the assertion. For me, things got a lot easier to keep track of when we started to speak about chains, predicates and subjects.

quentin-jaquier-sonarsource: Yes, make sense, we introduced this vocabulary exactly to improve the clarity, we should stick to it.

PR #2973, opened 2020-05-29
SONARJAVA-3428 Introduce new method to report on parts of String literal

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/regex/RegexCheck.java
```diff
@@ -0,0 +1,82 @@
...
+    public List<RegexIssueLocation> toSingleLocationItems() {
+      if (locations.size() == 1) {
+        return Collections.singletonList(this);
+      }
+      return locations.stream()
+        .map(loc -> new RegexIssueLocation(loc, message))
+        .collect(Collectors.toList());
+    }
+
+    private static List<AnalyzerMessage.TextSpan> textSpansFromRegexTree(RegexSyntaxElement tree) {
```
sebastian-hungerecker-sonarsource, line 69: Should probably be called textSpansFromRegexSyntaxElement to fit the parameter type.

sebastian-hungerecker-sonarsource:
LGTM

PR #2979, opened 2020-06-02
SONARJAVA-3402 Rule S5853: Consecutive AssertJ "assertThat" statement should be chained

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/AssertJConsecutiveAssertionCheck.java
```diff
@@ -0,0 +1,150 @@
...
+      }
+    }
+    return Optional.empty();
+  }
+
+  private static boolean areEquivalent(ExpressionTree e1, ExpressionTree e2) {
+    if (e1.is(Tree.Kind.METHOD_INVOCATION)) {
+      // Two method invocation can return different values.
+      return false;
+    }
```
alban-auzeill, line 136: Checking `METHOD_INVOCATION` only at the root of an `ExpressionTree` is not enough, we can still have false-positives like:
```
    assertThat(getRandomArray().length).isNotEqualTo(3); // false-positive
    assertThat(getRandomArray().length).isNotEqualTo(4); 
```
I would start from `e1`
for all `MEMBER_SELECT` replace by MemberSelectExpressionTree#expression()
And finally, check `METHOD_INVOCATION`

PR #2980, opened 2020-06-03
SONARJAVA-3393 SONARJAVA-3406 AssertJ assertion chain simplifications

johann-beleites-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/AssertJChainSimplificationCheck.java
```diff
@@ -0,0 +1,132 @@
...
+    if (wasIssueRaised || predicates.size() > 1) {
+      return;
+    }
+
+    checkPredicatesForSimplification(
+      predicates, SIMPLIFIERS_WITH_CONTEXT, (simplifier, predicate) -> simplifier.simplify(subjectMit, predicate),
+      (predicate, replacement) -> reportIssue(ExpressionUtils.methodName(predicate),
+        String.format("Use %s instead", replacement),
+        Collections.singletonList(
+          new JavaFileScannerContext.Location("", subjectMit)),
```
alban-auzeill, line 94: When the secondary location message is empty, SonarQube seems buggy, it's not possible to navigate to a secondary location. Here, it's the "assertion subject". Or we can probably highlight only the `subjectMit`'s first argument as "actual value".
Or we can also completely remove the secondary location so we can hard code `reportIssue` in `checkPredicatesForSimplification` and get rid of the `BiConsumer<MethodInvocationTree, String> reportingMethod`

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5838_java.html
```diff
@@ -0,0 +1,78 @@
...
+    <tr>
+      <td><code>assertThat(x &lt;= y).isNotPositive()</code></td>
+      <td><code>assertThat(x).isLessThanOrEqualTo(y)</code></td>
+    </tr>
+    <tr>
+      <td><code>assertThat(x &lt; y).isTrue(0)</code></td>
+      <td><code>assertThat(x).isLessThan(y)</code></td>
+    </tr>
+  </tbody>
+</table>
```
alban-auzeill, line 73: The paragraph `<h2>Noncompliant Code Example</h2>` is missing. It's because it should be `h2. Noncompliant Code Example` and not `h2. Noncompliant Solution` in the RSPEC.

java-checks-test-sources/src/main/java/checks/AssertJChainSimplificationCheckTest.java
```diff
@@ -0,0 +1,99 @@
...
+    assertThat(x.compareTo(y)).isNegative(); // Noncompliant {{Use assertThat(actual).isLessThan(expected) instead}}
+    assertThat(x.compareTo(y)).isNotNegative(); // Noncompliant {{Use assertThat(actual).isGreaterThanOrEqualTo(expected) instead}}
+    assertThat(x.compareTo(y)).isPositive(); // Noncompliant {{Use assertThat(actual).isGreaterThan(expected) instead}}
+    assertThat(x.compareTo(y)).isNotPositive(); // Noncompliant {{Use assertThat(actual).isLessThanOrEqualTo(expected) instead}}
+
+    assertThat(x.compareTo(y)).isOne(); // Compliant
+
+    //assertThatObject();
+
+  }
```
alban-auzeill, line 98: This file contains a lot of `Noncompliant` test cases compared to the number of `Compliant`. We need to also add a good proportion to `Compliant` cases to assert that we don't raise issues on compliant code.

java-checks/src/main/java/org/sonar/java/checks/tests/AssertJChainSimplificationCheck.java
johann-beleites-sonarsource: What was our final agreement in regards to the message here? Is it fine to leave the message blank? (I think that makes sense, although we should investigate the SQ issue at the same time).

PR #2982, opened 2020-06-03
Regex/sebastian/improve parser

sebastian-hungerecker-sonarsource:


java-frontend/src/main/java/org/sonar/java/regex/JavaCharacterParser.java
```diff
@@ -0,0 +1,152 @@
...
+
+  public boolean isNotAtEnd() {
+    return current != null;
+  }
+
+  public boolean currentIs(char ch) {
+    return current != null && current.getCharacter() == ch;
+  }
+
+  private JavaCharacter parseJavaCharacter() {
```
m-g-sonar, line 90: Should be annotated with `@CheckForNull`

java-frontend/src/main/java/org/sonar/java/regex/JavaCharacterParser.java
```diff
@@ -0,0 +1,152 @@
...
+ */
+public class JavaCharacterParser {
+
+  public static final int EOF = -1;
+
+  private final RegexSource source;
+
+  private final JavaUnicodeEscapeParser unicodeProcessedCharacters;
+
+  private JavaCharacter current;
```
m-g-sonar, line 38: I would annotate this field with `@Nullable` and explicitly document when it will be null (end of source?)

java-frontend/src/main/java/org/sonar/java/regex/JavaCharacterParser.java
```diff
@@ -0,0 +1,152 @@
...
+      return null;
+    }
+    if (javaCharacter.getCharacter() == '\\') {
+      return parseJavaEscapeSequence(javaCharacter);
+    }
+    unicodeProcessedCharacters.moveNext();
+    return javaCharacter;
+  }
+
+  private JavaCharacter parseJavaEscapeSequence(JavaCharacter backslash) {
```
m-g-sonar, line 102: Should be annotated with `@CheckForNull` as well, when encountering invalid string literals.

java-frontend/src/main/java/org/sonar/java/regex/JavaCharacterParser.java
```diff
@@ -0,0 +1,152 @@
...
+    JavaCharacter javaCharacter = unicodeProcessedCharacters.getCurrent();
+    if (javaCharacter == null) {
+      // Should only happen in case of syntactically invalid string literals
+      return null;
+    }
+    char ch = javaCharacter.getCharacter();
+    switch (ch) {
+      case 'n':
+        ch = '\n';
+        unicodeProcessedCharacters.moveNext();
```
m-g-sonar, line 113: I think I would move out of the `switch` the statement `unicodeProcessedCharacters.moveNext();`, I think it simplifies a bit:

```
    switch (ch) {
      case 'n':
        ch = '\n';
        break;
      case 'r':
        ch = '\r';
        break;
      case 'f':
        ch = '\f';
        break;
      case 'b':
        ch = '\b';
        break;
      case 't':
        ch = '\t';
        break;
      default:
        if (isOctalDigit(ch)) {
          StringBuilder codeUnit = new StringBuilder(3);
          for (int i = 0; i < 3 && javaCharacter != null && isOctalDigit(javaCharacter.getCharacter()); i++) {
            codeUnit.append(javaCharacter.getCharacter());
            unicodeProcessedCharacters.moveNext();
            javaCharacter = unicodeProcessedCharacters.getCurrent();
          }
          ch = (char) Integer.parseInt(codeUnit.toString(), 8);
          return new JavaCharacter(source, backslash.getRange().extendTo(getCurrentStartIndex()), ch);
        }
    }
    unicodeProcessedCharacters.moveNext();
    return new JavaCharacter(source, backslash.getRange().extendTo(getCurrentStartIndex()), ch);
```

PR #2983, opened 2020-06-03
[REGEX] Extends cases supported by AbstractRegexCheck

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/regex/AbstractRegexCheck.java
```diff
@@ -53,27 +62,84 @@
...
+  }
+
+  private static Optional<LiteralTree[]> getLiteralsFromFinalVariables(IdentifierTree identifier) {
+    Symbol symbol = identifier.symbol();
+    if (!symbol.isVariableSymbol()) {
+      return Optional.empty();
+    }
+
+    Symbol.VariableSymbol variableSymbol = (Symbol.VariableSymbol) symbol;
+    if (!variableSymbol.isFinal()) {
```
sebastian-hungerecker-sonarsource, line 125: Is there a way to check whether a variable is *effectively* final? Because I'd like this to work with locals and locals are rarely explicitly defined as final.

m-g-sonar: actually, thanks to ECJ, there is! I'm trying this!

PR #2994, opened 2020-06-10
Regex/sebastian/group modifier parsing

sebastian-hungerecker-sonarsource:


m-g-sonar:
Yep, I guessed `AbstractRegexVisitorCheck` was introduced for this purpose, but **Less is More**. Let's introduce it when we will really need it, especially as there are big chances we will have to adapt it depending on the need.

PR #3002, opened 2020-06-11
Fix ruling with SonarQube >= 8 because of bad qualityprofiles/activate_rule parameters

alban-auzeill:
Parameters not anymore supported and needed to be renamed:
profile_key => key
rule_key => rule


m-g-sonar:
LGTM!

PR #3007, opened 2020-06-12
SONARJAVA-3405 Rule S5863: Assertions should not compare an object to itself

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/tests/AssertionCompareToSelfCheck.java
```diff
@@ -0,0 +1,189 @@
...
+      arePackageRelated(
+        packageOf(actualExpressionType.symbol()),
+        packageOf(classSymbolOfTheUnitTest(actualExpression)));
+  }
+
+  private static boolean isPrimitiveOrNull(Type actualExpressionType) {
+    return actualExpressionType.isPrimitive() || "null".equals(actualExpressionType.symbol().name());
+  }
+
+  static boolean arePackageRelated(String actualTypePackage, String unitTestPackage) {
```
quentin-jaquier-sonarsource, line 164: We should use `@VisibleForTesting` here.
Same for `classSymbolOfTheUnitTest`.

java-checks/src/main/java/org/sonar/java/checks/tests/AssertionCompareToSelfCheck.java
```diff
@@ -0,0 +1,189 @@
...
+    return unitTestPackage.startsWith(actualTypePackage) || actualTypePackage.startsWith(unitTestPackage);
+  }
+
+  @CheckForNull
+  static Symbol classSymbolOfTheUnitTest(@Nullable ExpressionTree actualExpression) {
```
quentin-jaquier-sonarsource, line 169: Can't we use `JUtils.enclosingClass` here?

its/ruling/src/test/resources/sonar-server/java-S5863.json
```diff
@@ -0,0 +1,8 @@
+{
+'org.sonarsource.sonarqube:sonar-server:src/test/java/org/sonar/server/computation/task/projectanalysis/api/measurecomputer/MeasureComputerDefinitionImplTest.java':[
+156,
+],
+'org.sonarsource.sonarqube:sonar-server:src/test/java/org/sonar/server/user/DefaultUserTest.java':[
+33,
```
quentin-jaquier-sonarsource, line 6: I'm still worried about the noise we will generate when reporting similar FPs.
The fact that we will consider this obvious FP as a "major bug" doesn't help...

Maybe we should ask for a third mind?

java-checks-test-sources/src/main/java/checks/AssertJConsecutiveAssertionCheck.java
```diff
@@ -193,6 +193,11 @@ void assert_on_member_select_7() {
     assertThat(fof.field.list.length).isNotEqualTo(1);
   }
 
+  @Test
+  void assert_on_constructor() {
+    assertThat(new java.util.Random()).isNotEqualTo(new java.util.Random()); // Compliant
```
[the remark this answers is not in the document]
alban-auzeill: :confounded: I just repeat 10 times in my head `MAKE YOUR TEST FAIL FIRST`!

java-checks/src/main/java/org/sonar/java/checks/tests/AssertionCompareToSelfCheck.java
```diff
@@ -0,0 +1,189 @@
...
+    return MethodMatchers.or(
+      ASSERTJ_AND_FEST_ASSERT_SUBJECT_METHODS,
+      JUNIT5_ASSERTIONS,
+      JUNIT4_ASSERTIONS_WITH_MESSAGE,
+      JUNIT4_ASSERTIONS_WITHOUT_MESSAGE);
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    if (JUNIT4_ASSERTIONS_WITH_MESSAGE.matches(mit)) {
```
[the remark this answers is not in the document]
alban-auzeill: Good advice, I don't need to visit NEW_CLASS and METHOD_REFERENCE, and it's easier.

java-checks/src/main/java/org/sonar/java/checks/tests/AssertionCompareToSelfCheck.java
alban-auzeill: :metal:

its/ruling/src/test/resources/sonar-server/java-S5863.json
alban-auzeill: After discussions, we decided to change the strategy. Instead of having an exception using packages name, we will try to base the exception on unit test method name containing some keywords (e.g. "equals", "hash_code")

PR #3010, opened 2020-06-12
SONARJAVA-3410 S5838: Optional-related rules for AssertJ chain simplification

johann-beleites-sonarsource:


java-checks-test-sources/src/main/java/checks/AssertJChainSimplificationCheckTest.java
```diff
@@ -305,6 +310,17 @@ void pathRelatedAssertionChains() {
     assertThat(getPath().isAbsolute()).isFalse(); // Noncompliant	{{Use assertThat(actual).isRelative() instead}}
   }
 
+  void optionalRelatedAssertionChains() {
+    assertThat(Optional.empty().isPresent()).isTrue(); // Noncompliant {{Use assertThat(actual).isPresent() or assertThat(actual).isNotEmpty() instead}}
```
quentin-jaquier-sonarsource, line 314: Do we really want to propose `isNotEmpty()` here?
Same for the others, I feel that we should stick to one variant.

johann-beleites-sonarsource: I considered that as well but then went with both as I felt like different phrasings might make sense in different contexts. Maybe you've established to always use `isPresent()` and `isEmpty()`, then suggesting only `isNotPresent()` seems a little unfitting. But perhaps someone else usually uses `isPresent()` and `isNotPresent()` throughout their code, then suggesting `isEmpty()` would seem off. That's why I went for including both, especially as the docs don't seem to suggest one more than the other. WDYT?

quentin-jaquier-sonarsource: I see, it's true that it works fine for all other cases, there is just this one that sounds strange when you look at it separately, but as a whole, it makes sense to propose both.

johann-beleites-sonarsource: I see, I agree that isNotEmpty() seems like the least intuitive method of those. I have one more thought that came to mind: what do you think of the fact that for collections AssertJ offers `isEmpty()` and `isNotEmpty()`. So there might be a natural tendency to those methods for some users over `is(Not)Present()`. Should that be something to consider here?

PR #3014, opened 2020-06-15
Continuing S5856

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/regex/ast/BackReferenceTree.java
```diff
@@ -28,7 +28,7 @@
   private final JavaCharacter key;
 
   public BackReferenceTree(RegexSource source, JavaCharacter backslash, @Nullable JavaCharacter key, JavaCharacter start, JavaCharacter end) {
-    super(source, backslash.getRange().merge(end.getRange()));
+    super(source, backslash.getRange().extendTo(end.getRange().getBeginningOffset()));
```
m-g-sonar, line 31: There is something fishy here... related to reporting.

java-checks-test-sources/src/main/java/checks/regex/InvalidRegexCheck.java
```diff
@@ -13,10 +13,26 @@ void noncompliant(String str) {
...
+      "(?<g1>ab)"
+        + "\\k<g2>" // Noncompliant [[sc=12;ec=19;secondary=25,26]] {{Fix the back reference errors inside this regex.}}
+        + "\\k<g3>"
+        + "(?<g2>cd)"
+        + "(?<g3>ed)");
 
-    str.matches("(?<name>\\w+)-\\k<nae>"); // False Negative - group names not handled in rule
+    str.matches(
+      "(?<g1>ab)"
+        + "\\2" // Noncompliant [[sc=12;ec=15;secondary=32,33]] {{Fix the back reference errors inside this regex.}}
```
sebastian-hungerecker-sonarsource, line 32: Interestingly, unlike named back references, numbered back references don't cause an exception when they don't exist. They just don't match anything when the group doesn't exist yet. You could even produce a marginally useful regex that refers to a group defined later that can actually produce a match: `(?:(?:x|\\1)(.))+`. This will match strings where the first character is an x and every other odd-numbered character is either an x or equal to the previous character. The equivalent using named captures will throw an exception.

So we might consider either not complaining about numbered back references at all (so we'd only be complaining about errors that cause an exception) or only complaining if the numbered back reference refers to a group that does not exist in the regex at all (i.e. referring to a group defined later would be allowed when using numbers), in which case the regex would still not cause an exception, but definitely be wrong.

On the other hand, we might also just say "99% of the time referencing a not-yet-defined group is a mistake and the 1% FPs is acceptable" and just leave it as-is.

sebastian-hungerecker-sonarsource: On second thought, given that they do not cause an exception, it'd probably be best to not handle numeric back references here and instead handle them in the "this pattern can never match" rule.

m-g-sonar: Fine by me, I'll only take into account named reference. And it will simplify the code of the rule as well!

m-g-sonar:
@sebastian-hungerecker-sonarsource reworked to remove reporting on wrong group numbers!

PR #3016, opened 2020-06-15
Rule S5869: Character classes in regular expressions should not ...

sebastian-hungerecker-sonarsource:
... contain the same character twice


java-checks/src/main/java/org/sonar/java/checks/regex/DuplicatesInCharacterClassCheck.java
```diff
@@ -0,0 +1,113 @@
...
+@Rule(key = "S5869")
+public class DuplicatesInCharacterClassCheck extends AbstractRegexCheck {
+
+  private static final String MESSAGE = "Remove duplicates in this character class.";
+
+  @Override
+  public void checkRegex(RegexParseResult regexForLiterals, MethodInvocationTree mit) {
+    if (!regexForLiterals.hasSyntaxErrors()) {
+      DuplicateFinder duplicateFinder = new DuplicateFinder();
+      duplicateFinder.setActiveFlags(getFlags(mit));
```
m-g-sonar, line 47: I would prefer to have this logic within the constructor of the `DuplicateFinder`, so we could rewrite this whole block such as:
```
if (!regexForLiterals.hasSyntaxErrors()) {
  regexForLiterals.getResult().accept(new DuplicateFinder(mit));
}

...

private class DuplicateFinder extends RegexBaseVisitor {
  private DuplicateFinder(MethodInvocationTree mit) {
    setActiveFlags(getFlags(mit));
  }

  ...
}

```

Or, what about having a final method `setActiveFlags(MethodInvocationTree mit)` within the `RegexBaseVisitor`, which would itself calls the currently private `getFlags(mit)` ?
So in our case we could just call `myVisitor.setActiveFlags(mit)`. I have the feeling that we will always have to call getFlags and setActiveFlags together.

java-checks/src/main/java/org/sonar/java/checks/regex/DuplicatesInCharacterClassCheck.java
```diff
@@ -0,0 +1,113 @@
...
+          CharacterRangeTree range = (CharacterRangeTree) element;
+          char lower = range.getLowerBound().getCharacter();
+          char upper = range.getUpperBound().getCharacter();
+          processRange(duplicates, inCharacterClass, lower, upper, range);
+        }
+      }
+      if (!duplicates.isEmpty()) {
+        List<RegexIssueLocation> secondaries = duplicates.stream()
+          .skip(1)
+          .map(duplicate -> new RegexIssueLocation(duplicate, MESSAGE))
```
m-g-sonar, line 72: Why not using another message here? it seems to me to be strange to have the same message for all secondaries.

sebastian-hungerecker-sonarsource: What would the other message be? Conceptually, the first duplicated character isn't any different than the others, so in principle we could even report one issue for each duplicated character. But of course that would produce way too much noise if a character class contains a lot of duplicates, so I think making the subsequent duplicates secondaries is preferable, but I don't know how their message should differ.

sebastian-hungerecker-sonarsource:
Please don't squash

java-checks/src/main/java/org/sonar/java/checks/regex/DuplicatesInCharacterClassCheck.java
sebastian-hungerecker-sonarsource: As discussed, I'll take care of this later in another PR after related changes from #3033 are merged.

m-g-sonar:
LGTM!

PR #3017, opened 2020-06-15
SONARJAVA-3417 Rule S5850: Alternatives in regular expressions ...

sebastian-hungerecker-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/regex/AnchorPrecedenceCheck.java
```diff
@@ -0,0 +1,84 @@
...
+      }
+      SequenceTree sequence = (SequenceTree) tree;
+      List<RegexTree> items = sequence.getItems().stream()
+        .filter(item -> !isFlagSetter(item))
+        .collect(Collectors.toList());
+      if (items.isEmpty()) {
+        return false;
+      }
+      int index = position == Position.BEGINNING ? 0 : (items.size() - 1);
+      return items.get(index).is(RegexTree.Kind.BOUNDARY);
```
m-g-sonar, line 71: I feel that we should also check here that the boundary type is the one expected, and I would add:
```
BoundaryTree boundary = items.get(index).is(RegexTree.Kind.BOUNDARY)
return (position == Position.BEGINNING && boundary.type() == BoundaryTree.Type.LINE_START) 
  || (position == Position.END && boundary.type() == BoundaryTree.Type.LINE_END);
```

Otherwise, I fear that we might report FPs on regex such as:
```
"\bTest|assert\b"
```
Which I believe are valid cases.

PR #3030, opened 2020-06-17
SONARJAVA-3386 FP on S2187 when test class is a JUnit4 test class also inheriting from a JUnit3 TestCase

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/NoTestInTestClassCheck.java
```diff
@@ -81,16 +82,12 @@ private void checkClass(ClassTree classTree) {
...
-          } else if (isArchUnitTestClass(classSymbol)) {
-            testMethodAnnotations.add(ARCH_UNIT_TEST);
-            testFieldAnnotations.add(ARCH_UNIT_TEST);
-          }
-          checkJunit4AndAboveTestClass(simpleName, classSymbol, members);
+        List<Symbol> membersList = members.collect(Collectors.toList());
+        if (isJunit3TestClass && containsJUnit3Tests(membersList)) {
+          return;
+        }
+        if (isJunit3TestClass || classSymbol.name().endsWith("Test")) {
```
johann-beleites-sonarsource, line 89: What exactly are you checking for with `.endsWith("Test")`? Would it make sense to also check for `*Test`, `*Tests` and `*TestCase`?

quentin-jaquier-sonarsource: In fact, this code is not really new, and come from [old line 117](https://github.com/SonarSource/sonar-java/pull/3030/files#diff-0c30a77861bf3b124e43378746e197c4L117).

I agree that we could improve this rule, hovewer, I believe it deserves a new ticket.
Does it make sense?

PR #3033, opened 2020-06-17
SONARJAVA-3423 Rule S5866: Case insensitive Unicode regular expressions should enable the “UNICODE_CASE” flag

sebastian-hungerecker-sonarsource:
Should be merged after #3016


java-checks-test-sources/src/main/java/checks/regex/UnicodeCaseCheck.java
```diff
@@ -0,0 +1,28 @@
+package checks.regex;
+
+import java.util.regex.Pattern;
+
+public class UnicodeCaseCheck {
+
+  void noncompliant(String str) {
+    Pattern.compile("söme pättern", Pattern.CASE_INSENSITIVE); // Noncompliant [sc=37;ec=61] {{Also use "Pattern.UNICODE_CASE" to correctly handle non-ASCII letters.}}
```
m-g-sonar, line 8: be careful, in order to test the location of the issues, you need 2 opening/closing brackets! For instance:  `[[sc=37,ec=61]]`.
I didn't check the other rule but I might have missed it, and with this PR, the precise location is not tested because of this error.

java-checks/src/main/java/org/sonar/java/checks/regex/UnicodeCaseCheck.java
```diff
@@ -0,0 +1,81 @@
...
+      Visitor visitor = new Visitor();
+      visitor.setActiveFlags(getFlags(mit));
+      visitor.visit(regexForLiterals.getResult());
+      if (visitor.problematicFlagSetOutsideOfRegex) {
+        getFlagsTree(mit).ifPresent( flagsTree ->
+          reportIssue(flagsTree, String.format(MESSAGE, "\"Pattern.UNICODE_CASE\""))
+        );
+      }
+      for (JavaCharacter flag : visitor.problematicFlags) {
+        reportIssue(flag, String.format(MESSAGE, "the \"u\" flag"), null, Collections.emptyList());
```
m-g-sonar, line 50: I'm not sure that I saw any test case which would benefit from having a loop here and report multiple issues for a given regex. would it be possible to add a test for it?

java-checks-test-sources/src/main/java/checks/regex/UnicodeCaseCheck.java
sebastian-hungerecker-sonarsource: I fixed this and also grepped through the other tests to see whether I made the same mistake elsewhere (I didn't).

java-checks/src/main/java/org/sonar/java/checks/regex/UnicodeCaseCheck.java
sebastian-hungerecker-sonarsource: Test case added.

PR #3037, opened 2020-06-18
Refactoring to avoid duplicated code for flags and error handling

sebastian-hungerecker-sonarsource:
Make the result of parsing a regex contain information about which
flags the regex was created with.

Use this information in the RegexBaseVisitor to set the initial
flags, removing any need to call setFlags, which is now removed.

Also move the logic of checking whether the regex contains syntax
errors before visiting anything into the RegexBaseVisitor and the
methods before and after to it, so now rules should all be
implementable with only a suitable definition of a visitor class
and then `new MyVisitor(possibleArguments).visit(regexForLiterals);`.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/regex/ast/RegexBaseVisitor.java
+++ b/java-frontend/src/main/java/org/sonar/java/regex/ast/RegexBaseVisitor.java
@@ -21,17 +21,12 @@ package org.sonar.java.regex.ast;
 
 import com.google.common.annotations.VisibleForTesting;
 import java.util.List;
-import java.util.regex.Pattern;
 import javax.annotation.CheckForNull;
+import org.sonar.java.regex.RegexParseResult;
 
 public class RegexBaseVisitor implements RegexVisitor {
 
-  private FlagSet activeFlags = new FlagSet();
-
-  @Override
-  public void setActiveFlags(int activeFlags) {
-    this.activeFlags.add(activeFlags);
-  }
+  private FlagSet activeFlags;
 
   @VisibleForTesting
   protected int getActiveFlags() {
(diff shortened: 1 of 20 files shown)


java-frontend/src/test/java/org/sonar/java/regex/RegexParserTestUtils.java
```diff
@@ -47,22 +49,32 @@
 
   private RegexParserTestUtils() {}
 
-  public static RegexTree assertSuccessfulParse(String regex) {
+  public static RegexParseResult assertSuccessfulParse(String regex) {
```
m-g-sonar, line 52: I'm really not sure about this change. Why not returning always the RegexTree and use the other methods `parseRegex(String regex)` when we need to test the flags.
When I look at all the tests which have been changed to call `getResult()` just after, I feel it reduced the readability of the test.

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexBaseVisitor.java
```diff
@@ -52,10 +47,38 @@ protected JavaCharacter getJavaCharacterForFlag(int flag) {
...
+
   private void visit(List<RegexTree> trees) {
     trees.forEach(this::visit);
   }
 
+  @Override
+  public void visit(RegexParseResult regexParseResult) {
+    if (!regexParseResult.hasSyntaxErrors()) {
+      activeFlags = regexParseResult.getInitialFlags();
+      before(regexParseResult);
```
m-g-sonar, line 62: This one is actually not used. Do you need it for a coming rule?
If it's not the case I would simply drop it for the time being (less is more).

PR #3045, opened 2020-06-19
SONARJAVA-3421 Rule S5857: Regular expressions character classes should be preferred over non-greedy quantifiers

sebastian-hungerecker-sonarsource:
Should be merged after #3044


m-g-sonar:
LGTM!

PR #3051, opened 2020-06-22
SONARJAVA-3422 RSPEC-5860 Names of regular expressions named groups should be used

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/regex/UnusedGroupNamesCheck.java
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  private static class KnownGroupsCollector extends RegexBaseVisitor {
+
+    private final RegexTree target;
+    private Map<String, CapturingGroupTree> groupsByName = new HashMap<>();
+    private Map<Integer, CapturingGroupTree> groupsByNumber = new HashMap<>();
+    private boolean usesBackReferences = false;
+
+    public KnownGroupsCollector(RegexParseResult regex) {
```
sebastian-hungerecker-sonarsource, line 272: This seems like a good opportunity to re-introduce the `RegexBaseVisitor.before` method and use that instead of the constructor to set `target`. That way you don't have to pass the regex to both the constructor and the `visit` method.

m-g-sonar: Damn, you are right... I'll re-introduce it then. :)

PR #3052, opened 2020-06-22
Fix the locations of escape sequences

sebastian-hungerecker-sonarsource:
Some escape sequences had locations which included one character too
many and/or did not include the leading backslash.


m-g-sonar:
LGTM!

PR #3055, opened 2020-06-23
SONARJAVA-3415 Rule S5843: Regular expressions should not be too complicated

sebastian-hungerecker-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/regex/AbstractRegexCheck.java
```diff
@@ -94,12 +94,20 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
     }
     FlagSet flags = getFlags(mit);
     if (!flags.contains(Pattern.LITERAL)) {
-      getLiterals(args.get(0))
```
m-g-sonar, line 97: As discussed, I feel that we should not change this logic and introduce some new overridable methods. It blurs the entry point of regex check, by providing many different ways to analyze regex.

I feel it would be better to keep AbstractRegexCheck having only `checkRegex()` being available for override, and unchanged in the way we use it. The rest of the methods being private or final. It makes all the rules consistent.

java-checks/src/main/java/org/sonar/java/checks/regex/RegexComplexityCheck.java
```diff
@@ -0,0 +1,235 @@
...
+  private static final int DEFAULT_MAX = 15;
+
+  @RuleProperty(
+    key = "maxComplexity",
+    description = "The maximum authorized complexity.",
+    defaultValue = "" + DEFAULT_MAX)
+  private int max = DEFAULT_MAX;
+
+  @Override
+  protected void checkRegexConstruction(ExpressionTree regexArgument, FlagSet flags, MethodInvocationTree mit) {
```
m-g-sonar, line 62: I would drop this part, and just change the logic of `checkRegex()`, which will then simply not use the regex provided as an entry point, and recompute some new ones based on the literal parts. the rule is only interested in bits of regex and that's ok. anyway, the full regex will be computed by other rules and reused.

java-frontend/src/main/java/org/sonar/java/regex/RegexParser.java
```diff
@@ -537,17 +539,22 @@ private RegexTree parseCharacterClass() {
       expected("']'");
     }
     IndexRange range = openingBracket.getRange().extendTo(characters.getCurrentStartIndex());
-    return new CharacterClassTree(source, range, negated, contents);
+    return new CharacterClassTree(source, range, openingBracket, negated, contents);
```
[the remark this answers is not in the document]
sebastian-hungerecker-sonarsource: I need the opening bracket in order to report a "+1" secondary only on the bracket (reporting on the whole character class can lead to overlapping secondaries if something inside the class has a "+n" as well and, as we say, overlapping secondaries lead to ugly output in SonarQube). I don't need the closing bracket for anything, so that's why it's not there.

I agree that it'd be more consistent to have both, but I feel like it's a similar thing to having `before` and `after`: only add both if and when you need both of them.

java-checks/src/main/java/org/sonar/java/checks/regex/RegexComplexityCheck.java
```diff
@@ -0,0 +1,235 @@
...
+    List<RegexIssueLocation> components = new ArrayList<>();
+
+    private RegexSyntaxElement firstComponent = null;
+
+    private void increaseComplexity(RegexSyntaxElement syntaxElement, int increment) {
+      complexity += increment;
+      if (firstComponent == null) {
+        firstComponent = syntaxElement;
+      }
+      components.add(new RegexIssueLocation(syntaxElement, "+" + increment));
```
m-g-sonar, line 141: Looking at examples of issues from peach, I think it would be nice to add the amount caused by nesting (to be closer from cognitive complexity), such as:
```
+2 (incl 1 for nesting)
```

java-checks/src/main/java/org/sonar/java/checks/regex/RegexComplexityCheck.java
```diff
@@ -0,0 +1,299 @@
...
+
+  @RuleProperty(
+    key = "maxComplexity",
+    description = "The maximum authorized complexity.",
+    defaultValue = "" + DEFAULT_MAX)
+  private int max = DEFAULT_MAX;
+
+  private final List<RegexConstructionInfo> regexConstructions = new ArrayList<>();
+
+  private final Set<Integer> commentedLines = new HashSet<>();
```
m-g-sonar, line 66: you need to clear the list when leaving the file. It's the same instance of the rule which is used for all the checks. If not cleared, you store all the comments encountered along the analysis. To do so, simply override `leaveFile()` method as follow:

```
  @Override
    commentedLines.clear();
  }
```

m-g-sonar: Note that it might be the reason why sanity fails.

sebastian-hungerecker-sonarsource:
This PR has been merged, but does not show up as such because github was acting up.

PR #3057, opened 2020-06-24
[REGEX] Fix handling of escapes in unicode parser

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/regex/ast/SequenceTreeTest.java
```diff
@@ -34,4 +36,16 @@ void empty_string() {
...
   }
 
+  @Test
+  void escapedRegexEscape() {
+    // actual regular expression catching unicode characters in strings literals: "\\u+[a-fA-F0-9]{4}"
+    RegexTree regex = assertSuccessfulParse("\\\\\\\\u+");
+    assertKind(RegexTree.Kind.SEQUENCE, regex);
+    List<RegexTree> items = ((SequenceTree) regex).getItems();
+    assertThat(items).hasSize(2);
+//    assertPlainCharacter('\\', items.get(0));
```
m-g-sonar, line 46: @sebastian-hungerecker-sonarsource I'm really not sure about this. Shouldn't I have 3 items here ? I was expecting 2 backslashes.

sebastian-hungerecker-sonarsource: Note that there are three levels of escaping here (because this is written as a String inside the tests instead of in its own .java file as the rule tests are): "\\\\\\\\u+" creates a string containing four backslashes. The JavaCharacter parser turns this into two backslashes. The RegexParser then sees those two backslashes, parses them as an escape sequence and produces a single PlainCharacterTree representing a backslash. So it's all good.

java-frontend/src/main/java/org/sonar/java/regex/JavaUnicodeEscapeParser.java
```diff
@@ -62,7 +63,13 @@ public void moveNext() {
     }
     int startIndex = index;
     char ch;
-    if (index < sourceText.length() - 1 && sourceText.charAt(index) == '\\' && sourceText.charAt(index + 1) == 'u') {
+    if (sourceText.charAt(index) == '\\') {
+      isEscaping = !isEscaping;
```
sebastian-hungerecker-sonarsource, line 67: It looks to me as if something like `\t\u1234` isn't going to be recognized as a Unicode escape this way because the `\t` sets `isEscaping` to `true` and the backslash from `\u` sets it to false.

m-g-sonar:
Updated @sebastian-hungerecker-sonarsource !

PR #3059, opened 2020-06-25
SONARJAVA-3424 Implementation of S5867

johann-beleites-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/regex/UnicodeAwareCharClassesCheck.java
```diff
@@ -0,0 +1,110 @@
...
+    if (!unicodeUnawareCharClassFinder.unicodeAwareWithFlag.isEmpty() &&
+      !unicodeUnawareCharClassFinder.containsUnicodeCharacterFlag()) {
+
+      List<RegexCheck.RegexIssueLocation> secondaries = unicodeUnawareCharClassFinder.unicodeAwareWithFlag.stream()
+        .map(tree -> new RegexIssueLocation(tree, "Predefined/POSIX character class"))
+        .collect(Collectors.toList());
+      reportIssue(ExpressionUtils.methodName(mit),
+        "Enable the \"UNICODE_CHARACTER_CLASS\" flag or use a Unicode-aware alternative.", null, secondaries);
+    }
+  }
```
sebastian-hungerecker-sonarsource, line 69: I'd prefer to move the bulk of this method into `UnicodeUnawareCharClassFinder.after` (removing the `static` modifier from `UnicodeUnawareCharClassFinder`) and keep the tradition of `checkRegex` being implemented as only `new UnicodeUnawareCharClassFinder().visit(regexForLiterals);`.

java-checks/src/main/java/org/sonar/java/checks/regex/UnicodeAwareCharClassesCheck.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Override
+    protected void doVisitNonCapturingGroup(NonCapturingGroupTree tree) {
+      containsUnicodeCharacterFlag |= flagActive(Pattern.UNICODE_CHARACTER_CLASS);
+      super.doVisitNonCapturingGroup(tree);
+    }
+
+    public boolean containsUnicodeCharacterFlag() {
+      return containsUnicodeCharacterFlag;
+    }
```
sebastian-hungerecker-sonarsource, line 108: I think this getter can be removed. You access the other variables without a getter as well and if you move the code using the variables into `after`, there's no need for getters anyway.

PR #3061, opened 2020-06-26
SONARJAVA-3419 Rule S5854: Regex containing characters subjects to normalization should use the CANON_EQ flag

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/regex/CanonEqFlagInRegexCheck.java
```diff
@@ -0,0 +1,136 @@
...
+      String endOfMessage;
+      if (STRING_MATCHES.matches(mit) || PATTERN_MATCHES.matches(mit)) {
+        endOfMessage = "\"Pattern.compile(regex, CANON_EQ).matcher(input).matches()\"";
+      } else if (STRING_REPLACE_ALL.matches(mit)) {
+        endOfMessage = "\"Pattern.compile(pattern, CANON_EQ).matcher(input).replaceAll(replacement)\"";
+      } else if (STRING_REPLACE_FIRST.matches(mit)) {
+        endOfMessage = "\"Pattern.compile(pattern, CANON_EQ).matcher(input).replaceFirst(replacement)\"";
+      } else {
+        endOfMessage = "this pattern";
+      }
```
sebastian-hungerecker-sonarsource, line 82: Nice catch that enabling the flag might require using a different method and to include this in the message. :+1:

PR #3062, opened 2020-06-29
SONARJAVA-3426 Rule S5868: Unicode Grapheme Clusters should be avoided inside regex character classes

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/regex/GraphemeClustersInClassesCheck.java
```diff
@@ -0,0 +1,85 @@
...
+import org.sonar.java.regex.ast.JavaCharacter;
+import org.sonar.java.regex.ast.PlainCharacterTree;
+import org.sonar.java.regex.ast.RegexBaseVisitor;
+import org.sonar.java.regex.ast.RegexTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+
+@Rule(key = "S5868")
+public class GraphemeClustersInClassesCheck extends AbstractRegexCheck {
+
+  private static final String MESSAGE = "Extract these Grapheme Clusters from the character class.";
```
quentin-jaquier-sonarsource, line 38: The rule description proposes a more precise message when the regex is "simple enough", however, I'm not sure to get why are the benefits. If you feel that it would be useful, let's discuss it!

java-checks/src/main/java/org/sonar/java/checks/CheckList.java
```diff
@@ -398,6 +399,7 @@ private CheckList() {
       ForLoopUsedAsWhileLoopCheck.class,
       ForLoopVariableTypeCheck.class,
       GarbageCollectorCalledCheck.class,
+      GraphemeClustersInClassesCheck.class,
```
m-g-sonar, line 402: Not sure `'r'` is before `'e'`

java-checks-test-sources/src/main/java/checks/regex/GraphemeClustersInClassesCheck.java
```diff
@@ -0,0 +1,28 @@
+package checks.regex;
+
+import java.util.regex.Pattern;
+
+public class GraphemeClustersInClassesCheck {
+
+  void noncompliant(String str) {
+    Pattern.compile("aaa[è]aaa"); // Noncompliant [[sc=25;ec=29]] {{Extract these Grapheme Clusters from the character class.}}
```
m-g-sonar, line 8: The issue message is in the plural, while there is apparently only one cluster, I feel it is strange
 Why not reporting directly on it and use a singular form?

java-checks-test-sources/src/main/java/checks/regex/GraphemeClustersInClassesCheck.java
```diff
@@ -0,0 +1,28 @@
...
+    Pattern.compile("[èa-dè]"); // Noncompliant [[sc=22;ec=31]] {{Extract these Grapheme Clusters from the character class.}}
+    // Noncompliant@+1
```
m-g-sonar, line 10: instead of using this, you can use a String concatenation and a different line

java-checks/src/main/java/org/sonar/java/checks/regex/GraphemeClustersInClassesCheck.java
m-g-sonar: I think it would be valuable to use at least a singular/plural form if not using the simpler case.

java-checks/src/main/java/org/sonar/java/checks/regex/GraphemeClustersInClassesCheck.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        previousChar = null;
+      }
+
+      super.visitCharacterClassUnion(tree);
+    }
+
+    private boolean canCanonicallyCombine(@Nullable JavaCharacter previousChar, char currentChar) {
+      return previousChar != null
+        && Normalizer.normalize(new StringBuilder().append(previousChar.getCharacter()).append(currentChar), Normalizer.Form.NFC).length() == 1;
```
m-g-sonar, line 81: I feel it is not enough.

I was surprised to see that a simpler version of the test case from the `Noncompliant` part of the RSPEC was marked as Compliant, while it shouldn't. The following code, when executed:

```
"cc̈c".replaceAll("[c̈]", "X");
```
would print XXXX, instead of cXc, and it should be reported!

PR #3074, opened 2020-07-01
SONARJAVA-3413 RSPEC-5842 Regex repetition pattern's body should not match the empty String

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/regex/EmptyStringRepetitionCheck.java
```diff
@@ -0,0 +1,74 @@
...
+  public void checkRegex(RegexParseResult regex, MethodInvocationTree mit) {
+    new Visitor().visit(regex);
+  }
+
+  private class Visitor extends RegexBaseVisitor {
+
+    @Override
+    public void visitRepetition(RepetitionTree tree) {
+      if (matchEmptyString(tree.getElement())) {
+        reportIssue(tree, MESSAGE, null, Collections.emptyList());
```
sebastian-hungerecker-sonarsource, line 48: The issue should either be reported on `tree.getElement()` instead of `tree` or the message should be changed. The way it is now, the message would imply that in `(x*)*` the whole expression should be changed to not match the empty string when in fact only the inner part (`(x*)`) should be changed to not match the empty string.

PR #3083, opened 2020-07-06
Fix parsing of character ranges involving hex literals

sebastian-hungerecker-sonarsource:
Also add support for hex literals for rules S5869 and S5857.


java-checks-test-sources/src/main/java/checks/regex/DuplicatesInCharacterClassCheck.java
```diff
@@ -17,6 +17,7 @@ void nonCompliant() {
     str.matches("(?iU)[äÄ]"); // Noncompliant [[sc=25;ec=26]]
     str.matches("(?iu)[xX]"); // Noncompliant [[sc=25;ec=26]]
     str.matches("[\\\"\\\".]"); // Noncompliant [[sc=23;ec=27]]
+    str.matches("[\\x{1F600}-\\x{1F637}\\x{1F608}]"); // Noncompliant [[sc=40;ec=50]]]");
```
m-g-sonar, line 20: The non-compliant comment is not correctly ending.

PR #3095, opened 2020-07-07
SONARJAVA-3237 S1142 should be ignored in equals methods

sebastian-hungerecker-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/MethodWithExcessiveReturnsCheck.java
```diff
@@ -68,7 +69,13 @@ public void visitNode(Tree tree) {
   public void leaveNode(Tree tree) {
     Tree reportTree = null;
     if (tree.is(Tree.Kind.METHOD)) {
-      reportTree = ((MethodTree) tree).simpleName();
+      IdentifierTree methodName = ((MethodTree) tree).simpleName();
+      if ("equals".equals(methodName.name())) {
```
quentin-jaquier-sonarsource, line 73: You should use `MethodTreeUtils.isEqualsMethod` here

java-checks/src/test/files/checks/MethodWithExcessiveReturnsCheck.java
```diff
@@ -46,6 +46,23 @@ public void g() { // Compliant
     return;
     return;
   }
+
```
quentin-jaquier-sonarsource, line 49: We should also consider moving test files related to the new compiled module.

PR #3097, opened 2020-07-08
IT: Add a test which perform a self-analysis of sonar-java in order to prevent direct regression

m-g-sonar:


m-g-sonar:
closed, as it currently consumes too much time on ITs (+7min on PR)

PR #3103, opened 2020-07-09
SONARJAVA-3459 Only check for words in comments rather than any occurence

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/UselessImportCheck.java
```diff
@@ -52,6 +54,8 @@
 @Rule(key = "S1128")
 public class UselessImportCheck extends BaseTreeVisitor implements JavaFileScanner {
 
+  private static final Pattern NON_WORDS_OR_WHITE_SPACES_CHARACTERS = Pattern.compile("(?:\\W|\\s)+");
```
sebastian-hungerecker-sonarsource, line 57: The `\\s` is redundant here. `\\W` already matches everything matched by `\\s`. So this can just be `"\\W+"`

PR #3107, opened 2020-07-10
SONARJAVA-3400 Handle DateFormat.getDateInstance()

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/StaticMultithreadedUnsafeFieldsCheck.java
```diff
@@ -61,8 +68,20 @@ public void visitNode(Tree tree) {
   }
 
   private static boolean isForbiddenType(VariableTree variableTree) {
```
sebastian-hungerecker-sonarsource, line 70: The method should probably be renamed since it's no longer just about the type. Maybe just `isForbidden`.

java-checks/src/main/java/org/sonar/java/checks/StaticMultithreadedUnsafeFieldsCheck.java
```diff
@@ -61,8 +68,20 @@ public void visitNode(Tree tree) {
...
+    if (isForbiddenType(variableTree.type().symbolType())) {
+      return true;
+    }
     ExpressionTree initializer = variableTree.initializer();
-    return isForbiddenType(variableTree.type().symbolType()) || (initializer != null && !initializer.is(Tree.Kind.NULL_LITERAL) && isForbiddenType(initializer.symbolType()));
+    if (initializer == null) {
+      return false;
+    }
+    if (initializer.is(Tree.Kind.NULL_LITERAL)) {
+      return false;
```
sebastian-hungerecker-sonarsource, line 79: The two `if`s that return `false` can be combined into one:

    if (initializer == null || initializer.is(Tree.Kind.NULL_LITERAL)) {
      return false;
    }

java-checks/src/main/java/org/sonar/java/checks/StaticMultithreadedUnsafeFieldsCheck.java
```diff
@@ -61,8 +68,20 @@ public void visitNode(Tree tree) {
...
+    }
+    if (initializer.is(Tree.Kind.METHOD_INVOCATION)) {
+      return GET_DATE_INSTANCE.matches((MethodInvocationTree) initializer);
```
sebastian-hungerecker-sonarsource, line 82: This will break the case where the initializer is a method call returning a forbidden type (which should be added as a test case since currently there are only test where a constructor is called rather than a method). The check for `isForbiddenType(initializer.symbolType())` should come before this one. Or alternatively this one should only return if the method matcher matches, i.e.:

    if (initializer.is(Tree.Kind.METHOD_INVOCATION) && GET_DATE_INSTANCE.matches((MethodInvocationTree) initializer)) {
      return true;
    }

PR #3109, opened 2020-07-13
SONARJAVA-3460 S3457, S2275: Rework printf-style format rules

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/AbstractPrintfChecker.java
```diff
@@ -77,45 +83,22 @@
   protected static final MethodMatchers JAVA_UTIL_LOGGER_LOG_MATCHER = MethodMatchers.or(
```
alban-auzeill, line 83: All constants related to logger can be `private` and moved in the PrintfMisuseCheck class

PR #3125, opened 2020-08-03
SONARJAVA-3494: Improve performance of S2095

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -87,6 +89,8 @@ public String valueAsString() {
   public String excludedTypes = "";
   private final List<String> excludedTypesList = new ArrayList<>();
 
+  private final Set<TryStatementTree> tryWithResourcesTrees = new HashSet<>();
```
quentin-jaquier-sonarsource, line 94: It was not clear to me that the content of this set was used only to know if you already collected the resources from the try. Maybe renaming it as `visitedTryWithResourcesTrees ` or something like this would help?

java-frontend/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -129,18 +135,60 @@ public String valueAsString() {
...
+    if (syntaxNode.is(Tree.Kind.TRY_STATEMENT)) {
+      TryStatementTree tryStatementTree = (TryStatementTree) syntaxNode;
+      ListTree<Tree> resourceList = tryStatementTree.resourceList();
+      if (!resourceList.isEmpty() && tryWithResourcesTrees.add(tryStatementTree)) {
+        knownResources.addAll(collectAllResourceTrees(resourceList));
+      }
+    }
+  }
+
+  private static Set<Tree> collectAllResourceTrees(Tree tree) {
```
quentin-jaquier-sonarsource, line 168: Question: I would typically use a `BaseTreeVisitor` to do something like this, is there any reasons you are using something else?

m-g-sonar: I initially wanted to collect all possible trees, without exception, and with a `BaseTreeVisitor` I would have had to override all the methods to do so. On a second pass, I only collected identifiers, method invocations, and initializers, so the approach with a visitor is indeed easier to get.

PR #3129, opened 2020-08-06
SONARJAVA-3447 FN on S2259 when a method is annotated with spring's annotation @Nullable

quentin-jaquier-sonarsource:


java-frontend/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -222,7 +222,9 @@ public ProgramState checkPostStatement(CheckerContext context, Tree syntaxNode)
   }
 
   private static boolean isAnnotatedCheckForNull(MethodInvocationTree syntaxNode) {
-    return syntaxNode.symbol().metadata().isAnnotatedWith("javax.annotation.CheckForNull");
+    return syntaxNode.symbol().metadata().isAnnotatedWith("javax.annotation.CheckForNull")
```
m-g-sonar, line 225: you can avoid calling two times `syntaxNode.symbol().metadata()` by storing it first.

java-frontend/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -222,7 +222,9 @@ public ProgramState checkPostStatement(CheckerContext context, Tree syntaxNode)
...
+      // Despite the name, Spring Nullable is meant to be used as CheckForNull
+      || syntaxNode.symbol().metadata().isAnnotatedWith("org.springframework.lang.Nullable");
```
m-g-sonar, line 227: In fact, I think we should rather rely on the `NullnessAnnotationUtils` class here, and move this method there.
There is already a list of nullable annotations, and I believe we should do another pass on the *Nullable* annotations list, to check if some of them are in fact strong nullness constraints (*a la* javax `CheckForNull`), and not weak (*a la* javax `Nullable`), to be used in such context.

quentin-jaquier-sonarsource:
As proposed, I did a second path on the annotations that we support:

- "javax.annotation.CheckForNull"
Strong
- "javax.annotation.Nullable"
Weak
- "org.springframework.lang.Nullable"
Strong, see ticket
- "org.checkerframework.checker.nullness.compatqual.NullableDecl"
Weak: from the [doc](https://checkerframework.org/api/org/checkerframework/checker/nullness/compatqual/NullableDecl.html):
> "Identical as @Nullable [...]"
- "edu.umd.cs.findbugs.annotations.Nullable"
Weark, annotated with `@javax.annotation.Nonnull(when = When.UNKNOWN)` (same as @Nullable)

For all the others, it's not crystal clear from the doc, but they are closer to weak.

- ["android.support.annotation.Nullable"](https://android.googlesource.com/platform/frameworks/support/+/300475fc42431cefb0d8f1bfc70de1a7a052a501/annotations/src/main/java/android/support/annotation/Nullable.java)
- ["androidx.annotation.Nullable"](https://developer.android.com/reference/androidx/annotation/Nullable)
- ["org.eclipse.jdt.annotation.Nullable"](https://help.eclipse.org/2020-06/index.jsp?topic=/org.eclipse.jdt.doc.isv/reference/api/org/eclipse/jdt/annotation/Nullable.html)
- "[org.jetbrains.annotations.Nullable](https://www.jetbrains.com/help/idea/nullable-and-notnull-annotations.html#nullable)"
- ["org.checkerframework.checker.nullness.qual.Nullable"](https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/Nullable.html)

PR #3133, opened 2020-08-06
SONARJAVA-3026 S3518 Division by zero on floats and double should not mention ArithmeticException

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
Not sure if I missed something, but to me, there is no mention of ArithmeticException in any of the messages or secondary locations.

The message is:

> Make sure "xxx" can't be zero before doing this division.

Which suits well the situation where "xxx" is a double or float.
The only part that can be slightly confusing is the description, I updated it to clearly state that an error is not always the outcome, but the situation should still be avoided.

m-g-sonar:
LGTM!

PR #3135, opened 2020-08-07
SONARJAVA-3069 Adapt SE engine to Switch Expressions

sebastian-hungerecker-sonarsource:
Fix CFG generation for switch expressions.

The generated CFG for switch expressions previously included the
switch expression twice, once as a terminator (as it should be) and
once as an expression in the block that used the result of the switch
expression. This would make the switch expression the top of the stack
instead of the result and once the user popped the switch expression,
the result of the switch expression would be left on the stack when
the stack should either be empty or contain the results of other
subexpressions.

With the CFG fixed, switch expressions should now be handled properly
with no need to change the symbolic execution engine itself.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
+++ b/java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
@@ -513,12 +513,10 @@ public class ExplodedGraphWalker {
   }
 
   private void handleSwitch(CFG.Block programPosition, List<CaseGroupTree> caseGroups) {
-    ProgramState.Pop poppedSwitchValue = programState.unstackValue(1);
-    ProgramState.SymbolicValueSymbol switchValue = poppedSwitchValue.valuesAndSymbols.get(0);
-    ProgramState state = poppedSwitchValue.state;
+    ProgramState state = programState;
 
     Map<CaseGroupTree, List<ProgramState.SymbolicValueSymbol>> caseValues = new HashMap<>();
-    for (CaseGroupTree caseGroup : caseGroups) {
+    for (CaseGroupTree caseGroup : Lists.reverse(caseGroups)) {
       int numberOfCaseValues = caseGroup.labels()
         .stream()
         .map(CaseLabelTree::expressions)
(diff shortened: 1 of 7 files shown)


java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -502,6 +512,59 @@ private static ExpressionTree cleanupCondition(ExpressionTree condition) {
...
+    assert(elseBlock != null);
+    enqueue(new ProgramPoint(elseBlock), elseState, node.exitPath);
+  }
+
+  private static ProgramState setConstraint(ProgramState state, SymbolicValue condition, BooleanConstraint constraint) {
+    List<ProgramState> states = condition.setConstraint(state, constraint);
+    if (states.isEmpty()) {
+      return state;
+    }
+    assert(states.size() == 1);
```
sebastian-hungerecker-sonarsource, line 564: Is this always true or just in the case of boolean constraints? If not, under which conditions would `setConstraint` return a list of multiple items? And if so, why doesn't it return an option instead of a list?

java-frontend/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -502,6 +512,59 @@ private static ExpressionTree cleanupCondition(ExpressionTree condition) {
...
+      if (successor.isDefaultBlock()) {
+        assert(elseBlock == null);
+        elseBlock = successor;
+      }
+    }
+    assert(elseBlock != null);
+    enqueue(new ProgramPoint(elseBlock), elseState, node.exitPath);
+  }
+
+  private static ProgramState setConstraint(ProgramState state, SymbolicValue condition, BooleanConstraint constraint) {
```
sebastian-hungerecker-sonarsource, line 559: I'm a bit unsure about this method because I'm generally a bit unsure about what SymbolicValue.setConstraint does (as opposed to ProgramState.addConstraint). From the code it looks like it adds the constraint unless it's already there, but that doesn't explain why it failed to properly add the constraints when I used `ProgramState.addConstraint` instead of this method.

PR #3136, opened 2020-08-11
SONARJAVA-3186 SE rules should consider unknown exception as runtime.

quentin-jaquier-sonarsource:


java-frontend/src/main/java/org/sonar/java/se/ProgramState.java
```diff
@@ -533,7 +533,7 @@ public SymbolicValue exitValue() {
   public boolean exitingOnRuntimeException() {
```
quentin-jaquier-sonarsource, line 533: This helper is used by 5 rules:

- S3546: CustomUnclosedResourcesCheck
- S351: InvariantReturnCheck
- S2222: LocksNotUnlockedCheck
- S2095: UnclosedResourcesCheck
- S2755: XxeProcessingCheck

For me, when an exception is unknown, it's either:
- we failed to correctly resolve the semantic
- it's an exception added by the engine to model the fact that something could go wrong in certain situations.

In both cases, it makes sense to not report an issue to avoid FPs.

PR #3146, opened 2020-08-19
Duplicates #3144 for ITs

m-g-sonar:
Duplicates #3144  for ITs


m-g-sonar:
Breaks the unit tests, we need to adapt a few checks to make sure it works.

PR #3169, opened 2020-09-07
SONARJAVA-3372 Rule S5803: Class members annotated with @VisibleForTesting should not be accessed from production code

margarita-nedzelska-sonarsource:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5803_java.json
```diff
@@ -0,0 +1,12 @@
+{
+  "title": "Class members annotated with \"@VisibleForTesting\" should not be accessed from production code ",
```
m-g-sonar, line 2: There is a suspicious extra space at the end of the title. You might want to check on Jira.

java-checks/src/test/java/org/sonar/java/checks/VisibleForTestingUsageCheckTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.sonar.java.checks.verifier.JavaCheckVerifier;
+import org.sonar.java.testing.FilesUtils;
+
+import static org.sonar.java.CheckTestUtils.testSourcesPath;
+
+class VisibleForTestingUsageCheckTest {
+  @Test
+  void test() {
+    List<File> classPath = FilesUtils.getClassPath(FilesUtils.DEFAULT_TEST_JARS_DIRECTORY);
+    classPath.add(new File("../java-checks-test-sources/target/classes"));
```
m-g-sonar, line 34: Why not be even more precise and add the folder which will contain directly the classes we are searching for?

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5803_java.html
```diff
@@ -0,0 +1,30 @@
+<p><code>@VisibleForTesting</code>&nbsp;can be used to mark methods and fields whose visibility restrictions have been relaxed more than necessary for
```
m-g-sonar, line 1: `&nbsp;` is suspicious here. Also, RSPEC mentions classes, while here we only have `methods and fields`. I guess you should re-generate the  metadada.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5803_java.html
```diff
@@ -0,0 +1,30 @@
...
+the API to allow for easier unit testing.</p>
+<p>Access to such methods and fields only possible thanks to this relaxed visibility is fine for test code, but it should be avoided in production
+code.</p>
+<p>Supported framework:</p>
+<p> * Guava (com.google.common.annotations.VisibleForTestin)</p>
```
m-g-sonar, line 6: missing `g` at the end :)
Also, I fee that it would be nice to mark the fully qualified names with `<code>` blocks

java-checks/src/main/java/org/sonar/java/checks/VisibleForTestingUsageCheck.java
```diff
@@ -0,0 +1,68 @@
...
+    if (isFieldMethodOrClass(symbol) && !inTheSameFile
+      && (ANNOTATIONS.stream().anyMatch(metadata::isAnnotatedWith))) {
+      reportIssue(identifier, String.format("Remove this usage of \"%s\", it is annotated with @VisibleForTesting and should not be accessed from production code.",
+        identifier.name()));
+    }
+  }
+
+  private static boolean isFieldMethodOrClass(Symbol symbol) {
+    Symbol owner = symbol.owner();
+    return ((owner != null) && owner.isTypeSymbol()) || symbol.isTypeSymbol();
```
m-g-sonar, line 66: Here is the only missing condition in term of coverage (`owner == null` probably).
I feel that we might be missing the point. For me, the only way to have a missing owner is to have a symbol not correctly resolved, and, therefore, marked as being unknown (package symbols can not occur here). I would consequently only test like this:

```java
return !symbol.isUnknown() && (symbol.isTypeSymbol() || symbol.owner().isTypeSymbol());
```

Now, in order to cover it, simply create a test file under `java-checks-test-sources/src/main/files/non-compiling/checks`, and then add a test such as:

```java
  @Test
  void test_non_compiling() {
      .onFile(nonCompilingTestSourcesPath("checks/VisibleForTestingUsageCheck.java"))
      .withCheck(new VisibleForTestingUsageCheck())
      .verifyIssues();
  }
```

And as a test code something like:

```java
class VisibleForTestingUsageCheck {
  @com.google.common.annotations.VisibleForTesting 
  void bar() { }

  void foo() {
    unknown(); // Compliant
  }
}

class VisibleForTestingUsageCheck2 {
  void qix() {
    new VisibleForTestingUsageCheck()
      .bar(); // Noncompliant
  }
}
```

java-checks/src/main/java/org/sonar/java/checks/VisibleForTestingUsageCheck.java
```diff
@@ -0,0 +1,68 @@
...
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.IDENTIFIER);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    IdentifierTree identifier = (IdentifierTree) tree;
+    Symbol symbol = identifier.symbol();
+    SymbolMetadata metadata = symbol.metadata();
```
m-g-sonar, line 51: you can probably already return here if the symbol is unknown.

java-checks/src/main/java/org/sonar/java/checks/VisibleForTestingUsageCheck.java
```diff
@@ -0,0 +1,68 @@
...
+    if (metadata.annotations().isEmpty()) {
+      return;
+    }
+
+    boolean inTheSameFile = symbol.declaration() != null;
+    if (isFieldMethodOrClass(symbol) && !inTheSameFile
```
m-g-sonar, line 57: in fact, `isFieldMethodOrClass()` is probably simply `!symbol.isUnknown() && !symbol.isPackageSymbol()`.
By construction, it will then only be a method, a variable or a type... As we don't have other kinds of symbol.

java-checks/src/main/java/org/sonar/java/checks/VisibleForTestingUsageCheck.java
```diff
@@ -0,0 +1,68 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.IDENTIFIER);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    IdentifierTree identifier = (IdentifierTree) tree;
+    Symbol symbol = identifier.symbol();
+    SymbolMetadata metadata = symbol.metadata();
+    if (metadata.annotations().isEmpty()) {
```
m-g-sonar, line 52: I think I would move this below, after we are sure it's a valid symbol to be considered

its/ruling/src/test/resources/guava/java-S5803.json
```diff
@@ -0,0 +1,10 @@
+{
+'com.google.guava:guava:src/com/google/common/io/Closer.java':[
+252,
```
m-g-sonar, line 3: This one looks like a FP to me (while a bit borderline), we are inside a class itself annotated with `@VisibleForTesting`, so it feels to me that it's OK to do call a something which is also annotated with `VisibleForTesting`. What's your opinion of it?

m-g-sonar (changes requested):

java-checks/src/main/java/org/sonar/java/checks/VisibleForTestingUsageCheck.java
alban-auzeill: @m-g-sonar for methods and classes it's easy, we can test them with symbol.isMethodSymbol() for symbol.isTypeSymbol().
But for fields, symbol.isVariableSymbol() is not enough. We want to exclude local variables, method parameters, and lambda parameters. I think the owner of a local variable is the method declaration. What is the best way to test if a symbol is a class field?

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5803_java.json
margarita-nedzelska-sonarsource: In the ticket title in Jira there is a space at the end. Will fix on both sides

its/ruling/src/test/resources/guava/java-S5803.json
margarita-nedzelska-sonarsource: I think we should exclude those members, is the owner is annotated @VisibleForTesting too

margarita-nedzelska-sonarsource: I thought about it and I think we should report this issue. The 'Logger', which is used inside 'Closer' and 'Closer' are in different files, so if 'Logger' is marked as private, it is not visible in 'Closer'

m-g-sonar: Fine by me!

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5803_java.html
```diff
@@ -0,0 +1,43 @@
+<p>@VisibleForTesting can be used to mark methods, fields and classes whose visibility restrictions have been relaxed more than necessary for the API
+to allow for easier unit testing.</p>
+<p>Access to such methods, fields and classes only possible thanks to this relaxed visibility is fine for test code, but it should be avoided in
+production code.</p>
+<p>Supported framework:</p>
+<p> * Guava</p>
+<pre>
```
m-g-sonar, line 7: Wops, it seems to me that this HTML file is completely broken. did you double-check how it looks in SonarQube?

m-g-sonar (changes requested):
Except of the html description, it LGTM!

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5803_java.html
margarita-nedzelska-sonarsource: looking into it. Looks like need to update the Jira and regenerate

margarita-nedzelska-sonarsource: Updated

m-g-sonar (approved):

PR #3171, opened 2020-09-07
Add jetty project to the its/ruling

alban-auzeill:


its/ruling/src/test/java/org/sonar/java/it/JavaRulingTest.java
```diff
@@ -164,6 +164,18 @@ public void apache_commons_beanutils() throws Exception {
     executeBuildWithCommonProperties(build, projectName);
   }
 
+  @Test
+  public void eclipse_jetty() throws Exception {
+    String projectName = "eclipse-jetty";
+    MavenBuild build = test_project("org.eclipse.jetty:jetty-project", projectName)
+      .setProperty("sonar.exclusions", "" +
```
margarita-nedzelska-sonarsource, line 171: I think we can remove this empty string from here

margarita-nedzelska-sonarsource (approved):
Overall looks good to me!

PR #3172, opened 2020-09-09
SONARJAVA-3509 RSPEC-5958 AssertJ "assertThatThrownBy" should not be used alone

m-g-sonar:


margarita-nedzelska-sonarsource:
Looks Good to me.

margarita-nedzelska-sonarsource (approved):
I was thinking if it is possible to name the check differently, cause word "Alone" might be confusing. But I didn't come up with anything better than you already have. So everything is good.

PR #3173, opened 2020-09-11
SONARJAVA-3511 Rule S5961: Test methods should not contain too many assertions

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/helpers/UnitTestUtils.java
```diff
@@ -34,6 +36,38 @@
...
+  public static final Pattern TEST_METHODS_PATTERN = Pattern.compile("test.*|.*Test");
+
+  public static final MethodMatchers ASSERTION_INVOCATION_MATCHERS = MethodMatchers.or(
+    // fest 1.x / 2.X
+    MethodMatchers.create().ofSubTypes("org.fest.assertions.GenericAssert", "org.fest.assertions.api.AbstractAssert").anyName().withAnyParameters().build(),
+    // rest assured 2.0
+    MethodMatchers.create().ofTypes("io.restassured.response.ValidatableResponseOptions")
+      .name(name -> name.equals("body") ||
+        name.equals("time") ||
+        name.startsWith("time") ||
```
margarita-nedzelska-sonarsource, line 49: This case is not covered in tests because as I found there is only one method starting with "time", which is time itself and it is true for `name.equals("time")` if I remove this line, could hit performance a little(very little), I wanted to remove `startsWith(..)` but maybe I am missing something

m-g-sonar: It might be related to old version of restassured API, and so it would make us support them all, to be confirmed. For me it's OK to let it like this.

java-checks/src/main/java/org/sonar/java/checks/tests/TooManyAssertionsCheck.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.sonar.java.checks.helpers.UnitTestUtils.ASSERTION_METHODS_PATTERN;
+import static org.sonar.java.checks.helpers.UnitTestUtils.REACTIVE_X_TEST_METHODS;
+import static org.sonar.java.checks.helpers.UnitTestUtils.TEST_METHODS_PATTERN;
+import static org.sonar.java.checks.helpers.UnitTestUtils.isUnitTest;
+import static org.sonar.java.model.ExpressionUtils.methodName;
+
+@Rule(key = "S5961")
+public class TooManyAssertionsCheck extends BaseTreeVisitor implements JavaFileScanner {
+
+  private static final int DEFAULT_MAX = 10;
```
m-g-sonar, line 48: If the default 10 is too noisy, I would adjust it to a more appropriate number by default, to have a rule "already configured for most cases" when it is enabled. 50 looks like too far away from the default to me.
Could you have a look at the distribution of values on our test projects? Then the boundary could be set to catch 5-10% of them.

java-checks/src/main/java/org/sonar/java/checks/tests/AssertionsInTestsCheck.java
```diff
@@ -172,23 +173,6 @@ private static boolean expectAssertion(MethodTree methodTree) {
     return false;
   }
 
-  private static boolean isUnitTest(MethodTree methodTree) {
```
m-g-sonar, line 175: shouldn't you also remove the `ASSERTION_INVOCATION_MATCHERS` and `REACTIVE_X_TEST_METHODS` constants from this file, since you duplicated it in `UnitTestUtils`?

java-checks/src/main/java/org/sonar/java/checks/tests/TooManyAssertionsCheck.java
```diff
@@ -0,0 +1,123 @@
...
+
+    if (isUnitTest(methodTree) && numberOfAssertions(methodTree.symbol()) > maximumAssertionNumber) {
+      context.reportIssue(this, methodTree.simpleName(), String.format("Refactor this method in order to have less than %d assertions.", maximumAssertionNumber));
+    }
+  }
+
+  private int numberOfAssertions(Symbol symbol) {
+    if (!assertionsInMethod.containsKey(symbol)) {
+      assertionsInMethod.put(symbol, 0);
+      Tree declaration = symbol.declaration();
```
m-g-sonar, line 78: you already have the declaration in the `visitMethod()` method (L70), why not directly use it instead of the symbol?
So you would not have to jump into the declaration. You can these the symbol inside this method the.

java-checks/src/main/java/org/sonar/java/checks/tests/TooManyAssertionsCheck.java
```diff
@@ -0,0 +1,123 @@
...
+
+import static org.sonar.java.checks.helpers.UnitTestUtils.ASSERTION_INVOCATION_MATCHERS;
+import static org.sonar.java.checks.helpers.UnitTestUtils.ASSERTION_METHODS_PATTERN;
+import static org.sonar.java.checks.helpers.UnitTestUtils.REACTIVE_X_TEST_METHODS;
+import static org.sonar.java.checks.helpers.UnitTestUtils.TEST_METHODS_PATTERN;
+import static org.sonar.java.checks.helpers.UnitTestUtils.isUnitTest;
+import static org.sonar.java.model.ExpressionUtils.methodName;
+
+@Rule(key = "S5961")
+public class TooManyAssertionsCheck extends BaseTreeVisitor implements JavaFileScanner {
```
m-g-sonar, line 46: I don't see the interest of using a BaseTreeVisitor if you are only visiting methods. Use an `IssuableSubscriptionVisitor`, it will be simpler.

java-checks/src/main/java/org/sonar/java/checks/tests/TooManyAssertionsCheck.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  @Override
+  public void visitMethod(MethodTree methodTree) {
+    if (ModifiersUtils.hasModifier(methodTree.modifiers(), Modifier.ABSTRACT)) {
+      return;
+    }
+
+    if (isUnitTest(methodTree) && numberOfAssertions(methodTree.symbol()) > maximumAssertionNumber) {
+      context.reportIssue(this, methodTree.simpleName(), String.format("Refactor this method in order to have less than %d assertions.", maximumAssertionNumber));
```
m-g-sonar, line 71: 2 comments:
* With an `IssuableSubscriptionVisitor`, you could directly call `reportIssue(methodTree.simpleName())`

* I think it would be valuable to use secondaries on all the assertions, in order to proof our count.

java-checks/src/main/java/org/sonar/java/checks/tests/TooManyAssertionsCheck.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.sonar.java.checks.helpers.UnitTestUtils.isUnitTest;
+import static org.sonar.java.model.ExpressionUtils.methodName;
+
+@Rule(key = "S5961")
+public class TooManyAssertionsCheck extends BaseTreeVisitor implements JavaFileScanner {
+
+  private static final int DEFAULT_MAX = 10;
+
+  @RuleProperty(key = "MaximumAssertionNumber", description = "The maximum authorized number of assertions in a test method", defaultValue = "" + DEFAULT_MAX)
+  public int maximumAssertionNumber = DEFAULT_MAX;
```
m-g-sonar, line 51: the field name sounds complicated for a so simple purpose. Why not calling it `maximum` or `threshold` only?

m-g-sonar (changes requested):

java-checks/src/main/java/org/sonar/java/checks/tests/TooManyAssertionsCheck.java
margarita-nedzelska-sonarsource: I think we need declaration in case we want to detect a local method which contains assertions. 
`  @Test
  void test3() { // Noncompliant [[sc=8;ec=13]]{{Refactor this method in order to have less than 2 assertions.}}
    assertEquals(2, f(2));
    assertEquals(3, f(1));
    customAssert();
  }

  void customAssert() {
    assertEquals(2, f(2));
    assertEquals(3, f(1));
  }
`

In this case `tree` points to invocation of `customAssert()` while to understand whether this is an assert itself we need its declaration.

java-checks/src/main/java/org/sonar/java/checks/tests/TooManyAssertionsCheck.java
margarita-nedzelska-sonarsource: Sounds like a good idea. I checked in sonarqube locally. So the majority of issues contain <20 assertions. If we want to be even less noisy I'll suggest to put 25.

Wdyt, 20 or 25?

m-g-sonar (approved):

PR #3178, opened 2020-09-14
SONARJAVA-3524 Rule S5973: Tests should be stable

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/TestsStabilityCheck.java
```diff
@@ -0,0 +1,68 @@
...
+    return Arrays.asList(Tree.Kind.METHOD, Tree.Kind.CONSTRUCTOR);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodTree methodTree = (MethodTree) tree;
+
+    List<AnnotationTree> annotations = methodTree.modifiers().annotations();
+    Optional<Arguments> arguments = annotations.stream()
+      .filter(annotationTree -> annotationTree.symbolType().fullyQualifiedName().equals(ANNOTATION))
```
alban-auzeill, line 52: Can be replaced by 
`.filter(annotationTree -> annotationTree.symbolType().is(ANNOTATION))`

alban-auzeill (approved):

PR #3179, opened 2020-09-15
SONARJAVA-3527 Rule S5977: Tests should use fixed data instead of ran…

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/RandomizedTestDataCheck.java
```diff
@@ -0,0 +1,103 @@
...
+      reportedUUIDRandom = true;
+      List<JavaFileScannerContext.Location> locations = symbol.usages().stream()
+        .map(identifierTree -> new JavaFileScannerContext.Location(LOCATIONS_TEXT, identifierTree))
+        .collect(Collectors.toList());
+      reportIssue(tree, MESSAGE, locations, null);
+    }
+  }
+
+  @Override
+  public void leaveFile(JavaFileScannerContext context) {
```
quentin-jaquier-sonarsource, line 92: I tend to agree that reporting only one issue per test file is enough, we should just make sure that the [rspec](https://jira.sonarsource.com/browse/RSPEC-5977) description reflects this behavior (current Highlighting part mention that one issue per method will be reported).

java-checks/src/main/java/org/sonar/java/checks/tests/RandomizedTestDataCheck.java
```diff
@@ -0,0 +1,103 @@
...
+    if (tree.is(Tree.Kind.NEW_CLASS)) {
+      checkForRandomConstructorUsage(tree);
+    }
+    if (!reportedUUIDRandom && tree.is(Tree.Kind.METHOD_INVOCATION)) {
+      checkAndReportUUIDRandomUsage(tree);
+    }
+  }
+
+  private void checkForRandomConstructorUsage(Tree tree) {
+    NewClassTree newClassTree = (NewClassTree) tree;
```
quentin-jaquier-sonarsource, line 73: If you call this method naively with any `Tree`, the cast is not safe.
I would rather use `NewClassTree` as parameter's type and perform the cast when you call it.
Same for `checkAndReportUUIDRandomUsage`.

java-checks/src/main/java/org/sonar/java/checks/tests/RandomizedTestDataCheck.java
```diff
@@ -0,0 +1,103 @@
...
+import org.sonar.plugins.java.api.tree.NewClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S5977")
+public class RandomizedTestDataCheck extends IssuableSubscriptionVisitor {
+
+  private static final MethodMatchers RANDOM_CONSTRUCTOR_METHOD_MATCHER = MethodMatchers.create()
+    .ofTypes("java.util.Random")
+    .constructor()
+    .withAnyParameters()
```
quentin-jaquier-sonarsource, line 41: It means that we will report an issue for Random with seed.
Not so unpredictable, but using hardcoded value is still more readable, so it seems correct to me.
We should just add a test for this case, to show that we thought about this.

java-checks/src/main/java/org/sonar/java/checks/tests/RandomizedTestDataCheck.java
```diff
@@ -0,0 +1,103 @@
...
+      .build();
+
+  private static final MethodMatchers RANDOM_UUID_MATCHER = MethodMatchers.create()
+    .ofTypes("java.util.UUID")
+    .names("randomUUID")
+    .withAnyParameters()
```
quentin-jaquier-sonarsource, line 47: Can it actually take parameters? If not, we should consider using `addWithoutParametersMatcher()`.

quentin-jaquier-sonarsource (changes requested):

java-checks-test-sources/src/main/java/checks/NonRandomizedTestDataCheck.java
```diff
@@ -0,0 +1,35 @@
+package checks;
```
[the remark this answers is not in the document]
margarita-nedzelska-sonarsource: This was the intention to check the file without any issues. Because otherwise this case is not covered.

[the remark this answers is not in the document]
quentin-jaquier-sonarsource: It makes sense indeed.

java-checks-test-sources/src/main/java/checks/RandomizedTestDataCheck.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Random;
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+
+public class RandomizedTestDataCheck {
+
+  @Test
+  public void randomizedTest() {
+    int userAge = new Random().nextInt(42);  // Noncompliant[[sc=19;ec=31;secondary=11,25,27,29,31,33,42]]{{Replace randomly generated values with fixed ones.}}
+    UUID userID = UUID.randomUUID(); // Noncompliant[[sc=19;ec=36;secondary=12,26,28,30,32,34]]{{Replace randomly generated values with fixed ones.}}
```
quentin-jaquier-sonarsource, line 12: I don't think it makes sense to include the main location in the secondaries.

margarita-nedzelska-sonarsource: On the one hand yes.

What I thought about secondary locations is that there are different cases, If we report the main issue on a different level (example class) and highlight secondaries (example method) we do not report the primary one in a secondary.

When we're talking about this issue. Here we report the first occurrence and highlight other usages. So at this point this makes sense to add the first one to secondary locations. And moreover, this way we have a fair counter of all usages.

Wdyt?

quentin-jaquier-sonarsource (approved):
LGTM

PR #3180, opened 2020-09-15
SONARJAVA-3515 Rule S5960: Assertions should not be used in production code

alban-auzeill:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3182, opened 2020-09-16
SONARJAVA-3526 Rule S5976: Similar tests should be grouped in a single Parameterized test

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/ParameterizedTestCheck.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  static class CollectAndFilter implements BiPredicate<JavaTree, JavaTree> {
+
+    Set<JavaTree> nodeToParametrize = new HashSet<>();
+    private final Set<JavaTree> currentNodeToParameterize = new HashSet<>();
+
+    @Override
+    public boolean test(JavaTree leftNode, JavaTree rightNode) {
+      if (isLiteral(leftNode) && isLiteral(rightNode)) {
```
alban-auzeill, line 137: I disagree, the logic should be careful with the semantic kind of the literals, let's take this example:
```
class FooTest {
  @Test
  void test1() { // FP "compute(1)" and "compute(2.0f)" is not related to the same method
    int a = compute(1);
    assertEquals(a, 0);
  }

  @Test
  void test2() {
    int a = compute(2.0f);
    assertEquals(a, 0);
  }

  @Test
  void test3() {
    int a = compute(3);
    assertEquals(a, 0);
  }

  int compute(int a) { return 0;  }
  int compute(float a) { return 0; }
}
``` 
So, when the left and right literals do not have the same types, we should check if a parameter could have a compatible type that will really keep the same test semantic.

java-checks/src/main/java/org/sonar/java/checks/tests/ParameterizedTestCheck.java
```diff
@@ -0,0 +1,160 @@
...
+      .map(MethodTree.class::cast)
+      .filter(ParameterizedTestCheck::isParametrizedCandidate)
+      .collect(Collectors.toList());
+    if (methods.size() < MIN_SIMILAR_METHODS) {
+      return;
+    }
+
+    Set<MethodTree> handled = new HashSet<>();
+    for (int i = 0; i < methods.size(); i++) {
+      MethodTree method = methods.get(i);
```
alban-auzeill, line 79: I would add:
```
      if (handled.contains(method)) {
        continue;
      }
```

java-checks/src/main/java/org/sonar/java/checks/tests/ParameterizedTestCheck.java
```diff
@@ -0,0 +1,160 @@
...
+      List<MethodTree> equivalentMethods = methods.stream()
+        .skip(i + 1L)
+        // avoid reporting multiple times
+        .filter(otherMethod -> !handled.contains(otherMethod))
+        // only consider method syntactically equivalent, ignoring literals
+        .filter(otherMethod -> {
+          boolean areEquivalent = SyntacticEquivalence.areEquivalent(methodBody, otherMethod.block().body(), collectAndFilter);
+          if (areEquivalent) {
+            // If methods where not equivalent, we don't want to pollute the set of node to parameterize.
+            collectAndFilter.validateCollection();
```
alban-auzeill, line 94: I don't understand how this code can accidentally work.
- If a method is equivalent, we copy `currentNodeToParameterize` into `nodeToParametrize`, without cleaning `currentNodeToParameterize`, it's not optimum
- If a method is not equivalent, we do nothing! we keep `currentNodeToParameterize` dirty and it's content will be copied into `nodeToParametrize` by the following equivalent method :confounded:

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/tests/ParameterizedTestCheck.java
```diff
@@ -0,0 +1,160 @@
...
+          }
+          return areEquivalent;
+        })
+        .collect(Collectors.toList());
+
+      if (equivalentMethods.size() + 1 >= MIN_SIMILAR_METHODS) {
+        handled.add(method);
```
alban-auzeill, line 101: We probably don't need to add `method` itself to `handled` because the two loops will only look after `method` in the `methods` list.

java-frontend/src/main/java/org/sonar/java/model/SyntacticEquivalence.java
```diff
@@ -37,33 +37,53 @@ private SyntacticEquivalence() {
    * @return true, if nodes are syntactically equivalent
    */
   public static boolean areEquivalent(List<? extends Tree> leftList, List<? extends Tree> rightList) {
+    return areEquivalent(leftList, rightList, (t1, t2) -> false);
```
alban-auzeill, line 40: I don't like the bellow variable name `equivalenceFilter`, I expect a filter to reduce the result and not to extends it. Here we pass `(t1, t2) -> false` to a filter, it's confusing, the result should be empty if it was a filter.
Why not calling it: `tolerantEquivalence` or `permissiveEquivalence`?

java-checks/src/main/java/org/sonar/java/checks/tests/ParameterizedTestCheck.java
quentin-jaquier-sonarsource: Make sense, and in fact, even when the issue is correct (relate to the same method taking Object for example), using parameterized tests would not result in something super clear, so let's consider only literals of the same type.

java-frontend/src/main/java/org/sonar/java/model/SyntacticEquivalence.java
```diff
@@ -37,33 +37,54 @@ private SyntacticEquivalence() {
...
+  }
+
+  /**
+   * @return true, if nodes are syntactically equivalent
+   * Use permissiveEquivalence to force the equivalence of two nodes
+   * Use permissiveEquivalence to force the equivalence of two nodes
```
alban-auzeill, line 46: duplicated line

PR #3183, opened 2020-09-16
SONARJAVA-3530 Fix FPs in S3577 caused by abstract classes and update supported class names + SONARJAVA-3520 Make S3577 part of SonarWay

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/naming/BadTestClassNameCheck.java
```diff
@@ -38,7 +38,9 @@
 @Rule(key = "S3577")
 public class BadTestClassNameCheck extends IssuableSubscriptionVisitor {
 
-  private static final String DEFAULT_FORMAT = "^((Test|IT)[a-zA-Z0-9]+|[A-Z][a-zA-Z0-9]*(Test|Tests|TestCase|IT|ITCase))$";
+  private static final String STARTING_WITH_TEST = "(Test|IT)[a-zA-Z0-9_]+";
```
margarita-nedzelska-sonarsource, line 42: It looks like you can use '\w' in regex which means exactly the same : 
[a-zA-Z_0-9 (https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html)

On the other hand, it might not be that obvious for others, so up to you.

margarita-nedzelska-sonarsource (approved):
Looks good :)

java-checks/src/main/java/org/sonar/java/checks/naming/BadTestClassNameCheck.java
m-g-sonar: In fact, I can not really easily change this, or it will change the message being reported as well. So I'll let it like this, as I don't feel displaying `\w` in the issue message is nice, and replacing it before printing looks a bit too complex for what it is.

PR #3185, opened 2020-09-16
SONARJAVA-3522 Rule S3414: Tests should be kept in a dedicated source directory

margarita-nedzelska-sonarsource:
The main idea is that main sources should not contain any classes with tests.

### DIFF
--- a/java-checks-test-sources/src/main/java/checks/NoTestsInProductionFile.java
+++ b/java-checks-test-sources/src/main/java/checks/NoTestsInProductionFile.java
@@ -1,8 +1,6 @@
 package checks;
 
-import org.junit.jupiter.api.Test;
-
-public class NoTestsInProductionFile { // Noncompliant [[sc=14;ec=35]] {{Move this test class to a separate assembly in this solution.}}
+public class NoTestsInProductionFile { // Compliant
 
   public void not_a_test(){}
 
--- a/java-checks-test-sources/src/main/java/checks/NoTestsInProductionFile.java
+++ b/java-checks-test-sources/src/main/java/checks/NoTestsInProductionFile.java
@@ -1,0 +1,9 @@
+package checks;
+
+import org.junit.jupiter.api.Test;
+
+public class NoTestsInProductionFile { // Noncompliant [[sc=14;ec=35]] {{Move this test class to a separate assembly in this solution.}}
+
+  public void not_a_test(){}
+
+}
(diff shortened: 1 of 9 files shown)


java-checks-test-sources/src/main/java/checks/NoTestsInProductionFile.java
```diff
@@ -0,0 +1,9 @@
+package checks;
+
+import org.junit.jupiter.api.Test;
+
+public class NoTestsInProductionFile { // Noncompliant [[sc=14;ec=35]] {{Move this test class to a separate assembly in this solution.}}
```
m-g-sonar, line 5: Remove this `Noncompliant` message, and replace it by `compliant`, as we do not expect any issue here. I guess it's a copy-paste issue, but this is misleading.

java-checks-test-sources/src/main/java/checks/TestsInProductionFile.java
```diff
@@ -0,0 +1,10 @@
+package checks;
+
+import org.junit.jupiter.api.Test;
+
+public class TestsInProductionFile { // Noncompliant [[sc=14;ec=35]] {{Move this test class to a separate assembly in this solution.}}
```
m-g-sonar, line 5: The issue message seems to come from C# world: `Move this test class to a separate assembly in this solution.`. I think we should have a different one for java, as this vocabulary does not apply to our context.

java-checks/src/main/java/org/sonar/java/checks/TestsInSeparateFolderCheck.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+    if (isTestClass(classTree)) {
+      reportIssue(classTree.simpleName(), "Move this test class to a separate assembly in this solution.");
```
m-g-sonar, line 43: The message does not seem appropriate for java projects.

m-g-sonar (changes requested):

java-checks/src/main/java/org/sonar/java/checks/TestsInSeparateFolderCheck.java
margarita-nedzelska-sonarsource: Maybe "Move this test class to the test folder in current project." will sound better?

m-g-sonar (approved):

PR #3186, opened 2020-09-16
SONARJAVA-3516 Rule S5969: Mocking all non-private methods of a class…

sebastian-hungerecker-sonarsource:
… should be avoided


quentin-jaquier-sonarsource (approved):
LGTM, nice work.

PR #3187, opened 2020-09-16
Fix FP for Serializable with initializer

margarita-nedzelska-sonarsource:
I checked The main file without semantic. Almost all the fields were reported correctly. The only thing is annotations `@Inject` and `@EJB`. As far as they are not found without semantic, fields are reported as issues. So we have false positive.

Is it ok? or it's better o exclude analysis without semantic at all?

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
@@ -19,6 +19,9 @@
  */
 package org.sonar.java.checks.serialization;
 
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.Nullable;
 import org.sonar.check.Rule;
 import org.sonar.java.model.JUtils;
 import org.sonar.java.model.ModifiersUtils;
@@ -37,10 +40,6 @@ import org.sonar.plugins.java.api.tree.TypeTree;
 import org.sonar.plugins.java.api.tree.VariableTree;
 import org.sonar.plugins.java.api.tree.WildcardTree;
 
-import javax.annotation.Nullable;
-import java.util.Collections;
-import java.util.List;
-
 @Rule(key = "S1948")
 public class SerializableFieldInSerializableClassCheck extends IssuableSubscriptionVisitor {
 
@@ -51,9 +50,6 @@ public class SerializableFieldInSerializableClassCheck extends IssuableSubscript
 
   @Override
   public void visitNode(Tree tree) {
-    if(!hasSemantic()) {
-      return;
-    }
     ClassTree classTree = (ClassTree) tree;
     if (isSerializable(classTree)
       && !SerializableContract.hasSpecialHandlingSerializationMethods(classTree)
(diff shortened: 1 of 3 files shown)


java-checks/src/test/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheckTest.java
```diff
@@ -39,4 +39,13 @@ void test() {
       .verifyIssues();
   }
 
+  @Test
+  void testWithoutSemantic() {
+    JavaCheckVerifier.newVerifier()
+      .onFile(nonCompilingTestSourcesPath("checks/serialization/SerializableFieldInSerializableClassCheck.java"))
```
m-g-sonar, line 45: To answer your question, I believe we should not report Issues for lines with `@EJB` or `@Inject` if bytecode is incomplete.

There is however an easy way to cover this, we simply have to check if there is any unknown annotation. To do so, I would simply modify method `isTransientSerializableOrInjected` as follow:

```java
private static boolean isTransientSerializableOrInjected(VariableTree member) {
    if (ModifiersUtils.hasModifier(member.modifiers(), Modifier.TRANSIENT) || (isSerializable(member.type()) && !isSubtypeOfCollectionApi(member.type().symbolType()))) {
      return true;
    }
    return metadata.isAnnotatedWith("javax.inject.Inject")
      || metadata.isAnnotatedWith("javax.ejb.EJB")
      // we can not guarantee anything due to the presence of an unknown annotation
      || metadata.annotations().stream().anyMatch(annotationInstance -> annotationInstance.symbol().isUnknown());
  }
```

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -94,7 +95,7 @@ private static boolean isUnserializableCollection(Type type) {
   private void checkCollectionAssignments(List<IdentifierTree> usages) {
     for (IdentifierTree usage : usages) {
       Tree parentTree = usage.parent();
-      if (parentTree.is(Tree.Kind.ASSIGNMENT)) {
+      if (Objects.requireNonNull(parentTree).is(Tree.Kind.ASSIGNMENT)) {
```
m-g-sonar, line 98: This is not a required change. Is it IntelliJ who suggests you to do it?

On this, We (as SonarSource) are not aligned with their definition of nullness. `Tree.Parent()` is annotated with `@Nullable`, which is a *weak* nullness annotation, contrarily to `@CheckForNull`. Here, it's the responsibility of the developer calling the method to know when it will be null or not, and it usually depends on very precise conditions, and it can not happen here. Our NPE detection rule is not raising any issue, and I believe we are right.

Could you revert the change?

m-g-sonar (changes requested):

java-checks/src/test/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheckTest.java
margarita-nedzelska-sonarsource: Thanks. That helped me a lot :)

m-g-sonar (approved):

PR #3190, opened 2020-09-17
SONARJAVA-3477 Make S1214 report issues only when interfaces contain only constants

m-g-sonar:


java-checks-test-sources/src/main/java/checks/InterfaceAsConstantContainerCheck.java
```diff
@@ -3,52 +3,45 @@
 interface InterfaceAsConstantContainerCheck {
 }
 
-interface InterfaceAsConstantContainerCheckB {
-}
-
-interface InterfaceAsConstantContainerCheckC { // Noncompliant [[sc=11;ec=45;secondary=10,11]] {{Move constants defined in this interfaces to another class or enum.}}
+interface InterfaceAsConstantContainerCheckA { // Noncompliant [[sc=11;ec=45;secondary=7,8]] {{Move constants defined in this interfaces to another class or enum.}}
   int a = 0;
   int b = 0;
```
quentin-jaquier-sonarsource, line 8: The official quote is:
> “an interface which contains no methods, it consists solely of **static fields,** each exporting a constant.”

To me, static can be relaxed, and not only static field should be considered.
In any way, we should add a test with static fields.

quentin-jaquier-sonarsource (changes requested):
We could update the description to clearly explain when do we raise an issue:

> This rule raises an issue when an interface consists solely of fields, without any other members.


In addition, please have a look at my comment on the ticket and let me know what you think.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3191, opened 2020-09-17
SONARJAVA-3492 S1215 should detect "System.runFinalization()" the sam…

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/GarbageCollectorCalledCheck.java
```diff
@@ -19,18 +19,22 @@
  */
```
quentin-jaquier-sonarsource, line 19: Nice simplification! This is a perfect example to show that some rules are implemented since already quite some years, with the changes and improvement of the plugin, it is now possible to refactor some rule to something way cleaner.

quentin-jaquier-sonarsource (changes requested):
In addition, I think we can also update the description of the rule, to state that `runFinalization` is also something to avoid.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3197, opened 2020-09-21
SONARJAVA-3501 only ignore usued imports on lombok.val and lombok.var

m-g-sonar:
.


alban-auzeill (approved):

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -118,7 +118,7 @@
   public void visitImport(ImportTree tree) {
     String fullyQualifiedName = ExpressionsHelper.concatenate((ExpressionTree) tree.qualifiedIdentifier());
 
-    excludeLinesIfTrue(fullyQualifiedName.startsWith("lombok."), tree, UselessImportCheck.class);
+    excludeLinesIfTrue("lombok.var".equals(fullyQualifiedName) || LOMBOK_VAL.equals(fullyQualifiedName), tree, UselessImportCheck.class);
```
pzygielo, line 121: Strange that `fullyQualifiedName` once is compared with literal and once with constant...

PR #3199, opened 2020-09-21
SONARJAVA-3519 FP on S3878 when the argument before the vararg is also an array

quentin-jaquier-sonarsource:
Fixes SONARJAVA-3519.


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3200, opened 2020-09-21
SONARJAVA-3476 Improve issue location for S5843

sebastian-hungerecker-sonarsource:


java-frontend/src/main/java/org/sonar/java/regex/ast/OpeningQuote.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class OpeningQuote extends RegexSyntaxElement {
+
+  public OpeningQuote(RegexSource source) {
+    super(source, new IndexRange(-1, 0));
+  }
+
+  @Override
+  public String getText() {
+    throw new UnsupportedOperationException("getText should not be called on OpeningQuote objects.");
```
m-g-sonar, line 37: Why not simply return directly `"`? It does not hurt and will not make anything fail in the worst-case scenario!

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
```diff
@@ -31,7 +31,7 @@
  */
 public class RegexSource {
 
-  private final List<LiteralTree> stringLiterals;
+  final List<LiteralTree> stringLiterals;
```
m-g-sonar, line 34: I'm not sure it's good to have its visibility increased. I would prefer to have a getter which would return an unmodifiable list, such as:
```java
List<LiteralTree> stringLiterals() {
  return Collections.unmodifiableList(stringLiterals);
}
```
It would prevent any accidental modification of the list.

PR #3204, opened 2020-09-22
SONARJAVA-3534 FP S3077(VolatileNonPrimitiveFieldCheck) should consider enum as immutable

alban-auzeill:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3206, opened 2020-09-22
SONARJAVA-3491 FN S2789 (NullShouldNotBeUsedWithOptionalCheck) on null assignment

alban-auzeill:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3208, opened 2020-09-22
Fix too many assertions issue

margarita-nedzelska-sonarsource:
I just found this issue in Sonar java. And have a proposal how to make it more readable.
What do you think about this solution? Doest it look more readable for you?

### DIFF
--- a/external-reports/src/test/java/org/sonar/java/externalreport/PmdSensorTest.java
+++ b/external-reports/src/test/java/org/sonar/java/externalreport/PmdSensorTest.java
@@ -54,7 +54,7 @@ class PmdSensorTest {
   private static final Path PROJECT_DIR = Paths.get("src", "test", "resources", "pmd");
   private static final String PROJECT_ID = "pmd-test";
 
-  private static PmdSensor sensor = new PmdSensor();
+  private static final PmdSensor sensor = new PmdSensor();
 
   @Rule
   public LogTester logTester = new LogTester();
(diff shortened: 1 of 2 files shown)


quentin-jaquier-sonarsource:
My main concern is that you will still execute the same number of assertions, they are just hidden behind method calls. This will remove the issue, but in the end, I feel that the problem originally reported is still present (test can fail for too many reasons).

external-reports/src/test/java/org/sonar/java/externalreport/PmdSensorTest.java
```diff
@@ -218,3 +223,66 @@ private static String language(Path file) {
     return path.substring(path.lastIndexOf('.') + 1);
   }
 }
+
+class ExternalIssueAsserter {
```
m-g-sonar, line 227: Just a small comment about this: I'm not a big fan of having two classes at the same level in a compilation unit, just one being not public. I would prefer to make it a public static method inside the other one, or a dedicated class in its own compilation unit.

Also, what about making it inherit from AssertJ's `AbstractAssert`?
https://joel-costigliola.github.io/assertj/assertj-core-custom-assertions.html

margarita-nedzelska-sonarsource: I thought about AssertJ approach. Unfortunately, I am not sure we'll benefit much from it. Cause we still have to implement all needed assertions logic.

I also changed the logic of assertions, so there is only one point to get an assertion error and we'll have all the error messages collected.

m-g-sonar: Fine by me :)

I just believe that the more we stay closer from how AssertJ operate, the higher is the chance to have it reused somewhere in our own tests.

PR #3212, opened 2020-09-24
SONARJAVA-3414 Rule S4790: its content should be replaced by S2070

margarita-nedzelska-sonarsource:
.


margarita-nedzelska-sonarsource:
I I understood everything correctly, I just need to change the rule key of Rule S2070 to be S4790 and get rid of the original S4790. Correct me if I am wrong.

java-checks/src/main/java/org/sonar/java/checks/CheckList.java
```diff
@@ -345,7 +344,6 @@ private CheckList() {
       CustomUnclosedResourcesCheck.class,
       DanglingElseStatementsCheck.class,
       DataEncryptionCheck.class,
-      DataHashingCheck.class,
```
quentin-jaquier-sonarsource, line 348: As discussed this morning, we can not drop this rule directly, issues currently raised should not disappear, we should go through a deprecation period. 

At the end:

- S4790 implementation will change (and it happens to be matching closely S2070), to report fewer issues. And it's perfectly fine, it's like any improvement of a rule.
- S2070 will be deprecated. The implementation will be closely related to the new S4790, and it's not a problem either.

@m-g-sonar Does this solution looks good to you? You mention the possibility to use @DeprecatedRuleKey, but I'm unsure to see what it can improve.

quentin-jaquier-sonarsource (changes requested):
Don't forget to update the meta-data of the two rules.

Second commit LGTM.

java-checks/src/main/java/org/sonar/java/checks/DeprecatedHashAlgorithmCheck.java
```diff
@@ -39,168 +35,26 @@
 import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
-import static org.sonar.java.checks.DeprecatedHashAlgorithmCheck.InsecureAlgorithm.MD2;
-import static org.sonar.java.checks.DeprecatedHashAlgorithmCheck.InsecureAlgorithm.MD5;
-import static org.sonar.java.checks.DeprecatedHashAlgorithmCheck.InsecureAlgorithm.SHA1;
-import static org.sonar.plugins.java.api.semantic.MethodMatchers.ANY;
```
quentin-jaquier-sonarsource, line 45: Importing statically in one file and using a qualified name in another feels like cheating the duplication. In my opinion, we should either accept it or organize the code to really avoid duplication. 
In this case, isn't it possible to create an abstract class instead of a helper, and having two rules inheriting from it?

quentin-jaquier-sonarsource (changes requested):
In addition, don't forget to update the rule description.

java-checks/src/main/java/org/sonar/java/checks/DeprecatedHashAlgorithmCheck.java
margarita-nedzelska-sonarsource: I thought about it, but in the future we'll remove one of the rule, so we'll have only one class. Actually I can refactor it, so we won't have duplications without this trick

margarita-nedzelska-sonarsource:
Checked with Eric the description is already up-to-date.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3214, opened 2020-09-24
SONARJAVA-3411 Rule S5876 A new session should be created during user authentication

alban-auzeill:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5876_java.json
```diff
@@ -0,0 +1,25 @@
+{
+  "title": "A new session should be created during user authentication",
+  "type": "VULNERABILITY",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
+  },
+  "tags": [
```
m-g-sonar, line 9: Since we are targeting spring, I feel we should have the corresponding tag. I added it to the java-sub-task, but it requires to be re-generated.

PR #3215, opened 2020-09-25
SONARJAVA-3374 Rule S5804 allowing user enumeration is security-sensitive

margarita-nedzelska-sonarsource:
.


java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,123 @@
...
+  private static final String ABSTRACT_USER_DETAILS_AUTHENTICATION_PROVIDER = "org.springframework.security.authentication.dao.AbstractUserDetailsAuthenticationProvider";
+  private static final String USER_DETAILS_SERVICE = "org.springframework.security.core.userdetails.UserDetailsService";
+  private static final String USERNAME_NOT_FOUND_EXCEPTION = "org.springframework.security.core.userdetails.UsernameNotFoundException";
+  private static final String HIDE_USER_NOT_FOUND_EXCEPTIONS = "setHideUserNotFoundExceptions";
+  private static final String LOAD_USER_BY_USERNAME = "loadUserByUsername";
+  private static final String BOOLEAN = "boolean";
+  private static final String STRING = "java.lang.String";
+  private static final String THROWABLE = "java.lang.Throwable";
+
+  private final MethodMatchers setHideUserMatcher = MethodMatchers.create()
```
m-g-sonar, line 49: Why not making it `private static final`. I feel this is a constant. (same comment for the next one as well)

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,123 @@
...
+
+  private void checkHiddenUserNotFoundException(MethodInvocationTree methodInvocationTree, ExpressionTree expression) {
+    if (setHideUserMatcher.matches(methodInvocationTree) && isFalseLiteral(expression)) {
+      reportIssue(methodInvocationTree, MESSAGE);
+    }
+  }
+
+  private void checkThrowUsernameNotFoundException(ThrowStatementTree tree) {
+    if (tree.expression().symbolType().is(USERNAME_NOT_FOUND_EXCEPTION)) {
+      reportIssue(tree, MESSAGE);
```
m-g-sonar, line 102: The `tree` includes the `new` keyword. I feel we can be slightly more precise and report on the expression only.
I also wonder if we want to handle cases where we do not really own the exception, such as:
```java
try { 
 doSomethingThrowingUsernameNotFoundException();
} catch (UsernameNotFoundException e) {
  if (/* whatever test */) {
    doSomething();
  } else {
   throw e; // would report here
  }

}
```

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private void checkHiddenUserNotFoundException(MethodInvocationTree methodInvocationTree, ExpressionTree expression) {
+    if (setHideUserMatcher.matches(methodInvocationTree) && isFalseLiteral(expression)) {
+      reportIssue(methodInvocationTree, MESSAGE);
+    }
+  }
+
+  private void checkThrowUsernameNotFoundException(ThrowStatementTree tree) {
+    if (tree.expression().symbolType().is(USERNAME_NOT_FOUND_EXCEPTION)) {
```
m-g-sonar, line 101: According to the rule description, it should be authorized to throw such an exception if we are within the body of the `loadUserByUsername` method. I don't see check over this case, and it would therefore report FPs.

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private static boolean isFalseLiteral(ExpressionTree expression) {
+    return expression.is(Tree.Kind.BOOLEAN_LITERAL) && !Boolean.parseBoolean(((LiteralTree) expression).value());
+  }
+
+  private static boolean checkParentIsThrowable(Tree tree) {
+    Tree current = tree.parent();
+    while (current instanceof ExpressionTree || current instanceof Arguments) {
+      if (current.is(Tree.Kind.NEW_CLASS)) {
```
m-g-sonar, line 113: this can be simplified a bit to:
```java
if (current.is(Tree.Kind.NEW_CLASS) 
  && ((NewClassTree) current).symbolType().isSubtypeOf(THROWABLE)) {
  return true;
}
```

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  private void checkThrowUsernameNotFoundException(ThrowStatementTree tree) {
+    if (tree.expression().symbolType().is(USERNAME_NOT_FOUND_EXCEPTION)) {
+      reportIssue(tree, MESSAGE);
+    }
+  }
+
+  private static boolean isFalseLiteral(ExpressionTree expression) {
+    return expression.is(Tree.Kind.BOOLEAN_LITERAL) && !Boolean.parseBoolean(((LiteralTree) expression).value());
```
m-g-sonar, line 107: What about using `LiteralUtils.isFalse(tree)` here?
I also feel that we might be missing cases of constants, such as:
```
daoauth.setHideUserNotFoundExceptions(MY_CONSTANT);
```
From an expressionTree, you can probably call `asConstant(Boolean.class)`, and it might work.

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,123 @@
...
+
+    MethodInvocationTree methodInvocationTree = (MethodInvocationTree) tree;
+    Arguments arguments = methodInvocationTree.arguments();
+    if (arguments.isEmpty()){
+      return;
+    }
+    ExpressionTree expression = arguments.get(0);
+
+    checkHiddenUserNotFoundException(methodInvocationTree, expression);
+
```
m-g-sonar, line 81: no need of extra space here.

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,123 @@
...
+      checkThrowUsernameNotFoundException(((ThrowStatementTree) tree));
+      return;
+    }
+
+    MethodInvocationTree methodInvocationTree = (MethodInvocationTree) tree;
+    Arguments arguments = methodInvocationTree.arguments();
+    if (arguments.isEmpty()){
+      return;
+    }
+    ExpressionTree expression = arguments.get(0);
```
m-g-sonar, line 78: why not calling it simply `firstArgument` ?

m-g-sonar (changes requested):

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
margarita-nedzelska-sonarsource: I checked this case and it reports the issue if we're catching UsernameNotFoundException and throwing it again. To my mind, we should report here because from the code it's obvious that we will throw UsernameNotFoundException, which is tricky.

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,129 @@
...
+  private static final String ABSTRACT_USER_DETAILS_AUTHENTICATION_PROVIDER = "org.springframework.security.authentication.dao.AbstractUserDetailsAuthenticationProvider";
+  private static final String USER_DETAILS_SERVICE = "org.springframework.security.core.userdetails.UserDetailsService";
+  private static final String USERNAME_NOT_FOUND_EXCEPTION = "org.springframework.security.core.userdetails.UsernameNotFoundException";
+  private static final String HIDE_USER_NOT_FOUND_EXCEPTIONS = "setHideUserNotFoundExceptions";
+  private static final String LOAD_USER_BY_USERNAME = "loadUserByUsername";
+  private static final String BOOLEAN = "boolean";
+  private static final String STRING = "java.lang.String";
+  private static final String THROWABLE = "java.lang.Throwable";
+
+  private static final MethodMatchers setHideUserMatcher = MethodMatchers.create()
```
m-g-sonar, line 50: I'm a bit surprised not having a code smell here, I would have expected an issue to be raised by [java:S115](https://next.sonarqube.com/sonarqube/coding_rules?open=java%3AS115&rule_key=java%3AS115) on constants not using upper case names, as convention. Could you fix it?

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,129 @@
...
+    if (loadUserMatcher.matches(methodInvocationTree) && expression.is(Tree.Kind.IDENTIFIER)) {
+      IdentifierTree identifierTree = (IdentifierTree) expression;
+      identifierTree.symbol().usages()
+        .stream().filter(UserEnumerationCheck::checkParentIsThrowable)
+        .forEach(value -> reportIssue(value, MESSAGE));
+    }
+  }
+
+  private void checkHiddenUserNotFoundException(MethodInvocationTree methodInvocationTree, ExpressionTree expression) {
+    if (setHideUserMatcher.matches(methodInvocationTree) && !expression.asConstant(Boolean.class).orElse(true)) {
```
m-g-sonar, line 95: :+1:

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,129 @@
...
+        return true;
+      }
+      current = current.parent();
+    }
+    return false;
+  }
+
+  private static boolean isInsideLoadUserByUserName(Tree tree) {
+    Tree current = tree.parent();
+    while (current instanceof BlockTree || current instanceof MethodTree) {
```
m-g-sonar, line 119: I feel this condition is not enough, or will not report only on very trivial case. If I write something like the following, there is going to be 2 FPs:
```
 public static class MyUserDetailsService4 implements UserDetailsService {
    @Override
    public MyUserDetails loadUserByUsername(String s) throws UsernameNotFoundException {
      if (s.equals("yolo"))
        throw new UsernameNotFoundException(""); // Compliant

      switch(s) {
        case "admin":
          throw new UsernameNotFoundException(""); // Compliant
        case "batman":
          break;
      }
      return null;
    }
  }
```
In order to handle this case, I would use a stack of methodTree I'm entering, and just checking the top of the stack is NOT the `loadUserByUsername` method.

For the other _parent loop_, I can see it will anyway stop fast. For this one, it will stop too early, and making it more general will trigger the loop until the parent is null which we want to avoid.

margarita-nedzelska-sonarsource: I have a small question here.

Firstly, I refactored it to additional method visitor. So it checked each method which wasn't `loadUserByUsername` and reported an issue each time an exception `UserNotFoundException` was found.
Then I realized that this way we're going to check all methods in class. And a chance to meet this exception is quite low. I suppose It's better to check exact throw statement if it is in scope of the `loadUserByUsername` method. The only solution that came to my mind will be 
`private static boolean isInsideLoadUserByUserName(Tree tree) {
    Tree current = tree.parent();
    while (current != null) {
      if (current.is(Tree.Kind.METHOD) && LOAD_USER_MATCHER.matches(((MethodTree) current))) {
        return true;
      }
      current = current.parent();
    }
    return false;
  }`

Is it ok or am I missing something?

java-checks/src/main/java/org/sonar/java/checks/security/UserEnumerationCheck.java
```diff
@@ -0,0 +1,131 @@
...
+      return;
+    }
+    ExpressionTree firstArgument = arguments.get(0);
+
+    checkHiddenUserNotFoundException(methodInvocationTree, firstArgument);
+    checkLoadUserArgUsedInExceptions(methodInvocationTree, firstArgument);
+  }
+
+  @Override
+  public void leaveNode(Tree tree) {
```
m-g-sonar, line 96: I think I would also clear the stack when entering file, just to be sure that if the rules crash on one file (the analyzer will recover), we start over from a clear state on next file.

margarita-nedzelska-sonarsource: is it better to do in scanFile() at the beginning or in leaveFile() at the end?

alban-auzeill: If a rule is stateful, it's mandatory to clean the state when entering the file and it's usually done by overriding:
```
  @Override
    mystate.clear();
    super.setContext(context);
  }
```
Because you don't know if an exception previously occurs during the last analyzed file. When an exception occurs, `leaveNode` is not called, the sate is stale, and a recovery mechanism continues with the next file.

Finally, you can also keep the cleanup in the `leaveNode` method, it frees the memory before the next file.

m-g-sonar (approved):

PR #3216, opened 2020-09-25
SONARJAVA-3396 Rule S5808 Authorizations should be based on strong decisions

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/AuthorizationsStrongDecisionsCheck.java
```diff
@@ -0,0 +1,120 @@
...
+    public boolean takesStrongDecision() {
+      // We don't compute the return value when the expression is complex (method call, expression),
+      // and consider it as strong decision to avoid FP
+      return takesStrongDecision || containsComplexReturn;
+    }
+
+    @Override
+    public void visitReturnStatement(ReturnStatementTree tree) {
+      ExpressionTree expression = tree.expression();
+      if (expression != null && (expression instanceof LiteralTree || expression.is(Tree.Kind.IDENTIFIER))) {
```
alban-auzeill, line 100: An identifier could be different from `ACCESS_GRANTED`, `ACCESS_ABSTAIN`, `ACCESS_DENIED`, `true`, `false` like a variable, and in this case it could be complex. This is a FP:
```
    @Override
    public int vote(Authentication authentication, Object object, Collection collection) { // false-positive
      int access = ACCESS_DENIED;
      if(testAccess(authentication, object, collection)) {
        access = ACCESS_GRANTED;
      }
      return access;
    }
```

alban-auzeill (changes requested):

java-checks-test-sources/src/main/java/checks/security/AuthorizationsStrongDecisionsCheck.java
```diff
@@ -0,0 +1,298 @@
...
+
+  // AccessDecisionVoter =================================================
+
+  class WeakNightVoter implements AccessDecisionVoter {
+    @Override
+    public int vote(Authentication authentication, Object object, Collection collection) { // Noncompliant [[sc=16;ec=20]] {{"vote" method should return at least one time ACCESS_DENIED.}}
+      Calendar calendar = Calendar.getInstance();
+      int currentHour = calendar.get(Calendar.HOUR_OF_DAY);
+      if(currentHour >= 8 && currentHour <= 19) {
+        return ACCESS_GRANTED;
```
alban-auzeill, line 22: There's false-negatives because we don't support the second form:

1. `return ACCESS_GRANTED;`
2. `return AccessDecisionVoter.ACCESS_GRANTED;`

PR #3217, opened 2020-09-28
SONARJAVA-3479 FP in S2184 when return is in another scope

quentin-jaquier-sonarsource:


margarita-nedzelska-sonarsource (approved):
Looks good!

PR #3218, opened 2020-09-28
SONARJAVA-3525 S2333 supports redundant modifiers on nested interfaces and classes

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/RedundantModifierCheck.java
```diff
@@ -64,6 +64,13 @@ public void visitNode(Tree tree) {
         checkRedundantModifier(modifiers, Modifier.FINAL);
       } else if(member.is(Kind.CONSTRUCTOR) && tree.is(Kind.ENUM)) {
         checkRedundantModifier(((MethodTree) member).modifiers(), Modifier.PRIVATE);
+      } else if (member.is(Kind.CLASS, Kind.INTERFACE) && isInterfaceOrAnnotation(tree)) {
+        ClassTree nested = (ClassTree) member;
+        ModifiersTree modifiers = nested.modifiers();
+        checkRedundantModifier(modifiers, Modifier.PUBLIC);
+        checkRedundantModifier(modifiers, Modifier.STATIC);
+      } else if (member.is(Kind.INTERFACE) && tree.is(Kind.CLASS, Kind.ENUM)) {
+        checkRedundantModifier(((ClassTree) member).modifiers(), Modifier.STATIC);
```
margarita-nedzelska-sonarsource, line 73: In general I'll suggest to refactor it `switch {}` to have sth like this:

`      switch (member.kind()) {
        case METHOD:
          checkMethod(classTree, (MethodTree) member);
          break;
        case VARIABLE:
          checkVariable(classTree, (VariableTree) member);
          break;
        case CONSTRUCTOR: ...
        case CLASS: ...
        case INTERFACE: ...

      }`
To my mind it will be more readable, too many else if is quite difficult to understand

margarita-nedzelska-sonarsource (changes requested):
It looks like you have uncovered conditions. I guess it's a constructor in a class. And sth non-compiling.

Also it looks like this condition `if (member.is(Kind.INTERFACE) && tree.is(Kind.CLASS, Kind.ENUM)` is always true if the first part of it is true.

java-checks/src/main/java/org/sonar/java/checks/RedundantModifierCheck.java
```diff
@@ -47,35 +47,71 @@
...
+  }
+
+  private void checkVariable(VariableTree variableTree, ClassTree classTree) {
+    if (isInterfaceOrAnnotation(classTree)) {
+      ModifiersTree modifiers = variableTree.modifiers();
+      checkRedundantModifier(modifiers, Modifier.PUBLIC, Modifier.STATIC, Modifier.FINAL);
+    }
+  }
+
+  private void checkNestedClass(ClassTree nested, ClassTree classTree) {
```
margarita-nedzelska-sonarsource, line 92: Maybe it would be nice to rename it to sth more general like `checkNestedType`. Cause this method is used on both classes and interfaces, so a little bit confusing.

java-checks/src/main/java/org/sonar/java/checks/RedundantModifierCheck.java
```diff
@@ -47,35 +47,71 @@
...
+
   private static boolean isInterfaceOrAnnotation(Tree tree) {
     return tree.is(Tree.Kind.INTERFACE, Tree.Kind.ANNOTATION_TYPE);
   }
 
-  private void checkRedundantModifier(ModifiersTree modifiersTree, Modifier modifier) {
-    ModifierKeywordTree foundModifier = ModifiersUtils.getModifier(modifiersTree, modifier);
-    if (foundModifier != null) {
-      reportIssue(foundModifier, "\"" + modifier.toString().toLowerCase(Locale.US) + "\" is redundant in this context.");
+  private void checkRedundantModifier(ModifiersTree modifiersTree, Modifier... modifiers) {
```
margarita-nedzelska-sonarsource, line 109: Maybe 'Modifiers'?

java-checks/src/main/java/org/sonar/java/checks/RedundantModifierCheck.java
```diff
@@ -47,35 +47,71 @@
...
+    for (Modifier modifier : modifiers) {
+      ModifierKeywordTree foundModifier = ModifiersUtils.getModifier(modifiersTree, modifier);
+      if (foundModifier != null) {
+        reportIssue(foundModifier, "\"" + modifier.toString().toLowerCase(Locale.US) + "\" is redundant in this context.");
```
margarita-nedzelska-sonarsource, line 113: I'll suggest also to have a String representation of each modifier somewhere in utils or helpers. Cause it's not optimal to have `modifier.toString().toLowerCase(Locale.US)`  each time we need it.

Also, that's up to you, personally I prefer `String.format()` over String concat.

margarita-nedzelska-sonarsource (approved):
Good job!

java-checks/src/main/java/org/sonar/java/checks/RedundantModifierCheck.java
quentin-jaquier-sonarsource: I agree for the `String.format()`.

>it's not optimal to have modifier.toString().toLowerCase(Locale.US)

I refactored the code to have `foundModifier.keyword().text()`, I feel that it's slightly better (at least, we do not recompute `toLowerCase`...). 
In any way, we will only call this code when an issue is reported, so it should not be too frequent.

margarita-nedzelska-sonarsource: LGTM

PR #3220, opened 2020-09-28
SONARJAVA-3538 Cover custom assertions

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/tests/AssertJConsecutiveAssertionCheck.java
```diff
@@ -136,26 +140,49 @@ private void reportConsecutiveAssertions(List<StatementTree> statements) {
...
+      return SyntacticEquivalence.areEquivalent(arg, other.arg)
+        && (other.assertionType.isSubtypeOf(assertionType) || couldBeChained(other));
+    }
+
+    boolean couldBeChained(AssertSubject other) {
+      Tree otherParent = other.mit
+        // owning member select
+        .parent()
+        // chained method invocation
+        .parent();
```
quentin-jaquier-sonarsource, line 173: You can use `MethodTreeUtils.consecutiveMethodInvocation` to do this.

m-g-sonar: Oh nice!

PR #3222, opened 2020-09-29
SONARJAVA-3478 Support Collection and Map methods

m-g-sonar:


its/ruling/src/test/resources/commons-beanutils/java-S2201.json
```diff
@@ -0,0 +1,6 @@
+{
+'commons-beanutils:commons-beanutils:src/main/java/org/apache/commons/beanutils2/ContextClassLoaderLocal.java':[
+145,
```
m-g-sonar, line 3: Note that to me these issues are FP... but these are ugly things. WeakHashMaps have mechanisms to clear unused objects. Here, they force calls to clearing methods, in a way forcing the normal garbage collection. So even if it's a FP, I would leave with it as it is as long as no one complains. Also, detecting the runtime type of an object is NOT trivial.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3224, opened 2020-09-30
SONARJAVA-3539 FP in S5845 when BigDecimal and BigInteger are compared with string

quentin-jaquier-sonarsource:


margarita-nedzelska-sonarsource (approved):
LGTM!

PR #3225, opened 2020-09-30
SONARJAVA-3468 FP on S1905 when casted argument is an ambiguous method reference

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
In this implementation I check only whether there are overloads both for method reference itself and the method which accept it as a parameter. Unfortunately, I didn't take into account the type of the Method Reference and whether it can be automatically casted. This will lead to possible False Negatives (when there are overloads on both sides but only one can be applied)

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -172,6 +175,24 @@ private static boolean isUnnecessarySubtypeCast(Type childType, TypeCastTree typ
         || isUnnecessaryLambdaCast(childType, parentType));
   }
 
+  private static boolean isAmbiguousMethodReference(TypeCastTree typeCastTree) {
+    MethodReferenceTree methodReferenceTree = (MethodReferenceTree) typeCastTree.expression();
+    Symbol.TypeSymbol methodRefOwner = (Symbol.TypeSymbol) methodReferenceTree.method().symbol().owner();
+    long numberMethodRefsOverloads = getOverloadsCount(methodRefOwner,methodReferenceTree.method().symbol().name());
+
+    MethodInvocationTree methodInvocationTree = (MethodInvocationTree) typeCastTree.parent().parent();
+    Symbol.TypeSymbol owner = (Symbol.TypeSymbol) methodInvocationTree.symbol().owner();
```
quentin-jaquier-sonarsource, line 184:
```
 unknown((Supplier<String>) Overloaded::fff);
```
This code... **does not** raise a `ClassCastException`. :)
But an issue is still reported and I think it makes sense to not report an issue if we don't know the type expected. If the fix is easy, and it makes sense to include it with this PR, we could do it now.
Otherwise, we can simply create a new ticket to track this minor issue.

quentin-jaquier-sonarsource (changes requested):
S1905 is dealing with a similar problem, see [SONARJAVA-2795](https://jira.sonarsource.com/browse/SONARJAVA-2795) and [SONARJAVA-3378](https://jira.sonarsource.com/browse/SONARJAVA-3378), and especially this code: https://github.com/SonarSource/sonar-java/blob/b5144dc4d367e9d2fd038b08db764ef8eb1b9e9a/java-checks/src/main/java/org/sonar/java/checks/ReplaceLambdaByMethodRefCheck.java#L238

Can we re-use it for this rule? If it turns out to be too complex, the heuristic you used is good enough.

java-checks-test-sources/src/main/files/non-compiling/checks/RedundantTypeCastCheck.java
```diff
@@ -0,0 +1,164 @@
...
+    String s2 = (String) s1; // Noncompliant {{Remove this unnecessary cast to "String".}}
+    A a = (A) new B(); // Noncompliant {{Remove this unnecessary cast to "A".}}
+    A[][] as = (A[][]) new B[1][1]; // Noncompliant {{Remove this unnecessary cast to "A[][]".}}
+    B b = null;
+    fun(b);
+    fun((B) b); // Noncompliant
+    fun((A) b); // Compliant - exception to distinguish the method to call
+    funBParameter((A) b); // Noncompliant
+    List<B> bees = new java.util.ArrayList<B>();
+    List<A> aaas = (List) bees;
```
quentin-jaquier-sonarsource, line 35: All this code is compiling and is duplicated in the compiling file. I think we should keep only what is not compiling in this file, to avoid duplication.

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -163,13 +163,33 @@ private static Tree skipParentheses(Tree parent) {
...
       && (typeCastTree.type().symbolType().equals(childType)
         || (isArgument && childType.equals(parentType)) || (!isArgument && childType.isSubtypeOf(parentType)))
       && (!ExpressionUtils.skipParentheses(typeCastTree.expression()).is(Tree.Kind.LAMBDA_EXPRESSION)
-        || isUnnecessaryLambdaCast(childType, parentType));
+        || isUnnecessaryLambdaCast(childType, parentType))
+      && !(isArgument && isMandatoryMethodReferenceCast(typeCastTree, parentTree));
+  }
+
+  private static boolean isMandatoryMethodReferenceCast(TypeCastTree typeCastTree, Tree parentTree) {
+    if (typeCastTree.expression().is(Tree.Kind.METHOD_REFERENCE) && skipParentheses(parentTree.parent()).is(Tree.Kind.METHOD_INVOCATION)) {
```
quentin-jaquier-sonarsource, line 179: As a general observation, I tend to prefer to create an intermediate variable to store `typeCastTree.expression()`.
In my opinion, it increases the clarity of the code and avoids problems in case it returns differents values between two calls, or if the computation is expensive. 
Same for `parentTree.parent()`, to avoid having to call `skipParentheses()` twice.

java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -163,13 +163,33 @@ private static Tree skipParentheses(Tree parent) {
...
+      MethodInvocationTree methodInvocationTree = (MethodInvocationTree) skipParentheses(parentTree.parent());
+      Symbol methodAsArg = expression.method().symbol();
+      Symbol methodCaller = methodInvocationTree.symbol();
+      return hasOverloads(methodAsArg) && hasOverloads(methodCaller);
+    }
+    return false;
+  }
+
+  private static boolean hasOverloads(Symbol symbol) {
+    return symbol.owner().isTypeSymbol() && ((Symbol.TypeSymbol) symbol.owner()).memberSymbols().stream()
```
quentin-jaquier-sonarsource, line 190: Same here, let's create a variable to store `symbol.owner()`.

java-checks-test-sources/src/main/files/non-compiling/checks/RedundantTypeCastCheck.java
```diff
@@ -116,24 +116,49 @@ String qix2() {
...
+      return "";
     }
-  }}
+  }
+
+  void main() {
+    foo(Overloaded::f); // Does not compile without cast
+    foo(String::new); // Does not compile without cast
+
+    bar(((Supplier<String>) Overloaded::fff));
```
quentin-jaquier-sonarsource, line 136: I would also put this test in the compiling part, to be sure that unknown method does not prevent the Exception to happen.

quentin-jaquier-sonarsource (changes requested):

java-checks-test-sources/src/main/files/non-compiling/checks/RedundantTypeCastCheck.java
margarita-nedzelska-sonarsource: I tried to remove as many duplications as possible but it's kinda hard to make difference between what you need in non-compiling code. I've already removed much from non-compiling file.

PR #3226, opened 2020-09-30
SONARJAVA-3540 FP in S2175 when a primitive is auto-boxed into a subtype of Number

quentin-jaquier-sonarsource:


margarita-nedzelska-sonarsource (approved):
Well spotted! Looks good!

PR #3228, opened 2020-10-01
SONARJAVA-3542 RSPEC-5993 Constructors of an "abstract" class should not be declared "public"

margarita-nedzelska-sonarsource:
.


java-checks/src/main/java/org/sonar/java/checks/PublicConstructorInAbstractClassCheck.java
```diff
@@ -0,0 +1,68 @@
...
+      return;
+    }
+
+    Optional<ModifierKeywordTree> abstractKeyword = ModifiersUtils.findModifier(classTree.modifiers(), Modifier.ABSTRACT);
+
+    abstractKeyword.ifPresent(keyword -> {
+      JavaFileScannerContext.Location secondaryLocation = new JavaFileScannerContext.Location("This class is \"abstract\".", keyword);
+      classTree.members().stream()
+        .filter(PublicConstructorInAbstractClassCheck::isaConstructor)
+        .map(member -> ((MethodTree) member))
```
m-g-sonar, line 56: You can use `MethodTree.class::cast` instead of the lambda

java-checks/src/main/java/org/sonar/java/checks/PublicConstructorInAbstractClassCheck.java
```diff
@@ -0,0 +1,68 @@
...
+        .map(methodTree -> ModifiersUtils.findModifier(methodTree.modifiers(), Modifier.PUBLIC))
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .forEach(modifier -> reportIssue(modifier, "Change the visibility of this constructor to \"protected\".",
+          Collections.singletonList(secondaryLocation), null));
+    });
+  }
+
+  private static boolean isaConstructor(Tree member) {
```
m-g-sonar, line 65: I would prefer `isConstructor()`

java-checks/src/main/java/org/sonar/java/checks/PublicConstructorInAbstractClassCheck.java
```diff
@@ -0,0 +1,68 @@
...
+    abstractKeyword.ifPresent(keyword -> {
+      JavaFileScannerContext.Location secondaryLocation = new JavaFileScannerContext.Location("This class is \"abstract\".", keyword);
+      classTree.members().stream()
+        .filter(PublicConstructorInAbstractClassCheck::isaConstructor)
+        .map(member -> ((MethodTree) member))
+        .map(methodTree -> ModifiersUtils.findModifier(methodTree.modifiers(), Modifier.PUBLIC))
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .forEach(modifier -> reportIssue(modifier, "Change the visibility of this constructor to \"protected\".",
+          Collections.singletonList(secondaryLocation), null));
```
m-g-sonar, line 61: Let's make the list of secondaries outside this stream to not have to do it again.

java-checks/src/main/java/org/sonar/java/checks/PublicConstructorInAbstractClassCheck.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    Optional<ModifierKeywordTree> abstractKeyword = ModifiersUtils.findModifier(classTree.modifiers(), Modifier.ABSTRACT);
+
+    abstractKeyword.ifPresent(keyword -> {
+      JavaFileScannerContext.Location secondaryLocation = new JavaFileScannerContext.Location("This class is \"abstract\".", keyword);
+      classTree.members().stream()
+        .filter(PublicConstructorInAbstractClassCheck::isaConstructor)
+        .map(member -> ((MethodTree) member))
+        .map(methodTree -> ModifiersUtils.findModifier(methodTree.modifiers(), Modifier.PUBLIC))
```
m-g-sonar, line 57: I would inline this into a method `isPublic(MethodTree)` to be able to use a method reference again.

m-g-sonar (approved):

PR #3230, opened 2020-10-01
SONARJAVA-3535 Do not raise issues if singleton is annotated with @ConfigurationProperties

m-g-sonar:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3232, opened 2020-10-02
SONARJAVA-3414 Rule S4790: its content should be replaced by S2070

margarita-nedzelska-sonarsource:
Update reported message for a security hotspot

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/security/DataHashingCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/security/DataHashingCheck.java
@@ -19,9 +19,34 @@
  */
 package org.sonar.java.checks.security;
 
+import com.google.common.collect.ImmutableSet;
+import java.util.Optional;
+import java.util.Set;
 import org.sonar.check.Rule;
 import org.sonar.java.checks.AbstractHashAlgorithmChecker;
 
 @Rule(key = "S4790")
 public class DataHashingCheck extends AbstractHashAlgorithmChecker {
+
+  public static final Set<String> DEPRECATED_HASH_CLASSES = ImmutableSet.of(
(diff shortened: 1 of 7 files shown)


java-checks/src/main/java/org/sonar/java/checks/security/DataHashingCheck.java
```diff
@@ -19,9 +19,33 @@
...
+    DeprecatedSpringPasswordEncoder.NO_OP.classFqn,
+    DeprecatedSpringPasswordEncoder.STANDARD.classFqn
+  );
+
+  private static final String MESSAGE = "Make sure this weak hash algorithm is not used in a sensitive context here.";
+
+  @Override
+  protected String getMessageForClass(String className) {
+    return DEPRECATED_HASH_CLASSES.contains(className) ? MESSAGE : null;
+  }
```
quentin-jaquier-sonarsource, line 45: The result of this method is used without null check, relying on the fact that `DEPRECATED_HASH_CLASSES` set contains exactly all the value of `DeprecatedSpringPasswordEncoder`, synchronization that has to be done "by hand". I'm not sure it's a good pattern, I would rather return an Optional or consistently check it for null here.

margarita-nedzelska-sonarsource: I refactored it, but to my mind, as far as the implementation of AbstractMethodDetection and AbstractHashAlgorithmChecker guarantees there will never happen null. I someone will need to add a new class and forget to update it, there will be problem during implementation and testing. So will never go to production.

On the other hand, if we just hide someone's mistake by adding optional, that might not be a good experience to fix it.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3237, opened 2020-10-05
SONARJAVA-3547 Improve rule S1612 to replace casts with method reference

margarita-nedzelska-sonarsource:

### DIFF
--- a/java-checks-test-sources/src/main/files/non-compiling/checks/ReplaceLambdaByMethodRefCheck.java
+++ b/java-checks-test-sources/src/main/files/non-compiling/checks/ReplaceLambdaByMethodRefCheck.java
@@ -1,0 +1,265 @@
+package checks;
+
+import java.util.*;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+class A {
+  void fun() {
+    IntStream.range(1, 5)
+        .map((x) -> x * x)
+        .map(x -> square(x)) // Noncompliant [[sc=16;ec=18]] {{Replace this lambda with a method reference.}}
+        .map(x -> { // Noncompliant
+          return square(x);
+        })
+        .map(this::square) //Compliant
+        .forEach(System.out::println);
+    IntStream.range(1, 5).forEach(x -> System.out.println(x)); // Noncompliant
+    IntStream.range(1, 5).forEach(x -> { // Noncompliant
+          System.out.println(x);
+        });
+    IntStream.range(1, 5).forEach(x -> {return;}); // Compliant
+
+    Arrays.asList("bar").stream().filter(string -> string.startsWith("b")); // Compliant
+    Arrays.asList(new A()).stream().filter(a -> a.coolerThan(0, a)); // Compliant
+
+    biConsumer((x, y) -> x * y);
+    biConsumer((x, y) -> { ; });
+    biConsumer((x, y) -> { ;; });
+  }
+
+  void biConsumer(BiConsumer consumer) {
+  }
+
+  void runnable(Runnable runnable) {
+  }
+
+  void foo(List<String> list, String a) {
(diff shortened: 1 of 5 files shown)


java-checks/src/main/java/org/sonar/java/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -106,6 +127,23 @@ private void visitLambdaExpression(LambdaExpressionTree tree) {
...
+      return Optional.empty();
+    }
+    Tree expr = statement;
+    if (expr.is(Tree.Kind.PARENTHESIZED_EXPRESSION)) {
+      expr = ExpressionUtils.skipParentheses((ParenthesizedTree) statement);
+    }
+    if (expr.is(Tree.Kind.TYPE_CAST)) {
+      TypeCastTree typeCastTree = (TypeCastTree) expr;
+      if (isSingleParamCast(typeCastTree.expression(), tree)) {
+        return Optional.of(typeCastTree.type().symbolType().name());
```
m-g-sonar, line 141: what if the type is unknown but the symbols are matched by the frontend? I fear it's going to return "!unknown!" here.

Instead of:
```java
return Optional.of(typeCastTree.type().symbolType().name());
```
I would use something like so we use directly the TypeTree, which, even unresolved, will display the correct name:
```java
return Optional.of(ExpressionsHelper.concatenate(typeCastTree.type()));
```

java-checks/src/main/java/org/sonar/java/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -106,6 +127,23 @@ private void visitLambdaExpression(LambdaExpressionTree tree) {
     return Optional.empty();
   }
 
+  private static Optional<String> getTypeCast(@Nullable Tree statement, LambdaExpressionTree tree) {
+    if (statement == null) {
+      return Optional.empty();
+    }
+    Tree expr = statement;
+    if (expr.is(Tree.Kind.PARENTHESIZED_EXPRESSION)) {
```
m-g-sonar, line 135: I feel this check defeats a bit the purpose of using `ExpressionUtils.skipParentheses(...);`, Won't it be possible to already arrive here which a (Nullable) ExpressionTree instead? It would make things easier.

java-checks/src/main/java/org/sonar/java/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -87,6 +93,21 @@ private void visitLambdaExpression(LambdaExpressionTree tree) {
...
 
+  private static Optional<String> getTypeCast(LambdaExpressionTree lambda) {
+    Tree lambdaBody = lambda.body();
+    if (isBlockWithOneStatement(lambdaBody)) {
+      return getTypeCastFromReturn(((BlockTree) lambdaBody).body().get(0), lambda);
+    }
+    return getTypeCast(lambdaBody, lambda);
+  }
+
+  private static Optional<String> getTypeCastFromReturn(Tree statement, LambdaExpressionTree lambda) {
```
m-g-sonar, line 104: The only reason of keeping the lambda as 2nd parameter is to ultimately check for the parameter symbol (3 calls below). Wouldn't it be possible to check this earlier (only one param for instance) and only pass a symbol?

java-checks-test-sources/src/main/files/non-compiling/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -1,4 +1,10 @@
-import java.util.*;
```
m-g-sonar, line 1: This change does not seem required to me, you kept it in the other file.

m-g-sonar (changes requested):

m-g-sonar (approved):

PR #3250, opened 2020-10-19
S1948: a searializable collection with serializable generic-type doesn't have to be private or transient (false-positive)

xtermi2:
In my optinion rule `S1948` produces false-positives on searializable collections with serializable generic-types, e.g. 
```
class MyClass implements Serializable {
  ArrayList<String> arrayList; // Compliant
}
```
The argument `because they could be assigned non-Serializable values externally` is not true:
* I can't assign an non-Serializable `List` to the above `arrayList`
* OK, I can assign a `ArrayList<Object>` when I ignore all compiler warnings and IDE warnings, but in this case a ClassCast error is much more likely/serious then a serialization issue! And even if the field is private I could also assign this with a getter.

  - no Jira issue exists

### DIFF
--- a/its/ruling/src/test/resources/commons-beanutils/java-S1948.json
+++ b/its/ruling/src/test/resources/commons-beanutils/java-S1948.json
@@ -5,7 +5,6 @@
 ],
 'commons-beanutils:commons-beanutils:src/main/java/org/apache/commons/beanutils2/BasicDynaClass.java':[
 114,
-141,
 ],
 'commons-beanutils:commons-beanutils:src/main/java/org/apache/commons/beanutils2/BeanComparator.java':[
 52,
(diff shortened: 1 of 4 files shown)


m-g-sonar:
Hey Andreas,

Thanks for getting back to us with this (painful) rule. Indeed, another FP here. I created the following ticket to track it: [SONARJAVA-3571](https://jira.sonarsource.com/browse/SONARJAVA-3571)

Once again, I'm going to duplicate your PR to trigger our QA, and merge it if everything goes fine.

Cheers,
Michael

m-g-sonar:
Hey, 
I merged the duplicate (it was rebased on top of our main branch)!
 Thanks again for your contribution.

Like your other fix, it will also be part of our next release of the Java Analyzer (currently planned for mid-November).

Cheers,
Michael

PR #3259, opened 2020-11-09
SONARJAVA-3458 Revert native issue filtering to reuse SQ API (#3110)

alban-auzeill:
This reverts commit 696fa143e137d5be70395f863d7773e5ccce5210.

### DIFF
--- a/its/plugin/tests/src/test/java/com/sonar/it/java/JspTest.java
+++ b/its/plugin/tests/src/test/java/com/sonar/it/java/JspTest.java
@@ -17,8 +17,10 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-package com.sonar.it.java.suite;
+package com.sonar.it.java;
 
+import com.sonar.it.java.suite.JavaTestSuite;
+import com.sonar.it.java.suite.TestUtils;
 import com.sonar.orchestrator.Orchestrator;
 import com.sonar.orchestrator.OrchestratorBuilder;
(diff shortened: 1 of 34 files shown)


its/plugin/tests/src/test/java/com/sonar/it/java/suite/JavaTestSuite.java
```diff
@@ -52,12 +52,11 @@
   SuppressWarningTest.class,
   SonarLintTest.class,
   ExternalReportTest.class,
-  DuplicationTest.class,
-  JspTest.class
```
m-g-sonar, line 56: ohlala, nop, we just dropped a test here.

its/plugin/tests/src/test/java/com/sonar/it/java/JspTest.java
```diff
@@ -17,8 +17,10 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-package com.sonar.it.java.suite;
+package com.sonar.it.java;
```
m-g-sonar, line 20: really not sure about this.

its/plugin/tests/src/test/java/com/sonar/it/java/suite/JavaTestSuite.java
alban-auzeill: If you look at the log of the QA, you will see `com.sonar.it.java.JspTest`.
I move JspTest out of JavaTestSuite because it requires a different version of SQ.
But it is still executed because of the pom.xml `<include>com/sonar/it/java/*Test.java</include>`

PR #3262, opened 2020-11-12
SONARJAVA-3546 Issue message of S5961 should contains the number of actual assertions

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/java/checks/TooManyAssertionsCheckCustom2.java
```diff
@@ -8,7 +8,7 @@
 public class TooManyAssertionsCheckCustom2 {
 
   @Test
-  void test1() { // Noncompliant [[sc=8;ec=13;secondary=12,13,14]]{{Refactor this method in order to have less than 2 assertions.}}
+  void test1() { // Noncompliant [[sc=8;ec=13;secondary=12,13,14]]{{Refactor this method in order to not have 3 assertions, which is greater than 2 authorized.}}
```
m-g-sonar, line 11: Woa, I think we need to reformulate this. What about something simpler:
```
Refactor this method to reduce the number of assertions from 3 to less than 2.
```

quentin-jaquier-sonarsource: Yea, I tried to imitates other rules messages, but in the end, they all have a different one, and this one looks good.

PR #3267, opened 2020-11-13
SONARJAVA-3545 Rule S4973 shouldn't report an issue if "==" is used to compare Boolean constants

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
I still have doubts with this rule, by introducing these changes we'll add some FN

quentin-jaquier-sonarsource (approved):
Minor comment, otherwise LGTM.

_For the record, we discussed the problem together and concluded that we will go ahead with this solution, see if the results are good enough and adjust if needed._

PR #3268, opened 2020-11-13
SONARJAVA-2154 remove guava ImmutableList from java-checks module

margarita-nedzelska-sonarsource:


m-g-sonar (commented):
Overall, LGTM.
What about `ImmutableList` methods in the forbidden API configuration now?
Is there still some usage remaining?

margarita-nedzelska-sonarsource:
The forbidden API configuration is in parent pom. In this PR we're touching java-checks module, there are some usages in other modules. So I think it's better to remove it in the end.

PR #3269, opened 2020-11-16
SONARJAVA-2154 Remove usage of ImmutableMap from java-checks module

margarita-nedzelska-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/helpers/MapBuilder.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/helpers/MapBuilder.java
@@ -1,0 +1,46 @@
+/*
+ * SonarQube Java
+ * Copyright (C) 2012-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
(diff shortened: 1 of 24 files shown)


java-checks/src/main/java/org/sonar/java/checks/helpers/MapBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+    return new MapBuilder<>();
+  }
+
+  private final Map<K, V> map;
+
+  private MapBuilder() {
+    this.map = new HashMap<>();
+  }
+
+  public MapBuilder<K, V> add(K key, V value) {
```
m-g-sonar, line 38: Why not keep the usual map name of inserting items `put()` ?
It would make most of the changes transparent.

java-checks/src/main/java/org/sonar/java/checks/helpers/MapBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java.checks.helpers;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class MapBuilder<K, V> {
```
m-g-sonar, line 26: Why not add a reference, or small explanation here, telling that [Java9+ Immutable Map Static Factory Methods](https://docs.oracle.com/javase/9/docs/api/java/util/Map.html#immutable) are going to be useful in the future, and might be used to replace this class.

m-g-sonar (changes requested):

java-checks/src/main/java/org/sonar/java/checks/helpers/MapBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java.checks.helpers;
```
m-g-sonar, line 20: Also, in order to make it useful for everyone in this rework, I would recommend to put it in `org.sonar.java.collections` package from the `java-frontend` module.

m-g-sonar (approved):

PR #3273, opened 2020-11-16
SONARJAVA-3612 FP in S1185 when class is annotated "@Transactional"

quentin-jaquier-sonarsource:


margarita-nedzelska-sonarsource (changes requested):
As for the ticket, it looks good.
But as far as you've already touched this rule. Let's get rid of "hasSemantic()" check and move test files to test-sources module

java-checks/src/main/java/org/sonar/java/checks/MethodOnlyCallsSuperCheck.java
```diff
@@ -156,4 +164,9 @@ private static boolean hasAnnotationDifferentFromOverride(List<AnnotationTree> a
     return false;
   }
 
+  private static boolean isClassAnnotatedWithTransactional(MethodTree methodTree) {
+    SymbolMetadata metadata = methodTree.symbol().enclosingClass().metadata();
```
margarita-nedzelska-sonarsource, line 168: Let's double check we won't get NPE here

margarita-nedzelska-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/MethodOnlyCallsSuperCheck.java
quentin-jaquier-sonarsource: `enclosingClass()` returns null for package symbol, since we are on a MethodTree, this is not possible.
I'm still in favor of defensive programming though, even if we will miss coverage, I will add it.

java-checks/src/main/java/org/sonar/java/checks/MethodOnlyCallsSuperCheck.java
```diff
@@ -156,4 +161,14 @@ private static boolean hasAnnotationDifferentFromOverride(List<AnnotationTree> a
     return false;
   }
 
+  private static boolean isClassAnnotatedWithTransactional(MethodTree methodTree) {
+    Symbol.TypeSymbol typeSymbol = methodTree.symbol().enclosingClass();
+    if (typeSymbol != null){
+      SymbolMetadata metadata = typeSymbol.metadata();
+      return TRANSACTIONAL_ANNOTATIONS.stream().anyMatch(metadata::isAnnotatedWith);
+    }
+    // Defensive programming, should not be possible, enclosingClass returns null only for package symbol.
```
margarita-nedzelska-sonarsource, line 170: I checked a couple of variants, this check seems to be redundant. Even if we put method outside of the class, we're not able to parse the file. So I can't imagine the situation where the enclosing class will be null. So let's remove this null check. If we see any problems on peach with it we still can add the check and at least find the proper test case.

margarita-nedzelska-sonarsource (approved):
Overall looks good, I've just added a comment regarding that null check. It seems to me that it's better to remove it and see what we'll get on peach.

java-checks/src/main/java/org/sonar/java/checks/MethodOnlyCallsSuperCheck.java
quentin-jaquier-sonarsource: Okay, that makes sense. In addition, I double-checked and it seems there are other rules that do not perform the check, and nothing went wrong, so I believe it fine to not do it.

PR #3274, opened 2020-11-16
SONARJAVA-3617 S1170 should not raise an issue when the initializer contains "this" or "super"

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -96,7 +96,14 @@ private static boolean hasConstantInitializer(VariableTree variableTree) {
       if (init.is(Tree.Kind.NEW_ARRAY)) {
         return false;
       }
-      return !containsChildrenOfKind((JavaTree) init, Tree.Kind.METHOD_INVOCATION, Tree.Kind.NEW_CLASS);
+      return !containsChildMatchingPredicate((JavaTree) init, tree -> {
+          if (tree.is(Tree.Kind.IDENTIFIER)) {
```
margarita-nedzelska-sonarsource, line 100: Would be nice to extract it to a separate method. And combine these 2 conditions using `Predicate.or()` if possible.

margarita-nedzelska-sonarsource (changes requested):
As far as you've already touched this rule, let's move its test files to the test-sources module.

java-checks/src/main/java/org/sonar/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -96,7 +96,20 @@ private static boolean hasConstantInitializer(VariableTree variableTree) {
       if (init.is(Tree.Kind.NEW_ARRAY)) {
         return false;
       }
-      return !containsChildrenOfKind((JavaTree) init, Tree.Kind.METHOD_INVOCATION, Tree.Kind.NEW_CLASS);
+      return !containsChildMatchingPredicate((JavaTree) init,
+        ((Predicate<Tree>) ConstantsShouldBeStaticFinalCheck::isIgnoredKind).or(ConstantsShouldBeStaticFinalCheck::isThisOrSuper));
```
margarita-nedzelska-sonarsource, line 100: Looks like my suggestion was not that elegant. For me it will be more readable to use lambda. 
```suggestion
        (tree -> isIgnoredKind(tree) || isThisOrSuper(tree)));
```

margarita-nedzelska-sonarsource (approved):
Looks good!
A minor suggestion to use lambda here to make it more readable.

java-checks/src/main/java/org/sonar/java/checks/ConstantsShouldBeStaticFinalCheck.java
quentin-jaquier-sonarsource: Of course, thanks!

PR #3277, opened 2020-11-16
SONARJAVA-3552 Rule S5996: Regex boundaries should not be used in a way that can never match

sebastian-hungerecker-sonarsource:
Also included:

SONARJAVA-3551 Implement helper to find whether state in regex automaton is reachable without consuming input


java-checks-test-sources/src/main/java/checks/regex/ImpossibleBoundariesCheck.java
```diff
@@ -0,0 +1,44 @@
...
+    str.matches("^$");
+    str.matches("^(?i)$");
+    str.matches("^$(?i)");
+    str.matches("^abc$|^def$");
+    str.matches("(?i)^abc$");
+    str.matches("()^abc$");
+    str.matches("^abc$()");
+    str.matches("^abc$\\b");
+    str.matches("(?=abc)^abc$");
+    str.matches("(?=^abc$)abc");
```
sebastian-hungerecker-sonarsource, line 36: Added these two test cases

java-checks/src/main/java/org/sonar/java/checks/helpers/RegexTreeHelper.java
```diff
@@ -74,4 +79,29 @@ private static void addCurrentGrapheme(List<RegexCheck.RegexIssueLocation> resul
...
+      return false;
+    }
+    visited.add(start);
+    for (AutomatonState successor : start.successors()) {
+      // We don't generally consider elements behind backtracking edges to be 0-input reachable because what comes
+      // after the edge won't directly follow what's before the edge. However, we do consider the end-of-lookahead
+      // state itself reachable (but not any state behind it), so that we can check whether the end of the lookahead
+      // can be reached without input from a given place within the lookahead.
+      if ((successor.incomingTransitionType() == EPSILON && canReachWithoutConsumingInput(successor, goal, visited))
+        || (successor instanceof EndOfLookaroundState && successor == goal)) {
```
sebastian-hungerecker-sonarsource, line 100: Removed backtracking edges from the list of traversable edges and added the ||-exception instead, as well as the comment above.

java-checks/src/main/java/org/sonar/java/checks/regex/ImpossibleBoundariesCheck.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    public void visit(RegexParseResult regexParseResult) {
+      start = regexParseResult.getStartState();
+      end = regexParseResult.getFinalState();
+      super.visit(regexParseResult);
+    }
+
+    @Override
+    public void visitLookAround(LookAroundTree tree) {
```
sebastian-hungerecker-sonarsource, line 55: `visitLookAround` is new.

java-checks-test-sources/src/main/java/checks/regex/ImpossibleBoundariesCheck.java
```diff
@@ -0,0 +1,45 @@
...
+    str.matches("($)a"); // Noncompliant [[sc=19;ec=20]]
+    str.matches("a$|$a"); // Noncompliant [[sc=21;ec=22]]
+    str.matches("^a|a^"); // Noncompliant [[sc=22;ec=23]]
+    str.matches("a(b|^)"); // Noncompliant [[sc=22;ec=23]]
+    str.matches("(?=abc^)"); // Noncompliant [[sc=24;ec=25]]
+    str.matches("abc(?=^abc)"); // Noncompliant [[sc=24;ec=25]]
+    str.matches("abc(?<=$abc)"); // Noncompliant [[sc=25;ec=26]]
+    str.matches("abc(?<=abc$)def"); // Noncompliant [[sc=28;ec=29]]
+    str.matches("(?:abc(X|^))*Y?"); // Noncompliant [[sc=27;ec=28]]
+  }
```
alban-auzeill, line 23: We should add:
```
// FP we raises an issue below because the usage of $ in the middle of a regex is suspicious and redundant
str.matches("(?m)^1$\n2"); // Noncompliant
```

alban-auzeill (approved):

PR #3279, opened 2020-11-17
SONARJAVA-2154 Remove usages of ImmutableSet from java-checks module

margarita-nedzelska-sonarsource:
I've changed Set -> List in few places, where we're not searching element but iterating over them and in case of 2 elements.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/JacksonDeserializationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/JacksonDeserializationCheck.java
@@ -19,11 +19,9 @@
  */
 package org.sonar.java.checks;
 
-import com.google.common.collect.ImmutableSet;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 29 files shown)


java-checks/src/main/java/org/sonar/java/checks/JacksonDeserializationCheck.java
```diff
@@ -89,15 +87,15 @@ private static boolean isAnnotationOnClassOrField(AnnotationTree annotationTree)
       if (tree.is(Tree.Kind.ASSIGNMENT)) {
         AssignmentExpressionTree assignment = (AssignmentExpressionTree) tree;
         if (((IdentifierTree) assignment.variable()).name().equals("use")
-            && isJsonTypeIdEnumValue(assignment.expression(), ImmutableSet.of("CLASS", "MINIMAL_CLASS"))) {
```
m-g-sonar, line 92: 👍🏻

PR #3283, opened 2020-11-17
SONARJAVA-3557 Rule S6001: Back references in regular expressions should only refer to capturing groups that are matched before the reference

sebastian-hungerecker-sonarsource:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6001_java.html
```diff
@@ -0,0 +1,18 @@
+<p>When a back reference in a regex refers to a capturing group that hasn't been defined yet (or at all), it can never be matched. Named back
+references throw a <code>PatternSyntaxException</code> in that case; numeric back references fail silently when they can't match, simply making the
+match fail.</p>
+<p>When the group is defined before the back reference but on a different control path (like in <code>(.)|&amp;bsol;&amp;bsol;1</code> for example),
```
alban-auzeill, line 4: the `\` should not be `&amp;bsol;`, only HTML dev would understand :wink:

alban-auzeill (changes requested):

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6001_java.html
sebastian-hungerecker-sonarsource: How do I fix this? I wrote `&bsol;` in Jira because using a backslash there messes up the formatting in Jira.

alban-auzeill: You could try to escape `\` by `\\`

PR #3285, opened 2020-11-18
SONARJAVA-3608 Rule S6068: Call to Mockito method "verify", "when" or "given" should be simplified

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,94 @@
...
+    Foo foo = mock(Foo.class);
+    Object v1 = new Object();
+    Object v2 = new Object();
+    Object v3 = new Object();
+    Object v4 = new Object();
+    Object v5 = new Object();
+
+    given(foo.bar(eq(v1), eq(v2), eq(v3))).willReturn(null); // Noncompliant [[sc=19;ec=21;secondary=34,34]] {{Remove these useless "eq" invocation and directly use the value.}}
+    when(foo.baz(eq(v4), eq(v5))).thenReturn("foo"); // Noncompliant [[sc=18;ec=20;secondary=35]] {{Remove these useless "eq" invocation and directly use the value.}}
+    doThrow(new RuntimeException()).when(foo).quux(eq(42)); // Noncompliant [[sc=52;ec=54;secondary=]] {{Remove this useless "eq" invocation and directly use the value.}}
```
m-g-sonar, line 36: remove `secondary=` here

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,99 @@
...
+import org.sonar.plugins.java.api.tree.Arguments;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S6068")
+public class MockitoEqSimplificationCheck extends IssuableSubscriptionVisitor {
+  private static final String MOCKITO = "org.mockito.Mockito";
+
+  private static final MethodMatchers MOCKITO_EQ_USED_IN_ARGUMENTS = MethodMatchers.or(
```
m-g-sonar, line 41: `METHODS_USING_EQ_IN_ARGUMENTS`

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,99 @@
...
+    MethodMatchers.create().ofTypes(MOCKITO).names("when")
+      .addParametersMatcher(MethodMatchers.ANY).build(),
+    MethodMatchers.create().ofTypes("org.mockito.BDDMockito").names("given")
+      .addParametersMatcher(MethodMatchers.ANY).build()
+  );
+
+  private static final MethodMatchers MOCKITO_EQ_USED_IN_CONSECUTIVE_CALL = MethodMatchers.or(
```
m-g-sonar, line 48: `METHODS_USING_EQ_IN_CONSECUTIVE_CALL`

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,99 @@
...
+    MethodMatchers.create().ofTypes(MOCKITO).names("verify").withAnyParameters().build(),
+    MethodMatchers.create().ofTypes("org.mockito.InOrder").names("verify").withAnyParameters().build(),
+    MethodMatchers.create().ofTypes("org.mockito.stubbing.Stubber").names("when").withAnyParameters().build()
+  );
+
+  private static final MethodMatchers MOCKITO_EQ_MATCHERS = MethodMatchers.create()
```
m-g-sonar, line 54: `MOCKITO_EQ`

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,99 @@
...
+    return Collections.singletonList(Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodInvocationTree mit = (MethodInvocationTree) tree;
+
+    if (MOCKITO_EQ_USED_IN_ARGUMENTS.matches(mit)) {
+      ExpressionTree argument = mit.arguments().get(0);
+      if (argument.is(Tree.Kind.METHOD_INVOCATION)) {
```
m-g-sonar, line 69: be careful with nested parenthesis here

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,99 @@
...
+      }
+    } else if (MOCKITO_EQ_USED_IN_CONSECUTIVE_CALL.matches(mit)) {
+      MethodTreeUtils.consecutiveMethodInvocation(mit).ifPresent(m -> reportUselessEqUsage(m.arguments()));
+    }
+  }
+
+  private void reportUselessEqUsage(Arguments arguments) {
+    List<MethodInvocationTree> eqs = new ArrayList<>();
+
+    for (Tree t : arguments) {
```
m-g-sonar, line 80: skip parenthesis

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,99 @@
...
+      if (t.is(Tree.Kind.METHOD_INVOCATION) && MOCKITO_EQ_MATCHERS.matches((MethodInvocationTree) t)) {
+        eqs.add((MethodInvocationTree) t);
+      } else {
+        return;
```
m-g-sonar, line 84: worth a comment

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+
+    if (!eqs.isEmpty()) {
+      reportIssue(ExpressionUtils.methodName(eqs.get(0)), String.format(
+        "Remove %s useless \"eq\" invocation and directly use the value.", eqs.size() == 1 ? "this" : "these"),
```
m-g-sonar, line 90: I realize I forgot to add them in the rule description, but I tend to prefer adding parenthesis when discussing methods, to make it clear `"eq(...)"`

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,99 @@
...
+        eqs.stream()
+          .skip(1)
+          .map(eq -> new JavaFileScannerContext.Location("", ExpressionUtils.methodName(eq)))
```
m-g-sonar, line 93: I feel this will violate @nicolas-harraudeau-sonarsource 's rule about never having text in secondaries. Can you check?

java-checks-test-sources/src/main/java/checks/tests/MockitoEqSimplificationCheck.java
```diff
@@ -0,0 +1,94 @@
...
+    inOrder2.verify(foo).baz(v1, v2);
+
+    // No issue when mixed:
+
+    given(foo.bar(eq(v1), endsWith(""), v3)).willReturn(null);
+    given(foo.bar(eq(v1), any(), eq(v3))).willReturn(null);
+    given(foo.baz(eq(v1), captor.capture())).willReturn(null);
+
+    // When raw values are mixed with matchers, Mockito will throw InvalidUseOfMatchersException.
+    // S6073 should cover this case.
```
m-g-sonar, line 66: 👍🏻

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
nicolas-harraudeau-sonarsource: @m-g-sonar thanks for pinging me.
If I understand correctly we want to have secondary locations on every subsequent call to `eq()`. It seems to me that the calls to `given`, `when` and `verify` will rarely be split on many lines. If it is the case developers won't expect to navigate to `eq()` calls. They will all be visible right away.

As there is no need for navigation to these secondary locations, nor to to explain them (the primary message is clear enough), I would say that we don't need a secondary location message. Secondary locations are only there to improve issues' readability.
Does it make sense?

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
nicolas-harraudeau-sonarsource: Secondary locations are not highlighted for now in SonarLint for VSCode. Only a message is visible in the "Problems" bar. Thus I would change the primary message just a bit to replace "these". What do you think of:
"Remove this and every subsequent useless \"eq\" invocation; use directly the values."

That way we make more explicit what other "eq()" calls we refer to.

quentin-jaquier-sonarsource: Sounds good.

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoEqSimplificationCheck.java
quentin-jaquier-sonarsource: Make sense. And to be clear:

```
new JavaFileScannerContext.Location("", ExpressionUtils.methodName(eq))
```

is already a message-less secondary, as `methodName` returns a tree and is used for the location.

PR #3286, opened 2020-11-18
SONARJAVA-3588 Do not fail parsing of 'module-info.java' files if 'sonar.java.version' is misconfigured

m-g-sonar:


margarita-nedzelska-sonarsource (approved):
LGTM, just one thing in message.

PR #3287, opened 2020-11-18
SONARJAVA-3605 FP in S3305 when field has an initializer

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/java/checks/spring/SpringConfigurationWithAutowiredFieldsCheck.java
```diff
@@ -0,0 +1,133 @@
...
+  @Configuration
+  class A {
+
+    @Autowired private Bar singleUsage; // Noncompliant [[sc=28;ec=39]] {{Inject this field value directly into "method", the only method that uses it.}}
+    @Inject private Bar jsr330; // Noncompliant [[sc=25;ec=31]] {{Inject this field value directly into "jsr330", the only method that uses it.}}
+    @Autowired private Bar multipleUsage;
+    @Autowired private Bar notUsedInBeanMethod;
+    @Autowired private Bar notUsed;
+    private Bar notAutowired;
+    @Autowired(required=false) private Bar withInitializer = new Bar();
```
margarita-nedzelska-sonarsource, line 25: I'd suggest adding a case `@Autowired private Bar withInitializer = new Bar();` (maybe to non-compiling examples if doesn't compile) and decide what we want to do in such case.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringConfigurationWithAutowiredFieldsCheck.java
```diff
@@ -78,6 +78,10 @@ private static void collectAutowiredFields(Tree tree, Map<Symbol, VariableTree>
       return;
     }
     VariableTree variable = (VariableTree) tree;
+    if (variable.initializer() != null) {
```
margarita-nedzelska-sonarsource, line 81: As far as I remember Spring. `@Autowired` annotation by default has `required = true`. So the initializer won't make sense, because the project will fail if no `@Bean` found (or even won't compile). Maybe it will be nice to check that `required=false` there. On the other hand, it's an additional check, so maybe we can leave it as it is.

margarita-nedzelska-sonarsource (approved):
LGTM 👍

PR #3288, opened 2020-11-18
SONARJAVA-3622 Drop SE debug checks

m-g-sonar:


quentin-jaquier-sonarsource (approved):
LGTM

I like these changes, it's cleaner to not include debug feature in production!

PR #3290, opened 2020-11-19
SONARJAVA-3613 FP in S1193 when the catch block contains more code

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
I was wondering: if we use the "all or nothing" approach, wouldn't it be better to not report an issue per `instanceof` but one issue on the catch keyword with `instanceof` as secondaries? What do you think?

margarita-nedzelska-sonarsource:
> I was wondering: if we use the "all or nothing" approach, wouldn't it be better to not report an issue per `instanceof` but one issue on the catch keyword with `instanceof` as secondaries? What do you think?

I think this would make sense if we've just developed this rule. For now it might be confusing for our users if we change the way we're reporting. So I'd better leave the way we report issue at the moment

quentin-jaquier-sonarsource:
Yeah, make sense, the experience of the user will not be optional indeed, let's not change anything there.

margarita-nedzelska-sonarsource (approved):
LGTM!
Nice work 👍

PR #3291, opened 2020-11-20
SONARJAVA-3547 Fix cast recommendation with !Unknown! type

margarita-nedzelska-sonarsource:


java-checks-test-sources/src/main/java/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -143,6 +165,10 @@ void test2(Object param) {
     });
   }
 
+  private static <I, R> Function<I, R> castingIdentity() {
+    return i -> (R) i; // Compliant, 'R.class::cast' wont compile.
```
margarita-nedzelska-sonarsource, line 169:
```suggestion
    return i -> (R) i; // Compliant, 'R.class::cast' won't compile.
```

java-checks/src/main/java/org/sonar/java/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -130,12 +135,32 @@ private static boolean nullAgainstParam(ExpressionTree o1, ExpressionTree o2, Sy
...
+      return getTypeName(((ArrayTypeTree) type).type()).map(x -> x + "[]");
+    }
+    if (type.is(Tree.Kind.PRIMITIVE_TYPE)) {
+      return Optional.of(((PrimitiveTypeTree) type).keyword().text());
+    }
+    return Optional.empty();
+  }
 
+  private static boolean isGeneric(IdentifierTree identifierTree) {
+    return (identifierTree instanceof IdentifierTreeImpl) && ((IdentifierTreeImpl) identifierTree).binding.toString().matches("<[A-Z]>");
```
m-g-sonar, line 161: There is no such thing as an IdentifierTree not being an `IdentifierTreeImpl`, so the first test is useless.

Now, I'm afraid I don't understand what you try to achieve here.
Testing if we are in the presence of a raw type of a generic type? For instance `(List) obj` ?
Whatever it is, we have a way to express this through our own API rather than relying on the `toString()` of the eclipse binding.

margarita-nedzelska-sonarsource: That's a case when we're inside a parametrized method

```
  private static <I, R> Function<I, R> castingIdentity() {
    return i -> (R) i; // Compliant, 'R.class::cast' wont compile.
  }
```

'R.class' doesn't compile, but it's still catched as identifier.

m-g-sonar: So I think this method should help:

https://github.com/SonarSource/sonar-java/blob/master/java-frontend/src/main/java/org/sonar/java/model/JUtils.java#L94

margarita-nedzelska-sonarsource: Thanks, That helped :)

PR #3292, opened 2020-11-20
SONARJAVA-3482 Support character classes as operand to reluctant quantifier in rule S5857

margarita-nedzelska-sonarsource:
…

### DIFF
--- a/java-checks-test-sources/src/main/java/checks/regex/ReluctantQuantifierCheck.java
+++ b/java-checks-test-sources/src/main/java/checks/regex/ReluctantQuantifierCheck.java
@@ -7,17 +7,11 @@ public class ReluctantQuantifierCheck {
     str.matches("<\\S+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\s]++".}}
     str.matches("<\\D+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\d]++".}}
     str.matches("<\\W+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\w]++".}}
-    str.matches("<\\s+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\S]++".}}
-    str.matches("<\\d+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\D]++".}}
-    str.matches("<\\w+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\W]++".}}
 
     str.matches("<.{2,5}?>"); // Noncompliant [[sc=19;ec=26]] {{Replace this use of a reluctant quantifier with "[^>]{2,5}+".}}
     str.matches("<\\S{2,5}?>"); // Noncompliant [[sc=19;ec=28]] {{Replace this use of a reluctant quantifier with "[^>\\s]{2,5}+".}}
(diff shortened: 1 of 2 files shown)


java-checks-test-sources/src/main/java/checks/regex/ReluctantQuantifierCheck.java
```diff
@@ -4,7 +4,21 @@
 
   void noncompliant(String str) {
     str.matches("<.+?>"); // Noncompliant [[sc=19;ec=22]] {{Replace this use of a reluctant quantifier with "[^>]++".}}
+    str.matches("<\\S+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\s]++".}}
+    str.matches("<\\D+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\d]++".}}
+    str.matches("<\\W+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\w]++".}}
+    str.matches("<\\s+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\S]++".}}
+    str.matches("<\\d+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\D]++".}}
+    str.matches("<\\w+?>"); // Noncompliant [[sc=19;ec=24]] {{Replace this use of a reluctant quantifier with "[^>\\W]++".}}
```
sebastian-hungerecker-sonarsource, line 12: In these cases (and the later ones involving \s, \d and \w), the suggested replacement is too complicated because those classes don't contain `>`, so there's no need to exclude it. This also means that there's no potential for catastrophic backtracking behavior here, so using the `?` here is not really problematic, but also entirely unnecessary (the regex would behave the same without it and be somewhat more efficient).

So in cases like this there should either be no issue or the message should instead be something along the lines of "Make this regex greedy" or "Remove this '?'".

To check whether a character class overlaps with the character after it, `new SimplifiedRegexCharacterClass(base).intersects(new SimplifiedRegexCharacterClass(escapedClass), true)` can be used in the code.

sebastian-hungerecker-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/regex/ReluctantQuantifierCheck.java
```diff
@@ -104,9 +108,30 @@ private String makePossessive(Quantifier quantifier) {
       return Optional.of(result);
     }
 
+    private boolean hasNoIntersection(CharacterClassElementTree tree, @Nullable CharacterClassElementTree base) {
+      return base != null && !new SimplifiedRegexCharacterClass(base, new FlagSet()).intersects(new SimplifiedRegexCharacterClass(tree, new FlagSet()), false);
```
sebastian-hungerecker-sonarsource, line 112: Instead of `new FlagSet()`, we should be calling `getFlagSet()` here. Otherwise it won't take flags into account for cases like `"(?i)[a-z]*?X"` or `Pattern.compile("[a-z]*?X", Pattern.CASE_INSENSITIVE)`.

Even with that change we can still get wrong results if a flag is only enabled for part of the regex (e.g. `(?i:[a-z]*?)X` where the `i` flag would already be unset by the time we call `getFlagSet`), but that's actually pretty tricky to handle with the current API. @alban-auzeill  is currently working on a change that will make it easy to query the flags for a specific RegexTree, so it's best to just use `getFlags()` for now and then we can replace it with `base.getFlags()` and `tree.getFlags()` respectively afterwards once Alban adds that method.

sebastian-hungerecker-sonarsource (changes requested):

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3294, opened 2020-11-23
SONARJAVA-3608 Report on the methodSelect instead of only name

m-g-sonar:
Mockito's Matchers.eq() method is a static method, so there is no risk of
highlighting too much when raising on the 'methodSelect()' part of the
method invocations. Such calls can not be chained.


margarita-nedzelska-sonarsource (approved):
LGTM

PR #3295, opened 2020-11-23
SONARJAVA-2154 Remove Guava dependency from java-checks module

margarita-nedzelska-sonarsource:


java-frontend/src/main/java/org/sonar/java/annotations/VisibleForTesting.java
```diff
@@ -0,0 +1,23 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java.annotations;
+
+public @interface VisibleForTesting {
```
m-g-sonar, line 22: probably worth at least a comment on the utility of this new annotation, and how to use it internally.

java-checks/src/main/java/org/sonar/java/checks/VisibleForTestingUsageCheck.java
```diff
@@ -42,7 +42,8 @@
     "com.google.common.annotations.VisibleForTesting",
     "org.assertj.core.util.VisibleForTesting",
     "androidx.annotation.VisibleForTesting",
-    "org.apache.flink.annotation.VisibleForTesting"
+    "org.apache.flink.annotation.VisibleForTesting",
+    "org.sonar.java.annotation.VisibleForTesting"
```
m-g-sonar, line 46: I think that's not the best possible fix. This annotation is internal machinery, and it should probably not be included directly in the rule, to not encourage people using OUR implementation (even if the annotation is not described in the rule, some people might look at our open-source code to get the list of supported annotations).

What about another angle, and consider all the annotations called `@VisibleForTested`, whatever their fully qualified name? I think it would be cleaner (but requires another ticket)

margarita-nedzelska-sonarsource:
There are still  2 usages in test directory. But it's ok, cause it won't influence plugin size.
The dependency in pom can be removed, but we still have guava coming from java-frontend.

java-checks/src/main/java/org/sonar/java/checks/VisibleForTestingUsageCheck.java
margarita-nedzelska-sonarsource: Created a ticket for it https://jira.sonarsource.com/browse/SONARJAVA-3623

m-g-sonar (approved):

PR #3297, opened 2020-11-24
SONARJAVA-3564 Implement intersects and supersetOf helper for regex automata

alban-auzeill:


java-frontend/src/main/java/org/sonar/java/regex/RegexParser.java
```diff
@@ -81,6 +83,7 @@ public RegexParser(RegexSource source, FlagSet initialFlags) {
   }
 
   public RegexParseResult parse() {
+    this.activeFlags = initialFlags;
```
sebastian-hungerecker-sonarsource, line 86: It probably makes sense to initialize `activeFlags` in the constructor rather than `parse` and to get rid of `initialFlags` altogether (this would also entail getting rid of `RegexParseResult.initialFlags`, which should be fine because we can just use `parseResult.getResult().activeFlags()` instead).

java-frontend/src/main/java/org/sonar/java/regex/ast/DisjunctionTree.java
```diff
@@ -30,7 +30,7 @@
   private final List<JavaCharacter> orOperators;
 
   public DisjunctionTree(RegexSource source, IndexRange range, List<RegexTree> alternatives, List<JavaCharacter> orOperators) {
-    super(source, range);
+    super(source, range, FLAGS_HAVE_NO_EFFECT_ON_THIS_STATE);
```
sebastian-hungerecker-sonarsource, line 33: If we decide to remove `RegexParseResult.initialFlags`, `RegexTree`s should never use `FLAGS_HAVE_NO_EFFECT_ON_THIS_STATE`.

sebastian-hungerecker-sonarsource (changes requested):
Looks great so far. In addition to the individual comments I've left, you should also rebase onto master to get Margo's latest changes that use `RegexBaseVisitor.flagActive` to fix those as well (and add a test case that `(?i:[a-z]*?)X` now raises an issue).

java-frontend/src/main/java/org/sonar/java/regex/ast/DisjunctionTree.java
alban-auzeill: I have set correct activeFlags on all AutomatonState and removed FLAGS_HAVE_NO_EFFECT_ON_THIS_STATE

java-frontend/src/main/java/org/sonar/java/regex/RegexParser.java
alban-auzeill: I removed the field `initialFlags` but keept the getter `getInitialFlags()`.

alban-auzeill:
> rebase onto master to get Margo's latest changes that use `RegexBaseVisitor.flagActive` to fix those as well (and add a test case that `(?i:[a-z]*?)X` now raises an issue).

S5857 seems to only support DOT and ESCAPED_CHARACTER_CLASS before the `*?`
So I added this case instead `(?i:.*?[a-z])`

java-checks/src/main/java/org/sonar/java/checks/regex/PossessiveQuantifierContinuationCheck.java
```diff
@@ -0,0 +1,74 @@
...
+    private final FinalState finalState;
+
+    public Visitor(FinalState finalState) {
+      this.finalState = finalState;
+    }
+
+    @Override
+    public void visitRepetition(RepetitionTree repetitionTree) {
+      AutomatonState continuation = repetitionTree.continuation();
+      if (doesRepetitionContinuationAlwaysFail(repetitionTree) && continuation instanceof RegexSyntaxElement) {
```
sebastian-hungerecker-sonarsource, line 57: I believe we should still report the issue if the continuation is not a RegexSyntaxElement. If the non-regex-syntax continuation has a single successor (i.e. it's not a branch), we can just report on that successor instead (or, if it's still not a regex syntax element, keep following single edges until we find one or a branch). And if it's a branch (as might be the case for something like `(xx++)+x`) we should probably still report by only reporting on the repetition itself with the message adjusted accordingly.

sebastian-hungerecker-sonarsource (changes requested):

java-checks-test-sources/src/main/java/checks/regex/ReluctantQuantifierCheck.java
```diff
@@ -21,7 +21,8 @@ void noncompliant(String str) {
     str.matches(".*?\\P{L}"); // Noncompliant [[sc=18;ec=21]] {{Replace this use of a reluctant quantifier with "\\p{L}*+".}}
     str.matches("\\[.*?\\]"); // Noncompliant [[sc=21;ec=24]] {{Replace this use of a reluctant quantifier with "[^\\]]*+".}}
     str.matches(".+?[abc]"); // Noncompliant [[sc=18;ec=21]] {{Replace this use of a reluctant quantifier with "[^abc]++".}}
-    
+    str.matches("(?i:.*?[a-z])"); // Noncompliant [[sc=22;ec=25]] {{Replace this use of a reluctant quantifier with "[^a-z]*+".}}
```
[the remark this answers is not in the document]
alban-auzeill: Unfortunately `ReluctantQuantifierCheck` only supports `DOT` and `ESCAPED_CHARACTER_CLASS` before the `*?`, and only `PLAIN_CHARACTER`, `ESCAPED_CHARACTER_CLASS`, `CHARACTER_CLASS` after.
So it was not possible to have different active flags before and after `*?`. For this, I had to change the rule implementation :disappointed: to support `NON_CAPTURING_GROUP`.
See this commit: `SONARJAVA-3564 ReluctantQuantifierCheck support mixed active flags`

sebastian-hungerecker-sonarsource (approved):
LGTM

java-checks/src/test/java/org/sonar/java/checks/helpers/SimplifiedRegexCharacterClassTest.java
```diff
@@ -52,4 +69,226 @@ void testIntersectionWithTrueAsDefaultAnswer() {
...
+    // but a single code point
+    assertThat(maxCodePoint.codePoints().count()).isEqualTo(1);
+    // and java Pattern support it
+    assertThat(maxCodePoint).matches(".");
+
+    RegexTree result = parseRegex(maxCodePoint, new FlagSet()).getResult();
+    // FIXME result should be a single PlainCharacterTree and not a SequenceTree of two PlainCharacterTree s
+    // In oder to assert: assertIntersects(".", maxCodePoint, false, NO_FLAGS).isTrue();
+    // Instead of:
+    assertThat(result).isInstanceOf(SequenceTree.class);
```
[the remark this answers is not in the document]
alban-auzeill: Created https://jira.sonarsource.com/browse/SONARJAVA-3624

PR #3298, opened 2020-11-24
SONARJAVA-3623 Update rule S5803 to support all annotations named @VisibleForTesting

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/VisibleForTestingUsageCheck.java
```diff
@@ -60,7 +53,7 @@ public void visitNode(Tree tree) {
     }
     if (isMisusedVisibleForTesting(symbol)) {
       List<JavaFileScannerContext.Location> locations = symbol.usages().stream()
-        .filter(identifierTree -> isMisusedVisibleForTesting(identifierTree.symbol()))
```
margarita-nedzelska-sonarsource, line 63: already checked in line 54

PR #3299, opened 2020-11-24
SONARJAVA-3615 Fix FP in S1905 when casted argument is a method reference to a method with varargs argument

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -193,6 +193,10 @@ private static boolean hasOverloads(Symbol symbol) {
     return owner.isTypeSymbol() && calcOverloads((Symbol.TypeSymbol) owner, symbol.name()) > 1;
   }
 
+  private static boolean hasVarargArg(Symbol symbol) {
+    return symbol.isMethodSymbol() && ((Symbol.MethodSymbol) symbol).parameterTypes().stream().anyMatch(Type::isArray);
```
m-g-sonar, line 197: Checking for any array is too much here. Varargs can only be placed on last parameter. There is also a dedicated method in JUtils for that:
```
return symbol.isMethodSymbol() && JUtils.isVarArgsMethod((Symbol.MethodSymbol) symbol);
```

m-g-sonar (changes requested):
Coverage should be at 95% minimum. In order to pass the "isMethodSymbol()", the best way is to use an unknown (non-resolved) method

m-g-sonar (approved):

PR #3301, opened 2020-11-26
SONARJAVA-3567 Rule S6019: Reluctant quantifiers in regular expressions should be followed by an expression that can't match the empty string

sebastian-hungerecker-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/regex/AbstractRegexCheckTrackingMatchers.java
```diff
@@ -0,0 +1,224 @@
...
+    nodes.add(Tree.Kind.COMPILATION_UNIT);
+    return nodes;
+  }
+
+  @Override
+  public void leaveNode(Tree tree) {
+    if (tree.is(Tree.Kind.RETURN_STATEMENT)) {
+      collectReturnedVariables(((ReturnStatementTree) tree).expression());
+    } else if (tree.is(Tree.Kind.COMPILATION_UNIT)) {
+      for (Map.Entry<RegexParseResult, List<MethodInvocationTree>> entry : methodsCalledOnRegex.entrySet()) {
```
m-g-sonar, line 123: Looks to me that we could simplify this using a `foreach`:
```java
methodsCalledOnRegex.forEach((regex, invocations) -> checkRegex(regex, invocations, escapingRegexes.contains(regex)));
```

java-checks/src/main/java/org/sonar/java/checks/regex/AbstractRegexCheckTrackingMatchType.java
```diff
@@ -0,0 +1,79 @@
...
+      .ofTypes("java.util.regex.Pattern", "java.util.regex.Matcher", "java.lang.String")
+      .names("matches", "asMatchPredicate")
+      .withAnyParameters()
+      .build();
+
+  protected abstract void checkRegex(RegexParseResult regex, MatchType matchType);
+
+  @Override
+  protected MethodMatchers trackedMethodMatchers() {
+    return MethodMatchers.or(PARTIAL_MATCHERS, FULL_MATCHERS);
```
m-g-sonar, line 56: this should probably be stored in a constant, because `trackedMethodMatchers()` is called way too many times, and we probably don't want to perform the `or()` all the time.

Another option could be to cache it in `AbstractRegexCheckTrackingMatchers`, and then only call the cached one.

java-checks/src/main/java/org/sonar/java/checks/regex/AbstractRegexCheckTrackingMatchers.java
```diff
@@ -0,0 +1,224 @@
...
+import org.sonar.plugins.java.api.tree.AssignmentExpressionTree;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.ReturnStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+public abstract class AbstractRegexCheckTrackingMatchers extends AbstractRegexCheck {
```
m-g-sonar, line 47: I understand the problem of levels of hierarchy... and in fact, I believe we should modify the parent `AbstractRegexCheck` to not anymore rely on `AbstractMethodDetection`.

We already specialized the behavior of the initial `AbstractMethodDetection` too much. We would benefit in doing it simpler, by focusing only on method invocations directly in the `AbstractRegexCheck`.

m-g-sonar:
LGTM!

PR #3302, opened 2020-11-26
SONARJAVA-3621 Workaround unknown types in catch block union types

m-g-sonar:


quentin-jaquier-sonarsource (commented):
As discussed, it would make sense to add a unit test on the semantic side, to test this specific behavior and be aware if it changes at one point.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3303, opened 2020-11-27
SONARJAVA-3566 Rule S5855: Regex alternatives should not be redundant

alban-auzeill:


sebastian-hungerecker-sonarsource (approved):
Looks great and kudos for implementing relative locations for secondaries! Rebase and merge after merging SONARJAVA-3564.

PR #3304, opened 2020-11-27
SONARJAVA-3619 FP S2589 when Boolean variable doesn't always evaluate to TRUE/FALSE

margarita-nedzelska-sonarsource:


java-frontend/src/main/java/org/sonar/java/se/constraint/ConstraintManager.java
```diff
@@ -153,6 +157,14 @@ private SymbolicValue createIdentifierSymbolicValue(IdentifierTree identifier) {
     }
     return createDefaultSymbolicValue();
   }
+  
+  private SymbolicValue createVariableSymbolicValue(VariableTree identifier) {
+    final Type type = identifier.symbol().type();
+    if (type != null && type.is("java.lang.Boolean")) {
```
m-g-sonar, line 163: I'm really not sure about this `null` test here, I think it's impossible. At worst, an unknown symbol should return an unknown type., and so your test can probably simply be `if (type.is("java.lang.Boolean")) { ...`

java-frontend/src/main/java/org/sonar/java/se/constraint/ConstraintManager.java
```diff
@@ -153,6 +157,14 @@ private SymbolicValue createIdentifierSymbolicValue(IdentifierTree identifier) {
     }
     return createDefaultSymbolicValue();
   }
+  
+  private SymbolicValue createVariableSymbolicValue(VariableTree identifier) {
```
m-g-sonar, line 161: I would rename `identifier` into `variable`

java-frontend/src/main/java/org/sonar/java/se/symbolicvalues/RelationalSymbolicValue.java
```diff
@@ -172,35 +172,64 @@ RelationalSymbolicValue inverse() {
...
 
+  private boolean shouldAddNullConstraint(@Nullable BooleanConstraint constraint1, @Nullable BooleanConstraint constraint2) {
+    return !isEquality() && constraint1 == null && constraint2 != null;
+  }
+
+  private static List<ProgramState> getProgramStatesWithNullConstraints(List<ProgramState> copiedConstraints, SymbolicValue operand) {
+    List<ProgramState> nullConstraints = copiedConstraints.stream()
+      .flatMap(ps -> operand.setConstraint(ps, ObjectConstraint.NULL).stream())
+      .map(ps -> ps.removeConstraintsOnDomain(operand, BooleanConstraint.class)
+      ).collect(Collectors.toList());
```
m-g-sonar, line 205: Question of style, but I would really prefer to have the parenthesis on the previous line and this one starting with `.collect(` to align dots

java-frontend/src/main/java/org/sonar/java/se/symbolicvalues/RelationalSymbolicValue.java
```diff
@@ -172,35 +172,64 @@ RelationalSymbolicValue inverse() {
     return results;
   }
 
+  private boolean hasBooleanSymbolicValueAsOperand() {
+    return (leftOp instanceof BooleanSymbolicValue) || (rightOp instanceof BooleanSymbolicValue);
+  }
+
+  private boolean isEqualityCheck() {
+    return Kind.METHOD_EQUALS == kind || Kind.NOT_METHOD_EQUALS == kind || Kind.EQUAL == kind || Kind.NOT_EQUAL == kind;
```
m-g-sonar, line 180: What about:
```
return isEquality() || Kind.NOT_METHOD_EQUALS == kind || Kind.NOT_EQUAL == kind;
```
In order to mark the difference with isEquality ?

m-g-sonar (commented):
Almost there! Good job! Just a few minor things to fix.

java-frontend/src/main/java/org/sonar/java/se/symbolicvalues/RelationalSymbolicValue.java
margarita-nedzelska-sonarsource: Nice idea!

java-frontend/src/main/java/org/sonar/java/se/constraint/ConstraintManager.java
margarita-nedzelska-sonarsource: Good old copy/paste)

PR #3305, opened 2020-11-27
SONARJAVA-3561 AbstractRegexCheck should target more regex providers

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/regex/UnusedGroupNamesCheck.java
```diff
@@ -247,7 +248,13 @@ private void checkNoSuchName(KnownGroupsCollector knownGroups, ExpressionTree ar
   }
 
   @Override
-  public void checkRegex(RegexParseResult regexForLiterals, MethodInvocationTree mit) {
+  protected boolean filterAnnotation(AnnotationTree annotation) {
+    return false;
```
sebastian-hungerecker-sonarsource, line 252: I'd argue that we don't want to ignore annotations here: If a regex in an annotation defines a named group that's not used by a backreference, the named group is pointless and we should report it. That said, this rule is completely reworked in SONARJAVA-3567, so there's no point in changing anything here - I'll take care of it on my side.

sebastian-hungerecker-sonarsource (approved):
Nice work.

PR #3306, opened 2020-11-30
Extend S1181 to be fully appliable to ERR08-J

kjdavid:

Actually it fails the JspTest, but at my side it fails without my modifications so I hope it doesn't count.
image

I'm trying to implement rules from [this page](https://wiki.sei.cmu.edu/confluence/display/java/SEI+CERT+Oracle+Coding+Standard+for+Java), as I expanded your partially working rule [S1181](https://rules.sonarsource.com/java/RSPEC-1181) to be fully appliable to [ERR08-J. Do not catch NullPointerException or any of its ancestors](https://wiki.sei.cmu.edu/confluence/display/java/ERR08-J.+Do+not+catch+NullPointerException+or+any+of+its+ancestors) with [S1696](https://rules.sonarsource.com/java/RSPEC-1696). If you want I can merge these two rules into one.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CatchOfThrowableOrErrorCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CatchOfThrowableOrErrorCheck.java
@@ -69,7 +69,7 @@ public class CatchOfThrowableOrErrorCheck extends IssuableSubscriptionVisitor {
 
   private void checkType(TypeTree typeTree, CatchTree catchTree) {
     Type type = typeTree.symbolType();
-    if (type.is("java.lang.Error")) {
+    if (type.is("java.lang.Error") || type.is("java.lang.Exception") || type.is("java.lang.RuntimeException")) {
       insertIssue(typeTree, type);
     } else if (type.is(JAVA_LANG_THROWABLE)) {
       GuavaCloserRethrowVisitor visitor = new GuavaCloserRethrowVisitor(catchTree.parameter().symbol());
@@ -81,7 +81,7 @@ public class CatchOfThrowableOrErrorCheck extends IssuableSubscriptionVisitor {
   }
 
   private void insertIssue(TypeTree typeTree, Type type) {
-    reportIssue(typeTree, "Catch Exception instead of " + type.name() + ".");
+    reportIssue(typeTree, "Catch the proper exception instead of " + type.name() + ".");
   }
 
   private static class GuavaCloserRethrowVisitor extends BaseTreeVisitor {
--- a/java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1181_java.html
+++ b/java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1181_java.html
@@ -17,7 +17,7 @@ try { /* ... */ } catch (MyException e) { /* ... */ }
 <h2>See</h2>
 <ul>
   <li> <a href="http://cwe.mitre.org/data/definitions/396.html">MITRE, CWE-396</a> - Declaration of Catch for Generic Exception </li>
-  <li> <a href="https://www.securecoding.cert.org/confluence/x/BIB3AQ">CERT, ERR08-J.</a> - Do not catch NullPointerException or any of its ancestors
+  <li> <a href="https://wiki.sei.cmu.edu/confluence/x/_TdGBQ">CERT, ERR08-J.</a> - Do not catch NullPointerException or any of its ancestors
   </li>
 </ul>
 
--- a/java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1181_java.html
+++ b/java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1181_java.html
@@ -2,14 +2,16 @@
 meant to be caught by applications.</p>
 <p>Catching either <code>Throwable</code> or <code>Error</code> will also catch <code>OutOfMemoryError</code> and <code>InternalError</code>, from
 which an application should not attempt to recover.</p>
+<p>Programs must not catch the <code>NullPointerException</code>, so it is suggested to not catch any of its ancestor.</p>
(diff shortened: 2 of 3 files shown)


quentin-jaquier-sonarsource:
Hello, thanks for raising this question.

First, let me just clarify that the references cited in the rules are not meant to perfectly describe the scope of the rule, they are here to describe the problem from another perspective, give examples of what can go wrong, and so on.

Then concerning S1696, I'm not sure we can merge it, or extend it to report `RuntimeException` as well. To me, this is a different situation, in the first one, you are explicitly catching a NPE, there is a clear alternative as described in the rule. In the second one, you might inadvertently catch a NPE, but the situation is not as clear and the solution is not the same. It seems fine to keep it as it is.

Also, note that [S2221]( https://rules.sonarsource.com/java/RSPEC-2221) is already covering the case of `Exception`.

Finally, the only remaining case is catching `RuntimeException`. Indeed, this is something that should be used with care, since subject to the problem described in the different references. But this seems quite often used legitimately and does not feel as bad as catching an `Error` or a `Throwable`. I'm afraid this change will generate too much noise.

Does this make sense to you?

kjdavid:
Thank you for your comment @quentin-jaquier-sonarsource.

I admit that I overlooked the [S2221](https://rules.sonarsource.com/java/RSPEC-2221), I apolgize for that. 

I understand your concern of the noise that it could generate, but I would like to point out that security issues should have more priority than noise reduction, because our world relies on softwares almost everywhere, and this dependence between the world and the softwares will be tighter as time passes. If developers don't consider `RuntimeException` as a throwable exception, only as a wrapper class, this noise also can be reduced. As Sonar is a software which gives hints, and suggestions on how the code is more reliable, I think this is the way it should work. If we just put it in as a separate optional checker it can notice several issues which makes software quality better.

So, to sum it up, I'll remove the duplication due to [S2221](https://rules.sonarsource.com/java/RSPEC-2221) and run some tests on some bigger open source project to see how much noise will appear. I'll post the results here, after that you can decide whether I should make a separate checker for this `RuntimeException` case or it can stay together.

kjdavid:
Hi @quentin-jaquier-sonarsource,

Sorry for not checking in earlier, I had some trouble with builidng the Sonar-java into SonarLint, because I couldn't find anything about it in the documentations or just looked over it.
I removed the duplication and after that I ran the test which provided these results:

- RxJava

image

image

- Jenkins

image

image

- Elastic-Search

image

image

As we can see in these projects this change doesn't have too much effect on the number of issues compared to the total number of issues. I found 0 in RxJava, 54 (~0.67%) in Jenkins and 876 in Elastic search (~1.29%).
What do you think, should I put in a new rule or should we keep it merged instead?

quentin-jaquier-sonarsource:
What bothers me with your experiment is that the rule is already raising thousands of issues on these projects, so the numbers will always be pretty low. That almost 100 issues, maybe they are all perfectly fine? And to be clear, I'm absolutely not claiming that they have poor quality, but that this problem does not seem to be relevant for them, doing any change in this rule would anyway not have an impact for them, since it will be lost in the noise.

My feeling is that we should not touch [RSPEC-1696](https://rules.sonarsource.com/java/RSPEC-1696) or [RSPEC-2221](https://rules.sonarsource.com/java/RSPEC-2221), the problem they are targeting is not the same.
[RSPEC-1181](https://rules.sonarsource.com/java/RSPEC-1181) is closer, but I think there is still a difference between catching an `Error` and catching a `RuntimeException`. The former is, without doubt, a code smell, while the latter seems more acceptable.

So we could go for a new rule but... will it bring real value?

I'm not so sure, there are plenty of good reasons to catch a `Runtime`, and I'm afraid I would myself be reluctant to change the code. It can indeed warn an inexperienced user that something bad can happen, but this sounds minor compared to the value.

kjdavid:
> What bothers me with your experiment is that the rule is already raising thousands of issues on these projects

I don't quite understand your problem, because as I mentioned in my previous comment the sum of the 3 projects' newly detected issues are less than a thousand and there was also a project which had 0 new issues. Actually I didn't run only this rule, because I didn't know how to run it separately. I ran a whole analysis with my modification, and one without it. This found every smelling code that sonar-java could notice, and in the 2nd case it found the additional `RuntimeException`s in the catch conditions.

Little notice about SonarLint: Even if I used the **/*{!.java} exclusions for files, it checked everything, but when I exclude folders that's working, which is weird.

> So we could go for a new rule but... will it bring real value?

Yes, I think it will.

> I'm not so sure, there are plenty of good reasons to catch a Runtime, and I'm afraid I would myself be reluctant to change the code. It can indeed warn an inexperienced user that something bad can happen, but this sounds minor compared to the potential false positives.

I totally agree with this, and I think Sonar should mention that this could be wrong and the developer or the project manager should be able to decide whether they fix it or not. Giving people more options is never a bad thing. That's why previously I wrote that it could be an optional and separate checker, I just need to know whether you accept this or not.
> What do you think, should I put in a new rule or should we keep it merged instead?

quentin-jaquier-sonarsource:
>I didn't run only this rule

Okay, I missed that. So that 54 issues for Jenkins and 76 for Elastic. For the record, S1181 is raising 122 for Jenkins and 48 for Elastic (without changes). Even if it's a small percentage of the total issue, we are still speaking about hundreds of issues and we are not sure if they are false positive or not. We have more than 600 rules, we should be really careful when raising too many issues, to not hide other more interesting ones.

In any way, I am convinced that we should **not** modify any existing rule, as it would not fit well any of them.
Now what about creating a new rule? I am only half convinced. I understand the idea and the argument of having the choice, but at the same time, all the issues it would report on our own code would simply be noise, and I don't see any way to be sure to report interesting issues.

At this point, I believe it could be great to have more input from other users, to see if there is a common need. I will close this PR, as it anyway diverged from the original topic. If anyone has additional opinions, please reach the [SonarSource community](https://community.sonarsource.com/), it is the best place to discuss such topic.

PR #3309, opened 2020-12-01
SONARJAVA-3130 S3824: raise issue when "containsKey" is used

margarita-nedzelska-sonarsource:


its/ruling/src/test/resources/eclipse-jetty/java-S3824.json
```diff
@@ -1,4 +1,7 @@
 {
+'org.eclipse.jetty:jetty-project:jetty-jmx/src/main/java/org/eclipse/jetty/jmx/ConnectorServer.java':[
+127,
```
m-g-sonar, line 3: I'm a bit surprised that we are not catching the exact same pattern a few lines above (lines 120 and 125). I feel it's due to the fact that we are using constants as `MemberSelectExpressionTree`, which should be exactly the same.

Can you check that if the problem is on the SE engine side (the way we associate constants with SVs) or if it comes from the rule itself?

If it comes from the SE engine, then we need a ticket.

its/ruling/src/test/resources/sonar-server/java-S3824.json
```diff
@@ -11,10 +15,20 @@
 'org.sonarsource.sonarqube:sonar-server:src/main/java/org/sonar/server/issue/notification/NewIssuesStatistics.java':[
 53,
 ],
+'org.sonarsource.sonarqube:sonar-server:src/main/java/org/sonar/server/issue/ws/SearchAction.java':[
+408,
```
m-g-sonar, line 19: very nice findings!

java-checks-test-sources/src/main/java/symbolicexecution/checks/MapComputeIfAbsentOrPresentCheck.java
```diff
@@ -1,11 +1,14 @@
+package symbolicexecution.checks;
+
 import com.google.common.base.Preconditions;
 
 import javax.annotation.CheckForNull;
 
 import java.util.Map;
 import java.util.Objects;
+import org.elasticsearch.common.MacAddressProvider;
```
m-g-sonar, line 9: unused import

java-frontend/src/main/java/org/sonar/java/se/checks/MapComputeIfAbsentOrPresentCheck.java
```diff
@@ -153,60 +184,99 @@ public void checkEndOfExecution(CheckerContext context) {
     checkIssues.stream().filter(checkIssue -> checkIssue.isOnlyPossibleIssueForReportTree(checkIssues)).forEach(issue -> issue.report(context, check));
   }
 
-  private static class CheckIssue {
+  private abstract static class CheckIssue {
```
m-g-sonar, line 187: Love this part! pretty nice job!

java-frontend/src/main/java/org/sonar/java/se/checks/MapComputeIfAbsentOrPresentCheck.java
```diff
@@ -126,15 +151,21 @@ private static boolean isMethodInvocationThrowingCheckedException(ExpressionTree
   }
 
   private static boolean isInsideIfStatementWithNullCheckWithoutElse(MethodInvocationTree mit) {
+    return getIfStatementParent(mit).map(ifStatementTree -> ifStatementTree.elseStatement() == null
```
m-g-sonar, line 154: From how it is implemented, `getIfStatementParent(mit)` can be pretty costly (it potentially explore the whole AST), and we are going to make such calls for each `map.containsKey()` or `map.get()`.

What about precomputing them or coaching them to avoid having to reexplore the AST for all these invocations?

m-g-sonar (commented):
Nice work! A few changes required, but looks pretty good already!

its/ruling/src/test/resources/eclipse-jetty/java-S3824.json
margarita-nedzelska-sonarsource: I checked and we do not report this issues because SV associated with key is different in containsKey and in put, despite it's actually the same constant. Also true for Enums and reproduces with Map.get()

It seems to come from engine :( Will create a ticket.

margarita-nedzelska-sonarsource: Created a ticket https://jira.sonarsource.com/browse/SONARJAVA-3626

m-g-sonar (approved):

PR #3312, opened 2020-12-02
SONARJAVA-3609 Make sure JParser releases jars after parsing

m-g-sonar:


henryju (approved):
Tested with SonarLint on Windows. Files are no more locked after the analysis has completed.

PR #3314, opened 2020-12-02
SONARJAVA-3599 FP in S2226 for non final Servlet fields initialized in init() method without parameters

margarita-nedzelska-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3315, opened 2020-12-02
SONARJAVA-3470 Fix FP rule S107 by adding more ignored annotations

margarita-nedzelska-sonarsource:

### DIFF
--- a/java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S107_java.html
+++ b/java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S107_java.html
@@ -14,6 +14,13 @@ public void doSomething(int param1, int param2, int param3, String param4) {
 }
 </pre>
 <h2>Exceptions</h2>
-<p>Methods annotated with Spring's <code>@RequestMapping</code> (and related shortcut annotations, like <code>@GetRequest</code>) or
-<code>@JsonCreator</code> may have a lot of parameters, encapsulation being possible. Such methods are therefore ignored.</p>
+<p>Methods annotated with :</p>
+<ul>
+  <li> Spring's <code>@RequestMapping</code> (and related shortcut annotations, like <code>@GetRequest</code>) </li>
+  <li> JAX-RS API annotations (like <code>@javax.ws.rs.GET</code>) </li>
+  <li> Bean constructor injection with <code>@org.springframework.beans.factory.annotation.Autowired</code> </li>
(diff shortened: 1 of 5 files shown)


margarita-nedzelska-sonarsource:
In the ticket there is a suggestion to add one more parameter to the rule but actually I have doubts whether it will bring much value

margarita-nedzelska-sonarsource:
After discussing with @m-g-sonar we decided not to add an additional parameter to the rule, because all the cases were about JAX-RS annotations and we covered it here. And The general rule is to keep checks simple.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S107_java.html
```diff
@@ -14,6 +14,11 @@ <h2>Compliant Solution</h2>
...
 </pre>
 <h2>Exceptions</h2>
-<p>Methods annotated with Spring's <code>@RequestMapping</code> (and related shortcut annotations, like <code>@GetRequest</code>) or
-<code>@JsonCreator</code> may have a lot of parameters, encapsulation being possible. Such methods are therefore ignored.</p>
+<p>Methods annotated with :</p>
+<p> * Spring's <code>@RequestMapping</code> (and related shortcut annotations, like <code>@GetRequest</code>)</p>
+<p> * JAX-RS API annotations (like <code>@javax.ws.rs.GET</code>)</p>
+<p> * Bean constructor injection with <code>@org.springframework.beans.factory.annotation.Autowired</code></p>
+<p> * CDI constructor injection with <code>@javax.inject.Inject</code></p>
+<p> * <code>@com.fasterxml.jackson.annotation.JsonCreator</code></p>
```
quentin-jaquier-sonarsource, line 22: This should be an HTML list.
It's a ruleapi bug, you have to remove the space before the `*` in the description.

I already did it, you just have to generate the metadata to solve the problem.

quentin-jaquier-sonarsource (approved):
Also make sure that the [description of the ticket](https://jira.sonarsource.com/browse/SONARJAVA-3470) is up to date with the current list.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S107_java.html
margarita-nedzelska-sonarsource: Thanks, didn't know about it 👍

PR #3316, opened 2020-12-02
SONARJAVA-2154 Remove usages of guava from sonar-java-plugin module

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
Removed gson dependency as it is transitively coming from java-frontend module

m-g-sonar (approved):

m-g-sonar:
LGTM!

PR #3318, opened 2020-12-03
SONARJAVA-3560 Rule S6002: Regex lookahead assertions should not be contradictory

alban-auzeill:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3319, opened 2020-12-04
Add RegexDatabase in ITs

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/regex/ReluctantQuantifierCheck.java
```diff
@@ -64,7 +64,7 @@ public void visitSequence(SequenceTree tree) {
           .ifPresent(negatedClass -> {
             String newQuantifier = makePossessive(repetition.getQuantifier());
             String message = String.format("Replace this use of a reluctant quantifier with \"%s%s\".", negatedClass, newQuantifier);
-            reportIssue(repetition, String.format(message, negatedClass), null, Collections.emptyList());
+            reportIssue(repetition, message, null, Collections.emptyList());
```
alban-auzeill, line 67: Fix this bug making the rule throwing exceptions on some regex containing `^`

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexSource.java
```diff
@@ -61,7 +61,7 @@ public int length() {
     return sourceText.length();
   }
 
-  List<LiteralTree> getStringLiterals() {
+  public List<LiteralTree> getStringLiterals() {
```
alban-auzeill, line 64: Required by the export script that collects string literals

its/ruling/src/test/resources/regex-examples/java-S5994.json
```diff
@@ -0,0 +1,5 @@
+{
+'org.regex-examples:regex-examples:src/main/java/org/regex/examples/RegexDatabase5.java':[
+135,
```
alban-auzeill, line 3: Seems to be a false-positive, I will investigate

sebastian-hungerecker-sonarsource: The possessive quantifier check should be changed to not apply to repetitions with a fixed upper bound.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3320, opened 2020-12-04
Fix junit tests not running in Intellij Idea 2020.3 by bumping junit jupiter version

alban-auzeill:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3321, opened 2020-12-04
Fix quality flaw: use method references and assertJ simplification

quentin-jaquier-sonarsource:


margarita-nedzelska-sonarsource (approved):
LGTM :)

PR #3322, opened 2020-12-04
SONARJAVA-3130 Fix FP when the value put in map is null literal

margarita-nedzelska-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3326, opened 2020-12-07
SONARJAVA-2154 Remove usages of guava annotations from java-frontend module

margarita-nedzelska-sonarsource:


java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
```diff
@@ -305,4 +304,21 @@ private final void forEach(Collection<SubscriptionVisitor> visitors, Consumer<Su
       }
     }
   }
+  private static Throwable getRootCause(Throwable throwable) {
```
quentin-jaquier-sonarsource, line 307: If I understand correctly, this code is inspired by the one from Guava, and is used to detect loops at the same time. Still, I believe it could use a few comments explaining the idea behind it. 

In addition, we should cover this code somehow.

quentin-jaquier-sonarsource (commented):
In addition, let's wait for @m-g-sonar opinion on the removal of `@Beta` annotations.

java-frontend/src/main/java/org/sonar/java/model/VisitorsBridge.java
margarita-nedzelska-sonarsource: Yes, the code is actually the same as in guava. It uses the standard algorithm to find a loop, which is having 2 pointers. one normal and other 2x times slower, increments on each second iteration. And if at some point 2 pointers are equal it means that we have a cycle here.

margarita-nedzelska-sonarsource:
This PR should be "rebased and merged"

quentin-jaquier-sonarsource (approved):
LGTM, well done!

PR #3327, opened 2020-12-07
Fix rule descriptions relative links to other rules

alban-auzeill:


quentin-jaquier-sonarsource (approved):
LGTM.

This happens when the rule key is not separated by whitespace in the rule description. I made sure to fix the descriptions of the impacted rules.

PR #3330, opened 2020-12-07
SONARJAVA-3567 Fix FPs on full matches

sebastian-hungerecker-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/helpers/RegexTreeHelper.java
```diff
@@ -253,4 +253,28 @@ public static boolean isEndBoundary(AutomatonState state) {
...
+    return onlyMatchesEmptySuffix(start, new HashSet<>());
+  }
+
+  private static boolean onlyMatchesEmptySuffix(AutomatonState start, Set<AutomatonState> visited) {
+    if (start instanceof FinalState || visited.contains(start)) {
+      return true;
+    }
+    visited.add(start);
+    if (start instanceof LookAroundTree) {
+      return onlyMatchesEmptySuffix(start.continuation());
```
alban-auzeill, line 266: Why not `return onlyMatchesEmptySuffix(start.continuation(), visited);`?

sebastian-hungerecker-sonarsource: That was a mistake, good catch.

PR #3342, opened 2020-12-09
SONARJAVA-2154 Remove usages of guava Immutable Collections from java-frontend module

margarita-nedzelska-sonarsource:
Only last commit should be reviewed. Merge after this [pr](https://github.com/SonarSource/sonar-java/pull/3337).

PR #3343, opened 2020-12-10
SONARJAVA-2154 Remove usages of guava Sets, Iterables and AbstractIterator from java-frontend module

margarita-nedzelska-sonarsource:
Depends on this [pr](https://github.com/SonarSource/sonar-java/pull/3342). Should be rebased

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/collections/SetUtils.java
+++ b/java-frontend/src/main/java/org/sonar/java/collections/SetUtils.java
@@ -53,4 +53,12 @@ public final class SetUtils {
       .filter(elem -> !set2.contains(elem))
       .collect(Collectors.toSet());
   }
+
+  public static <T> T getOnlyElement(Set<T> set) {
+    if (set.size() == 1) {
+      return set.iterator().next();
+    }
+    throw new IllegalArgumentException(String.format("Expected list of size 1, but was list of size %d.", set.size()));
+  }
+
 }
--- a/java-frontend/src/main/java/org/sonar/java/collections/SetUtils.java
+++ b/java-frontend/src/main/java/org/sonar/java/collections/SetUtils.java
@@ -49,16 +49,16 @@ public final class SetUtils {
   }
 
   public static <T> Set<T> difference(Set<T> set1, Set<T> set2) {
-    return set1.stream()
-      .filter(elem -> !set2.contains(elem))
-      .collect(Collectors.toSet());
+    Set<T> newSet1 = new HashSet<>(set1);
+    newSet1.removeAll(set2);
+    return newSet1;
   }
 
   public static <T> T getOnlyElement(Set<T> set) {
     if (set.size() == 1) {
       return set.iterator().next();
     }
-    throw new IllegalArgumentException(String.format("Expected list of size 1, but was list of size %d.", set.size()));
+    throw new IllegalArgumentException(String.format("Expected set of size 1, but was set of size %d.", set.size()));
   }
 
 }
--- a/java-frontend/src/main/java/org/sonar/java/collections/SetUtils.java
+++ b/java-frontend/src/main/java/org/sonar/java/collections/SetUtils.java
@@ -47,4 +47,10 @@ public final class SetUtils {
       .flatMap(Set::stream)
       .collect(Collectors.toSet());
   }
+
+  public static <T> Set<T> difference(Set<T> set1, Set<T> set2) {
+    return set1.stream()
+      .filter(elem -> !set2.contains(elem))
+      .collect(Collectors.toSet());
+  }
 }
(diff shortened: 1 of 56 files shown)


java-frontend/src/main/java/org/sonar/java/collections/SetUtils.java
```diff
@@ -47,4 +47,18 @@ private SetUtils() {
       .flatMap(Set::stream)
       .collect(Collectors.toSet());
   }
+
+  public static <T> Set<T> difference(Set<T> set1, Set<T> set2) {
+    return set1.stream()
```
m-g-sonar, line 52: I'm not sure about this implementation. According to the documentation of Guava, why not go for:
```
Set<T> newSet1 = new HashSet<>(set1);
newSet1.removeAll(set2);
return newSet1;
```

java-frontend/src/main/java/org/sonar/java/collections/SetUtils.java
```diff
@@ -47,4 +47,18 @@ private SetUtils() {
...
+      .filter(elem -> !set2.contains(elem))
+      .collect(Collectors.toSet());
+  }
+
+  public static <T> T getOnlyElement(Set<T> set) {
+    if (set.size() == 1) {
+      return set.iterator().next();
+    }
+    throw new IllegalArgumentException(String.format("Expected list of size 1, but was list of size %d.", set.size()));
```
m-g-sonar, line 61: this is a `set`, not a `list`

java-frontend/src/main/java/org/sonar/java/model/VisitorsBridgeForTests.java
```diff
@@ -58,7 +58,8 @@ public VisitorsBridgeForTests(Iterable<? extends JavaCheck> visitors, @Nullable
   }
 
   public VisitorsBridgeForTests(Iterable<? extends JavaCheck> visitors, List<File> projectClasspath, @Nullable SonarComponents sonarComponents) {
-    super(visitors, projectClasspath, sonarComponents, SymbolicExecutionMode.getMode(Iterables.toArray(visitors, JavaCheck.class)));
+    super(visitors, projectClasspath, sonarComponents, SymbolicExecutionMode.getMode(
```
m-g-sonar, line 61: I think it might be simpler to write a new method in SymbolicExecutionMode accepting an Iterable instead of an array.

java-frontend/src/main/java/org/sonar/java/model/statement/SwitchExpressionTreeImpl.java
```diff
@@ -101,7 +101,7 @@ public void accept(TreeVisitor visitor) {
 
   @Override
   public Iterable<Tree> children() {
-    return Iterables.concat(
+    return ListUtils.concat(
```
m-g-sonar, line 104: The more I read about all these changes, the more I think that having `children()`  returning `Iterable<Tree>` in the `JavaTree` is stupid... we should have go for a `List<Tree>` from day one, or a least a `Collection`. It does not make sense to use an Iterable anymore. If you are searching for something else to improve, go for it!

java-frontend/src/main/java/org/sonar/java/se/NullableAnnotationUtils.java
```diff
@@ -65,28 +67,31 @@ private NullableAnnotationUtils() {
...
+    "org.eclipse.jdt.annotation.Nullable",
+    "org.eclipse.jgit.annotations.Nullable",
+    "org.jetbrains.annotations.Nullable",
+    "org.jmlspecs.annotation.Nullable",
+    "org.netbeans.api.annotations.common.NullAllowed",
+    "org.netbeans.api.annotations.common.NullUnknown");
+
+  private static final Set<String> NULLABLE_ANNOTATIONS = Collections.unmodifiableSet(
+    Stream.of(STRONG_NULLABLE_ANNOTATIONS, WEAK_NULLABLE_ANNOTATIONS)
+    .flatMap(Set::stream).collect(Collectors.toSet()));
```
m-g-sonar, line 92: style, can you please align it as in others?

```
  Stream.of(STRONG_NULLABLE_ANNOTATIONS, WEAK_NULLABLE_ANNOTATIONS)
    .flatMap(Set::stream)
    .collect(Collectors.toSet()));
```

java-frontend/src/main/java/org/sonar/java/model/statement/SwitchExpressionTreeImpl.java
margarita-nedzelska-sonarsource: It looks like there are cases where we use custom implementation of Iterable. So maybe we can update this in other PR

margarita-nedzelska-sonarsource: Here it is https://github.com/SonarSource/sonar-java/pull/3345 :)

PR #3345, opened 2020-12-10
Refactor JavaTree children() method to return List instead of Iterable

margarita-nedzelska-sonarsource:
Only last commit is relevant


java-frontend/src/main/java/org/sonar/java/ast/parser/ListTreeImpl.java
```diff
@@ -67,45 +64,8 @@ public Kind kind() {
...
-      iterators = Arrays.asList(list.iterator(), separators.iterator());
-    }
-
-    @Override
-    public Iterator<Tree> iterator() {
-      return new InterleaveIterator<>(iterators);
-    }
-  }
-
-  private static class InterleaveIterator<E> extends AbstractIterator<E> {
```
m-g-sonar, line 88: so cool we can drop the AbstractIterator!

java-frontend/src/main/java/org/sonar/java/model/JavaTree.java
```diff
@@ -129,7 +129,7 @@ public void setParent(Tree parent) {
    *
    * @throws java.lang.UnsupportedOperationException if {@link #isLeaf()} returns {@code true}
    */
-  protected abstract Iterable<Tree> children();
+  protected abstract List<Tree> children();
 
   public List<Tree> getChildren() {
     if(children == null) {
```
m-g-sonar, line 135: So we cab probably change next line to: 
```java
children = children().stream()
  .filter(Objects::nonNull)

File: java-frontend/src/main/java/org/sonar/java/model/JavaTree.java [hunk 3]
```diff
@@ -129,7 +129,7 @@ public void setParent(Tree parent) {
    *
    * @throws java.lang.UnsupportedOperationException if {@link #isLeaf()} returns {@code true}
    */
-  protected abstract Iterable<Tree> children();
+  protected abstract List<Tree> children();
```
m-g-sonar, line 132: There are some unit tests that add the result of this method into another list to do assertions. It worth modifying them to simplify a bit.

m-g-sonar (changes requested):

java-frontend/src/main/java/org/sonar/java/ast/parser/ListTreeImpl.java
margarita-nedzelska-sonarsource: I guess this class (fully covered btw :) ) lived for a couple of hours

m-g-sonar (approved):

PR #3347, opened 2020-12-11
SONARJAVA-3626 Fix FN Rule S3824: Apply The same SymbolicValue for static constants or enums when used as MemberSelect

margarita-nedzelska-sonarsource:


quentin-jaquier-sonarsource (approved):
The idea in this ticket and the actual fix implemented sounds good to me.

Concerning the new issues found for S2583 and S3824, they look like nice TP.
For S4165, it's not that clear, I did not do the effort to understand exactly the code, but at first glance, the same could happen with any variable instead of the enum, and the different enum values refer correctly to the same one. So if there is a problem, it is in the engine/rule, and not related to this fix.

All in all, LGTM!

margarita-nedzelska-sonarsource:
Yes, I should have mentioned. The issues found in ruling in S4165 are caused by the problem in the engine itself and were revealed because enum are fixed here. The thing is the only way to reach the second switch is that <code>_state</code> actual value is <code>PRODUCE</code> So from this point of view the check is right cause there is no point in reassigning the same value. The tricky part is that this case is actually unreachable code and will never be executed. So the problem is actually in the way we handle **switch** statements. 

So we decided with @m-g-sonar to accept this ruling issue for now and then to fix switch handling in other ticket.

PR #3348, opened 2020-12-11
SONARJAVA-3578 FP in S2147 when the type of the Exception is needed inside the body

quentin-jaquier-sonarsource:


java-frontend/src/main/java/org/sonar/java/model/SyntacticEquivalence.java
```diff
@@ -62,27 +65,27 @@ public static boolean areEquivalent(List<? extends Tree> leftList, List<? extend
...
    */
   public static boolean areEquivalent(@Nullable Tree leftNode, @Nullable Tree rightNode) {
-    return areEquivalent(leftNode, rightNode, (t1, t2) -> false);
+    return areEquivalent(leftNode, rightNode, (t1, t2) -> false, true);
   }
 
   /**
    * @return true, if nodes are syntactically equivalent
-   * Use permissiveEquivalence to force the equivalence of two nodes
+   * Use "overwriteEquivalence" to force the equivalence or not of two nodes. When it returns true, the method will return "equivalenceValue".
```
margarita-nedzelska-sonarsource, line 73: What about "overwritten/overridden"
and
"to force the equivalence check result"

java-checks/src/main/java/org/sonar/java/checks/CombineCatchCheck.java
```diff
@@ -55,6 +58,33 @@ public void visitNode(Tree tree) {
     }
   }
 
+  private static boolean areNotSameMethodCalls(JavaTree leftNode, JavaTree rightNode) {
```
margarita-nedzelska-sonarsource, line 61: Could this be a part of general equivalence check? To my mind, we can't say that 2 blocks are equivalent, if they're look the same but referencing different methods (in general, not only for this rule)

This change might also influence other rules, so need to check.

quentin-jaquier-sonarsource:
As discussed, I refactored the code.

We still have the two strange arguments `BiPredicate<JavaTree, JavaTree> overwriteEquivalence` and `boolean equivalenceValue`, but hidden as implementation details. The `public` methods are cleaner this way, and the old interface is unchanged. Let me know what you think!

Concerning applying this to other rules, it can be something to investigate, but it does not enter well in the scope of this ticket, I suggest not doing it in this PR.

margarita-nedzelska-sonarsource (approved):
LGTM!

PR #3349, opened 2020-12-14
SONARJAVA-3637 add new property 'sonar.java.jdkHome'

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/classpath/ClasspathProperties.java
```diff
@@ -37,25 +38,30 @@ private ClasspathProperties() {
...
+    extensions.add(PropertyDefinition.builder(SONAR_JAVA_JDK_HOME)
+      .description("Path to jdk directory used by the project under analysis.")
+      .hidden()
+      .build()
+    );
     extensions.add(PropertyDefinition.builder(SONAR_JAVA_BINARIES)
-            .description("Comma-separated paths to directories containing the binary files (directories with class files).")
-            .hidden()
-            .build()
+      .description("Comma-separated paths to directories containing the binary files (directories with class files).")
```
henryju, line 47: Unrelated, but would be great to declare those properties as multi-valued

java-frontend/src/main/java/org/sonar/java/classpath/AbstractClasspath.java
```diff
@@ -62,18 +63,38 @@
...
+    return settings.get(ClasspathProperties.SONAR_JAVA_JDK_HOME)
+      .flatMap(AbstractClasspath::fileOrLog)
+      .map(File::toPath)
+      .map(JavaSdkUtil::getJdkClassesRoots)
+      .orElse(Collections.emptyList());
+  }
+
+  private static Optional<File> fileOrLog(String path) {
+    File file = new File(path);
+    if (!file.exists()) {
```
henryju, line 90: Maybe also check if this is a directory

java-frontend/src/main/java/org/sonar/java/classpath/AbstractClasspath.java
```diff
@@ -62,18 +63,38 @@
...
 
+  protected List<File> getJdkJars() {
+    return settings.get(ClasspathProperties.SONAR_JAVA_JDK_HOME)
+      .flatMap(AbstractClasspath::fileOrLog)
+      .map(File::toPath)
+      .map(JavaSdkUtil::getJdkClassesRoots)
+      .orElse(Collections.emptyList());
+  }
+
+  private static Optional<File> fileOrLog(String path) {
```
henryju, line 88:
```suggestion
  private static Optional<File> existingDirectoryOrLog(String path) {
```

henryju (approved):
Minor feedback but LGTM

java-frontend/src/main/java/org/sonar/java/classpath/ClasspathProperties.java
m-g-sonar: since they are hidden anyway, i'm not sure it would change things, but indeed.

PR #3350, opened 2020-12-14
SONARJAVA-3620 FP in S2384 when unmodifiable collection is returned

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/MutableMembersUsageCheck.java
```diff
@@ -149,18 +149,21 @@ private static boolean isThis(ExpressionTree expression) {
     return expression.is(Tree.Kind.IDENTIFIER) && ((IdentifierTree) expression).name().equals("this");
   }
 
-  private static boolean isImmutableFinalVariable(Symbol.VariableSymbol symbol) {
-    if (symbol.isFinal()) {
-      VariableTree declaration = symbol.declaration();
-      // symbol is private, so declaration can only be null if assignment is done in static block
```
quentin-jaquier-sonarsource, line 155: > // symbol is private, so declaration can only be null if assignment is done in static block

I did not manage to reproduce a case where the declaration is null, but I still added a null check because I was not convinced by this comment. 
I think it makes sense to keep it for defensive programming.

margarita-nedzelska-sonarsource:
Overall looks good. I'm just wondering, do we want to check for other common immutable collections (Apache Commons, Guava and Results of Java 9  methods like listOf, setOf, etc...)

quentin-jaquier-sonarsource:
We had [SONARJAVA-3429[](https://jira.sonarsource.com/browse/SONARJAVA-3429) doing something similar.

It seems that we are only missing Apache Commons, if you think it's worth to add it, we should create a ticket.

java-checks/src/main/java/org/sonar/java/checks/MutableMembersUsageCheck.java
margarita-nedzelska-sonarsource: Maybe Objects.requireNonNull will be a better option here, cause I can't imagine a situation when declaration is null for private fields. But null check is also fine for me

margarita-nedzelska-sonarsource (approved):
Overall looks good!

java-checks/src/main/java/org/sonar/java/checks/MutableMembersUsageCheck.java
quentin-jaquier-sonarsource: Both alternatives sound good to me aswell. The behavior when the declaration is null (for whatever reason) is to proceed with the check of the usages as if it was not initialized, so it seems fine to me to keep it as it is.

PR #3351, opened 2020-12-14
SONARJAVA-3570 S5411: Relax rule against boxed booleans whose first usage is a null check

dorian-burihabwa-sonarsource:


java-checks-test-sources/src/main/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -341,4 +341,72 @@ Boolean False() {
...
+
+  void forStatementCheckedBeforeUseIsCompliant() {
+    Boolean checkedBeforeUse = getCustomNonNull();
+    if (checkedBeforeUse != null) {
+      for (; checkedBeforeUse; )  {} // Compliant because the variable is checked before use
+    }
+
+    Boolean checkedBeforeUse2 = getCustomNonNull();
+    if (checkedBeforeUse2 == null) {
+      for (; checkedBeforeUse2; )  {} // Compliant because the variable is checked before use
```
margarita-nedzelska-sonarsource, line 353: I think this code should not be compliant, cause it's producing NPE

java-checks-test-sources/src/main/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -341,4 +341,72 @@ Boolean False() {
...
+  }
+
+  void whileStatementCheckedBeforeUseIsCompliant() {
+    Boolean checkedBeforeUse = getCustomNonNull();
+    if (checkedBeforeUse != null) {
+      while (checkedBeforeUse)  {} // Compliant because the variable is checked before use
+    }
+    Boolean checkedBeforeUse2 = getCustomNonNull();
+    if (checkedBeforeUse2 == null) {
+      while (checkedBeforeUse2)  {} // Compliant because the variable is checked before use
```
margarita-nedzelska-sonarsource, line 364: I think this code should not be compliant, cause it's producing NPE

java-checks-test-sources/src/main/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -341,4 +341,72 @@ Boolean False() {
...
+    }
+
+    Boolean checkedBeforeUse2 = getCustomNonNull();
+    if (checkedBeforeUse2 == null) {
+      for (; checkedBeforeUse2; )  {} // Compliant because the variable is checked before use
+    }
+  }
+
+  void whileStatementCheckedBeforeUseIsCompliant() {
+    Boolean checkedBeforeUse = getCustomNonNull();
```
margarita-nedzelska-sonarsource, line 358: I think you don't need to use <code>getCustomNonNull()</code>, but sth like <code>customNullable</code> because the main reason to use @NonNull annotations is to get rid of null checks. And I think the cases with annotation are already covered.

java-checks-test-sources/src/main/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -341,4 +341,72 @@ Boolean False() {
 
   Object foo() { return new Object(); }
   Object bar() { return new Object(); }
+
```
margarita-nedzelska-sonarsource, line 344: I will also suggest adding more test cases:

- where condition is in else branch

-  where there are other usages of checked variable

- when the tested variable is actually null

-  etc.

margarita-nedzelska-sonarsource (changes requested):
There are a few comments. Feel free to contact me if you have any questions.

java-checks-test-sources/src/main/java/checks/BoxedBooleanExpressionsCheck.java
dorian-burihabwa-sonarsource: The problem with this specific case and the one that follows, at lines 363-364, stems from the fact that the example is not very realistic:  this code is written to crash.
However, if the test against null was followed by an early return, the code would be valid.

Without changing the logic of the check, I can rewrite the example to be more realistic.
Otherwise, we would have to change the logic of the check and look into symbolic execution to distinguish a good test from a bad one.
WDYT?

margarita-nedzelska-sonarsource: Yep, I see you wanted to cover early returns, but this is actually a SE job, and as we discussed we can return to it later. The thing I'm afraid of is when we're relaxing check for early returns we might introduce some FNs, which is tricky.

So, the simple solution could be:
just add additional test cases with TP, and these ones mark as known FN, which we're going to remove in the future.

Or:
One more possible solution is to check early returns, that the last statement inside null check is actually return. So we will cover early returns and won't introduce these new FN

dorian-burihabwa-sonarsource: I will go with you first suggestion then and add tests with TP and mark those with FN.

margarita-nedzelska-sonarsource (approved):
Loooks good!

PR #3352, opened 2020-12-14
SONARJAVA-3628 FP in S5853 when assertions "flatExtracting" prevent the chaining

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/AssertJConsecutiveAssertionCheck.java
```diff
@@ -57,7 +57,7 @@
 
   public static final MethodMatchers ASSERTJ_SET_CONTEXT_METHODS = MethodMatchers.create()
     .ofSubTypes("org.assertj.core.api.AbstractAssert")
-    .name(name -> name.startsWith("extracting") || name.startsWith("using") || name.startsWith("filtered"))
+    .name(name -> name.startsWith("extracting") || name.startsWith("using") || name.startsWith("filtered") || "flatExtracting".equals(name))
```
dorian-burihabwa-sonarsource, line 60: Unless there is a reason to believe that `name` might be null, which should be tested before entering the predicate, it might be better to change the line to `name.equals("flatExtracting")` for stylistic consistency.

dorian-burihabwa-sonarsource (changes requested):
Apart from that sylistic issue, LGTM.

dorian-burihabwa-sonarsource (approved):
Looks good to me :+1:

PR #3353, opened 2020-12-15
SONARJAVA-3467 FP on S1948 when using both field and setter/constructor injection

margarita-nedzelska-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
@@ -21,6 +21,9 @@ package org.sonar.java.checks.serialization;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.sonar.check.Rule;
 import org.sonar.java.model.JUtils;
@@ -33,6 +36,7 @@ import org.sonar.plugins.java.api.tree.AssignmentExpressionTree;
 import org.sonar.plugins.java.api.tree.ClassTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.Modifier;
 import org.sonar.plugins.java.api.tree.ParameterizedTypeTree;
 import org.sonar.plugins.java.api.tree.Tree;
@@ -42,6 +46,8 @@ import org.sonar.plugins.java.api.tree.WildcardTree;
 
 @Rule(key = "S1948")
 public class SerializableFieldInSerializableClassCheck extends IssuableSubscriptionVisitor {
+
+  private static final String JAVAX_INJECT = "javax.inject.Inject";
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
@@ -54,12 +60,44 @@ public class SerializableFieldInSerializableClassCheck extends IssuableSubscript
     if (isSerializable(classTree)
       && !SerializableContract.hasSpecialHandlingSerializationMethods(classTree)
       && !classTree.symbol().type().isSubtypeOf("javax.servlet.http.HttpServlet")) {
-      for (Tree member : classTree.members()) {
-        if (member.is(Tree.Kind.VARIABLE)) {
-          checkVariableMember((VariableTree) member);
-        }
-      }
+      
+      Set<String> constructorInjectedParams = getConstructorInjectedFields(classTree);
+      Set<String> setterInjectedParams = getSetterInjectedFields(classTree);
+      
+      classTree.members().stream()
+        .filter(member -> member.is(Tree.Kind.VARIABLE))
+        .map(VariableTree.class::cast)
+        .filter(variableTree -> 
+          !(constructorInjectedParams.contains(variableTree.simpleName().name()) ||
+            setterInjectedParams.contains(variableTree.simpleName().name())))
+        .forEach(this::checkVariableMember);
     }
+  }
+
+  private static Set<String> getSetterInjectedFields(ClassTree classTree) {
+    return classTree.members().stream()
+      .filter(member -> member.is(Tree.Kind.METHOD))
+      .map(MethodTree.class::cast)
+      .filter(SerializableFieldInSerializableClassCheck::isAnnotatedWithInject)
+      .map(methodTree -> methodTree.simpleName().name())
+      .filter(name -> name.startsWith("set"))
+      .map(name -> name.substring(3).toLowerCase(Locale.ROOT))
+      .collect(Collectors.toSet());
+  }
+
+  private static Set<String> getConstructorInjectedFields(ClassTree classTree) {
+    return classTree.members().stream()
+      .filter(member -> member.is(Tree.Kind.CONSTRUCTOR))
+      .map(MethodTree.class::cast)
+      .filter(SerializableFieldInSerializableClassCheck::isAnnotatedWithInject)
+      .flatMap(methodTree -> methodTree.parameters().stream())
+      .map(variableTree -> variableTree.simpleName().name())
+      .collect(Collectors.toSet());
+  }
+
+  private static boolean isAnnotatedWithInject(MethodTree methodTree) {
+    return methodTree.symbol().metadata().isAnnotatedWith(JAVAX_INJECT)
+      || hasAnnotationsWithIncompleteSemantic(methodTree.symbol().metadata());
   }
 
   private void checkVariableMember(VariableTree variableTree) {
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -54,14 +60,46 @@ public void visitNode(Tree tree) {
...
+      .map(MethodTree.class::cast)
+      .filter(SerializableFieldInSerializableClassCheck::isAnnotatedWithInject)
+      .flatMap(methodTree -> methodTree.parameters().stream())
+      .map(variableTree -> variableTree.simpleName().name())
+      .collect(Collectors.toSet());
+  }
+
+  private static boolean isAnnotatedWithInject(MethodTree methodTree) {
+    return methodTree.symbol().metadata().isAnnotatedWith(JAVAX_INJECT)
+      || methodTree.symbol().metadata().annotations().stream().anyMatch(annotation -> annotation.symbol().isUnknown());
```
margarita-nedzelska-sonarsource, line 100: Without semantic this annotation won't be found. So this line protects of FPs in case of incomplete semantic., the same check is done in line 179

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -54,14 +60,46 @@ public void visitNode(Tree tree) {
...
-        }
-      }
+      
+      Set<String> constructorInjectedParams = getConstructorInjectedFields(classTree);
+      Set<String> setterInjectedParams = getSetterInjectedFields(classTree);
+      
+      classTree.members().stream()
+        .filter(member -> member.is(Tree.Kind.VARIABLE))
+        .map(VariableTree.class::cast)
+        .filter(variableTree ->
```
quentin-jaquier-sonarsource, line 70: We are doing a first filtering here, and then a second in `checkVariableMember` thanks to `isExcluded`. We could move the second filtering here.
It will avoid having the differents kind of injections being filtered at two different places.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -135,10 +173,14 @@ private static boolean isTransientSerializableOrInjected(VariableTree member) {
       return true;
     }
     SymbolMetadata metadata = member.symbol().metadata();
-    return metadata.isAnnotatedWith("javax.inject.Inject")
+    return metadata.isAnnotatedWith(JAVAX_INJECT)
       || metadata.isAnnotatedWith("javax.ejb.EJB")
       || metadata.isAnnotatedWith("org.apache.wicket.spring.injection.annot.SpringBean")
-      || metadata.annotations().stream().anyMatch(annotation -> annotation.symbol().isUnknown());
+      || hasAnnotationsWithIncompleteSemantic(metadata);
```
quentin-jaquier-sonarsource, line 179: `isAnnotatedWith` will go through all annotations, so with this code, you end up going through the list 4 times when the field is not annotated by any of the annotations (which I expect to be the most frequent case). We could do everything in one shot, without using `isAnnotatedWith`. 

The same happens in `isAnnotatedWithInject`.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -54,14 +60,46 @@ public void visitNode(Tree tree) {
...
+      .collect(Collectors.toSet());
+  }
+
+  private static Set<String> getConstructorInjectedFields(ClassTree classTree) {
+    return classTree.members().stream()
+      .filter(member -> member.is(Tree.Kind.CONSTRUCTOR))
+      .map(MethodTree.class::cast)
+      .filter(SerializableFieldInSerializableClassCheck::isAnnotatedWithInject)
+      .flatMap(methodTree -> methodTree.parameters().stream())
+      .map(variableTree -> variableTree.simpleName().name())
```
quentin-jaquier-sonarsource, line 94: I'm wondering if we can really rely on the naming of parameters here. As an example, we will not support the example described in the ticket.

What is your opinion on it?

margarita-nedzelska-sonarsource: Yes, that's true. I was just thinking what would be a better option.
So I changed the logic here to check assignments inside the constructor, and exclude fields that are assigned values inside constructor annotated with @Inject.
 Wdyt about this solution?

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -54,32 +65,65 @@ public void visitNode(Tree tree) {
...
-          checkVariableMember((VariableTree) member);
-        }
-      }
+
+      Set<String> constructorInjectedParams = getConstructorInjectedFields(classTree);
+      Set<String> setterInjectedParams = getSetterInjectedFields(classTree);
+
+      classTree.members().stream()
+        .filter(member -> member.is(Tree.Kind.VARIABLE))
+        .map(VariableTree.class::cast)
```
quentin-jaquier-sonarsource, line 74: You compute this two times (the first one in `getConstructorInjectedFields`), what about storing it in an intermediate variable instead?

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
quentin-jaquier-sonarsource: Nice one, this solution looks better to me! What about using it for the setter as well? Doing both in this method?

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -54,32 +65,65 @@ public void visitNode(Tree tree) {
...
+
+    AssignmentsVisitor assignmentsVisitor = new AssignmentsVisitor(fields);
+    
+    classTree.members().stream()
+      .filter(member -> member.is(Tree.Kind.CONSTRUCTOR))
+      .map(MethodTree.class::cast)
+      .filter(methodTree -> isAnnotatedWith(methodTree.symbol().metadata(), JAVAX_INJECT))
+      .forEach(methodTree -> methodTree.accept(assignmentsVisitor));
+    
+    return assignmentsVisitor.getUsedFields();
```
quentin-jaquier-sonarsource, line 108: This should work fine, I would just add a test when you have multiple constructors reaching the `forEach`, to make sure we are not losing anything across different visits.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -168,4 +225,30 @@ private static boolean implementsSerializable(@Nullable Type type) {
...
+    }
+
+    @Override
+    public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+      if (ExpressionUtils.isSelectOnThisOrSuper(tree)) {
+        MemberSelectExpressionTree select = ((MemberSelectExpressionTree) tree.variable());
+        usedFields.add(select.identifier().name());
+      } else if (tree.variable().is(Tree.Kind.IDENTIFIER)) {
+        IdentifierTree identifier = ((IdentifierTree) tree.variable());
+        if (fields.contains(identifier.name())) {
```
quentin-jaquier-sonarsource, line 243: Do we really need this check? If we collect any variable assigned, it should work fine, the set returned will just contain unrelated names.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
margarita-nedzelska-sonarsource: If we remove the check in AssignmentVisitor, there is no need to have all fields before iterating

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -135,10 +179,23 @@ private static boolean isTransientSerializableOrInjected(VariableTree member) {
...
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private static boolean isOneOf(Type type, String[] fullyQualifiedNames) {
+    return Arrays.stream(fullyQualifiedNames)
+      .map(type::is)
+      .reduce(false, Boolean::logicalOr);
```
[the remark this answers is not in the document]
margarita-nedzelska-sonarsource: Looks nice!

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
margarita-nedzelska-sonarsource: Actually this is impossible situation for compiling example, cause only one constructor can be annotated with <code>@Inject</code> But I added it in non-compiling example

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
margarita-nedzelska-sonarsource: It was a small space opitimization. As far as we return Set, should not be a big deal regarding performance

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
margarita-nedzelska-sonarsource: I thought about it. "Setters" are actually a very common practice and initially I thought it will be much faster to just rely on naming convention, but you're right <code>@Inject</code> can be used on any method and it will be nice to reduce noise for them as well

quentin-jaquier-sonarsource (approved):
LGTM

PR #3354, opened 2020-12-16
Fix CheckVerifier when testing multiples files

quentin-jaquier-sonarsource:
Only the issues reported by the execution of the rules on the last file were taken into consideration, while we should consider any issue raised during the analysis.


quentin-jaquier-sonarsource:
Postponing it since it requires clarification.
See [SONARJAVA-3643](https://jira.sonarsource.com/browse/SONARJAVA-3643).

PR #3357, opened 2020-12-17
SONARJAVA-3555 S4830: Report issues on empty implementations of X509ExtendedTrustManager methods

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/ServerCertificatesCheck.java
```diff
@@ -38,12 +38,22 @@
...
+    MethodMatchers.create()
+      .ofSubTypes("javax.net.ssl.X509ExtendedTrustManager")
+      .names("checkClientTrusted", "checkServerTrusted")
+      .addParametersMatcher(X509_CERTIFICATE_ARRAY, JAVA_LANG_STRING, "java.net.Socket")
+      .addParametersMatcher(X509_CERTIFICATE_ARRAY, JAVA_LANG_STRING, "javax.net.ssl.SSLEngine")
+      .build()
+  );
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
```
margarita-nedzelska-sonarsource, line 59: I suppose you haven't face it yet. But each time there is a call to ```hasSemantic()``` we're removing this condition. It was useful before frontend was migrated to Eclipse now it's not actually emulating real behaviour, because Eclipse try to resolve as much as possible and put ```unknown``` in case uf incomplete semantic.
So now we're removing it and adding tests ```.withoutSemantic()``` (If it fails you need to review issues and understand why this is happening)

margarita-nedzelska-sonarsource (commented):
Overall looks good!
One thing should be done before merge

margarita-nedzelska-sonarsource (approved):
LGTM

PR #3361, opened 2020-12-17
SONARJAVA-3637 document property

m-g-sonar:


sonar-java-plugin/src/main/resources/static/documentation.md
```diff
@@ -35,6 +35,13 @@ Key | Value
 [[warning]]
 | Android users, Jack doesn't provide the required `.class` files.
 
+## Project's specific JDK
+
+In some situations, you might have to analyze a project built with a different version of Java than the one executing the analysis. The most common case is to run the analysis with **Java 11**, while the project itself uses **Java 8** or before for its build. This case is normally automatically handled when using Maven or Gradle, as well as with any flavor of SonarLint.
+
+If it is your case, and you are **NOT** using Maven or Gradle, you will need to set the property `sonar.java.jdkHome` manually (see below) to specify which JDK classes the analyzer must refer to during the analysis. Not setting this property, while it would have been required, usually leads to inconsistent or even impossible to fix issues being reported, especially in relation with native JDK classes.
```
margarita-nedzelska-sonarsource, line 42: The first sentence is quite long and complicated. Maybe it will be better to split it in two.

```suggestion
If it is your case, and you are **NOT** using Maven or Gradle, you will need to set the property `sonar.java.jdkHome` manually (see below). By doing this you'll specify which JDK classes the analyzer must refer to during the analysis. Not setting this property, while it would have been required, usually leads to inconsistent or even impossible to fix issues being reported, especially in relation with native JDK classes.
```

margarita-nedzelska-sonarsource (approved):
Overall looks good and very clear to me 👍 
Just one small suggestion

PR #3364, opened 2020-12-22
SONARJAVA-3633 FP in S4032 when there are several source directories

christophe-zurn-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/UselessPackageInfoCheck.java
```diff
@@ -20,23 +20,57 @@
...
+    if (hasOtherFiles) {
+      knownPackagesWithOtherFiles.add(packageName);
+    } else if (packageInfoFile.isFile()) {
+      unneededPackageInfoFiles.put(packageName, context);
+    }
+  }
+
+  @Override
+  public void endOfAnalysis() {
+    unneededPackageInfoFiles.keySet().removeAll(knownPackagesWithOtherFiles);
```
dorian-burihabwa-sonarsource, line 63: The behavior of `scanFile` seems to be that it adds a package to one or the other but not both. Is the explicit removal of valid packages from the the set of unneeded packages necessary?

java-checks/src/test/java/org/sonar/java/checks/UselessPackageInfoCheckTest.java
```diff
@@ -45,7 +45,17 @@ void withOtherFile() {
   @Test
   void notAPackageInfo() {
     JavaCheckVerifier.newVerifier()
-      .onFile(testSourcesPath("checks/UselessPackageInfoCheck/packageWithNoOtherFilesButNotPackageInfo/HelloWorld.java"))
+      .onFiles(
```
dorian-burihabwa-sonarsource, line 48: To avoid  the silent corruption of one of the test files,  we should probably repeat the test with a call to singular `onFile`.

dorian-burihabwa-sonarsource (approved):

java-checks/src/main/java/org/sonar/java/checks/UselessPackageInfoCheck.java
christophe-zurn-sonarsource: Yes, it is needed to avoid FP in case the analysis scan the same package multiple times (from different source folder). In such a case, the package may be present in both collections, but we do not want to raise an issue.

PR #3365, opened 2020-12-24
SONARJAVA-3480 S5917: DateTimeFormatters should not use mismatched year and week numbers

dorian-burihabwa-sonarsource:


java-checks-test-sources/src/main/java/checks/DateTimeFormatterMismatch.java
```diff
@@ -0,0 +1,80 @@
...
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.temporal.ChronoField;
+import java.time.temporal.WeekFields;
+import java.util.Locale;
+
+public class DateTimeFormatterMismatch {
+  public void createUsingPatterns() {
+    DateTimeFormatter.ofPattern("Y-ww"); // Compliant
+    DateTimeFormatter.ofPattern("YY-ww"); // Compliant
```
christophe-zurn-sonarsource, line 12: For completeness, we could add 2 tests for `ofPattern`: one where the argument is not a string literal, and another where the pattern does not contain any of the character we're looking for

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,145 @@
...
+  private static final MethodMatchers DATE_TIME_FORMATTER_BUILDER = MethodMatchers.create()
+    .ofTypes("java.time.format.DateTimeFormatterBuilder")
+    .names("toFormatter")
+    .addParametersMatcher()
+    .build();
+
+  private static final Pattern WEEK_PATTERN = Pattern.compile(".*ww{1,2}.*");
+  private static final Pattern YEAR_OF_ERA_PATTERN = Pattern.compile(".*[uy]+.*");
+
+  private static final String CHANGE_WEEK_FORMAT_MESSAGE = "Change this week format to use the week of week-based year instead.";
```
christophe-zurn-sonarsource, line 60: RSPEC message in that case is `Change this year format to use ChronoField.YEAR instead`. Is there a reason why the message here is different ? I think the message from RSPEC makes more sense following the primary and secondary locations that we want to highlight

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,145 @@
...
+      }
+      wanderer = mit.methodSelect();
+    }
+    if (wanderer == null) {
+      return;
+    }
+    ExpressionTree lastExpression = ((MemberSelectExpressionTree) wanderer).expression();
+    if (!lastExpression.is(Tree.Kind.NEW_CLASS)) {
+      return;
+    }
```
christophe-zurn-sonarsource, line 116: Why do we want to check the last expression type here ? We could also try to catch the following cases:
```
DateTimeFormatterBuilder builder = new DateTimeFormatterBuilder();
return builder.appendValue(WeekFields.ISO.weekBasedYear(), 4) // Noncompliant
      .appendLiteral('-')
      .appendValue(ChronoField.ALIGNED_WEEK_OF_YEAR, 2)
      .toFormatter();
```

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,145 @@
...
+    if (usesWeekBasedYear && !usesWeekOfWeekBasedYear) {
+      reportIssue(invocation, CHANGE_WEEK_FORMAT_MESSAGE);
+    } else if (!usesWeekBasedYear && usesWeekOfWeekBasedYear) {
+      reportIssue(invocation, CHANGE_YEAR_FORMAT_MESSAGE);
```
christophe-zurn-sonarsource, line 120: From the RSPEC, I believe the tree we want to report the issue on is not the `toFormatter` method invocation, but the `appendValue` on which we have the year, with secondary location being the `appendValue` invocation on which the weeks are appended.

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Walking back through an invocation chain from a call to DateTimeFormatterBuilder.toFormatter looking for calls to appendValue.
+   * If conflicting week and year settings are detected, an issue is reported
+   * @param invocation A call to DateTimeFormatterBuilder.toFormatter
+   */
+  private void visitBuildChain(MethodInvocationTree invocation) {
+    boolean usesWeekBasedYear = false;
+    boolean usesWeekOfWeekBasedYear = false;
+    Tree wanderer = invocation.methodSelect();
+    while (wanderer != null && wanderer.is(Tree.Kind.MEMBER_SELECT)) {
```
christophe-zurn-sonarsource, line 98: I think we can remove the `wanderer != null` check here: `methodSelect()` is not annotated as nullable, so it is assumed never to be null

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,145 @@
...
+    if (wanderer == null) {
+      return;
+    }
+    ExpressionTree lastExpression = ((MemberSelectExpressionTree) wanderer).expression();
+    if (!lastExpression.is(Tree.Kind.NEW_CLASS)) {
+      return;
+    }
+    if (usesWeekBasedYear && !usesWeekOfWeekBasedYear) {
+      reportIssue(invocation, CHANGE_WEEK_FORMAT_MESSAGE);
+    } else if (!usesWeekBasedYear && usesWeekOfWeekBasedYear) {
```
christophe-zurn-sonarsource, line 119: With these checks, we will currenlty raise FP in the following 2 cases:
```
    new DateTimeFormatterBuilder() // Noncompliant - FP
      .appendValue(WeekFields.ISO.weekBasedYear(), 4)
      .toFormatter();

    new DateTimeFormatterBuilder() // Noncompliant - FP
      .appendValue(WeekFields.ISO.weekOfWeekBasedYear(), 2)
      .toFormatter();
```
We should only take into account cases where the year and week are explicitely set to mismatching format

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,145 @@
...
+    ExpressionTree argument = arguments.get(0);
+    if (argument.is(Tree.Kind.METHOD_INVOCATION)) {
+      MethodInvocationTree call = (MethodInvocationTree) argument;
+      Symbol symbol = call.symbol();
+      return symbol.name().equals("weekBasedYear");
+    }
+    return false;
+  }
+
+  private static boolean isWeekOfWeekBasedYearUsed(MethodInvocationTree invocation) {
```
christophe-zurn-sonarsource, line 135: We could use `MethodMatchers` for checking these invocations

christophe-zurn-sonarsource (changes requested):

java-checks-test-sources/src/main/java/checks/DateTimeFormatterMismatch.java
```diff
@@ -0,0 +1,103 @@
...
+      .appendLiteral('-')
+      .appendValue(ChronoField.ALIGNED_WEEK_OF_YEAR, 2)
+      .toFormatter();
+
+    DateTimeFormatterBuilder builder = new DateTimeFormatterBuilder()
+      .appendValue(ChronoField.YEAR, 4) // Noncompliant [[sc=20;ec=36]]
+      .appendLiteral('-')
+      .appendValue(WeekFields.ISO.weekOfWeekBasedYear(), 2);
+
+    builder.toFormatter();
```
christophe-zurn-sonarsource, line 93: For completeness, we could add an additional test case to document that we do not consider consecutive calls, which is ok:
```
DateTimeFormatterBuilder builder2 = new DateTimeFormatterBuilder();
builder2.appendValue(ChronoField.YEAR, 4); // FN - consecutive calls on builders are not taken into account
builder2.appendLiteral('-');
builder2.appendValue(WeekFields.ISO.weekOfWeekBasedYear(), 2);
builder2.toFormatter();
```

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,216 @@
...
+  private static final MethodMatchers APPEND_VALUE_MATCHER = MethodMatchers.create()
+    .ofTypes("java.time.format.DateTimeFormatterBuilder")
+    .names("appendValue")
+    .addParametersMatcher("java.time.temporal.TemporalField", "int")
+    .build();
+
+  private static final MethodMatchers WEEK_BASED_YEAR_MATCHER = MethodMatchers.create()
+    .ofTypes("java.time.temporal.WeekFields")
+    .names("weekBasedYear")
+    .addParametersMatcher()
```
christophe-zurn-sonarsource, line 59: For clarity, I suggest to replace this with `addWithoutParametersMatcher`:
```suggestion
    .addWithoutParametersMatcher()
```

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,216 @@
...
+    .build();
+
+  private static final MethodMatchers WEEK_OF_WEEK_BASED_YEAR_MATCHER = MethodMatchers.create()
+    .ofTypes("java.time.temporal.WeekFields")
+    .names("weekOfWeekBasedYear")
+    .addParametersMatcher()
```
christophe-zurn-sonarsource, line 65: For clarity, I suggest to replace this with `addWithoutParametersMatcher`:
```suggestion
    .addWithoutParametersMatcher()
```

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,216 @@
...
+  public static boolean isChronoFieldYear(ExpressionTree argument) {
+    if (argument.is(Tree.Kind.MEMBER_SELECT)) {
+      MemberSelectExpressionTree select = (MemberSelectExpressionTree) argument;
+      IdentifierTree identifier = select.identifier();
+      return identifier.name().equals("YEAR");
+    }
+    return false;
+  }
+
+  static class ChainVisitor extends BaseTreeVisitor {
```
christophe-zurn-sonarsource, line 178: We can make this inner class `private`

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,216 @@
...
+
+  private static boolean isWeekBasedYearUsed(ExpressionTree argument) {
+    if (argument.is(Tree.Kind.METHOD_INVOCATION)) {
+      MethodInvocationTree call = (MethodInvocationTree) argument;
+      return WEEK_BASED_YEAR_MATCHER.matches(call);
+    }
+    return false;
+  }
+
+  public static boolean isChronoFieldYear(ExpressionTree argument) {
```
christophe-zurn-sonarsource, line 169: You can make all these `public static` methods private, they are not needed to be public, and they can all be moved inside the inner `ChainVisitor` class

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,216 @@
...
+      inspectCall(callToAppendValue);
+      ExpressionTree wanderer = callToAppendValue.methodSelect();
+      while (wanderer.is(Tree.Kind.MEMBER_SELECT)) {
+        ExpressionTree expression = ((MemberSelectExpressionTree) wanderer).expression();
+        if (!expression.is(Tree.Kind.METHOD_INVOCATION)) {
+          break;
+        }
+        MethodInvocationTree mit = (MethodInvocationTree) expression;
+        inspectCall(mit);
+        wanderer = mit.methodSelect();
```
christophe-zurn-sonarsource, line 197: Can we reverse the condition on this `if` here and not use a `break` as this is not necessary anymore

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,216 @@
...
+      }
+      VariableTree variable = (VariableTree) declaration;
+      ExpressionTree initializer = variable.initializer();
+      if (initializer == null || !initializer.is(Tree.Kind.STRING_LITERAL)) {
+        return;
+      }
+      String pattern = ((LiteralTree) initializer).value();
+      if (isInfringingPattern(pattern)) {
+        reportIssue(invocation, CHANGE_YEAR_FORMAT_WEEK_BASED_MESSAGE);
+      }
```
christophe-zurn-sonarsource, line 128: I believe it would be easier to read if we had less if/return statements by adding an helper method to retrieve the string value: ex: `getStringLiteralValue`

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
christophe-zurn-sonarsource, line 178: This is personal, and you can leave it as is if you prefer, but I'd find it slightly clearer it we renamed this class. Ex `ChainedInvocationVisitor`

christophe-zurn-sonarsource (changes requested):
I've left mostly minor improvement suggestions, otherwise it is a nice update on the original implementation :+1:

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
dorian-burihabwa-sonarsource: The test can only be moved higher in the while loop condition because the code we will either need to exit the loop or try to access method select on an object that does not support it.

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,221 @@
...
+      Tree declaration = symbol.declaration();
+      Optional<String> pattern = getStringLiteralValue(declaration);
+      if (pattern.isPresent() && isInfringingPattern(pattern.get())) {
+        reportIssue(invocation, CHANGE_YEAR_FORMAT_WEEK_BASED_MESSAGE);
+      }
+    }
+  }
+
+  private static Optional<String> getStringLiteralValue(Tree declaration) {
```
christophe-zurn-sonarsource, line 125: We should add the `Nullable` annotation to the `Tree declaration` parameter

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,221 @@
...
+      return WEEK_OF_WEEK_BASED_YEAR_MATCHER.matches(call);
+    }
+    return false;
+  }
+
+  private static boolean isChronoFieldWeek(ExpressionTree argument) {
+    if (argument.is(Tree.Kind.MEMBER_SELECT)) {
+      MemberSelectExpressionTree select = (MemberSelectExpressionTree) argument;
+      IdentifierTree identifier = select.identifier();
+      return identifier.name().equals("ALIGNED_WEEK_OF_YEAR");
```
christophe-zurn-sonarsource, line 157: For safety, we could also check the experssion underlying type:
`return select.symbolType().is("java.time.temporal.ChronoField") && identifier.name().equals("ALIGNED_WEEK_OF_YEAR");`

java-checks/src/main/java/org/sonar/java/checks/DateTimeFormatterMismatchCheck.java
```diff
@@ -0,0 +1,221 @@
...
+      return WEEK_BASED_YEAR_MATCHER.matches(call);
+    }
+    return false;
+  }
+
+  private static boolean isChronoFieldYear(ExpressionTree argument) {
+    if (argument.is(Tree.Kind.MEMBER_SELECT)) {
+      MemberSelectExpressionTree select = (MemberSelectExpressionTree) argument;
+      IdentifierTree identifier = select.identifier();
+      return identifier.name().equals("YEAR");
```
christophe-zurn-sonarsource, line 178: Same here:
`return select.symbolType().is("java.time.temporal.ChronoField") && identifier.name().equals("YEAR");`

christophe-zurn-sonarsource (approved):
LGTM.
Please have a look at some last possible improvements, as well as the code smells before merging

PR #3366, opened 2020-12-28
SONARJAVA-3630 Rule S6103: AssertJ assertions with "Consumer" arguments should contain assertion inside consumers

christophe-zurn-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM

PR #3370, opened 2021-01-05
SONARJAVA-3651 S4925: add support for DB2 JDBC Type 4 driver

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/java/checks/JdbcDriverExplicitLoadingCheck_no_version.java
```diff
@@ -1,5 +1,7 @@
-class A {
-  void doStuff() {
+package checks;
+
+class JdbcDriverExplicitLoadingCheck_no_version {
+  void doStuff() throws ClassNotFoundException {
     Class.forName("org.h2.Driver"); // Noncompliant [[sc=11;ec=18]] {{Remove this "Class.forName()", it is useless. (sonar.java.source not set. Assuming 6 or greater.)}}
     Class.forName("java.lang.String");
     Class.forName(javax.print.ServiceUIFactory.DIALOG_UI);
```
dorian-burihabwa-sonarsource, line 7: For completeness, we should add a call with a non compliant string defined in the scope of this class

quentin-jaquier-sonarsource: If I understand correctly, you mean to add code like this:

```java
  private static final String DRIVER = "com.mysql.jdbc.Driver";

  void doStuff() throws ClassNotFoundException {
    Class.forName(DRIVER); // Noncompliant
  }
```

In this file? My feeling is that we already test this behavior in the other file, this one is here to test the message with an unknown version and should not duplicate what is tested in the other one.

Does it make sense?

dorian-burihabwa-sonarsource: Then I am wondering why we are using two test-source files with overlapping cases, including this one that has a limited number of test cases.
Couldn't we use a single test-source file for this rule?

quentin-jaquier-sonarsource: In fact, we are testing that the message is not the same when the version of java is not set. But I agree that if we really want to keep this file minimal, we should remove the line `Class.forName(javax.print.ServiceUIFactory.DIALOG_UI);`.

dorian-burihabwa-sonarsource: Good plan

PR #3373, opened 2021-01-06
SONARJAVA-3654 Add support of MongoDB nullness annotations

m-g-sonar:


quentin-jaquier-sonarsource (commented):
I'm not sure we are reliably testing the code about global annotations. The two tests testing the logic for Spring and Eclipse annotations are ignored in `NullableAnnotationUtilsTest` and failing when run locally.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3374, opened 2021-01-06
SONARJAVA-3614, SONARJAVA-3655: Mockito argument checkers use

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/MockitoArgumentMatchersUsedOnAllParametersCheck.java
```diff
@@ -52,25 +52,25 @@
...
     }
 
-    if (argumentMatchers.isEmpty()) {
-      return;
-    }
-    if (argumentMatchers.size() < arguments.size()) {
-      reportIssue(secondaries.get(0),
-        "Add an \"eq()\" argument matcher on this/these parameters",
-        secondaries.stream()
+    if (0 < nonMatchers.size() && nonMatchers.size() < arguments.size()) {
```
quentin-jaquier-sonarsource, line 66: You can extract the `size()` call to a variable (used later aswell) and use `isEmpty()` here.

quentin-jaquier-sonarsource (commented):
Except for one little refactoring, LGTM.

At this point, I'm not sure we will be able to have clean separate commits... 
So don't insist too much, I think squashing everything on master is good enough.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3376, opened 2021-01-07
Clean ITs and add some new for @SuppressWarnings and Package annotations

m-g-sonar:


quentin-jaquier-sonarsource (approved):
LGTM, thanks for cleaning and improving its!

PR #3377, opened 2021-01-08
SONARJAVA-3480: Revert implementation of S5917

dorian-burihabwa-sonarsource:
Revert implementation of S5917 for future integration to S3986 (see SONARJAVA-3657)


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3382, opened 2021-01-12
Revert "SONARJAVA-3576 S2095 should raise an issue when a stream is created from an ignored stream"

quentin-jaquier-sonarsource:


m-g-sonar:
LGTM!

PR #3384, opened 2021-01-12
SONARJAVA-3487 & SONARJAVA-3489

sebastian-hungerecker-sonarsource:
I included both fixes in one PR because my fix for SONARJAVA-3489 depends on the fix for SONARJAVA-3487.


java-checks/src/main/java/org/sonar/java/checks/MissingDeprecatedCheck.java
```diff
@@ -48,12 +47,7 @@ public void setContext(JavaFileScannerContext context) {
 
   @Override
   public void visitNode(Tree tree) {
-    boolean isLocalVar = false;
-    if (tree.is(Tree.Kind.VARIABLE)) {
-      isLocalVar = currentParent.peek().is(METHOD_KINDS);
-    } else {
-      currentParent.push(tree);
-    }
+    boolean isLocalVar = tree.is(Tree.Kind.VARIABLE) && ((VariableTree) tree).symbol().owner().isMethodSymbol();
```
margarita-nedzelska-sonarsource, line 50: This check seems to be quite useful. Maybe a good candidate to move it in some utils?

java-checks/src/main/java/org/sonar/java/checks/MissingDeprecatedCheck.java
```diff
@@ -27,16 +27,15 @@
...
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.Tree.Kind;
+import org.sonar.plugins.java.api.tree.VariableTree;
 import org.sonarsource.analyzer.commons.annotations.DeprecatedRuleKey;
 
 @DeprecatedRuleKey(ruleKey = "MissingDeprecatedCheck", repositoryKey = "squid")
 @Rule(key = "S1123")
 public class MissingDeprecatedCheck extends AbstractDeprecatedChecker {
 
   private static final Kind[] CLASS_KINDS = PublicApiChecker.classKinds();
```
margarita-nedzelska-sonarsource, line 37: Do you think we should add Records to CLASS_KINDS? It looks like it might affect other checks so is out of the scope for this ticket.

margarita-nedzelska-sonarsource (approved):
Looks good!

java-checks/src/main/java/org/sonar/java/checks/MissingDeprecatedCheck.java
sebastian-hungerecker-sonarsource: As I understand it, when it comes to preview features, we want to do the least possible changes necessary to keep the analyzer from crashing. Records being missing from `CLASS_KINDS` doesn't seem to cause any crashes, so I'd say it should remain unchanged for now.

margarita-nedzelska-sonarsource: Yep, let's leave it as is.

margarita-nedzelska-sonarsource (approved):
NIce work!

PR #3387, opened 2021-01-12
SONARJAVA-3488 [Java 14 - Records preview feature] NPE when computing…

sebastian-hungerecker-sonarsource:
… metrics of methods

### DIFF
--- a/java-frontend/src/test/files/metrics/Methods.java
+++ b/java-frontend/src/test/files/metrics/Methods.java
@@ -1,8 +1,8 @@
 class MyClass {
-  public MyClass() {
+  public MyClass() {  // counted
   }
 
-  public void method() {
+  public void method() {  // counted
   }
 
   public <T> List<T> genericMethod(T param) {  // counted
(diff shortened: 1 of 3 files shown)


java-frontend/src/test/files/metrics/Methods.java
```diff
@@ -36,4 +36,8 @@ int method(){  // counted
     }
   };
   abstract int method();  // counted
-}
\ No newline at end of file
+}
+
+record R(int x) {
+  int neg() { return -x; }
```
dorian-burihabwa-sonarsource, line 42: We probably should add "//counted" comment next to the method for documentation consistency

dorian-burihabwa-sonarsource (commented):
LGTM. Adding a //counted comment would help with the understanding of the test file.

java-frontend/src/test/files/metrics/Methods.java
sebastian-hungerecker-sonarsource: Good catch. Will do. Looks like I didn't look at the existing code carefully enough.

dorian-burihabwa-sonarsource (approved):

PR #3388, opened 2021-01-13
SONARJAVA-3658 Support TEXT_BLOCKs

margarita-nedzelska-sonarsource:
Should be rebased.

Firstly, I thought that this ticket is also regarding rules. So I started adding support in rules. I think if we do "rebase & merge" it's fine to leave it here.

For the convenience it's better to review by commits.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/ControlCharacterInLiteralCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ControlCharacterInLiteralCheck.java
@@ -54,7 +54,7 @@ public class ControlCharacterInLiteralCheck extends IssuableSubscriptionVisitor 
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.STRING_LITERAL, Tree.Kind.CHAR_LITERAL);
+    return Arrays.asList(Tree.Kind.STRING_LITERAL, Tree.Kind.CHAR_LITERAL, Tree.Kind.TEXT_BLOCK);
   }
 
   @Override
--- a/java-frontend/src/main/java/org/sonar/java/AnalyzerMessage.java
+++ b/java-frontend/src/main/java/org/sonar/java/AnalyzerMessage.java
@@ -24,6 +24,7 @@ import java.util.List;
 import javax.annotation.Nullable;
 import org.sonar.api.batch.fs.InputComponent;
 import org.sonar.plugins.java.api.JavaCheck;
+import org.sonar.plugins.java.api.tree.LiteralTree;
 import org.sonar.plugins.java.api.tree.SyntaxToken;
 import org.sonar.plugins.java.api.tree.Tree;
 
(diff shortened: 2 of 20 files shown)


java-frontend/src/main/java/org/sonar/java/AnalyzerMessage.java
```diff
@@ -117,6 +117,10 @@ public boolean isEmpty() {
   }
 
   public static AnalyzerMessage.TextSpan textSpanFor(Tree syntaxNode) {
+    if (syntaxNode.is(Tree.Kind.TEXT_BLOCK)) {
+      SyntaxToken syntaxToken = getNonEmptyTree(syntaxNode).firstToken();
+      return textSpanForTextBlock(syntaxToken);
```
m-g-sonar, line 122: if we already know it's a `TEXT_BLOCK`, I feel we could bypass the `getNonEmptyTree()` and directly call:
```java
return textSpanForTextBlock(((LiteralTree) syntaxNode).token());
```

java-frontend/src/main/java/org/sonar/java/AnalyzerMessage.java
```diff
@@ -141,6 +145,23 @@ public boolean isEmpty() {
     return location;
   }
 
+  private static AnalyzerMessage.TextSpan textSpanForTextBlock(SyntaxToken syntaxToken) {
+    String text = syntaxToken.text();
+    String[] lines = text.split("\n");
```
m-g-sonar, line 150: I'm not sure about this. Depending of the OS, it might not be only `\n`. I would use: `"(\r)?\n|\r"`

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/Tree.java
```diff
@@ -547,6 +547,11 @@
      */
     STRING_LITERAL(LiteralTree.class),
 
+    /**
+     * {@link LiteralTree}
+     */
+    TEXT_BLOCK(LiteralTree.class),
```
m-g-sonar, line 553: You might want to modify the LiteralTree class as well to mark in the comments that there is also this kind to consider.

java-checks/src/main/java/org/sonar/java/checks/ControlCharacterInLiteralCheck.java
```diff
@@ -54,7 +54,7 @@
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.STRING_LITERAL, Tree.Kind.CHAR_LITERAL);
+    return Arrays.asList(Tree.Kind.STRING_LITERAL, Tree.Kind.CHAR_LITERAL, Tree.Kind.TEXT_BLOCK);
```
m-g-sonar, line 57: maybe should have been in another commit?

m-g-sonar (commented):
Great workü
On top of my comments, I would remove the brackets around the ticket number in each commit title. No need to introduce something new, we always prefix by the ticket, it's simple enough.

m-g-sonar (approved):

PR #3390, opened 2021-01-14
SONARJAVA-3563 Report 10 slowest analyzed files

alban-auzeill:


alban-auzeill:
I executed two times the "ruling" of this PR and here is the result:
* 1st time:
```
[INFO] Slowest analyzed files: /tmp/cirrus-ci-build/its/sources/sonarqube-6.5/server/sonar-server/src/main/java/org/sonar/server/issue/IssueQueryFactory.java (2563ms), /tmp/cirrus-ci-build/its/sources/sonarqube-6.5/server/sonar-server/src/main/java/org/sonar/ce/CeModule.java (1695ms), /tmp/cirrus-ci-build/its/sources/sonarqube-6.5/server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java (1326ms), /tmp/cirrus-ci-build/its/sources/sonarqube-6.5/server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java (1011ms)
```
* current "ruling" log:
```
[INFO] Slowest analyzed files: /tmp/cirrus-ci-build/its/sources/sonarqube-6.5/server/sonar-server/src/main/java/org/sonar/server/issue/IssueQueryFactory.java (2008ms), /tmp/cirrus-ci-build/its/sources/sonarqube-6.5/server/sonar-server/src/main/java/org/sonar/ce/CeModule.java (1358ms), /tmp/cirrus-ci-build/its/sources/sonarqube-6.5/server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java (1302ms)
```
We can see that `IssueQueryFactory.java` (with 370 lines and 15'231 bytes) is the slowest file to analyze.
We should probably investigate why on another ticket...

java-frontend/src/main/java/org/sonar/java/ast/JavaAstScanner.java
```diff
@@ -64,14 +66,18 @@ public void scan(Iterable<? extends InputFile> inputFiles) {
...
     try {
+      ExecutionTimeReport executionTimeReport = new ExecutionTimeReport(Clock.systemUTC());
       for (InputFile inputFile : inputFiles) {
         if (analysisCancelled()) {
           cancelled = true;
           break;
         }
+        executionTimeReport.start(inputFile.uri().getPath());
         simpleScan(inputFile);
+        executionTimeReport.end();
```
m-g-sonar, line 77: I feel that this call to `end()` and probably a call to `report()` should be part of the finally block as well, in case of interruption.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
```diff
@@ -0,0 +1,101 @@
...
+    private final long analysisTime;
+
+    public ExecutionTime(String file, long analysisTime) {
+      this.file = file;
+      this.analysisTime = analysisTime;
+    }
+  }
+
+  private static final Comparator<ExecutionTime> ORDER_BY_ANALYSIS_TIME_DESCENDING_AND_FILE_ASCENDING = (a, b) -> {
+    int compare = -Long.compare(a.analysisTime, b.analysisTime);
```
m-g-sonar, line 47: why making it minus? if you know that b is after, just invert args.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
```diff
@@ -0,0 +1,101 @@
...
+    if (compare == 0) {
+      compare = a.file.compareTo(b.file);
```
m-g-sonar, line 49: worth a comment that we default to filename, also, you can directly return instead of reassigning.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
```diff
@@ -0,0 +1,101 @@
...
+    this.currentFile = currentFile;
+    currentFileStartTimeMS = clock.millis();
+  }
+
+  public void end() {
+    long currentAnalysisTime = clock.millis() - currentFileStartTimeMS;
+    if (currentAnalysisTime >= minRecordedOrderedExecutionTime) {
+      recordedOrderedExecutionTime.add(new ExecutionTime(currentFile, currentAnalysisTime));
+      recordedOrderedExecutionTime.sort(ORDER_BY_ANALYSIS_TIME_DESCENDING_AND_FILE_ASCENDING);
+      if (recordedOrderedExecutionTime.size() > MAX_REPORTED_FILES) {
```
m-g-sonar, line 77: To be honest I feel that we should filter out only at the end of the full analysis. We might be interested in seeing everything, maybe if set logs to DEBUG.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
```diff
@@ -0,0 +1,101 @@
...
+    if (analysisEndTimeMS >= MIN_TOTAL_ANALYSIS_TIME_TO_REPORT_MS && !recordedOrderedExecutionTime.isEmpty()) {
+      LOG.info("Slowest analyzed files: " + toString());
+    }
+  }
+
+  @Override
+  public String toString() {
+    return recordedOrderedExecutionTime.stream()
+      .map(e -> e.file + " (" + e.analysisTime + "ms)")
+      .collect(Collectors.joining(", "));
```
m-g-sonar, line 98: maybe on different lines?

java-frontend/src/main/java/org/sonar/java/ast/JavaAstScanner.java
```diff
@@ -64,14 +66,18 @@ public void scan(Iterable<? extends InputFile> inputFiles) {
     boolean successfullyCompleted = false;
     boolean cancelled = false;
     try {
+      ExecutionTimeReport executionTimeReport = new ExecutionTimeReport(Clock.systemUTC());
       for (InputFile inputFile : inputFiles) {
         if (analysisCancelled()) {
           cancelled = true;
           break;
         }
+        executionTimeReport.start(inputFile.uri().getPath());
```
m-g-sonar, line 75: I would use the inputFile.toString(), it displays a readable understandable by end users

java-frontend/src/main/java/org/sonar/java/ast/JavaAstScanner.java
alban-auzeill: Good point, in case of a never-ending analysis, it could be nice to see the report. I'm moving the `report()` in the finally block and make it works when `end()` is missing.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
alban-auzeill: I started by inverting the args and then thought it would be more explicit to keep the same order in all this method and materialize the `descending` order by a minus. But, apparently, it generates more questions than answers, so ok, I'm inverting the args.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
alban-auzeill: The goal of the extra-long variable name `ORDER_BY_ANALYSIS_TIME_DESCENDING_AND_FILE_ASCENDING` was to not require a comment.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
alban-auzeill: My fear was to store a lot of data or manipulate too often the `recordedOrderedExecutionTime` list.
That's why there's this non-costly test `if (currentAnalysisTime >= minRecordedOrderedExecutionTime) {`
Once the list is full and the `minRecordedOrderedExecutionTime` strat to increase, it's super rare entering this test.
For example, on the ruling, the list contains only 3 or 4 elements and we never enter into this filter.

But I like the idea of logging more execution time in DEBUG mode. My proposal is to log each execution time, file by file, but probably more in TRACE mode than in DEBUG mode.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
alban-auzeill: I know that it's not easy to read like this, but there's no other log INFO on multiple lines and it should not be confused with a stack trace. So I prefer to keep it like this.

java-frontend/src/main/java/org/sonar/java/ast/JavaAstScanner.java
alban-auzeill: ok, I will try

alban-auzeill: good try, paths are more readable:
```
[INFO] Slowest analyzed files: src/main/java/org/sonar/server/issue/IssueQueryFactory.java (2223ms), src/main/java/org/sonar/ce/CeModule.java (1507ms), src/main/java/org/sonar/server/rule/RegisterRules.java (1298ms)
```

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
```diff
@@ -0,0 +1,107 @@
...
+  public void end() {
+    long currentAnalysisTime = clock.millis() - currentFileStartTimeMS;
+    if (LOG.isTraceEnabled()) {
+      LOG.trace("Analysis time of " + currentFile + " (" + currentAnalysisTime + "ms)");
+    } else if (currentAnalysisTime >= MIN_REPORTED_ANALYSIS_TIME_MS && LOG.isDebugEnabled()) {
+      LOG.debug("Analysis time of " + currentFile + " (" + currentAnalysisTime + "ms)");
+    }
+    if (currentAnalysisTime >= minRecordedOrderedExecutionTime) {
+      recordedOrderedExecutionTime.add(new ExecutionTime(currentFile, currentAnalysisTime));
+      recordedOrderedExecutionTime.sort(ORDER_BY_ANALYSIS_TIME_DESCENDING_AND_FILE_ASCENDING);
```
m-g-sonar, line 78: I'm not sure we would need to provide a Comparator if you simply make `ExecutionTime` a `Comparable`, then you can just implement the `Compare()` method.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
```diff
@@ -0,0 +1,107 @@
...
+    if (analysisEndTimeMS >= MIN_TOTAL_ANALYSIS_TIME_TO_REPORT_MS && !recordedOrderedExecutionTime.isEmpty()) {
+      LOG.info("Slowest analyzed files: " + toString());
+    }
+  }
+
+  @Override
+  public String toString() {
+    return recordedOrderedExecutionTime.stream()
+      .map(e -> e.file + " (" + e.analysisTime + "ms)")
+      .collect(Collectors.joining(", "));
```
m-g-sonar, line 104: I really think it is OK to have them on multiple lines, especially since we are displaying the path of each input file (it take place) and only 10 files. I prefer have 10 explicit lines rather than one big chunk of text,

m-g-sonar (commented):
Pretty nice work! almost there!

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
alban-auzeill: Doing so will also require to implement the equals/hashCode methods to comply with rule S1210. So I prefer keeping the implementation like this.

java-frontend/src/main/java/org/sonar/java/ExecutionTimeReport.java
alban-auzeill: ok

alban-auzeill: The new logs look like this:
```
[INFO] 1370/1370 source files have been analyzed
[INFO] Slowest analyzed files:
    src/main/java/org/sonar/server/issue/IssueQueryFactory.java (2143ms)
    src/main/java/org/sonar/ce/CeModule.java (1363ms)
    src/main/java/org/sonar/server/rule/RegisterRules.java (1288ms)
[INFO] Java Main Files AST scan (done) | time=87830ms
```

PR #3391, opened 2021-01-14
Fix its expecting default Java version in maven

margarita-nedzelska-sonarsource:
Starting from [this commit](https://github.com/SonarSource/sonar-scanner-maven/commit/0d618b5b8a622aa8c97232bb5acb3fb488387e8a) release version configured in pom.xml will be considered as `sonar.java.source` version.

Our plugin test projects are all configured to run on java 11 using this property in parent's pom.xml. 
To keep testing older versions of java in this test I removed parent section only from this project. So now `sonar.java.source` version won't be overridden and we'll get the right fallback from maven, which is Java 5


quentin-jaquier-sonarsource (approved):
LGTM, thanks for investigating this problem

PR #3392, opened 2021-01-14
SONARJAVA-3606 S5979: Ensure that annotated objects are properly initialized

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
```diff
@@ -0,0 +1,177 @@
...
+public class MockitoAnnotatedObjectsShouldBeInitializedCheck extends IssuableSubscriptionVisitor {
+  private static final List<String> TARGET_ANNOTATIONS = Arrays.asList(
+    "Captor",
+    "InjectMocks",
+    "Mock",
+    "Spy"
+  );
+
+  private static final List<String> EXPECTED_CLASS_ANNOTATIONS = Arrays.asList(
+    "ExtendWith",
```
alban-auzeill, line 49: We should add one test about `@ExtendWith`

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
```diff
@@ -0,0 +1,177 @@
...
+
+  private static boolean mocksAreProperlyInitialized(ClassTree testClass) {
+    return isClassProperlyAnnotated(testClass) ||
+      isMockitoJUnitRuleInvoked(testClass) ||
+      areMocksInitializedInSetup(testClass);
+  }
+
+  public static boolean isClassProperlyAnnotated(ClassTree clazz) {
+    List<AnnotationTree> annotations = clazz.modifiers().annotations();
+    return !getAnnotations(annotations, EXPECTED_CLASS_ANNOTATIONS).isEmpty();
```
alban-auzeill, line 101: We don't test if it's `MockitoJUnitRunner` or `MockitoExtension`. It's only false-negatives, but we should add a test to make it explicit by adding a false-negative with a class annotated with for example `@RunWith(Parameterized.class)`

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  private static boolean isMockitoJUnitRuleInvoked(ClassTree clazz) {
+    List<VariableTree> collected = clazz.members().stream()
+      .filter(member -> member.is(Tree.Kind.VARIABLE))
+      .map(VariableTree.class::cast)
+      .collect(Collectors.toList());
+    for (VariableTree field : collected) {
+      String type = field.type().symbolType().fullyQualifiedName();
+      if (type.equals("org.mockito.junit.MockitoRule")) {
```
alban-auzeill, line 111: we can use `field.type().symbolType().is("org.mockito.junit.MockitoRule")`

java-checks/src/test/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheckTest.java
```diff
@@ -0,0 +1,42 @@
...
+  void test() {
+    JavaCheckVerifier.newVerifier()
+      .onFile(testSourcesPath("checks/tests/MockitoAnnotatedObjectsShouldBeInitialized.java"))
+      .withCheck(new MockitoAnnotatedObjectsShouldBeInitializedCheck())
+      .verifyIssues();
+
+    JavaCheckVerifier.newVerifier()
+      .onFile(nonCompilingTestSourcesPath("checks/tests/MockitoAnnotatedObjectsShouldBeInitialized.java"))
+      .withCheck(new MockitoAnnotatedObjectsShouldBeInitializedCheck())
+      .verifyNoIssues();
```
alban-auzeill, line 40: We should move those lines in a different test with a name that explains why we need some code that does not compile. Reading this PR I don't see the reason.

alban-auzeill (changes requested):

dorian-burihabwa-sonarsource:
Thanks for the feedback. Correcting now.

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
```diff
@@ -0,0 +1,191 @@
...
+    return false;
+  }
+
+  private static boolean areMocksInitializedInSetup(ClassTree clazz) {
+    List<MethodTree> methods = new ArrayList<>();
+    Optional<ClassTree> parent = Optional.of(clazz);
+    while (parent.isPresent()) {
+      ClassTree tree = parent.get();
+      methods.addAll(getSetupMethods(tree));
+      parent = getParentClass(tree);
```
alban-auzeill, line 134: In our unit test, it's a corner case where from a ClassTree we can access to the super ClassTree because all code is in the same file.
Usually, from a ClassTree we can only access to the semantic of the superclass, but not the AST.
With the semantic, it's not possible to apply SetupMethodVisitor.
So, I would just not support the case where the ClassTree have superClass() as a limitation of the rule, with a unit test that show a false-negative, and remove this getParentClass logic.

alban-auzeill (changes requested):

alban-auzeill (approved):

mminot-yseop:
@dorian-burihabwa-sonarsource 

Hi there.

This rule doesn’t seem to take JUnit 5’s `@Nested` classes into consideration (or rather, it considers them too individually, without looking at the parents).
This working test triggers the rule:

```java
@ExtendWith(MockitoExtension.class)
public class S5979Test {

    @Nested
    class NestedClass {

        @Mock
        Object myMock;

        @Test
        void myTest() {
            assertThat(myMock).isNotNull();
        }
    }
}
```

… and I’m forced to add a `@ExtendWith(MockitoExtension.class)` on the nested class itself, which seems redundant, or to move it from the root class to the nested class, but then I cannot add mock fields to the root class (and I’d run into the same issue again in other, sibbling `@Nested` classes anyway).

(Sorry for popping up like this on the PR; 🙇🏻‍♀️ creating an account on the community thingy in the last few minutes of the last working day of the week seemed kinda overkill when the rule-creating PR was right under my nose. I was hoping you could create an issue yourself or whatever.)

# Extra info

My real case looked like this:

```java
@ExtendWith(MockitoExtension.class)
@MockitoSettings(strictness = Strictness.STRICT_STUBS)
class YseopManagerCmdUtilsTest {

    @Nested
    class BuildRestTemplate {

        @Mock
        RestTemplateBuilder mockRestBuilder;

        @Mock
        RestTemplate mockRest;

        @BeforeEach
        void beforeEach() {
            when(mockRestBuilder.basicAuthentication("uuu", "ppp", StandardCharsets.UTF_8))
                    .thenReturn(mockRestBuilder);
[…]
```

We’re using SonarCloud: https://sonarcloud.io/organizations/yseop
Thanks for reading.

dorian-burihabwa-sonarsource:
Hello @mminot-yseop,
Thank you for your feedback. A [ticket](https://jira.sonarsource.com/browse/SONARJAVA-3687) was created to investigate the issue.
In the future please make sure to report issues on the [community forums](https://community.sonarsource.com/) as your feedback might prove useful to many.

mminot-yseop:
Thanks! (btw, there can be nested classes in nested classes, etc.)

PR #3393, opened 2021-01-15
SONARJAVA-3255 Rule S5663: Simple string literal should be used for single line strings

margarita-nedzelska-sonarsource:


java-checks-test-sources/src/main/files/non-compiling/checks/SimpleStringLiteralForSingleLineStringsCheck.java
```diff
@@ -0,0 +1,19 @@
...
+  
+  public void str() { // Noncompliant@+1 [sc=23;endColumn=42]]{{Use simple literal for a single-line string.}}
+    String question = """
+              What's the point, really?""";
+
+    String question1 = "What's the point, really?"; // Compliant
+
+    String question2 = """
+              What's the point, really
+              ?"""; // Compliant, 3 lines
```
quentin-jaquier-sonarsource, line 13: This is 3 lines for our engine (when you call `textBlock.value()`) but it represents a two lines string in java, right?

quentin-jaquier-sonarsource (approved):
LGTM
Just a quick question, no need to change anything.

java-checks-test-sources/src/main/files/non-compiling/checks/SimpleStringLiteralForSingleLineStringsCheck.java
margarita-nedzelska-sonarsource: yes. It's actually 2 lines, because it's not possible to write `"""What's the point, really?"""`

PR #3394, opened 2021-01-15
SONARJAVA-2961 Rule S4977: Type parameters should not shadow other type parameters

quentin-jaquier-sonarsource:
My first implementation was raising hundreds of issues on static members. This is arguable because it can indeed be kind of confusing, but you are not technically shadowing anything because you can not use the type parameter from an outer scope in a static member.

So I went for not reporting static members, as Intellij is doing.

Let me know what you think!


sebastian-hungerecker-sonarsource:
I 100% agree with ignoring static members. In fact, in many cases I would say it's bad style if the type parameters of a static method *didn't* have the same names as those of the class - e.g. in a class `MyMap<K, V>` something like `MyMap<K, V> of<K, V>(K k1, V v1)` makes a lot more sense than if the type parameters were called something else.

java-checks/src/main/java/org/sonar/java/checks/TypeParametersShadowingCheck.java
```diff
@@ -0,0 +1,109 @@
...
+
+@Rule(key = "S4977")
+public class TypeParametersShadowingCheck extends BaseTreeVisitor implements JavaFileScanner {
+
+  private static final String ISSUE_MESSAGE = "Rename \"%s\" which hides a type parameter from the outer scope.";
+
+  private JavaFileScannerContext context;
+
+  private Map<String, IdentifierTree> currentTypeParametersInScope;
+  private Deque<Map<String, IdentifierTree>> scopes;
```
sebastian-hungerecker-sonarsource, line 47: You don't need an explicit stack when using `BaseTreeVisitor`. You can just set `currentTypeParametersInScope` at the beginning of `processTree` (storing the old value in a local variable) and set it back to its old value at the end.

java-checks-test-sources/src/main/java/checks/TypeParametersShadowingCheck.java
```diff
@@ -0,0 +1,108 @@
+package checks;
+
+import java.util.List;
+import java.util.function.Function;
+
+public class TypeParametersShadowingCheck<T0> {
+  class TypeParameterHidesAnotherType<T> {
+    public class Inner<T> { // Noncompliant [[sc=24;ec=25;secondary=-1]] {{Rename "T" which hides a type parameter from the outer scope.}}
+      //...
+    }
```
sebastian-hungerecker-sonarsource, line 10: I'd like a second inner class here that shadows T0 instead of T (`public class Inner2<T0> {}`) to test the fact that we don't forget about `TypeParametersShadowingCheck`'s type parameters when entering `TypeParameterHidesAnotherType`.

java-checks/src/main/java/org/sonar/java/checks/TypeParametersShadowingCheck.java
```diff
@@ -0,0 +1,109 @@
...
+      String name = id.toString();
+
+      IdentifierTree shadowedId = currentTypeParametersInScope.get(name);
+      if (shadowedId != null) {
+        context.reportIssue(this, id,
+          String.format(ISSUE_MESSAGE, name),
+          Collections.singletonList(new JavaFileScannerContext.Location("Shadowed type parameter", shadowedId)
+        ), null);
+      } else {
+        // Both collections updated only in the else part, because we want to store only the first and outer most apparition of a type.
```
sebastian-hungerecker-sonarsource, line 101: s/apparition/appearance/

java-checks/src/main/java/org/sonar/java/checks/TypeParametersShadowingCheck.java
```diff
@@ -0,0 +1,109 @@
...
+    processTree(tree, tree.typeParameters(), tree.symbol().isStatic(), super::visitMethod);
+  }
+
+  private <T> void processTree(T tree, TypeParameters typeParameters, boolean isStatic, Consumer<T> visitTree) {
+    if (isStatic) {
+      pushNewScope();
+    }
+    List<String> declaredTypeParameters = processAndGetTypeParameters(typeParameters);
+    visitTree.accept(tree);
+    declaredTypeParameters.forEach(currentTypeParametersInScope::remove);
```
sebastian-hungerecker-sonarsource, line 73: I think it would be better if the new entries were put into `currentTypeParametersInScope` here instead of in `processAndTypeParameters`:

    List<String> declaredTypeParameters = processAndGetTypeParameters(typeParameters);
    currentTypeParamtersInScope.putAll(declaredTypeParameters);
    visitTree.accept(tree);
    declaredTypeParameters.forEach(currentTypeParametersInScope::remove);

This way the logic of "add entries, visit children, remove entries" becomes more apparent and the addition step and the remove step are more symmetric.

sebastian-hungerecker-sonarsource (changes requested):
I've left some notes. Looks good otherwise.

java-checks/src/main/java/org/sonar/java/checks/TypeParametersShadowingCheck.java
quentin-jaquier-sonarsource: Indeed, nice improvment!

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3395, opened 2021-01-19
SONARJAVA-3661 FP on S2259 (Null Pointer Dereference) when using MapUtils from Apache Collections

alban-auzeill:


alban-auzeill:
Note for the reviewer:

- `defaultIfNull` has been generated using `SymbolicExecutionVisitor` and `MethodBehaviorJsonAdapter.gson().toJson`
- `isEmpty` and `isNotEmpty` have been copied from the other existing behaviors and does not match what the `SymbolicExecutionVisitor` would have generated:
```
{
  "signature": "........#isEmpty(LIterator;)Z",
  "varArgs": false,
  "declaredExceptions": [],
  "yields": [
    {
      "parametersConstraints": [
        ["NOT_NULL"]
      ],
      "exception": null
    },
    {
      "parametersConstraints": [
        ["NULL"]
      ],
      "resultIndex": -1,
      "resultConstraint": [
        "NOT_NULL",
        "TRUE"
      ]
    },
    {
      "parametersConstraints": [
        ["NOT_NULL"]
      ],
      "resultIndex": -1,
      "resultConstraint": null
    }
  ]
}
```

```
{
  "signature": "........#isNotEmpty(LMap;)Z",
  "varArgs": false,
  "declaredExceptions": [],
  "yields": [
    {
      "parametersConstraints": [
        ["NOT_NULL"]
      ],
      "exception": null
    },
    {
      "parametersConstraints": [
        ["NOT_NULL"]
      ],
      "resultIndex": -1,
      "resultConstraint": null
    },
    {
      "parametersConstraints": [
        ["NULL"]
      ],
      "resultIndex": -1,
      "resultConstraint": [
        "NOT_NULL",
        "FALSE"
      ]
    }
  ]
}
```

m-g-sonar:
The modification LGTM, so I'm OK to merge it.
However, there is something I don't get from the behavior you just warned about:
```
{
  "signature": "........#isNotEmpty(LMap;)Z",
  "varArgs": false,
  "declaredExceptions": [],
  "yields": [
    {
      "parametersConstraints": [
        ["NOT_NULL"]
      ],
      "exception": null  <<<<<<<<<<<<<<<<< this is strange
    },
    {
      "parametersConstraints": [
        ["NOT_NULL"]
      ],
      "resultIndex": -1,
      "resultConstraint": null
    },
    {
      "parametersConstraints": [
        ["NULL"]
      ],
      "resultIndex": -1,
      "resultConstraint": [
        "NOT_NULL",
        "FALSE"
      ]
    }
  ]
}
```
Two comments:
* There is apparently a `ExceptionalMethodBehavior` being returned if the parameter is NOT_NULL... but the exception is unknown. I bet it comes from any "unexpected" issue which could occur when calling `isEmpty()` method on the arg, as detected by the engine. If it's the case, I feel we simply discarded such similar options when generating the first method behaviors.
* Did you try to use these generated behaviors to see if they are generating different results? I have the feeling they are equivalent!

alban-auzeill:
The json can not contains `"exception": null`, so the generated json is invalid, but is I change it to `"exception": "java.lang.IllegalArgumentException"` then I can replace in the json configuration the above isEmpty/isNotEmpty behaviors and the unit tests still pass.

PR #3396, opened 2021-01-19
RSPEC-5665 Escape sequences should not be used in text blocks

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
I've decided to report issues at line having improper character instead of the full Text Block, because it might have a lot of lines and it will be hard to locate the issue properly.

margarita-nedzelska-sonarsource:
While working on this issue faced a problem with line numbers and reported a bug in JDT: https://bugs.eclipse.org/bugs/show_bug.cgi?id=570472

java-checks/src/main/java/org/sonar/java/checks/UnnecessaryEscapeSequencesInTextBlockCheck.java
```diff
@@ -0,0 +1,60 @@
...
+import org.sonar.java.collections.SetUtils;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S5665")
+public class UnnecessaryEscapeSequencesInTextBlockCheck extends IssuableSubscriptionVisitor {
+  private static final String MESSAGE = "Remove this unnecessary escape sequence.";
+  private static final String TRIPLE_QUOTE_MESSAGE = "Use '\\\"\"\"' to escape \"\"\".";
+  private static final Set<String> ESCAPED = SetUtils.immutableSetOf("\\n", "\\t", "\\'", "\\\"");
```
sebastian-hungerecker-sonarsource, line 35: I wouldn't include `\t` in the list of escapes we complain about. I feel that using `\t` is a much better idea than putting literal tab characters into your string - whether it's inside a normal string literal or a text block.

java-checks/src/main/java/org/sonar/java/checks/UnnecessaryEscapeSequencesInTextBlockCheck.java
```diff
@@ -0,0 +1,60 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.TEXT_BLOCK);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    LiteralTree textBlock = (LiteralTree) tree;
+    int startLine = textBlock.token().line();
+    String value = textBlock.value();
+    String[] lines = value.split("(\r?\n)|\r");
```
sebastian-hungerecker-sonarsource, line 47: The parentheses around `\r?\n` are redundant.

sebastian-hungerecker-sonarsource (approved):
See comments. LGTM otherwise.

java-checks/src/main/java/org/sonar/java/checks/UnnecessaryEscapeSequencesInTextBlockCheck.java
margarita-nedzelska-sonarsource: This does make sense, thanks

PR #3397, opened 2021-01-19
SONARJAVA-3665 Add support of Text Blocks in S2973 (Escaped unicode characters)

margarita-nedzelska-sonarsource:
Waits for https://github.com/SonarSource/sonar-java/pull/3399 to be merged

### DIFF
--- a/java-checks-test-sources/src/main/files/non-compiling/checks/EscapedUnicodeCharactersTextBlocksCheck.java
+++ b/java-checks-test-sources/src/main/files/non-compiling/checks/EscapedUnicodeCharactersTextBlocksCheck.java
@@ -2,6 +2,17 @@ package checks;
 
 class EscapedUnicodeCharactersCheck {
   void method() {
+    String prefix0 = """
+      
+      
+      
+      
+      n\uuuu00E9e
+      
+      
+      
+      
+      """; // Noncompliant@-10 [[sc=22;ec=9]] {{Remove this Unicode escape sequence and use the character instead.}}
     String prefix = """
       n\uuuu00E9e"""; // Noncompliant@-1 [[sc=21;ec=20]] {{Remove this Unicode escape sequence and use the character instead.}}
     String prefix2= """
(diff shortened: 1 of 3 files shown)


java-checks/src/test/java/org/sonar/java/checks/EscapedUnicodeCharactersCheckTest.java
```diff
@@ -34,4 +35,12 @@ void test() {
       .verifyIssues();
   }
 
+  @Test
+  void testTextBlocks() {
```
dorian-burihabwa-sonarsource, line 39: We should rename this method in snake_case for consistency.

java-checks/src/main/java/org/sonar/java/checks/EscapedUnicodeCharactersCheck.java
```diff
@@ -54,6 +54,9 @@ public void visitNode(Tree node) {
       return;
     }
     String value = LiteralUtils.trimQuotes(((LiteralTree) node).value());
+    if (node.is(Kind.TEXT_BLOCK)) {
+      value = value.replaceAll("(\r?\n|\r)\\s*+", "");
```
dorian-burihabwa-sonarsource, line 58: Is the `+`  operator necessary here? The '*' just before already covers the same case. Or is the `+` supposed o apply to the entire expression?

java-checks-test-sources/src/main/files/non-compiling/checks/EscapedUnicodeCharactersTextBlocksCheck.java
```diff
@@ -0,0 +1,116 @@
+package checks;
+
+class EscapedUnicodeCharactersCheck {
+  void method() {
+    String prefix = """
+      n\uuuu00E9e"""; // Noncompliant@-1 [[sc=21;ec=20]] {{Remove this Unicode escape sequence and use the character instead.}}
```
dorian-burihabwa-sonarsource, line 6: We should probably include examples where multiple line breaks follow one another.
```java
    String prefix = """


      n\uuuu00E9e

""";
```

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/EscapedUnicodeCharactersCheck.java
margarita-nedzelska-sonarsource: Yes, possessive quantifier is odd in this context, copy/paste issue

dorian-burihabwa-sonarsource (approved):
LGTM :+1:  Merging now

PR #3398, opened 2021-01-19
SONARJAVA-3256 Rule S5664: Whitespace for text block indent should be…

sebastian-hungerecker-sonarsource:
… consistent


java-checks/src/main/java/org/sonar/java/checks/TextBlockTabsAndSpacesCheck.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  private static boolean isIndentation(char c) {
+    return c == ' ' || c == '\t';
+  }
+
+  private static boolean containsWrongIndentation(String[] lines) {
+    char indentationCharacter = indentationCharacter(lines);
+    for (String line : lines) {
+      for (int i = 0; i < line.length() && isIndentation(line.charAt(i)); i++) {
```
margarita-nedzelska-sonarsource, line 53: What do you think if we consider the indentation length of the first line as indentation length for all lines in text block:
```
String s = """
....<html>
....    <body>
....        <tag>
....        </tag>
....    </body>
....</html>""";
```

Like in this example we consider only the 4 first characters in each line as indentation.

java-checks/src/main/java/org/sonar/java/checks/TextBlockTabsAndSpacesCheck.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+    return false;
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    LiteralTree textBlock = (LiteralTree) tree;
+    String[] lines = textBlock.value().split("\r?\n");
+    if (containsWrongIndentation(lines)) {
+      reportIssue(tree, "Use only spaces or only tabs for indentation");
```
margarita-nedzelska-sonarsource, line 67: I am not exactly sure, but maybe we should report on each violated line rather than on the whole text block. So the issue might be easier to found. Wdyt?

margarita-nedzelska-sonarsource (commented):
Overall looks good, just see my comments

java-checks/src/main/java/org/sonar/java/checks/TextBlockTabsAndSpacesCheck.java
sebastian-hungerecker-sonarsource: I decided to use the actual indentation as determined by Java's rules rather than the indentation of the first line. For that I copied the code to determine that from your PR and put it into its own helper method. Depending on which PR gets merged first, the other one should be adjusted according, so the code doesn't end up duplicated.

java-checks/src/main/java/org/sonar/java/checks/TextBlockTabsAndSpacesCheck.java
sebastian-hungerecker-sonarsource: I thought about that, but I don't want to create multiple issues for one string literal. What I'd really want is to create a single issue that has the first problematic line as its primary and the others as secondary locations. In fact I wouldn't even want to mark the whole lines, but merely the indentation. Sadly, we don't have fitting a `reportIssue` overload for any of that.

So I'll leave it as-is for now and create a ticket to improve our reporting capabilities here.

margarita-nedzelska-sonarsource (approved):
LGTM 👍 Good job!

PR #3399, opened 2021-01-19
SONARJAVA-3658 Fix TextBlocks in SE engine, fix issue location between Text Blocks

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
After testing on different input, I've found out the Text_Blocks are not fully supported. So I ptovided a small fix for SE and for Locations reporting between 2 trees, when the last one is Text Block (it requires to calculate the end of the tree)

java-frontend/src/main/java/org/sonar/java/AnalyzerMessage.java
```diff
@@ -139,29 +142,48 @@ public boolean isEmpty() {
       lastSyntaxToken.line(),
       lastSyntaxToken.column() + lastSyntaxToken.text().length()
     );
+    checkLocation(firstSyntaxToken, location);
+    return location;
+  }
+
+  private static AnalyzerMessage.TextSpan textSpanEndsWithTextBlock(SyntaxToken firstSyntaxToken, SyntaxToken lastSyntaxToken) {
```
m-g-sonar, line 149: what about calling this method `textSpanBetween(SyntaxToken firstSyntaxToken, LiteralTree textBlockTree)` to cover this case explicitly instead?

java-checks-test-sources/src/main/files/non-compiling/checks/SimpleStringLiteralForSingleLineStringsCheck.java
```diff
@@ -2,7 +2,7 @@
 
 public class SimpleStringLiteralForSingleLineStringsCheck {
   
-  public void str() { // Noncompliant@+1 [sc=23;endColumn=42]]{{Use simple literal for a single-line string.}}
+  public void str() { // Noncompliant@+1 [[sc=23;endColumn=43]]{{Use simple literal for a single-line string.}}
```
m-g-sonar, line 5: wow, nice finding and undetected issue in precise location testing... 👍🏻 
worth a ticket to improve the verifier, don't you think?

margarita-nedzelska-sonarsource: Was surprised too. Yep, we should definitely do sth with verify to detect such typos)

m-g-sonar:
LGTM!

PR #3401, opened 2021-01-20
SONARJAVA-3670 Rule S6126: String multiline concatenation can be replaced with a Text block

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
I was thinking about the way to implement this rule in a best way. I checked the same rule in Intellij, and they suggest replacing in case there will be more than 2 lines in the end. So I applied the same limitation

java-checks-test-sources/src/main/java/checks/StringConcatToTextBlockCheck.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final int SOME_INT = 5;
+  
+  public static void main(String[] args) {
+
+    String s = "<html>\n" + // Noncompliant [[sc=16;ec=25]] {{Replace this String concatenation with Text block.}}
+               "    <body>\n" +
+               "        <tag>\n" +
+               "        </tag>\n" +
+               "    </body>\n" +
+               "</html>";
```
alban-auzeill, line 14: What's a good example of what we would like to find with this rule:

- a list of string literal concatenation
- each string literal end with \n (it is optional for the last one)

The current rule implementation is super simple, that's nice.
Unfortunately, splitting on `"\\\\n"` raises the following false-positive:
```
    System.out.println("\\n-" +
      "\\n-\\n");
```
Also, I don't like proposing a text block solution that is longer, like for:
```
    System.out.println("-\n-\n-\n-\n-" +
      "\n-\n");
```
Text bock is a longer solution:
```
    System.out.println(
      """
      -
      -
      -
      -
      -
      -
      """);
```
And some string with spaces before \n could be easier to read like:
```
    System.out.println(" \n" +
      "  \n   \n");
```
compare to:
```
    System.out.println(
      """
      \u0020
      \u0020\u0020
      \u0020\u0020\u0020
      """);
```

java-checks-test-sources/src/main/java/checks/StringConcatToTextBlockCheck.java
```diff
@@ -0,0 +1,66 @@
...
+                "        <tag>\n" +
+                i2 +
+                "        </tag>\n" +
+                "    </body>\n" +
+                "</html>";
+
+    String s6 = "<html>\n" + // Noncompliant, constant inside
+                "    <body>\n" +
+                "        <tag>\n" +
+                SOME_INT +
```
alban-auzeill, line 59: If some code depends on a constant, it's to manage the value of this constant at a central point in the application or in its dependencies. It prevents duplication of the value and prevents partial replacement when the value needs to be updated. Unfortunately, text blocks do not support string interpolation and force the user to duplicate the value of the constant expression, which is IMO a bad practice.

java-checks/src/main/java/org/sonar/java/checks/StringConcatToTextBlockCheck.java
```diff
@@ -0,0 +1,69 @@
...
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.JavaVersion;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S6126")
+public class StringConcatToTextBlockCheck extends IssuableSubscriptionVisitor implements JavaVersionAwareVisitor {
+
+  private static final String MESSAGE = "Replace this String concatenation with Text block.";
+  private static final Set<Tree> VISITED_NODES = new HashSet<>();
```
alban-auzeill, line 38: the visited nodes variable does not need to be `static`

java-checks/src/main/java/org/sonar/java/checks/StringConcatToTextBlockCheck.java
```diff
@@ -0,0 +1,69 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.PLUS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    VISITED_NODES.add(tree);
+    if (VISITED_NODES.contains(tree.parent())) {
+      return;
+    }
```
alban-auzeill, line 55: A nice trick to prevent reentrance.
I just wonder if we should fixe this case of issue duplication:
```
    System.out.println("<html>\n" + // Noncompliant, 1st issue
                "    <body>\n" +
                (
                "        <tag>\n" + // <== 2nd issue here
                "          xxx\n" +
                "        </tag>\n"
                ) +
                "    </body>\n" +
                "</html>");
```

java-checks/src/main/java/org/sonar/java/checks/StringConcatToTextBlockCheck.java
```diff
@@ -0,0 +1,69 @@
...
+      String value = (String) constant;
+      if (value.split("\\\\n").length > 2) {
+        reportIssue(tree, MESSAGE);
+      }
+    }
+  }
+
+  @Override
+  public void leaveNode(Tree tree) {
+    VISITED_NODES.remove(tree);
```
alban-auzeill, line 67: If there's an exception during the analysis of one file, the analysis continues on the next file, and methods like `leaveNode(Tree tree)` or `leaveFile(JavaFileScannerContext context)` are not called.
That why we use this pattern that ensure that everything is clear at the beginning of the analysis of one file:
```
  @Override
    visitedNodes.clear();
    super.setContext(context);
  }
```
Then, I don't think it's useful to remove the tree each time we leave the node, the global clean will be enough.

alban-auzeill (changes requested):

margarita-nedzelska-sonarsource:
I've updated the implementation. So we now gather all operands of plus. Then filter out those having something else except Literals. Then all strings not having '\n' are filtered. The strings containing '\n' are split into parts (if '\n' is somewhere in the middle), and small parts are not counted. So the final counter includes only lines having length > 7. And the count should be more than 2.

java-checks/src/main/java/org/sonar/java/checks/StringConcatToTextBlockCheck.java
margarita-nedzelska-sonarsource: I think yes, we should report once here

java-checks/src/main/java/org/sonar/java/checks/StringConcatToTextBlockCheck.java
```diff
@@ -0,0 +1,109 @@
...
+      .map(LiteralTree.class::cast)
+      .map(LiteralTree::value)
+      .filter(value -> value.contains("\\n"))
+      .flatMap(value -> Stream.of(value.split("\\\\n")))
+      .filter(str -> str.length() >= MINIMAL_LINE_LENGTH)
+      .count();
+
+    if (outputLines >= MINIMAL_NUMBER_OF_LINES) {
+      reportIssue(tree, MESSAGE);
+    }
```
alban-auzeill, line 80: I like the result of this implementation but I would like to propose another approach.

1. I want to split the lines of the full concatenated string content and not some fragment of it to have more consistent findings.
```
    String e = "abcdef\n" + "abcdef\n"; // currently non-compliant
    String f = "abcdef\nabc" + "def\n"; // currently compliant
```

2. I don't want to take into account the double quotes delimiter in the computation
```
    String h = "123456\n1234567\n" + "\n"; // currently non-compliant, it's <<"123456>> and <<1234567>>
    String g = "1234567\n123456\n" + "\n"; // currently compliant, it's <<"1234567>> and <<123456>>
```

3. It should be easy to ignore when it's not a new line `"\n"` but two distinct characters `"\\n"`
 
4. it should be possible to get rid of `Set<Tree> visitedNodes` by looking at the parent node kind.

So here is my proposal:
```
  public static final int MINIMAL_CONTENT_LENGTH = 19;
  public static final int MINIMAL_NUMBER_OF_NEW_LINES = 2;
  // ...
  @Override
    if (isChildOfPlusOperator(tree)) {
      return;
    }
    StringBuilder concatenatedContent = new StringBuilder();
    if (concatStringLiterals(concatenatedContent, tree)) {
      String content = concatenatedContent.toString();
      if (content.length() >= MINIMAL_CONTENT_LENGTH &&
        !content.contains("\\\\n") &&
        StringUtils.countMatches(content, "\\n") >= MINIMAL_NUMBER_OF_NEW_LINES) {
        reportIssue(tree, MESSAGE);
      }
    }
  }

  private static boolean isChildOfPlusOperator(Tree child) {
    Tree parent = child.parent();
    while (parent.is(Tree.Kind.PARENTHESIZED_EXPRESSION)) {
      parent = parent.parent();
    }
    return parent.is(Tree.Kind.PLUS);
  }

  private static boolean concatStringLiterals(StringBuilder concatenatedContent, Tree tree) {
    if (tree.is(Tree.Kind.PLUS)) {
      BinaryExpressionTree binaryExpression = (BinaryExpressionTree) tree;
      return concatStringLiterals(concatenatedContent, ExpressionUtils.skipParentheses(binaryExpression.leftOperand())) &&
             concatStringLiterals(concatenatedContent, ExpressionUtils.skipParentheses(binaryExpression.rightOperand()));
    } else if (tree instanceof LiteralTree) {
      String treeValue = ((LiteralTree) tree).value();
      if (tree.is(Tree.Kind.STRING_LITERAL)) {
        treeValue = LiteralUtils.trimQuotes(treeValue);
      }
      concatenatedContent.append(treeValue);
      return true;
    } else {
      return false;
    }
  }
```

WDYT?

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/StringConcatToTextBlockCheck.java
margarita-nedzelska-sonarsource: I liked the idea of concatStringLiterals method. I changed it a little.
As for isChildOfPlusOperator() method. It's kinda questionable what's better, caching visited nodes, or calculating. I'd rather leave caching here.

java-checks/src/main/java/org/sonar/java/checks/StringConcatToTextBlockCheck.java
```diff
@@ -19,29 +19,31 @@
...
 
 @Rule(key = "S6126")
 public class StringConcatToTextBlockCheck extends IssuableSubscriptionVisitor implements JavaVersionAwareVisitor {
 
   private static final String MESSAGE = "Replace this String concatenation with Text block.";
-  public static final int MINIMAL_LINE_LENGTH = 7;
+  public static final int MINIMAL_CONTENT_LENGTH = 19;
   public static final int MINIMAL_NUMBER_OF_LINES = 2;
+  // matches '\n' characters, but skips '\\n'
+  public static final Pattern EOL = Pattern.compile("(?<!\\\\)\\\\n");
```
alban-auzeill, line 46: :smile: I didn't dare to propose regular expression lookbehind previously because I was unsure that tmembers of our java bubble team would be comfortable with it. But with a comment, it seems easy to understand :+1:

alban-auzeill (approved):

PR #3404, opened 2021-01-21
SONARJAVA-3671 Improve rule S1192 to Support Text blocks

margarita-nedzelska-sonarsource:


java-checks-test-sources/src/main/java/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -36,22 +35,36 @@ default void foo() {
   private static final String REPORT_WITHOUT_THRESHOLD = "blabla";
 
   void test() {
-    System.out.println("constant"); // Noncompliant [[secondary=40]] {{Use already-defined constant 'A' instead of duplicating its value here.}}
+    System.out.println("constant"); // Noncompliant [[secondary=39]] {{Use already-defined constant 'A' instead of duplicating its value here.}}
```
sebastian-hungerecker-sonarsource, line 38: We can use relative line numbers for secondaries now (`secondary=+1`), so we should do that whenever updating a secondary location, so it's the last time we have to update it :wink: 

The same applies to the secondary locations in the other test files.

sebastian-hungerecker-sonarsource (changes requested):
See comments. Looks good other than that.

java-frontend/src/main/java/org/sonar/java/model/LiteralUtils.java
```diff
@@ -148,4 +150,28 @@ public static boolean isOne(ExpressionTree tree) {
   public static boolean isNegOne(ExpressionTree tree) {
     return tree.is(Kind.UNARY_MINUS) && isOne(((UnaryExpressionTree) tree).expression());
   }
+  
+  public static String getAsStringValue(LiteralTree tree) {
+    if (!tree.is(Kind.TEXT_BLOCK)) {
+      return trimQuotes(tree.value());
+    }
+    String[] lines = trimQuotes(tree.value()).split("\r?\n");
+    int indent = Arrays.stream(lines).skip(1).mapToInt(LiteralUtils::getIndentation)
```
sebastian-hungerecker-sonarsource, line 159: I missed this before, but empty lines (or lines containing only white space) are ignored for the purposes of determining indentation. So if all non-empty lines are indented by 4 spaces and all empty lines contain no spaces (which will be common since many editors/IDEs remove trailing spaces), the indentation should be 4, not 0.

java-frontend/src/main/java/org/sonar/java/model/LiteralUtils.java
```diff
@@ -148,4 +150,28 @@ public static boolean isOne(ExpressionTree tree) {
...
+    return Arrays.stream(lines)
+      .skip(1)
+      .map(s -> s.substring(indent))
+      .map(s -> s.replaceAll("[\t ]*$", ""))
+      .collect(Collectors.joining("\n"));
+  }
+  
+  private static int getIndentation(String str) {
+    for (int i = 0; i < str.length(); ++i) {
+      if (str.charAt(i) != ' ' && str.charAt(i) != '\t' ) {
```
sebastian-hungerecker-sonarsource, line 171: Another thing I missed the first time (and in my own rule implementation as well): Form feeds (`\f`) are also considered as space characters - both for the purposes of counting indentation and stripping trailing spaces. That isn't a really a big deal because how would a form feed even end up in a Java file? But if you're going to reopen this issue anyway, you might as well add a `\f` here and on the `replaceAll` call.

That's it though: spaces, tabs and form feeds. Unicode spaces count as regular characters - so good luck figuring out why your indentation isn't stripped if one of your indentation spaces happens to be a Unicode space.

sebastian-hungerecker-sonarsource (changes requested):
Sorry, I missed two issues before.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3405, opened 2021-01-22
SONARJAVA-3662 Improve rule S2142 to check methods called inside catch block

margarita-nedzelska-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
@@ -64,9 +64,6 @@ public class InterruptedExceptionCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public void visitNode(Tree tree) {
-    if(!hasSemantic()) {
-      return;
-    }
     TryStatementTree tryStatementTree = (TryStatementTree) tree;
     withinInterruptingFinally.addFirst(isFinallyInterrupting(tryStatementTree.finallyBlock()));
     for (CatchTree catchTree : tryStatementTree.catches()) {
@@ -101,9 +98,6 @@ public class InterruptedExceptionCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public void leaveNode(Tree tree) {
-    if (!hasSemantic()) {
-      return;
-    }
     withinInterruptingFinally.removeFirst();
   }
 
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
```diff
@@ -131,16 +137,31 @@ public BlockVisitor(Symbol catchedException) {
...
+          block.accept(this);
+        }
+      }
+      depth--;
       super.visitMethodInvocation(tree);
     }
 
     @Override
     public void visitThrowStatement(ThrowStatementTree tree) {
+      if (threadInterrupted || depth >= 3) {
```
sebastian-hungerecker-sonarsource, line 162: `depth` should be compared against `MAX_DEPTH`, not a hard coded `3`, but I don't think it should be compared here at all (see other comment).

java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
```diff
@@ -131,16 +137,31 @@ public BlockVisitor(Symbol catchedException) {
 
     @Override
     public void visitMethodInvocation(MethodInvocationTree tree) {
-      Symbol symbol = tree.symbol();
-      if (!symbol.isUnknown() && symbol.owner().type().isSubtypeOf("java.lang.Thread") && "interrupt".equals(symbol.name()) && tree.arguments().isEmpty()) {
+      if (threadInterrupted || depth >= MAX_DEPTH) {
```
sebastian-hungerecker-sonarsource, line 140: Right now, when the maximum depth is reached, we stop looking at any method calls (including interrupt) and at thrown exceptions. This means that if we're at depth `MAX_DEPTH - 1` and see a call to another method, we will then visit that method, but won't look at anything inside that method. So visiting that method would be pointless.

What I think should be happening is that we stop recursing into new methods when the maximum depth is reached, but we still look for `interrupt` calls and `throw`s in the method that reached the final depth. For that purpose, this if statement and the following `depth++` should be moved to after the `INTERRUPT_MATCHERS` if-statement and the `depth` check in `visitThrowStatement` should be removed.

sebastian-hungerecker-sonarsource (changes requested):
See comments. Looks good otherwise.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3407, opened 2021-01-25
SONARJAVA-3659 S2755 should not raise an issue when "EntityResolver" is customized

alban-auzeill:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3408, opened 2021-01-26
SONARJAVA-3114 Log issues related to unresolved types

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -325,4 +336,22 @@ public InputComponent project() {
...
+
+  public void logUndefinedTypes() {
+    if (!undefinedTypes.isEmpty()) {
+      javaClasspath.logSuspiciousEmptyLibraries();
+      javaTestClasspath.logSuspiciousEmptyLibraries();
+
+      LOG.debug(String.format("Unresolved imports/types have been detected during analysis. Logging the %d first:", LOGGED_UNDEFINED_TYPES));
+      LOG.debug(undefinedTypes
+        .stream()
+        .sorted()
```
margarita-nedzelska-sonarsource, line 352: As far as I understood `.sorted()` will sort strings in alphabetical order and `limit()` will show only first N.
In current implementation you will get first N elements after sorting them, so original order will not make sense and you probably don't need LinkedHashSet. 
If you want to get first N elements in the appearing order and then sort them, you need to swap calls  `limit` and `sorted`

java-frontend/src/test/java/org/sonar/java/classpath/ClasspathForTestTest.java
```diff
@@ -56,6 +65,34 @@ void no_interaction_with_FileSystem_at_initialization() {
...
+
+    String warning = "Dependencies/libraries were not provided for analysis of test files. The 'sonar.java.test.libraries' property is empty. "
+      + "Verify your configuration, as you might end up with less precise results.";
+    assertThat(logTester.logs(LoggerLevel.WARN))
+      .hasSize(1)
+      .contains(warning);
+  }
+
+  @Test
+  void no_warning_for_missing_bytecode_when_libraries_empty_and_have_jno_ava_sources() {
```
margarita-nedzelska-sonarsource, line 85:
```suggestion
  void no_warning_for_missing_bytecode_when_libraries_empty_and_have_no_java_sources() {
```

java-frontend/src/test/java/org/sonar/java/classpath/ClasspathForMainTest.java
```diff
@@ -92,14 +91,32 @@ void when_property_not_defined_project_classpath_null_getElements_should_be_empt
...
+    String warning = "Dependencies/libraries were not provided for analysis of source files. The 'sonar.java.libraries' property is empty. "
+      + "Verify your configuration, as you might end up with less precise results.";
+    verify(analysisWarnings).addUnique(warning);
+    assertThat(logTester.logs(LoggerLevel.WARN))
+      .hasSize(1)
+      .contains(warning);
+  }
+
+  @Test
+  void no_warning_for_missing_bytecode_when_libraries_empty_and_have_no_ava_sources() {
```
margarita-nedzelska-sonarsource, line 111:
```suggestion
  void no_warning_for_missing_bytecode_when_libraries_empty_and_have_no_java_sources() {
```

java-frontend/src/main/java/org/sonar/java/model/JSema.java
```diff
@@ -135,4 +138,8 @@ ITypeBinding resolveType(String name) {
   public void cleanupEnvironment() {
     ASTUtils.cleanupEnvironment(ast);
   }
+
+  public Set<String> undefinedTypes() {
```
margarita-nedzelska-sonarsource, line 142: Maybe it's better to make field private and return LinkedHashSet here too, if you want to keep order

margarita-nedzelska-sonarsource:
I left some comments, overall looks good. Nice work!

java-frontend/src/main/java/org/sonar/java/model/JSema.java
margarita-nedzelska-sonarsource: It looks like this method is never used. Do we need it?

java-frontend/src/test/java/org/sonar/java/SonarComponentsTest.java
```diff
@@ -409,6 +418,93 @@ void jsp_classpath_should_include_plugin() throws Exception {
...
+      assertThat(debugLogs)
+        .hasSize(2)
+        .startsWith("Unresolved imports/types have been detected during analysis. Logging the 50 first:");
+      String list = debugLogs.get(1);
+
+      assertThat(list)
+        .endsWith("- ...")
+        .doesNotContain("- Y cannot be resolved to a type")
+        .doesNotContain("- Z cannot be resolved to a type");
+      for (int i = 0; i < 26; i++) {
```
margarita-nedzelska-sonarsource, line 463: It's just a matter of taste, so what do you think about this variant:

```
      IntStream.rangeClosed(0, 25).forEach(i -> {
          char typeName = (char) ('A' + i);
          assertThat(list).contains(String.format("- The import org.package%s cannot be resolved", formatter.format(i + 1)));
          if (((char) ('A' + i)) < 'Y') {
            assertThat(list).contains(String.format("- %c cannot be resolved to a type", typeName));
          }
        });
```

java-frontend/src/test/java/org/sonar/java/classpath/ClasspathForMainTest.java
```diff
@@ -92,14 +91,32 @@ void when_property_not_defined_project_classpath_null_getElements_should_be_empt
...
-      "you might end up with less precise results. Bytecode can be provided using sonar.java.libraries property.";
-    verify(analysisWarnings).addUnique(eq(warning));
+
+    javaClasspath.logSuspiciousEmptyLibraries();
+
+    String warning = "Dependencies/libraries were not provided for analysis of source files. The 'sonar.java.libraries' property is empty. "
+      + "Verify your configuration, as you might end up with less precise results.";
+    verify(analysisWarnings).addUnique(warning);
+    assertThat(logTester.logs(LoggerLevel.WARN))
+      .hasSize(1)
```
margarita-nedzelska-sonarsource, line 106: I think we can replace it with single call to `containsExactly()`

java-frontend/src/test/java/org/sonar/java/classpath/ClasspathForTestTest.java
```diff
@@ -56,6 +65,34 @@ void no_interaction_with_FileSystem_at_initialization() {
...
+    javaTestClasspath.init();
+    assertThat(javaTestClasspath.getFilesFromProperty(ClasspathProperties.SONAR_JAVA_TEST_LIBRARIES)).isEmpty();
+    assertThat(javaTestClasspath.hasJavaSources()).isTrue();
+
+    javaTestClasspath.logSuspiciousEmptyLibraries();
+
+    String warning = "Dependencies/libraries were not provided for analysis of test files. The 'sonar.java.test.libraries' property is empty. "
+      + "Verify your configuration, as you might end up with less precise results.";
+    assertThat(logTester.logs(LoggerLevel.WARN))
+      .hasSize(1)
```
margarita-nedzelska-sonarsource, line 80: the same as for ClasspathForMainTest

java-frontend/src/test/java/org/sonar/java/model/JParserTest.java
```diff
@@ -96,13 +97,22 @@ void should_throw_RecognitionException_in_case_of_lexical_error() {
   @Test
   void err() {
     try {
-      // ASTNode.METHOD_DECLARATION with flag ASTNode.MALFORMED
+      // ASTNode.METHOD_DECLARATION with flag ASTNode.MALFORMED: missing return type
       test("interface Foo { public foo(); // comment\n }");
       fail("exception expected");
```
margarita-nedzelska-sonarsource, line 102: I see, it's not from your code. But it's not the best way to assert exceptions and as you've already touched this test, you can replace it with `assertThrows()`

java-frontend/src/test/java/org/sonar/java/model/JParserTest.java
```diff
@@ -96,13 +97,22 @@ void should_throw_RecognitionException_in_case_of_lexical_error() {
...
     }
   }
 
+  @Test
+  void unknown_types_are_collected() {
+    // import org.foo missing, type Bar unknown
+    CompilationUnitTree cut = test("import org.foo.Bar;\n class Foo {\n void foo(Bar b) {}\n }\n");
+    assertThat(((CompilationUnitTreeImpl) cut).sema.undefinedTypes)
+      .hasSize(2)
+      .contains("The import org.foo cannot be resolved", "Bar cannot be resolved to a type");
```
margarita-nedzelska-sonarsource, line 113: Maybe `containsExactly` or `containsExactlyInAnyOrder` will fit here?

margarita-nedzelska-sonarsource (commented):
Overall looks good!
Just few minor comments

java-frontend/src/main/java/org/sonar/java/model/JSema.java
m-g-sonar: It is used by the `JavaAstScanner`, at the end of the analysis.

java-frontend/src/test/java/org/sonar/java/SonarComponentsTest.java
m-g-sonar: I definitely feel better letting the test without requiring to streams there, so I'm going to keep as it is for now! But thanks for the suggestion!

margarita-nedzelska-sonarsource (approved):

PR #3409, opened 2021-01-27
SONARJAVA-3672 S1213: Check ordering of class and instance variables

dorian-burihabwa-sonarsource:
* Check ordering of class and instance variables
* Move test file to test-sources
* Update ruling expected results


alban-auzeill (approved):
We should also update the RSPEC by replacing:
```
- Class and instance variables
- Constructors
- Methods
```
with
```
- Class variables
- Instance variables
- Constructors
- Methods
```

alban-auzeill (approved):

PR #3410, opened 2021-01-28
SONARJAVA-3680 Update rules metadata

m-g-sonar:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5663_java.html
```diff
@@ -1,39 +1,17 @@
-<div class="paragraph">
```
m-g-sonar, line 1: I have some doubts about this one. I feel that's the one generated from Github. What was the conclusion of the discussion in LT? Should we keep all these useless classes?

I also see these useless nodes when using `gh_generate` command, but I kind of remember that we said they should be gone. Am I using an outdated version of rule-api?

margarita-nedzelska-sonarsource: I think this question wasn't resolved yet, whether we need to keep additional decorations from asciidoctor or not.
As for this rule. I tested it locally and on peach and it looks pretty much the same as others.

On the other hand, to feel more confident we can keep the version from old-fasioned Jira, because we still don't know whether it will break anything else or not.

So I'd rather use the version generated directly from Jira at the moment to stay safe.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1210_java.html
```diff
@@ -1,6 +1,6 @@
 <p>According to the Java <code>Comparable.compareTo(T o)</code> documentation:</p>
 <blockquote>
-  <p>It is strongly recommended, but not strictly required that <code>(x.compareTo(y)==0) == (x.equals(y))</code>.</p>
+  <p>It is strongly recommended, but not strictly required that <code>(x.compareTo( y )==0) == (x.equals( y ))</code>.</p>
```
margarita-nedzelska-sonarsource, line 3: I am not really sure spaces are done right here. I'm more a person of `(x.compareTo(y) == 0) == (x.equals(y))`

margarita-nedzelska-sonarsource (approved):
Looks good,
Just 1 style comment.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1210_java.html
m-g-sonar: agreed, I'm reverting the the change.

PR #3412, opened 2021-02-01
SONARJAVA-3114 Make sure we log only once

m-g-sonar:


quentin-jaquier-sonarsource (approved):
LGTM.

The logs are much cleaner now, good job!

PR #3414, opened 2021-02-01
SONARJAVA-3674 Rule S5659: JWT should be signed and verified with strong cipher algorithms

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/security/JWTWithStrongCipherCheck.java
```diff
@@ -0,0 +1,166 @@
...
+      if (declaration instanceof VariableTree) {
+        ExpressionTree initializer = ((VariableTree) declaration).initializer();
+        return initializer instanceof MethodInvocationTree && isSigned((MethodInvocationTree) initializer);
+      }
+    }
+    // Can be signed anywhere (field, other file), we consider it as signed
+    return true;
+  }
+
+  private static boolean isLocalVariable(Symbol symbol) {
```
sebastian-hungerecker-sonarsource, line 162: This method already exists in the `JUtils` class.

sebastian-hungerecker-sonarsource (approved):
See comment. LGTM other than that.

PR #3415, opened 2021-02-01
SONARJAVA-2929 Rule S2053: Hashes should include an unpredictable salt

margarita-nedzelska-sonarsource:


quentin-jaquier-sonarsource (commented):
As discussed, here is the first part of the review, I let you have a look.

java-checks/src/main/java/org/sonar/java/checks/security/UnpredictableSaltCheck.java
```diff
@@ -0,0 +1,164 @@
...
+      ExpressionTree expression = ((MemberSelectExpressionTree) methodSelect).expression();
+      if (expression.is(Tree.Kind.IDENTIFIER)) {
+        return Optional.of(((IdentifierTree) expression).symbol());
+      }
+    }
+    return Optional.empty();
+  }
+
+  private static boolean isInitializedWithLiteral(IdentifierTree identifier) {
+    Tree declaration = identifier.symbol().declaration();
```
quentin-jaquier-sonarsource, line 139: It's a corner case, but you should check that the variable is not re-assigned, otherwise you could be subject to FP.
You can probably use `symbol.isFinal() || JUtils.isEffectivelyFinal(symbol)` to do this.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S2053_java.html
```diff
@@ -0,0 +1,43 @@
...
+<h2>Recommended Secure Coding Practices</h2>
+<ul>
+  <li> Use hashing functions generating their own salt or generate a long random salt of at least 32 bytes. </li>
+  <li> The salt is at least as long as the resulting hash value. </li>
+  <li> Provide the salt to a safe hashing function such as PBKDF2. </li>
+  <li> Save both the salt and the hashed value in the relevant database record; during future validation operations, the salt and hash can then be
+  retrieved from the database. The hash is recalculated with the stored salt and the value being validated, and the result compared to the stored
+  hash. </li>
+</ul>
+<h2>Noncompliant Code Example</h2>
```
quentin-jaquier-sonarsource, line 16: We should add an example with `PBEKeySpec` to show in the description that we also support this library.

java-checks-test-sources/src/main/java/checks/security/UnpredictableSaltCheck.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  public void testPBESpec(char[] chars, byte[] salt) throws NoSuchAlgorithmException {
+    PBEKeySpec spec = new PBEKeySpec(chars); // Noncompliant [[sc=23;ec=44]]
+    PBEKeySpec spec1 = new PBEKeySpec(chars, salt, 1); // Compliant 
+  }
+
+  public void testNonRandomPBESpec(char[] chars, String str) throws NoSuchAlgorithmException {
+    byte[] salt = "notrandom".getBytes();
+    PBEKeySpec spec = new PBEKeySpec(chars, salt, 1); // Noncompliant [[sc=23;ec=53]] {{Make this salt unpredictable.}}
```
quentin-jaquier-sonarsource, line 48: In the case the salt is predictable and comes from a variable, we might want to add a secondary location to clearly point why the salt is predictable.

quentin-jaquier-sonarsource (approved):
LGTM, well done.

PR #3419, opened 2021-02-08
SONARJAVA-3683 S4502: raise an issue when CSRF protection is disabled on a path

dorian-burihabwa-sonarsource:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3422, opened 2021-02-09
SONARJAVA-3677 Rule S5443: Using publicly writable directories is security-sensitive

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/helpers/ExpressionsHelper.java
```diff
@@ -235,4 +237,19 @@ public static boolean alwaysReturnSameValue(ExpressionTree expression) {
...
+      if (expression.is(Tree.Kind.IDENTIFIER)) {
+        return Optional.of(((IdentifierTree) expression).symbol());
+      }
+    }
+    return Optional.empty();
+  }
+
+  public static boolean isNotReassigned(Symbol symbol) {
+    return symbol.isFinal() || (symbol.isVariableSymbol() && JUtils.isEffectivelyFinal(((Symbol.VariableSymbol) symbol)));
+  }
```
quentin-jaquier-sonarsource, line 253: Good idea to extract this method. You can also extract it in `OSCommandsPathCheck.java`, using the same code.

java-checks/src/main/java/org/sonar/java/checks/security/PubliclyWritableDirectoriesCheck.java
```diff
@@ -0,0 +1,169 @@
...
+    "/dev/mqueue",
+    "/run/lock",
+    "/var/run/lock",
+    "/Library/Caches",
+    "/Users/Shared",
+    "/private/tmp",
+    "/private/var/tmp",
+    "\\Windows\\Temp",
+    "\\Temp",
+    "\\TMP");
```
quentin-jaquier-sonarsource, line 59: I think we should test every single one in the unit tests.

java-checks-test-sources/src/main/files/non-compiling/checks/security/PubliclyWritableDirectories.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.attribute.FileAttribute;
+import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
+import java.util.Map;
+import java.util.Scanner;
+import java.util.Set;
```
quentin-jaquier-sonarsource, line 15: You can remove the unused import, to keep the file clean.

java-checks/src/main/java/org/sonar/java/checks/security/PubliclyWritableDirectoriesCheck.java
```diff
@@ -0,0 +1,169 @@
...
+      .constructor()
+      .addParametersMatcher(STRING_TYPE)
+      .addParametersMatcher(STRING_TYPE, STRING_TYPE)
+      .addParametersMatcher(STRING_TYPE, "java.nio.charset.Charset")
+      .build(),
+    MethodMatchers.create()
+      .ofTypes("java.nio.file.Paths", "java.nio.file.Path")
+      .names("get", "of")
+      .withAnyParameters()
+      .build());
```
quentin-jaquier-sonarsource, line 75: You are not fully testing all the methods that this matcher can match.

java-checks/src/main/java/org/sonar/java/checks/security/PubliclyWritableDirectoriesCheck.java
```diff
@@ -0,0 +1,169 @@
...
+    "/Users/Shared",
+    "/private/tmp",
+    "/private/var/tmp",
+    "\\Windows\\Temp",
+    "\\Temp",
+    "\\TMP");
+
+  private static final Set<String> TMP_DIR_ENV = SetUtils.immutableSetOf("TMP", "TMPDIR");
+
+  private static final MethodMatchers CREATE_FILE_MATCHERS = MethodMatchers.or(
```
quentin-jaquier-sonarsource, line 63: This matcher can match both constructors and method invocations and is only used when we already know the kind. This means that we can already say that half of the matcher will always fail when you are using it. You should split it in two.

java-checks/src/main/java/org/sonar/java/checks/security/PubliclyWritableDirectoriesCheck.java
```diff
@@ -0,0 +1,169 @@
...
+      .filter(ExpressionsHelper::isNotReassigned)
+      .map(Symbol::declaration)
+      .filter(decl -> decl.is(Tree.Kind.VARIABLE))
+      .map(VariableTree.class::cast)
+      .map(VariableTree::initializer)
+      .map(ExpressionUtils::skipParentheses)
+      .filter(initializer -> initializer.is(Tree.Kind.METHOD_INVOCATION))
+      .map(MethodInvocationTree.class::cast)
+      .map(SYSTEM_GETENV::matches)
+      .orElse(false);
```
quentin-jaquier-sonarsource, line 166: 😄

margarita-nedzelska-sonarsource: I've tried to write it firstly using imperative way, and the code looked very complicated and had a lot of null checks. So I decided to apply declarative approach here.

As for me, it looks quite readable and understandable, but feel free to challenge this approach. Cause it might be just my Scala brain deformation 😅

quentin-jaquier-sonarsource: I also like this code... I probably have the same brain deformation though.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3424, opened 2021-02-09
SONARJAVA-3675 Rule S5332: Using clear-text protocols is security-sensitive

sebastian-hungerecker-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3426, opened 2021-02-09
SONARJAVA-3676 Rule S5689: Disclosing fingerprints from web application technologies is security-sensitive

margarita-nedzelska-sonarsource:


java-checks-test-sources/src/main/java/checks/security/DisclosingTechnologyFingerprints.java
```diff
@@ -0,0 +1,91 @@
...
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletResponseWrapper;
+import org.apache.wicket.protocol.http.BufferedWebResponse;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+public class DisclosingTechnologyFingerprints {
```
dorian-burihabwa-sonarsource, line 17: We should add a test for `org.rapidoid.http.Resp` even if it has to be in a non-compiling source file

dorian-burihabwa-sonarsource (changes requested):
We should add add a test for `org.rapidoid.http.Resp#header` even if not compiling. Otherwise this looks good to me :+1:

dorian-burihabwa-sonarsource (approved):
Thanks for adding the test. Looks good to me 👍🏿

PR #3427, opened 2021-02-10
SONARJAVA-3663 S2976 implementation moved to S5445

margarita-nedzelska-sonarsource:


quentin-jaquier-sonarsource (changes requested):
I think it makes sense to deprecate S2976 in this PR. Running ruleapi will not work, because the replacement rule is not yet merged. You can work around this by manually setting java in the "covered language" of the S5445.

In addition, this rule is now a "Blocker" (S2976 was "critical"). I feel that keeping the original severity for the new rule makes more sense.

margarita-nedzelska-sonarsource:
I asked AppSec about the severity, they're going to think about it next week, so asked to keep it as is for now.
We'll still be able to adjust changes before the release

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/Sonar_way_profile.json
```diff
@@ -380,6 +380,7 @@
     "S5411",
     "S5413",
     "S5443",
+    "S5445",
```
quentin-jaquier-sonarsource, line 383: I would expect to see S2976 removed from the SonarWay profile, is it expected to keep it?

margarita-nedzelska-sonarsource: I asked Eric, he said, it should be done automatically, once we deprecate it.
I don't understand exactly what is the process here. rule-api, didn't remove it, when I run scripts.

I went through this[ doc ](https://xtranet-sonarsource.atlassian.net/wiki/spaces/LANG/pages/665976/Coding+Rules+Guidelines) and it looks like we need to update RSPEC itself and remove it from SonarWay, but I am not absolutely sure what's done with script and what should be done manualy.

We can also wait till tomorrow to see if any jobs did some updates on the rule.

quentin-jaquier-sonarsource: Indeed, "sonartech" did some changes to the RSPEC this night, removing the default profile.

margarita-nedzelska-sonarsource: Updated PR :)

quentin-jaquier-sonarsource (approved):

PR #3430, opened 2021-02-11
SONARJAVA-3679 Rule S5693: Allowing requests with excessive content length is security-sensitive

quentin-jaquier-sonarsource:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3433, opened 2021-02-12
SONARJAVA-2929 Rule S2053: Fix FP with weak hash algorithms

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/helpers/ExpressionsHelper.java
```diff
@@ -221,6 +221,15 @@ private static boolean isAssignedToNonSerializable(ExpressionTree expression) {
     } else {
       return Stream.concat(Stream.of(initializer), assignedExpressionStream);
     }
+  }  
+  
+  public static Optional<ExpressionTree> initializedAndAssignedExpression(Symbol symbol) {
```
quentin-jaquier-sonarsource, line 226: I'm not sure this name is well describing what this method does. What about removing the `AndAssigned` part? Or `initializedAndNotAssignedExpression`?

In addition, maybe we can use `getSingleWriteUsage` from the same file, doing something similar?

margarita-nedzelska-sonarsource: Thanks, that's make sense to reuse it.

quentin-jaquier-sonarsource (approved):

PR #3435, opened 2021-02-15
SONARJAVA-3694 Improve rule S1612 to replace instanceof lambda with method reference

margarita-nedzelska-sonarsource:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3439, opened 2021-02-16
SONARJAVA-3678 FP in S5853 when map/flatMap is used

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/java/checks/AssertJConsecutiveAssertionCheck.java
```diff
@@ -13,10 +14,11 @@
   String myString = "42";
   String myOtherString = "24";
   List<String> myList = new ArrayList<>();
+  Optional<String> myOptional = Optional.of("abc");
```
dorian-burihabwa-sonarsource, line 17: Is the use of an `Optional` necessary here since we don't explicitly check for `Optional` in the rule?

quentin-jaquier-sonarsource: As discussed, the "map" and "flatMap" we want to test are defined in "AbstractOptionalAssert", we need an Optional to start such assertion.

PR #3441, opened 2021-02-16
SONARJAVA-3687: S5979 should not report on classes annotated with JUnit5's @Nested

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
```diff
@@ -73,16 +79,32 @@
   @Override
   public void visitNode(Tree tree) {
     ClassTree testClass = (ClassTree) tree;
+    if (visited.contains(testClass)) {
+      return;
+    }
+
+    if (isClassProperlyAnnotated(testClass)) {
+      List<ClassTree> children = getChildrenCoveredByAnnotations(testClass);
+      visited.addAll(children);
```
quentin-jaquier-sonarsource, line 88: You can probably return here, and remove the redundant `isClassProperlyAnnotated` computation in `mocksAreProperlyInitialized`.

quentin-jaquier-sonarsource (changes requested):
I'm wondering if the initialization via a Rule or during the setup in a parent works for the Nested class aswell. Did you check that?

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
```diff
@@ -65,6 +69,8 @@
 
   private static final String MESSAGE = "Initialize mocks before using them.";
 
+  private Set<ClassTree> visited = new HashSet<>();
```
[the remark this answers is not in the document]
dorian-burihabwa-sonarsource: Renamed

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
dorian-burihabwa-sonarsource: Modified

dorian-burihabwa-sonarsource:
> I'm wondering if the initialization via a Rule or during the setup in a parent works for the Nested class aswell. Did you check that?

Experimenting with it, and looking up the different constructs, we can discard the @Rule solution as it's a JUnit4 construct and is not expected to be mixed with JUnit5's @Nested. It makes sense to cover `ExtendWith` as it comes from JUnit5. But `openMocks` and `initMocks` used in setup methods cannot cover nested classes as they need to target and instance of the class they run in (eg `MockitoAnnotations.openMocks(this)` or `MockitoAnnotations.openMocks(this.myMockedField)`) and thus cannot cover mocks in inner classes.

quentin-jaquier-sonarsource (approved):
LGTM.

Thanks for investigating what is or not possible.

PR #3444, opened 2021-02-16
SONARJAVA-3649 FP on S1157 (CaseInsensitiveComparisonCheck) when only one side is upper or lower case

sebastian-hungerecker-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM, nice refactoring!

PR #3445, opened 2021-02-17
Update tests for S3252 and S1942

quentin-jaquier-sonarsource:
During the Guava removal, 2 tests ended up not testing anything, since they `verifyNoIssues` despite the fact that the rule should still raise issues.

Only the tests are affected, the behavior of the rules are correct and unchanged.


margarita-nedzelska-sonarsource (approved):
LGTM

PR #3447, opened 2021-02-17
SONARJAVA-3679 Report issue with AnalyzerMessage

quentin-jaquier-sonarsource:
See [DefaultJavaFileScannerContext.java#L263](https://github.com/SonarSource/sonar-java/blob/master/java-frontend/src/main/java/org/sonar/java/model/DefaultJavaFileScannerContext.java#L263).

I believe this is a protection, to not store references to trees across different files analysis.
In this case, I don't think it was relevant: the issue was directly reported during the analysis of the file (and not at the end of the analysis), but we can report with AnalyzerMessage anyway since it does not change anything at the end.


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3451, opened 2021-02-18
SONARJAVA-3688 FP on S5860(UnusedGroupNamesCheck) for name referenced by dollar curly braces

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/regex/UnusedGroupNamesCheck.java
```diff
@@ -47,13 +48,27 @@
...
+      .build(),
+    MethodMatchers.create()
+      .ofTypes(JAVA_UTIL_REGEX_MATCHER)
+      .names("appendReplacement")
+      .addParametersMatcher(MethodMatchers.ANY, MethodMatchers.ANY)
+      .build(),
+    MethodMatchers.create()
+      .ofTypes(JAVA_UTIL_REGEX_MATCHER)
+      .names("replaceAll", "replaceFirst")
+      .addParametersMatcher(MethodMatchers.ANY)
```
quentin-jaquier-sonarsource, line 70: I was wondering: is there any specific reason to chose `ANY` instead of `java.lang.String`?

quentin-jaquier-sonarsource (approved):

java-checks/src/main/java/org/sonar/java/checks/regex/UnusedGroupNamesCheck.java
alban-auzeill: Because java 9 introduces more method signatures to the `Matcher`:
```
replaceAll(Function<MatchResult, String> replacer)
replaceFirst(Function<MatchResult, String> replacer)
```
And finding all replaceFirst, replaceAll and appendReplacement methods is important to not have false-positive.
Not being able to convert the argument to a string constant is not a problem in the rule implementation.
And I didn't add a unit test about such false-positive because we don't support java 9 in our compiled unit tests. And the rule requires a project that compiles to have the semantic about replaceFirst, replaceAll.

quentin-jaquier-sonarsource: Make sense, thanks for the explanation. 👍

PR #3452, opened 2021-02-18
SONARJAVA-3698 Extract Check Verifier from "java-frontend" module int…

margarita-nedzelska-sonarsource:
…o testkit


java-checks-testkit/src/test/java/org/sonar/java/checks/verifier/TestUtils.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import org.sonar.api.batch.fs.InputFile;
+import org.sonar.api.batch.fs.internal.TestInputFileBuilder;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class TestUtils {
```
alban-auzeill, line 31: Unfortunately, after adding this `TestUtils` class we will have 4 identical `TestUtils` classes duplicated:
```
java-frontend/src/test/java/org/sonar/java/TestUtils.java
java-checks-testkit/src/test/java/org/sonar/java/checks/verifier/TestUtils.java
java-symbolic-execution/src/test/java/org/sonar/java/se/utils/TestUtils.java
java-checks/src/test/java/org/sonar/java/CheckTestUtils.java
```
Every module depends on `java-frontend` but it's not possible to use the test code of this module in others.
For `java-checks-testkit` it's a different story, we can move `TestUtils` from `src/test/java` to `src/main/java` so we can use it in `java-symbolic-execution` and `java-checks`.
Moving `TestUtils`, require `junit-jupiter` dependency just for one line:
```
assertTrue(file.exists(), "Path '" + path + "' should exist.");
```
We could replace this line with:
```
    if (!file.exists()) {
      throw new IllegalStateException("Path '" + path + "' should exist.");
    }
```
And once we have `java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/TestUtils.java` we can remove 2 of the `TestUtils` classes:
```
java-checks/src/test/java/org/sonar/java/CheckTestUtils.java
java-symbolic-execution/src/test/java/org/sonar/java/se/utils/TestUtils.java
java-symbolic-execution/src/test/java/org/sonar/java/se/utils/TestUtilsTest.java
```
By renaming in all files:
`org.sonar.java.se.utils.TestUtils` => `org.sonar.java.checks.verifier.TestUtils`
`org.sonar.java.CheckTestUtils` => `org.sonar.java.checks.verifier.TestUtils`
`CheckTestUtils.testSourcesPath(` => `TestUtils.testSourcesPath(`
`CheckTestUtils.inputFile(` => `TestUtils.inputFile(`

java-checks-testkit/src/test/java/org/sonar/java/checks/verifier/TestUtils.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  public static int[] computeLineEndOffsets(int[] lineStartOffsets, int lastValidOffset) {
+    int[] lineEndOffsets = new int[lineStartOffsets.length];
+    for (int i = 0; i < lineStartOffsets.length - 1; i++) {
+      lineEndOffsets[i] = lineStartOffsets[i + 1] - 1;
+    }
+    lineEndOffsets[lineEndOffsets.length - 1] = lastValidOffset - 1;
+    return lineEndOffsets;
+  }
```
alban-auzeill, line 64: `computeLineEndOffsets` seems to be useful only in `java-frontend/src/test/java/org/sonar/java/TestUtils.java` but and be removed from here.

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #3455, opened 2021-02-24
SONARJAVA-3603 FP on S4276 when Function is using "compose" or "andThen" methods

margarita-nedzelska-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3456, opened 2021-02-24
SONARJAVA-3669 S2325 should not raise on empty methods

margarita-nedzelska-sonarsource:


java-checks/src/test/java/org/sonar/java/checks/StaticMethodCheckTest.java
```diff
@@ -22,14 +22,25 @@
...
       .withCheck(new StaticMethodCheck())
       .verifyIssues();
   }
 
+  @Test
+  void test_non_compiling() {
+    JavaCheckVerifier.newVerifier()
+      .onFile(nonCompilingTestSourcesPath("checks/StaticMethodCheck.java"))
+      .withCheck(new StaticMethodCheck())
+      .verifyNoIssues();
```
quentin-jaquier-sonarsource, line 43: "no issues", really? There is a `Noncompliant` comment in this file, it should not pass!

quentin-jaquier-sonarsource (changes requested):
LLGTM

java-checks/src/test/java/org/sonar/java/checks/StaticMethodCheckTest.java
margarita-nedzelska-sonarsource: That was a leftover

quentin-jaquier-sonarsource (approved):

PR #3457, opened 2021-02-24
SONARJAVA-3558 Issue filter should extends its filter to IDE-specific suppressed warnings

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/filters/SuppressWarningFilter.java
```diff
@@ -60,8 +61,20 @@
...
+      .put("boxing", SetUtils.immutableSetOf("java:S2153", "java:S5411"))
+      .put("hiding", Collections.singleton("java:S4977"))
+      .put("javadoc", Collections.singleton("java:S1176"))
+      .put("null", Collections.singleton("java:S2259"))
+      .put("resource", SetUtils.immutableSetOf("java:S2093", "java:S2095"))
+      .put("serial", Collections.singleton("java:S2057"))
+      .put("static-access", SetUtils.immutableSetOf("java:S2696", "java:S2209"))
+      .put("static-method", Collections.singleton("java:S2325"))
+      .put("sync-override", Collections.singleton("java:S3551"))
+      .put("unused", SetUtils.immutableSetOf("java:S1481","java:S1065","java:S1854","java:S1068"
```
margarita-nedzelska-sonarsource, line 76: Just for consistency, it's better to keep spaces after comas and keep comas  at the end of the previous line.

java-checks/src/test/java/org/sonar/java/filters/SuppressWarningFilterTest.java
```diff
@@ -69,4 +86,44 @@ void verify() {
     );
   }
 
+  @Test
+  void verify_2() {
+    FilterVerifier.verify("src/test/files/filters/SuppressWarningFilter_2.java", new SuppressWarningFilter(),
```
margarita-nedzelska-sonarsource, line 91: Do you think it makes sense to move these files to java-checks-test-sources module?

margarita-nedzelska-sonarsource (commented):
Overall Looks good. Just 2 small comments

java-checks/src/test/java/org/sonar/java/filters/SuppressWarningFilterTest.java
quentin-jaquier-sonarsource: I tried the same way we do for unit tests, but it's not working. I'm not sure what is the exact issue, but at first glance, it doesn't seem trivial to fix.

It would be nice to have, but I don't think it's a must, I think we can let it as it is. Is it fine for you?

margarita-nedzelska-sonarsource: Fine by me

margarita-nedzelska-sonarsource (approved):

PR #3459, opened 2021-02-25
SONARJAVA-3636 Improve secondary message for regex rules when issues are reported across different string literals

alban-auzeill:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3460, opened 2021-02-25
SONARJAVA-2745 FN on S2142: no issue raised when catching the generic Exception

quentin-jaquier-sonarsource:
Small precision: 
I decided to only support "InterruptedException" and not "ThreadDeath" because it is an (unchecked) error, I don't expect it to catch anything more than what is currently detected.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
@@ -80,13 +80,14 @@ public class InterruptedExceptionCheck extends IssuableSubscriptionVisitor {
     withinInterruptingFinally.addFirst(isFinallyInterrupting(tryStatementTree.finallyBlock()));
     for (CatchTree catchTree : tryStatementTree.catches()) {
       VariableTree catchParameter = catchTree.parameter();
-      Optional<Type> interruptType = getIfCatchMatch(catchParameter, INTERRUPTING_TYPE_PREDICATE);
+      List<Type> caughtTypes = getCaughtTypes(catchParameter);
+      Optional<Type> interruptType = caughtTypes.stream().filter(INTERRUPTING_TYPE_PREDICATE).findFirst();
       if (interruptType.isPresent()) {
         if (handleInCorrectlyInterruption(catchTree)) {
           reportIssue(catchParameter, String.format(MESSAGE, interruptType.get().name()));
         }
         return;
-      } else if (getIfCatchMatch(catchParameter, GENERIC_EXCEPTION_PREDICATE).isPresent()) {
+      } else if (caughtTypes.stream().anyMatch(GENERIC_EXCEPTION_PREDICATE)) {
         reportIfThrowInterruptInBlock(tryStatementTree.block(), catchTree);
         return;
       }
@@ -113,17 +114,13 @@ public class InterruptedExceptionCheck extends IssuableSubscriptionVisitor {
     return !blockVisitor.threadInterrupted && !isWithinInterruptingFinally();
   }
 
-  private static Optional<Type> getIfCatchMatch(VariableTree parameter, Predicate<Type> predicate) {
+  private static List<Type> getCaughtTypes(VariableTree parameter) {
     if (parameter.type().is(Tree.Kind.UNION_TYPE)) {
       return ((UnionTypeTree) parameter.type()).typeAlternatives().stream()
         .map(TypeTree::symbolType)
-        .filter(predicate)
-        .findFirst();
+        .collect(Collectors.toList());
     }
-    return Optional.of(parameter)
-      .map(VariableTree::symbol)
-      .map(Symbol::type)
-      .filter(predicate);
+    return Collections.singletonList(parameter.symbol().type());
   }
 
   private boolean isWithinInterruptingFinally() {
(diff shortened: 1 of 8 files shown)


java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
```diff
@@ -68,31 +77,64 @@ public void leaveFile(JavaFileScannerContext context) {
...
-        if(!blockVisitor.threadInterrupted && !isWithinInterruptingFinally()) {
-          reportIssue(catchTree.parameter(), "Either re-interrupt this method" +
-            " or rethrow the \""+interruptType.get().name()+"\".");
+      VariableTree catchParameter = catchTree.parameter();
+      Optional<Type> interruptType = getIfCatchMatch(catchParameter, INTERRUPTING_TYPE_PREDICATE);
+      if (interruptType.isPresent()) {
+        interruptCaught = true;
+        if (handleInCorrectlyInterruption(catchTree)) {
+          reportIssue(catchParameter, String.format(MESSAGE, interruptType.get().name()));
         }
```
margarita-nedzelska-sonarsource, line 91: It seems to me we should stop iterating, once we found appropriate Exception handler (InterruptedException, Exception or Throwable) because only one catch block will be executed once exception is caught.

margarita-nedzelska-sonarsource (changes requested):
I feel we need to support these cases (currently we report incorrectly):

```
    try {
      throwsInterruptedException();
    } catch (Exception e) { // Noncompliant
      LOGGER.log(Level.WARN, "Interrupted!", e);
    } catch (InterruptedException e) {
      LOGGER.log(Level.WARN, "Interrupted!", e);
    }
    
    try {
      throwsInterruptedException();
    } catch (Exception e) { // Noncompliant
      LOGGER.log(Level.WARN, "Interrupted!", e);
    } catch (InterruptedException e) {
      Thread.currentThread().interrupt();
    }
    
    try {
      throwsInterruptedException();
    } catch (Exception e) { // Compliant
      Thread.currentThread().interrupt();
    } catch (InterruptedException e) {
      LOGGER.log(Level.WARN, "Interrupted!", e);
    }
```

In the current implementation we first check if InterruptedException was caught and then the general one. Which is actually not true, because if Exception catch block comes first, it will catch the InterruptedException. 

The problem that catch block are in a wrong order should be caught by a different issue.

java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
quentin-jaquier-sonarsource: It makes a lot of sense, doing this will clean the flow of the logic, and consistently support all the examples.

java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
```diff
@@ -68,31 +76,51 @@ public void leaveFile(JavaFileScannerContext context) {
...
+    if (!invocationInterrupting.isEmpty() && handleInCorrectlyInterruption(catchTree)) {
+      reportIssue(catchTree.parameter(), String.format(MESSAGE, "InterruptedException"),
+        invocationInterrupting.stream()
+          .map(t -> new JavaFileScannerContext.Location("Method invocation throwing InterruptedException.", t))
+          .collect(Collectors.toList()),
+        null);
+    }
+  }
+
+  private boolean handleInCorrectlyInterruption(CatchTree catchTree) {
```
margarita-nedzelska-sonarsource, line 111: maybe rename this method to `wasNotInterrupted` or invert condition and name it `wasInterrupted`

java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
```diff
@@ -68,31 +76,51 @@ public void leaveFile(JavaFileScannerContext context) {
...
-  private static Optional<Type> findInterruptingType(VariableTree parameter) {
+  private void reportIfThrowInterruptInBlock(BlockTree blockTree, CatchTree catchTree) {
+    MethodTreeUtils.MethodInvocationCollector collector = new MethodTreeUtils.MethodInvocationCollector(InterruptedExceptionCheck::throwInterruptedException);
+    blockTree.accept(collector);
+    List<Tree> invocationInterrupting = collector.getInvocationTree();
+
+    if (!invocationInterrupting.isEmpty() && handleInCorrectlyInterruption(catchTree)) {
+      reportIssue(catchTree.parameter(), String.format(MESSAGE, "InterruptedException"),
+        invocationInterrupting.stream()
+          .map(t -> new JavaFileScannerContext.Location("Method invocation throwing InterruptedException.", t))
```
margarita-nedzelska-sonarsource, line 105: It looks like rule description doesn't contain secondary locations. Do you think we should update rule description?

margarita-nedzelska-sonarsource (approved):
Looks good, just minor comments

java-checks/src/main/java/org/sonar/java/checks/InterruptedExceptionCheck.java
quentin-jaquier-sonarsource: Yes, even if it does not change the final product, it makes sense to keep the RSPEC up to date with what is in the implementation.

PR #3463, opened 2021-02-25
SONARJAVA-3635 S2384 should not raise an issue when mutable members in temporary variable are not stored

quentin-jaquier-sonarsource:


alban-auzeill (approved):
LGTM

PR #3464, opened 2021-02-25
SONARJAVA-3624 Regex FP/FN with Supplementary Multilingual Plane

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/helpers/RegexTreeHelper.java
```diff
@@ -30,7 +30,7 @@
 import org.sonar.java.regex.ast.AutomatonState;
 import org.sonar.java.regex.ast.AutomatonState.TransitionType;
 import org.sonar.java.regex.ast.BoundaryTree;
-import org.sonar.java.regex.ast.CharacterTree;
+import org.sonar.java.regex.ast.PlainCharacterTree;
```
sebastian-hungerecker-sonarsource, line 33: Given the amount of changes like this, it might make more sense to rename `PlainCharacterTree` to `CharacterTree` (and the corresponding enum value to `CHARACTER`). The qualifier "plain" isn't really necessary anymore since there's no other type of character tree to distinguish it from. Plus, since previously "plain" meant "single code unit", it's no longer accurate anyway and was confusing nomenclature to begin with.

java-frontend/src/main/java/org/sonar/java/regex/RegexParser.java
```diff
@@ -820,14 +824,34 @@ private CharacterClassElementTree parseCharacterRange(CharacterTree startCharact
...
+      } else {
+        LOG.warn("Couldn't parse '{}{}', two high surrogate characters in a row. Please check your encoding.", c1, c2);
+      }
+    }
     return plainCharacter(character, character.getRange());
   }
 
   private PlainCharacterTree plainCharacter(JavaCharacter character, IndexRange range) {
-    return new PlainCharacterTree(source, range, character, activeFlags);
+    return new PlainCharacterTree(source, range, character.getCharacter(), character.isEscapeSequence(), activeFlags);
```
sebastian-hungerecker-sonarsource, line 851: I was worried that only checking for multi-unit characters in `plainCharacter(JavaCharacter)` and not `plainCharacter(JavaCharacter, IndexRange)` might lead to breakage, but apparently the latter is only ever called by the former. To me that means it's useless and should just be inlined into `plainCharacter(JavaCharacter)`, so no one accidentally calls it instead of `plainCharacter(JavaCharacter)` in the future.

java-frontend/src/main/java/org/sonar/java/regex/ast/RegexBaseVisitor.java
```diff
@@ -68,15 +68,7 @@ public void visitPlainCharacter(PlainCharacterTree tree) {
...
-  @Override
-  public void visitUnicodeCodePoint(UnicodeCodePointTree tree) {
-    visitCharacter(tree);
-  }
-
-  /**
-   * Override to visit both PlainCharacterTrees and UnicodeCodePointTrees
-   */
-  protected void visitCharacter(CharacterTree tree) {
+  protected void visitCharacter(PlainCharacterTree tree) {
```
sebastian-hungerecker-sonarsource, line 71: `visitPlainCharacter` should be removed as it's now redundant to have both.

sebastian-hungerecker-sonarsource (changes requested):
Good work. Some (hopefully minor) changes.

sebastian-hungerecker-sonarsource (approved):
LGTM :+1:

PR #3466, opened 2021-02-26
SONARJAVA-3706 FP in S2384, S2386: support any unmodifiable and immutable methods

alban-auzeill:


quentin-jaquier-sonarsource (approved):
LGTM, glad we ended up with something really clean!

PR #3468, opened 2021-03-01
SONARJAVA-3570 S5411: Relax rule when test is done in related branch

dorian-burihabwa-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
@@ -20,6 +20,8 @@
 package org.sonar.java.checks;
 
 import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import org.sonar.check.Rule;
 import org.sonar.plugins.java.api.JavaFileScanner;
@@ -127,15 +129,51 @@ public class BoxedBooleanExpressionsCheck extends BaseTreeVisitor implements Jav
   private static boolean isFirstUsageANullCheck(ExpressionTree boxedBoolean) {
     if (boxedBoolean.is(Kind.IDENTIFIER)) {
       IdentifierTree identifier = (IdentifierTree) boxedBoolean;
+      // Usages are not guaranteed to be ordered
       List<IdentifierTree> usages = identifier.symbol().usages();
-      if (!usages.isEmpty()) {
-        Tree parent = usages.get(0).parent();
-        if (parent.is(Kind.NOT_EQUAL_TO, Kind.EQUAL_TO) && isNullCheck((ExpressionTree) parent)) {
-          return true;
-        }
+      Tree firstUsage = usages.get(0).parent();
+      // Test if the first usage in our list is a null check
+      if (firstUsage.is(Kind.EQUAL_TO, Kind.NOT_EQUAL_TO) && isNullCheck((ExpressionTree) firstUsage)) {
+        return true;
       }
+      // Return false if the only usage is not a null check
+      if (usages.size() == 1) {
+        return false;
+      }
+      // Fetch the first null check in the usages list
+      Optional<ExpressionTree> firstNullCheck = getFirstNullCheck(usages);
+      if (!firstNullCheck.isPresent()) {
+        return false;
+      }
+      // Test if the first null check and the first usage are part of the same higher if structure
+      Optional<IfStatementTree> test = getParentConditionalBranch(firstNullCheck.get());
+      Optional<IfStatementTree> parentConditionalBranch = getParentConditionalBranch((ExpressionTree) firstUsage);
+      return test.equals(parentConditionalBranch);
     }
     return false;
+  }
+
+  private static Optional<ExpressionTree> getFirstNullCheck(List<IdentifierTree> usages) {
+    List<ExpressionTree> nullChecks = usages.stream()
+      .map(IdentifierTree::parent)
+      .filter(tree -> tree.is(Kind.EQUAL_TO, Kind.NOT_EQUAL_TO) && isNullCheck((ExpressionTree) tree))
+      .map(ExpressionTree.class::cast)
+      .collect(Collectors.toList());
+    if (nullChecks.isEmpty()) {
+      return Optional.empty();
+    }
+    return Optional.of(nullChecks.get(0));
+  }
+
+  private static Optional<IfStatementTree> getParentConditionalBranch(ExpressionTree tree) {
+    Tree parent = tree;
+    while (parent != null && !parent.is(Kind.IF_STATEMENT)) {
+      parent = parent.parent();
+    }
+    if (parent == null) {
+      return Optional.empty();
+    }
+    return Optional.of((IfStatementTree) parent);
   }
 
   @CheckForNull
--- a/java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
@@ -21,7 +21,6 @@ package org.sonar.java.checks;
 
 import java.util.List;
 import java.util.Optional;
-import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import org.sonar.check.Rule;
 import org.sonar.plugins.java.api.JavaFileScanner;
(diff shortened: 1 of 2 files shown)


margarita-nedzelska-sonarsource (changes requested):
Actually, I think we can improve this solution and make it less heavy

What I propose is every time we face null-check put their branches into one of two sets : safeBlocks or dangerousBlocks. For example:

```
    if (condition.is(Kind.EQUAL_TO) && isNullCheck(condition)) {
      dangerousBlocks.add(thenStatement);
      if (elseStatement != null) {
        safeBlocks.add(elseStatement);
      }
    } else if (condition.is(Kind.NOT_EQUAL_TO) && isNullCheck(condition)) {
      safeBlocks.add(thenStatement);
      if (elseStatement != null) {
        dangerousBlocks.add(elseStatement);
      }
    }
```

And then we can check whether our tested expression belongs to one of these blocks. This approach won't cover early returns, so for early returns we still need to check the first usage.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -127,17 +135,61 @@ private boolean isSafeBooleanExpression(ExpressionTree tree) {
...
+      .findFirst();
+  }
+
+  private static Optional<IfStatementTree> getParentConditionalBranch(ExpressionTree tree) {
+    if (ifStatementCache.containsKey(tree)) {
+      return ifStatementCache.get(tree);
+    }
+    Tree parent = tree;
+    while (parent != null && !parent.is(Kind.IF_STATEMENT)) {
+      parent = parent.parent();
```
margarita-nedzelska-sonarsource, line 186: the point of caching here is that you might have already calculated the 'IfStatement parent' for some parent of this tree hierarchy. So you don't need to go through the whole loop of parents, just find the first parent. whose closest IfStatement we already know.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -127,17 +135,61 @@ private boolean isSafeBooleanExpression(ExpressionTree tree) {
...
+      // Test if the first usage in our list is a null check
+      if (firstUsage.is(Kind.EQUAL_TO, Kind.NOT_EQUAL_TO) && isNullCheck((ExpressionTree) firstUsage)) {
+        return true;
+      }
+      // Return false if the only usage is not a null check
+      if (usages.size() == 1) {
+        return false;
+      }
+      // Fetch the first null check in the usages list
+      Optional<ExpressionTree> firstNullCheck = getFirstNullCheck(identifier);
```
margarita-nedzelska-sonarsource, line 150: It looks like you don't need to put there identifier as an argument, Just symbol or symbol and usages.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -127,17 +135,61 @@ private boolean isSafeBooleanExpression(ExpressionTree tree) {
...
+      // Test if the first null check and the first usage are part of the same higher if structure
+      Optional<IfStatementTree> ifStatementWithNullCheck = getParentConditionalBranch(firstNullCheck.get());
+      Optional<IfStatementTree> ifStatementWithFirstUsage = getParentConditionalBranch((ExpressionTree) firstUsage);
+      return ifStatementWithNullCheck.equals(ifStatementWithFirstUsage);
     }
     return false;
   }
 
+  private static Optional<ExpressionTree> getFirstNullCheck(IdentifierTree identifier) {
+    if (firstNullCheckCache.containsKey(identifier)) {
```
margarita-nedzelska-sonarsource, line 163: If you look at the coverage report, you can notice that this condition will never be true.
It's because you're using 'IdentifierTree' as a key.

In case you want to cache the result for the variable, that might appear several times in the code, you need to use Symbol as a Key of map.

On the other hand, do you feel this situation is likely enough, so we need to cache? I don't know, so I am ok with and without caching.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -127,17 +135,61 @@ private boolean isSafeBooleanExpression(ExpressionTree tree) {
...
+      return firstNullCheckCache.get(identifier);
+    }
+    List<IdentifierTree> usages = identifier.symbol().usages();
+    Optional<ExpressionTree> firstNullCheck = getFirstNullCheck(usages);
+    firstNullCheckCache.put(identifier, firstNullCheck);
```
margarita-nedzelska-sonarsource, line 168: I believe we should not put Optional in cache. Just put it there if value is present

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -127,17 +135,61 @@ private boolean isSafeBooleanExpression(ExpressionTree tree) {
...
+  private static Optional<IfStatementTree> getParentConditionalBranch(ExpressionTree tree) {
+    if (ifStatementCache.containsKey(tree)) {
+      return ifStatementCache.get(tree);
+    }
+    Tree parent = tree;
+    while (parent != null && !parent.is(Kind.IF_STATEMENT)) {
+      parent = parent.parent();
+    }
+    Optional<IfStatementTree> result = Optional.ofNullable((IfStatementTree) parent);
+    ifStatementCache.put(tree, result);
```
margarita-nedzelska-sonarsource, line 189: Here too, let's not put empty optionals in cache, just it's value, when present

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -127,17 +135,61 @@ private boolean isSafeBooleanExpression(ExpressionTree tree) {
...
       }
+      // Test if the first null check and the first usage are part of the same higher if structure
+      Optional<IfStatementTree> ifStatementWithNullCheck = getParentConditionalBranch(firstNullCheck.get());
+      Optional<IfStatementTree> ifStatementWithFirstUsage = getParentConditionalBranch((ExpressionTree) firstUsage);
+      return ifStatementWithNullCheck.equals(ifStatementWithFirstUsage);
     }
     return false;
   }
 
+  private static Optional<ExpressionTree> getFirstNullCheck(IdentifierTree identifier) {
```
margarita-nedzelska-sonarsource, line 162: It seems to me it will be ok for this method to return @Nullable value. Optionals are great, but in some cases, it will make code easier to read and understand if we keep things simple.

margarita-nedzelska-sonarsource (commented):
Nice work!
We're almost there, just a few last comments.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
dorian-burihabwa-sonarsource: I picked a half-way solution where the method still returns an Optional but the cache only contains direct references.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
dorian-burihabwa-sonarsource: The cache is now stores all the parents leading up to an `IfStatementTree` are cached at the end of the method.
We also add parents that lead nowhere to make better use of the cache.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -127,17 +137,66 @@ private boolean isSafeBooleanExpression(ExpressionTree tree) {
...
+  private static Optional<ExpressionTree> getFirstNullCheck(Symbol symbol) {
+    if (firstNullCheckCache.containsKey(symbol)) {
+      return Optional.ofNullable(firstNullCheckCache.get(symbol));
+    }
+    Optional<ExpressionTree> firstNullCheck = symbol.usages().stream()
+      .map(IdentifierTree::parent)
+      .filter(tree -> tree.is(Kind.EQUAL_TO, Kind.NOT_EQUAL_TO) && isNullCheck((ExpressionTree) tree))
+      .map(ExpressionTree.class::cast)
+      .findFirst();
+    firstNullCheckCache.put(symbol, firstNullCheck.isPresent() ? firstNullCheck.get() : null);
```
margarita-nedzelska-sonarsource, line 173:
```suggestion
    firstNullCheckCache.put(symbol, firstNullCheck.orElse(null));
```

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -127,17 +137,66 @@ private boolean isSafeBooleanExpression(ExpressionTree tree) {
...
+    return firstNullCheck;
+  }
+
+
+  private static Optional<IfStatementTree> getParentConditionalBranch(ExpressionTree tree) {
+    Deque<Tree> trees = new ArrayDeque<>();
```
margarita-nedzelska-sonarsource, line 179: Nothing to change, just curious why you chose this kind of Collection. As far as I understand the logic, it doesn't really matter because you don't need the order and you'll loose the order anyway after putting in a HashMap.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -127,17 +137,66 @@ private boolean isSafeBooleanExpression(ExpressionTree tree) {
...
+      }
+      // Fetch the first null check in the usages list
+      Optional<ExpressionTree> firstNullCheck = getFirstNullCheck(identifier.symbol());
+      if (!firstNullCheck.isPresent()) {
+        return false;
       }
+      // Test if the first null check and the first usage are part of the same higher if structure
+      Optional<IfStatementTree> ifStatementWithNullCheck = getParentConditionalBranch(firstNullCheck.get());
+      Optional<IfStatementTree> ifStatementWithFirstUsage = getParentConditionalBranch((ExpressionTree) firstUsage);
+      return ifStatementWithNullCheck.equals(ifStatementWithFirstUsage);
```
margarita-nedzelska-sonarsource, line 159: What do you think if we rewrite this in a more functional style?

```suggestion
return getFirstNullCheck(identifier.symbol())
        .map(expressionTree -> getParentConditionalBranch(expressionTree)
          .flatMap(st -> getParentConditionalBranch(((ExpressionTree) firstUsage))
            .map(st::equals))
          .orElse(false))
        .orElse(false);
```

margarita-nedzelska-sonarsource (approved):
Nice job!

I suggested few minor improvements.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
dorian-burihabwa-sonarsource: Yeah, I initially used a stack but a List could the job as well. Will change that.

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
dorian-burihabwa-sonarsource: Not sure this makes the code more readable

PR #3473, opened 2021-03-02
SONARJAVA-3696 S2755 should not raise when a xml document is build

quentin-jaquier-sonarsource:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3475, opened 2021-03-04
SONARJAVA-3641 FN in S5998

sebastian-hungerecker-sonarsource:


quentin-jaquier-sonarsource (approved):
In addition to the fix on the test, you could refactor the description of the ticket, as it does not reflect what is actually implemented.

Otherwise, LGTM

PR #3477, opened 2021-03-09
SONARJAVA-3701 Prepare the move of the regex parser into its own project

sebastian-hungerecker-sonarsource:
This removes all references to interna of SonarJava from the general
regex classes and moves it to the SonarJava-specific JavaRegexSource


java-checks/src/main/java/org/sonar/java/checks/helpers/RegexTreeHelper.java
```diff
@@ -51,37 +49,35 @@ private RegexTreeHelper() {
...
   }
 
   private static boolean isMark(CharacterTree currentChar) {
     return MARK_PATTERN.matcher(currentChar.characterAsString()).matches();
   }
 
-  private static void addCurrentGrapheme(List<RegexCheck.RegexIssueLocation> result, @Nullable RegexSyntaxElement start, @Nullable RegexSyntaxElement end) {
-    if (start != null && end != null) {
-      result.add(new RegexCheck.RegexIssueLocation(start, end, ""));
+  private static void addCurrentGrapheme(List<RegexCheck.RegexIssueLocation> result, List<RegexSyntaxElement> codePoints) {
```
m-g-sonar, line 78: the list is called `codeUnits` in the previous method code, and here `codePoints`. I would uniformize it.

java-checks/src/main/java/org/sonar/java/checks/helpers/RegexTreeHelper.java
```diff
@@ -51,37 +49,35 @@ private RegexTreeHelper() {
...
+    if (codePoints.size() > 1) {
+      result.add(new RegexCheck.RegexIssueLocation(codePoints, ""));
```
m-g-sonar, line 80: I always fear when giving a reference to a list to another object constructor (what if we modify again this object here, would it have a consequence in the `RegexIssueLocation` ?)
Looking at the code, I can see that it's not the case, but I feel it might be safer to pass a copy anyway, just in case.

java-frontend/src/main/java/org/sonar/java/regex/ast/JavaRegexSource.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  @Override
+  public String getSourceText() {
+    return sourceText;
+  }
+
+  public List<TextSpan> textSpansFor(IndexRange range) {
+    List<TextSpan> result = new ArrayList<>();
+    Map.Entry<Integer, TextSpan> startEntry = indexToTextSpan.floorEntry(range.getBeginningOffset());
```
m-g-sonar: I really wonder about using `Map.Entry<>`. Can't we use a dedicated type that would encode the logic of the key/value couple with more meaningful names?

I feel it would radically simplify the readiness of the method.

m-g-sonar (commented):
Just a few comments which I feel could slightly improve things. Otherwise, looks promising!

PR #3481, opened 2021-03-11
SONARJAVA-3625 Possible FP in S5998 when using backreferences to large groups

alban-auzeill:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3485, opened 2021-03-12
SONARJAVA-3653 S5996 should raise issues even if the regex can match the empty string

margarita-nedzelska-sonarsource:


sebastian-hungerecker-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/regex/ImpossibleBoundariesCheck.java
```diff
@@ -69,26 +81,74 @@ public void visitLookAround(LookAroundTree tree) {
...
+    }
+
+    public Set<BoundaryTree> foundBoundaries() {
+      return new HashSet<>(foundBoundaries);
+    }
+  }
+
+  private boolean canReachWithConsumingInput(AutomatonState start, AutomatonState goal, Set<AutomatonState> visited) {
+    if (start == goal 
+      || start instanceof LookAroundTree
```
sebastian-hungerecker-sonarsource, line 137: Returning false when encountering a LookAroundTree is wrong. Given the regex `a(?=.)bc`, the path from a to c does consume input, whereas for `a(?=.)c` it does not. So we should indeed ignore characters consumed inside the lookaround, but we shouldn't discard the entire path because of it. Instead we should ignore the contents of the lookaround and continue with the continuation:

```
if (start instanceof LookAroundTree) {
    return canReachWithConsumingInput(start.continuation(), goal, visited);
}
```

(after the `visited.add(start);` line).

java-checks/src/main/java/org/sonar/java/checks/regex/ImpossibleBoundariesCheck.java
```diff
@@ -69,26 +81,74 @@ public void visitLookAround(LookAroundTree tree) {
...
+
+    public Set<BoundaryTree> foundBoundaries() {
+      return new HashSet<>(foundBoundaries);
+    }
+  }
+
+  private boolean canReachWithConsumingInput(AutomatonState start, AutomatonState goal, Set<AutomatonState> visited) {
+    if (start == goal 
+      || start instanceof LookAroundTree 
+      || start.continuation() instanceof LookAroundTree
```
sebastian-hungerecker-sonarsource, line 138: With the changes suggested above, this line should be removed as well.

java-checks/src/main/java/org/sonar/java/checks/regex/ImpossibleBoundariesCheck.java
```diff
@@ -20,19 +20,31 @@
...
+import static org.sonar.java.regex.ast.AutomatonState.TransitionType.CHARACTER;
+
 @Rule(key = "S5996")
 public class ImpossibleBoundariesCheck extends AbstractRegexCheck {
 
   private static final String MESSAGE = "Remove or replace this boundary that will never match because it appears %s mandatory input.";
+  private static final String SOFT_MESSAGE =
+    "Remove or replace this boundary that can only match if the previous part matched the empty string because it appears %s mandatory input.";
+  private final Set<RegexTree> excluded = new HashSet<>();
+  private final RegexReachabilityChecker regexReachabilityChecker = new RegexReachabilityChecker(false);
```
sebastian-hungerecker-sonarsource, line 47: This checker should have its cache cleared when visiting a new regex (i.e. on every call to `checkRegex`). Having the states of the old regex in the cache won't buy us anything (it will never lead to a cache hit) and just increase the chance that we hit the cache size limit.

java-checks/src/main/java/org/sonar/java/checks/regex/ImpossibleBoundariesCheck.java
```diff
@@ -69,26 +81,74 @@ public void visitLookAround(LookAroundTree tree) {
...
+      || visited.contains(start)) {
+      return false;
+    }
+
+    visited.add(start);
+
+    for (AutomatonState successor : start.successors()) {
+      AutomatonState.TransitionType transition = successor.incomingTransitionType();
+      if (((transition == CHARACTER) && regexReachabilityChecker.canReach(successor, goal))
+        || canReachWithConsumingInput(successor, goal, visited)) {
```
sebastian-hungerecker-sonarsource, line 148: This still leads to some unnecessary work when `canReach` returns false because it will still try `canReachWithConsumingInput` even though it can never return true in that case (which is also why you can't get full coverage here). You should change it so that `canReachWithConsumingInput` is only called when the edge is not a character edge.

sebastian-hungerecker-sonarsource (changes requested):
Looks good, but some changes required

java-checks/src/main/java/org/sonar/java/checks/regex/ImpossibleBoundariesCheck.java
```diff
@@ -69,26 +82,74 @@ public void visitLookAround(LookAroundTree tree) {
...
           // Do nothing
       }
     }
+
+    private boolean probablyShouldConsumeInput(AutomatonState start, AutomatonState goal) {
+      return  canReachWithConsumingInput(start, goal, new HashSet<>());
+    }
+  }
+
+  private static class BoundaryInDisjunctionFounder extends RegexBaseVisitor {
```
sebastian-hungerecker-sonarsource, line 123: s/Founder/Finder/

sebastian-hungerecker-sonarsource (approved):
One small typo that I didn't notice before. Other than that, everything's great.

PR #3486, opened 2021-03-12
SONARJAVA-3733 ReDoS: Don't call cubic and worse runtimes quadratic

sebastian-hungerecker-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/regex/RedosCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/regex/RedosCheck.java
@@ -206,9 +206,7 @@ public class RedosCheck extends AbstractRegexCheckTrackingMatchType {
      * When used for partial matches, a regex acts as if it had `(?s:.*)` attached to its beginning and end unless anchored.
      */
     private boolean overlapsWithImplicitMatchAlls(RepetitionTree tree) {
-      return isUsedForPartialMatch &&
-        (canReachWithoutConsumingInputOrGoingThroughBoundaries(startOfRegex, tree)
-          || canReachWithoutConsumingInputOrGoingThroughBoundaries(tree, endOfRegex));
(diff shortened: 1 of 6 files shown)


java-checks/src/main/java/org/sonar/java/checks/regex/RedosCheck.java
```diff
@@ -181,13 +211,16 @@ private void checkForOverlappingRepetitions(RepetitionTree tree) {
           if (reachabilityChecker.canReach(repetition, tree)) {
             SubAutomaton auto1 = new SubAutomaton(repetition.getElement(), tree.continuation(), false);
             SubAutomaton auto2 = new SubAutomaton(repetition.continuation(), tree.continuation(), false);
+            intersectionChecker.clearCache();
```
quentin-jaquier-sonarsource, line 214: Not sure to understand why we need to clear the cache here, could you explain this change?

sebastian-hungerecker-sonarsource: Damn, I forgot to remove that line. I added it while debugging when `intersects` kept returning different results than what it clearly should and what it did return in self-contained tests. So I suspected a caching bug (which turned out to be the case) and added the `clearCache` to verify. The issue went away, so that confirmed the caching bug. Then I (thought I) fixed the caching bug, but forgot to remove the `clearCache` before testing. Turns out the bug was only partially fixed and I'll have to investigate further :frowning_face:

sebastian-hungerecker-sonarsource: Turns out the part that wasn't fully fixed was due to another bug that I fixed in the PR that Alban merged earlier. So rebasing on master fixed the issue. :smile:

quentin-jaquier-sonarsource (approved):
LGTM

PR #3487, opened 2021-03-14
Return all overrides, not just the first one

mc1arke:
The `overriddenSymbol` method currently returns a single method that the current method overrides - checking through the parent interfaces, then super classes, and finally `Object` for the first matching method - even where the method is overridden from multiple sources, such as both an interface and parent class, or implemented from multiple interfaces. Where a rule needs to check all overridden methods for certain information - such as annotations being replicated onto classes from all parent interfaces - this incomplete list prevents these rules from performing a full check.

This change introduces an `overriddenSymbols` method that returns a list of methods that are being overridden across all parent interfaces and super classes, and updates the `overriddenSymbol` method to delegate to the new method and return the first matched method to retain backwards compatibility. As use of the existing `overriddenSymbol` method could potentially lead to misleading results, this method has been marked as deprecated in favour of the new method, although the usage in the existing rules have not been migrated since their usages do not seem to be directly impacted.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
+++ b/java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
@@ -19,17 +19,20 @@
  */
 package org.sonar.java.model;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
 import org.eclipse.jdt.core.dom.ASTUtils;
 import org.eclipse.jdt.core.dom.IMethodBinding;
 import org.eclipse.jdt.core.dom.ITypeBinding;
 import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.semantic.Type;
 import org.sonar.plugins.java.api.tree.MethodTree;
-
-import javax.annotation.Nullable;
-import java.util.List;
-import java.util.Objects;
-import java.util.function.Predicate;
 
 final class JMethodSymbol extends JSymbol implements Symbol.MethodSymbol {
 
@@ -51,7 +54,7 @@ final class JMethodSymbol extends JSymbol implements Symbol.MethodSymbol {
   /**
    * Cache for {@link #overriddenSymbol()}.
    */
-  private MethodSymbol overriddenSymbol = Symbols.unknownMethodSymbol;
+  private List<MethodSymbol> overriddenSymbols;
 
   private final String signature;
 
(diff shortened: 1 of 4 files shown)


m-g-sonar:
Nice work @mc1arke !

This could indeed be a nice improvement to our API.
We won't have the time to include it in version 6.14 (scheduled for tomorrow), but we will get back to your PR quickly!

Cheers,
Michael

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
```diff
@@ -96,45 +99,81 @@ public TypeSymbol returnType() {
...
+      .collect(Collectors.toList());
+  }
+
+  private List<IMethodBinding> find(Predicate<IMethodBinding> predicate, ITypeBinding typeBinding) {
+    List<IMethodBinding> bindings = new ArrayList<>();
+
+    addMissing(findSuperclassBindings(predicate, typeBinding.getSuperclass(), true), bindings);
+    addMissing(findInterfaceBindings(predicate, typeBinding.getInterfaces()), bindings);
+
+    ITypeBinding objectTypeBinding = Objects.requireNonNull(sema.resolveType("java.lang.Object"));
```
m-g-sonar, line 128: I'm really not sure we need to do it for every class for the object. this should probably be done only once. For the last superclass in the hierarchy, or the class itself if it does not have any super class.

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
```diff
@@ -96,45 +99,81 @@ public TypeSymbol returnType() {
...
-        return r;
+  private List<MethodSymbol> convertOverriddenSymbols() {
+    IMethodBinding methodBinding = methodBinding();
+    return find(methodBinding::overrides, methodBinding.getDeclaringClass()).stream()
+      .map(sema::methodSymbol)
+      .collect(Collectors.toList());
+  }
+
+  private List<IMethodBinding> find(Predicate<IMethodBinding> predicate, ITypeBinding typeBinding) {
+    List<IMethodBinding> bindings = new ArrayList<>();
```
m-g-sonar, line 123: If it's a list, we will hit the problem of duplicates, and so you have to explore it all the time to check if it's not already there. having a LinkedHashSet preserve the order of introduction in the collection, and will guarantee unicity already. I feel it might be easier.

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
```diff
@@ -96,45 +99,81 @@ public TypeSymbol returnType() {
...
-      if (r != null) {
-        return r;
+  private List<MethodSymbol> convertOverriddenSymbols() {
+    IMethodBinding methodBinding = methodBinding();
+    return find(methodBinding::overrides, methodBinding.getDeclaringClass()).stream()
+      .map(sema::methodSymbol)
+      .collect(Collectors.toList());
+  }
+
+  private List<IMethodBinding> find(Predicate<IMethodBinding> predicate, ITypeBinding typeBinding) {
```
m-g-sonar, line 122: we can probably get rid of the predicates everywhere, since it's always the same which is used.

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
```diff
@@ -96,45 +99,81 @@ public TypeSymbol returnType() {
...
     }
-    return null;
+
+    bindings.addAll(findSuperclassBindings(predicate, typeBinding.getSuperclass(), checkClassMethods));
+    bindings.addAll(findInterfaceBindings(predicate, typeBinding.getInterfaces()));
+
+    return bindings;
+  }
+
+  private static <T> void addMissing(Collection<T> items, Collection<T> target) {
```
m-g-sonar, line 171: Probably not required when using a LinkedHashSet.

java-frontend/src/main/java/org/sonar/plugins/java/api/semantic/Symbol.java
```diff
@@ -187,10 +187,20 @@
      * Note that if the method returns null, the method is not overriding any method for sure.
      *
      * @return the overridden symbol, null if the method is not overriding any method or overriding can not be determined (incomplete semantic)
+     * @deprecated This method only returns the first symbol that has been overridden, rather than all symbols that are being overridden. Use {@link #overriddenSymbols()} to get a list of all symbols
      */
     @Nullable
+    @Deprecated
     Symbol.MethodSymbol overriddenSymbol();
```
m-g-sonar, line 194: I think it worths having a look at usages in our rules to migrate them.

java-frontend/src/test/java/org/sonar/java/model/JMethodSymbolTest.java
```diff
@@ -45,6 +48,69 @@ void test() {
       .containsOnly(cu.sema.type(Objects.requireNonNull(cu.sema.resolveType("java.lang.Exception"))));
   }
 
+  @Test
+  void testSingleInheritance() {
+    JavaTree.CompilationUnitTreeImpl cu = test("interface A { void a(); }  class C implements A { public void a() { } }");
```
m-g-sonar, line 53: The test cases could be slightly rewritten to ease the readiness, for instance having one class/interface by line.

m-g-sonar (commented):
Pretty nice work @mc1arke, I feel this would definitely be a valuable change in our API!

I added a few comments to your PR, and I will handle them from there.

* I created the following ticket to track the change in our backlog: [SONARJAVA-3781](https://jira.sonarsource.com/browse/SONARJAVA-3781)
* I will keep your commit, to be sure you appear as one of our contributors.
* I will open a new PR, having a commit on top of yours to extends it a bit with respect to my comments.

Thanks again!

Michael

m-g-sonar:
Other PR open (#3511), so I'll close this one!
Thanks again!

PR #3488, opened 2021-03-16
SONARJAVA-3712 Support boundaries in RegexTreeHelper.intersects and supersetOf methods

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/helpers/IntersectAutomataChecker.java
```diff
@@ -31,13 +35,20 @@ protected boolean neutralAnswer() {
...
-    SimplifiedRegexCharacterClass characterClass1 = SimplifiedRegexCharacterClass.of(auto1.start);
-    SimplifiedRegexCharacterClass characterClass2 = SimplifiedRegexCharacterClass.of(auto2.start);
-    return ((characterClass1 != null) && (characterClass2 != null)) ?
-      (characterClass1.intersects(characterClass2, defaultAnswer) &&
-        auto1.anySuccessorMatch(successor1 -> auto2.anySuccessorMatch(successor2 ->
-          check(successor1, successor2, true)))) :
-      defaultAnswer;
+    boolean start1IntersectsStart2;
+    if (auto1.incomingTransitionType() == BOUNDARY && auto2.incomingTransitionType() == BOUNDARY) {
+      start1IntersectsStart2 = ((BoundaryTree) auto1.start).type() == ((BoundaryTree) auto2.start).type();
```
sebastian-hungerecker-sonarsource, line 40: We should return `defaultAnswer`, not `false`, when the boundaries have different types because they could still intersect. Alternatively we could check the intersection of different boundary types like this:

The start of line/input boundaries intersect each other as well as the word and grapheme boundaries.
The end of line/input (with or without final terminator) boundaries intersect each other as well as the word and grapheme boundaries.
The non-word boundary intersects only itself.

java-checks/src/main/java/org/sonar/java/checks/helpers/IntersectAutomataChecker.java
```diff
@@ -31,13 +35,20 @@ protected boolean neutralAnswer() {
...
+    } else {
+      SimplifiedRegexCharacterClass characterClass1 = SimplifiedRegexCharacterClass.of(auto1.start);
+      SimplifiedRegexCharacterClass characterClass2 = SimplifiedRegexCharacterClass.of(auto2.start);
+      if (characterClass1 == null || characterClass2 == null) {
+        return defaultAnswer;
+      }
+      start1IntersectsStart2 = characterClass1.intersects(characterClass2, defaultAnswer);
+    }
+    return start1IntersectsStart2 &&
```
sebastian-hungerecker-sonarsource, line 49: I think the method would be a bit nicer to read if we moved the previous if-else into its own method `charactersIntersect` and replaced `start1IntersectsStart2` with `charactersIntersect(auto1.start, auto2.start)` here.

This would also fix a bug I've just noticed where, if one of the character classes is null and the `defaultAnswer` is `true`, we return true without checking the successors.

java-checks/src/main/java/org/sonar/java/checks/helpers/IntersectAutomataChecker.java
```diff
@@ -31,13 +35,20 @@ protected boolean neutralAnswer() {
...
+      auto1.anySuccessorMatch(successor1 -> auto2.anySuccessorMatch(successor2 ->
+      check(successor1, successor2, true)));
```
sebastian-hungerecker-sonarsource, line 51: I think this line should be indented further than the previous line because we're inside a lambda.

java-checks/src/main/java/org/sonar/java/checks/helpers/SupersetAutomataChecker.java
```diff
@@ -31,13 +35,20 @@ protected boolean neutralAnswer() {
...
-    SimplifiedRegexCharacterClass characterClass1 = SimplifiedRegexCharacterClass.of(auto1.start);
-    SimplifiedRegexCharacterClass characterClass2 = SimplifiedRegexCharacterClass.of(auto2.start);
-    return ((characterClass1 != null) && (characterClass2 != null)) ?
-      (characterClass1.supersetOf(characterClass2, defaultAnswer) &&
-        auto2.allSuccessorMatch(successor2 -> auto1.anySuccessorMatch(successor1 ->
-          check(successor1, successor2, true)))) :
-      defaultAnswer;
+    boolean start1SupersetOfStart2;
+    if (auto1.incomingTransitionType() == BOUNDARY && auto2.incomingTransitionType() == BOUNDARY) {
+      start1SupersetOfStart2 = ((BoundaryTree) auto1.start).type() == ((BoundaryTree) auto2.start).type();
```
sebastian-hungerecker-sonarsource, line 40: Here we should use `defaultAnswer` instead of `false` when the types are not equal. Alternatively, we can implement a partial hierarchy for boundaries like this:

UNICODE_EXTENDED_GRAPHEME_CLUSTER ⊇WORD ⊇ LINE_START ⊇ INPUT_START
UNICODE_EXTENDED_GRAPHEME_CLUSTER ⊇WORD ⊇ LINE_END ⊇ INPUT_END_FINAL_TERMINATOR ⊇INPUT_END
NON_WORD isn't a superset or subset of any other boundary.

sebastian-hungerecker-sonarsource (changes requested):
Good work, some comments.

alban-auzeill:
The implementation not so easy, this attempt is a failure. I close it because it is now too far from the master branch to be reused.

PR #3490, opened 2021-03-17
SONARJAVA-3689: Improve rule S110 to not report when hierarchy is too big already in library code

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/DepthOfInheritanceTreeCheck.java
```diff
@@ -71,9 +77,11 @@ public void visitClass(ClassTree tree) {
     if (!isBodyOfEnumConstantTree(tree)) {
       Type superClass = tree.symbol().superClass();
       int dit = 0;
+      List<WildcardPattern> exclusionPatterns = getPatterns();
```
m-g-sonar, line 80: This is not equivalent to what we were doing with `getPatterns()` before. We need to cache the patterns in the instance of the check to not recompute them every time we enter a class.

java-checks/src/main/java/org/sonar/java/checks/DepthOfInheritanceTreeCheck.java
```diff
@@ -102,10 +110,11 @@ void setMax(int max) {
   }
 
   private List<WildcardPattern> getPatterns() {
-    if (filteredPatterns == null) {
-      filteredPatterns = Arrays.asList(PatternUtils.createPatterns(filteredClasses));
-    }
-    return filteredPatterns;
+    String patterns = String.join(",",
```
m-g-sonar, line 113: this should be changed to be computed only once!

java-checks/src/test/java/org/sonar/java/checks/DepthOfInheritanceTreeCheckTest.java
```diff
@@ -22,50 +22,53 @@
...
+import static org.sonar.java.checks.verifier.TestUtils.nonCompilingTestSourcesPath;
 import static org.sonar.java.checks.verifier.TestUtils.testSourcesPath;
 
-class DITCheckTest {
+class DepthOfInheritanceTreeCheckTest {
 
   @Test
   void defaults() {
     JavaCheckVerifier.newVerifier()
-      .onFile("src/test/files/checks/DitOk.java")
```
m-g-sonar, line 32: I would also rename the test files to contain the name of the check in their name.
```
DitOk.java -> DepthOfInheritanceTreeCheckOk.java
``` 
This comment applies to other files as well.

java-checks-test-sources/src/main/java/checks/DitFrameworkExclusion.java
```diff
@@ -0,0 +1,23 @@
...
+  class MyJFrame extends JFrame {
+  }
+
+  class MyLogPermission extends LogPermission {
+    public MyLogPermission(String name, String actions) {
+      super(name, actions);
+    }
+  }
+
+  class MyDataAccessException extends DataAccessException {
```
m-g-sonar, line 17: To me we still miss a test case where there is actually N levels BEFORE reaching an ignored class, and so we would raise an issue.

PR #3493, opened 2021-03-18
SONARJAVA-3731: S5786 should not report on abstract classes or overridding test methods

dorian-burihabwa-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3494, opened 2021-03-18
SONARJAVA-3747 FPs in S5852 when repetition overlaps with non-repetition part

sebastian-hungerecker-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3498, opened 2021-03-22
SONARJAVA-3715: Add size of file to slowest files analyzed output

dorian-burihabwa-sonarsource:

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/ast/JavaAstScanner.java
+++ b/java-frontend/src/main/java/org/sonar/java/ast/JavaAstScanner.java
@@ -20,7 +20,6 @@
 package org.sonar.java.ast;
 
 import com.sonar.sslr.api.RecognitionException;
-import java.io.IOException;
 import java.io.InterruptedIOException;
 import java.time.Clock;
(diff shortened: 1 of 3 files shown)


java-frontend/src/main/java/org/sonar/java/ast/JavaAstScanner.java
```diff
@@ -73,7 +74,13 @@ public void scan(Iterable<? extends InputFile> inputFiles) {
           cancelled = true;
           break;
         }
-        executionTimeReport.start(inputFile.toString());
+        int lengthInBytes = -1;
+        try {
+          lengthInBytes = inputFile.contents().length();
```
alban-auzeill, line 79: Because some implementations of InputFile load the file content at each `contents()` call, it would be better to have `executionTimeReport.start(inputFile)` without content length computation and only compute the length in the `if (currentAnalysisTime >= minRecordedOrderedExecutionTime) {` block.

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #3500, opened 2021-03-22
Refactoring: consistency in checks targeting specific Java version, remove useless flatMap call

margarita-nedzelska-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/FilesExistsJDK8Check.java
```diff
@@ -45,7 +45,7 @@
 
   @Override
   public boolean isCompatibleWithJavaVersion(JavaVersion version) {
-    return version.asInt() == 8;
+    return version.isJava8Compatible();
```
dorian-burihabwa-sonarsource, line 48: Looking at the [description of the rule](https://jira.sonarsource.com/browse/RSPEC-3725), it seems that we suggest not using `File.exists` with Java 8 but `isJava8Compatible` checks if the version is greater or equal to 8.
So this change would alter the behavior of the rule and the analyzer.
Is it the change we expected to implement?

java-checks/src/test/files/checks/FilesExistsJDK8Check_no_version.java
```diff
@@ -3,13 +3,13 @@
 class A {
   void foo() {
     Path path;
```
dorian-burihabwa-sonarsource, line 5: This change is due to the change in behavior of S3725

dorian-burihabwa-sonarsource (changes requested):
LGTM as there are no old-style version checks when a suitable method exists :+1:
But if we want to change the behavior of S3725 then the spec should be updated in kind.

java-checks/src/main/java/org/sonar/java/checks/FilesExistsJDK8Check.java
margarita-nedzelska-sonarsource: Decided to leave the check as is, cause our compatibility message says that if not set, we assume 8 or greater, not just 8

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

java-checks/src/main/java/org/sonar/java/checks/StandardCharsetsConstantsCheck.java
```diff
@@ -313,8 +313,7 @@ private static String getMethodRef(Symbol symbol) {
   private static Optional<String> getConstantName(ExpressionTree argument) {
     return argument.asConstant(String.class)
       .map(String::toUpperCase)
-      .map(ALIAS_TO_CONSTANT::get)
-      .flatMap(Optional::ofNullable);
+      .map(ALIAS_TO_CONSTANT::get);
```
m-g-sonar, line 316: nice... we could have a rule for that!

m-g-sonar (approved):

PR #3501, opened 2021-03-22
SONARJAVA-3716 Provide a user property to produce performance metrics

alban-auzeill:


dorian-burihabwa-sonarsource (changes requested):
LGTM apart from pattern of calling stop on the Duration object

dorian-burihabwa-sonarsource (approved):

dorian-burihabwa-sonarsource:
When the checks have passed, you can go ahead and merge it 👍🏿

PR #3502, opened 2021-03-22
SONARJAVA-3732 Execute the move of the regex parser into analyzer-commons

sebastian-hungerecker-sonarsource:
The pom currently contains a dependency on a snapshot versions of commons, so this PR shouldn't be merged until the corresponding commons PR is merged and released and the pom is updated.


java-frontend/src/main/java/org/sonar/java/regex/JavaRegexSourceTrackingTextSpans.java
```diff
@@ -28,31 +28,31 @@
...
-import org.sonar.java.regex.ast.IndexRange;
 import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonarsource.analyzer.commons.regex.JavaRegexSource;
+import org.sonarsource.analyzer.commons.regex.ast.IndexRange;
 
-public class JavaRegexSource implements RegexSource {
-
-  private final String sourceText;
-
+public class JavaRegexSourceTrackingTextSpans extends JavaRegexSource {
```
m-g-sonar, line 35: I'm not a huge fan of this big name. Anything simpler maybe, such as `JavaRegexTextSpans` ?

sebastian-hungerecker-sonarsource: `JavaRegexTextSpans` sounds like it's a collection of `TextSpan`s. I mean, it does contain a collection of `TextSpan`s (well, one `TextSpan` in most cases), but what it *is* is an implementation of the `RegexSource` interface for Java regular expressions that also allows you to get the `TextSpan` corresponding to each `RegexTree`. And I can't think of a shorter name that expresses that.

m-g-sonar: `JavaAnalyzerRegexSource`

PR #3503, opened 2021-03-23
SONARJAVA-3722 Compute measurement cost in performance metrics

alban-auzeill:


java-frontend/src/main/java/org/sonar/java/PerformanceMeasure.java
```diff
@@ -162,14 +165,51 @@ public void stop() {
...
 
+    private static void appendMeasurementCost() {
+      int numberOfSamples = 99;
+      String[] sampleNames = IntStream.range(0, numberOfSamples).mapToObj(i -> "m" + i).toArray(String[]::new);
+      Duration totalDuration = start("#MeasurementCost_v1");
+      PerformanceMeasure measurementCost = currentMeasure;
+      Duration temporaryDuration = start("#temporary");
+      measurementCost.getOrCreateChild("nanoTime").add(median(IntStream.range(0, numberOfSamples).mapToLong(i -> {
+        long start = System.nanoTime();
+        return System.nanoTime() - start;
```
dorian-burihabwa-sonarsource, line 188: No need for a variable, we could probably one line this one

java-frontend/src/main/java/org/sonar/java/PerformanceMeasure.java
```diff
@@ -162,14 +165,51 @@ public void stop() {
...
+      })));
+      measurementCost.getOrCreateChild("createChild").add(median(IntStream.range(0, numberOfSamples).mapToLong(i -> {
+        long start = System.nanoTime();
+        start(sampleNames[i]).stop();
+        return System.nanoTime() - start;
+      })));
+      measurementCost.getOrCreateChild("emptyDuration").add(median(Arrays.stream(sampleNames)
+        .map(n -> currentMeasure.childrenMap.get(n)).mapToLong(m -> m.totalDurationNanos)));
```
dorian-burihabwa-sonarsource, line 196: Should we name this counter `getChild`?

java-frontend/src/main/java/org/sonar/java/PerformanceMeasure.java
```diff
@@ -162,14 +165,51 @@ public void stop() {
...
+        long start = System.nanoTime();
+        start("measure").stop();
+        return System.nanoTime() - start;
+      })));
+      temporaryDuration.stop();
+      measurementCost.childrenMap.remove("#temporary");
+      totalDuration.stop();
+    }
+
+    protected static long median(LongStream measures) {
```
dorian-burihabwa-sonarsource, line 208: Should the scope of this method be private like the one above?

dorian-burihabwa-sonarsource (changes requested):
LGTM but I think a few tweaks would help

java-frontend/src/main/java/org/sonar/java/PerformanceMeasure.java
alban-auzeill: Replaced with `observationCost`

java-frontend/src/main/java/org/sonar/java/PerformanceMeasure.java
alban-auzeill: I think it's hard to understand with `-(System.nanoTime() - System.nanoTime())` because you have to understand the order of operands evaluation. So I keep it like this.

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #3506, opened 2021-03-25
SONARJAVA-3778 Fix performance hotspots in S103 due to slow regex

sebastian-hungerecker-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/TooLongLineCheck.java
```diff
@@ -99,10 +99,11 @@ private void visitFile() {
   }
```
alban-auzeill, line 100: At line 93 we are calling `removeIgnoredPatterns(origLine)` for all lines, without checking if `origLine.length() > maximumLineLength`, I do not expect that a lot of lines exceed maximumLineLength, we should test this before calling `removeIgnoredPatterns(origLine)`.

alban-auzeill (changes requested):

PR #3511, opened 2021-03-30
SONARJAVA-3781 Provide access to all method overrides and not only the first one

m-g-sonar:
Complete #3487


java-checks-test-sources/src/main/java/checks/AnonymousClassShouldBeLambdaCheck.java
```diff
@@ -89,19 +92,20 @@ public String handle() {
       @Override
       public String handle() {
         return "";
-      }; // this empty statement should not be counted!
+      } // this empty statement should not be counted!
```
quentin-jaquier-sonarsource, line 95: This change does not make sense to me, we should either add the empty statement or remove the comment.

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
```diff
@@ -96,45 +99,57 @@ public TypeSymbol returnType() {
...
-  @Nullable
-  private MethodSymbol convertOverriddenSymbol() {
-    IMethodBinding overrides = find(
-      methodBinding()::overrides,
-      methodBinding().getDeclaringClass()
-    );
-    if (overrides == null) {
-      return null;
+  @Override
+  public List<MethodSymbol> overriddenSymbols() {
```
quentin-jaquier-sonarsource, line 110: In the different rules using this method, we often use the first element of the list. However, this list is internally built thanks to a Set, I'm afraid we could introduce inconsistency due to the fact that the first element is not always the same one.
Same concern for `firstOverridenSymbol`, it will be the same between two calls, but not necessarily between two analyses.

java-frontend/src/main/java/org/sonar/plugins/java/api/semantic/Symbol.java
```diff
@@ -183,14 +183,24 @@
...
+    @Deprecated
     Symbol.MethodSymbol overriddenSymbol();
 
+    /**
+     * Retrieve the overridden symbols in all the known type hierarchy.
+     * Note that it will only returns the symbols which might be determined from known types. The list might therefore not be complete in case of missing dependencies.
+     *
+     * @return the overridden symbols, or an empty list if the method is not overriding any method or overriding can not be determined (incomplete semantics)
+     */
+    List<Symbol.MethodSymbol> overriddenSymbols();
```
quentin-jaquier-sonarsource, line 202: We changed the public API, we should document it in the API changes [documentation](https://github.com/SonarSource/sonar-java/blob/master/sonar-java-plugin/src/main/resources/static/documentation.md).

quentin-jaquier-sonarsource (commented):
A few minor comments together with deeper questions. Feel free to call me if you want to discuss them further.

java-checks-test-sources/src/main/java/checks/AnonymousClassShouldBeLambdaCheck.java
m-g-sonar: indeed, reverted.

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
```diff
@@ -99,81 +99,57 @@ public TypeSymbol returnType() {
...
-
-    if (checkClassMethods) {
-      for (IMethodBinding candidate : typeBinding.getDeclaredMethods()) {
-        if (predicate.test(candidate)) {
-          bindings.add(candidate);
-          checkClassMethods = false;
-        }
-      }
+  private Set<IMethodBinding> findOverridesInTypes(Predicate<IMethodBinding> overridesCondition, ITypeBinding... types) {
+    Set<IMethodBinding> overrides = new LinkedHashSet<>();
```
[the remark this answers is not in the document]
m-g-sonar: Indeed, I will replace them with an accumulator set that I pass along. Now, the `LinkedHashSet()` preserve order or insertion.

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
m-g-sonar: The set(s) used to build the list is done through `linkedHashSet`, and preserve the order of insertion.
Also, the list of parents types is always explored in the same order, being the declaration order.

java-frontend/src/main/java/org/sonar/plugins/java/api/semantic/Symbol.java
m-g-sonar: indeed. I will document it.

java-frontend/src/main/java/org/sonar/plugins/java/api/semantic/Symbol.java
```diff
@@ -183,19 +183,19 @@
     List<Type> thrownTypes();
 
     /**
-     * Retrieve the overridden symbol, which may may not be able to determine (returning 'unknown' symbol).
-     * Note that if the method returns null, the method is not overriding any method for sure.
+     * Retrieve the overridden symbol, which might not be able to determined (returning 'unknown' symbol).
```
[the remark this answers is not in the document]
m-g-sonar: reformulated

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
quentin-jaquier-sonarsource: Make sense, thanks for the clarification.

quentin-jaquier-sonarsource (approved):
LGTM, nice work!

PR #3513, opened 2021-03-30
SONARJAVA-3761 Rule S6213: Restricted Identifiers should not be used …

margarita-nedzelska-sonarsource:
…as Identifiers


java-checks/src/test/java/org/sonar/java/checks/RestrictedIdentifiersUsageCheckTest.java
```diff
@@ -0,0 +1,55 @@
...
+      .withJavaVersion(8)
+      .verifyIssues();
+  }
+
+  @Test
+  void test_non_compiling() {
+    JavaCheckVerifier.newVerifier()
+      .onFile(nonCompilingTestSourcesPath("checks/RestrictedIdentifiersUsageCheck.java"))
+      .withCheck(new RestrictedIdentifiersUsageCheck())
+      .withJavaVersion(14)
```
quentin-jaquier-sonarsource, line 44: Just mentioning: this will not change the actual JDK used, we are always using Java 11 currently. This is usually used when using a `JavaVersionAwareVisitor`. This test makes sense anyway, it makes sure the rule does not change his behavior with a different version.

margarita-nedzelska-sonarsource: Actually, it sets the version of Java, against which our code will be compiled. And it makes sense, because custom method "yield()" is not allowed to be called outside of the switch expression in Java 13-14, but is absolutely legal in Java 8. And it makes sense to make these versions explicit in tests.

java-checks/src/test/java/org/sonar/java/checks/RulesSanityTest.java
```diff
@@ -120,7 +120,7 @@ void test() throws Exception {
       .collect(Collectors.toList());
 
     assertThat(errorLogs)
-      .hasSize(6)
+      .hasSize(8)
```
margarita-nedzelska-sonarsource, line 123: In test for Java 8 `yield()` method is allowed to be outside of switch expression, while in the recent versions of Java it doesn't parse.

java-checks/src/test/java/org/sonar/java/checks/RestrictedIdentifiersUsageCheckTest.java
quentin-jaquier-sonarsource: Make sense

quentin-jaquier-sonarsource (approved):
LGTM

PR #3515, opened 2021-03-30
SONARJAVA-3751 Extend S4738 to suggest Java 9 "List.of", "Map.of", "S…

margarita-nedzelska-sonarsource:
…et.of" instead of Guava

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/ReplaceGuavaWithJavaCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ReplaceGuavaWithJavaCheck.java
@@ -33,13 +33,16 @@ import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.VariableTree;
 
 @Rule(key = "S4738")
-public class ReplaceGuavaWithJava8Check extends AbstractMethodDetection implements JavaVersionAwareVisitor {
+public class ReplaceGuavaWithJavaCheck extends AbstractMethodDetection implements JavaVersionAwareVisitor {
 
   private static final String USE_INSTEAD = "Use \"%s\" instead.";
 
   private static final String GUAVA_BASE_ENCODING = "com.google.common.io.BaseEncoding";
   private static final String GUAVA_OPTIONAL = "com.google.common.base.Optional";
   private static final String GUAVA_FILES = "com.google.common.io.Files";
+  private static final String GUAVA_IMMUTABLE_SET = "com.google.common.collect.ImmutableSet";
+  private static final String GUAVA_IMMUTABLE_LIST = "com.google.common.collect.ImmutableList";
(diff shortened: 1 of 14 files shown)


java-checks/src/main/java/org/sonar/java/checks/ReplaceGuavaWithJavaCheck.java
```diff
@@ -33,13 +33,16 @@
 import org.sonar.plugins.java.api.tree.VariableTree;
 
 @Rule(key = "S4738")
-public class ReplaceGuavaWithJava8Check extends AbstractMethodDetection implements JavaVersionAwareVisitor {
+public class ReplaceGuavaWithJavaCheck extends AbstractMethodDetection implements JavaVersionAwareVisitor {
```
quentin-jaquier-sonarsource, line 36: In addition to renaming this class, we shoud also rename the rule title in the RPSEC:
"Java features should be preferred to Guava"

java-frontend/src/main/java/org/sonar/plugins/java/api/JavaVersion.java
```diff
@@ -96,4 +96,6 @@
    * @return empty string if java version is properly set, default message otherwise.
    */
   String java8CompatibilityMessage();
+
+  boolean isJava9Compatible();
```
quentin-jaquier-sonarsource, line 100: We should put this method next to the other similar one (line 53), with proper javadoc since it will be part of the public API.

java-frontend/src/main/java/org/sonar/java/model/JavaVersionImpl.java
```diff
@@ -103,6 +104,11 @@ public String java8CompatibilityMessage() {
     return compatibilityMessage(JAVA_8);
   }
 
+  @Override
+  public boolean isJava9Compatible() {
```
quentin-jaquier-sonarsource, line 108: We should put this method next to the other similar one (line 77).

java-frontend/src/test/java/org/sonar/java/model/JavaVersionImplTest.java
```diff
@@ -86,11 +91,24 @@ void java_12() {
...
+  void java_15() {
+    JavaVersion version = new JavaVersionImpl(15);
+    assertThat(version.isJava6Compatible()).isTrue();
+    assertThat(version.isJava7Compatible()).isTrue();
+    assertThat(version.isJava8Compatible()).isTrue();
+    assertThat(version.isJava9Compatible()).isTrue();
+    assertThat(version.isJava12Compatible()).isTrue();
+    assertThat(version.isJava15Compatible()).isTrue();
+    assertThat(version.asInt()).isEqualTo(15);
+  }
```
quentin-jaquier-sonarsource, line 110: You added a test for Java 15, but not for Java 9, do you think it would be useful to add it?

quentin-jaquier-sonarsource (approved):
LGTM

PR #3516, opened 2021-03-31
SONARJAVA-3787 Children of Switch Statement should not be a Switch Expression

quentin-jaquier-sonarsource:
Internally, we represent Switch Statement thanks to a Switch Expression. It makes sense since both share the same structure.

However, this is only an internal detail, when we call "getChildren" on a Switch Statement, it should not contain the Switch Expression, but the actual children of the node. Similarly, when we visit a SwitchStatement, we should not scan the SwitchExpression, but directly visit the children, as we would do for a SwitchExpression.

This result in strange behavior in the `SubscriptionVisitor`:
If you register only on `SWITCH_EXPRESSION` but the code contains a `SWITCH_STATEMENT`, the nested expression will actually be visited. Similarly when using a BaseTreeVisitor, `visitSwitchExpression` will be called when your code only contains a Switch Statement.

Currently, we are never using `SWITCH_EXPRESSION` or overriding `visitSwitchExpression`, so this problem does not have any impact. However, it becomes a problem when we want to explicitly support Switch Expression in rules. The risk is that when registering on Switch Expression, the logic will actually be called two times for every Switch Statement (once for the Statement, and once for the nested Expression). In addition, if we want a rule supporting only Switch Expression, it is currently not possible.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/model/statement/SwitchStatementTreeImpl.java
+++ b/java-frontend/src/main/java/org/sonar/java/model/statement/SwitchStatementTreeImpl.java
@@ -19,8 +19,10 @@
  */
 package org.sonar.java.model.statement;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import org.sonar.java.collections.ListUtils;
 import org.sonar.java.model.JavaTree;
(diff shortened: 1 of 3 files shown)


quentin-jaquier-sonarsource:
This ticket will be solved differently, see #3521.

PR #3518, opened 2021-03-31
SONARJAVA-3753 S6208: Comma-separated labels should be used in Switch with colon case

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/SwitchCasesShouldBeCommaSeparatedCheck.java
```diff
@@ -0,0 +1,84 @@
...
+      int size = labels.size();
+      if (size == 1) {
+        continue;
+      }
+      while (size >= 2 && labels.get(size - 1).caseOrDefaultKeyword().text().equals("default")) {
+        size--;
+      }
+      if (size == 1) {
+        return;
+      }
```
alban-auzeill, line 64: We can not filter "default" only when it is at the end, here we raise the following FP:
```
    switch (mode) {
      default:
      case "a": // false-positive
        doSomething();
    }
```
And we should not `return` line 63, to not miss the following issues.
This my proposal:
```
    for (CaseGroupTree aCase : switchExpression.cases()) {
      List<CaseLabelTree> caseAndDefaultLabels = aCase.labels();
      if (caseAndDefaultLabels.size() == 1) {
        continue;
      }
      Deque<CaseLabelTree> caseLabels = aCase.labels().stream()
        .filter(label -> "case".equals(label.caseOrDefaultKeyword().text()))
        .collect(Collectors.toCollection(ArrayDeque::new));

      if (caseLabels.size() > 1) {
        CaseLabelTree lastLabel = caseLabels.removeLast();
        List<JavaFileScannerContext.Location> secondaries = caseLabels.stream()
          .map(label -> new JavaFileScannerContext.Location("", label))
        reportIssue(lastLabel, MESSAGE, secondaries, null);
      }
    }
```

java-checks/src/main/java/org/sonar/java/checks/SwitchCasesShouldBeCommaSeparatedCheck.java
```diff
@@ -0,0 +1,84 @@
...
+    return Arrays.asList(
+      Tree.Kind.SWITCH_EXPRESSION,
+      Tree.Kind.SWITCH_STATEMENT
+    );
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    SwitchTree switchExpression = (SwitchTree) tree;
+    if (usesArrows(switchExpression)) {
```
alban-auzeill, line 50: Small comment: If we test `!useColons` instead of `useArrows`, it's almost the same code but when java will introduce "yet another switch expression", this rule will continue to work without effort.

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #3521, opened 2021-04-01
SONARJAVA-3787 Children of Switch Statement should not be a Switch Expression

alban-auzeill:


sonar-java-plugin/src/main/resources/static/documentation.md
```diff
@@ -97,6 +97,19 @@ The tutorial [Writing Custom Java Rules 101](https://redirect.sonarsource.com/do
...
+
+* **Switch representation change in the AST**
+
+  Previously, Switch Statements were represented thanks to a Switch Expression. It means that the child of the Switch
+  Statement was a Switch Expression. This is no longer the case, a Switch statement is now a distinct node in the AST
+  and does not contain a Switch Expression anymore. This may impact existing custom rules relying explicitly on Switch
+  Expressions, via the kind SWITCH_EXPRESSION or the method visitSwitchExpression. More rarely, this could also impact
+  rules relying on parents of Tree, as the overall shape of the AST may also change.
+
+* **Deprecated**
```
quentin-jaquier-sonarsource, line 110: Maybe in addition to this point, we should add something about the new tree:

**New kind of tree: SwitchTree**

_Switch Expression and Switch Statement share the same fields, it can sometimes make sense to manipulate a Switch as either one. Previously, it was possible to use the method `asSwitchExpression()` to do this. Since this method is now deprecated, you can now use the new Kind of tree SwitchTree, containing all the elements shared between Switch Expressions and Statements._

PR #3522, opened 2021-04-01
SONARJAVA-3743 Rule S6203: Text blocks should not be used in complex expression

margarita-nedzelska-sonarsource:


margarita-nedzelska-sonarsource:
I will suggest to not merge this PR before we get Java 15 in CI. So we can put language level to 15 and put the into compiled folder

java-checks-test-sources/src/main/files/non-compiling/checks/TextBlocksInComplexExpressionsCheck.java
```diff
@@ -0,0 +1,74 @@
+package test;
+
+class TextBlocksInComplexExpressionsCheck {
+  
+  void fun() {
+    listOfString.stream()
+      
+      .map(str -> { // Noncompliant@+1 [[sc=18;ec=8]]{{Refactor this text block to a local variable or a static final field.}}
```
alban-auzeill, line 8: When the issue is on several lines we can also test the end line: `[[sc=18;el=+14;ec=8]]`

java-checks/src/main/java/org/sonar/java/checks/TextBlocksInComplexExpressionsCheck.java
```diff
@@ -0,0 +1,80 @@
...
+  private static final String MASSAGE = "Refactor this text block to a local variable or a static final field.";
+  private static final int DEFAULT_LINES_NUMBER = 5;
+  @RuleProperty(key = "MaximumNumberOfLines", 
+    description = "The maximum number of lines in a text block that can be nested into a complex expression.",
+    defaultValue = "" + DEFAULT_LINES_NUMBER)
+  private int linesNumber = DEFAULT_LINES_NUMBER;
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.LAMBDA_EXPRESSION);
```
alban-auzeill, line 45: I believe that just testing if the string is in a lambda is not enough. I consider the following example as false-positive:
```
    Supplier<String> supplier = () -> """
    <project>
      <modelVersion>4.0.0</modelVersion>
      <parent>
        <groupId>com.mycompany.app</groupId>
        <artifactId>my-app</artifactId>
        <version>1</version>
      </parent>

      <groupId>com.mycompany.app</groupId>
      <artifactId>my-module</artifactId>
      <version>1</version>
    </project>
    """;
```
Because the lambda is not used in a chain of method calls.

java-checks/src/main/java/org/sonar/java/checks/TextBlocksInComplexExpressionsCheck.java
```diff
@@ -0,0 +1,80 @@
...
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.LambdaExpressionTree;
+import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S6203")
+public class TextBlocksInComplexExpressionsCheck extends IssuableSubscriptionVisitor {
+
+  private static final String MASSAGE = "Refactor this text block to a local variable or a static final field.";
```
alban-auzeill, line 36: Good idea :laughing: ! But no. Unfortunately, our analyzer does not yet provide `MASSAGE` to reduce the stress of our users.

alban-auzeill: To me, the message is confusing, because even if I move the string in a local variable I still get an issue:
```
    Runnable f1 = () -> {
      // Noncompliant@+1
      String value = """
      <project>
        <modelVersion>4.0.0</modelVersion>
        <parent>
          <groupId>com.mycompany.app</groupId>
          <artifactId>my-app</artifactId>
          <version>1</version>
        </parent>

        <groupId>com.mycompany.app</groupId>
        <artifactId>my-module</artifactId>
        <version>1</version>
      </project>
      """;
      return x ? value : "";
    };
```
And because we allow TextBlockFinder to continue on visitClass, visitNewClass, we could also have:
```
    Runnable f2 = () -> {
      class A {
        // Noncompliant@+1
        static final String VALUE = """
          <project>
            <modelVersion>4.0.0</modelVersion>
            <parent>
              <groupId>com.mycompany.app</groupId>
              <artifactId>my-app</artifactId>
              <version>1</version>
            </parent>

            <groupId>com.mycompany.app</groupId>
            <artifactId>my-module</artifactId>
            <version>1</version>
          </project>
          """;
      }
      return VALUE;
    };
```
And because we allow TextBlockFinder to continue on visitLambdaExpression, we could have duplicated issues like:
```
    Runnable f1 = () -> {
      // Noncompliant@+2
      // Noncompliant@+1
      Supplier<String> value = () -> """
      <project>
        <modelVersion>4.0.0</modelVersion>
        <parent>
          <groupId>com.mycompany.app</groupId>
          <artifactId>my-app</artifactId>
          <version>1</version>
        </parent>

        <groupId>com.mycompany.app</groupId>
        <artifactId>my-module</artifactId>
        <version>1</version>
      </project>
      """;
      return value.get();
    };
```
To avoid all the above problems, we could prevent visiting blocks in `TextBlockFinder`:
```
    @Override
    public void visitBlock(BlockTree tree) {
      // skip
    }
 ```
The rule `S6203 Text blocks should not be used in complex expression` will have false-negative for lambda with text block `-> { }`, but I think it kind of ok because if the user complains, we can ask him to activate `S5612 Lambdas should not have too many lines`

File: java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6203_java.json [hunk 4]
```diff
@@ -0,0 +1,16 @@
+{
+  "title": "Text blocks should not be used in complex expression",
```
alban-auzeill, line 2: I would say `complex expressions`

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/TextBlocksInComplexExpressionsCheck.java
margarita-nedzelska-sonarsource: We talked with @quentin-jaquier-sonarsource about such cases and decided to not report issues here, becasue it's almost the same as having text block in a variable.

java-checks/src/main/java/org/sonar/java/checks/TextBlocksInComplexExpressionsCheck.java
margarita-nedzelska-sonarsource: Ok, I would rather ay that message is confusing. We should suggest to move TextBlocks somewhere outside from lambda.
Because any big text block in a lambda is not readable.

java-checks/src/main/java/org/sonar/java/checks/TextBlocksInComplexExpressionsCheck.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private final List<Tree> misusedTextBlocks = new ArrayList<>();
+
+    @Override
+    public void visitLiteral(LiteralTree tree) {
+      if (tree.is(Tree.Kind.TEXT_BLOCK)) {
+        String value = tree.value();
+        if (value.split("(\r?\n)|\r").length > maxLines) {
+          misusedTextBlocks.add(tree);
+        }
```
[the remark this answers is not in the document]
margarita-nedzelska-sonarsource: As far as I understand there is no need to make it split ("\r?\n|\r", -1), because we never get a trailing empty line in the text block. Every text block starts with '"""\n' and the first element of the split is always 'triple quotes'.

java-checks/src/main/java/org/sonar/java/checks/TextBlocksInComplexExpressionsCheck.java
margarita-nedzelska-sonarsource: Decided to filter out all lambdas which are not parts of the argument list. Wdyt? I suppose it will cove the majority of TP and filter the majority of FP.

java-checks/src/main/java/org/sonar/java/checks/TextBlocksInComplexExpressionsCheck.java
margarita-nedzelska-sonarsource: And I think lambdas with body in curly braces are also supported by this rule '{}'. You can see it in test examples

alban-auzeill (approved):

PR #3525, opened 2021-04-06
SONARJAVA-3759 Rule S6212: Local-Variable Type Inference (var) should…

margarita-nedzelska-sonarsource:
… be used


its/ruling/src/test/resources/eclipse-jetty/java-S6212.json
```diff
@@ -0,0 +1,4066 @@
+{
```
margarita-nedzelska-sonarsource, line 1: I checked nearly 500 issues. The majority of them where TP, some of them where FP to me, so I updated the rule implementation with some constraints, where it's not possible/not recommended to use "var"

java-checks/src/test/java/org/sonar/java/checks/VarCanBeUsedCheckTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.sonar.java.checks.verifier.JavaCheckVerifier;
+
+import static org.sonar.java.checks.verifier.TestUtils.testSourcesPath;
+
+class VarCanBeUsedCheckTest {
+
+  @Test
+  void test() {
+    JavaCheckVerifier.newVerifier()
+      .onFile(testSourcesPath("checks/VarCanBeUsedCheck.java"))
```
quentin-jaquier-sonarsource, line 32: You can probably extract `testSourcesPath("checks/VarCanBeUsedCheck.java")` into a field.

java-checks-test-sources/src/main/java/checks/VarCanBeUsedCheck.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+class MyTestClass {}
+
+public class VarCanBeUsedCheck {
+
+  public static final MyTestClass MyTest = new MyTestClass(); // Compliant
+  
+
```
quentin-jaquier-sonarsource, line 12: Extra empty lines.

java-checks/src/main/java/org/sonar/java/checks/VarCanBeUsedCheck.java
```diff
@@ -0,0 +1,95 @@
...
+      return false;
+    }
+    int line = firstToken.line();
+    if (typeAssignmentLine == line) {
+      return true;
+    }
+    typeAssignmentLine = line;
+    SyntaxToken token = variableTree.endToken();
+    return token != null && token.text().equals(",");
+  }
```
quentin-jaquier-sonarsource, line 94: This code looks strange but I'm not sure there is another way to write it... So I guess we will keep it as it is.

quentin-jaquier-sonarsource (commented):
The implementation LGTM. Let's wait the outcome of the discussion about the severity/profile before merging.

java-checks/src/main/java/org/sonar/java/checks/VarCanBeUsedCheck.java
margarita-nedzelska-sonarsource: I tried to find how it's handled in other rules and it was a similar approach. In situations like this `int a = 0, b = 1;` we can't put `var`, but there is no dedicated Kind. Maybe we should have it in the future.

margarita-nedzelska-sonarsource:
I decided to filter out primitives and parameterized types at the moment. The rule still suggests a lot. Let's see how it goes and probably we can increase the scope in the future.

quentin-jaquier-sonarsource (commented):
Almost there, let's finish this rule!

java-checks/src/main/java/org/sonar/java/checks/VarCanBeUsedCheck.java
```diff
@@ -0,0 +1,96 @@
...
+      !JUtils.isLocalVariable(variableTree.symbol()) ||
+      initializer.is(Tree.Kind.CONDITIONAL_EXPRESSION) ||
+      symbolType.isPrimitive() ||
+      symbolType.isParameterized()) {
+      return;
+    }
+
+    Type initializerType = initializer.symbolType();
+    if (symbolType.fullyQualifiedName().equals(initializerType.fullyQualifiedName())) {
+      reportIssue(variableTree.simpleName(), "");
```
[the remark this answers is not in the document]
margarita-nedzelska-sonarsource: Ooops

java-checks/src/main/java/org/sonar/java/checks/VarCanBeUsedCheck.java
```diff
@@ -0,0 +1,135 @@
...
+    if (symbolType.fullyQualifiedName().equals(initializerType.fullyQualifiedName()) &&
+      (isSelfExplained(initializer) ||
+        typeWasMentionedInTheName(identifierTree, symbolType) ||
+        typeWasMentionedInTheInitializer(initializer, symbolType))) {
+      reportIssue(identifierTree, MESSAGE);
+    }
+  }
+
+  private static boolean typeWasMentionedInTheName(IdentifierTree variable, Type type) {
+    return isLogicallyReferable(variable.name(), type.name());
```
quentin-jaquier-sonarsource, line 94: You call this method many times with the same second argument, and end up doing multiple times the same operation: `type.name().toLowerCase(Locale.ROOT)`. We should compute it only once.

quentin-jaquier-sonarsource (commented):
The current issues reported by the rule look good, I'm happy with the result!

Just one comment concerning the implementation.

quentin-jaquier-sonarsource (approved):
LGTM!

I'm quite happy with this implementation, looking forward to seeing it in action.

PR #3527, opened 2021-04-06
SONARJAVA-3789 Fix ClassCastException in S6202 (IsInstanceMethodCheck…

margarita-nedzelska-sonarsource:
…:70)


margarita-nedzelska-sonarsource:
The problem occurred on the input "int[].class.isInstance()", because ".class" can also be called on arrays, not only on class identifiers.

PR #3533, opened 2021-04-07
SONARJAVA-3744 Extend existing rules to support Switch Expression

quentin-jaquier-sonarsource:
A lot of lines changed, but that mainly for tests, the actual changes are trivial.


java-checks/src/main/java/org/sonar/java/checks/SwitchDefaultLastCaseCheck.java
```diff
@@ -19,37 +19,37 @@
...
-    getDefaultLabelAtWrongPosition(switchStatementTree).ifPresent(defaultLabel -> reportIssue(defaultLabel, "Move this default to the end of the switch."));
+    SwitchTree switchTree = (SwitchTree) tree;
+    getDefaultLabelAtWrongPosition(switchTree).ifPresent(defaultLabel -> reportIssue(defaultLabel, "Move this default to the end of the switch."));
   }
 
-  private static Optional<CaseLabelTree> getDefaultLabelAtWrongPosition(SwitchStatementTree switchStatementTree) {
-    for (int i = 0; i < switchStatementTree.cases().size(); i++) {
-      List<CaseLabelTree> labels = switchStatementTree.cases().get(i).labels();
+  private static Optional<CaseLabelTree> getDefaultLabelAtWrongPosition(SwitchTree switchTree) {
+    for (int i = 0; i < switchTree.cases().size(); i++) {
```
margarita-nedzelska-sonarsource, line 47: maybe `switchTree.cases()` can be extracted?

java-checks/src/main/java/org/sonar/java/checks/SwitchWithTooManyCasesCheck.java
```diff
@@ -19,23 +19,21 @@
...
-import org.sonar.plugins.java.api.tree.SwitchStatementTree;
+import org.sonar.plugins.java.api.tree.SwitchTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-
 @Rule(key = "S1479")
 public class SwitchWithTooManyCasesCheck extends IssuableSubscriptionVisitor {
```
margarita-nedzelska-sonarsource, line 35: I've just realized I have some doubts about exactly this rule and a Switch Expression. As far as I understand in the expression you need to cover all the cases, otherwise it's a compile-time error. So if we have an Enum with 10 values or a sealed class in the future. It could bring some FPs, because you have to cover all cases. wdyt?

java-checks/src/main/java/org/sonar/java/checks/SwitchWithTooManyCasesCheck.java
```diff
@@ -46,28 +44,28 @@
...
-    List<CaseGroupTree> cases = switchStatementTree.cases();
+    List<CaseGroupTree> cases = switchTree.cases();
     int size = cases.size();
     if (size > maximumCases) {
       List<JavaFileScannerContext.Location> secondary = new ArrayList<>();
       for (CaseGroupTree element : cases) {
         secondary.add(new JavaFileScannerContext.Location("+1", element.labels().get(0)));
       }
-      reportIssue(switchStatementTree.switchKeyword(), "Reduce the number of non-empty switch cases from " + size + " to at most " + maximumCases + ".", secondary, null);
+      reportIssue(switchTree.switchKeyword(), "Reduce the number of non-empty switch cases from " + size + " to at most " + maximumCases + ".", secondary, null);
```
margarita-nedzelska-sonarsource, line 64: Maybe `String.format()` ?

java-checks/src/main/java/org/sonar/java/checks/NestedSwitchCheck.java
```diff
@@ -19,28 +19,30 @@
...
   public List<Tree.Kind> nodesToVisit() {
-    return Collections.singletonList(Tree.Kind.SWITCH_STATEMENT);
+    return Arrays.asList(Tree.Kind.SWITCH_STATEMENT, Tree.Kind.SWITCH_EXPRESSION);
   }
 
   @Override
   public void visitNode(Tree tree) {
     NestedSwitchVisitor visitor = new NestedSwitchVisitor();
-    ((SwitchStatementTree) tree).cases().stream().forEach(c -> c.accept(visitor));
+    ((SwitchTree) tree).cases().stream().forEach(c -> c.accept(visitor));
```
margarita-nedzelska-sonarsource, line 45: I see it was there before. I just don't like when there are more than 2 calls in one line, especially with lambdas. Maybe it will be better if you move `foreach` to the next line.
It minor and just a matter of taste.

margarita-nedzelska-sonarsource (commented):
Overall looks good!
Nice work!

Just have a question about 1 rule and a few minor comments.

java-checks/src/main/java/org/sonar/java/checks/SwitchWithTooManyCasesCheck.java
quentin-jaquier-sonarsource: In fact, that already the case for Switch Statements. You don't have to, but covering all possible values of an enum is something you might want to do. This is why this rule is not triggered with enum. it might become an issue for sealed classes, we might want to exclude them in the future.

margarita-nedzelska-sonarsource (approved):

PR #3539, opened 2021-04-08
SONARJAVA-3744 Extend existing rules to support Switch Expression

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/java/checks/LeftCurlyBraceEndLineCheck.java
```diff
@@ -12,10 +16,12 @@
   }
   
   public enum MyEnum {
-    A
+    A(1)
+    // Duplicated issue...
+    // Noncompliant@+1
```
dorian-burihabwa-sonarsource, line 21: What is the same issue reported twice on an enum?

java-checks-test-sources/src/main/java/checks/NestedIfStatementsCheck.java
```diff
@@ -52,15 +54,15 @@ private void f() {
 
     for (int i = 0; i < 0; i++) { // Compliant - 1
       for (Object o: getObjects()) { // Compliant - 2
-        while (false) { // Compliant - 3
+        while (cond) { // Compliant - 3
 
-          for (int i = 0; i < 0; i++) { // Noncompliant
+          for (int j = 0; i < 0; i++) { // Noncompliant
```
dorian-burihabwa-sonarsource, line 59: Why isn't local variable `j` used instead of `i`?

dorian-burihabwa-sonarsource (changes requested):
LGTM but if we can clarify in the comments why we have an issue reported twice on the same enum that would be nice

java-checks-test-sources/src/main/java/checks/LeftCurlyBraceEndLineCheck.java
quentin-jaquier-sonarsource: The issue was in fact already reported twice before, but the code was not compiling and therefore the strange behavior was not visible.
In fact, this issue was there for many years since the rule is not in Sonarway and the related code is not that frequent, it was unnoticed. Still, it might deserve a ticket to fix this, but with low priority.

java-checks-test-sources/src/main/java/checks/NestedIfStatementsCheck.java
quentin-jaquier-sonarsource: Indeed, the code looks strange, but the code does not have to be meaningful as long as it compiles and test the rule correctly. I think it's fine to keep it.

PR #3543, opened 2021-04-09
SONARJAVA-3796 Fix possible Catastrophic backtracking in regex for S3…

margarita-nedzelska-sonarsource:
…518: Division by zero rule

### DIFF
--- a/java-symbolic-execution/src/main/java/org/sonar/java/se/checks/DivisionByZeroCheck.java
+++ b/java-symbolic-execution/src/main/java/org/sonar/java/se/checks/DivisionByZeroCheck.java
@@ -20,11 +20,14 @@
 package org.sonar.java.se.checks;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.sonar.check.Rule;
+import org.sonar.java.cfg.CFG;
 import org.sonar.java.model.ExpressionUtils;
 import org.sonar.java.se.CheckerContext;
 import org.sonar.java.se.Flow;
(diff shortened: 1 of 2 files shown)


margarita-nedzelska-sonarsource:
I wanted to use `asConstant()` first. But imo, this method generates too much overhead with string replace and parsing literals, while this path is considered to be executed very often. So I suggest doing this low level iteration.

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/DivisionByZeroCheck.java
```diff
@@ -407,12 +407,47 @@ private void handleLiteral(LiteralTree tree) {
...
         addZeroConstraint(sv, ZeroConstraint.ZERO);
       } else if (tree.is(Tree.Kind.INT_LITERAL, Tree.Kind.LONG_LITERAL, Tree.Kind.DOUBLE_LITERAL, Tree.Kind.FLOAT_LITERAL)) {
-        addZeroConstraint(sv, isNumberZero(value) ? ZeroConstraint.ZERO : ZeroConstraint.NON_ZERO);
+        addZeroConstraint(sv, isZeroLiteral(tree) ? ZeroConstraint.ZERO : ZeroConstraint.NON_ZERO);
       }
     }
 
-    private static boolean isNumberZero(String literalValue) {
-      return !(literalValue.matches("(.)*[1-9]+(.)*") || literalValue.matches("(0x|0X){1}(.)*[1-9a-fA-F]+(.)*") || literalValue.matches("(0b|0B){1}(.)*[1]+(.)*"));
+    // LiteralTree.asConstant() could fit here logically but is not an optimal algorithm for such a hot paths
```
m-g-sonar, line 414: Worth writing it in a proper javadoc comment, and give a bit more of the why.

But I'm now wondering, what about caching the couples (literal tree value; boolean)? That way we would not have to recompute it all the time as well.

m-g-sonar:
Nice work! 👍🏻

PR #3546, opened 2021-04-09
Fix quality flaw: simplify regex

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -101,7 +101,7 @@ private static BlockTagKey of(BlockTag tag, @Nullable String name) {
   private static final Tree.Kind[] CLASS_KINDS = PublicApiChecker.classKinds();
   private static final Tree.Kind[] METHOD_KINDS = PublicApiChecker.methodKinds();
   private static final List<String> GENERIC_EXCEPTIONS = Arrays.asList("Exception", "java.lang.Exception");
-  private static final Pattern BLOCK_TAG_LINE_PATTERN = Pattern.compile("^@\\S+.*");
+  private static final Pattern BLOCK_TAG_LINE_PATTERN = Pattern.compile("^@.+");
```
quentin-jaquier-sonarsource, line 104: The regex itself is subject to catastrophic backtracking, when the input is like "@aaa\n", with thousands of "a". In practice, we already process the string that will be matched by the regex, so I don't think we are really subject to a problem here. Still, the regex can be simplified because there is no point in being that precise to detect "Block tag", a line starting with "@" is already enough.

sebastian-hungerecker-sonarsource: `@\\S.*` would be equivalent to the original without the backtracking issue. I don't know Javadoc well enough to know whether "the `@` sign must be followed by a non-space" is an important property to check or not. If not, you can probably get rid of the regex altogether and just use `.startsWith("@")` or `.charAt(0) == '@'`.

java-checks/src/main/java/org/sonar/java/checks/HardcodedIpCheck.java
```diff
@@ -36,13 +36,13 @@
 @Rule(key = "S1313")
 public class HardcodedIpCheck extends BaseTreeVisitor implements JavaFileScanner {
 
-  private static final Pattern IP_V4_REGEX = Pattern.compile("([^\\d.]*\\/)?((?<ip>(?:\\d{1,3}\\.){3}\\d{1,3})(:\\d{1,5})?(?!\\d|\\.))(\\/.*)?");
+  private static final Pattern IP_V4_REGEX = Pattern.compile("([^\\d.]*\\/)?((?<ip>(?:\\d{1,3}\\.){3}\\d{1,3})(:\\d{1,5})?(?![\\d.]))(\\/.*)?");
```
sebastian-hungerecker-sonarsource, line 39: This will probably be more readable if we split the regex into multiple named (private static final String) variables, PROTOCOL, IP, PORT and PATH. That will hopefully make it more readable, fix the regex complexity issue and avoid the confusion I had initially about what the rule is even matching.

sebastian-hungerecker-sonarsource (changes requested):
See comments. Looks good otherwise.

java-checks/src/main/java/org/sonar/java/checks/HardcodedIpCheck.java
sebastian-hungerecker-sonarsource: The protocol, port and path parts could then also be reused in the IP_V6_URL pattern.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3549, opened 2021-04-12
Use contains on a Set for (possible) perf improvement

henryju:
While investigating a UI freeze issue in SonarLint, I ended up finding this possible performance hotspot. 
Note that I don't think it is critical, the fact that the stacktrace ended up on the `contains` call might just be coincidental.

### DIFF
--- a/sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
+++ b/sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
```diff
@@ -68,16 +69,17 @@ public void define(Context context) {
       .createRepository(CheckList.REPOSITORY_KEY, Java.KEY)
       .setName("SonarAnalyzer");
     List<Class<?>> checks = CheckList.getChecks();
-    new RulesDefinitionAnnotationLoader().load(repository, checks.toArray(new Class[]{}));
+    new RulesDefinitionAnnotationLoader().load(repository, checks.toArray(new Class[] {}));
     JavaSonarWayProfile.Profile profile = JavaSonarWayProfile.readProfile();
+    Set<String> sonarWayRuleKeys = new HashSet<>(profile.ruleKeys);
```
henryju, line 74: `profile.ruleKeys` is an ArrayList (at least with this current implementation of Gson) so the `contains` method is not very efficient.

PR #3560, opened 2021-04-19
SONARJAVA-3808 NPE in JMethodSymbol.overriddenSymbol

quentin-jaquier-sonarsource:


java-frontend/src/test/java/org/sonar/java/model/JMethodSymbolTest.java
```diff
@@ -154,6 +154,16 @@ void testUnknownInheritance() {
       retrieveMethodSymbol("KnownInterface", "a", cu));
   }
 
+  @Test
+  void testUnknownExtends() {
```
quentin-jaquier-sonarsource, line 158: To be fair, this test does not reproduce the problem reported in the ticket, I did not manage to reproduce the NPE...

m-g-sonar: So the test is probably useless, and does not reflect what we want to try. I would not add it, to be honest.

In your test I don't see anything unknown as well. What about starting from a method invocation?
it's also maybe possible to manually craft the situation by mocking some incomplete ECJ binding.

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
```diff
@@ -137,6 +137,10 @@ private void findOverridesInParentTypes(Collection<MethodSymbol> accumulator, Pr
 
   private void findOverridesInTypes(Collection<MethodSymbol> accumulator, Predicate<IMethodBinding> overridesCondition, ITypeBinding... types) {
     for (ITypeBinding type : types) {
+      if (type == null) {
+        // Can happen for unknown reason.
+        continue;
```
m-g-sonar, line 142: fix LGTM, but I really would like to find the root cause.
Did you try analyzer jetty on your machine?

java-frontend/src/test/java/org/sonar/java/model/JMethodSymbolTest.java
quentin-jaquier-sonarsource: Indeed, the type extended was supposed to be Unknown. I think it makes sense to keep it (the corrected version), to test that it behaves correctly both with an unknown parent and unknown interfaces.

Testing this code with mock turns out to be a bit tricky: it is not possible to mock/spy `JMethodSymbol` (it is final), and not easy to mock the arguments either, because the constructor is using an external helper method. I ended up making the method visible for testing.

java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
quentin-jaquier-sonarsource: I managed to reproduce the issue locally when analyzing jetty. The problem comes when analyzing [this line](https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/CustomRequestLog.java#L1095), more specifically when we call `MethodInvocation.symbol().overriddenSymbols()` for the method invocation `getServletContext()`. It goes through the parents just fine until it reaches `AbstractHandler`, where the call to `ITypeBinding.getSuperclass()` returns null for some unknown reasons. It seems to me that digging deeper at this point will take a long time, do you think it is still worth it?

PR #3561, opened 2021-04-19
SONARJAVA-3804 FP in S3077 when volatile is used with javax.annotation.concurrent.immutable and threadsafe annotations

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/VolatileNonPrimitiveFieldCheck.java
```diff
@@ -85,11 +86,27 @@ public void visitNode(Tree tree) {
...
   }
 
+  private static boolean isUnsafeVolatile(Type symbolType) {
+    return !(symbolType.isPrimitive()
+      || isImmutableType(symbolType)
+      || isSafelyAnnotated(symbolType.symbol().metadata()));
+  }
+
+  private static boolean isSafelyAnnotated(SymbolMetadata metadata) {
+    for (SymbolMetadata.AnnotationInstance a : metadata.annotations()) {
```
m-g-sonar, line 100: what about writing this as stream instead?
```java
return metadata.annotations().stream()
  .map(SymbolMetadata.AnnotationInstance::symbol)
  .map(Symbol::type)
  .anyMatch(type -> type.is("javax.annotation.concurrent.Immutable") || type.is("javax.annotation.concurrent.ThreadSafe"));
```

quentin-jaquier-sonarsource: Looks better indeed.

PR #3562, opened 2021-04-19
SONARJAVA-3785 Support SpringBootApplication 'scanBasePackages' configuration

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -97,12 +99,28 @@ public void visitNode(Tree tree) {
...
+    return Objects.requireNonNull(classSymbolMetadata.valuesForAnnotation(SPRING_BOOT_APP_ANNOTATION))
+      .stream()
+      .filter(v -> "scanBasePackages".equals(v.name()))
+      .map(SymbolMetadata.AnnotationValue::value)
+      .findFirst()
+      // list of packages to scan
+      .filter(Object[].class::isInstance)
+      .map(Object[].class::cast)
+      .map(Arrays::asList)
+      .map(list -> list.stream().map(String.class::cast).collect(Collectors.toList()))
```
quentin-jaquier-sonarsource, line 119: It seems unlikely to happen, but the values of annotations can be Identifiers, making this cast unsafe. Example:
```
class MyClass {
  private final String PACKAGE = "my.package";

  @SpringBootApplication(scanBasePackages = PACKAGE)
  public class SpringBootApp {
  }
}
```

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -97,12 +99,28 @@ public void visitNode(Tree tree) {
...
+      // package in which the annotated class is defined
+      .orElse(Collections.singletonList(classPackageName));
```
quentin-jaquier-sonarsource, line 121: This matches my understanding of the annotation and seems correct. We could be a bit clearer in the comment, to say that this is the default behavior of the annotation, for example:


Using this annotation without arguments tells Spring to scan the current package and all of its sub-packages.

java-checks-test-sources/src/main/java/checks/spring/s4605/springBootApplication/thirdApp/controller/Controller.java
```diff
@@ -0,0 +1,6 @@
+package checks.spring.s4605.springBootApplication.thirdApp.controller;
+
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class Controller { } // // Noncompliant [[sc=14;ec=24]] {{'Controller' is not reachable by @ComponentsScan or @SpringBootApplication. Either move it to a package configured in @ComponentsScan or update your @ComponentsScan configuration.}}
```
quentin-jaquier-sonarsource, line 6:
```suggestion
public class Controller { } // Noncompliant [[sc=14;ec=24]] {{'Controller' is not reachable by @ComponentsScan or @SpringBootApplication. Either move it to a package configured in @ComponentsScan or update your @ComponentsScan configuration.}}
```
It does not break the test, just to clean it.

java-checks/src/test/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheckTest.java
```diff
@@ -68,4 +76,19 @@ void testSpringBootApplication() {
...
+    final String testFolder = BASE_PATH + "springBootApplication/";
+    List<String> files = Arrays.asList(
+      testSourcesPath(testFolder + "thirdApp/SpringBootApp.java"),
+      testSourcesPath(testFolder + "thirdApp/domain/SomeClass.java"),
+      testSourcesPath(testFolder + "thirdApp/controller/Controller.java"));
+
+    JavaCheckVerifier.newVerifier()
+      .onFiles(files)
+      .withCheck(new SpringBeansShouldBeAccessibleCheck())
+      .verifyIssues();
```
quentin-jaquier-sonarsource, line 90: What bothers me when testing multiples files is that if you put the `Noncompliant` comment in `SomeClass.java`, line 6 and remove it from `Controller.java`, the test will pass despite the fact that the issue is not raised where you would expect.
I would rather test one file at a time, and set the classPath (`withClassPath`) if needed.

quentin-jaquier-sonarsource (changes requested):

java-checks/src/test/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheckTest.java
quentin-jaquier-sonarsource: (See related ticket describing the problem: [SONARJAVA-3643](https://jira.sonarsource.com/browse/SONARJAVA-3643))

m-g-sonar: Well, its indeed a limitation that I had to fight already when moving the other test files. However, your solution won't work either. The rule is based on an `EndOfAnalysisCheck` which requires analyzing all the files... So using the classpath won't work. The collects first annotations and at the end of the analysis, report.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
m-g-sonar: good catch

m-g-sonar: In fact nop... I forgot about it but we are not relying anymore on the tree when getting annotation values... but directly on the semantic engine. So it's always going to be an `Object[]` filled with `String`s. I anyway added a test to prove it.

java-checks/src/test/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheckTest.java
quentin-jaquier-sonarsource: Okay, thanks for the prescision, I missed that indeed.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
quentin-jaquier-sonarsource: As discussed, most of the time it will work, but there are still corner cases where the object could be something other than a String.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3563, opened 2021-04-19
SONARJAVA-3809: S5979 should not report on objects initialized with MockitoJUnit.rule() followed by options

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
```diff
@@ -154,16 +155,33 @@ private static boolean isMockitoJUnitRuleInvoked(ClassTree clazz) {
...
     return false;
   }
 
+  private static boolean isInitializedWithRule(MethodInvocationTree mit) {
+    MethodInvocationTree current = mit;
+    while (true) {
+      if (MOCKITO_JUNIT_RULE.matches(current)) {
+        return true;
+      }
+      MemberSelectExpressionTree memberSelect = (MemberSelectExpressionTree) current.methodSelect();
```
m-g-sonar, line 176: How do you know it's not simply a method invocation without a member select!?

The following code makes the rule crash:
```java
  public class FooTest3 {
    @Rule
    public MockitoRule rule = foo();

    @Mock // Compliant
    private Bar bar;

    private MockitoRule foo() { return null; }
  }
```

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
```diff
@@ -154,16 +155,33 @@ private static boolean isMockitoJUnitRuleInvoked(ClassTree clazz) {
     for (VariableTree field : collected) {
       if (field.type().symbolType().is("org.mockito.junit.MockitoRule")) {
         ExpressionTree initializer = field.initializer();
+        Symbol symbol = field.symbol();
         if (initializer != null && initializer.is(Tree.Kind.METHOD_INVOCATION) &&
-          MOCKITO_JUNIT_RULE.matches((MethodInvocationTree) initializer) &&
-          field.symbol().metadata().isAnnotatedWith(RULE_ANNOTATION)) {
+          symbol.metadata().isAnnotatedWith(RULE_ANNOTATION) &&
+          symbol.isPublic() &&
```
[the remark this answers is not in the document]
dorian-burihabwa-sonarsource: This is a JUnit4 way of initializing mocks and if the field is not public, the unit test crashes.
The whole idea of this rule it to catch errors of this kind but I do agree that without an appropriate error message it is not super helpful.
From this point, we can choose the "isPublic" test or change the error message. WDYT?

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoAnnotatedObjectsShouldBeInitializedCheck.java
dorian-burihabwa-sonarsource: I will integrate that

PR #3569, opened 2021-04-22
SONARJAVA-3795 Improve tests for S2638

quentin-jaquier-sonarsource:
This PR will not solve the ticket but improve the test to better illustrate the problems.
The actual fix will be done in a second step.


m-g-sonar (approved):
LGTM! Nice preparatory work!

PR #3587, opened 2021-05-03
Remove obsolete FIXMEs and TODOs

m-g-sonar:


java-symbolic-execution/src/test/files/se/FlowComputationRelSV.java
```diff
@@ -58,7 +58,6 @@ void npe1(Object a) {
 
   void npe2(Object a, Object b) {
     if (a == b) {
-      // FIXME SONARJAVA-2272
```
quentin-jaquier-sonarsource, line 61: The situation is not fixed, it should refer to https://jira.sonarsource.com/browse/SONARJAVA-2303 now.

java-frontend/src/main/java/org/sonar/java/model/InternalSyntaxTrivia.java
```diff
@@ -63,7 +63,6 @@ public boolean isLeaf() {
 
   @Override
   public void accept(TreeVisitor visitor) {
-    //FIXME do nothing
```
quentin-jaquier-sonarsource, line 66: Maybe we should keep the comment, without the `FIXME`, to be explicit that this method does nothing on purpose? Same in `InternalSyntaxToken.java`.

java-symbolic-execution/src/test/files/se/FlowComputationRelSV.java
m-g-sonar: missed this while reading SONARJASVA-2272

quentin-jaquier-sonarsource (approved):
LGTM

PR #3590, opened 2021-05-10
S2694 false positive on JUnit nested tests

Bananeweizen:
JUnit nested tests must be inner classes and cannot be converted to
static classes. See
https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested

I didn't get the integration tests running. Feel free to add and edit.


quentin-jaquier-sonarsource:
Hello @Bananeweizen 

What bothers me with this change is that S2694 targets only Main code, it is not supposed to raise an issue on Test code in the first place. Since test annotations are not supposed to be in Main code, the change is not so interesting.

Did I miss something?

Bananeweizen:
Seems like the project where I noticed this is actually configured wrongly. You are right, in that I don't see the same issue for correctly configured projects.

PR #3595, opened 2021-05-17
SONARJAVA-3806 Improve S1656 to rely on ECJ warnings

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/SelfAssignementCheck.java
```diff
@@ -19,29 +19,61 @@
...
 
-import java.util.Collections;
-import java.util.List;
-
 @Rule(key = "S1656")
 public class SelfAssignementCheck extends IssuableSubscriptionVisitor {
 
+  private static final String ISSUE_MESSAGE = "Remove or correct this useless self-assignment.";
+  private final Set<JWarning> warnings = new HashSet<>();
+  private final Set<JWarning> reportedWarnings = new HashSet<>();
```
sebastian-hungerecker-sonarsource, line 42: Instead of having two sets here and subtracting one from the other at the end, it would be simpler to just have the warnings set and remove entries from it as they're reported. This should also improve runtime complexity as we'd no longer loop over all the warnings in `markReportedWarnings` every time a warning is reported.

java-frontend/src/main/java/org/sonar/java/model/JParser.java
```diff
@@ -166,6 +166,8 @@ public static CompilationUnitTree parse(
     options.put(JavaCore.COMPILER_SOURCE, version);
     if (MAXIMUM_SUPPORTED_JAVA_VERSION.equals(version)) {
       options.put(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, "enabled");
+      // enabling all supported compiler warnings
+      JWarning.Type.compilerOptions().forEach(option -> options.put(option, "warning"));
```
sebastian-hungerecker-sonarsource, line 170: This shouldn't be inside the if statement.

sebastian-hungerecker-sonarsource (changes requested):
Looks good; some changes required.

java-checks/src/main/java/org/sonar/java/checks/SelfAssignementCheck.java
m-g-sonar: nice finding. I didn't think about this.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3596, opened 2021-05-18
SONARJAVA-3851 Rules targeting methods calls should work with incomplete semantic

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #3597, opened 2021-05-19
SONARJAVA-3846 Rules targeting returns should work with incomplete semantic

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/BooleanMethodReturnCheck.java
```diff
@@ -53,6 +55,10 @@ private static boolean returnsBoolean(MethodTree methodTree) {
     return methodTree.returnType().symbolType().is("java.lang.Boolean");
   }
 
+  private static boolean hasUnknownAnnotation(SymbolMetadata symbolMetadata) {
```
dorian-burihabwa-sonarsource, line 58: This looks like a good candidate for a new method on `SymbolMetadata` when the sprint is done

dorian-burihabwa-sonarsource (approved):
LGTM

PR #3598, opened 2021-05-19
SONARJAVA-3844: Rules targeting tests should work with incomplete semantic

dorian-burihabwa-sonarsource:


java-checks/src/test/java/org/sonar/java/checks/tests/IgnoredTestsCheckTest.java
```diff
@@ -33,4 +34,13 @@ void test() {
       .withCheck(new IgnoredTestsCheck())
       .verifyIssues();
   }
+
+  @Test
+  void test_without_semantic() {
+    JavaCheckVerifier.newVerifier()
+      .onFile(testSourcesPath("checks/IgnoredTestsCheck.java"))
+      .withCheck(new IgnoredTestsCheck())
+      .withClassPath(Collections.emptyList())
```
quentin-jaquier-sonarsource, line 43: Maybe we could consider at the end of this sprint adding a test that checks that all rules do not raise FP when the classPath is empty. It will help to not have to do this effort again in the future.
I'm a bit reluctant though because it will probably slow the (already slow) build quite a bit.
Let's keep the idea in mind and see if we can do something.

its/plugin/tests/src/test/java/com/sonar/it/java/suite/SonarLintTest.java
```diff
@@ -124,7 +124,6 @@ public void simpleTestFileJava() throws Exception {
     sonarlintEngine.analyze(standaloneAnalysisConfiguration, issues::add, null, null);
 
     assertThat(issues).extracting("ruleKey", "startLine", "inputFile.path", "severity").containsOnly(
-      tuple("java:S1607", 4, inputFile.getPath(), "MAJOR"),
```
quentin-jaquier-sonarsource, line 127: We should also explain why this issue is not raised anymore (requires semantic) in the code of the test (line 105).

quentin-jaquier-sonarsource (changes requested):
The second commit (Narrow down location of issues reported by S2187) seems a bit superfluous, we can test the location for a few cases, but I'm not sure we need to consistently test it.

quentin-jaquier-sonarsource (approved):

PR #3600, opened 2021-05-19
SONARJAVA-3807 Improve and simplify S4970 by relying on ECJ warnings

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/JWarning.java
```diff
@@ -89,7 +90,7 @@ boolean isMatching(IProblem warning) {
             warning.getSourceLineNumber(),
             astNode.getColumnNumber(warning.getSourceStart()),
             astNode.getLineNumber(warning.getSourceEnd()),
-            astNode.getColumnNumber(warning.getSourceEnd()));
+            astNode.getColumnNumber(warning.getSourceEnd()) + 1);
```
m-g-sonar, line 93: was also noticeable in review on PEACH. The end of highlighting was wrong by 1 character.

java-checks/src/main/java/org/sonar/java/checks/UnreachableCatchCheck.java
```diff
@@ -48,203 +42,67 @@
...
-
-      reportIssue(typeToTypeTree.get(hiddenType),
-        "Remove this type because it is unreachable as hidden by previous catch blocks.",
-        subtypesHiding.stream().map(type -> new JavaFileScannerContext.Location("Already catch the exception", typeToTypeTree.get(type)))
-          .collect(Collectors.toList()),
+  private void reportWarning(Map<JWarning, TypeTree> exceptionTreeByWarning, Map<TypeTree, Type> exceptionTypes) {
+    for (Map.Entry<JWarning, TypeTree> treeByWarning : exceptionTreeByWarning.entrySet()) {
+      TypeTree reportTree = treeByWarning.getValue();
+      reportIssue(reportTree,
+        "Remove this type because it is unreachable, hidden by previous catch block(s).",
```
sebastian-hungerecker-sonarsource, line 92: I believe we should still provide different messages for a single type being unreachable vs. the whole catch block being unreachable. IIRC we introduced that distinction specifically because only talking about types confused some users.

sebastian-hungerecker-sonarsource (changes requested):
Looks good; some changes.

java-checks/src/main/java/org/sonar/java/checks/UnreachableCatchCheck.java
```diff
@@ -48,203 +42,67 @@
...
-          for (Type hiddenType : hiddenTypes) {
-            reportSingleType(hiddenType);
-          }
-        }
-      }
+  private void checkWarnings(TryStatementTree tryStatementTree) {
+    Map<JWarning, TypeTree> exceptionTreeByWarning = new HashMap<>();
+    Map<TypeTree, Type> exceptionTypes = new HashMap<>();
+    for (CatchTree catchTree : tryStatementTree.catches()) {
+      caughtExceptionTypes(catchTree).stream()
```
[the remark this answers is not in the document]
m-g-sonar: I can even simplify the whole loops.

java-checks-test-sources/src/main/java/checks/UnreachableCatchCheck.java
```diff
@@ -13,26 +13,26 @@ void unreachable(boolean cond) {
...
       if (true) throw new ExtendsExtendsCustomException();
-      else throw new ExtendsOtherExtendsCustomException();
+      throw new ExtendsOtherExtendsCustomException();
     } catch (ExtendsExtendsCustomException e) {
       // ...
     } catch (ExtendsOtherExtendsCustomException e) {
-      // ...
-    } catch (ExtendsCustomException|OtherExtendsCustomException e) { // Noncompliant [[sc=7;ec=12;secondary=31,33]] {{Remove or refactor this catch clause because it is unreachable as hidden by previous catch blocks.}}
+      // Noncompliant@+1
+    } catch (ExtendsCustomException | OtherExtendsCustomException e) { // Noncompliant
```
sebastian-hungerecker-sonarsource, line 35: Here we still tell the user to remove all types from the catch clause rather than telling them to just remove the catch clause.

sebastian-hungerecker-sonarsource (changes requested):

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3601, opened 2021-05-20
SONARJAVA-3794 Improve S1905 (Redundant cast) rule precision by relying on compiler warnings

sebastian-hungerecker-sonarsource:
This drops support for lambdas and method references as our previous
implementation caused too many FPs and Eclipse's version of the rule
doesn't support them either.


java-checks/src/main/java/org/sonar/java/checks/RedundantTypeCastCheck.java
```diff
@@ -19,223 +19,69 @@
...
-    int i = 0;
-    do {
-      // typeCastTree is necessarily within the arguments
-      Tree arg = ExpressionUtils.skipParentheses(arguments.get(i));
-      if (typeCastTree.equals(arg)) {
-        return i;
+  private static boolean matchesWarning(JWarning warning, TypeCastTree tree) {
+    SyntaxToken startToken = tree.openParenToken();
+    // When a cast expression is nested inside one or more parenthesized expression, Eclipse raises the warning on
+    // the outermost parenthesized expression rather than the cast expression, so we need to take that into account
```
sebastian-hungerecker-sonarsource, line 76: I assume this also applies to other Eclipse warnings that are raised on expressions and should be taken into account when we implement the logic of mapping warnings to trees.

PR #3603, opened 2021-05-20
SONARJAVA-3847 Rules targeting parameters should work with incomplete semantic

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #3605, opened 2021-05-20
SONARJAVA-3820 Update remediation times and cost

m-g-sonar:


sonarpedia.json
```diff
@@ -3,7 +3,7 @@
   "languages": [
     "JAVA"
   ],
-  "latest-update": "2021-04-26T08:13:38.034417Z",
+  "latest-update": "2021-05-20T20:48:15.570465835Z",
```
quentin-jaquier-sonarsource, line 6: Should we really commit this, despite the fact that we only partially updated the metadata?

quentin-jaquier-sonarsource (approved):
Small comment, otherwise, LGTM

sonarpedia.json
m-g-sonar: indeed. removed

PR #3606, opened 2021-05-21
Rework "MethodTree.isOverriding()" to match the contract in case of unknowns in hierarchy

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
Will be reworked in #3636.

PR #3607, opened 2021-05-21
SONARJAVA-3845 Fix S1144 in case of unknown arguments

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
Looks good to me!

PR #3608, opened 2021-05-21
SONARJAVA-3848: Rules targeting types should work with incomplete semantics

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -75,11 +76,25 @@ public void visitNode(Tree tree) {
...
+  public void scanFile(JavaFileScannerContext context) {
+    visited.clear();
+    super.scanFile(context);
+  }
+
+  @Override
+  // This specific override is meant to ensure that the visited set is cleared even for scans that do not go through scanFile (such as the sanity test).
+  public void leaveFile(JavaFileScannerContext context) {
+    visited.clear();
+    super.leaveFile(context);
```
m-g-sonar, line 89: no reason to call this super method, it does nothing purposely.

java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -75,11 +76,25 @@ public void visitNode(Tree tree) {
...
   }
 
+  @Override
+  public void scanFile(JavaFileScannerContext context) {
+    visited.clear();
+    super.scanFile(context);
+  }
+
+  @Override
+  // This specific override is meant to ensure that the visited set is cleared even for scans that do not go through scanFile (such as the sanity test).
```
m-g-sonar, line 86: no need for the comment, it is a common pattern from many rules: any collections storing trees should be cleared at the end or beginning of processing of a file.

java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -56,6 +56,8 @@
 
   private final Deque<Set<String>> exclusions = new ArrayDeque<>();
 
+  private Set<Tree> visited = new HashSet<>();
```
m-g-sonar, line 60: can be made `final`

m-g-sonar (commented):
Almost there. Good work already!

java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -57,15 +56,20 @@
...
+    // We register on compilation units to clear the visited data structure when scanning a new file
+    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.COMPILATION_UNIT, Tree.Kind.ENUM, Tree.Kind.INTERFACE, Tree.Kind.ANNOTATION_TYPE);
   }
 
   @Override
   public void visitNode(Tree tree) {
+    if (tree.is(Tree.Kind.COMPILATION_UNIT)) {
+      // We clear the visited data structure when entering every file
+      visited.clear();
+    }
```
m-g-sonar, line 72: No! You should return!
If not, you are going to trigger the `exclusionsVisitor` on a way too large scope, and end up scanning everything!
You can also probably get rid of the comment, the body of the if is self-explanatory.

```java
   if (tree.is(Tree.Kind.COMPILATION_UNIT)) {
      visited.clear();
      return;
   }

File: java-checks/src/test/java/org/sonar/java/checks/AbstractClassNoFieldShouldBeInterfaceCheckTest.java [hunk 5]
```diff
@@ -23,25 +23,30 @@
 import org.junit.jupiter.api.Test;
 import org.sonar.java.checks.verifier.JavaCheckVerifier;
 
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 import static org.junit.jupiter.api.Assertions.assertFalse;
```
m-g-sonar, line 27: `assertTrue` and `assertFalse` are not used anymore and the imports should be dropped.

m-g-sonar (changes requested):

java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -56,26 +56,37 @@
...
     }
     exclusions.push(currentClassExclusions);
     tree.accept(new DeprecatedTypeVisitor());
   }
 
+
   @Override
   public void leaveNode(Tree tree) {
-    exclusions.pop();
+    if (!exclusions.isEmpty()) {
```
m-g-sonar, line 87: nop, sorry I missed this.
`leaveNode()` is called as many times as `visitNode()`, and both methods are going to be called for all the kind we registered to.

For all the nodes but COMPILATION_UNIT, we push something in the stack, so for each of them we should pop(), and not pop() because it's not empty.

I would replace this by:
```java
if(!tree.is(Tree.Kind.COMPILATION_UNIT)) {
  exclusions.pop();
}

PR #3610, opened 2021-05-21
(WIP) Map ECJ warnings to tree nodes

sebastian-hungerecker-sonarsource:
I've made the code compile by making SubscriptionVisitor.scanTree
public, but clearly that's not the right solution.

The tests for redundant casts and self assignments are all passing, so
that's great. To make the self assignment tests pass, I had to change
the expected column numbers, but that seems like a good
thing. Previously ECJ-reported warnings would use the whole assignment
for the location while issues that we found ourselves would only mark
the equals sign - so locations were inconsistent. Now issues are
reported on the equals sign no matter where they come from.

The unreachable catch check doesn't work because I listed
Tree.Kind.CATCH as its tree type, which is wrong. I believe it reports
on the type trees, but there's no kind for that (or rather there's a
lot of kinds for that), so we may have to rethink the approach there.

And unused imports doesn't work because I just commented out the code
to make it compile, so obviously it can't do anything.


m-g-sonar:
Reworked in #3637

PR #3612, opened 2021-05-26
SONARJAVA-3849 Rules targeting control flow should work with incomplete semantics

dorian-burihabwa-sonarsource:
This is mostly a validation that shows that both S1181 and S4838 work fine with incomplete semantics


java-checks/src/test/java/org/sonar/java/checks/CatchOfThrowableOrErrorCheckTest.java
```diff
@@ -22,12 +22,14 @@
 import org.junit.jupiter.api.Test;
 import org.sonar.java.checks.verifier.JavaCheckVerifier;
 
+import static org.sonar.java.checks.verifier.TestUtils.nonCompilingTestSourcesPath;
+
 class CatchOfThrowableOrErrorCheckTest {
```
m-g-sonar, line 27: I feel we are missing the test of the absence of semantic, aren't we?
If I add the following test, it fails with what looks to me like FPs.
```
  @Test
  void testNoSemantic() {
      .onFile(nonCompilingTestSourcesPath("checks/CatchOfThrowableOrErrorCheck.java"))
      .withCheck(new CatchOfThrowableOrErrorCheck())
      .withClassPath(Collections.emptyList())
      .verifyIssues();
  }
```

dorian-burihabwa-sonarsource: Good catch, I will look into that

java-checks/src/main/java/org/sonar/java/checks/CatchOfThrowableOrErrorCheck.java
```diff
@@ -111,7 +111,10 @@ public void visitThrowStatement(ThrowStatementTree tree) {
     private boolean isGuavaCloserRethrow(ExpressionTree expression) {
       if (expression.is(Tree.Kind.METHOD_INVOCATION)) {
         MethodInvocationTree mit = (MethodInvocationTree) expression;
-        if (MATCHERS.matches(mit)) {
+        // In case of broken semantics, the method matcher cannot identify the method invocation.
+        // To avoid raising FPs in these cases, we accept is as is and look at the first argument
+        if (MATCHERS.matches(mit) ||
```
m-g-sonar, line 116: To me, this is the wrong fix. if the invocation is unknown, you can not assume it's equivalent to guava's closer, especially relying only on number of arguments! Without semantic, some issues should simply not be raised, because when methods are unknown, just reporting nothing is way safer. We don't assume anything.

Also, if we don't know what is in the thrown from body of the `try { }`, then reporting an issue on any `catch` is wrong. and I don't see any tests of this.
```java
try {
  unknown();
} catch (Throwable e) { } // should not raise any issue - the user might not have the choice and we don't know it!
```

m-g-sonar (changes requested):

m-g-sonar (approved):

PR #3613, opened 2021-05-26
SONARJAVA-3850: Rules targeting class members should work with incomplete semantics

dorian-burihabwa-sonarsource:


java-checks/src/test/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheckTest.java
```diff
@@ -40,16 +41,16 @@ void test() {
   }
 
   @Test
-  void testWithoutSemantic() {
+  void test_without_semantic() {
     JavaCheckVerifier.newVerifier()
       .onFile(testSourcesPath("checks/serialization/SerializableFieldInSerializableClassCheck.java"))
       .withCheck(new SerializableFieldInSerializableClassCheck())
-      .withoutSemantic()
+      .withClassPath(Collections.emptyList())
```
m-g-sonar, line 48: I'm not sure this is needed. `.withoutSemantic()` is supposed to do exactly this. To me, the test is already valid and the rule operating as expected.

java-checks/src/main/java/org/sonar/java/checks/PublicStaticMutableMembersCheck.java
```diff
@@ -220,6 +220,11 @@ private static boolean isAcceptedTypeOrUnmodifiableMethodCall(MethodInvocationTr
   }
 
   private static boolean isAcceptedType(Type type, List<String> accepted) {
+    // In case of broken semantics, the type is unknown and can therefore not be matched against an accepted one.
+    // To avoid raising FPs, we consider that an unknown type is most likely an accepted one.
+    if (type.isUnknown()) {
```
m-g-sonar, line 225: This is now redundant with line 219.

Why not dropping the `type.isUnknown() ||` part of method `isAcceptedTypeOrUnmodifiableMethodCall()` ? it anyway calls `isAcceptedType(type, ACCEPTED_TYPES)` right after.

m-g-sonar (changes requested):

PR #3617, opened 2021-05-27
SONARJAVA-3818 Add new custom rule example targeting TEST sources

m-g-sonar:


docs/java-custom-rules-example/src/main/resources/org/sonar/l10n/java/rules/java/NoIfStatementInTests.html
```diff
@@ -0,0 +1,27 @@
+<p>This rule is an example of custom rules targeting only TEST code.It is not a common good practice, but a simple illustration of what can be achieved when writing rules.</p>
+<p>In particular, this rules target test methods (method annotated with JUnit4 @Test annotation) which contains a "if" statement, completely arbitrarily. The only embrio of reasoning
```
dorian-burihabwa-sonarsource, line 2:
```suggestion
<p>In particular, this rule targets test methods (method annotated with JUnit4 @Test annotation) that contain an "if" statement, completely arbitrarily. A possible reasoning
```

dorian-burihabwa-sonarsource (changes requested):
Apart from few typos the rule description, the custom rule looks  good to me 👍🏿

PR #3619, opened 2021-05-30
fix nonsense example

Bananeweizen:
.hasSameSize() is used with iterables/arrays as argument, not their
length. Probably a copy/paste error.


quentin-jaquier-sonarsource (approved):

quentin-jaquier-sonarsource:
This change will be merged in #3621.

Thanks!

PR #3620, opened 2021-05-30
S5838 misses some length()/size() > 0 assertions

Bananeweizen:
assertThat(file.length() > 0).isTrue() and similar comparisons of
length/size greater zero should be mapped to isNotEmpty().

This was missing for file, collection and string. The existing optimizations would only lead to size().isPositive().

- [X] Unit tests are passing and you provided a unit test for your fix


quentin-jaquier-sonarsource (approved):
LGTM

quentin-jaquier-sonarsource:
Hello @Bananeweizen 

Indeed, it makes sense to improve the rule to add the support of this case.
I created a Jira ticket to document your change: [SONARJAVA-3858](https://jira.sonarsource.com/browse/SONARJAVA-3858).
Then, I had to duplicate your code in order to trigger the QA, see #3621. The ruling test was failing, I just added the new (correct) issue that this rule now detects to the expected list.

Thanks for taking the time to suggest this change and have a nice day.

Quentin

PR #3621, opened 2021-05-31
SONARJAVA-3858 S5838 should support "length()"/"size()" followed by "isPositive()" simplification

quentin-jaquier-sonarsource:
Duplicates #3619 and #3620 to trigger QA.

### DIFF
--- a/its/ruling/src/test/resources/sonar-server/java-S5838.json
+++ b/its/ruling/src/test/resources/sonar-server/java-S5838.json
@@ -649,6 +649,9 @@
 426,
 439,
 ],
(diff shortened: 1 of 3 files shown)


Bananeweizen:
@quentin-jaquier-sonarsource Just out of curiosity: Did I miss something with my original PRs, such that they did not trigger that bot? Would of course like to do it better next time.

quentin-jaquier-sonarsource:
Concerning the JSON file, this is the result of running the [ruling test](https://github.com/SonarSource/sonar-java#ruling-test), using the modified version of the plugin on a set of a fixed project to see early the effect of the change.

PR #3623, opened 2021-05-31
SONARJAVA-3822: S6073 should not report on method invocation arguments that actually return an argument matcher

dorian-burihabwa-sonarsource:
This pull request adds the capability to the check to look into methods that might return an argument matcher.
If the method cannot be resolved or its code explored, then it is assumed that the method might return an argument matcher and thus no issue is reported.

### DIFF
--- a/java-checks-test-sources/src/main/files/non-compiling/checks/tests/MockitoArgumentMatchersUsedOnAllParameters.java
+++ b/java-checks-test-sources/src/main/files/non-compiling/checks/tests/MockitoArgumentMatchersUsedOnAllParameters.java
@@ -1,0 +1,100 @@
+package checks.tests;
+
+
+import java.util.Collection;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.ArgumentMatchers;
+import org.mockito.Mockito;
+
+import static org.mockito.AdditionalMatchers.gt;
+import static org.mockito.AdditionalMatchers.not;
+import static org.mockito.AdditionalMatchers.or;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.anyList;
+import static org.mockito.ArgumentMatchers.anySet;
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/tests/MockitoArgumentMatchersUsedOnAllParametersCheck.java
```diff
@@ -104,4 +130,30 @@ private static ExpressionTree skipCasts(ExpressionTree tree) {
...
+        .filter(statement -> statement.is(Tree.Kind.RETURN_STATEMENT))
+        .map(ReturnStatementTree.class::cast)
+        .map(ReturnStatementTree::expression)
+        .filter(expression -> expression.is(Tree.Kind.METHOD_INVOCATION))
+        .collect(Collectors.toList());
+      if (terminalMethodInvocations.isEmpty()) {
+        return;
+      }
+      returnsAnArgumentMatcher = terminalMethodInvocations.stream()
+        .allMatch(MockitoArgumentMatchersUsedOnAllParametersCheck::isArgumentMatcherLike);
```
quentin-jaquier-sonarsource, line 155: As discussed, recursively searching for an argument matcher is a bit overkill here. An approximation that seems good enough would be to consider any method call at this point as an argument matcher.

This will also simplify the implementation.

java-checks-test-sources/src/main/files/non-compiling/checks/tests/MockitoArgumentMatchersUsedOnAllParameters.java
```diff
@@ -0,0 +1,100 @@
...
+public class MockitoArgumentMatchersUsedOnAllParameters {
+
+  @Test
+  public void nonCompliant() {
+    verify(foo).bar(
+      returnRawValue(), // Noncompliant [[sc=7;ec=23]] {{Add an "eq()" argument matcher on this parameter.}}
+      any(), any()
+    );
+    verify(foo).bar(
+      returnRawValueThroughLayers(), // Noncompliant [[sc=7;ec=36]] {{Add an "eq()" argument matcher on this parameter.}}
```
quentin-jaquier-sonarsource, line 36: If we apply the change suggested in the other comment, this issue will become an FN. It seems acceptable, I don't expect this to happen a lot.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3624, opened 2021-05-31
SONARJAVA-3825 Drop deprecated methods and classes

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
```diff
@@ -98,16 +98,6 @@ public TypeSymbol returnType() {
...
   }
 
-  @Nullable
-  @Override
-  public MethodSymbol overriddenSymbol() {
-    if (overriddenSymbols == null) {
-      // only compute overridenSymbols once
-      overriddenSymbols();
-    }
-    return firstOverridenSymbol;
```
quentin-jaquier-sonarsource, line 108: `firstOverridenSymbol` is still cached but not used anymore, we can remove this field aswell.

quentin-jaquier-sonarsource (changes requested):

quentin-jaquier-sonarsource (approved):
LGTM

PR #3625, opened 2021-05-31
SONARJAVA-3862 Check for nullness of "isOverriding" in S1448 and AbstractJUnit5NotCompliantModifierChecker

quentin-jaquier-sonarsource:
This PR will not resolve SONARJAVA-3862, but is a necessary prerequisite, as it will help to avoid unwanted NPE when we will rework `MethodTree.isOverriding()`.


dorian-burihabwa-sonarsource (approved):
LGTM!

PR #3626, opened 2021-05-31
SONARJAVA-3856 Handle parenthesis and assignments

m-g-sonar:


java-checks/src/test/files/checks/StringConcatenationInLoopCheck.java
```diff
@@ -58,4 +58,17 @@ static MyObject newInstance() {
...
+      while (true) {
+        (this).f += "sdasd"; // Noncompliant
+
+        (m = this).f += "asdasd"; // not supported... what to do with this?
+        ((M) m).f += "asdas"; // not supported... what to do with this?
+      }
+    }
+  }
+
 }
```
dorian-burihabwa-sonarsource, line 74: Should we move this file to the non-compiling sources folder of `java-checks-test-sources` ?

dorian-burihabwa-sonarsource (changes requested):
Simple and effective, LGTM. But we should probably move the test source file to `java-checks-test-sources`.

java-checks/src/test/files/checks/StringConcatenationInLoopCheck.java
m-g-sonar: Indeed, we should. no reason to not make it clean.

dorian-burihabwa-sonarsource (approved):
Great, thanks for the changes :)

PR #3628, opened 2021-06-01
SONARJAVA-3825 Revert removal of "overriddenSymbol"

quentin-jaquier-sonarsource:


pom.xml
```diff
@@ -98,6 +98,8 @@
 
     <!-- FIXME: Overrides OSS parent pom for java 16 compatibility -->
     <version.artifactory.plugin>3.2.1</version.artifactory.plugin>
+
+    <sonar.skip>true</sonar.skip>
```
dorian-burihabwa-sonarsource, line 102: Is this a temporary change? If so, it might be worth leaving a meaningful comment next to it.

dorian-burihabwa-sonarsource (approved):
LGTM but we shouldn't forget to revert the skip later

pom.xml
quentin-jaquier-sonarsource: Yes, but I think it's fine as it is, I will directly create a PR to not forget it.

PR #3636, opened 2021-06-04
SONARJAVA-3862 Rework "MethodTree.isOverriding()" to match the contract in case of unknowns in hierarchy

quentin-jaquier-sonarsource:
Notes:
- the rework fixed some known FP in unit tests in case of incomplete semantic
- the rework removed many issues in jboss, a project misconfigured on purpose (missing binaries). This is expected, we remove potential FPs.
- the last commit reworks the logic in the case where `isOverriding` returns null in some rules, to ensure the rule takes the path that does not report an issue in such cases.


java-frontend/src/main/java/org/sonar/java/model/declaration/MethodTreeImpl.java
```diff
@@ -272,17 +281,29 @@ public CFG cfg() {
...
+      isOverriding = true;
+    } else {
+      Symbol.MethodSymbol symbol = symbol();
+      List<Symbol.MethodSymbol> overriddenSymbols = symbol.overriddenSymbols();
+      if (overriddenSymbols.isEmpty()) {
+        if (hasUnknownTypeInHierarchy(symbol)) {
+          isOverriding = null;
+        } else {
+          isOverriding = false;
+        }
```
dorian-burihabwa-sonarsource, line 301:
```suggestion
      if (overriddenSymbols.isEmpty() && !hasUnknownTypeInHierarchy(symbol)) {
        isOverriding = false;
```

dorian-burihabwa-sonarsource (changes requested):
We could use a few changes around typos, test file locations and a test logic in `isoverriding`

java-frontend/src/main/java/org/sonar/java/model/declaration/MethodTreeImpl.java
quentin-jaquier-sonarsource: Not sure about this one, the value null has a meaning in this context. Am I missing something?

java-checks/src/test/files/checks/UndocumentedApiCheck/UndocumentedApi.java
```diff
@@ -94,20 +94,20 @@ protected void doSomething() {                 // Compliant - not public
    * Valid descriptions.
```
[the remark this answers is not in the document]
quentin-jaquier-sonarsource: Probably yes.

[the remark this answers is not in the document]
quentin-jaquier-sonarsource: In fact, we should ideally move the files to the compiled module, but it's quite a big change, I'm afraid it will pollute the diff of this PR which is already non-trivial. I suggest to not move the files in this PR.

java-frontend/src/main/java/org/sonar/java/model/declaration/MethodTreeImpl.java
dorian-burihabwa-sonarsource: The value is indeed null by default. My issue is with the number of nested branches that can be simplified either by changing the test in the if statement or doing early returns. WDYT?

dorian-burihabwa-sonarsource (approved):

PR #3637, opened 2021-06-04
SONARJAVA-3860 Map ECJ Warnings with our Trees

m-g-sonar:


quentin-jaquier-sonarsource (commented):
Except for the coverage that we can slightly improve (`isMorePreciseTree` never returns false), this looks good, nice work!

quentin-jaquier-sonarsource (approved):
LGTM, good job

PR #3638, opened 2021-06-07
Add whitesource to build pipeline

dorian-burihabwa-sonarsource:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -61,6 +61,30 @@ build_task:
     - ./check-license-compliance.sh
   cleanup_before_cache_script: cleanup_maven_repository
 
+ws_scan_task:
+  depends_on:
+    - build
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -81,6 +105,7 @@ qa_os_win_task:
   cleanup_before_cache_script: cleanup_maven_repository
 
 plugin_qa_task:
+  only_if: $CIRRUS_BRANCH == "master" || $CIRRUS_BRANCH =~ "branch-.*"
```
saberduck, line 108: why this addition?

saberduck (approved):

.cirrus.yml
dorian-burihabwa-sonarsource: Good catch. That line should not be there. The change should now be reverted.

PR #3648, opened 2021-06-14
SONARJAVA-3873 Order rules based on execution time to make the best of issue streaming

alban-auzeill:


alban-auzeill:
Erroneous PR not ordering checks because sonarComponents.checkClasses() shuffled the checks

PR #3653, opened 2021-06-17
SONARJAVA-3738 Upgrade ECJ to 3.26.0 and update parser to support Java 16 final features

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/JParser.java
```diff
@@ -2082,8 +2081,18 @@ private ExpressionTree createExpression(Expression node) {
...
+          convertExpression(e.getLeftOperand())
+        );
+      }
+      case ASTNode.PATTERN_INSTANCEOF_EXPRESSION: {
+        PatternInstanceofExpression e = (PatternInstanceofExpression) node;
+        SingleVariableDeclaration v = e.getRightOperand();
+        return new InstanceOfTreeImpl(
+          firstTokenAfter(e.getLeftOperand(), TerminalTokens.TokenNameinstanceof),
+          convertType(v.getType()),
+          convertSimpleName(v.getName())
```
alban-auzeill, line 2095: We should also ensure that `InstanceOfTree#patternVariable().symbol().declaration()` matches this `InstanceOfTree`

alban-auzeill (changes requested):

java-frontend/src/main/java/org/sonar/java/model/JParser.java
m-g-sonar: I introduced a new KIND (`PATTERN_INSTANCE_OF`) and Tree interface (`PatternInstanceOfTree`) to do that.

PR #3656, opened 2021-06-21
SONARJAVA-3799 Make sure we visit record's componnent

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/model/declaration/ClassTreeImplTest.java
```diff
@@ -51,10 +49,38 @@ void getLine() {
...
+      "record A "
+        // components
+        + "(Object o, String s) {\n"
+        // compact constructor
+        + "  A { }\n"
+        + "}").types().get(0);
+
+    assertThat(classTree.declarationKeyword()).is("record");
+
+    String[] componentsVariableNames = {"", ""};
```
alban-auzeill, line 72: I would use `List<String>` and `add(...)` instead of an array and `i++`

alban-auzeill:
This PR only contains test code because the problem has been already fixed.

alban-auzeill (approved):

m-g-sonar:
The original fix was done when bringing support to Java 16, in this [commit](https://github.com/SonarSource/sonar-java/commit/fbbe0bd54f66a7ac2c807d3cffe4300cefd2c9ba#diff-98677bbfec6e3318e3b52cb56a9ff71e4c291fb13a70c9f7517864415c02c1caR66).

PR #3658, opened 2021-06-22
SONARJAVA-3775 Rule S6220: Functional interfaces should not be sealed

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/SealedFunctionalInterfaceCheck.java
```diff
@@ -0,0 +1,60 @@
...
+      .filter(modifierTree -> isSealedModifierOnFunctionalInterface(modifierTree, classTree))
+      .forEach(sealedModifierTree -> reportIssue(sealedModifierTree, MESSAGE));
+  }
+
+  private static boolean isSealedModifierOnFunctionalInterface(ModifierKeywordTree modifierTree, ClassTree classTree) {
+    if (modifierTree.modifier() != Modifier.SEALED || !classTree.superInterfaces().isEmpty()) {
+      return false;
+    }
+    List<Tree> members = classTree.members();
+    return members.size() == 1 &&
```
sebastian-hungerecker-sonarsource, line 55: A functional interface can have default methods in addition to its one abstract method. So `members.size()` doesn't have to be 1.

sebastian-hungerecker-sonarsource (changes requested):
I want to challenge this rule. To me it seems perfectly legit to have a sealed interface with only one member. For example I think there's absolutely nothing wrong with the below code:

```java
public sealed interface AstNode permits ExpressionNode, StatementNode, DeclarationNode, TerminalNode {
    Location location();
}
```

I have written code like that in other languages many times and I'd have been annoyed if a tool complained about it. Sure, I can't use lambdas to implement that interface, but in that example using a lambda to implement an AstNode wouldn't make any sense.

In other words, I don't think it makes sense to assume that an interface is intended to be used as a functional interface just because it meets the requirements.

I originally wanted to suggest to only raise this issue if the interface is annotated with `@FunctionalInterface`, but it turns out that trying to annotate a sealed interface with `@FunctionalInterface` is a compilation error anyway. So I would suggest abandoning the rule or at least not making it part of a Sonar Way.

alban-auzeill:
We decided to drop this rule. We do not trust anymore that a user could accidentally add the `sealed` keyword to a functional interface. First, the `sealed` keyword alone does not compile, you need also to specify all allowed derived interfaces after the `permits` keyword. Then mixing `@FunctionalInterface` and `sealed` does not compile. So in the end the number of possible false positives could be greater than the number of true findings and the value of the rule is too low to select this strategy.

PR #3659, opened 2021-06-22
Clean JParser by reducing complexity, extract method and simplify some implementation classes

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/JParser.java
```diff
@@ -977,12 +960,8 @@ private int processFieldDeclaration(FieldDeclaration fieldDeclaration, List<Tree
...
   }
 
-  private ArgumentListTreeImpl convertArguments(
-    InternalSyntaxToken openParen,
-    List<?> list,
-    InternalSyntaxToken closeParen
-  ) {
-    ArgumentListTreeImpl arguments = new ArgumentListTreeImpl(new ArrayList<>(), new ArrayList<>()).complete(openParen, closeParen);
+  private ArgumentListTreeImpl convertArguments(@Nullable InternalSyntaxToken openParen, List<?> list, @Nullable InternalSyntaxToken closeParen) {
+    ArgumentListTreeImpl arguments = ArgumentListTreeImpl.emptyList().complete(openParen, closeParen);
```
alban-auzeill, line 964: I would expect that every call to `complete(openParen, closeParen)` is only done at the end of `ArgumentListTreeImpl` construction. But here there's `arguments.add` after.

alban-auzeill (approved):

java-frontend/src/main/java/org/sonar/java/model/JParser.java
m-g-sonar: I won't change this for the time being.

PR #3660, opened 2021-06-22
SONARJAVA-3739 Implement rule S6201: Use Pattern Matching on instanceof to substitute instanceof + cast

sebastian-hungerecker-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/InstanceOfPatternMatchingCheck.java
```diff
@@ -0,0 +1,133 @@
...
+        IfStatementTree ifStatement = (IfStatementTree) tree;
+        handleConditional(ifStatement.condition(), ifStatement.thenStatement());
+        break;
+      case CONDITIONAL_AND:
+        BinaryExpressionTree and = (BinaryExpressionTree) tree;
+        handleConditional(and.leftOperand(), and.rightOperand());
+        break;
+      default: // CONDITIONAL_EXPRESSION
+        ConditionalExpressionTree conditional = (ConditionalExpressionTree) tree;
+        handleConditional(conditional.condition(), conditional.trueExpression());
```
m-g-sonar, line 70: I feel we need to check the `elseExpression` depending of the condition. This works:
```
int size = !(o instanceof String) ? 42 : ((String) o).length();
```
and can be written
```
int size = !(o instanceof String s) ? 42 : s.length();
```

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -1099,6 +1103,11 @@ private void buildInstanceOf(InstanceOfTree instanceOfTree) {
     build(instanceOfTree.expression());
   }
 
+  private void buildInstanceOf(PatternInstanceOfTree instanceOfTree) {
+    currentBlock.elements.add(instanceOfTree);
+    build(instanceOfTree.expression());
```
m-g-sonar, line 1108: what about the variable declaration?
```suggestion
    build(instanceOfTree.expression());
    build(instanceOfTree.variable());
```

java-frontend/src/main/java/org/sonar/java/model/SyntacticEquivalence.java
```diff
@@ -130,13 +141,26 @@ private static boolean areLeafsEquivalent(JavaTree leftNode, JavaTree rightNode)
     }
   }
 
+  private static boolean areDifferentVariables(JavaTree leftNode, JavaTree rightNode) {
```
m-g-sonar, line 144: 👍🏻

java-frontend/src/main/java/org/sonar/plugins/java/api/JavaVersion.java
```diff
@@ -91,6 +91,14 @@
    */
   boolean isJava15Compatible();
 
+  /**
+   * Test if java version of the project is greater than or equal to 16.
+   * Remark - Contrary to other isJava*Compatible methods, this one will NOT return true if version is not set
+   * @return true if java version used is >= 16
+   * @since SonarJava 7.1: Support of Java 16
+   */
+  boolean isJava16Compatible();
```
m-g-sonar, line 100: 👍🏻

m-g-sonar (commented):
Good work! However, some cases are not supported yet!

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
sebastian-hungerecker-sonarsource: @m-g-sonar Do we do anything with variable declarations (that don't have an initializer) in the CFG?

sebastian-hungerecker-sonarsource: To answer my own question: When a variable declaration without an initializer is added to the CFG, the SE engine uses this to mark the declared variables as null. So since an `instance of`-declared variable will not be null, we should not add the variable declaration to the CFG. If anything, we should make the SE engine generate a not-null constraint when it sees an instance-of pattern, but that's outside of the scope of this PR.

java-checks/src/main/java/org/sonar/java/checks/InstanceOfPatternMatchingCheck.java
```diff
@@ -0,0 +1,133 @@
...
+import org.sonar.plugins.java.api.tree.InstanceOfTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.TypeCastTree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = "S6201")
+public class InstanceOfPatternMatchingCheck extends IssuableSubscriptionVisitor implements JavaVersionAwareVisitor {
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Arrays.asList(Tree.Kind.IF_STATEMENT, Tree.Kind.CONDITIONAL_AND, Tree.Kind.CONDITIONAL_EXPRESSION);
```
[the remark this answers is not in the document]
sebastian-hungerecker-sonarsource: I did add cases for the loop (as well as OR since we now also support negated `instanceof`s), but I didn't end up going bottom-up. I feel like that could have ended being inefficient for deeply nested expressions - especially ones containing multiple `instanceof` checks.

java-checks/src/main/java/org/sonar/java/checks/InstanceOfPatternMatchingCheck.java
sebastian-hungerecker-sonarsource: Done.

PR #3662, opened 2021-06-23
SONARJAVA-3874:  FP in S1168 when using classes with the same unqualified name as collections

dorian-burihabwa-sonarsource:
FP in S1168 when using classes with the same unqualified name as collections


java-checks/src/main/java/org/sonar/java/checks/ReturnEmptyArrayNotNullCheck.java
```diff
@@ -46,49 +44,10 @@
...
-    "PriorityQueue",
-    "RoleList",
-    "RoleUnresolvedList",
-    "Stack",
-    "SynchronousQueue",
-    "TreeSet",
-    "Vector");
+  private static final List<String> COLLECTION_TYPES = Arrays.asList(
+    "java.util.Collection",
+    "javax.print.attribute.standard.JobStateReason"
```
alban-auzeill, line 49: Remove `JobStateReason`, it is not a collection. `JobStateReasons` is, but it also inherits from Collection.
So in the end, we can simplify to only support "java.util.Collection"

java-checks/src/main/java/org/sonar/java/checks/ReturnEmptyArrayNotNullCheck.java
```diff
@@ -121,7 +80,11 @@ private static boolean isCollection(Tree methodReturnType) {
       } else if (methodReturnType.is(Tree.Kind.MEMBER_SELECT)) {
         identifierTree = ((MemberSelectExpressionTree) methodReturnType).identifier();
       }
-      return identifierTree != null && COLLECTION_TYPES.contains(identifierTree.name());
+      if (identifierTree == null) {
+        return false;
+      }
+      Type type = identifierTree.symbol().type();
+      return COLLECTION_TYPES.stream().anyMatch(type::isSubtypeOf);
```
alban-auzeill, line 87: Proposal:
```
return identifierTree != null && identifierTree.symbolType().isSubtypeOf("java.util.Collection");
```

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #3663, opened 2021-06-23
SONARJAVA-3740 Extend rule S1481 to report on unused variables in pattern matching on instanceof

sebastian-hungerecker-sonarsource:


its/ruling/src/test/resources/eclipse-jetty/java-S6212.json
```diff
@@ -71,6 +71,10 @@
 590,
 592,
 599,
+842,
```
alban-auzeill, line 74: This looks like an FP because `i` is used:
```
for (int i = 0; i < __preprepared.length; i++)
```

its/ruling/src/test/resources/eclipse-jetty/java-S6212.json
```diff
@@ -71,6 +71,10 @@
...
+844,
```
alban-auzeill, line 75: Same FP here, `code` is used:
```
            HttpStatus.Code code = HttpStatus.getCode(i);
            if (code == null)
                continue;
            String reason = code.getMessage();
```

alban-auzeill (changes requested):

its/ruling/src/test/resources/eclipse-jetty/java-S6212.json
dorian-burihabwa-sonarsource: I guess the issue here is with S6212 that suggests using `var` on the local variable, even though the PR does not touch that rule

its/ruling/src/test/resources/eclipse-jetty/java-S6212.json
dorian-burihabwa-sonarsource: Similar issue here

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -19,128 +19,99 @@
...
     }
   }
 
-  private void addAssignment(ExpressionTree variable) {
-    if (variable.is(Tree.Kind.IDENTIFIER)) {
-      addAssignment((IdentifierTree) variable);
-    }
+  private boolean isProperLocalVariable(VariableTree variable) {
+    return isLocalVariable(variable.symbol())
+      && !variable.parent().is(Tree.Kind.METHOD, Tree.Kind.CONSTRUCTOR, Tree.Kind.LAMBDA_EXPRESSION, Tree.Kind.CATCH)
```
dorian-burihabwa-sonarsource, line 109: By checking whether the parent is none of this types, we ensure that we target variables that are unused and can be effectively removed.
My question is (and I am not sure it is possible): Is there a way to express the types that **the parent could be** in a  concise way?
I think that would help with understanding  what kind of changes the rule can suggest

dorian-burihabwa-sonarsource (commented):
The rule change looks good to me. The FPs that arise for rules S4174 and S6212 indeed show us that the assumptions on what a local variable need to be adjusted on a per rule basis and these checks should be fixed in the future.

As for the changes in semantics, it does look good to me but another set of eyes could help

sebastian-hungerecker-sonarsource:
> The FPs that arise for rules S4174 and S6212

Those aren't FPs, they're fixed FNs. There's no reason why S4174 and S6212 shouldn't apply to local variables defined in initializer blocks. They just didn't because `isLocalVariable` used to return the wrong answer in that case.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
sebastian-hungerecker-sonarsource: I will rewrite this part to `&& !isParameter(variable.symbol()) && !isDefinedInTryCatch(variable.symbol())` using `JUtils.isParameter` and defining `isDefinedInTryCatch` as a private helper. That should make it easier to follow and also also be more robust in detecting parameters.

its/ruling/src/test/resources/eclipse-jetty/java-S6212.json
alban-auzeill: My Bad, I confused S6212 result with S1481. It's not a S1481 false-positive.

its/ruling/src/test/resources/eclipse-jetty/java-S6212.json
alban-auzeill: I confused S6212 result with S1481. It's not a S1481 false-positive.

java-frontend/src/main/java/org/sonar/java/model/JInitializerBlockSymbol.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.Nullable;
+import org.sonar.plugins.java.api.semantic.Symbol;
+import org.sonar.plugins.java.api.semantic.SymbolMetadata;
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+
+public class JInitializerBlockSymbol implements Symbol.MethodSymbol {
```
m-g-sonar, line 31: to be consistent with other `JSymbol` class, I would make the class `final` and package-visible only.
I feel it would also make sense to write a little javadoc explaining WHY we need it.

java-frontend/src/main/java/org/sonar/java/model/JInitializerBlockSymbol.java
```diff
@@ -0,0 +1,185 @@
...
+  private final TypeSymbol owner;
+  private final boolean isStatic;
+
+  public JInitializerBlockSymbol(TypeSymbol owner, boolean isStatic) {
+    this.owner = owner;
+    this.isStatic = isStatic;
+  }
+
+  @Override
+  public MethodTree declaration() {
```
m-g-sonar, line 41: I'm a bit annoyed by this. I would like to have a link to its declaration, especially if it exists.
What about implementing _only_ the `Symbol` interface instead of `MethodSymbol`?

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -19,128 +19,99 @@
...
+  /**
+   * An identifier is being used as an r-value if it is not used as the left operand of an assignment nor as the
+   * operand of a stand-alone increment
+   */
+  private static boolean isRValue(IdentifierTree tree) {
+    if (tree.parent() instanceof AssignmentExpressionTree) {
+      AssignmentExpressionTree assignment = (AssignmentExpressionTree) tree.parent();
+      return assignment.variable() != tree;
     }
+    return !(tree.parent().is(INCREMENT_KINDS) && tree.parent().parent().is(Tree.Kind.EXPRESSION_STATEMENT));
```
m-g-sonar, line 98: what about parenthesized expressions in parent? we can probably s kip them.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -19,128 +19,99 @@
...
-      Symbol symbol = variableTree.symbol();
-      if (symbol.usages().size() == assignments.getOrDefault(symbol, Collections.emptyList()).size()) {
-        reportIssue(variableTree.simpleName(), "Remove this unused \"" + symbol.name() + "\" local variable.");
-      }
+  /**
+   * An identifier is being used as an r-value if it is not used as the left operand of an assignment nor as the
+   * operand of a stand-alone increment
+   */
+  private static boolean isRValue(IdentifierTree tree) {
+    if (tree.parent() instanceof AssignmentExpressionTree) {
```
m-g-sonar, line 94: I would extract `tree.parent()` in a variable instead of calling it 4 times.

java-frontend/src/main/java/org/sonar/java/model/JInitializerBlockSymbol.java
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  @Override
+  public List<IdentifierTree> usages() {
+    return Collections.emptyList();
+  }
+
+  @Override
+  public String name() {
+    return isStatic ? "<clinit>" : "<init>";
```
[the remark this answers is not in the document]
sebastian-hungerecker-sonarsource: I've added " (initialize block)" at the end of both class and instance initializer names.

java-frontend/src/main/java/org/sonar/java/model/JInitializerBlockSymbol.java
sebastian-hungerecker-sonarsource: Done.

java-frontend/src/main/java/org/sonar/java/model/JInitializerBlockSymbol.java
sebastian-hungerecker-sonarsource: I've thought about it, but left it as-is. If the class didn't implement `MethodSymbol`, then any code doing "if `isMethodSymbol()`, then cast to `MethodSymbol`" would break - and we do that a lot. The fix would be to make `isMethodSymbol()` return `false` and introduce `isInitializerBlockSymbol()` which would return `true`. That might be better, but would mean introducing a new method to the API.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
sebastian-hungerecker-sonarsource: Done

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
sebastian-hungerecker-sonarsource: Done

PR #3672, opened 2021-06-30
SONARJAVA-3745 Implement rule S6204: Use Stream.toList() instead of collectors

sebastian-hungerecker-sonarsource:


dorian-burihabwa-sonarsource (changes requested):
There is a minor issue left is the type used to store the variable when retrieving symbols but apart from that this PR looks good to me 👍🏿

dorian-burihabwa-sonarsource (approved):

PR #3673, opened 2021-07-01
SONARJAVA-3768: S6216: Reflection should not be used to update record's field value

dorian-burihabwa-sonarsource:


java-checks-test-sources/src/main/java/checks/AccessibilityChangeCheckRecord.java
```diff
@@ -0,0 +1,71 @@
...
+
+    void modifyAccessibilityOfField() {
+      setAccessible(true); // Noncompliant
+      this.setAccessible(true); // Noncompliant
+    }
+  }
+
+  record Person(String name, int age) {
+    void doSomething() throws NoSuchFieldException, IllegalAccessException {
+      this.getClass().getField("name").setAccessible(true); // Compliant because reported by S6216
```
sebastian-hungerecker-sonarsource, line 25: What happens if the `this` is removed here?

java-checks/src/main/java/org/sonar/java/checks/AbstractAccessibilityChangeChecker.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+    VariableTree variable = (VariableTree) declaration;
+    ExpressionTree initializer = variable.initializer();
+    if (initializer == null || !initializer.is(Tree.Kind.METHOD_INVOCATION)) {
+      return Optional.empty();
+    }
+    return Optional.of((MethodInvocationTree) initializer);
+  }
+
+  private static Optional<IdentifierTree> getClassIdentifier(MemberSelectExpressionTree memberSelect) {
```
sebastian-hungerecker-sonarsource, line 130: I kind of want to replace this entire method with simply checking with simply getting the symbol type of the expression that `getField` is called on and asking it whether its type argument inherits `Record`. This wouldn't work in the case where the `Class` object gets passed around as `Class<? extends Object>`, but the current approach doesn't work in that case either (indeed the current way doesn't work in any case where the `Class` object is being passed around).

Sadly, type symbols don't currently know their type arguments, so it's not that simple. The ECJ API does have that information though, so we might be able to add it to `TypeSymbol` without too much trouble.

sebastian-hungerecker-sonarsource (changes requested):
Good work, I've requested some changes (and one rather vague "Wouldn't it be nice if" comment, which we can discuss further on Monday).

java-checks/src/main/java/org/sonar/java/checks/AccessibilityChangeOnRecordsCheck.java
```diff
@@ -38,36 +37,45 @@
...
+  public void leaveNode(Tree tree) {
+    if (tree.is(Tree.Kind.COMPILATION_UNIT)) {
+      for (Map.Entry<Symbol, MethodInvocationTree> entry : primaryTargets.entrySet()) {
+        Symbol symbol = entry.getKey();
+        MethodInvocationTree setInvocation = entry.getValue();
+        List<JavaFileScannerContext.Location> secondaries = secondaryTargets.getOrDefault(symbol, Collections.emptyList())
+          .stream()
+          .map(mit -> new JavaFileScannerContext.Location(SECONDARY_MESSAGE, mit))
+          .collect(Collectors.toList());
+        if (secondaries.isEmpty()) {
```
sebastian-hungerecker-sonarsource, line 71: This `if` doesn't seem necessary. You can use `reportIssue(setInvocation, MESSAGE, secondaries, null)` regardless of whether the list is empty or not.

java-checks/src/main/java/org/sonar/java/checks/AbstractAccessibilityChangeChecker.java
```diff
@@ -145,4 +151,17 @@ protected static boolean isModifyingFieldFromRecord(MethodInvocationTree mit) {
...
     return Optional.of((IdentifierTree) expression);
   }
+
+  private static boolean identifierPointsToRecord(IdentifierTree classIdentifier) {
+    if (classIdentifier.name().equals("getClass") &&
+      classIdentifier.parent().is(Tree.Kind.METHOD_INVOCATION) &&
+      GET_CLASS_MATCHER.matches((MethodInvocationTree) classIdentifier.parent())) {
+      Tree walker = classIdentifier;
+      while (!walker.is(Tree.Kind.METHOD, Tree.Kind.CONSTRUCTOR)) {
+        walker = walker.parent();
```
sebastian-hungerecker-sonarsource, line 161: I'm a bit worried about accessing the parent in a loop. This could lead to performance issues for deeply nested code.

sebastian-hungerecker-sonarsource (changes requested):
Looks good, one small case of redundant code. I'd also still like to look into getting rid of `getClassIdentifier` (and by extension the new `identifierPointsToRecord`) in favor of getting that information from the semantics.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3674, opened 2021-07-01
SONARJAVA-3746 Extend rule S2201 to support 'Stream' non-void terminal methods

sebastian-hungerecker-sonarsource:


dorian-burihabwa-sonarsource (approved):
Looks good to me 👍🏿

PR #3675, opened 2021-07-02
SONARJAVA-3771 S6218: Equals should be overridden in the record with array fields

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/MissingOverridesInRecordWithArrayComponentCheck.java
```diff
@@ -0,0 +1,125 @@
...
+    .ofAnyType()
+    .names("toString")
+    .addWithoutParametersMatcher()
+    .build();
+
+  private static final String MESSAGE_TEMPLATE = "Override %s to consider array's content in the method";
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return
```
sebastian-hungerecker-sonarsource, line 57: Remove this line break

java-checks/src/main/java/org/sonar/java/checks/MissingOverridesInRecordWithArrayComponentCheck.java
```diff
@@ -0,0 +1,125 @@
...
+    ClassTree targetRecord = (ClassTree) tree;
+
+    boolean recordHasArrayComponent = targetRecord.recordComponents().stream()
+      .anyMatch(component -> component.symbol().type().isArray());
+    if (!recordHasArrayComponent) {
+      return;
+    }
+
+    Optional<String> message = inspectRecord(targetRecord);
+    if (message.isPresent()) {
```
sebastian-hungerecker-sonarsource, line 72: `ifPresent`

java-checks/src/main/java/org/sonar/java/checks/MissingOverridesInRecordWithArrayComponentCheck.java
```diff
@@ -0,0 +1,125 @@
...
+      } else if (HASH_CODE_MATCHER.matches(method)) {
+        hashCodeIsOverridden = true;
+      } else if (TO_STRING_MATCHER.matches(method)) {
+        toStringIsOverridden = true;
+      }
+    }
+    return Optional.ofNullable(computeMessage(equalsIsOverridden, hashCodeIsOverridden, toStringIsOverridden));
+  }
+
+  private static String computeMessage(boolean equalsIsOverridden, boolean hashCodeIsOverridden, boolean toStringIsOverridden) {
```
sebastian-hungerecker-sonarsource, line 97: This should either have a `@CheckForNull` annotation or return `Optional`. Since `inspectRecord` returns an `Optional`, it makes sense that this one should as well (or alternatively both should return a nullable). That way you can get rid of `Optional.ofNullable` as well.

sebastian-hungerecker-sonarsource (changes requested):
Looks good, some minor style comments.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3676, opened 2021-07-05
 SONARJAVA-3760 Improve rule S2094: 'Classes should not be empty' to support Records

sebastian-hungerecker-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM!

PR #3678, opened 2021-07-06
SONARJAVA-3773 Implement rule S6219: Don't set 'serialVersionUID' to '0L' in records

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/serialization/SerialVersionUidInRecordCheck.java
```diff
@@ -0,0 +1,76 @@
...
+  private static boolean isSerialVersionUIDField(VariableTree variable) {
+    Symbol symbol = variable.symbol();
+    return symbol.isFinal() &&
+      symbol.type().is("long") &&
+      symbol.name().equals("serialVersionUID");
+  }
+
+  private static boolean setsTheValueToZero(VariableTree variable) {
+    ExpressionTree initializer = variable.initializer();
+    if (initializer == null || !initializer.is(Tree.Kind.LONG_LITERAL)) {
```
sebastian-hungerecker-sonarsource, line 67: I would remove the check for `LONG_LITERAL` here. If the initializer is some other constant expression, like the name of another `static final long` in the same file, there's no reason we shouldn't still support it (unless I'm missing something). And if it's not, `asConstant` is going to return false anyway.

Side benefit: after removing this check, the check for `aLong.isPresent` can actually be false, so you'll increase test coverage.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerialVersionUidInRecordCheck.java
```diff
@@ -0,0 +1,76 @@
...
+      return false;
+    }
+    Optional<Long> aLong = initializer.asConstant(Long.class);
```
sebastian-hungerecker-sonarsource, line 70: I would also check for an integer constant here as nothing is stopping users from writing `static final long serialVersionUID = 0;`.

java-checks-test-sources/src/main/java/checks/serialization/SerialVersionUidInRecordCheck.java
```diff
@@ -0,0 +1,49 @@
...
+    @Serial private static final long serialVersionUID = getConstantValue(); // Compliant as this is derived from a method call
+
+    private static long getConstantValue() {
+      return 0L;
+    }
+  }
+
+  record FieldWithValueFromStaticVariable(String name, int age) implements Serializable {
+    static long DEFAULT_VALUE = 0L;
+    @Serial private static final long serialVersionUID = DEFAULT_VALUE; // Compliant as this is derived from a static variable
```
sebastian-hungerecker-sonarsource, line 47: Is this meant to be an FN / limitation of the rule? Otherwise I don't understand why this is compliant.

sebastian-hungerecker-sonarsource (changes requested):
Looks good, some improvements.

java-checks-test-sources/src/main/java/checks/serialization/SerialVersionUidInRecordCheck.java
dorian-burihabwa-sonarsource: Correct it should be marked as a false negative.

java-checks/src/main/java/org/sonar/java/checks/serialization/SerialVersionUidInRecordCheck.java
dorian-burihabwa-sonarsource: The issue is that I am not sure if this does not encourage [setting the value blindly](https://rules.sonarsource.com/java/tag/serialization/RSPEC-4926).

Basically, we can recover the value from:
1. A static variable
2. A static method

I can extend the current version to cover the first one but I am not sure I want to do the second one. WDYT?

sebastian-hungerecker-sonarsource: I don't think I understand your point about setting the value blindly. If we raise an issue in more cases, then, if anything, that'd discourage setting the value, no?

And regarding static methods: I'm just talking about removing the `.is` check and supporting whatever constant constructs `asConstant` already supports natively. Not supporting static methods is fine.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3680, opened 2021-07-06
SONARJAVA-3748 RSPEC-6206 Records should be used instead of ordinary classes when representing immutable data structure

m-g-sonar:


java-checks-test-sources/src/main/java/checks/RecordInsteadOfClassCheck.java
```diff
@@ -0,0 +1,101 @@
...
+    int getBase() { return base; }
+  }
+
+
+  class ClassWithSetter {
+    private final int sum;
+
+    ClassWithSetter(int sum) { this.sum = sum; }
+    int getSum() { return sum; }
+    void setSum(int value) { }
```
sebastian-hungerecker-sonarsource, line 67: I'm not sure it makes sense to me that we make an exception if and only if there's a method that exactly matches the typical signature for a setter. Is this about the possibility of using reflection to set the field? I don't see the use case where a user would define the fields as final, then define normal-looking setter methods and use reflection inside them to get rid of the finality.

java-checks/src/main/java/org/sonar/java/checks/RecordInsteadOfClassCheck.java
```diff
@@ -0,0 +1,213 @@
...
+      .filter(m -> isGetter(m, fieldNames))
+      .map(Symbol::name)
+      .map(RecordInsteadOfClassCheck::toFieldName)
+      .collect(Collectors.toSet());
+    return gettersForField.containsAll(fieldNames);
+  }
+
+  private static boolean isGetter(Symbol.MethodSymbol method, Set<String> fieldNames) {
+    String methodName = method.name();
+    return (methodName.startsWith("get") || methodName.startsWith("is"))
```
sebastian-hungerecker-sonarsource, line 152: I feel like we should also support getters whose name is equal to that of the field since a) that's a somewhat common way to name getters and b) that's what the getters will actually be called after converting to a record.

java-checks/src/main/java/org/sonar/java/checks/RecordInsteadOfClassCheck.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  private static boolean hasParameterForEveryField(Symbol.MethodSymbol constructor, Set<String> fieldNames) {
+    Set<String> parameterNames = constructor.declaration()
+      .parameters()
+      .stream()
+      .map(VariableTree::simpleName)
+      .map(IdentifierTree::name)
+      .collect(Collectors.toSet());
+    return parameterNames.size() == fieldNames.size() && parameterNames.containsAll(fieldNames);
```
sebastian-hungerecker-sonarsource, line 173: return parameterNames.equals(fieldNames);

sebastian-hungerecker-sonarsource (changes requested):
Looks good, some changes/questions

java-checks-test-sources/src/main/java/checks/RecordInsteadOfClassCheck.java
m-g-sonar: I'm also a bit puzzled by this condition. If fields are final and private, setters do not make sense. Probably we can drop the condition from the rule

java-checks/src/main/java/org/sonar/java/checks/RecordInsteadOfClassCheck.java
m-g-sonar: 👍🏻

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3682, opened 2021-07-07
SONARJAVA-3748 RSPEC-6209 Members ignored during record serialization should not be used

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/serialization/RecordSerializationIgnoredMembersCheck.java
```diff
@@ -0,0 +1,88 @@
...
+    methodMatcher("writeExternal", "java.io.ObjectOutput"));
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.RECORD);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    for (Tree member : ((ClassTree) tree).members()) {
```
dorian-burihabwa-sonarsource, line 50: Should we check that the record implements `Serializable` or `Externalizable` before looking at its members?

java-checks/src/main/java/org/sonar/java/checks/serialization/RecordSerializationIgnoredMembersCheck.java
```diff
@@ -0,0 +1,88 @@
...
+      if (member.is(Tree.Kind.VARIABLE)) {
+        checkField((VariableTree) member);
+      } else if (member.is(Tree.Kind.METHOD)) {
```
dorian-burihabwa-sonarsource, line 53: To increase precision, we could even check that the `Serializable` related methods are private and that the `Externalizable` methods are overriding. But that could also be tackled as an improvement to the rule

java-checks/src/test/java/org/sonar/java/checks/serialization/RecordSerializationIgnoredMembersCheckTest.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.sonar.java.checks.verifier.TestUtils.testSourcesPath;
+
+class RecordSerializationIgnoredMembersCheckTest {
+
+  @Test
+  void test() {
+    CheckVerifier.newVerifier()
+      .onFile(testSourcesPath("checks/serialization/RecordSerializationIgnoredMembersCheck.java"))
+      .withCheck(new RecordSerializationIgnoredMembersCheck())
+      .withJavaVersion(16)
```
dorian-burihabwa-sonarsource, line 34: It may not be necessary to set the version explicitly as parsing the code with versions prior to 16 will not generate any issue

dorian-burihabwa-sonarsource (changes requested):
We could still a make a few tweaks to increase precision even though the rule already covers the canonical examples pretty well

java-checks/src/main/java/org/sonar/java/checks/serialization/RecordSerializationIgnoredMembersCheck.java
m-g-sonar: I don't think it is relevant, you don't have to implement `Serializable` interface to write the private methods, and still be serializable in practice.

Then, if you have the signature of  `Externalizable` methods, then it should be enough. Also, there is already a few rules targeting wrong serializable contracts.

java-checks/src/main/java/org/sonar/java/checks/serialization/RecordSerializationIgnoredMembersCheck.java
m-g-sonar: Since the method signatures are matching, I'm not sure this is required, and I feel it's OK to let it that way.

java-checks/src/test/java/org/sonar/java/checks/serialization/RecordSerializationIgnoredMembersCheckTest.java
m-g-sonar: Well, records are only available with 16 and we anyway parse with this version. So indeed, I'll drop the line. However, we still parse and raise issues. If I force a version below 16, it crash with a parsing issue.

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #3683, opened 2021-07-08
SONARJAVA-3752 Implement rule S6207: Avoid redundant constructors/methods in records

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.RECORD);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree targetRecord = (ClassTree) tree;
+
+    List<VariableTree> components = targetRecord.recordComponents();
+    Map<String, VariableTree> componentsByName = components.stream()
```
m-g-sonar, line 57: You only use `componentsByName` for the component names. What about simply using a `set` called `componentsName` ?

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+      .collect(Collectors.toMap(component -> component.symbol().name(), component -> component));
+
+    for (Tree member : targetRecord.members()) {
+      if (member.is(Tree.Kind.CONSTRUCTOR)) {
+        MethodTree constructor = (MethodTree) member;
+        // Report if the constructor is empty
+        if (constructor.block().body().isEmpty() || onlyDoesSimpleAssignments(constructor, components)) {
+          reportIssue(constructor.simpleName(), "Remove this redundant constructor which is the same as a default one.");
+        }
```
m-g-sonar, line 66: I would extract this into a `checkConstructor` method

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+      } else if (member.is(Tree.Kind.METHOD)) {
+        MethodTree method = (MethodTree) member;
+        String methodName = method.symbol().name();
+        if (!componentsByName.containsKey(methodName)) {
+          continue;
+        }
+        if (onlyReturnsRawValue(method, components)) {
+          reportIssue(method.simpleName(), "Remove this redundant method which is the same as a default one.");
+        }
```
m-g-sonar, line 75: I would extract this into a `checkMethod` method

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+      }
+      ExpressionTree rightHandSide = assignment.get().expression();
+      if (!rightHandSide.is(Tree.Kind.IDENTIFIER)) {
+        continue;
+      }
+      Symbol valueSymbol = ((IdentifierTree) rightHandSide).symbol();
+      boolean valueIsAParameter = parameters.stream().anyMatch(parameter -> parameter.symbol().equals(valueSymbol));
+      if (valueIsAParameter && variableSymbol.name().equals(valueSymbol.name())) {
+        componentsAssignedInConstructor.add(variableSymbol);
+      }
```
m-g-sonar, line 132: extract this to a given `void` method which populates the set `componentsAssignedInConstructor`, so you will get rid of the `continue` statements, and replace them by simple empty `return;`

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+    return componentsAssignedInConstructor.size() == components.size();
+  }
+
+  private static Optional<AssignmentExpressionTree> extractAssignment(StatementTree statement) {
+    if (!statement.is(Tree.Kind.EXPRESSION_STATEMENT)) {
+      return Optional.empty();
+    }
+    ExpressionStatementTree initialStatement = (ExpressionStatementTree) statement;
```
m-g-sonar, line 141: directly use a variable `expression`, to not have to reuse `initialStatement.expression()`:

```java
ExpressionTree expression = ((ExpressionStatementTree) statement).expression();
```

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+    for (StatementTree statement : statements) {
+      Optional<AssignmentExpressionTree> assignment = extractAssignment(statement);
+      if (!assignment.isPresent()) {
+        continue;
+      }
+      ExpressionTree leftHandSide = assignment.get().variable();
+      if (!leftHandSide.is(Tree.Kind.MEMBER_SELECT)) {
+        continue;
+      }
+      Symbol variableSymbol = ((MemberSelectExpressionTree) leftHandSide).identifier().symbol();
```
m-g-sonar, line 119: You can also check that the member_select is `this` using 
```java
ExpressionUtils.isThis(((MemberSelectExpressionTree) leftHandSide).expression())
```

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.RECORD);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree targetRecord = (ClassTree) tree;
+
+    List<VariableTree> components = targetRecord.recordComponents();
```
m-g-sonar, line 56: It seems to me that having a `Set<Symbol.VariableSymbol>` would make more sense. You always check the symbols, not the variable trees.

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+      Optional<AssignmentExpressionTree> assignment = extractAssignment(statement);
+      if (!assignment.isPresent()) {
+        continue;
+      }
+      ExpressionTree leftHandSide = assignment.get().variable();
+      if (!leftHandSide.is(Tree.Kind.MEMBER_SELECT)) {
+        continue;
+      }
+      Symbol variableSymbol = ((MemberSelectExpressionTree) leftHandSide).identifier().symbol();
+      boolean variableIsAComponent = components.stream().anyMatch(component -> component.symbol().equals(variableSymbol));
```
m-g-sonar, line 120: with a set, it would be extremely simple: `components.contains(variableSymbol)`

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+      }
+      Symbol identifierSymbol = identifier.symbol();
+      boolean identifierMatchesComponent = components.stream().anyMatch(component -> component.symbol().equals(identifierSymbol));
+      onlyReturnsComponents |= identifierMatchesComponent;
+    }
+    return onlyReturnsComponents;
+  }
+
+  public static boolean onlyDoesSimpleAssignments(MethodTree constructor, List<VariableTree> components) {
+    List<VariableTree> parameters = constructor.parameters();
```
m-g-sonar, line 104: Having a `Set<Symbol.VariableSymbol>` would make more sense as well here.

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+      if (!rightHandSide.is(Tree.Kind.IDENTIFIER)) {
+        continue;
+      }
+      Symbol valueSymbol = ((IdentifierTree) rightHandSide).symbol();
+      boolean valueIsAParameter = parameters.stream().anyMatch(parameter -> parameter.symbol().equals(valueSymbol));
+      if (valueIsAParameter && variableSymbol.name().equals(valueSymbol.name())) {
+        componentsAssignedInConstructor.add(variableSymbol);
+      }
+    }
+    return componentsAssignedInConstructor.size() == components.size();
```
m-g-sonar, line 134: I think testing equality would be simpler, once components is a set of symbol.

java-checks/src/main/java/org/sonar/java/checks/RedundantConstructorsAndMethodsShouldBeAvoidedCheck.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+        if (onlyReturnsRawValue(method, components)) {
+          reportIssue(method.simpleName(), "Remove this redundant method which is the same as a default one.");
+        }
+      }
+    }
+  }
+
+  public static boolean onlyReturnsRawValue(MethodTree method, Collection<VariableTree> components) {
+    List<ReturnStatementTree> returnStatements = method.block().body().stream()
```
m-g-sonar, line 81: I feel this should be done simpler. Also, I don't see any test with more than one statement, so I don't know if the loop is working fine. To me, it is suspcisious.

Try using something similar as in: 

https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/java/org/sonar/java/checks/GettersSettersOnRightFieldCheck.java#L135

Also, I try to achieve the same thing on my PR, here: https://github.com/SonarSource/sonar-java/pull/3685/files#diff-98bea908beb6ebd2a216e7e338702537b1ff7479f297b953001cd1ede773383fR79

m-g-sonar (changes requested):
On the good track, but the approach can be simplified to rely more on semantic (symbols) and less on syntax (trees).

Also, please reduce the number of commits in your PR to a **single one** before submitting it to review (or at max 2-3 if it makes sense).  Your work commits are irrelevant for review and having a single commit would allow merge-rebase without having to squash!

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6207_java.html
```diff
@@ -0,0 +1,48 @@
+<p>In Java 16 records represent a brief notation for immutable data structures. Records have autogenerated implementations for constructors with all
+parameters, <code>getters</code>, <code>equals</code>, <code>hashcode</code> and <code>toString</code>. Although these methods can&nbsp;still be
```
m-g-sonar, line 2: there is a suspicious character `&nbsp;`  here, as well as on almost every example of code!

java-checks/src/main/java/org/sonar/java/checks/RedundantRecordMethodsCheck.java
```diff
@@ -0,0 +1,179 @@
...
+    List<Symbol.VariableSymbol> parameters = constructor.parameters().stream()
+      .map(parameter -> (Symbol.VariableSymbol) parameter.symbol())
+      .collect(Collectors.toList());
+    List<AssignmentExpressionTree> assignments = extractAssignments(constructor.block().body());
+    Set<Symbol> componentsAssignedInConstructor = new HashSet<>();
+    for (AssignmentExpressionTree assignment : assignments) {
+      Optional<Symbol.VariableSymbol> variableSymbol = assignsParameterToComponent(assignment, components, parameters);
+      if (variableSymbol.isPresent()) {
+        componentsAssignedInConstructor.add(variableSymbol.get());
+      }
```
m-g-sonar, line 123: I think this can be rewritten even simpler as:
```java
assignsParameterToComponent(assignment, components, parameters)
  .ifPresent(componentsAssignedInConstructor::add);
```

PR #3685, opened 2021-07-08
SONARJAVA-3758 RSPEC-6211 Prefer overriding default record's getter

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/RecordDuplicatedGetterCheck.java
```diff
@@ -0,0 +1,150 @@
...
+    return String.format("Remove this getter '%s()' from record and override an existing one '%s()'.", getter.simpleName().name(), component.name());
+  }
+
+  private static List<Symbol.VariableSymbol> recordComponents(Symbol.TypeSymbol recordSymbol) {
+    return recordSymbol
+      .memberSymbols()
+      .stream()
+      .filter(Symbol::isVariableSymbol)
+      .map(Symbol.VariableSymbol.class::cast)
+      .filter(component -> !"this".equals(component.name()))
```
sebastian-hungerecker-sonarsource, line 130: I don't understand what this check does. Why/how would a member's name be `this`? Does that happen in case of inner classes? If so, nested records are always static, so they'll never hold a reference to the outer `this`.

m-g-sonar: My bad. In our previous semantic (before ECJ), we were artificially introducing a "this" symbol in the list of member symbols. This is not done, nor required anymore.

java-checks-test-sources/src/main/java/checks/RecordDuplicatedGetterCheck.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override public int age() { return age; }
+    public int getAge() {        return this.age; }
+
+    @Override public double size() { return size; }
+    public double getSize() {        return size(); }
+
+    @Override public boolean alive() { return alive; }
+    public boolean isAlive() { return Being.CTHULU.alive(); } // Noncompliant
+
+    public Color getColor() { return this.color(); }
```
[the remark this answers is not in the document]
m-g-sonar: I added a test (`power`) and it was already not raising an issue (which I think is fine).

java-checks-test-sources/src/main/java/checks/RecordDuplicatedGetterCheck.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override public float power() { return getPower(); }
+    public float getPower() {        return power; }
```
sebastian-hungerecker-sonarsource, line 40: I think the only reason we don't raise an issue here is just because `getPower()` directly returns the corresponding component. I believe if we change the body of `getPower` to something like `return Math.abs(power);`, it will raise an issue (but shouldn't because both `power` and `getPower` both return the same thing).

sebastian-hungerecker-sonarsource (changes requested):

java-checks-test-sources/src/main/java/checks/RecordDuplicatedGetterCheck.java
m-g-sonar: mmmh indeed.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3686, opened 2021-07-12
SONARJAVA-3763 Update rules targeting classes to also target records

m-g-sonar:


java-checks-test-sources/src/main/java/checks/TooManyMethodsCheck.java
```diff
@@ -48,15 +57,15 @@ public void method10() {}
...
-    public void method1() {}
-    public void method2() {}
-    public void method3() {}
-    public void method4() {}
-    public void method5() {}
+    @Override public void method1() {}
+    @Override public void method2() {}
+    @Override public void method3() {}
+    @Override public void method4() {}
+    @Override public void method5() {}
```
quentin-jaquier-sonarsource, line 64: It seems that the annotation was omitted on purpose, to test that the override is detected without the help of the annotation. We should not add them, or add an equivalent test without.

java-checks-test-sources/src/main/java/checks/MembersDifferOnlyByCapitalizationCheck.java
```diff
@@ -1,35 +1,40 @@
...
+package checks;
+
+class MembersDifferOnlyByCapitalizatuionCheck {
+
+abstract class A implements MyInterface {
   private static final long var1 = -9215047833775013803L; // Compliant
   public long var2 = 0L;
-  
+
   public static void equAls(Object obj) {} // Noncompliant [[sc=22;ec=28]] {{Rename method "equAls" to prevent any misunderstanding/clash with method "equals" defined in superclass "java.lang.Object".}}
```
quentin-jaquier-sonarsource, line 9: This code does not compile on my side: "inner class can not have static declarations".

quentin-jaquier-sonarsource (approved):
Except for 3 minor comments, LGTM, nice work!

quentin-jaquier-sonarsource (changes requested):
In fact, after having a last look, it seems to me that the commits related to the additions of secondary locations in S1845 make sense, but are independent of the topic and should be extracted. You can probably use [SONARJAVA-3902](https://jira.sonarsource.com/browse/SONARJAVA-3902) for this.

m-g-sonar:
Okay, I extracted the commit modifying S1845 into a dedicated branch (https://github.com/SonarSource/sonar-java/tree/improve_s1845_secondary) and used [SONARJAVA-3902](https://jira.sonarsource.com/browse/SONARJAVA-3902) as prefix for the commit message.

java-checks-test-sources/src/main/java/checks/MembersDifferOnlyByCapitalizationCheck.java
m-g-sonar: that's strange. it compile on my side.

java-checks-test-sources/src/main/java/checks/TooManyMethodsCheck.java
m-g-sonar: i'll remove some of them

java-checks-test-sources/src/main/java/checks/MembersDifferOnlyByCapitalizationCheck.java
m-g-sonar: removing the static modifier

quentin-jaquier-sonarsource (approved):
LGTM

PR #3688, opened 2021-07-14
SONARJAVA-3872 "JSymbol.convertMetadata" should not throw an Exception when ecj fails

quentin-jaquier-sonarsource:


java-frontend/src/test/java/org/sonar/java/model/JParserSemanticTest.java
```diff
@@ -1207,6 +1208,30 @@ void declaration_method_parameterized_recovered() {
     assertThat(metadata.annotations()).isEmpty();
   }
 
+  @Test
+  void ecj_exception_when_computing_metadata_should_be_caught() {
+    String source = "" +
+      " public class C {\n" +
+      "  interface I1 {}\n" +
+      "  interface I2 {\n" +
+      "    @I1(\"\")\n" +
```
m-g-sonar, line 1217: worth adding a comment that it explicitly does not compile, as `I1` is not an `@interface`.

PR #3689, opened 2021-07-14
SONARJAVA-3892 Exclude "com.sun.jersey" and "com.sun.faces" from S1191 by default

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/files/non-compiling/checks/SunPackagesUsedCheckCustom.java
```diff
@@ -2,7 +2,7 @@
 import com.sun.jersey.api.client.ClientHandlerException; // Compliant
 import java.util.ArrayList;
 
-class A {
+class SunPackagesUsedCheck {
```
dorian-burihabwa-sonarsource, line 5: We should rename the class to `SunPackagesUsedCheckCustom` to match the file name and avoid confusion when looking at the two test files side by side.

java-checks-test-sources/src/main/files/non-compiling/checks/SunPackagesUsedCheckCustom.java
```diff
@@ -2,7 +2,7 @@
 import com.sun.jersey.api.client.ClientHandlerException; // Compliant
 import java.util.ArrayList;
```
dorian-burihabwa-sonarsource, line 3: We should add a non compliant import from `com.sun.faces.*` or `com.sun.jersey.*" element to show that the `exclude` pattern is overwritten and not extended on top of its default value.

dorian-burihabwa-sonarsource (changes requested):
LGTM but a few tweaks could help the test source files easier to understand

java-checks-test-sources/src/main/files/non-compiling/checks/SunPackagesUsedCheckCustom.java
quentin-jaquier-sonarsource: Good idea

dorian-burihabwa-sonarsource (approved):

PR #3690, opened 2021-07-14
SONARJAVA-3898 Don't apply S5838 for calls to equals in methods with "equals" in the name

quentin-jaquier-sonarsource:


sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3693, opened 2021-07-15
SONARJAVA-3901 FP in S2245 (PseudeRandomCheck) when passing a SecureRandom object as parameter

sebastian-hungerecker-sonarsource:
Also fix an FN when a random method is imported statically

### DIFF
--- a/java-checks-test-sources/src/main/java/checks/PseudoRandomCheck.java
+++ b/java-checks-test-sources/src/main/java/checks/PseudoRandomCheck.java
@@ -1,10 +1,14 @@
 package checks;
 
+import java.security.SecureRandom;
+import java.util.Locale;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
 import org.apache.commons.lang.math.JVMRandom;
 import org.apache.commons.lang.math.RandomUtils;
 import org.apache.commons.lang.RandomStringUtils;
+
+import static org.apache.commons.lang.RandomStringUtils.random;
 
 class PseudoRandomCheck {
   void fun() {
@@ -26,7 +30,7 @@ class PseudoRandomCheck {
     double abs = Math.abs(12); // Compliant
 
     // java.util.concurrent.ThreadLocalRandom
-    int rand2 = ThreadLocalRandom.current().nextInt();  // Noncompliant
+    int rand2 = ThreadLocalRandom.current().nextInt();  // Noncompliant [[sc=35;ec=42]]
 
     // org.apache.commons.lang.math.RandomUtils
     RandomUtils randomUtils = new RandomUtils();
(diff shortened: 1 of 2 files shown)


java-checks-test-sources/src/main/java/checks/PseudoRandomCheck.java
```diff
@@ -42,12 +46,24 @@ void fun() {
...
     org.apache.commons.lang3.RandomStringUtils randomStringUtils2 = new org.apache.commons.lang3.RandomStringUtils();
     String rand10 = randomStringUtils.random(1); // Noncompliant
     String rand11 = org.apache.commons.lang3.RandomStringUtils.random(1); // Noncompliant
+    rand11 = org.apache.commons.lang3.RandomStringUtils.random(1, 0, 0, false, false, null); // Noncompliant [[sc=57;ec=63]]
+    // Here we should raise an issue only on the `new Random`, not on the call to `random` itself
+    rand11 = org.apache.commons.lang3.RandomStringUtils.random(1, 0, 0, false, false, null, new Random()); // Noncompliant [[sc=97;ec=103]]
+    // Here there should be no issue because `random` will use the supplied secure source of randomness
+    rand11 = org.apache.commons.lang3.RandomStringUtils.random(1, 0, 0, false, false, null, new SecureRandom()); // Compliant
 
+    String rand12 = random(42).toLowerCase(Locale.ROOT); // FN because we ignore calls inside chains to avoid raising multiple issues on line 33
```
sebastian-hungerecker-sonarsource, line 66: This is an old FN, not introduced by this PR. I just felt it should be mentioned explicitly. I also wonder whether it wouldn't be okay to have two issues on line 33 and avoid this FN.

java-checks-test-sources/src/main/java/checks/PseudoRandomCheck.java
```diff
@@ -26,7 +30,7 @@ void fun() {
     double abs = Math.abs(12); // Compliant
 
     // java.util.concurrent.ThreadLocalRandom
-    int rand2 = ThreadLocalRandom.current().nextInt();  // Noncompliant
+    int rand2 = ThreadLocalRandom.current().nextInt();  // Noncompliant [[sc=35;ec=42]]
```
sebastian-hungerecker-sonarsource, line 33: Here the issue location changed from the call to `nextInt` to the call to `current`. I think this is fine and is anyway more consistent with how we handle `java.util.Random` (where we would also raise on the constructor call, not the call to `nextInt`).

java-checks-test-sources/src/main/java/checks/PseudoRandomCheck.java
```diff
@@ -42,12 +46,24 @@ void fun() {
...
     org.apache.commons.lang3.RandomStringUtils randomStringUtils2 = new org.apache.commons.lang3.RandomStringUtils();
     String rand10 = randomStringUtils.random(1); // Noncompliant
     String rand11 = org.apache.commons.lang3.RandomStringUtils.random(1); // Noncompliant
+    rand11 = org.apache.commons.lang3.RandomStringUtils.random(1, 0, 0, false, false, null); // Noncompliant [[sc=57;ec=63]]
+    // Here we should raise an issue only on the `new Random`, not on the call to `random` itself
+    rand11 = org.apache.commons.lang3.RandomStringUtils.random(1, 0, 0, false, false, null, new Random()); // Noncompliant [[sc=97;ec=103]]
+    // Here there should be no issue because `random` will use the supplied secure source of randomness
+    rand11 = org.apache.commons.lang3.RandomStringUtils.random(1, 0, 0, false, false, null, new SecureRandom()); // Compliant
 
+    String rand12 = random(42).toLowerCase(Locale.ROOT); // Noncompliant [[sc=21;ec=27]]
```
sebastian-hungerecker-sonarsource, line 66: This is a fixed FN - previously we didn't report on statically imported methods because we used the receiver expression to determine the class of the method.

java-checks-test-sources/src/main/java/checks/PseudoRandomCheck.java
```diff
@@ -42,12 +46,24 @@ void fun() {
     RandomStringUtils randomStringUtils = new RandomStringUtils();
     String rand8 = randomStringUtils.random(1); // Noncompliant
     String rand9 = RandomStringUtils.random(1); // Noncompliant
+    rand9 = random(1); // Noncompliant
```
sebastian-hungerecker-sonarsource, line 49: This is a FN - previously we didn't report on statically imported methods because we used the receiver expression to determine the class of the method

java-checks-test-sources/src/main/java/checks/PseudoRandomCheck.java
sebastian-hungerecker-sonarsource, line 66: This is another fixed FN - previously this would be ignored because it's part of a method chain (and also because of the static import).

quentin-jaquier-sonarsource (approved):
LGTM

PR #3698, opened 2021-07-16
SONARJAVA-3867 S2479 allows tabs in string literals when the allowTabs flag is set to true

dorian-burihabwa-sonarsource:


alban-auzeill (changes requested):
Now we can remove the "Exceptions" § in the rule description.

alban-auzeill (approved):

PR #3699, opened 2021-07-16
SONARJAVA-3904 Fix highlighting

m-g-sonar:


java-frontend/src/test/java/org/sonar/java/ast/visitors/SyntaxHighlighterVisitorTest.java
```diff
@@ -192,13 +191,57 @@ void text_block() throws Exception {
...
+   */
+  @Test
+  void switch_expression() throws Exception {
+    this.eol = "\n";
+    InputFile inputFile = generateTestFile("src/test/files/highlighter/SwitchExpression.java");
+    scan(inputFile);
+
+    String componentKey = inputFile.key();
+    assertThatHasBeenHighlighted(componentKey, 9, 12, 9, 14, TypeOfText.KEYWORD); // yield true;
+    assertThatHasBeenHighlighted(componentKey, 13, 12, 13, 14, TypeOfText.KEYWORD); // yield false;
```
quentin-jaquier-sonarsource, line 205: It seems to work as expected, but we should probably use 9 for `startColumn` instead of 12 for consistency.

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/ClassTree.java
```diff
@@ -86,6 +86,14 @@
 
   ListTree<TypeTree> superInterfaces();
 
+  /**
+   * @since Java 15
+   * @deprecated Preview Feature
+   */
+  @Deprecated
+  @Nullable
+  SyntaxToken permitsKeyword();
```
quentin-jaquier-sonarsource, line 95: Should we document this change in the [documentation: API changes](https://github.com/SonarSource/sonar-java/blob/master/sonar-java-plugin/src/main/resources/static/documentation.md)?

m-g-sonar: Nop, we do not document changes related to preview features yet.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3704, opened 2021-07-22
SONARJAVA-3965 Provide a new extensible API for issue reporting

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/reporting/InternalJavaIssueBuilder.java
```diff
@@ -0,0 +1,236 @@
...
+          .collect(Collectors.toList()));
+      }
+    }
+
+    newIssue.save();
+  }
+
+  private static TextRange range(InputFile file, JavaFileScannerContext.Location location) {
+    AnalyzerMessage.TextSpan textSpan = AnalyzerMessage.textSpanFor(location.syntaxNode);
+    if (textSpan.onLine()) {
```
m-g-sonar, line 203: I feel that this condition will never be possible. A `location` holds a `Tree` (`syntaxNode`), and it cannot be at an invalid position which would trigger an `onLine()` test to be true (`startCharacter` has to be `-1`).

What about removing it?

java-frontend/src/main/java/org/sonar/java/model/DefaultJavaFileScannerContext.java
```diff
@@ -261,4 +262,9 @@ protected static void throwIfEndOfAnalysisCheck(JavaCheck javaCheck) {
     }
     return Optional.empty();
   }
+
+  @Override
+  public JavaIssueBuilder newIssue() {
+    return new InternalJavaIssueBuilder(inputFile, sonarComponents);
```
m-g-sonar, line 268: I wonder if we should rework the various `reportIssue()` methods to use the new API...

java-frontend/src/main/java/org/sonar/java/reporting/InternalJavaIssueBuilder.java
quentin-jaquier-sonarsource: We took the code from another class that was able to reach this situation. Since it is not possible currently with this new way, we can probably remove it indeed.

java-frontend/src/main/java/org/sonar/java/model/DefaultJavaFileScannerContext.java
quentin-jaquier-sonarsource: I would suggest considering doing this as a second step, this PR is getting quite big, and I would rather not add too much to it.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3726, opened 2021-08-05
SONARJAVA-3990 S1120 should not crash on code containing line breaking control characters

dorian-burihabwa-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3727, opened 2021-08-09
SONARJAVA-3854 Rule S5329: Collection constructors should not be used as java.util.function.Function

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/CollectionConstructorReferenceCheck.java
```diff
@@ -0,0 +1,97 @@
...
+      if (intArgumentName != null) {
+        reportIssue(methodReference, String.format(
+          "Replace this method reference by a lambda to explicitly show the usage of %1$s(int %2$s) or %1$s().",
+          constructorType.name(),
+          intArgumentName));
+      }
+    });
+  }
+
+  private static boolean isConstructorWithSingleIntArgument(MethodReferenceTree methodReference) {
```
quentin-jaquier-sonarsource, line 78: Can't we use a method matcher instead of computing this manually?

java-checks/src/main/java/org/sonar/java/checks/CollectionConstructorReferenceCheck.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @Override
+  protected void onMethodReferenceFound(MethodReferenceTree methodReference) {
+    if ("java.util.function.Function".equals(methodReference.symbolType().fullyQualifiedName())) {
+      Type methodOwnerType = ((ExpressionTree) methodReference.expression()).symbolType();
+      reportIssue(methodReference, String.format(
+        "Replace this method reference by a lambda to explicitly show the usage of %1$s(int %2$s) or %1$s().",
+        methodOwnerType.name(),
+        "IdentityHashMap".equals(methodOwnerType.name()) ? "expectedMaxSize" : "initialCapacity"));
```
quentin-jaquier-sonarsource, line 66: You can extract `methodOwnerType.name()` to a variable since you are calling it two times.

quentin-jaquier-sonarsource (approved):
One little comment, otherwise, LGTM, nice refactoring!

PR #3734, opened 2021-08-13
SONARJAVA-3946 Add quick fixes for S1488 (ImmediatelyReturnedVariableCheck)

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ImmediatelyReturnedVariableCheck.java
```diff
@@ -79,12 +84,23 @@ public void visitBlock(BlockTree tree) {
             // Can only happen for non-compilable code, still, we should not report anything.
             return;
           }
-          context.reportIssue(
-            this, initializer, "Immediately " + lastTypeForMessage + " this expression instead of assigning it to the temporary variable \"" + identifier + "\".");
+          ((InternalJavaIssueBuilder) ((DefaultJavaFileScannerContext) context).newIssue())
+            .forRule(this)
+            .onTree(initializer)
+            .withMessage(String.format("Immediately %s this expression instead of assigning it to the temporary variable \"%s\".", lastTypeForMessage, identifier))
```
m-g-sonar, line 90: `withMessage()` has already a built-in support of message format, just write
```java
.withMessage("Immediately %s this expression instead of assigning it to the temporary variable \"%s\".", lastTypeForMessage, identifier)
```

java-checks/src/main/java/org/sonar/java/checks/ImmediatelyReturnedVariableCheck.java
```diff
@@ -79,12 +84,23 @@ public void visitBlock(BlockTree tree) {
...
+            .report();
         }
       }
     }
+  }
 
+  private static JavaQuickFix quickFix(StatementTree butLastStatement, StatementTree lastStatement, VariableTree variableTree, String lastTypeForMessage) {
+    // Equal token can not be null at this point, we checked before the presence of the initializer
+    return JavaQuickFix.newQuickFix("Inline expression")
+      .addTextEdit(JavaTextEdit.replaceBetweenTree(variableTree.modifiers(), variableTree.equalToken(), lastTypeForMessage),
```
m-g-sonar, line 101: modifiers has been always emtpy in our examples. I would like to see one with `final`, such as:
```java
final int i = 42;
return i;
```
Also, I wonder about the end tree. What if the initializer is on a different line? or after many spaces?
Maybe it would be simpler to just replace from the beginning of the modifiers (if there is any) included until the beginning of the initializer, not included.

java-checks/src/test/java/org/sonar/java/checks/ImmediatelyReturnedVariableCheckTest.java
```diff
@@ -22,14 +22,16 @@
...
 import org.sonar.java.checks.verifier.CheckVerifier;
 import org.sonar.java.checks.verifier.TestUtils;
+import org.sonar.java.checks.verifier.internal.InternalCheckVerifier;
 
 class ImmediatelyReturnedVariableCheckTest {
 
   @Test
   void test() {
-    CheckVerifier.newVerifier()
+    ((InternalCheckVerifier) CheckVerifier.newVerifier())
```
m-g-sonar, line 31: You don't need to cast. instead of 
```
((InternalCheckVerifier) CheckVerifier.newVerifier())
```
You can call directly
```
InternalCheckVerifier.newInstance()
```

m-g-sonar (commented):
There is typo in 2nd commit message `Compliable` ➡️ `Compilable`

java-checks/src/main/java/org/sonar/java/checks/ImmediatelyReturnedVariableCheck.java
quentin-jaquier-sonarsource: Indeed, sounds cleaner to do it this way.

PR #3735, opened 2021-08-13
SONARJAVA-3915 Add quick fixes for S1124 (ModifiersOrderCheck)

m-g-sonar:


java-checks-test-sources/src/main/java/checks/ModifiersOrderCheckWithQuickFixes.java
```diff
@@ -0,0 +1,45 @@
+package checks;
+
+class ModifiersOrderCheckWithQuickFixes {
```
quentin-jaquier-sonarsource, line 3: In our code, we add annotations on a first line, then modifier on the following:
```
@Annotation2
final private void foo() {}
```
We should make sure this correctly get fixed to

```
@Annotation2
private final void foo() {}
```

quentin-jaquier-sonarsource (approved):

PR #3737, opened 2021-08-16
SONARJAVA-3951 Add quick fixes for S5164 (ThreadLocalCleanupCheck)

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ThreadLocalCleanupCheck.java
```diff
@@ -63,7 +69,18 @@ public void visitNode(Tree tree) {
     } else {
       MethodInvocationTree mit = (MethodInvocationTree) tree;
       if (THREADLOCAL_SET.matches(mit) && mit.arguments().get(0).is(Tree.Kind.NULL_LITERAL)) {
-        reportIssue(mit, "Use \"remove()\" instead of \"set(null)\".");
+        InternalJavaIssueBuilder builder = ((InternalJavaIssueBuilder) ((DefaultJavaFileScannerContext) context).newIssue())
+          .forRule(this)
+          .onTree(mit)
+          .withMessage("Use \"remove()\" instead of \"set(null)\".");
+        ExpressionTree expressionTree = mit.methodSelect();
+        // Defensive programming: should never be something else than a member select.
```
alban-auzeill, line 77: nop, nop! it's not defensive programing, replace the last class of the test file by:
```java
class ThreadLocalCleanupExtends extends ThreadLocal {
  private static final ThreadLocal<UserSession> DELEGATE = new ThreadLocal<>(); // Compliant, extends ThreadLocal
  void foo() {
    set(null); // Noncompliant [[sc=5;ec=14;quickfixes=qf3]] {{Use "remove()" instead of "set(null)".}}
    // fix@qf3 {{Replace with "remove()"}}
    // edit@qf3 [[sc=5;ec=14]] {{remove()}}
  }
}
```

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #3738, opened 2021-08-16
SONARJAVA-3909 Add quick fixes for S1481 (UnusedLocalVariableCheck)

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -112,4 +134,74 @@ private static boolean isTryResource(VariableTree variable) {
...
+  }
+
+  private static Optional<VariableTree> getPrecedingVariable(VariableTree current) {
+    Tree parent = current.parent();
+    List<Tree> children;
+    if (parent.is(Tree.Kind.BLOCK, Tree.Kind.INITIALIZER)) {
+      children = ((BlockTreeImpl) parent).children();
+    } else {
+      children = ((StaticInitializerTreeImpl) parent).children();
+    }
```
alban-auzeill, line 189: Can be simplified because StaticInitializerTreeImpl extends BlockTreeImpl
```suggestion
    List<Tree> children = ((BlockTreeImpl) parent).children();
```

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -112,4 +134,74 @@ private static boolean isTryResource(VariableTree variable) {
...
+    return Optional.empty();
+  }
+
+  private static Optional<SyntaxToken> getPrecedingComma(VariableTree variable) {
+    Optional<VariableTree> precedingVariable = getPrecedingVariable(variable);
+    if (!precedingVariable.isPresent()) {
+      return Optional.empty();
+    }
+    return Optional.of(precedingVariable.get().lastToken());
```
alban-auzeill, line 179:
```suggestion
    return getPrecedingVariable(variable).map(Tree::lastToken);
```

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -112,4 +134,74 @@ private static boolean isTryResource(VariableTree variable) {
...
+    Tree parent = current.parent();
+    List<Tree> children;
+    if (parent.is(Tree.Kind.BLOCK, Tree.Kind.INITIALIZER)) {
+      children = ((BlockTreeImpl) parent).children();
+    } else {
+      children = ((StaticInitializerTreeImpl) parent).children();
+    }
+    int currentIndex = children.indexOf(current);
+    // If the variable is the first element that follows the opening token, there is no predecessor to return
+    if (currentIndex == 1) {
```
alban-auzeill, line 192: To add some defensive programming, I suggest testing `if (currentIndex <= 1)` to also include `-1`, when `current` is not found.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -112,4 +134,74 @@ private static boolean isTryResource(VariableTree variable) {
...
+    Tree child = children.get(currentIndex - 1);
+    if (!child.is(Tree.Kind.VARIABLE)) {
+      return Optional.empty();
+    }
+    // If the preceding element is a variable, we check if that it is part of the same declaration
+    VariableTree preceding = (VariableTree) child;
+    if (preceding.firstToken().equals(current.firstToken())) {
+      return Optional.of(preceding);
+    }
+    return Optional.empty();
```
alban-auzeill, line 205:
```suggestion
    // If there is a predecessor, we check that it is a variable, and it is part of the same declaration
    Tree preceding = children.get(currentIndex - 1);
    if (preceding.is(Tree.Kind.VARIABLE) && preceding.firstToken().equals(current.firstToken())) {
      return Optional.of((VariableTree) preceding);
    }
    return Optional.empty();
```

alban-auzeill (changes requested):

PR #3739, opened 2021-08-17
SONARJAVA-3929 Add quick fixes for S1172 (UnusedMethodParameterCheck)

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
```diff
@@ -95,23 +104,53 @@ public void visitNode(Tree tree) {
...
+        AnalyzerMessage.TextSpan textSpanToRemove;
+        if (isFirst && isLast) {
+          // no need to remove any comma
+          textSpanToRemove = AnalyzerMessage.textSpanBetween(methodTree.openParenToken(), false, methodTree.closeParenToken(), false);
+        } else if (isLast) {
+          // also remove the previous comma and spaces between previousParameter and parameter
+          VariableTree previousParameter = parameters.get(i - 1);
+          textSpanToRemove = AnalyzerMessage.textSpanBetween(previousParameter.endToken(), true, methodTree.closeParenToken(), false);
+        } else {
+          // also remove the next comma and spaces between parameter and nextParameter
```
dorian-burihabwa-sonarsource, line 144: The approach is sound here provided that all quickfixes are not applied in at once. Otherwise. we need to think of how to  generate the list of contiguous segments to delete.

dorian-burihabwa-sonarsource (approved):

PR #3743, opened 2021-08-18
SONARJAVA-3948 Add quick fixes for S2153 (ImmediateReverseBoxingCheck)

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ImmediateReverseBoxingCheck.java
```diff
@@ -112,10 +117,19 @@ private void visitMethodInvocationTree(MethodInvocationTree mit) {
...
 
-  private void checkForUselessUnboxing(Type targetType, Tree reportTree, ExpressionTree arg0) {
+  private void checkForUselessUnboxing(Type targetType, Tree reportTree, ExpressionTree arg0, Tree originalTree) {
     Type argType = arg0.symbolType();
     if (argType.is(targetType.fullyQualifiedName())) {
-      reportIssue(reportTree, String.format("Remove the boxing to \"%s\"; The argument is already of the same type.", argType.name()));
+      ((InternalJavaIssueBuilder) ((DefaultJavaFileScannerContext) context).newIssue())
+        .forRule(this)
+        .onTree(reportTree)
+        .withMessage(String.format("Remove the boxing to \"%s\"; The argument is already of the same type.", argType.name()))
```
m-g-sonar, line 126: You can use directly the formatter here.

java-checks/src/main/java/org/sonar/java/checks/ImmediateReverseBoxingCheck.java
```diff
@@ -157,13 +171,25 @@ private void checkForBoxing(ExpressionTree expression) {
     return null;
   }
 
-  private void addBoxingIssue(Tree tree, Symbol classSymbol, Tree boxingArg) {
+  private void addBoxingIssue(Tree tree, Symbol classSymbol, Tree boxingArg, Tree originalTree) {
+    String message;
     if (boxingArg.is(Tree.Kind.IDENTIFIER)) {
       IdentifierTree identifier = (IdentifierTree) boxingArg;
-      reportIssue(tree, "Remove the boxing of \"" + identifier.name() + "\".");
+      message = String.format("Remove the boxing of \"%s\".", identifier.name());
```
m-g-sonar, line 178: you could just set the name and use the formatted directly on line 186.

java-checks/src/main/java/org/sonar/java/checks/ImmediateReverseBoxingCheck.java
```diff
@@ -112,10 +117,19 @@ private void visitMethodInvocationTree(MethodInvocationTree mit) {
...
-  private void checkForUselessUnboxing(Type targetType, Tree reportTree, ExpressionTree arg0) {
+  private void checkForUselessUnboxing(Type targetType, Tree reportTree, ExpressionTree arg0, Tree originalTree) {
     Type argType = arg0.symbolType();
     if (argType.is(targetType.fullyQualifiedName())) {
-      reportIssue(reportTree, String.format("Remove the boxing to \"%s\"; The argument is already of the same type.", argType.name()));
+      ((InternalJavaIssueBuilder) ((DefaultJavaFileScannerContext) context).newIssue())
+        .forRule(this)
+        .onTree(reportTree)
+        .withMessage(String.format("Remove the boxing to \"%s\"; The argument is already of the same type.", argType.name()))
+        .withQuickFix(() -> JavaQuickFix.newQuickFix("Remove the boxing")
```
m-g-sonar, line 127: it seems to me that the following code  is duplicated:
```java
() -> JavaQuickFix.newQuickFix("Remove the boxing")
        .addTextEdit(
          JavaTextEdit.replaceTextSpan(AnalyzerMessage.textSpanBetween(originalTree, true, boxingArg, false), ""),
          JavaTextEdit.replaceTextSpan(AnalyzerMessage.textSpanBetween(boxingArg, false, originalTree, true), ""))
        .build()
```
it can maybe be extracted into a dedicated static method building a quickfix with 2 args

java-checks/src/test/java/org/sonar/java/checks/ImmediateReverseBoxingCheckTest.java
```diff
@@ -21,13 +21,24 @@
...
 import org.sonar.java.checks.verifier.CheckVerifier;
+import org.sonar.java.checks.verifier.TestUtils;
+import org.sonar.java.checks.verifier.internal.InternalCheckVerifier;
 
 class ImmediateReverseBoxingCheckTest {
 
   @Test
   void test() {
+    InternalCheckVerifier.newInstance()
+      .onFile(TestUtils.testSourcesPath("checks/ImmediateReverseBoxingCheck.java"))
```
m-g-sonar, line 32: in almost all the other rules, we use static imports for `testSourcesPath` and `nonCompilingTestSourcesPath`. I would suggest to keep doing this!

java-checks/src/main/java/org/sonar/java/checks/ImmediateReverseBoxingCheck.java
quentin-jaquier-sonarsource: In fact, the message changes slightly between the two cases: `boxing of` vs `boxing to`.

java-checks/src/main/java/org/sonar/java/checks/ImmediateReverseBoxingCheck.java
quentin-jaquier-sonarsource: Indeed, nice catch!

m-g-sonar (approved):

PR #3744, opened 2021-08-18
SONARJAVA-3922 Add quick fixes for S3415 (AssertionArgumentOrderCheck)

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/model/DefaultJavaFileScannerContext.java
```diff
@@ -59,6 +59,10 @@
   private final JavaVersion javaVersion;
   private final boolean fileParsed;
 
+  private List<String> lines = null;
+  private List<String> linesWithEnding = null;
```
quentin-jaquier-sonarsource, line 63: Maybe we should consider protecting these lists from modification, as they will be used in many rules.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3748, opened 2021-08-18
SONARJAVA-3952 Add quick fix for BigDecimal with two arguments

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/java/checks/BigDecimalDoubleConstructorCheck.java
```diff
@@ -13,14 +13,28 @@
   BigDecimal bd2 = new BigDecimal(2.0); // Noncompliant [[sc=20;ec=39;quickfixes=qf1]] {{Use "BigDecimal.valueOf" instead.}}
   // fix@qf1 {{Replace with BigDecimal.valueOf}}
   // edit@qf1 [[sc=20;ec=34]] {{BigDecimal.valueOf}}
-  BigDecimal bd4 = new BigDecimal(2.0, mc); // Noncompliant [[sc=20;ec=43;quickfixes=!]] {{Use "BigDecimal.valueOf" instead.}}
+  BigDecimal bd4 = new BigDecimal(2.0, mc); // Noncompliant [[sc=20;ec=43;quickfixes=qf1_2]] {{Use "BigDecimal.valueOf" instead.}}
+  // fix@qf1_2 {{Replace with BigDecimal("2.0")}}
```
alban-auzeill, line 17: The replacement message is misleading, it shows a missing `MathContext` argument.
And without the MathContext, the value is not the same, for example:
```
    MathContext mathContext = new MathContext(3, RoundingMode.FLOOR);
    System.out.println(new BigDecimal(1.2345f, mathContext)); // => 1.23
    System.out.println(new BigDecimal("1.2345")); // => 1.2345
```
We could perhaps say `Replace with BigDecimal("2.0",`

java-checks/src/main/java/org/sonar/java/checks/BigDecimalDoubleConstructorCheck.java
```diff
@@ -56,16 +59,34 @@ public void visitNode(Tree tree) {
...
-  private static JavaQuickFix quickFix(NewClassTree newClassTree) {
+  private static JavaQuickFix valueOfQuickFix(NewClassTree newClassTree) {
     return JavaQuickFix.newQuickFix("Replace with BigDecimal.valueOf")
       .addTextEdit(JavaTextEdit.replaceBetweenTree(newClassTree.newKeyword(), newClassTree.identifier(), "BigDecimal.valueOf"))
       .build();
   }
+
+  private static JavaQuickFix stringConstructorQuickFix(LiteralTree argument) {
+    String argumentValue = argument.value();
+    if (argumentValue.endsWith("f") || argumentValue.endsWith("d")) {
```
alban-auzeill, line 84: Warning, the decimal suffix is case insensitive: `1.0f` == `1.0F` and `1.0d` == `1.0D`

alban-auzeill (changes requested):

quentin-jaquier-sonarsource:
I took the opportunity to also fix the message, suggesting to use `valueOf` even when there is a MathContext. We should suggest using the String constructor instead.

alban-auzeill (approved):

PR #3750, opened 2021-08-19
SONARJAVA-3996 FP S2200(CompareToResultTestCheck) for unusual zero literal

alban-auzeill:


quentin-jaquier-sonarsource (approved):
Expect for the small code smell, LGTM

PR #3751, opened 2021-08-19
SONARJAVA-3955 Add quick fixes for S4973 (CompareStringsBoxedTypesWithEqualsCheck)

dorian-burihabwa-sonarsource:


java-checks-test-sources/src/main/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -1,58 +1,118 @@
...
+      // fix@qf0 {{Replace with boxed comparison}}
+      // edit@qf0 [[sc=23;ec=23]]{{)}}
+      // edit@qf0 [[sc=15;ec=19]]{{, }}
+      // edit@qf0 [[sc=11;ec=11]]{{java.util.Objects.equals(}}
+      if (str1 == "blue") {} // Noncompliant [[sc=16;ec=18;quickfixes=qf1]]
+      // fix@qf1 {{Replace with boxed comparison}}
+      // edit@qf1 [[sc=25;ec=25]]{{)}}
+      // edit@qf1 [[sc=15;ec=19]]{{, }}
+      // edit@qf1 [[sc=11;ec=11]]{{java.util.Objects.equals(}}
+      if ("blue" == str1) {} // Noncompliant [[sc=18;ec=20;quickfixes=qf2]]
```
quentin-jaquier-sonarsource, line 25: As discussed, let's try to use `"x".equals(y)` when there is one literal.

quentin-jaquier-sonarsource (commented):
I'm not sure to understand the idea behind the second commit, you will lose all blame information. Maybe simply renaming the two classes is enough?

java-checks/src/main/java/org/sonar/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -19,29 +19,107 @@
...
+
   @Override
   protected void checkEqualityExpression(BinaryExpressionTree tree) {
     Type leftOpType = tree.leftOperand().symbolType();
     Type rightOpType = tree.rightOperand().symbolType();
     if (!isNullComparison(leftOpType, rightOpType)
       && !isCompareWithBooleanConstant(tree.leftOperand(), tree.rightOperand())
       && (isStringType(leftOpType, rightOpType) || isBoxedType(leftOpType, rightOpType))) {
-      reportIssue(tree.operatorToken());
+      ((InternalJavaIssueBuilder) ((DefaultJavaFileScannerContext) this.context).newIssue())
```
quentin-jaquier-sonarsource, line 53: You can now use the new helper to avoid the cast.

java-checks/src/main/java/org/sonar/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -19,29 +19,107 @@
...
 import org.sonar.plugins.java.api.semantic.Type;
 import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.Tree;
 
 @Rule(key = "S4973")
 public class CompareStringsBoxedTypesWithEqualsCheck extends CompareWithEqualsVisitor {
 
+  private static final String ISSUE_MESSAGE = "Strings and Boxed types should be compared using \"equals()\"";
```
quentin-jaquier-sonarsource, line 42: Issues messages should end with a dot.

java-checks-test-sources/src/main/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -1,58 +1,117 @@
...
-    if (b4 == Boolean.TRUE) {} // Compliant
-    if (Boolean.TRUE == b4) {} // Compliant
-    if (b5 == Boolean.TRUE) {} // FN
-    if (b5 == b6) {} // Noncompliant
-  }
-}
+import java.util.Locale;
+
+class CompareStringsBoxedTypesWithEqualsCheck {
+  class Boxed {
```
quentin-jaquier-sonarsource, line 6: Instead of nesting the classes into a single one, you could rename directly the three classes to keep the meta data of the changes.

java-checks/src/main/java/org/sonar/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -19,29 +19,107 @@
...
+  }
+
+  private static JavaQuickFix computeDefaultQuickFix(BinaryExpressionTree tree) {
+    String callToEquals = tree.is(Tree.Kind.NOT_EQUAL_TO) ? "!java.util.Objects.equals(" : "java.util.Objects.equals(";
+    AnalyzerMessage.TextSpan interOperandSpace = AnalyzerMessage.textSpanBetween(
+      tree.leftOperand().lastToken(), false,
+      tree.rightOperand().firstToken(), false
+    );
+    return JavaQuickFix.newQuickFix(QUICK_FIX_MESSAGE)
+      .addTextEdit(JavaTextEdit.insertAfterTree(tree.rightOperand().lastToken(), ")"))
```
quentin-jaquier-sonarsource, line 119:
```suggestion
      .addTextEdit(JavaTextEdit.insertAfterTree(tree.rightOperand(), ")"))
```
Inserting after the rightOperand or after the last token is the same, you can remove the call. (In practice, the method anyway takes the last token.)

quentin-jaquier-sonarsource: And in fact, you can nearly always omit the call to last/first token, even when calling `textSpanBetween`.

java-checks/src/main/java/org/sonar/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -19,29 +19,107 @@
...
     }
   }
-  
+
   private static boolean isCompareWithBooleanConstant(ExpressionTree left, ExpressionTree right) {
     return ExpressionsHelper.getConstantValueAsBoolean(left).value() != null ||
       ExpressionsHelper.getConstantValueAsBoolean(right).value() != null;
   }
 
+  private static List<JavaQuickFix> computeQuickFix(BinaryExpressionTree tree) {
```
quentin-jaquier-sonarsource, line 67: No need to return a List when you have a single JavaQuickFix, you can directly return it, and call `withQuickFix`.

java-checks/src/main/java/org/sonar/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -19,29 +19,107 @@
...
+    Optional<JavaQuickFix> conciseQuickFix = computeConciseQuickFix(tree);
+    if (conciseQuickFix.isPresent()) {
+      return Collections.singletonList(conciseQuickFix.get());
+    }
+    return Collections.singletonList(computeDefaultQuickFix(tree));
```
quentin-jaquier-sonarsource, line 72: If you return a JavaQuickFix directly, the code can become:
```
return computeConciseQuickFix(tree).orElseGet(() -> computeDefaultQuickFix(tree));
```

java-checks/src/main/java/org/sonar/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -19,29 +19,107 @@
...
+      }
+      JavaQuickFix.Builder quickFix = JavaQuickFix.newQuickFix(QUICK_FIX_MESSAGE)
+        .addTextEdit(JavaTextEdit.insertAfterTree(tree.rightOperand().lastToken(), ")"))
+        .addTextEdit(JavaTextEdit.replaceTextSpan(interOperandSpace, DOT_EQUALS_AND_OPENING_PARENTHESIS));
+      if (tree.is(Tree.Kind.NOT_EQUAL_TO)) {
+        quickFix.addTextEdit(JavaTextEdit.insertBeforeTree(tree.leftOperand().firstToken(), "!"));
+      }
+      return Optional.of(quickFix.build());
+    } else if (tree.rightOperand().is(Tree.Kind.STRING_LITERAL)) {
+      String callEqualsOnLiteral = ((LiteralTree) tree.rightOperand()).value() + DOT_EQUALS_AND_OPENING_PARENTHESIS;
```
quentin-jaquier-sonarsource, line 95: You are calling multiple times `tree.leftOperand()` and `tree.rightOperand()`, you could extract them into a variable before the if/else if.

java-checks/src/main/java/org/sonar/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -19,29 +19,107 @@
...
+      String callToEquals = tree.is(Tree.Kind.NOT_EQUAL_TO) ? ("!" + callEqualsOnLiteral) : callEqualsOnLiteral;
+      AnalyzerMessage.TextSpan leftOfOperatorToEndOfComparison = AnalyzerMessage.textSpanBetween(
+        tree.leftOperand().lastToken(), false,
+        tree.rightOperand().lastToken(), true
+      );
+      return Optional.of(
+        JavaQuickFix.newQuickFix(QUICK_FIX_MESSAGE)
+          .addTextEdit(JavaTextEdit.removeTextSpan(leftOfOperatorToEndOfComparison))
+          .addTextEdit(JavaTextEdit.insertAfterTree(tree.leftOperand().lastToken(), ")"))
```
quentin-jaquier-sonarsource, line 104: You can use `replaceTextSpan` to do the two steps in one edit.

java-checks/src/main/java/org/sonar/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
```diff
@@ -19,29 +19,107 @@
...
+      AnalyzerMessage.TextSpan interOperandSpace = AnalyzerMessage.textSpanBetween(
+        tree.leftOperand().lastToken(), false,
+        tree.rightOperand().firstToken(), false
+      );
+      List<JavaTextEdit> edits = new ArrayList<>();
+      edits.add(JavaTextEdit.insertAfterTree(tree.rightOperand().lastToken(), ")"));
+      edits.add(JavaTextEdit.replaceTextSpan(interOperandSpace, DOT_EQUALS_AND_OPENING_PARENTHESIS));
+      if (tree.is(Tree.Kind.NOT_EQUAL_TO)) {
+        edits.add(JavaTextEdit.insertBeforeTree(tree.leftOperand().firstToken(), "!"));
+      }
```
quentin-jaquier-sonarsource, line 86: Useless code?

quentin-jaquier-sonarsource (changes requested):

java-checks-test-sources/src/main/java/checks/CompareStringsBoxedTypesWithEqualsCheck.java
dorian-burihabwa-sonarsource: I am not sure I understand this comment

quentin-jaquier-sonarsource: When looking at the diff, the whole file will appear as new code, while you only modified part of it. To avoid this, you should not nest everything into another class.

quentin-jaquier-sonarsource: As discussed, the main point is to avoid name clashes in the test module, therefore to not have names like "Boxed" at the top level. Both the current code and my suggestion do the trick. which one is not so important in the end, it is fine as it is.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3754, opened 2021-08-23
SONARJAVA-3925 Add quick fixes for S1197 (ArrayDesignatorOnVariableCheck)

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/ArrayDesignatorOnVariableCheck.java
```diff
@@ -45,18 +51,61 @@ public void visitNode(Tree tree) {
...
       ArrayTypeTree arrayTypeTree = (ArrayTypeTree) type;
-      SyntaxToken arrayDesignatorToken = arrayTypeTree.ellipsisToken();
-      if (arrayDesignatorToken == null) {
-        arrayDesignatorToken = arrayTypeTree.openBracketToken();
-      }
-      if (isInvalidPosition(arrayDesignatorToken, identifierToken)) {
-        reportIssue(arrayDesignatorToken, "Move the array designator from the variable to the type.");
+      SyntaxToken openBracketToken = arrayTypeTree.openBracketToken();
+      if (openBracketToken != null && isInvalidPosition(openBracketToken, identifierToken)) {
+        ((InternalJavaIssueBuilder) ((DefaultJavaFileScannerContext) context).newIssue())
```
m-g-sonar, line 56: this could benefit from the `QuickFixHelper` utility class. :)

java-checks/src/main/java/org/sonar/java/checks/ArrayDesignatorOnVariableCheck.java
```diff
@@ -45,18 +51,61 @@ public void visitNode(Tree tree) {
...
+          arrayTypeTree.closeBracketToken(),
+          ""))
+        .addTextEdit(JavaTextEdit.insertAfterTree(
+          arrayTypeTree.type(),
+          "[]"))
+        .build());
+  }
+
+  private static boolean isDeclarationTypeUsedBySeveralVariable(VariableTree current) {
+    Tree parent = current.parent();
```
m-g-sonar, line 85: I would cast the parent into a `JavaTree`, then you can call `children()`

java-checks/src/main/java/org/sonar/java/checks/ArrayDesignatorOnVariableCheck.java
```diff
@@ -45,18 +51,61 @@ public void visitNode(Tree tree) {
...
+    } else if (parent instanceof BlockTreeImpl) {
+      children = ((BlockTreeImpl) parent).children();
+    } else if (parent.is(Tree.Kind.METHOD)) {
+      children = ((MethodTree) parent).parameters();
+    } else if (parent.is(Tree.Kind.LIST)) {
+      children = (List<? extends Tree>) parent;
+    }
+    int index = children.indexOf(current);
+    if (index == -1) {
+      // FOR_EACH_STATEMENT has no list of children
```
m-g-sonar, line 98: i think this would not be the case with the javaTree approach.

java-checks/src/main/java/org/sonar/java/checks/ArrayDesignatorOnVariableCheck.java
alban-auzeill: It makes sense for `BlockTree` and Tree.Kind.LIST but for `ClassTree` it's not efficient, instead of directly iterating the `members` you will previously iterate:
```
modifiers, atToken, declarationKeyword, simpleName, typeParameters..., recordComponents..., extendsKeyword,
superClass, implementsKeyword, superInterfaces..., permittedTypes, openBraceToken, and finally members
```
Same for `MethodTree` with `modifiers, typeParameters, returnType, simpleName, openParenToken, and finally members`

java-checks/src/main/java/org/sonar/java/checks/ArrayDesignatorOnVariableCheck.java
```diff
@@ -45,18 +50,55 @@ public void visitNode(Tree tree) {
...
+
+  private static boolean isDeclarationTypeUsedBySeveralVariable(VariableTree current) {
+    Tree parent = current.parent();
+    List<? extends Tree> children;
+    if (parent instanceof ClassTree) {
+      children = ((ClassTree) parent).members();
+    } else if (parent.is(Tree.Kind.METHOD)) {
+      children = ((MethodTree) parent).parameters();
+    } else {
+      children = ((JavaTree)parent).getChildren();
```
m-g-sonar, line 91: missing a space between the cast and `parent`

PR #3755, opened 2021-08-23
SONARJAVA-3940 Add quick fixes for S1858 (StringToStringCheck)

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/files/non-compiling/checks/TextBlockToString.java
```diff
@@ -6,6 +6,8 @@
               
   hello
   
-  """.toString(); // Noncompliant@-4 [[sc=19;ec=6]] {{there's no need to call "toString()" on a text block.}}
+  """.toString(); // Noncompliant@-4 [[sc=19;ec=6;quickfixes=qf1]] {{there's no need to call "toString()" on a text block.}}
```
dorian-burihabwa-sonarsource, line 9: We should add the starting and ending lines for the issue for as there might be some confusion due to ec < sc

java-checks-test-sources/src/main/files/non-compiling/checks/TextBlockToString.java
```diff
@@ -6,6 +6,8 @@
...
+  // fix@qf1 {{Remove "toString()"}}
+  // edit@qf1 [[sc=6;ec=17]] {{}}
 
 }
```
dorian-burihabwa-sonarsource, line 13: We should move this entire file to the compiling sources since we now support Java 16

dorian-burihabwa-sonarsource (changes requested):

java-checks-test-sources/src/main/files/non-compiling/checks/TextBlockToString.java
quentin-jaquier-sonarsource: Indeed :)

java-checks-test-sources/src/main/files/non-compiling/checks/TextBlockToString.java
quentin-jaquier-sonarsource: The starting line is defined thanks to the `@+4`, but we should indeed check the end line here.

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #3757, opened 2021-08-23
SONARJAVA-3945 Add quick fixes for S1107 (RightCurlyBraceSameLineAsNextBlockCheck)

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/RightCurlyBraceSameLineAsNextBlockCheck.java
```diff
@@ -20,18 +20,30 @@
...
   @Override
   protected void checkTokenPosition(SyntaxToken syntaxToken, BlockTree previousBlock) {
     if (syntaxToken.line() != previousBlock.closeBraceToken().line()) {
-      reportIssue(syntaxToken, "Move this \"" + syntaxToken.text() + "\" on the same line that the previous closing curly brace.");
+      QuickFixHelper.newIssue(context)
+        .forRule(this)
+        .onTree(syntaxToken)
+        .withMessage("Move this \"%s\" on the same line that the previous closing curly brace.", syntaxToken.text())
+        .withQuickFix(() -> JavaQuickFix.newQuickFix("Move to the same line as the closing curly brace")
+          .addTextEdit(JavaTextEdit.replaceTextSpan(textSpanBetween(previousBlock, false, syntaxToken, false), " ")
```
dorian-burihabwa-sonarsource, line 44: What happens if there is a comment between the closing curly brace and the next block?

java-checks-test-sources/src/main/java/checks/RightCurlyBraceSameLineAsNextBlockCheck.java
```diff
@@ -0,0 +1,70 @@
...
+    else {                              // Compliant
+      generateError();
+    }
+    {
+      try {
+        generateOrder();
+      }
+      catch (RuntimeException e) {         // Noncompliant [[sc=7;ec=12;quickfixes=qf2]] {{Move this "catch" on the same line that the previous closing curly brace.}}
+                                           // fix@qf2 {{Move to the same line as the closing curly brace}}
+                                           // edit@qf2 [[sl=-1;sc=8;ec=7]] {{ }}
```
dorian-burihabwa-sonarsource, line 31: Adding the end line  would help in understanding the span of the quick fix

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/RightCurlyBraceSameLineAsNextBlockCheck.java
quentin-jaquier-sonarsource: As discussed, the comment will be dropped. In fact, this behavior is the same in all rules where we remove a range, if a comment is present, it will be removed. We might want to do something about it in the future, but for now, I think it's fine, the user can anyway revert the change if needed.

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #3759, opened 2021-08-23
SONARJAVA-3998 Add quick fixes for S1195 (ArrayDesignatorAfterTypeCheck)

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/ArrayDesignatorAfterTypeCheck.java
```diff
@@ -47,13 +49,25 @@ public void visitNode(Tree tree) {
...
+          .withQuickFix(() -> createQuickFixes(arrayTypeTree))
+          .report();
         break;
       }
       returnType = arrayTypeTree.type();
     }
   }
 
+  private static JavaQuickFix createQuickFixes(ArrayTypeTree type) {
+    return JavaQuickFix.newQuickFix("Move [] to the method type")
```
m-g-sonar, line 65: I would prefer `"Move [] to the return type"`

java-checks/src/main/java/org/sonar/java/checks/ArrayDesignatorAfterTypeCheck.java
```diff
@@ -47,13 +49,25 @@ public void visitNode(Tree tree) {
...
+      .addTextEdit(JavaTextEdit.replaceBetweenTree(type.openBracketToken(), type.closeBracketToken(), ""))
```
m-g-sonar, line 66: Maybe we could add a new method `JavaTextEdit.removeBetweenTree(Tree start, Tree end)` to avoid having to type `""` all the time.

PR #3760, opened 2021-08-23
SONARJAVA-3939 Add quick fixes for S1116 (EmptyStatementUsageCheck)

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (changes requested):
LGTM but I would add some explicit start and end line information for quickfixes that span multiple lines

dorian-burihabwa-sonarsource (approved):
LGTM!

PR #3761, opened 2021-08-24
SONARJAVA-3942 Add quick fixes for S2209 (StaticMembersAccessCheck)

alban-auzeill:


java-checks-test-sources/src/main/java/checks/StaticMembersAccessCheck.java
```diff
@@ -33,18 +36,38 @@ public StaticMembersAccessCheckD d() {
...
+    // fix@qf2 {{Replace "second" by "StaticMembersAccessCheckA"}}
+    // edit@qf2 [[sc=5;ec=11]] {{StaticMembersAccessCheckA}}
     third.counter ++; // Noncompliant
     first.d.counter++; // Noncompliant
-    first.c.d.counter++; // Noncompliant
-    first.d().counter++; // Noncompliant
+    first.c.d.counter++; // Noncompliant [[sc=5;ec=22;quickfixes=qf3]]
+    // fix@qf3 {{Replace "first.c.d" by "StaticMembersAccessCheckD"}}
+    // edit@qf3 [[sc=5;ec=14]] {{StaticMembersAccessCheckD}}
+    first.d().counter++; // Noncompliant // Noncompliant [[sc=5;ec=22;quickfixes=qf4]]
```
m-g-sonar, line 51: typo? or two issues are expected?
I would explicitly mark one on another line with `//Noncompliant@+1

java-checks-test-sources/src/main/java/checks/StaticMembersAccessCheck.java
```diff
@@ -33,18 +36,38 @@ public StaticMembersAccessCheckD d() {
...
+    // fix@qf4 {{Replace "first.d()" by "StaticMembersAccessCheckD"}}
+    // edit@qf4 [[sc=5;ec=14]] {{StaticMembersAccessCheckD}}
     d().counter++; // Noncompliant
-    ((StaticMembersAccessCheckA.StaticMembersAccessCheckD) d()).counter++; // Noncompliant [[sc=5;ec=72]]
+    // Noncompliant@+1 [[sc=5;el=+3;ec=14;quickfixes=qf5]]
+    (
+      (StaticMembersAccessCheckA.StaticMembersAccessCheckD) d()
+    ).counter++;
+    // fix@qf5 {{Replace "( (StaticMembersAccessCheckA.StaticMembersAccessCheckD) d() )" by "StaticMembersAccessCheckD"}}
```
m-g-sonar, line 59: quickfix messages should not be too long. I feel here it can escalate fastly. What about using the following message instead when it's not a "simple case": `Replace the expression by "StaticMembersAccessCheckD"` I feel it would make it way nicer.

java-checks/src/main/java/org/sonar/java/checks/StaticMembersAccessCheck.java
```diff
@@ -19,35 +19,75 @@
...
+          .onTree(memberSelect)
+          .withMessage("Change this instance-reference to a static reference.")
+          .withQuickFix(() -> createQuickFixes(leftOperand, memberSelectSymbol.owner().type()))
+          .report();
       }
     }
   }
+
+  private JavaQuickFix createQuickFixes(ExpressionTree leftOperand, Type type) {
+    String leftOperandAsText = QuickFixHelper.contentForTree(leftOperand, context).replaceAll("\\s+", " ");
```
m-g-sonar, line 80: I think this should be done in a simpler way. If it's an identifier (or a member select) we just use its name, if not, we simply mark "this expression"

java-checks-test-sources/src/main/java/checks/StaticMembersAccessCheck.java
alban-auzeill: it's a typo

java-checks-test-sources/src/main/java/checks/StaticMembersAccessCheck.java
```diff
@@ -46,26 +46,26 @@ public void noncompliant() {
     third.counter ++; // Noncompliant
     first.d.counter++; // Noncompliant
     first.c.d.counter++; // Noncompliant [[sc=5;ec=22;quickfixes=qf3]]
-    // fix@qf3 {{Replace "first.c.d" by "StaticMembersAccessCheckD"}}
+    // fix@qf3 {{Replace the expression by "StaticMembersAccessCheckD"}}
```
m-g-sonar, line 49: In fact, looking at where the main issue is located, I feel it might not be the best phrasing (sorry I missed that in my first review).
What about using
`Replace by a static reference to "StaticMembersAccessCheckD"` 
instead of 
`Replace the expression by "StaticMembersAccessCheckD"`

Because we highlight the whole expression, it sounds strange.

Now, if you want to try it first and see how it goes once dogfooded, it's OK for me as well.

PR #3764, opened 2021-08-24
SONARJAVA-3926 Add quick fixes for S1125 (BooleanLiteralCheck)

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
In my opinion, the 2 code smells should be resolved as "won't fix", the Boolean values are checked for null before in the logic, it seems redundant to use `Boolean.TRUE.equals(b)` here.

java-checks/src/main/java/org/sonar/java/checks/BooleanLiteralCheck.java
```diff
@@ -77,4 +93,206 @@ public void visitNode(Tree tree) {
...
+    } else if (right != null) {
+      // Defensive programming, if we reached this point, right must be a boolean literal
+      addForConditionalWhenRightIsLiteral(edits, right, leftOperand, rightOperand, conditionalOr);
+    }
+
+    return edits;
+  }
+
+  private static void addEditsConditionalWhenBothLiterals(BinaryExpressionTree tree, List<JavaTextEdit> edits, Boolean left, Boolean right, boolean conditionalOr) {
+    boolean replacement =
```
dorian-burihabwa-sonarsource, line 250: Following the flow of the check, we can understand that `!conditionalOr` means that the operator is `&&`.
However, I would make this more explcit by:
1. Renaming the parameter to `operator`
2. Explicitly extracting the value of the operator as two variables `boolean orOperator = operator, andOperator = !operator;`

dorian-burihabwa-sonarsource: The variable replacement can also be renamed to `value`

java-checks/src/main/java/org/sonar/java/checks/BooleanLiteralCheck.java
```diff
@@ -77,4 +93,206 @@ public void visitNode(Tree tree) {
...
+    boolean found = false;
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree tree) {
+      found = true;
+    }
+
+  }
+
+  @Nullable
+  private static Boolean getBooleanValue(Tree expression) {
```
dorian-burihabwa-sonarsource, line 291: We can move this method higher in the file closer to where it is first used

dorian-burihabwa-sonarsource (changes requested):
First reviewing pass over the code, the code added by this PR looks like a good candidate to a independent class that exposes a `JavaQuickFix proposeQuickFix(Tree tree)` method.

A part from that, here are a few comments
* The extensive test cases are really nice to validate assumptions
* A number of methods that were extracted `add...` could benefit from having their signature changed to return values in order to reduce the number of parameters

PR #3765, opened 2021-08-24
SONARJAVA-3940 Quick-Fix for S1858: Do not remove parenthesis after the expression

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM

PR #3774, opened 2021-08-27
SONARJAVA-3944 Make S1168 correctly handle arrays and missing imports when needed

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
```diff
@@ -118,4 +136,99 @@ public static String contentForTree(Tree tree, JavaFileScannerContext context) {
 
     return sb.toString();
   }
+
+  public static Optional<JavaTextEdit> newImportEdit(String requiredType, Tree base) {
```
quentin-jaquier-sonarsource, line 140: What does `base` means in this context?

java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
```diff
@@ -118,4 +136,99 @@ public static String contentForTree(Tree tree, JavaFileScannerContext context) {
...
+    if (!requiresImportOf(requiredType, base)) {
+      return Optional.empty();
+    }
+
+    CompilationUnitTree cut = ((JavaTree) base).root();
```
quentin-jaquier-sonarsource, line 145: The first step of the method `requiresImportOf` is to compute this value, you could give this `cut` as argument, and avoid computing it a second time.

java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
```diff
@@ -118,4 +136,99 @@ public static String contentForTree(Tree tree, JavaFileScannerContext context) {
...
+      // no package declaration and no type? Should be impossible or an empty file
+      return Optional.empty();
+    }
+
+    List<ImportTree> nonStaticImports = imports
+      .stream()
+      .filter(importClauseTree -> importClauseTree.is(Tree.Kind.IMPORT))
+      .map(ImportTree.class::cast)
+      .filter(importTree -> !importTree.isStatic())
+      .sorted(IMPORT_TREE_COMPARATOR)
```
quentin-jaquier-sonarsource, line 169: I'm wondering if we should try to find something smarter than sorting the whole list of imports for every call to this method, "just" to find the closest matching import or if it would be overengineering because it will be called only on rare occasions when we have an issue with a quick fix?

quentin-jaquier-sonarsource (changes requested):
The QA fails on windows, my guess is that it is due to the use of `%n` in `String.format`.

java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
m-g-sonar: It was to be used as "any base tree to start from", but in fact, we don't need it, we can simply rely on the context, which has an access to the CompilationUnitTree.

java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
m-g-sonar: it is not "computed" per se, as it's always available. But I agree it's not needed.

java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
m-g-sonar: to be honest, I feel it's OK to do it every time. I also don't see any "smart" way to do it only a few times.

m-g-sonar:
I dropped the commit adding visibility to `root()`, as we don't need it.

quentin-jaquier-sonarsource (approved):
LGTM, nice work!

PR #3776, opened 2021-08-30
SONARJAVA-3955 Automatically add import instead of fully qualified name

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
@dorian-burihabwa-sonarsource FYI

PR #3778, opened 2021-08-31
SONARJAVA-4003 Fix Deadlock on ProgressMonitor

quentin-jaquier-sonarsource:


java-frontend/src/main/java/org/sonar/java/ProgressMonitor.java
```diff
@@ -39,8 +40,19 @@
...
   private int processedWork = 0;
 
+  /**
+   * The report loop can not rely only on Thread.interrupted() to end, according to
+   * interrupted() javadoc, a thread interruption can be ignored because a thread was
+   * not alive at the time of the interrupt. This could happen if done() is being called
+   * before ProgressReport's thread becomes alive.
+   * So this boolean flag ensures that ProgressReport never enter an infinite loop when
+   * Thread.interrupted() failed to be set to true.
+   */
```
alban-auzeill, line 50: You should replace 2 times `ProgressReport` with `ProgressMonitor` if you don't want that someone notices your copy and paste :wink:

alban-auzeill (approved):

java-frontend/src/main/java/org/sonar/java/ProgressMonitor.java
quentin-jaquier-sonarsource: 😅

PR #3779, opened 2021-08-31
SONARJAVA-3991 S6212 should not suggest using "var" when the initializer is an array initializers

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/java/checks/VarCanBeUsedCheck.java
```diff
@@ -66,9 +66,14 @@ void f(int[] array) {
...
 
     byte[] input = new byte[array.length]; // Noncompliant
-    
     var input2 = new byte[array.length]; // Compliant
 
+    String[] arrayInit = new String[]{"A", "B", "C"}; // Noncompliant
+    var arrayInit2 = new String[]{"A", "B", "C"}; // Compliant
+
+    String[] arrayInit3 = {"A", "B", "C"}; // Compliant, for array initializer, without type, it is not possible to change to var
+    //var arrayInit4 = (String[]) {"A", "B", "C"}; // does not compile
```
dorian-burihabwa-sonarsource, line 75: Ift this line does not compile, we should consider moving it to a non-compiling file or deleting it altogether
```suggestion
```

dorian-burihabwa-sonarsource (changes requested):
The PR is good but we should consider removing the commented example in the test source file

dorian-burihabwa-sonarsource (approved):

PR #3780, opened 2021-08-31
SONARJAVA-4004 Extend QuickFixHelper with nextVariable and previousVariable

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
```diff
@@ -102,6 +113,72 @@ public static SyntaxToken previousToken(Tree tree) {
...
+   *
+   * @param current The target variable
+   * @return An Optional with the preceding variable, Optional.empty otherwise.
+   */
+  public static Optional<VariableTree> previousVariable(VariableTree current) {
+    Tree parent = current.parent();
+    if (parent.is(Tree.Kind.BLOCK, Tree.Kind.INITIALIZER, Tree.Kind.STATIC_INITIALIZER)) {
+      List<Tree> children = ((BlockTreeImpl) parent).children();
+      return previousVariable(current, children, TRIMMING_OFFSET_IN_BLOCK_CHILDREN);
+    } else if (parent.is(Tree.Kind.CLASS)) {
```
alban-auzeill, line 127: There's not only CLASS but ENUM, INTERFACE, ANNOTATION_TYPE, RECORD (all instances of ClassTree)
And for RECORD the variable can be in `recordComponents()` or `members()`. If the variable is in `recordComponents()`, it has no previous variable or next variable.

java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
```diff
@@ -102,6 +113,72 @@ public static SyntaxToken previousToken(Tree tree) {
...
+      List<Tree> members = ((ClassTree) parent).members();
+      return previousVariable(current, members, TRIMMING_OFFSET_IN_CLASS_MEMBERS);
+    } else {
```
alban-auzeill, line 130: Warning, there's also `parent.is(Tree.Kind.METHOD)` where we could probably return  no previous variable or next variable. No need to look in `((MethodTree) parent).parameters()`
Same for `parent.is(Tree.Kind.LIST)`

java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
```diff
@@ -102,6 +113,72 @@ public static SyntaxToken previousToken(Tree tree) {
...
   }
 
+  /**
+   * Returns the preceding variable in a mutli-variable declaration.
+   *
+   * @param current The target variable
+   * @return An Optional with the preceding variable, Optional.empty otherwise.
+   */
+  public static Optional<VariableTree> previousVariable(VariableTree current) {
+    Tree parent = current.parent();
```
alban-auzeill, line 123: It would be simpler to maintain if the logic to find children was in a common method used by previousVariable and nextVariable. The method could return a list of trees without the `firstRelevantIndex`, it'ŝ not so useful.

java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
```diff
@@ -102,6 +113,72 @@ public static SyntaxToken previousToken(Tree tree) {
...
+    // If the variable is the first element that follows the opening token, there is no predecessor to return
+    if (currentIndex <= firstRelevantIndex) {
+      return Optional.empty();
+    }
+    // If there is a predecessor, we check that it is a variable and that it is part of the same declaration
+    Tree preceding = trees.get(currentIndex - 1);
+    if (preceding.is(Tree.Kind.VARIABLE) && preceding.firstToken().equals(current.firstToken())) {
+      return Optional.of((VariableTree) preceding);
+    }
+    return Optional.empty();
```
alban-auzeill, line 146: You can probably write:
```java
return Optional.of(trees.get(currentIndex - 1))
  .filter(preceding -> preceding.is(Tree.Kind.VARIABLE) && preceding.firstToken().equals(current.firstToken()))
  .map(VariableTree.class::cast);
```
(disclaimer: code written without IDE)

PR #3783, opened 2021-08-31
SONARJAVA-3925 S1195 & S1197 Improve quick fix to support multiple array designators at once

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/ArrayDesignatorOnVariableCheck.java
```diff
@@ -94,8 +89,41 @@ private static boolean isVariableDeclarationOfTheSameType(VariableTree variable,
     return otherTree.is(Tree.Kind.VARIABLE) && variable.firstToken().equals(otherTree.firstToken());
   }
 
-  private static boolean isInvalidPosition(SyntaxToken arrayDesignatorToken, SyntaxToken identifierToken) {
-    return identifierToken.range().start().isBefore(arrayDesignatorToken.range().start());
+  static boolean isInvalidPosition(ArrayTypeTree arrayTypeTree, SyntaxToken identifierToken) {
```
quentin-jaquier-sonarsource, line 92: Should be private inside `MisplacedArray`.

java-checks/src/main/java/org/sonar/java/checks/ArrayDesignatorOnVariableCheck.java
```diff
@@ -94,8 +89,41 @@ private static boolean isVariableDeclarationOfTheSameType(VariableTree variable,
...
+    SyntaxToken openBracketToken = arrayTypeTree.openBracketToken();
+    return openBracketToken != null && identifierToken.range().start().isBefore(openBracketToken.range().start());
+  }
+
+  static class MisplacedArray {
+    ArrayTypeTree firstArray;
+    ArrayTypeTree lastArray;
+    int misplacedCount;
```
quentin-jaquier-sonarsource, line 100: What about storing in the field the replacement as a String directly, instead of the `misplacedCount`, to avoid logic in `replacement()`?

quentin-jaquier-sonarsource (approved):
LGTM

PR #3786, opened 2021-09-01
SONARJAVA-3964 S1612 should not suggest method reference casting when the type is a primitive

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
After discussion, the fix looks good to me

PR #3788, opened 2021-09-01
SONARJAVA-3986 FP in S2583 with Java 16 pattern matching

quentin-jaquier-sonarsource:
For:
```
if (x instanceof String) {}
```
The basic block of the CFG looks like this:

- IDENTIFIER (x)
- INSTANCE_OF
- IF_STATEMENT

When the ExplodedGraphWalker reaches "INSTANCE_OF", the symbolic value of "x" will be on top of the stack. We will wrap it into a "InstanceOfSymbolicValue". (This new symbolic value is not doing much, as we currently do not track the types constraints.) This new symbolic value will replace the top of the stack. "IF_STATEMENT" will then consume this symbolic value.

For:
```
if (x instanceof String str) {} else if (x instanceof Integer i) {}
```
the corresponding cfg is:
- IDENTIFIER (x)
- PATTERN_INSTANCE_OF
- IF_STATEMENT
- IDENTIFIER (x)
- PATTERN_INSTANCE_OF
- IF_STATEMENT

The problem is that we are doing nothing when PATTERN_INSTANCE_OF is reached. It means that the first "IF_STATEMENT" will consume the symbolic value of "x", and assign the constraint `false` in the false branch. When the second "IF_STATEMENT" is reached, "x" has the constraint "false", generating a false positive.

The first commit solves this problem (and the problem of the ticket) by treating "PATTERN_INSTANCE_OF" exactly as "INSTANCE_OF". It makes sense because the two trees has the same impact on "x". Note that it does nothing special for the new variable "str".

The second commit is an improvement suggestion, to remove false negatives in the same rule (and potentially in others): it will set the symbolic value of "str" to the one from "x".

### DIFF
--- a/java-checks-test-sources/src/main/files/non-compiling/symbolicexecution/checks/ConditionAlwaysTrueOrFalseCheck.java
+++ b/java-checks-test-sources/src/main/files/non-compiling/symbolicexecution/checks/ConditionAlwaysTrueOrFalseCheck.java
@@ -250,6 +250,14 @@ class Class extends SuperClass {
     }
   }
 
+  public void instanceOfPatternMathing() {
+    Object object = new Object();
+    // Java 16 pattern matching instance of
+    if (object instanceof String s) { // Compliant
+    } else if (object instanceof Integer i) { // Compliant
+    }
+  }
+
   public void literals() {
     // literals
     if (!true) { // Noncompliant {{Change this condition so that it does not always evaluate to "false"}}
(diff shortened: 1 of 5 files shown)


quentin-jaquier-sonarsource:
@alban-auzeill let me know if you think we should go for the second commit, or if the first one is already enough!
Note that we should probably create a ticket if we keep the second one.

PR #3799, opened 2021-09-22
Fix dead link in custom rule tutorial

quentin-jaquier-sonarsource:


docs/CUSTOM_RULES_101.md
```diff
@@ -543,7 +543,7 @@ Check this example: [SecurityAnnotationMandatoryRule.java](https://github.com/So
 
 In the `pom.xml`, define in the `Maven Dependency Plugin` part all the JARs you need to run your Unit Tests. For example, if you sample code used in your Unit Tests is having a dependency on Spring, add it there.
 
-See: [pom.xml#L133-L187](https://github.com/SonarSource/sonar-java/blob/custom_rules_example/docs/java-custom-rules-example/pom_SQ_8_9_LTS.xml#L133-L187)
+See: [pom.xml#L137-L197](https://github.com/SonarSource/sonar-java/blob/master/docs/java-custom-rules-example/pom_SQ_8_9_LTS.xml#L137-L197)
```
dorian-burihabwa-sonarsource, line 546: Could we replace this with a relative link to avoid having the same issue in the future?

```suggestion
See: [pom.xml#L137-L197](./java-custom-rules-example/pom_SQ_8_9_LTS.xml#L137-L197)
```

dorian-burihabwa-sonarsource (changes requested):
We could change the link to be relative so that we are not tied to a specific branch in the future

docs/CUSTOM_RULES_101.md
quentin-jaquier-sonarsource: Sure, good idea!

dorian-burihabwa-sonarsource (approved):
LGTM!

PR #3800, opened 2021-09-27
SONARJAVA-3868 Rule S6288: Authorizing non-authenticated users to use keys in the Android KeyStore is security-sensitive

quentin-jaquier-sonarsource:
The two commits of this PR should be rebased separately into master.


quentin-jaquier-sonarsource:
Before this PR, we already had rules using android types. The unit tests for these rules are working thanks to implicit and undocumented tricks:
We are redefining android classes in `sonar-java\java-checks\src\test\java\android`. Then, the CheckVerifier adds (implicitly) the classes of this package into the classpath.

The first commit aims to improve the situation:
I don't think we can do much about redefining android classes, it is not convenient to add these dependencies, especially in a maven context. However, having them in the tests of `java-checks` is not clean, we should move them into `java-checks-test-sources`. Then, I think it's cleaner if we explicitly mention that we will use these "fake" android classes in the unit tests:

```
    CheckVerifier.newVerifier()
      .onFile(testSourcesPath("checks/security/ReceivingIntentsCheck.java"))
      .withCheck(new ReceivingIntentsCheck())
      .withClassPath(ANDROID_FAKE_CLASS_PATH)
      .verifyIssues();
```

mpaladin (approved):

PR #3801, opened 2021-09-29
SONARJAVA-3866 Rule S6293: Using a biometric authentication independent of a cryptographic solution is security-sensitive

mpaladin:


quentin-jaquier-sonarsource (approved):
LGTM

PR #3802, opened 2021-09-29
SONARJAVA-4028 Extend S5324 to support direct access to properties

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
Unlike Kotlin, in Java is it not possible to call the properties directly.

PR #3804, opened 2021-09-30
SONARJAVA-4031 Rule S6300: Using unencrypted files in mobile applications is security-sensitive

quentin-jaquier-sonarsource:


java-frontend/src/main/java/org/sonar/java/classpath/AbstractClasspath.java
```diff
@@ -130,6 +131,9 @@ static void logResolvedFiles(String property, Collection<File> files) {
         }
         validateLibraries = validateLibs;
         result.addAll(libraryFilesForPattern);
+        if (pathPattern.endsWith("android.jar")) {
+          inAndroidContext = true;
+        }
```
quentin-jaquier-sonarsource, line 136: The important lines of the first commit are here, all the rest is just passing the value around and adding the possibility to test in unit tests.

mpaladin (changes requested):

mpaladin (approved):

PR #3805, opened 2021-09-30
SONARJAVA-4029 Rule S6301: Mobile database encryption keys should not be disclosed

mpaladin:


java-checks/src/main/java/org/sonar/java/checks/security/AndroidMobileDatabaseEncryptionKeysCheck.java
```diff
@@ -0,0 +1,121 @@
...
+    return MethodMatchers.or(
+      SQLITE_DATABASE_CONSTRUCTOR,
+      SQLITE_DATABASE_METHODS,
+      REALM_CONFIGURATION_BUILDER_ENCRYPTION_KEY);
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    Arguments arguments = mit.arguments();
+    String messageArg = REALM_CONFIGURATION_BUILDER_ENCRYPTION_KEY.matches(mit) ? "encryptionKey" : "password";
```
quentin-jaquier-sonarsource, line 77: The downside of using `AbstractMethodDetection` is that you don't know which matcher matched. It means that you have to apply this matcher a second time. You can go for an `IssuableSubscriptionVisitor` and avoid redundant computation.

java-checks/src/main/java/org/sonar/java/checks/security/AndroidMobileDatabaseEncryptionKeysCheck.java
```diff
@@ -0,0 +1,121 @@
...
+    ExpressionTree passwordArg = arguments.size() == 1 ? arguments.get(0) : arguments.get(1);
```
quentin-jaquier-sonarsource, line 78: You are using `withAnyParameters()` in the matchers, it means that you have no guarantee that the argument list is not empty here. Even if there is no method without arguments currently, I would rather explicitly make sure we do not crash, by making sure the argument list is not empty.

Note: if you think specifying the types of every parameter is overkill, you can use `MethodMatchers.ANY`:
```
.addParametersMatcher(MethodMatchers.ANY)
.addParametersMatcher(MethodMatchers.ANY, MethodMatchers.ANY)
```
Alternatively, you can also play with the method taking a predicate to make sure the list is not empty:
```
.addParametersMatcher(l -> !l.isEmpty())
```

java-checks/src/main/java/org/sonar/java/checks/security/AndroidMobileDatabaseEncryptionKeysCheck.java
```diff
@@ -0,0 +1,121 @@
...
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.NewClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+import static org.sonar.java.checks.helpers.ExpressionsHelper.getSingleWriteUsage;
+
+@Rule(key = "S6301")
+public class AndroidMobileDatabaseEncryptionKeysCheck extends AbstractMethodDetection {
```
quentin-jaquier-sonarsource, line 38: I realized that the behavior of AbstractMethodDetection is not ideal: it will visit `Tree.Kind.METHOD_INVOCATION, Tree.Kind.NEW_CLASS, Tree.Kind.METHOD_REFERENCE` and run the matchers on all of them, despite the fact that most of the matchers only target method reference.
This situation is not new, but together with the other remark on AbstractMethodDetection, I believe we should not use it here.

java-checks/src/main/java/org/sonar/java/checks/security/AndroidMobileDatabaseEncryptionKeysCheck.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  @CheckForNull
+  private static ExpressionTree tryGetOriginStringFromByteOrCharArray(ExpressionTree givenExpression) {
+    ExpressionTree expression = givenExpression;
+    if (expression.is(Tree.Kind.IDENTIFIER)) {
+      IdentifierTree identifier = (IdentifierTree) expression;
+      final ExpressionTree singleWriteUsage = getSingleWriteUsage(identifier.symbol());
+      if (singleWriteUsage != null && singleWriteUsage.is(Tree.Kind.METHOD_INVOCATION)) {
```
quentin-jaquier-sonarsource, line 108:
```suggestion
      if (singleWriteUsage != null) {
```
The second part of the binary is redundant: it will be checked at line 112.

java-checks-test-sources/src/main/java/checks/security/AndroidMobileDatabaseEncryptionKeysCheck.java
```diff
@@ -0,0 +1,78 @@
...
+  String getPassword() {
+    return null;
+  }
+
+  void SQLiteDatabase_openOrCreateDatabase(String databasePath, SQLiteDatabaseHook hook, SQLiteDatabase.CursorFactory cursorFactory) {
+    String password = "pwd";
+    SQLiteDatabase database;
+
+    database = new SQLiteDatabase("path", "pwd".toCharArray(), cursorFactory, 0); // Noncompliant
+    database = new SQLiteDatabase("path", password.toCharArray(), cursorFactory, 0); // Noncompliant
```
quentin-jaquier-sonarsource, line 21:
```
char[] a = "pwd".toCharArray();
database = new SQLiteDatabase("path", a, cursorFactory, 0);
```
We should add a test for this situation to show that we are also supporting this case.

quentin-jaquier-sonarsource (approved):

PR #3826, opened 2021-10-14
Replace GNU LGPL 3.0 license by a more permissive one for  java-custom-rules-example

alban-auzeill:


henryju:
@gaudol Maybe you may like to have a look before it gets merged?

PR #3836, opened 2021-11-08
SONARJAVA-4052 Provide quick fix availability to SQ

mpaladin:


java-frontend/src/main/java/org/sonar/java/reporting/InternalJavaIssueBuilder.java
```diff
@@ -227,10 +227,12 @@ public void report() {
       }
     }
 
-
     if (!quickFixes.isEmpty() && sonarComponents.isQuickFixCompatible()) {
       addQuickFixes(ruleKey.get(), (NewSonarLintIssue) newIssue);
     }
+    if (!quickFixes.isEmpty() && sonarComponents.hasMethodSetQuickFixAvailable()) {
```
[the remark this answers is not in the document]
mpaladin: @quentin-jaquier-sonarsource ok, thank you for the note.

java-frontend/src/main/java/org/sonar/java/reporting/InternalJavaIssueBuilder.java
```diff
@@ -227,32 +232,40 @@ public void report() {
       }
     }
 
-
-    if (!quickFixes.isEmpty() && sonarComponents.isQuickFixCompatible()) {
-      addQuickFixes(ruleKey.get(), (NewSonarLintIssue) newIssue);
+    final List<JavaQuickFix> flatQuickFixes = quickFixes.stream()
+      .flatMap(s -> s.get().stream())
+      .collect(Collectors.toList());
```
quentin-jaquier-sonarsource, line 237: As discussed, we are always computing the quick fixes here, even when we have nothing to do with them.
I agree that we will eventually always compute them for all products and delaying the computation is just a temporary minor performance gain. Still, we already have the logic in place, we can benefit from it by refactoring the logic here.
Even if it is less readable, I think it is worth it.

java-frontend/src/main/java/org/sonar/java/reporting/InternalJavaIssueBuilder.java
```diff
@@ -227,32 +232,40 @@ public void report() {
...
+    if (!flatQuickFixes.isEmpty()) {
+      if (isQuickFixCompatible) {
+        addQuickFixes(inputFile, ruleKey.get(), flatQuickFixes, (NewSonarLintIssue) newIssue);
+      } else if (methodSetQuickFixAvailable != null) {
+        try {
+          methodSetQuickFixAvailable.invoke(newIssue, true);
+        } catch (ReflectiveOperationException e) {
+          throw new RuntimeException(e);
```
quentin-jaquier-sonarsource, line 245: Throwing this exception here will stop the analysis of the whole file, even if it is unlikely, it seems a bit overkill.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3837, opened 2021-11-09
SONARJAVA-4069 Improve Nullability annotations support in S2638 (ChangeMethodContractCheck)

quentin-jaquier-sonarsource:


java-frontend/src/main/java/org/sonar/java/model/JVariableSymbol.java
```diff
@@ -37,4 +43,78 @@ public VariableTree declaration() {
     return (VariableTree) super.declaration();
   }
 
+
+  static class ParameterPlaceholderSymbol extends Symbols.DefaultSymbol implements Symbol.VariableSymbol {
```
alban-auzeill, line 47: We should fix the 3 locations in org.sonar.java.model.JUtils that always cast a `VariableSymbol` using `(JVariableSymbol)`. It's not anymore the case, it can be for example:
```
  public static boolean isEffectivelyFinal(Symbol.VariableSymbol variableSymbol) {
    return (variableSymbol instanceof JVariableSymbol) && ((IVariableBinding) ((JVariableSymbol) variableSymbol).binding).isEffectivelyFinal();
  }
```

java-frontend/src/main/java/org/sonar/plugins/java/api/semantic/Symbol.java
```diff
@@ -174,6 +174,12 @@
      */
     List<Type> parameterTypes();
 
+    /**
+     * Symbols of parameters declared by this method.
+     * @return empty list if methods has not parameters
+     */
+    List<Symbol> declarationParameters();
```
alban-auzeill, line 181: We should add in the javadoc of this API:

- For `List<Type> parameterTypes();`: in case of generic types of method invokations, this list of types is more specialized than `declarationParameters().stream().map(Symbol::type)`, for exemple it could be `String` instead of `T`.
- For `List<Symbol> declarationParameters();` useful to access parameter annotations but the drawback is: 1) in case of generic types of method invokations, this list of types is less specialized than `parameterTypes()`,  for exemple it could be `T` instead of `String`  2) when the declaration come from binaries, the name of the symbol will be generated `arg...`

alban-auzeill (approved):

PR #3838, opened 2021-11-10
SONARJAVA-4068 S2118-S2441: Fix StackOverflowError raised for self assigned variables

mpaladin:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #3840, opened 2021-11-10
SONARJAVA-4064 S2055: Fix FP when the semantic is incomplete

mpaladin:


java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableSuperConstructorCheck.java
```diff
@@ -60,7 +60,10 @@ public void visitNode(Tree tree) {
   }
 
   private static boolean isNotSerializableMissingNoArgConstructor(@Nullable Type superclass) {
-    return superclass != null && !isSerializable(superclass) && !hasNonPrivateNoArgConstructor(superclass);
+    return superclass != null
```
mpaladin, line 63: this is the uncovered condition that was already uncovered, I believe it is there for defensive programming because it is always going to be not null at this point.

dorian-burihabwa-sonarsource: Indeed, it looks like we always get a `Type` object. Let's keep it for safety then.

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #3847, opened 2021-11-17
SONARJAVA-4079 Improve Nullability annotations support in S4682 (PrimitivesMarkedNullableCheck)

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/PrimitivesMarkedNullableCheck.java
```diff
@@ -46,10 +50,20 @@ public void visitNode(Tree tree) {
     if (returnType.symbolType().isPrimitive()) {
       SymbolMetadata.NullabilityData nullabilityData = methodTree.symbol().metadata().nullabilityData();
       if (nullabilityData.isNullable(METHOD, true, false)) {
-        reportIssue(returnType, String.format("\"@%s\" annotation should not be used on primitive types",
-          nullabilityData.annotation().symbol().name()),
-          Collections.singletonList(new JavaFileScannerContext.Location("Child annotation", nullabilityData.declaration())),
-          null);
+        // Both "annotation" and "declaration" should never be null, as we only target directly annotated methods
+        String annotationName = nullabilityData.annotation().symbol().name();
+        Tree annotationTree = nullabilityData.declaration();
```
quentin-jaquier-sonarsource, line 55: Both `annotationName` and `annotationTree` are nullable here. I decided to not check them as we only target directly annotated methods.
It feels only partially safe to me, what do you think @alban-auzeill ?

quentin-jaquier-sonarsource:
Before merging this PR, we should not forget to document the presence of quick fix in the metadata.

java-checks/src/main/java/org/sonar/java/checks/PrimitivesMarkedNullableCheck.java
alban-auzeill: IMO `isNullable(METHOD` does not guarantee enough the fact that `annotationName` and `annotationTree` is not null.
I would recommend checking them as defensive programming.

alban-auzeill (approved):

alban-auzeill (approved):

PR #3848, opened 2021-11-18
SONARJAVA-4081 Improve Nullability annotations support in S4454 (EqualsParametersMarkedNonNullCheck)

alban-auzeill:


java-checks-test-sources/src/main/java/checks/EqualsParametersMarkedNonNullCheck.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  static class E {
+    public boolean equals(@Nonnull C c) { // Compliant
+      return false;
+    }
+  }
+
+  static class F {
```
quentin-jaquier-sonarsource, line 43: We could add a test to check that equals in a class annotated "@org.eclipse.jdt.annotation.NonNullByDefault" is not reported by the rule as expected.

java-checks/src/main/java/org/sonar/java/checks/EqualsParametersMarkedNonNullCheck.java
```diff
@@ -40,14 +47,27 @@
...
-        reportIssue(variable, "\"equals\" method parameters should not be marked \"@Nonnull\".");
-
-      }
+    if (!"equals".equals(methodTree.symbol().name()) || methodTree.parameters().size() != 1) {
+      return;
+    }
+    VariableTree variable = methodTree.parameters().get(0);
+    if (!variable.symbol().type().is("java.lang.Object")) {
+      return;
+    }
```
quentin-jaquier-sonarsource, line 56: We can use `MethodTreeUtils.isEqualsMethod(methodTree)` here instead.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3850, opened 2021-11-19
Sergiomarotco patch 1

sergiomarotco:
add russian words for finding passwords in russian configs


quentin-jaquier-sonarsource:
Hello @sergiomarotco, thanks for the suggestion.

The goal of the **default** credential words in this rule is not to list every possible word in every possible language, it would be an endless work. It is mainly to list common words and examples. You are of course encouraged to extend the list to suit your needs thanks to the rule's parameter.

Hope it makes sense.
Best,
Quentin

sergiomarotco:
Ok! What about word “token”?

quentin-jaquier-sonarsource:
I guess we can apply the same reasoning here: we do not aim to list every single possible value, just examples, and you are encouraged to extend the list.

In addition, my feeling is that "token" can be used in different contexts, and does not always represent hardcoded credentials. Adding it by default would end up raising false positives.

In any case, if you want to discuss further this topic, feel free to open a new thread on the [community forum](https://community.sonarsource.com/).

PR #3852, opened 2021-11-19
SONARJAVA-4080 Improve Nullability annotations support in S2637 and S4449

quentin-jaquier-sonarsource:
Please rebase the 3 commits separately into master.


java-symbolic-execution/src/main/java/org/sonar/java/se/checks/NonNullSetToNullCheck.java
```diff
@@ -143,15 +146,20 @@ private static boolean callsThisConstructor(MethodTree constructor) {
...
     }
     if (isUndefinedOrNull(context, symbol)) {
-      context.reportIssue(tree.simpleName(), this, MessageFormat.format("\"{0}\" is marked \"{1}\" but is not initialized in this constructor.", symbol.name(), nonNullAnnotation));
+      context.reportIssue(tree.simpleName(), this,
+        MessageFormat.format("\"{0}\" is marked \"{1}\" but is not initialized in this constructor.", symbol.name(), nonnullAnnotationAsString.get()));
     }
   }
 
+  private static boolean isJavaxValidationConstraint(Symbol symbol) {
+    return symbol.metadata().nullabilityData().annotation().symbol().type().fullyQualifiedName().startsWith("javax.validation.constraints.");
```
dorian-burihabwa-sonarsource, line 160: Should we check that `symbol.metadata().nullabilityData().annotation()` is not null here?

java-symbolic-execution/src/test/java/org/sonar/java/se/NullabilityDataUtilsTest.java
```diff
@@ -0,0 +1,61 @@
...
+  void test_unknown_nullability_data() {
+    JavaTree.CompilationUnitTreeImpl cut = (JavaTree.CompilationUnitTreeImpl) JParserTestUtils.parse("" +
+      "class A {" +
+      " @Unknown" +
+      " Object o;" +
+      "}");
+    SymbolMetadata.NullabilityData nullabilityData = getNullabilityDataOfFirstMember(cut);
+    assertThat(nullabilityAsString(nullabilityData)).isEmpty();
+  }
+
```
dorian-burihabwa-sonarsource, line 55: Should we also add a test with a valid annotation for completeness? (knowing well that the compiliation unit does not provide semantic and that this would return `Optional.empty()`)

dorian-burihabwa-sonarsource (changes requested):
Looks good overall, we should check some return values for safety

java-symbolic-execution/src/test/java/org/sonar/java/se/NullabilityDataUtilsTest.java
quentin-jaquier-sonarsource: We don't have the semantic of dependencies in this test. We could add them, but in any way, this method is tested thoroughly in the unit tests of the rules, I don't think it is worth it to do the modification.

dorian-burihabwa-sonarsource (approved):

PR #3857, opened 2021-11-23
Fix quality flaw: make sure to honor nullability contract of parent method

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM

PR #3858, opened 2021-11-23
SONARJAVA-4078 Improve Nullability annotations support in S2789 (NullShouldNotBeUsedWithOptionalCheck)

dorian-burihabwa-sonarsource:


java-checks-test-sources/src/main/java/annotations/nullability/no_default/NullabilityAtVariableLevel.java
```diff
@@ -239,4 +243,45 @@ public Object id1077_type_NO_ANNOTATION_level_PACKAGE(
...
+      try (@Nullable InputStream id3003_type_WEAK_NULLABLE_level_VARIABLE = null) {
+      } catch (@CheckForNull IOException id3004_type_STRONG_NULLABLE_level_VARIABLE) {
+      }
+
+      for (Object id3005_type_NO_ANNOTATION_level_PACKAGE : new ArrayList<Object>()) {
+      }
+      for (@Nonnull Object id3006_type_NON_NULL_level_VARIABLE : new ArrayList<Object>()) {
+      }
+
+      // and the direct annotations are not yet supported on local variables
```
quentin-jaquier-sonarsource, line 279:
```suggestion
```

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -163,10 +164,21 @@ private static boolean isNull(ExpressionTree expression) {
...
-        String.format(messageFormat, "@" + annotation.annotationType().symbolType().name())));
+  private void checkNullability(Symbol symbol, String messageFormat) {
+    SymbolMetadata.NullabilityLevel level;
+    if (symbol.isVariableSymbol()) {
+      level = SymbolMetadata.NullabilityLevel.VARIABLE;
+    } else {
+      level = SymbolMetadata.NullabilityLevel.METHOD;
+    }
+    SymbolMetadata.NullabilityData nullabilityData = symbol.metadata().nullabilityData();
+    if (nullabilityData.isNullable(level, false, false)) {
```
quentin-jaquier-sonarsource, line 175:
```suggestion
    if (nullabilityData.isNullable(level, true, false)) {
```
When an element is annotated nullable via meta-annotation, we should not report an issue, as this annotation may not be removed.

java-checks/src/main/java/org/sonar/java/checks/NullShouldNotBeUsedWithOptionalCheck.java
```diff
@@ -163,10 +164,21 @@ private static boolean isNull(ExpressionTree expression) {
...
+      context.reportIssue(
+        this,
+        nullabilityData.declaration(),
+        String.format(messageFormat, "@" + nullabilityData.annotation().symbol().name())
```
quentin-jaquier-sonarsource, line 179: `nullabilityData.annotation()` is nullabe. Even if I don't think we can come up with an example to would make this code break, I think it makes sense to check it for defensive programming.
Or maybe use the helper method that I added in the [other PR](https://github.com/SonarSource/sonar-java/pull/3852/files#diff-b8e775a1b8262a3f713e6394e3f70eb7d5423ebf44500a19533cee068a1d2e0dR31)?

quentin-jaquier-sonarsource (approved):

PR #3863, opened 2021-11-24
BUILD-1180 Rotate secret ecj_batch

malena-ebert-sonarsource:


quentin-jaquier-sonarsource:
This is an old branch and not maintained anymore, I will delete it.

PR #3875, opened 2021-12-06
SONARJAVA-4092 Control the amount of data loaded in batch mode

dorian-burihabwa-sonarsource:


java-frontend/src/main/java/org/sonar/java/JavaFrontend.java
```diff
@@ -38,19 +38,21 @@
...
 import org.sonar.java.model.VisitorsBridge;
 import org.sonar.plugins.java.api.JavaCheck;
 import org.sonar.plugins.java.api.JavaResourceLocator;
 import org.sonar.plugins.java.api.JavaVersion;
+import org.sonarsource.analyzer.commons.collections.ListUtils;
 import org.sonarsource.performance.measure.PerformanceMeasure;
 import org.sonarsource.performance.measure.PerformanceMeasure.Duration;
 
 public class JavaFrontend {
+  public static final long MAX_BATCH_SIZE = (long) (Runtime.getRuntime().totalMemory() * 0.1);
```
alban-auzeill, line 52: With AutoScan mode, there's no semantic from project `.class` files. We should never limit the size of the batch in this mode. For now, I propose to keep the `batch size`/`memory size` computation for later.

java-frontend/src/main/java/org/sonar/java/JavaFrontend.java
```diff
@@ -133,30 +135,71 @@ public void scan(Iterable<InputFile> sourceFiles, Iterable<InputFile> testFiles,
...
-      } finally {
-        astScanner.endOfAnalysis();
-        astScannerForTests.endOfAnalysis();
-        astScannerForGeneratedFiles.endOfAnalysis();
+      int batchFirstPos = 0;
+      while (batchFirstPos < files.size()) {
+        long batchSize = files.get(batchFirstPos).contents().length();
+        int batchLastPosExclusive = batchFirstPos + 1;
+        while (batchSize < minBatchModeSize && batchLastPosExclusive < files.size()) {
+          batchSize += files.get(batchLastPosExclusive).contents().length();
```
alban-auzeill, line 171: Because we always want all the files in AutoScan mode, we should not load the content of all source files in memory at the same time, in this case, it's useless.
And when we are not in the AutoScan context we should exclude batch mode if `sonarComponents.isSonarLintContext()` so we can compute the size using `new File(inputFile.absolutePath()).length()`.

java-frontend/src/main/java/org/sonar/java/JavaFrontend.java
```diff
@@ -133,30 +135,71 @@ public void scan(Iterable<InputFile> sourceFiles, Iterable<InputFile> testFiles,
...
 
+  private<T extends InputFile> void scanBatch(List<T> allFiles) {
+    try {
+      JParserConfig.Mode.BATCH
+        .create(JParserConfig.effectiveJavaVersion(javaVersion), globalClasspath)
+        .parse(allFiles, this::analysisCancelled, this::scanAsBatchCallback);
+    } finally {
+      astScanner.endOfAnalysis();
+      astScannerForTests.endOfAnalysis();
+      astScannerForGeneratedFiles.endOfAnalysis();
```
alban-auzeill, line 199: In the previous batch mode, there's was only one call to `endOfAnalysis()` on each scanner. Now there's one call per batch, we are changing the behavior. Is it expected?

java-frontend/src/main/java/org/sonar/java/JavaFrontend.java
```diff
@@ -133,30 +135,71 @@ public void scan(Iterable<InputFile> sourceFiles, Iterable<InputFile> testFiles,
...
+        throw new AnalysisException(BATCH_ERROR_MESSAGE, e);
       }
     }
   }
 
+  private<T extends InputFile> void scanBatch(List<T> allFiles) {
+    try {
+      JParserConfig.Mode.BATCH
+        .create(JParserConfig.effectiveJavaVersion(javaVersion), globalClasspath)
+        .parse(allFiles, this::analysisCancelled, this::scanAsBatchCallback);
```
alban-auzeill, line 195: In the `parse` method, we log the following block for each batch, we could have a lot of:
```
[INFO] Using ECJ batch to parse source files.
[INFO] Starting batch processing.
[INFO] 100% analyzed
[INFO] Batch processing: Done!
```

java-frontend/src/main/java/org/sonar/java/JavaFrontend.java
```diff
@@ -133,30 +135,71 @@ public void scan(Iterable<InputFile> sourceFiles, Iterable<InputFile> testFiles,
...
-        throw new AnalysisException("Batch Mode failed, analysis of Java Files stopped.", e);
+        throw new AnalysisException(BATCH_ERROR_MESSAGE, e);
       }
     }
   }
 
+  private<T extends InputFile> void scanBatch(List<T> allFiles) {
+    try {
+      JParserConfig.Mode.BATCH
+        .create(JParserConfig.effectiveJavaVersion(javaVersion), globalClasspath)
```
alban-auzeill, line 194: We should keep `globalClasspath` only for AutoScan mode to help investigating differences

alban-auzeill (changes requested):

java-frontend/src/main/java/org/sonar/java/JavaFrontend.java
```diff
@@ -122,9 +124,9 @@ boolean analysisCancelled() {
     return sonarComponents != null && sonarComponents.analysisCancelled();
   }
 
+
   public void scan(Iterable<InputFile> sourceFiles, Iterable<InputFile> testFiles, Iterable<? extends InputFile> generatedFiles) {
     if (isBatchModeEnabled()) {
```
alban-auzeill, line 129: We should never use batch mode when `sonarComponents.isSonarLintContext()` because it uses `inputFile.absolutePath()` and not `inputFile.contents()`.

java-frontend/src/main/java/org/sonar/java/JavaFrontend.java
alban-auzeill: In fact `endOfAnalysis()` should be called only once per project.

dorian-burihabwa-sonarsource (commented):
LGTM

java-frontend/src/main/java/org/sonar/java/JavaFrontend.java
```diff
@@ -179,25 +179,18 @@ private void scanAsBatch(BatchModeContext context, Iterable<? extends InputFile>
     }
   }
 
-  private void scanInBatches(BatchModeContext context, List<InputFile> allInputFiles) throws IOException {
+  private void scanInBatches(BatchModeContext context, List<InputFile> allInputFiles) {
     long batchModeSizeInKB = sonarComponents.getBatchModeSizeInKB();
```
alban-auzeill, line 183: Rename SONAR_BATCH_MIN_SIZE_KEY into SONAR_BATCH_SIZE_KEY

alban-auzeill (approved):

PR #3884, opened 2021-12-22
SONARJAVA-4010 Fix NPE in JSymbol.hashCode()

peter-mansour:


m-g-sonar:
Hey @peter-mansour !

Thanks for the contribution! Unfortunately, most of the team is currently in Christmas Holidays already, and we won't be able to review your PR before early January! Thanks for the time you took to have a look at this NPE, be sure that we will get back to it very soon!

Cheers,
Michael

m-g-sonar:
Hey @peter-mansour ,

After reviewing the ticket, I picked another approach to avoid NPEs.
Rather than considering that `null` is a possible value as an owner of some types, I prefer to consider that primitive types' owner is the default root package (see #3917). It should not have any impact on any rules but prevent any further NPEs.

While I'm closing your PR as we won't merge it, I still want to thank you for the contribution, I reused parts of your tests. They made sense!

Cheers,
Michael

PR #3888, opened 2022-01-04
Update dependencies of sonar-checks-test-sources

m-g-sonar:
All the changes only impact the `java-checks-test-sources` module, which only contains test code, nothing being shipped with the analyzer.

* Updated dependencies to match versions used by sonar-java parent pom.
* Updated logback version to not use one being vulnerable and get rid of the warning (even if we do strictly nothing with this library)


quentin-jaquier-sonarsource (approved):
LGTM

PR #3891, opened 2022-01-04
SONARJAVA-4101 Update ECJ to 3.28.0

quentin-jaquier-sonarsource:


java-frontend/src/test/java/org/sonar/java/JavaFrontendTest.java
```diff
@@ -302,17 +301,48 @@ void exceptions_outside_rules_as_batch_should_interrupt_analysis_if_fail_fast()
       .hasMessage("Batch Mode failed, analysis of Java Files stopped.");
   }
 
+  @Test
+  void test_preview_feature_in_max_supported_version_do_not_log_message() throws IOException {
+    // When the the actual version match the maximum supported version (currently 17), the preview features flag is
+    // enable in the parser config and we made sure to be able to parse preview features, no need to log anything.
```
m-g-sonar, line 307: typo: enabled*

java-frontend/src/test/java/org/sonar/java/JavaFrontendTest.java
```diff
@@ -302,17 +301,48 @@ void exceptions_outside_rules_as_batch_should_interrupt_analysis_if_fail_fast()
...
+    // we can not guarantee the correct behavior of preview features and log a message.
+    logTester.setLevel(LoggerLevel.DEBUG);
+    scan(new MapSettings().setProperty(JavaVersion.SOURCE_VERSION, "18"),
+      SONARLINT_RUNTIME, "class A { void m(String s) { switch(s) { case null: default: } } }");
+    assertThat(sensorContext.allAnalysisErrors()).isEmpty();
+    assertTrue(logTester.logs(LoggerLevel.WARN).stream().noneMatch(l -> l.endsWith("Unresolved imports/types have been detected during analysis. Enable DEBUG mode to see them.")));
+    assertTrue(logTester.logs(LoggerLevel.WARN).stream().anyMatch(l -> l.endsWith("Use of preview features have been detected during analysis. Enable DEBUG mode to see them.")));
+    // We should keep this message or we won't have anything actionable in the debug logs to understand the warning
+    assertTrue(logTester.logs(LoggerLevel.DEBUG).stream().anyMatch(l -> l.replace("\r\n", "\n").endsWith("Use of preview features:\n" +
+      "- Pattern Matching in Switch is a preview feature and disabled by default. Use --enable-preview to enable")));
```
m-g-sonar, line 328: This is a little bit out of the scope of the ticket, but is it a message we create or directly the message from ECJ?
Because I'm not sure that using `-enable-preview` would change anything here!

In my opinion, we should change the message, as they are probably already using the flag or their project won't compile. It's just that we are not aligned yet.

It probably worthes another ticket.

quentin-jaquier-sonarsource: Indeed, this is a message from ECJ. The problem was already there before, it makes sense to create a ticket: [SONARJAVA-4120](https://jira.sonarsource.com/browse/SONARJAVA-4120).

PR #3892, opened 2022-01-05
SONARJAVA-4118 Introduce Java 17's Sealed Classes as final feature

quentin-jaquier-sonarsource:


quentin-jaquier-sonarsource:
We did an initial investigation to extend the semantic with Sealed Classes information, but it turns out that it was not so easy. The reason is not clear, but it seems that the data that we need is kept by ECJ for internal usage only.

We can probably dig further to find a solution or workaround, but currently none of our rules need this information, it is acceptable to not implement it for the moment.

m-g-sonar (approved):

PR #3893, opened 2022-01-05
SONARJAVA-4120 Logs about preview features should not suggest "-enable-preview"

quentin-jaquier-sonarsource:


java-frontend/src/main/java/org/sonar/java/model/JProblem.java
```diff
@@ -49,6 +49,18 @@ public String message() {
...
   @Override
   public String toString() {
+    if (type == JProblem.Type.PREVIEW_FEATURE_USED) {
+      return cleanMessage(message);
+    }
+    return message;
+  }
+
+  private static String cleanMessage(String message) {
+    String enablePreviewMessage = " Use --enable-preview to enable";
```
dorian-burihabwa-sonarsource, line 59: We should probably set this string as a constant of the class rather than a variable as it is not likely to change

dorian-burihabwa-sonarsource (approved):
LGTM. We could set the message to suppress as a constant rather than a variable, but I don't see this as a blocker.

PR #3894, opened 2022-01-05
SONARJAVA-4100 Abstract classes should be excluded from S5790

dorian-burihabwa-sonarsource:
Inner classes that are abstract but contain test code should not be
reported by S5790.


quentin-jaquier-sonarsource (approved):
LGTM

PR #3898, opened 2022-01-06
SONARJAVA-4102 S6204 should not raise an issue when removeIf is called on the list

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM!

PR #3901, opened 2022-01-06
SONARJAVA-4063 FN in S3688 (disallowed classes) in case of Reflection

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM!

PR #3902, opened 2022-01-06
SONARJAVA-3863 Drop deprecated method "MethodSymbol.overriddenSymbol()"

dorian-burihabwa-sonarsource:


java-frontend/src/main/java/org/sonar/java/model/JMethodSymbol.java
```diff
@@ -119,16 +119,6 @@ public TypeSymbol returnType() {
...
   }
 
-  @Nullable
-  @Override
-  public MethodSymbol overriddenSymbol() {
-    if (overriddenSymbols == null) {
-      // only compute overridenSymbols once
-      overriddenSymbols();
-    }
-    return firstOverridenSymbol;
```
quentin-jaquier-sonarsource, line 129: You can also remove the field `firstOverridenSymbol`. Note that this code does raise an issue in SQ, but not in the PR because the line was unchanged...

quentin-jaquier-sonarsource (changes requested):

sonar-java-plugin/src/main/resources/static/documentation.md
```diff
@@ -97,6 +97,10 @@ The tutorial [Writing Custom Java Rules 101](https://redirect.sonarsource.com/do
 
 ### API changes
 
+#### **7.7**
+
+* **Method `MethodSymbol.overriddenSymbol()` was dropped.**. Despite, the changes described in version 7.0, `overriddenSymbol()` was still shipped as part of the API because it was still used by sonar-security. As this is no longer the case, the method has now been safely removed.
```
quentin-jaquier-sonarsource, line 102: This list of changes are for custom rules developers, I don't think all the details are relevant here. What about this instead?
```suggestion
* **Method `MethodSymbol.overriddenSymbol()`, deprecated in 7.0, was dropped.**. Use `MethodSymbol.overriddenSymbols()` instead.
```

dorian-burihabwa-sonarsource: The thing is: it was deprecated in 6.15. What about?
```suggestion
* **Method `MethodSymbol.overriddenSymbol()` was dropped.**. Deprecated in 6.15 and planned to be dropped in 7.0. `overriddenSymbol()` has now been removed from the API. Use `MethodSymbol.overriddenSymbols()` instead.
```

quentin-jaquier-sonarsource: Sounds good to me.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3904, opened 2022-01-06
SONARJAVA-4060 FP in S3252 when owner type is unknown

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/files/non-compiling/checks/StaticMemberAccess.java
```diff
@@ -19,3 +20,9 @@ public StaticMemberAccessChild() {
     StaticMemberAccessParent.unknown(); // Compliant
   }
 }
+
+class StaticMemberAccessCheckWithUnknown {
+  public unknownGenericType() {
```
dorian-burihabwa-sonarsource, line 25:
```suggestion
  public void unknownGenericType() {
```

java-checks/src/main/java/org/sonar/java/checks/StaticMemberAccessCheck.java
```diff
@@ -56,7 +56,9 @@ public void visitNode(Tree tree) {
     if (symbol.isStatic()  && !isListOrSetOf(mse)) {
       ExpressionTree expression = mse.expression();
       Type staticType = symbol.owner().type();
-      if (!expression.symbolType().erasure().equals(staticType.erasure())) {
+      Type expressionType = expression.symbolType();
+      if (!staticType.isUnknown() && !expressionType.isUnknown()
```
dorian-burihabwa-sonarsource, line 60: In practice, is `expressionType.isUnknown()` ever true?

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/StaticMemberAccessCheck.java
quentin-jaquier-sonarsource: Indeed, I did not manage to come up with a use-case that cover this case, but I decided to keep it for defensive programming.

dorian-burihabwa-sonarsource (approved):

PR #3909, opened 2022-01-11
SONARJAVA-4057 Do not generate FP when rules don't have semantic

quentin-jaquier-sonarsource:
We should probably rebase and merge the commits separately into master.


its/ruling/src/test/resources/jboss-ejb3-tutorial/java-S1258.json
```diff
@@ -1,119 +1,17 @@
 {
-'jboss-ejb3-tutorial:blob/src/org/jboss/tutorial/blob/bean/BlobEntity.java':[
```
quentin-jaquier-sonarsource, line 2: The missing issues are due to the fact that this project is misconfigured on purpose, since the rule now skip classes with unknown annotation, these issues dissapear.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedMethodParameterCheck.java
```diff
@@ -217,4 +223,8 @@ private static boolean isUsedAsMethodReference(MethodTree tree) {
       // no need to check which side of method reference, from an identifierTree, it's the only possibility as direct parent
       .anyMatch(identifier -> identifier.parent().is(Tree.Kind.METHOD_REFERENCE));
   }
+
+  private static boolean hasUnknownAnnotation(SymbolMetadata symbolMetadata) {
```
alban-auzeill, line 227: This method is duplicated 3 times, we should move it to a helper.

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #3917, opened 2022-01-12
SONARJAVA-4010 Make sure primitive type symbols have an owner

m-g-sonar:
Relates to #3884.

I decided to get rid of the `null` approach in order to avoid any future NPE, and assigned ownership of primitive type symbols to the root package.


quentin-jaquier-sonarsource (approved):
LGTM

PR #3919, opened 2022-01-12
SONARJAVA-4070 S1874(CallToDeprecatedMethodCheck) should ignore incomplete method signature

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/CallToDeprecatedMethodCheck.java
```diff
@@ -56,4 +74,26 @@ private boolean nonAbstractOrFlaggedForRemoval(Symbol.MethodSymbol method) {
...
+   * @return true if the arguments' types match types. false otherwise.
+   */
+  private static boolean argumentsMatchSignature(Arguments arguments, List<Type> types) {
+    if (arguments.size() != types.size()) {
+      return false;
+    }
+    for (int i = 0; i < arguments.size(); i++) {
+      ExpressionTree argument = arguments.get(i);
+      Type type = types.get(i);
+      if (!argument.symbolType().equals(type)) {
```
quentin-jaquier-sonarsource, line 93: It seems to me that this will be true only when `argument.symbolType()` is actually unknown. If this is correct, we can reduce the whole logic to something like: `return arguments.stream().noneMatch(a -> a.symbolType().isUnknown());`.
In addition, it will remove the FN with varargs.

What do you think about this solution?

quentin-jaquier-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/CallToDeprecatedMethodCheck.java
dorian-burihabwa-sonarsource: Good point, checking for "unknowness" seems to do the trick. Updating the PR now

quentin-jaquier-sonarsource (approved):
LGTM

PR #3923, opened 2022-01-14
SONARJAVA-4119 Handle new switch patterns in CFG

m-g-sonar:


java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -815,7 +837,9 @@ private static boolean switchWithoutFallThrough(SwitchTree switchTree) {
   }
 
   private static boolean containsDefaultCase(List<CaseLabelTree> labels) {
-    return labels.stream().anyMatch(caseLabel -> "default".equals(caseLabel.caseOrDefaultKeyword().text()));
+    return labels.stream().anyMatch(caseLabel -> "default".equals(caseLabel.caseOrDefaultKeyword().text())
+      // FIXME JDK 17 preview feature
```
quentin-jaquier-sonarsource, line 841: We can probably remove this `FIXME`.

java-symbolic-execution/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -734,6 +737,20 @@ private void visit(Tree tree, @Nullable Tree terminator) {
...
         programState = programState.stackValue(constraintManager.createSymbolicValue(tree));
         break;
+      case DEFAULT_PATTERN:
+        programState = programState.stackValue(constraintManager.createSymbolicValue(tree));
+        break;
+      case TYPE_PATTERN:
+        programState = programState.stackValue(constraintManager.createSymbolicValue(tree));
+        break;
+      case NULL_PATTERN:
+        // FIXME might cause NPE in the block
```
quentin-jaquier-sonarsource, line 747: From this message, it is not clear if the analyzer is subject to a NPE or if the code we are analyzing could be subject to a NPE.
I guess it's the latter, so we could go for something like:
```suggestion
        // FIXME we could add a null constraint to "o"
```

java-symbolic-execution/src/test/java/org/sonar/java/se/ExplodedGraphWalkerTest.java
```diff
@@ -160,6 +160,19 @@ void switchExpression() {
       .verifyIssues();
   }
 
+  /**
+   * Just checking Java 17 patterns do not fail SE engine
```
quentin-jaquier-sonarsource, line 164:
```suggestion
   * Checking that Java 17 patterns do not fail SE engine
```

quentin-jaquier-sonarsource (approved):

PR #3924, opened 2022-01-17
Bump JDK version from 16 to 17 to build java-checks-test-sources using the latest

alban-auzeill:


java-checks-test-sources/pom.xml
```diff
@@ -841,12 +841,9 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <release>16</release>
-          <compilerArgs>
```
m-g-sonar, line 845: I think we still need this if we want to test preview-features of jdk 17 (pattern in switches)

m-g-sonar:
LGTM!

PR #3926, opened 2022-01-18
SONARJAVA-4059 Rule S6373 XML parsers should not allow inclusion of arbitrary files

alban-auzeill:


java-symbolic-execution/src/main/java/org/sonar/java/se/checks/XxeProcessingCheck.java
```diff
@@ -157,7 +151,6 @@
           || c.hasConstraint(FeatureIsSupportingExternalEntities.SECURED)
           || c.hasConstraint(XxeEntityResolver.CUSTOM_ENTITY_RESOLVER))
       .put(DOCUMENT_BUILDER_FACTORY_NEW_INSTANCE, XxeProcessingCheck::documentBuilderFactoryIsSecured)
-      .put(NEW_DOCUMENT_BUILDER, c -> c.hasConstraint(XxeEntityResolver.CUSTOM_ENTITY_RESOLVER))
```
quentin-jaquier-sonarsource, line 150: This change is a refactor of the fix from [SONARJAVA-3684](https://jira.sonarsource.com/browse/SONARJAVA-3684). The fix was working fine but was not suited to support new rules.

It will indeed introduce new FN, but if we look closely, we are subject to such FN for any "TRANSFERRING_METHOD_CALLS". I think this use case will hardly ever happen, it is fine to not support it for now. In any case, I created a new ticket to document the problem: [SONARJAVA-4130](https://jira.sonarsource.com/browse/SONARJAVA-4130)

quentin-jaquier-sonarsource: For the record: the fix was not working so well: it was causing FP, see [SONARJAVA-3842](https://jira.sonarsource.com/browse/SONARJAVA-3842) for example. This refactoring will also solve the problem mentioned in the ticket.

PR #3928, opened 2022-01-18
DBD-98 Add Dataflow Bug Detection rules to the default quality profile

guillaume-dequenne-sonarsource:


sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaSonarWayProfile.java
```diff
@@ -93,40 +98,48 @@ private static String readResource(URL resource) {
...
+      LOG.debug(String.format("%s is not found, no %s rules added to Sonar way java profile: %s", className, rulesCategory, e.getMessage()));
     } catch (NoSuchMethodException e) {
-      LOG.debug("Method is not found, no security rules added to Sonar way java profile: " + e.getMessage());
+      LOG.debug(String.format("Method is not found, no %s rules added to Sonar way java profile: %s", rulesCategory, e.getMessage()));
     } catch (IllegalAccessException e) {
-      LOG.debug("[IllegalAccessException] no security rules added to Sonar way java profile: " + e.getMessage());
+      LOG.debug(String.format("[IllegalAccessException] no %s rules added to Sonar way java profile: %s", rulesCategory, e.getMessage()));
     } catch (InvocationTargetException e) {
-      LOG.debug("[InvocationTargetException] no security rules added to Sonar way java profile: " + e.getMessage());
+      LOG.debug(String.format("[InvocationTargetException] no %s rules added to Sonar way java profile: %s", rulesCategory, e.getMessage()));
```
sebastian-hungerecker-sonarsource, line 134: I think it would be a lot nicer to combine all of these into a single multi-catch block (`catch (ClassNotFoundException|NoSuchMethodExcpetion|IllegalAccessException|InvocationTargetException e)` and unify the message to `"[%s] No %s rules added to Sonar way java profile: %s"`. This slightly changes the message for the first two cases, but that should not matter - we don't lose any information because the class name is still present in the exception message. And it just feels more consistent to have the same message in all cases. And most importantly we won't have four catch blocks that all log basically the same message.

As an added bonus, this should get the coverage to 100%.

sebastian-hungerecker-sonarsource (changes requested):
I'd like to merge the four catch blocks. Looks good otherwise.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaSonarWayProfile.java
guillaume-dequenne-sonarsource: Makes a lot of sense, thanks for the suggestion! At first I didn't want to change the existing behavior for security rules but indeed I think this makes the code much clearer.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #3936, opened 2022-01-25
SONARJAVA-4067 RSPEC-6377 XML signatures should be validated securely

m-g-sonar:


java-symbolic-execution/src/main/java/org/sonar/java/se/checks/XmlValidatedSighatureCheck.java
```diff
@@ -0,0 +1,219 @@
...
+      if (DOM_VALIDATE_CONTEXT_CONSTRUCTOR.matches(newClass)) {
+        context.getConstraintManager().setValueFactory(() -> new DomValidateContextSymbolicValue(newClass.identifier()));
+      }
+    }
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree mit) {
+      if (SET_PROPERTY.matches(mit)) {
+        Arguments args = mit.arguments();
+        if (!SECURE_VALIDATION_PROPERTY.equalsIgnoreCase(args.get(0).asConstant(String.class).orElse(""))) {
```
quentin-jaquier-sonarsource, line 133: When we can not resolve the value as a constant, we will then consider that it is not the expected property. Even if this may result in FP for strange code, I think it is fine anyway. Still, it may be worth a comment.

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/XmlValidatedSighatureCheck.java
```diff
@@ -0,0 +1,219 @@
...
+        SymbolicValue domSv = programState.peekValue(args.size());
+        BooleanConstraint booleanConstraint = programState.getConstraint(programState.peekValue(0), BooleanConstraint.class);
+        if (booleanConstraint == BooleanConstraint.FALSE) {
+          programState = programState.addConstraint(domSv, DomSecureValidation.DISABLED);
+        } else if (booleanConstraint == BooleanConstraint.TRUE) {
+          programState = programState.addConstraint(domSv, DomSecureValidation.ENABLED);
+        } else {
+          // no idea what it could be, better say nothing
+          programState = programState.addConstraint(domSv, DomSecureValidation.ENABLED);
+        }
```
quentin-jaquier-sonarsource, line 145: What about removing the redundant `else if`?

java-checks-test-sources/src/main/java/symbolicexecution/checks/S6377_XmlValidatedSighatureCheck.java
```diff
@@ -0,0 +1,85 @@
...
+
+    XMLSignature signature = fac.unmarshalXMLSignature(valContext);
+
+    signature.validate(valContext);
+  }
+
+  void noncompliant_3(Node node, KeySelector selector) throws Exception {
+    DOMValidateContext valContext = new DOMValidateContext(selector, node);
+    valContext.setProperty("org.jcp.xml.dsig.secureValidation", Boolean.TRUE);
+    valContext.setProperty("org.jcp.xml.dsig.secureValidation", Boolean.FALSE); // Noncompliant {{Use the secure validation mode when validating this XML signature.}}
```
quentin-jaquier-sonarsource, line 71: We will report on the `setProperty` only if the constraint changes after the constructor, and not when the constraint was already "DISABLED":

```
DOMValidateContext valContext = new DOMValidateContext(selector, node);  // Noncompliant {{Use the secure validation mode when validating this XML signature.}}
valContext.setProperty("org.jcp.xml.dsig.secureValidation", Boolean.FALSE); // Issue could be here?
```

We should either always report on `setProperty` (when present), or always on the constructor (with secondary location on the problematic `setProperty`). We use the second solution for S2755.

m-g-sonar: I think it's better to always highlight the setter when it is explicit to FALSE, and the constructor if there is no such call. I don't really like the secondary approach since the real place when we should fix things is in the setter.

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/XmlValidatedSighatureCheck.java
```diff
@@ -0,0 +1,221 @@
...
+      if (DOM_VALIDATE_CONTEXT_CONSTRUCTOR.matches(newClass)) {
+        context.getConstraintManager().setValueFactory(() -> new DomValidateContextSymbolicValue(newClass.identifier()));
+      }
+    }
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree mit) {
+      if (SET_PROPERTY.matches(mit)) {
+        Arguments args = mit.arguments();
+        if (!SECURE_VALIDATION_PROPERTY.equalsIgnoreCase(args.get(0).asConstant(String.class).orElse(""))) {
```
quentin-jaquier-sonarsource, line 134: What do you think about this code:
```suggestion
        if (args.get(0).asConstant(String.class).filter(SECURE_VALIDATION_PROPERTY::equalsIgnoreCase).isEmpty()) {
```
to avoid the negation and the possible comparison with an empty String?

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/XmlValidatedSighatureCheck.java
```diff
@@ -0,0 +1,221 @@
...
+          return;
+        }
+        SymbolicValue domSv = programState.peekValue(args.size());
+        if (programState.getConstraint(programState.peekValue(0), BooleanConstraint.class) == BooleanConstraint.FALSE) {
+          programState = programState.addConstraint(domSv, DomSecureValidation.EXPLICITLY_DISABLED);
+        } else {
+          // directly set to true, or no idea what the value could be, better say nothing
+          programState = programState.removeConstraintsOnDomain(domSv, DomSecureValidation.class);
+        }
+      } else if (XML_SIGNATURE_VALIDATE.matches(mit)) {
```
quentin-jaquier-sonarsource, line 145: I checked a bit the usages of `XMLValidateContext`, and it seems that there are other ways to use the context. For example:

- [XMLSignature](https://docs.oracle.com/javase/7/docs/api/javax/xml/crypto/dsig/XMLSignature.html#validate(javax.xml.crypto.dsig.XMLValidateContext))

```
XMLSignature signature;
signature.getSignatureValue().validate(context);
```

- [Reference](https://docs.oracle.com/javase/7/docs/api/javax/xml/crypto/dsig/Reference.html#validate(javax.xml.crypto.dsig.XMLValidateContext))
```
org.apache.jcp.xml.dsig.internal.dom.DOMReference dom;
dom.validate(context);
```

Maybe we should double-check with Appsec if we should report this kind of calls.

java-checks/src/main/java/org/sonar/java/checks/CheckList.java
```diff
@@ -780,12 +780,12 @@ private CheckList() {
       OptionalGetBeforeIsPresentCheck.class,
       StreamConsumedCheck.class,
       RedundantAssignmentsCheck.class,
+      XmlValidatedSighatureCheck.class,
```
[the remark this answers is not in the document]
m-g-sonar: indeed -_-

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/XmlValidatedSighatureCheck.java
m-g-sonar: nice!

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/XmlValidatedSignatureCheck.java
```diff
@@ -0,0 +1,222 @@
...
+    if (!(sv instanceof DomValidateContextSymbolicValue)) {
+      return;
+    }
+    DomValidateContextSymbolicValue domSv = (DomValidateContextSymbolicValue) sv;
+    if (domSv.isField) {
+      return;
+    }
+    Optional.ofNullable(ps.getConstraint(domSv, DomSecureValidation.class))
+      .map(constraint -> getIssueLocation(ps, context, domSv, constraint))
+      .ifPresent(tree -> context.reportIssue(tree, this, "Use the secure validation mode when validating this XML signature."));
```
quentin-jaquier-sonarsource, line 173: 👍

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/XmlValidatedSignatureCheck.java
```diff
@@ -0,0 +1,222 @@
...
+
+  private static final MethodMatchers SET_PROPERTY = MethodMatchers.create()
+    .ofSubTypes(JAVAX_XML_CRYPTO_VALIDATE_CONTEXT)
+    .names("setProperty")
+    .addParametersMatcher("java.lang.String", "java.lang.Object")
+    .build();
+
+  private static final MethodMatchers XML_SIGNATURE_VALIDATE = MethodMatchers.create()
+    .ofAnyType()
+    .names("validate")
```
quentin-jaquier-sonarsource, line 64: Nice one, it does the trick nicely!

quentin-jaquier-sonarsource (approved):
LGTM, nice work!

quentin-jaquier-sonarsource:
@m-g-sonar I let you merge the PR as soon as you fix the last code smell!

PR #3937, opened 2022-01-26
SONARJAVA-4008 S2755 should accept setExpandEntityReferences solution for openJDK >= 13

quentin-jaquier-sonarsource:


m-g-sonar:
Pretty nice work! LGTM!

PR #3938, opened 2022-01-26
SONARJAVA-3899 SONARJAVA-3842 Test correct behavior of "newDocumentBuilder" usage

quentin-jaquier-sonarsource:
The problems of both SONARJAVA-3899 and SONARJAVA-3842 were solved in https://github.com/SonarSource/sonar-java/pull/3926, more specifically in https://github.com/SonarSource/sonar-java/pull/3926/commits/496c487c251791a3935595c500f6c2ceb6429b7f.

This PR aims to test the correct behavior of this change with regard to the two tickets.

### DIFF
--- a/java-checks-test-sources/src/main/java/symbolicexecution/checks/S2755_XxeProcessingCheck_DocumentBuilderFactory.java
+++ b/java-checks-test-sources/src/main/java/symbolicexecution/checks/S2755_XxeProcessingCheck_DocumentBuilderFactory.java
@@ -177,3 +177,25 @@ class DocumentBuilderFactoryTest {
     }
   }
 }
+
+class DocumentBuilderFactoryTest_InStaticBlock {
+  private static final DocumentBuilderFactory documentBuilderFactory = DocumentBuilderFactory.newInstance();
+
+  static {
+    documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+    documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+  }
+
+  public static DocumentBuilder noIssue() throws Exception {
+    // No issue here, the code is equivalent to the one in the other method.
+    // We only report an issue on the declaration of "DocumentBuilderFactory", exactly to avoid such cases where the DocumentBuilder is created somewhere else.
+    return documentBuilderFactory.newDocumentBuilder();
+  }
+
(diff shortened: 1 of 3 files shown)


java-checks-test-sources/src/main/java/symbolicexecution/checks/S2755_XxeProcessingCheck_DocumentBuilderFactory.java
```diff
@@ -165,3 +165,25 @@ DocumentBuilderFactory two_factory4(boolean b) throws ParserConfigurationExcepti
...
+class DocumentBuilderFactoryTest_InStaticBlock {
+  private static final DocumentBuilderFactory documentBuilderFactory = DocumentBuilderFactory.newInstance();
+
+  static {
+    documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+    documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+  }
+
+  public static DocumentBuilder noIssue() throws Exception {
+    // No issue here, the code is equivalent to the one in the other method.
```
dorian-burihabwa-sonarsource, line 178: The "other method" is [equivalent](https://github.com/SonarSource/sonar-java/pull/3938/files#diff-61a413c2c2ca3340a7e6d425863e05ca0d559c750d012961a5520961522226f0R183)? Should we name it explicitly?
```suggestion
    // No issue here, the code is equivalent to the one in the other method {@link symbolicexecution.checks.DocumentBuilderFactoryTest_InStaticBlock#equivalent}.
```

dorian-burihabwa-sonarsource (approved):
LGTM

java-checks-test-sources/src/main/java/symbolicexecution/checks/S2755_XxeProcessingCheck_DocumentBuilderFactory.java
quentin-jaquier-sonarsource: I like the idea, but I can not click the link (at least on my IDE). What do you think about simply:

```suggestion
    // No issue here, the code is equivalent to the one in the next method.
```

dorian-burihabwa-sonarsource: Sound good to me 👍🏿

PR #3939, opened 2022-01-26
SONARJAVA-3721 FP in S6212 when a method has parameterized return types

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM!

PR #3941, opened 2022-01-27
Remove GSON from plugin modules POMs because it is only used in ITs

m-g-sonar:


alban-auzeill (approved):

m-g-sonar:
GSON is still used transitively in many places.

PR #3955, opened 2022-02-02
SONARJAVA-3887 Rule S5808 should not raise when an exception is thrown

quentin-jaquier-sonarsource:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S5808_java.html
```diff
@@ -87,6 +87,8 @@ <h2>Compliant Solution</h2>
     }
 }
 </pre>
+<h2>Exceptions</h2>
+<p>No issue is reported when the method is throwing an exception as it might be used to indicate a strong decision.</p>
```
dorian-burihabwa-sonarsource, line 91: To update with the RSPEC

dorian-burihabwa-sonarsource (approved):
LGTM!

PR #3956, opened 2022-02-02
SONARJAVA-3893 Document S128 fall through exception

dorian-burihabwa-sonarsource:


quentin-jaquier-sonarsource (approved):
LGTM.

As a general observation, I think that for changes that imply only description changes, it's okay to only create the PR on rpsec side and let the "update metadata" do the work, to avoid having to update two PRs in case changes are required.

PR #3957, opened 2022-02-03
SONARJAVA-4148 Duplicated "Using ECJ batch to parse source files" logs

alban-auzeill:


java-frontend/src/main/java/org/sonar/java/AnalysisProgress.java
```diff
@@ -0,0 +1,53 @@
...
+    return totalFileCount;
+  }
+
+  public int analysedFileCount() {
+    return analysedFileCount;
+  }
+
+  public int currentBatchSize() {
+    return currentBatchSize;
+  }
```
quentin-jaquier-sonarsource, line 43: Suggestion: 

Instead of having these getters and using them extensively in the other part of the code, what do you think about putting the logic inside the class, and having methods like `isLastBatch`, `isFirstBatch`, `getGlobalPercentage` and `batchPercentage` in this class?

quentin-jaquier-sonarsource (approved):
LGTM

PR #3958, opened 2022-02-04
SONARJAVA-4144 S2699 and S6103 should not report an issue in case of incomplete semantic

quentin-jaquier-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM!

PR #3965, opened 2022-02-11
Add test to compare projects with and without binaries

m-g-sonar:


its/ruling/src/test/java/org/sonar/java/it/SonarWayTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @ClassRule
+  public static Orchestrator orchestrator = Orchestrator.builderEnv()
+    .setSonarVersion(System.getProperty("sonar.runtimeVersion", "LATEST_RELEASE[8.9]"))
+    .addPlugin(FileLocation.byWildcardMavenFilename(new File("../../sonar-java-plugin/target"), "sonar-java-plugin-*.jar"))
+    .addPlugin(MavenLocation.of("org.sonarsource.sonar-lits-plugin", "sonar-lits-plugin", "0.10.0.2181"))
+    .build();
+
+  @Test
+  public void javaCheckTestSources() throws Exception {
```
m-g-sonar, line 52: The test requires java 17 to run, while the ruling itself requires Java 11. I wonder how to do that in practice.
Should we move the whole ruling to Java 17?

its/ruling/src/test/java/org/sonar/java/it/SonarWayTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    // 5min sleep
+    LOG.info("WAITING 5MIN BEFORE COMPARING RESULTS (SERVER: " + orchestrator.getServer().getUrl() + ")");
+    Thread.sleep(1000 * 60 * 5);
+
+    /**
+     * 3. Check differences
+     */
+    String differences = new String(Files.readAllBytes(FileLocation.of("target/" + projectKey + "-no-binaries_differences").getFile().toPath()), StandardCharsets.UTF_8);
+    Assertions.assertThat(differences).isEmpty();
```
m-g-sonar, line 114: This is definitely failing for the time being. What I would like to check is that there is no "new issues", but only "missing" ones.

alban-auzeill (approved):

PR #3977, opened 2022-02-17
Move test sources to `java-checks-test-sources` (part 5)

dorian-burihabwa-sonarsource:


java-checks/src/test/java/org/sonar/java/checks/spring/SpringScanDefaultPackageCheckTest.java
```diff
@@ -22,20 +22,22 @@
...
 
+import static org.sonar.java.checks.verifier.TestUtils.nonCompilingTestSourcesPath;
+
 class SpringScanDefaultPackageCheckTest {
 
   @Test
   void test() {
     CheckVerifier.newVerifier()
-      .onFile("src/test/files/checks/spring/SpringScanDefaultPackageCheck.java")
+      .onFile(nonCompilingTestSourcesPath("checks/spring/SpringScanDefaultPackageCheck.java"))
```
m-g-sonar, line 32: what about moving these file directly in the root package then?! We can do that

m-g-sonar (commented):
Did you try to put the files of S4602 directly at the root of the sources? I feel it could work,

m-g-sonar:
LGTM!

PR #3978, opened 2022-02-18
SONARJAVA-4177 Provide OWASP Top 10 2021 security standards for rules metadata

vilchik-elena:


vilchik-elena:
As soon as I bumped sonar-plugin-api to 9.3 some tests related to 'setQuickFixAvailable' started to fail, so I dropped the reflection as there is no need in it anymore.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
```diff
@@ -151,10 +160,16 @@ private static void addMetadata(NewRule rule, @Nullable RuleMetadata metadata) {
     addSecurityStandards(rule, metadata.securityStandards);
   }
 
-  private static void addSecurityStandards(NewRule rule, SecurityStandards securityStandards) {
-    for (String s : securityStandards.OWASP) {
+  private void addSecurityStandards(NewRule rule, SecurityStandards securityStandards) {
+    for (String s : securityStandards.OWASP_2017) {
       rule.addOwaspTop10(OwaspTop10.valueOf(s));
```
vilchik-elena, line 165: I am surprised Next didn't report usage of this deprecated API.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
```diff
@@ -151,10 +160,16 @@ private static void addMetadata(NewRule rule, @Nullable RuleMetadata metadata) {
...
     }
+    boolean isOwaspByVersionSupported = sonarRuntime.getApiVersion().isGreaterThanOrEqual(Version.create(9, 3));
+    if (isOwaspByVersionSupported) {
+      for (String s : securityStandards.OWASP_2021) {
+        rule.addOwaspTop10(OwaspTop10Version.Y2021, OwaspTop10.valueOf(s));
```
m-g-sonar, line 170: I don't even understand how this can build. The class is not imported anywhere.

m-g-sonar: Okay, so this class is directly defined in the parent interface `RulesDefinition`

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
m-g-sonar: since you didn't modify the line, it might have excluded the issue from the report.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
```diff
@@ -151,10 +160,16 @@ private static void addMetadata(NewRule rule, @Nullable RuleMetadata metadata) {
     addSecurityStandards(rule, metadata.securityStandards);
   }
 
-  private static void addSecurityStandards(NewRule rule, SecurityStandards securityStandards) {
-    for (String s : securityStandards.OWASP) {
+  private void addSecurityStandards(NewRule rule, SecurityStandards securityStandards) {
+    for (String s : securityStandards.OWASP_2017) {
       rule.addOwaspTop10(OwaspTop10.valueOf(s));
     }
+    boolean isOwaspByVersionSupported = sonarRuntime.getApiVersion().isGreaterThanOrEqual(Version.create(9, 3));
```
m-g-sonar, line 167: I feel that this value should be computed only once, at the instantiation of the RulesDefinition class. Currently, it will be called for each rule,

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaRulesDefinition.java
m-g-sonar, line 170: I think I would prefer to have the enums `OwaspTop10Version` and `OwaspTop10` prefixed with `RulesDefinition` to be clear about where they come from:

```
RulesDefinition.OwaspTop10Version
```

ghost (commented):
I would appreciate the little changes.
Also, could you rebase your branch on top of master?

ghost (approved):

m-g-sonar:
LGTM! Thank you @vilchik-elena !

PR #3979, opened 2022-02-21
Ruling jdk 17

m-g-sonar:


.cirrus.yml
```diff
@@ -170,7 +171,15 @@ ruling_win_task:
...
-  <<: *WINDOWS_VM_DEFINITION
+  gce_instance:
+    image_project: language-team
+    image_family: lt-base-windows-jdk17
+    platform: windows
+    zone: us-central1-a
+    preemptible: false
+    disk: 128
+    use_ssd: true
+    type: n2d-standard-4
```
alban-auzeill, line 182: You removed the last reference to `&WINDOWS_VM_DEFINITION` without removing the definition of it. And now we have two duplicated `gce_instance` blocks. Instead, I suggest keeping `&WINDOWS_VM_DEFINITION` definition, changing `image_family: lt-base-windows-jdk17` and `type: n2d-standard-4`, and using the reference 2 times, for `qa_os_win_task` and `ruling_win_task`.

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #3984, opened 2022-02-28
SONARJAVA-4181 Introduce rule selection for AutoScan

quentin-jaquier-sonarsource:


alban-auzeill (approved):

java-checks/src/main/java/org/sonar/java/checks/CheckList.java
```diff
@@ -217,634 +218,716 @@
...
+    NullDereferenceCheck.class,
+    ObjectOutputStreamCheck.class,
+    OptionalGetBeforeIsPresentCheck.class,
+    ParameterNullnessCheck.class,
+    RedundantAssignmentsCheck.class,
+    StreamConsumedCheck.class,
+    StreamNotConsumedCheck.class,
+    UnclosedResourcesCheck.class,
+    XmlParserLoadsExternalSchemasCheck.class,
+    XxeProcessingCheck.class,
```
quentin-jaquier-sonarsource, line 866: We should not hardcode the list of rules relying on SE, but dynamically exclude them, to be sure that we are not missing any.

alban-auzeill (approved):

PR #3985, opened 2022-03-01
SONARJAVA-4184 do not raise issue if unknown methods and handle constructors

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/RawExceptionCheck.java
```diff
@@ -71,37 +72,51 @@ public void visitMethod(MethodTree tree) {
     exceptionsThrownByMethodInvocations.clear();
   }
 
+  private static boolean hasNoUnknownMethod(MethodTree method) {
+    MethodTreeUtils.MethodInvocationCollector unknownMethodVisitor = new MethodTreeUtils.MethodInvocationCollector(Symbol::isUnknown);
```
quentin-jaquier-sonarsource, line 76: At this point in the logic, we already visited all methods invocations inside the body of the method to collect the exceptions thrown. We can probably determine if an invocation is unknown at the same time to avoid this second visit.

m-g-sonar: @quentin-jaquier-sonarsource  I thought about it, however, I can not do that easily, since the check is a `BaseTreeVisitor`, it will also enter anonymous classes and lambdas. I didn't want to radically change the logic here, as I would have to introduce a notion of state.

Do you believe I should rewrite completely the rule to be a subscription visitor (on methods and constructors?)

quentin-jaquier-sonarsource: Indeed, rewriting the whole rule would solve both problems, but seems out of scope for now. Since this code does fix correctly the problem, we can go ahead and merge this PR.

quentin-jaquier-sonarsource (approved):
LGTM

PR #3997, opened 2022-03-08
SONARJAVA-4194 Avoid recursing on BinaryExpressionTree and limit targeted trees

m-g-sonar:


java-checks/src/main/java/org/sonar/java/checks/CollectionIsEmptyCheck.java
```diff
@@ -54,38 +57,62 @@
...
-      }
+  public void visitNode(Tree tree) {
+    if (tree.is(CLASS_TREES)) {
+      boolean isCollection = ((ClassTree) tree).symbol().type().isSubtypeOf(JAVA_UTIL_COLLECTION);
+      IS_COLLECTION_STACK.push(isCollection);
+      return;
+    }
+    // Necessarily a BinaryExpressionTree - the rule only raise issues when size() comparison
+    // are "poorly" used from something which is NOT a Collection
+    if (Boolean.FALSE.equals(IS_COLLECTION_STACK.peek())) {
```
dorian-burihabwa-sonarsource, line 97: Should we rewrite the test as a separate method and exit early?
```
if (isNodeInACollectionType()) {
  return;
}
```

dorian-burihabwa-sonarsource: I think the test can be misled because we may be testing size() against 0 in an inner class.
```
public class MyCollectionType<T> implements Collection<T> {

  @Override
  public boolean isEmpty() {
    return false;
  }

  @NotNull
  @Override
  public Iterator<T> iterator() {
    return new MyIterator<>();
  }
  
  class MyIterator<T> implements Iterator<T> {

    @Override
    public boolean hasNext() {
      return MyCollectionType.this.size() == 0;
    }
  }
}
```
Should we consider this case?
```suggestion
    if (IS_COLLECTION_STACK.stream().allMatch(Boolean.FALSE::equals)) {
```

java-checks/src/main/java/org/sonar/java/checks/CollectionIsEmptyCheck.java
```diff
@@ -54,38 +57,62 @@
...
+    Tree.Kind.GREATER_THAN_OR_EQUAL_TO
+  };
+  private static final Tree.Kind[] CLASS_TREES = {
+    Tree.Kind.CLASS,
+    Tree.Kind.ENUM,
+    Tree.Kind.INTERFACE,
+    Tree.Kind.RECORD,
+    Tree.Kind.ANNOTATION_TYPE
+  };
+  private static final Deque<Boolean> IS_COLLECTION_STACK = new LinkedList<>();
```
dorian-burihabwa-sonarsource, line 75: We should rename this static variable to `enclosingTypes`

dorian-burihabwa-sonarsource (changes requested):
This looks good overall. Some minor changes could help. We could also look into covering the use of `size()` in Maps (perhaps in a new onboarding ticket).

java-checks/src/main/java/org/sonar/java/checks/CollectionIsEmptyCheck.java
m-g-sonar: damned. This question of inner class vs nested class is making the rule much more complex than expected.

java-checks/src/main/java/org/sonar/java/checks/CollectionIsEmptyCheck.java
```diff
@@ -54,38 +58,81 @@
...
-  public void visitClass(ClassTree tree) {
-    for (Tree member : tree.members()) {
-      if (!tree.symbol().type().isSubtypeOf(JAVA_UTIL_COLLECTION) || !member.is(Tree.Kind.METHOD)) {
-        scan(member);
-      }
+  public void visitNode(Tree tree) {
+    if (tree.is(CLASS_TREES)) {
+      handleClassTree((ClassTree) tree);
+    } else {
+      // Necessarily a BinaryExpressionTree - the rule only raise issues when size() comparison
```
dorian-burihabwa-sonarsource, line 94:
```suggestion
      // Necessarily a BinaryExpressionTree - the rule only raises issues on size() comparisons
```

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4003, opened 2022-03-18
SONARJAVA-4094 Fix support of records and permitted types in S1105 and S1106

m-g-sonar:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4010, opened 2022-03-23
SONARJAVA-4200 Document sonar.java.skipUnchanged property

dorian-burihabwa-sonarsource:


sonar-java-plugin/src/main/resources/static/documentation.md
```diff
@@ -63,6 +63,19 @@ Example: `sonar.java.source=1.6`
 
 If the property is provided, the analysis will take the source version into account, and execute related rules accordingly. At run time, each of these rules will be executed – or not – depending of the Java version used by sources within the project. For instance, on a correctly configured project built with Java 6, rules targeting Java 7 and Java 8 will never raise issues, even though they are enabled in the associated rule profile.
 
+## Skipping unchanged files
+Starting with SonarQube 9.4, and by default, the Java analyzer optimizes the analysis of unchanged files in pull requests.
```
agigleux, line 67: Is that documentation visible in SonarCloud? If yes, then we should find a way to remove "SonarQube 9.4".
Or say: "Starting from April 2022"

sonar-java-plugin/src/main/resources/static/documentation.md
```diff
@@ -63,6 +63,19 @@ Example: `sonar.java.source=1.6`
...
+In practice, this means that on a file that has not changed, the analyzer only runs a restricted list of checks.
```
agigleux, line 68: I would replace "checks" by "rules" because this is the wording we use in our products.

sonar-java-plugin/src/main/resources/static/documentation.md
```diff
@@ -63,6 +63,19 @@ Example: `sonar.java.source=1.6`
...
+To get a better understanding of the rule exclusion mechanism, keep in mind that:
+* Rules that need to run on multiple files to decide whether they need to raise issues are always executed
+* Rules that need to run at the end of the analysis to decide whether they need to raise issues are always executed
+* Rules that are defined out of the `sonar.java.checks` package are always executed
+
+This last criteria implies that checks defined in custom plugins cannot be skipped.
```
agigleux, line 74: I would replace that by:

"This last criteria implied that [custom rules] cannot be skipped". Where [custom rules] is a link to the doc about how to create custom rules.

agigleux (changes requested):

sonar-java-plugin/src/main/resources/static/documentation.md
dorian-burihabwa-sonarsource: It looks like we are sourcing the [sonarcloud documentation](https://docs.sonarcloud.io/advanced-setup/languages/java/)  from the same file. I will change the text to refer to April 2022 rather than a specific version.

agigleux (approved):

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -86,7 +86,11 @@
   public static final String SONAR_AUTOSCAN_CHECK_FILTERING = "sonar.internal.analysis.autoscan.filtering";
   public static final String SONAR_BATCH_SIZE_KEY = "sonar.java.experimental.batchModeSizeInKB";
   public static final String SONAR_FILE_BY_FILE = "sonar.java.fileByFile";
-  public static final String SONAR_CAN_SKIP_UNCHANGED_FILES_KEY = "sonar.java.internal.skipUnchanged";
+  /**
+   * Enables optimized analysis of unchanged by skipping some checks.
+   * Defaults to true.
+   */
```
johann-beleites-sonarsource, line 92: Technically, this is the override flag, which is `null` when not set. And whether or not we skip unchanged files depends on the context & whether or not SQ/SC supports it. Perhaps the `Defaults to true` could be confusing, as it actually won't always be enabled by default?

sonar-java-plugin/src/main/resources/static/documentation.md
```diff
@@ -63,6 +63,19 @@ Example: `sonar.java.source=1.6`
 
 If the property is provided, the analysis will take the source version into account, and execute related rules accordingly. At run time, each of these rules will be executed – or not – depending of the Java version used by sources within the project. For instance, on a correctly configured project built with Java 6, rules targeting Java 7 and Java 8 will never raise issues, even though they are enabled in the associated rule profile.
 
+## Skipping unchanged files
+Starting from April 2022, and by default, the Java analyzer optimizes the analysis of unchanged files in pull requests.
+In practice, this means that on a file that has not changed, the analyzer only runs a restricted list of rules.
+To get a better understanding of the rule exclusion mechanism, keep in mind that:
+* Rules that need to run on multiple files to decide whether they need to raise issues are always executed
+* Rules that need to run at the end of the analysis to decide whether they need to raise issues are always executed
+* Rules that are defined out of the `sonar.java.checks` package are always executed
```
johann-beleites-sonarsource, line 72:
```suggestion
* Rules that are defined outside of the `sonar.java.checks` package are always executed
```

sonar-java-plugin/src/main/resources/static/documentation.md
```diff
@@ -63,6 +63,19 @@ Example: `sonar.java.source=1.6`
...
+
+This last criteria implies that [custom rules](https://redirect.sonarsource.com/doc/java-custom-rules-guide.html) cannot be skipped.
+
+If you wish to disable this optimization, you can set the value of the analysis parameter `sonar.java.skipUnchanged` to `false`.
+Leaving the parameter untouched is equivalent to having `sonar.java.skipUnchanged` set to `true`.
```
johann-beleites-sonarsource, line 77: Technically, leaving it unset means that we let SQ/SC decide, and this may sometimes be false. We can, in fact, force incremental mode to be used even when outside of a PR analysis by overriding the parameter with `true`.

johann-beleites-sonarsource (approved):
A couple of small suggestions, otherwise LGTM!

PR #4025, opened 2022-03-31
SONARJAVA-4214 Compiler flag "enablePreviewFeatures" should be enabled

quentin-jaquier-sonarsource:
… for java version >= maximum supported version

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/model/JParserConfig.java
+++ b/java-frontend/src/main/java/org/sonar/java/model/JParserConfig.java
@@ -33,7 +33,6 @@ import java.util.function.BiFunction;
 import java.util.function.BooleanSupplier;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
-import javax.annotation.Nullable;
 import org.eclipse.core.runtime.OperationCanceledException;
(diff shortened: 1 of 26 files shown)


java-frontend/src/main/java/org/sonar/java/model/JParserConfig.java
```diff
@@ -173,7 +173,7 @@ public void acceptAST(String sourceFilePath, CompilationUnit ast) {
             executionTimeReport.start(inputFile);
             Result result;
             try {
-              result = new Result(JParser.convert(javaVersion, inputFile.filename(), inputFile.contents(), ast));
+              result = new Result(JParser.convert(javaVersion.toString(), inputFile.filename(), inputFile.contents(), ast));
```
alban-auzeill, line 176: Looking at `javaVersion.toString()` the reader could think there's a risk to return `"none"` to the lexer. In fact, it's not possible because of a previous call to `JParserConfig.effectiveJavaVersion`, but it's not obvious to understand. It would be easier to understand by moving the logic to JavaVersion and have a getter like `String javaVersion.effectiveVersion()`.

PR #4031, opened 2022-04-01
SONARJAVA-4215 S1943 (default system encoding) should not report an issue for Java >= 18

alban-auzeill:


java-checks/src/test/java/org/sonar/java/checks/DefaultEncodingUsageCheckTest.java
```diff
@@ -20,18 +20,44 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.sonar.java.checks.verifier.TestUtils.testSourcesPath;
 
 class DefaultEncodingUsageCheckTest {
 
   @Test
-  void test() {
+  void test_java_version_not_set() {
+    // Once JavaVersionImpl.MAX_SUPPORTED == 18, the following assertion should become verifyNoIssues()
+    assertThat(JavaVersionImpl.MAX_SUPPORTED).isNotEqualTo(18);
```
quentin-jaquier-sonarsource, line 36: When the version is not set, we will **parse** the code with the maximum supported version, but from a rule perspective, the value will stay at unset (`-1`). It means that this assertion is not needed, as the test will pass even with a `MAX_SUPPORTED` = 18.

quentin-jaquier-sonarsource (approved):
LGTM

PR #4036, opened 2022-04-05
SONARJAVA-4222  Rule S6418: Hard-coded secrets are security-sensitive

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/AbstractHardCodedCredentialChecker.java
```diff
@@ -0,0 +1,228 @@
...
+  private Stream<Pattern> variablePatterns() {
+    if (variablePatterns == null) {
+      variablePatterns = toPatterns("");
+    }
+    return variablePatterns.stream();
+  }
+
+  private Stream<Pattern> literalPatterns() {
+    if (literalPatterns == null) {
+      literalPatterns = toPatterns("=\\s*+([^\\\\ &;#]+)");
```
alban-auzeill, line 86: We could add ',' and '|' as separators.

quentin-jaquier-sonarsource (approved):

PR #4039, opened 2022-04-07
SONARJAVA-4223 S5693: Remove requirement to re-parse files on each PR

dorian-burihabwa-sonarsource:


johann-beleites-sonarsource (changes requested):
As discussed, let's refactor to use boolean fields instead of lists, we'll do the cache optimizations in a separate PR

java-checks/src/main/java/org/sonar/java/checks/security/ExcessiveContentRequestCheck.java
```diff
@@ -226,4 +280,56 @@ private static Optional<String> getIdentifierName(ExpressionTree expression) {
...
+  public static void keepForNextAnalysis(CacheContext cacheContext, InputFile inputFile, boolean instantiates, boolean setsMaximumSize) {
+    JavaWriteCache writeCache = cacheContext.getWriteCache();
+    try {
+      if (instantiates) {
+        writeCache.copyFromPrevious(CACHE_KEY_INSTANTIATE + ":" + inputFile.key());
+      }
+      if (setsMaximumSize) {
+        writeCache.copyFromPrevious(CACHE_KEY_SET_MAXIMUM_SIZE + ":" + inputFile.key());
+      }
+      writeCache.copyFromPrevious(CACHE_KEY_CACHED + ":" + inputFile.key());
```
johann-beleites-sonarsource, line 311: Can we also use `computeCacheKey` here?

johann-beleites-sonarsource (approved):
One small comment, otherwise LGTM!

PR #4051, opened 2022-04-21
SONARJAVA-4173 Rule S6411 Types used as keys in Maps should implement Comparable

chrislain-razafimahefa-sonarsource:
Uses ParameterizedTypeTree as the basis of the solution.


alban-auzeill:
Usually, we set the PR title to: `jira id` + `jira summary`, so it's ready to be `squash and merge`:
`SONARJAVA-4173 Rule S6411 Types used as keys in Maps should implement Comparable`

java-checks-test-sources/src/main/files/non-compiling/checks/MapKeyNotComparableCheck.java
```diff
@@ -0,0 +1,41 @@
...
+
+  static class MyComparableType implements Comparable {
+    @Override
+    public int compareTo(@NotNull Object o) {
+      return 0;
+    }
+  }
+
+  static class Program {
+    Map<UnknownType, Object> nonComparable = new HashMap<>(); // Compliant false positive
```
alban-auzeill, line 27: If it's compliant it's never a false positive, it could just be a false negative

java-checks/src/main/java/org/sonar/java/checks/MapKeyNotComparableCheck.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  private static boolean isGeneric(IdentifierTree mapKeyTree) {
+    return JUtils.isTypeVar(mapKeyTree.symbolType());
+  }
+
+  private static Optional<IdentifierTree> getMapKeyTree(ParameterizedTypeTree typeTree) {
+    TypeArguments typeArgs = typeTree.typeArguments();
+    if (typeArgs.size() > 1) {
```
alban-auzeill, line 61: I would test `(typeArgs.size() == 2)` because if you have a subtype of Map like `MyMap<Object, Object, Object>` you are not sure if the key type is the first or the second.

java-checks/src/main/java/org/sonar/java/checks/MapKeyNotComparableCheck.java
```diff
@@ -0,0 +1,73 @@
...
+    return Collections.singletonList(Tree.Kind.PARAMETERIZED_TYPE);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ParameterizedTypeTree typeTree = (ParameterizedTypeTree) tree;
+    Type type = typeTree.symbolType();
+    if (type.isSubtypeOf("java.util.Map")) {
+      getMapKeyTree(typeTree).ifPresent(mapKeyTree -> {
+        if (!isGeneric(mapKeyTree) && !implementsComparable(mapKeyTree.symbolType())) {
```
alban-auzeill, line 48: If `mapKeyTree.symbolType()` is unknown. Then `isGeneric` and `isSubtypeOf` are always `false`. So `!isGeneric` and `!isSubtypeOf` are always `true`, and we could report a false positive.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6411_java.html
```diff
@@ -0,0 +1,54 @@
...
+  href="https://github.com/openjdk/jdk/blob/4927ee426aedbeea0f4119bac0a342c6d3576762/src/hotspot/share/runtime/synchronizer.cpp#L760-L798">https://github.com/openjdk/jdk/blob/4927ee426aedbeea0f4119bac0a342c6d3576762/src/hotspot/share/runtime/synchronizer.cpp#L760-L798</a> </li>
+  <li> <a
+  href="https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Comparable.html">https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Comparable.html</a> </li>
+</ul>
+<h3>Implementation Specification</h3>
+<h3>Message</h3>
+<p>The key type should implement Comparable.</p>
+<h3>Highlighting</h3>
+<p>Variable type, method return type and expressions.</p>
+
```
alban-auzeill, line 54: When we generate the html file from rspec this paragraph should not be present. We should fix rspec.

java-checks/src/main/java/org/sonar/java/checks/MapKeyNotComparableCheck.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static boolean isGeneric(IdentifierTree mapKeyTree) {
+    return JUtils.isTypeVar(mapKeyTree.symbolType());
+  }
+
+  private static Optional<IdentifierTree> getMapKeyTree(ParameterizedTypeTree typeTree) {
+    TypeArguments typeArgs = typeTree.typeArguments();
+    if (typeArgs.size() > 1) {
+      Tree mapKeyTree = typeArgs.get(0);
+      if (mapKeyTree.is(Tree.Kind.IDENTIFIER)) {
```
alban-auzeill, line 63: `mapKeyTree.is(Tree.Kind.IDENTIFIER) == false` is not covered by tests, and we could also support MEMBER_SELECT like in: `Map<Map.Entry, Object>`

alban-auzeill (changes requested):

alban-auzeill (approved):

java-checks-test-sources/src/main/files/non-compiling/checks/MapKeyNotComparableCheck.java
```diff
@@ -22,8 +22,8 @@ public int compareTo(@NotNull Object o) {
   static class UnknownClass implements Unknown {}
 
   static class Program {
-    Map<UnknownClass, Object> nonComparable = new HashMap<>(); // Noncompliant false positive
-    Map<UnknownType, Object> nonComparable = new HashMap<>(); // Potential false negative
+    Map<UnknownClass, Object> nonComparable = new HashMap<>(); // Potential false negative cause by unknown hierarchy
```
chrislain-razafimahefa-sonarsource, line 25: It should be 'caused' instead of 'cause'.

PR #4052, opened 2022-04-21
SONARJAVA-4225 Use caching for S1228

johann-beleites-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/AbstractPackageInfoChecker.java
```diff
@@ -0,0 +1,99 @@
...
+      // default package
+      return;
+    }
+
+    processFile(context, packageName);
+  }
+
+  protected static Optional<String> getPackageFromCache(InputFileScannerContext inputFileScannerContext) {
+    var cacheKey = cacheKey(inputFileScannerContext.getInputFile());
+    try (var in = inputFileScannerContext.getCacheContext().getReadCache().read(cacheKey)) {
```
alban-auzeill, line 78: What a strange way of reading from a cache: `InputStream read(String key)`. It means the `read` method has to create an exception with its stack trace just to indicate that the cache does not contain this key...
When a project is not analyzed incrementally, this rule will generate a lot of exceptions and a lot of `LOG.debug` messages even if there are no analyzer errors. And I see only complexity and disadvantages when I see an `InputStream` instead of a `byte[]`.

java-checks/src/test/java/org/sonar/java/checks/MissingPackageInfoCheckTest.java
```diff
@@ -0,0 +1,186 @@
...
+        this.context = inputFileScannerContext;
+        return super.scanWithoutParsing(inputFileScannerContext);
+      }
+
+      @Override
+      public void scanFile(JavaFileScannerContext context) {
+        this.context = context;
+        super.scanFile(context);
+      }
+    });
```
alban-auzeill, line 115: It's strange to override the class that you want to test. How could you be sure that the class will behave the same without those overrides?
And the class `MissingPackageInfoCheck` receives `InputFileScannerContext context` as a method parameter for `processFile` and as a field for `endOfAnalysis`.
Why not add a `setContext` method that will be also used in the test, and change `processFile` to also depends on the field `context`.

alban-auzeill (changes requested):

PR #4065, opened 2022-04-26
SONARJAVA-4217 report only when the argument type is the same as the wrapper

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ToStringUsingBoxingCheck.java
```diff
@@ -85,13 +88,41 @@ public void visitNode(Tree tree) {
...
+        return true;
+      }
+      if (argumentType.is("int")) {
+        return memberSelectAsPrimitive.is("double")
+          || memberSelectAsPrimitive.is("float")
+          || memberSelectAsPrimitive.is("long");
+      }
+      if (argumentType.is("float")) {
+        return memberSelectAsPrimitive.is("double");
+      }
```
alban-auzeill, line 114: I think we miss the `short` type:
```
    byte myByte = 4;
    new Short(myByte).toString(); // Noncompliant {{Call the static method Short.toString(...) instead of instantiating a temporary object.}}
```
Perhaps we can just test `argumentType.isPrimitive()`...

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/ToStringUsingBoxingCheck.java
quentin-jaquier-sonarsource: Good idea, let's do this!

alban-auzeill (approved):

PR #4066, opened 2022-04-26
SONARJAVA-4156 Refactor regex rules using sonar-analyzer-commons

quentin-jaquier-sonarsource:
This first PR migrates all the rules that are strictly equivalent (almost token by token) in Analyzer Commons.


m-g-sonar:
LGTM! Nice work!

PR #4068, opened 2022-04-26
SONARJAVA-4152 Rule S6397: Character classes in regular expressions should not contain only one character

quentin-jaquier-sonarsource:
Related rspec PR: https://github.com/SonarSource/rspec/pull/765


m-g-sonar:
LGTM, except maybe for the rule description, which I believe could be explicited with some borderline case.

quentin-jaquier-sonarsource:
@m-g-sonar Something like this: https://github.com/SonarSource/rspec/pull/974?

PR #4070, opened 2022-04-26
SONARJAVA 4172 S6206 should not report on non-final classes

chrislain-razafimahefa-sonarsource:


quentin-jaquier-sonarsource (changes requested):
As a general observation, we try to group what is only related to the ticket at the same place and keep the changes minimal.
The extra tests and changes that you made (non-compiling) make sense, you should just try to find a way to make it explicit that it is unrelated to the ticket.
For example, keeping two commits: one for the ticket changes, and one for "Improve and clean tests case for S6206". Or stating it explicitly in the PR.

chrislain-razafimahefa-sonarsource:
@quentin-jaquier-sonarsource It is ready now. I have added my fixes.

quentin-jaquier-sonarsource (approved):
LGTM, thanks for taking into account my comments!

PR #4072, opened 2022-04-27
SONARJAVA-4240 S5693 stores a single cache entry per file

dorian-burihabwa-sonarsource:


johann-beleites-sonarsource (changes requested):
As discussed, the only improvement here would be to abstract the usage of integer values in certain positions to mean something. I.e. use a different power-of-2 constant defined in a static field for both states and add ( == bitwise or) them together to get the value stored in the cache. Otherwise looks good!

johann-beleites-sonarsource (approved):
I like the usage of the `CachedResult` here! One more comment, otherwise LGTM!

PR #4076, opened 2022-04-28
SONARJAVA-4160 Refactor S5868 using sonar-analyzer-commons

quentin-jaquier-sonarsource:
The implementation is in fact almost the same, it's just that the helper class code has been moved around.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/regex/CanonEqFlagInRegexCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/regex/CanonEqFlagInRegexCheck.java
@@ -24,17 +24,17 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Pattern;
 import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
 import org.sonarsource.analyzer.commons.regex.RegexParseResult;
 import org.sonarsource.analyzer.commons.regex.ast.CharacterClassTree;
 import org.sonarsource.analyzer.commons.regex.ast.CharacterTree;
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -82,7 +82,7 @@
 
     <sonar.version>9.4.0.54424</sonar.version>
     <sonarlint.plugin.api.version>6.3.0.36253</sonarlint.plugin.api.version>
-    <analyzer.commons.version>1.24.0.969</analyzer.commons.version>
```
quentin-jaquier-sonarsource, line 85: TODO: update this version to a one from master...

m-g-sonar: PR on commons merged!

java-checks/src/main/java/org/sonar/java/checks/regex/CanonEqFlagInRegexCheck.java
```diff
@@ -93,7 +94,9 @@ private static class CharacterVisitor extends RegexBaseVisitor {
 
     @Override
     public void visitSequence(SequenceTree tree) {
-      subjectToNormalization.addAll(getGraphemeInList(tree.getItems()));
+      subjectToNormalization.addAll(getGraphemeInList(tree.getItems()).stream()
```
m-g-sonar, line 97: what about not collecting into a list but directly add in `subjectToNormalization`?
```java
getGraphemeInList(tree.getItems())
  .stream()
  .map(RegexIssueLocation::fromCommonsRegexIssueLocation)
  .forEach(subjectToNormalization::add)
```

m-g-sonar (commented):
LGTM, except that we can probably slightly simplify the collection of the grapheme.

java-checks/src/main/java/org/sonar/java/checks/regex/CanonEqFlagInRegexCheck.java
quentin-jaquier-sonarsource: Indeed, way better!

pom.xml
quentin-jaquier-sonarsource: I updated the version to rely on `1.24.0.973`.

PR #4077, opened 2022-04-28
SONARJAVA-4219 Fix handling of assignments in switch cases

m-g-sonar:


java-checks-test-sources/src/main/java/checks/AssignmentInSubExpressionCheck.java
```diff
@@ -128,4 +128,61 @@ void fun(List<Integer> list) {
...
+    private String b = "toto";
+    void qix() {
+      String s = switch (RAND.nextInt(10)) {
+        case 1 -> b = "yolo"; // Noncompliant
+        case 2 -> {
+          b = "tagada"; // Compliant
+          yield "turlutte";
+        }
+        default -> "boom";
+      };
```
quentin-jaquier-sonarsource, line 185: I may sound grumpy, but I think that we should avoid such string names and try to find ones describing what is happening for example (especially when some can have hidden meaning 😆)

quentin-jaquier-sonarsource (approved):

java-checks-test-sources/src/main/java/checks/AssignmentInSubExpressionCheck.java
m-g-sonar: i agree with you. I actually forgot that I wrote this...

PR #4085, opened 2022-05-02
Fix quality flow: Simplify useless reluctant quantifier

quentin-jaquier-sonarsource:


java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/internal/Expectations.java
```diff
@@ -363,7 +363,7 @@ static class Parser extends IssuableSubscriptionVisitor {
     private Pattern shift = Pattern.compile(NONCOMPLIANT_FLAG + "@(\\S+)");
 
     private static final Pattern FLOW_COMMENT = Pattern.compile("//\\s+flow");
-    private static final Pattern FLOW = Pattern.compile("flow@(?<ids>\\S+).*?(?=flow@)?");
```
quentin-jaquier-sonarsource, line 366: This code was initially spotted by [S6019](https://sonarsource.github.io/rspec/#/rspec/S6019). The goal of this regex is to collect all "ids". 
```
// flow@id1,id2
```
You can have multiples `flow@` at the same line. 
```
// flow@id1,id2 something else flow@id3
```
The second part was probably added to "consume everything until the next `flow@` comment". This second part is in fact only matching the empty string (see S6019) and is in fact useless as the first part will do the job just fine.

m-g-sonar (approved):
LGTM! thank you for looking at it!

PR #4087, opened 2022-05-02
SONARJAVA-4155 Refactor S5843 using sonar-analyzer-commons

quentin-jaquier-sonarsource:
I had to modify/improve the code in Analyzer Commons. Please review the PR on the Commons side first: https://github.com/SonarSource/sonar-analyzer-commons/pull/226.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/regex/RegexComplexityCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/regex/RegexComplexityCheck.java
@@ -35,23 +35,11 @@ import org.sonar.plugins.java.api.tree.LiteralTree;
 import org.sonar.plugins.java.api.tree.SyntaxTrivia;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonarsource.analyzer.commons.regex.RegexParseResult;
-import org.sonarsource.analyzer.commons.regex.ast.BackReferenceTree;
-import org.sonarsource.analyzer.commons.regex.ast.CharacterClassIntersectionTree;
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -82,7 +82,7 @@
 
     <sonar.version>9.4.0.54424</sonar.version>
     <sonarlint.plugin.api.version>6.3.0.36253</sonarlint.plugin.api.version>
-    <analyzer.commons.version>1.24.0.973</analyzer.commons.version>
+    <analyzer.commons.version>1.24.0.977</analyzer.commons.version>
```
quentin-jaquier-sonarsource, line 85: TODO: update this version.

java-checks/src/main/java/org/sonar/java/checks/regex/RegexComplexityCheck.java
```diff
@@ -177,106 +165,6 @@ private boolean isCommented(LiteralTree regexPart) {
 
   }
 
-  private class ComplexityCalculator extends RegexBaseVisitor {
```
quentin-jaquier-sonarsource, line 180: I was initially confused because the code in the Java analyzer side is way more complex than the implementation on Commons. It turns out that the implementation on Commons contains only a subpart of the one from the analyzer, the analyzer having additional logic to count the complexity of each part individually if the regex is made out of multiple parts.
The implementation of this "ComplexityCalculator" is the same, so the migration is trivial.

pom.xml
quentin-jaquier-sonarsource: Done

alban-auzeill (approved):

PR #4088, opened 2022-05-03
SONARJAVA-2940 S4968: Upper bound of type variable should not be final

chrislain-razafimahefa-sonarsource:
- The rspec page will certainly need some changes, so let's not worry about that for now


chrislain-razafimahefa-sonarsource:
@quentin-jaquier-sonarsource Here is another attempt. It solves all the points you mentioned above (e.g., it does not trigger the rule for overridden methods).
- Integration tests are still not in order.

chrislain-razafimahefa-sonarsource:
Here is the latest version.
- I have not merged autoscan since there is a result I don't understand: why is rule S1128 reported and why are they false negatives?
- I am not super familiar with Optional so if there is a better way than what I did I am very pleased to hear about it.
- There is also a PR for the changes in RSPEC: https://github.com/SonarSource/rspec/pull/980

java-checks/src/main/java/org/sonar/java/checks/TypeUpperBoundNotFinalCheck.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  // Returns true if 'type' is part of the signature of an overriding method.
+  private static boolean inOverridingMethodDeclaration(TypeTree type) {
+    Optional<MethodTree> method = getMethodDeclaration(type);
+    if (method.isPresent()) {
+      return isOverriding(method.get());
+    } else {
+      return false;
+    }
```
quentin-jaquier-sonarsource, line 87:
```suggestion
return getMethodDeclaration(type).filter(TypeUpperBoundNotFinalCheck::isOverriding).isPresent();
```

java-checks/src/main/java/org/sonar/java/checks/TypeUpperBoundNotFinalCheck.java
```diff
@@ -0,0 +1,105 @@
...
+      handleBounds(((TypeParameterTree) tree).bounds(), tree);
+    } else if (tree.is(Tree.Kind.EXTENDS_WILDCARD)) {
+      handleBounds(Collections.singletonList(((WildcardTree) tree).bound()), tree);
+    }
+  }
+
+  private void handleBounds(List<TypeTree> bounds, Tree treeToReport) {
+    for (TypeTree bound : bounds) {
+      if (reportIssueIfBoundIsFinal(bound, treeToReport))
+        return;
```
quentin-jaquier-sonarsource, line 57: Missing curly braces (also reported by SQ).

java-checks/src/main/java/org/sonar/java/checks/TypeUpperBoundNotFinalCheck.java
```diff
@@ -0,0 +1,105 @@
...
+      }
+    }
+    return false;
+  }
+
+  private static boolean isFinal(IdentifierTree bound) {
+    return bound.symbol().isFinal();
+  }
+
+  // Returns true if 'type' is part of the signature of an overriding method.
```
quentin-jaquier-sonarsource, line 80: You can probably use javadoc style instead.

java-checks/src/main/java/org/sonar/java/checks/TypeUpperBoundNotFinalCheck.java
```diff
@@ -0,0 +1,105 @@
...
+    Optional<MethodTree> method = getMethodDeclaration(type);
+    if (method.isPresent()) {
+      return isOverriding(method.get());
+    } else {
+      return false;
+    }
+  }
+
+  // Returns the method where 'type' is in its signature, if any.
+  private static Optional<MethodTree> getMethodDeclaration(TypeTree type) {
```
quentin-jaquier-sonarsource, line 91: We may start from "treeToReport" instead of the type, to spare a few parent() calls.

java-checks/src/main/java/org/sonar/java/checks/TypeUpperBoundNotFinalCheck.java
```diff
@@ -0,0 +1,105 @@
...
+    if (bound.is(Tree.Kind.IDENTIFIER)) {
+      if (isFinal((IdentifierTree) bound) && !inOverridingMethodDeclaration(bound)) {
+        reportIssue(treeToReport, "Replace this type parametrization by the 'final' type.");
+        return true;
+      }
+    } else if (bound.is(Tree.Kind.PARAMETERIZED_TYPE)) {
+      ParameterizedTypeTree type = (ParameterizedTypeTree) bound;
+      if (reportIssueIfBoundIsFinal(type.type(), treeToReport)) {
+        return true;
+      }
```
quentin-jaquier-sonarsource, line 71:
```suggestion
      return reportIssueIfBoundIsFinal(type.type(), treeToReport);
```

java-checks/src/main/java/org/sonar/java/checks/TypeUpperBoundNotFinalCheck.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private void handleBounds(List<TypeTree> bounds, Tree treeToReport) {
+    for (TypeTree bound : bounds) {
+      if (reportIssueIfBoundIsFinal(bound, treeToReport))
+        return;
+    }
+  }
+
+  private boolean reportIssueIfBoundIsFinal(TypeTree bound, Tree treeToReport) {
```
quentin-jaquier-sonarsource, line 61: Let's try to replace this boolean return by a more functional code!

```
  private void handleBounds(List<TypeTree> bounds, Tree treeToReport) {
    boolean hasFinalBound = bounds.stream()
      .map(TypeUpperBoundNotFinalCheck::getIdentifier)
      .flatMap(Optional::stream)
      .allMatch(bound -> isFinal(bound) && !inOverridingMethodDeclaration(bound));
    if (hasFinalBound) {
      reportIssue(treeToReport, "Replace this type parametrization by the 'final' type.");
    }
  }

  private static Optional<IdentifierTree> getIdentifier(TypeTree t) {
    if (t.is(Tree.Kind.IDENTIFIER)) {
      return Optional.of((IdentifierTree) t);
    } else if (t.is(Tree.Kind.PARAMETERIZED_TYPE)) {
      ParameterizedTypeTree type = (ParameterizedTypeTree) t;
      return getIdentifier(type.type());
    }

    return Optional.empty();
  }
```

chrislain-razafimahefa-sonarsource:
It is ready again. Thanks for all the reviews. It made me see a lot.

chrislain-razafimahefa-sonarsource:
And if we could look at the autoscan issue when you have time. it would be great.

quentin-jaquier-sonarsource (approved):
LGTM, thanks for taking into consideration my comments, we ended up with a nice implementation, good work!

chrislain-razafimahefa-sonarsource:
@quentin-jaquier-sonarsource Thanks for all the help!

PR #4089, opened 2022-05-03
SONARJAVA-4159 Refactor S5867 using sonar-analyzer-commons

quentin-jaquier-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/regex/UnicodeAwareCharClassesCheck.java
```diff
@@ -19,113 +19,28 @@
...
-
-  private static final Map<Character, Character> unicodeUnawareCharacterRanges = MapBuilder.<Character, Character>newMap()
-    .put('a', 'z')
-    .put('A', 'Z')
-    .build();
-
   @Override
   public void checkRegex(RegexParseResult regexForLiterals, ExpressionTree methodInvocationOrAnnotation) {
-    new UnicodeUnawareCharClassFinder(methodInvocationOrAnnotation).visit(regexForLiterals);
+    new UnicodeUnawareCharClassFinder(this::reportIssueFromCommons, (message, cost, secondaries) -> {
```
quentin-jaquier-sonarsource, line 35: Small subtility: the Finder takes a second argument, in the case where we want to report an issue on the invocation (and not on the regex).

java-checks/src/main/java/org/sonar/java/checks/regex/UnicodeAwareCharClassesCheck.java
```diff
@@ -19,113 +19,28 @@
...
+      String flagName = methodInvocationOrAnnotation.is(Tree.Kind.ANNOTATION) ? "(?U)" : "UNICODE_CHARACTER_CLASS";
+      reportIssue(methodOrAnnotationName(methodInvocationOrAnnotation),
+        message.replace("\"u\"", String.format("\"%s\"", flagName)),
```
quentin-jaquier-sonarsource, line 38: The difference on the Java side compared to the current implementation on Commons is that we have a small logic to suggest the best flag to use. We can still apply the logic and adjust the message that we receive.

alban-auzeill (approved):

PR #4090, opened 2022-05-04
SONARJAVA-4154 Rule S6331: Regular expressions should not contain empty groups

quentin-jaquier-sonarsource:


its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json
```diff
@@ -2735,6 +2735,12 @@
     "falseNegatives": 0,
     "falsePositives": 0
   },
+  {
+    "ruleKey": "S6331",
+    "truePositives": 47,
+    "falseNegatives": 1,
```
quentin-jaquier-sonarsource, line 2741: This FN is due to a regex present in an annotation. Without the semantic, we are not able to find it.
The is also the reason why we have to increase the "Issues differences" counter in "AutoScanTest"

alban-auzeill (approved):

PR #4091, opened 2022-05-05
SONARJAVA-4163 Refactor S5996 using sonar-analyzer-commons

quentin-jaquier-sonarsource:


java-checks-test-sources/src/main/java/checks/regex/ImpossibleBoundariesCheck.java
```diff
@@ -87,4 +85,16 @@ void compliant(String str) {
...
+  void withMultilineFlag() {
+    Pattern.compile("a\\n^b$", Pattern.MULTILINE); // Compliant, matches "a\nb"
+    Pattern.compile("(?m)a\\n^b$"); // Compliant
+    Pattern.compile("a\\n^b$"); // Noncompliant
+
+    // Even if the usage of "$" is suspicious and redundant, we do not raise an issue
+    Pattern.compile("a$\nb", Pattern.MULTILINE); // Compliant, matches "a\nb"
+    Pattern.compile("(?m)a$\nb"); // Compliant
+    Pattern.compile("(?m)^1$\n2"); // Compliant
+    Pattern.compile("a$\nb"); // Noncompliant
```
quentin-jaquier-sonarsource, line 97: The implementation in Analyzer Commons is taking into account the multiline flag and does report an issue for "start of line" ("^") and "end of line" ("$") with the "Multiline" flag.
The rest of the differences are due to code moves, the implementation (and behavior) are equivalent.

PR #4097, opened 2022-05-10
SONARJAVA-4204 FP on S1221 when a method is overridden

chrislain-razafimahefa-sonarsource:
- I moved the test cases from `~/workspace/sonar-java/java-checks/src/test/files/checks/` to `~/workspace/sonar-java/java-checks-test-sources/src/main/java/checks/` where they are now compilable.


java-checks/src/main/java/org/sonar/java/checks/naming/MethodNamedHashcodeOrEqualCheck.java
```diff
@@ -37,16 +37,23 @@ public List<Tree.Kind> nodesToVisit() {
...
   @Override
   public void visitNode(Tree tree) {
-    IdentifierTree methodIdentifier = ((MethodTree) tree).simpleName();
+    MethodTree method = (MethodTree) tree;
+    IdentifierTree methodIdentifier = method.simpleName();
     switch (methodIdentifier.name()) {
       case "hashcode":
-        report(methodIdentifier, "hashCode()");
+        if (notOverriding(method)) {
+          report(methodIdentifier, "hashCode()");
```
m-g-sonar, line 45: To not duplicate 3 times `notOverriding()` in 3 distincts places, what about renaming `report()` into `reportIfNotOverriding()` and do like this:
```
case "hashcode":
  reportIfNotOverriding(method, "hashCode()");
```

and later replace:
```
private void report(IdentifierTree methodIdentifier, String substitute) {
...
```
by
```
private void reportIfNotOverriding(MethodTree method, String substitute) {
  if (notOverriding(method)) {
    reportIssue(method.simpleName(), "Either override Object." + substitute + ", or totally rename the method to prevent any confusion.");
  }
}
```

m-g-sonar:
Overall, look good to me! the approach is correct! I would just prefer to slightly simplify the code to not duplicate the logic 3 times, since it's always the same.

chrislain-razafimahefa-sonarsource:
Good point! I have made the change. Thanks a lot.

PR #4098, opened 2022-05-11
SONARJAVA-4246 Expose EndOfAnalysis in internal part of public API

johann-beleites-sonarsource:


java-checks-testkit/src/test/java/org/sonar/java/checks/verifier/internal/InternalCheckVerifierTest.java
```diff
@@ -968,7 +970,7 @@ void withCache_effectively_sets_the_caches_for_scanWithoutParsing() {
       .verifyNoIssues();
 
     verify(check, times(1)).scanWithoutParsing(argThat(context -> equivalent(cacheContext, context.getCacheContext())));
-    verify(check, times(1)).endOfAnalysis(cacheContext);
+    verify(check, times(1)).endOfAnalysis(any());
```
dorian-burihabwa-sonarsource, line 973:
```suggestion
    verify(check, times(1)).endOfAnalysis(argThat(moduleContext -> equivalent(cacheContext, moduleContext.getCacheContext())));
```

java-frontend/src/main/java/org/sonar/java/model/DefaultModuleScannerContext.java
```diff
@@ -72,6 +73,10 @@ public CacheContext getCacheContext() {
     return cacheContext;
   }
 
+  public void reportIssue(AnalyzerMessage message) {
```
dorian-burihabwa-sonarsource, line 76: Should we promote this at the interface level?

dorian-burihabwa-sonarsource (changes requested):
Looks good overall, a small fix would help ensure that we pass the right caches.
Also should we advertize `reportIssue(AnalyzerMessage)` in the `ModuleScannerContext` interface?

dorian-burihabwa-sonarsource (approved):
All good

PR #4099, opened 2022-05-11
Rely on Analyzer Commons for regex helper classes

quentin-jaquier-sonarsource:
Now that every ticket of [MMF-2661](https://jira.sonarsource.com/browse/MMF-2661) are done, we can clean the rest of the duplicated code.
I compared the files that I removed with the implementation on Analyzer commons and made sure there was equivalent (they are strictly equivalent for most of them).
I just had to add a public method on Commons side: https://github.com/SonarSource/sonar-analyzer-commons/pull/229.


its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json
```diff
@@ -2425,7 +2425,7 @@
   },
   {
     "ruleKey": "S5852",
-    "truePositives": 73,
+    "truePositives": 75,
```
quentin-jaquier-sonarsource, line 2428: To be honest, I'm really unsure how it is possible to have a different result here, without modifying the test code...

its/ruling/src/test/resources/regex-examples/java-S5852.json
```diff
@@ -1,6 +1,7 @@
 {
 'org.regex-examples:regex-examples:src/main/java/org/regex/examples/RegexDatabase1.java':[
 33,
+49,
```
quentin-jaquier-sonarsource, line 4: The differences in this rule are due to the fix made on Analyzer Commons side: https://github.com/SonarSource/sonar-analyzer-commons/pull/159.

I double-checked, and the change makes sense: we are now correctly able to determine when a regex is anchored at end.  The new issues looks like new true positives!

pom.xml
```diff
@@ -82,7 +82,7 @@
 
     <sonar.version>9.4.0.54424</sonar.version>
     <sonarlint.plugin.api.version>6.3.0.36253</sonarlint.plugin.api.version>
-    <analyzer.commons.version>1.24.0.986</analyzer.commons.version>
+    <analyzer.commons.version>1.24.0.990</analyzer.commons.version>
```
quentin-jaquier-sonarsource, line 85: TODO: change this version to the master one.

m-g-sonar: `1.24.0.991`

PR #4100, opened 2022-05-11
SONARJAVA-4188 S4423 should not report an issue when the version is not set

chrislain-razafimahefa-sonarsource:
- Assumes that when the project version is not set, the version is >= Java 8 (a very likely proposition because someone using Sonar is most likely using Java >= 8). Therefore when the version is not set, the rule now behaves as if the version is >= Java 8.
- Had to add a new test source file specifically for the case where the java version is set and set to Java 7, because this was not covered before.


java-checks/src/test/java/org/sonar/java/checks/WeakSSLContextCheckTest.java
```diff
@@ -34,6 +34,15 @@ void test() {
       .verifyIssues();
   }
 
+  @Test
+  void test_java_7() {
+    CheckVerifier.newVerifier()
+      .onFile(testSourcesPath("checks/WeakSSLContextCheckJava7.java"))
```
quentin-jaquier-sonarsource, line 40: Instead of adding a new file with only a subset of the test, I suggest:
- Use `WeakSSLContextCheckJava8.java` for both `test_java_8`  and `test` (the behavior should be the same).
- Use the `WeakSSLContextCheck.java` for `test_java_7` (without the code changes).

java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
```diff
@@ -56,11 +56,11 @@ public class WeakSSLContextCheck extends IssuableSubscriptionVisitor {
     .withAnyParameters()
     .build();
 
-  private boolean projectHasJava8OrHigher;
+  private boolean projectNotSetOrHasJava8OrHigher;
```
quentin-jaquier-sonarsource, line 59: Not sure that "project not set" makes sense, it is the java version that is not set.
Maybe rename it to `projectHasJavaNotSetOr8OrHigher` or even something like `javaVersionNotSetOrSuperiorTo8`.

java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
```diff
@@ -93,7 +93,8 @@ public void visitNode(Tree tree) {
   }
 
   private boolean isStrongProtocol(String protocol) {
-    return STRONG_PROTOCOLS.contains(protocol) || (projectHasJava8OrHigher && STRONG_AFTER_JAVA_8.contains(protocol));
+    // A project not set is very likely to be >= Java8
```
quentin-jaquier-sonarsource, line 96:
```suggestion
    // A project with a Java version not set is very likely to be >= Java8
```

java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
chrislain-razafimahefa-sonarsource: Right. Changed to `javaVersionNotSetOr8OrHigher`.

java-checks/src/test/java/org/sonar/java/checks/WeakSSLContextCheckTest.java
chrislain-razafimahefa-sonarsource: Good one!

quentin-jaquier-sonarsource (approved):
LGTM

PR #4101, opened 2022-05-11
Fix bug where we were not properly calling canSkipUnchangedFiles API …

johann-beleites-sonarsource:
…due to the incorrect usage of reflection


dorian-burihabwa-sonarsource (approved):
Pair-programmed and reviewed 👍🏿

PR #4108, opened 2022-05-16
SONARJAVA-4243 FP in S6205 when the content of the block is not an expression

chrislain-razafimahefa-sonarsource:


java-checks-test-sources/src/main/java/checks/SwitchRedundantKeywordCheck.java
```diff
@@ -42,6 +42,29 @@ int switchExpression(String mode) {
...
+      case "b" -> {
+        if (mode.equals("b")) {
+          doSomethingElse();
+        }
+      }
+      case "c" -> { // compliant, return is not an expression
+        return 1;
+      }
+      case "d" -> {
+        throw new IllegalArgumentException();
```
alban-auzeill, line 63: This is a false negative. We should report an issue because the throw statement does not need to be in a block.

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #4117, opened 2022-06-15
Disable DBD for now, to test incremental analysis

johann-beleites-sonarsource:


dorian-burihabwa-sonarsource:
No longer needed as DBD should support incremental analysis in the very near future.

PR #4123, opened 2022-06-22
Sonarjava 4138 Rule S6262 AWS region should not be set with a hardcoded String

chrislain-razafimahefa-sonarsource:


chrislain-razafimahefa-sonarsource:
Modification to RSPEC for rule S6262 needs to be reviewed as well:
https://github.com/SonarSource/rspec/pull/1069

java-checks/src/main/java/org/sonar/java/checks/aws/AwsRegionSetterCheck.java
```diff
@@ -0,0 +1,54 @@
...
+  @Override
+  protected MethodMatchers getMethodInvocationMatchers() {
+    return REGION_SETTER_MATCHER;
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree tree) {
+    // The methodmatcher ensures that there is one and only one argument.
+    ExpressionTree argument = tree.arguments().get(0);
+    if (argument.is(Kind.STRING_LITERAL, Kind.IDENTIFIER)) {
```
alban-auzeill, line 50: With STRING_LITERAL we will not have any FP, but with IDENTIFIER we need to check that it's a constant of this class that can be easily changed by the developer. Because in the case of an IDENTIFIER matching a method argument, the value could come from a location outside of developer responsibility, and I consider this as an FP.

java-checks/src/test/java/org/sonar/java/checks/aws/AwsRegionSetterCheckTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.sonar.java.checks.verifier.CheckVerifier;
+
+class AwsRegionSetterCheckTest {
+
+  @Test
+  void test() {
+    CheckVerifier.newVerifier()
+      .onFile("src/test/files/checks/aws/AwsRegionSetterCheck.java")
```
alban-auzeill, line 30: the test file should be in the `java-checks-test-sources/src/main/java/checks/aws` directory, and it should be referenced with `.onFile(mainCodeSourcesPath("checks/aws/...`

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/aws/AwsRegionSetterCheck.java
alban-auzeill, line 50: We could use `org.sonar.java.checks.helpers.ExpressionsHelper#getConstantValueAsString(ExpressionTree)` if we are able to check that the location are in the same file or package.

its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json
```diff
@@ -361,9 +361,9 @@
   },
   {
     "ruleKey": "S1181",
-    "truePositives": 3,
-    "falseNegatives": 12,
-    "falsePositives": 0
+    "truePositives": 0,
+    "falseNegatives": 10,
+    "falsePositives": 3
```
chrislain-razafimahefa-sonarsource, line 366: I don't understand why S1181 enters into the picture and why we have 3 new false positives.

PR #4124, opened 2022-06-22
SONARJAVA-4133 Rule S6241 Region should be set explicitly when creating a new AwsClient

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/aws/AwsRegionShouldBeSetExplicitlyCheck.java
```diff
@@ -0,0 +1,217 @@
...
+        return;
+      }
+      // If no call to region is found in the call, we go to the other usages
+      List<IdentifierTree> usages = actualDeclaration.symbol().usages();
+      // If one of the usages is passing the builder to a method, we assume it might set there
+      if (usages.stream().anyMatch(AwsRegionShouldBeSetExplicitlyCheck::isPassedAsArgument)) {
+        return;
+      }
+      // If the usage sets the region, we return
+      if (usages.stream().anyMatch(AwsRegionShouldBeSetExplicitlyCheck::setsRegion)) {
```
johann-beleites-sonarsource, line 90: We are creating & consuming the same stream twice here - WDYT of having `or`ed boolean expressions in the `anyMatch` only once?

java-checks/src/main/java/org/sonar/java/checks/aws/AwsRegionShouldBeSetExplicitlyCheck.java
```diff
@@ -0,0 +1,217 @@
...
+  private static class CallChainVisitor extends BaseTreeVisitor {
+    protected boolean methodIsInvoked;
+    private MethodMatchers target;
+
+    CallChainVisitor(MethodMatchers matcher) {
+      this.target = matcher;
+    }
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree tree) {
```
johann-beleites-sonarsource, line 158: This kind of code looks like an excellent candidate for generalization. Since you're going to do the ticket about setting the credentials provider next, I assume you already plan to generalize it then anyway

johann-beleites-sonarsource (approved):
Looks good! I only have a couple of minor comments. Do you think we can increase the coverage a little?

java-checks/src/main/java/org/sonar/java/checks/aws/AwsRegionShouldBeSetExplicitlyCheck.java
dorian-burihabwa-sonarsource: Will do

java-checks/src/main/java/org/sonar/java/checks/aws/AwsRegionShouldBeSetExplicitlyCheck.java
```diff
@@ -0,0 +1,217 @@
...
+        return;
+      }
+      // If the usage sets the region, we return
+      if (usages.stream().anyMatch(AwsRegionShouldBeSetExplicitlyCheck::setsRegion)) {
+        return;
+      }
+      reportIssue(actualDeclaration, "Set the region explicitly on this builder.");
+    } else {
+      reportIssue(invocation, "Set the region explicitly on this builder.");
+    }
```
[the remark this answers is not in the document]
dorian-burihabwa-sonarsource: I will try and fix that

java-checks/src/main/java/org/sonar/java/checks/aws/AwsRegionShouldBeSetExplicitlyCheck.java
dorian-burihabwa-sonarsource: Indeed that is part of what can be easily extracted

PR #4126, opened 2022-06-23
SONARJAVA-4135 Reusable resources should be initialized at constructi…

johann-beleites-sonarsource:
…on time of Lambda functions


dorian-burihabwa-sonarsource (approved):
Good job on the deeper search for method invocations 👍🏿 Technically, I think that we should exclude constructor calls that use one of the `handleRequest` arguments as one of its own. This way we avoid reporting on legitimate method invocations. WDYT?

PR #4130, opened 2022-06-27
SONARJAVA-4134 Rule S6242: Credentials Provider should be set explicitly when creating a new "AwsClient"

dorian-burihabwa-sonarsource:


java-checks-test-sources/src/main/files/non-compiling/checks/aws/AwsRegionShouldBeSetExplicitlyCheck.java
```diff
@@ -0,0 +1,8 @@
+package checks.aws;
```
johann-beleites-sonarsource, line 1: Correct me if I'm wrong - if we introduce this file, don't we also need a corresponding test that uses `.onFiles(nonCompilingTestSourcesPath("checks/aws/AwsRegionShouldBeSetExplicitlyCheck.java"))`?

java-checks/src/main/java/org/sonar/java/checks/aws/AwsBuilderMethodFinder.java
```diff
@@ -0,0 +1,222 @@
...
+  protected static final String AWS_CLIENT_BUILDER_TYPE = "software.amazon.awssdk.awscore.client.builder.AwsClientBuilder";
+  private static final MethodMatchers BUILD_METHOD = MethodMatchers.create()
+    .ofSubTypes(SDK_CLIENT_BUILDER_TYPE)
+    .names("build")
+    .addWithoutParametersMatcher()
+    .build();
+
+  protected MethodMatchers getBuildMethod() {
+    return BUILD_METHOD;
+  }
```
johann-beleites-sonarsource, line 50: Wen we use `BUILD_METHOD` directly instead of wrapping it in this method?

johann-beleites-sonarsource (approved):
Looks good, only a couple of minor remarks, then LGTM!

java-checks/src/main/java/org/sonar/java/checks/aws/AwsBuilderMethodFinder.java
dorian-burihabwa-sonarsource: Indeed, I will remove the method

PR #4131, opened 2022-06-27
SONARJAVA-4137 Rule S6246 Lambdas should not invoke other lambdas synchronously

chrislain-razafimahefa-sonarsource:


chrislain-razafimahefa-sonarsource:
There is also RSPEC to review: https://github.com/SonarSource/rspec/pull/1082

java-checks/src/main/java/org/sonar/java/checks/aws/AwsLambdaSyncCallCheck.java
```diff
@@ -0,0 +1,184 @@
...
+     * Returns true if in the call chain starting at 'tree' there is a call that
+     * sets 'InvocationType' to async.
+     *
+     * @param tree an Identifier or a MethodInvocation as starting point of a call chain
+     * @return true if InvocationType is set to async
+     */
+    private static boolean callChainSetsAsyncCall(Tree tree) {
+      Tree treeParent = tree.parent();
+      if (treeParent != null && treeParent.parent() != null &&
+        treeParent.parent().is(Tree.Kind.METHOD_INVOCATION)) {
```
johann-beleites-sonarsource, line 134: Take the following case:
```
      var invokeRequest4 = new InvokeRequest();
      (invokeRequest4).setInvocationType("Event");
      awsLambda.invoke(invokeRequest4);
```
We don't detect the async setter here, as the invocation target is in parentheses, leading to an FP.

johann-beleites-sonarsource (changes requested):
Nice work, this rule is quite involved! I think we can improve to avoid throwing exceptions or raising FPs in a couple of places - see comments below.

java-checks/src/main/java/org/sonar/java/checks/aws/AwsLambdaSyncCallCheck.java
```diff
@@ -0,0 +1,191 @@
...
+      declaration.accept(asyncSetterVisitor);
+      return asyncSetterVisitor.found();
+    }
+
+    private static final class AsyncInvocationTypeSetterFinder extends BaseTreeVisitor {
+      private boolean found = false;
+
+      @Override
+      public void visitMethodInvocation(MethodInvocationTree methodCall) {
+        found = setsInvocationTypeToAsync(methodCall) || found;
```
johann-beleites-sonarsource, line 168: Would turning around the conditions be more efficient (i.e. once `found == true`, we wouldn't have to call `setsInvocationTypeToAsync` anymore)?

java-checks/src/main/java/org/sonar/java/checks/aws/AwsLambdaSyncCallCheck.java
```diff
@@ -0,0 +1,191 @@
...
+    private static boolean statementSetsAsyncCall(Tree tree) {
+      MethodInvocationTree methodCall = findMethodInvocationTreeAncestor(tree);
+      if (methodCall != null) {
+        if (setsInvocationTypeToAsync(methodCall)) {
+          return true;
+        } else {
+          return statementSetsAsyncCall(methodCall);
+        }
+      }
+      return false;
```
johann-beleites-sonarsource, line 135: I believe this could be simplified to:
```suggestion
      return (methodCall != null && (setsInvocationTypeToAsync(methodCall) || statementSetsAsyncCall(methodCall)));
```

johann-beleites-sonarsource (changes requested):
Thanks for the improvements! There are really some tricky edge cases in a rule like this one, I found one more and a couple of optimization suggestions.

java-checks/src/main/java/org/sonar/java/checks/aws/AwsLambdaSyncCallCheck.java
```diff
@@ -0,0 +1,191 @@
...
+     * Returns true if the declaration of the 'invokeRequest' variable leads to async lambda calls.
+     *
+     * @param invokeRequest the 'invokeRequest' varible being declared
+     * @return true if the declaration of the 'invokeRequest' variable leads to async lambda calls
+     */
+    private static boolean declarationSetsAsyncCall(IdentifierTree invokeRequest) {
+      Tree declaration = invokeRequest.symbol().declaration();
+      AsyncInvocationTypeSetterFinder asyncSetterVisitor = new AsyncInvocationTypeSetterFinder();
+      // There has to be a declaration (if code compiled)
+      declaration.accept(asyncSetterVisitor);
```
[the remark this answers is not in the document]
chrislain-razafimahefa-sonarsource: Indeed! I tested this "manually" and it failed.
The fix is shown to pass on the manual test.

PR #4132, opened 2022-06-28
SONARJAVA-4139 S6263: Using long-term access keys is security-sensitive

johann-beleites-sonarsource:


java-checks-test-sources/src/main/java/checks/aws/AwsLongTermAccessKeysCheck.java
```diff
@@ -0,0 +1,18 @@
+package checks.aws;
+
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.auth.BasicSessionCredentials;
+
+public class AwsLongTermAccessKeysCheck {
+  void noncompliant() {
+    BasicAWSCredentials foo = new BasicAWSCredentials("", ""); // Noncompliant [[sc=35;ec=54]] {Make sure using a Long-term access key is safe here.}
```
dorian-burihabwa-sonarsource, line 8: The issue message should be checked by putting the expected string between double accolades.
```suggestion
    BasicAWSCredentials foo = new BasicAWSCredentials("", ""); // Noncompliant [[sc=35;ec=54]] {{Make sure using a Long-term access key is safe here.}}
```

java-checks/src/main/java/org/sonar/java/checks/aws/AwsLongTermAccessKeysCheck.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.sonar.check.Rule;
+import org.sonar.java.checks.methods.AbstractMethodDetection;
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
+import org.sonar.plugins.java.api.tree.NewClassTree;
+
+@Rule(key = "S6263")
+public class AwsLongTermAccessKeysCheck extends AbstractMethodDetection {
+
+  private static final String MESSAGE = "Make sure using a Long-term access key is safe here.";
```
dorian-burihabwa-sonarsource, line 30: I am seeing this in the rspec and the ticket as well but should "Long" be capitalized here? Does it refer to some kind of IP on a technology?

dorian-burihabwa-sonarsource (changes requested):
LGTM but let's make sure we test the issue message

dorian-burihabwa-sonarsource (approved):

PR #4136, opened 2022-06-30
SONARJAVA-4276 Message of S4968 should end with a full stop

quentin-jaquier-sonarsource:


chrislain-razafimahefa-sonarsource (approved):
@quentin-jaquier-sonarsource Thanks for taking care of the fix.

PR #4145, opened 2022-07-29
SONARJAVA-4131 FP in SE related to org.springframework.util.StringUtils#isEmpty

chrislain-razafimahefa-sonarsource:


java-checks-test-sources/src/main/java/symbolicexecution/behaviorcache/SpringStringUtilsMethods.java
```diff
@@ -0,0 +1,17 @@
+package symbolicexecution.behaviorcache;
+
+import org.springframework.util.StringUtils;
+
+class SpringStringUtilsMethods {
+
+  String test_isEmpty() {
+    String fileName = null;
+    if (!StringUtils.isEmpty(fileName)) {
+      return fileName.substring(1);  // FP, fileName cannot be null when isEmpty is false
```
Wohops, line 10: The comments should be edited to remove FP.

java-checks-test-sources/src/main/java/symbolicexecution/behaviorcache/SpringStringUtilsMethods.java
```diff
@@ -0,0 +1,17 @@
+package symbolicexecution.behaviorcache;
+
+import org.springframework.util.StringUtils;
+
+class SpringStringUtilsMethods {
+
+  String test_isEmpty() {
+    String fileName = null;
+    if (!StringUtils.isEmpty(fileName)) {
```
Wohops, line 9: `StringUtils.isEmpty()` seems to be deprecated in latest versions of Spring. They recommend using other methods instead which we should cover as well:
* `org.springframework.util.StringUtils.hasLength(@Nullable String)`
* `org.springframework.util.StringUtils.hasText(@Nullable String)`
* `org.springframework.util.ObjectUtils.isEmpty(@Nullable Object)`

its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json
```diff
@@ -1191,7 +1191,7 @@
     "ruleKey": "S2259",
     "truePositives": 41,
     "falseNegatives": 0,
-    "falsePositives": 129
+    "falsePositives": 130
```
Wohops, line 1194: The new FP comes from the new file that we introduced. In degraded mode, method behaviors can not be recognized (basically, almost all methods invocation are unknown methods), and so the fix can not be taken into account.

This should be handled by the following ticket: [SONARJAVA-3843](https://sonarsource.atlassian.net/browse/SONARJAVA-3843)

Wohops (changes requested):

java-checks-test-sources/src/main/java/symbolicexecution/behaviorcache/SpringStringUtilsMethods.java
chrislain-razafimahefa-sonarsource: All those methods' method behaviors were already implemented. I have just added test cases to cover them.

Wohops:
LGTM! Nice work!

PR #4146, opened 2022-08-03
SONARJAVA-4174 Fix handling of "Optional.ofNullable" when parameter has no known constraints

Wohops:


java-symbolic-execution/src/main/java/org/sonar/java/se/checks/OptionalGetBeforeIsPresentCheck.java
```diff
@@ -209,52 +211,101 @@ public void visitMethodInvocation(MethodInvocationTree tree) {
...
+      SymbolicValue symbolicValue;
+      if(optional instanceof OptionalSymbolicValue) {
+        symbolicValue = ((OptionalSymbolicValue) optional).wrappedValue;
+      } else {
+        symbolicValue = constraintManager.createSymbolicValue(tree);
+      }
+      List<ProgramState> psPresent = optional.setConstraint(pop.state.stackValue(symbolicValue), OptionalConstraint.PRESENT);
+      psEmpty.forEach(context::addTransition);
+      psPresent.forEach(context::addTransition);
+      programState = null;
```
chrislain-razafimahefa-sonarsource, line 277: What about adding the '// interrupt current path to only use transitions' comment before this line?

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/OptionalGetBeforeIsPresentCheck.java
```diff
@@ -209,52 +211,101 @@ public void visitMethodInvocation(MethodInvocationTree tree) {
...
+        context.addExceptionalYield(optionalSV, programState, "java.util.NoSuchElementException", check);
+        reportIssue(tree);
+        // continue exploration after reporting, assuming the optional is now present (killing any noise after the initial issue)
+        programState = programState.addConstraint(optionalSV, OptionalConstraint.PRESENT);
+      }
+    }
+
+    private String useIsEmptyPrefix(String owner) {
+      return java11 ? String.format(" or \"!%sisEmpty()\"", owner) : "";
+    }
```
chrislain-razafimahefa-sonarsource, line 249: Why not move this method just before method 'reportIssue' since it is only used there?

chrislain-razafimahefa-sonarsource (approved):
I just suggested 2 minor changes, but otherwise, it looks very good.

PR #4158, opened 2022-09-27
SONARJAVA-4330 Rule S2722: Update method invokation matching condition

Swalkyn:


alban-auzeill:
In general, it's better in commit messages to summarize "why this change". Because the "what this change is" can be extracted from the commit diff, but the "why" has to be puzzle out, and it is not always obvious. I would propose: `SONARJAVA-4330 FP S2722: Should not raise issues in the presence of method calls throwing NoSuchElementException`

alban-auzeill (approved):

PR #4163, opened 2022-10-10
SONARJAVA-3529 S3958: Fix incorrect issue location in case of exceptional flow.

johann-beleites-sonarsource:


Wohops (approved):
LGTM. Nice fix.

PR #4165, opened 2022-10-10
SONARJAVA-3668 S1186: Ignore methods annotated with org.aspectj.lang.annotation.Pointcut.

johann-beleites-sonarsource:


java-checks-test-sources/src/main/java/checks/EmptyMethodsCheckNoSemantics.java
```diff
@@ -0,0 +1,194 @@
...
+
+    class E {
+      // Compliant - not public
+      E() {
+      }
+    }
+
+    class F {
+      // Compliant - not a no-arg constructor
+      public F(int i) {
```
dorian-burihabwa-sonarsource, line 151: The documentation in RSPEC is inconsistent. This is a non-default constructor because it has an argument, so an issue should be raised or the [exception better documented](https://sonarsource.github.io/rspec/#/rspec/S1186/java).

dorian-burihabwa-sonarsource (changes requested):
LGTM. There is perhaps a leftover import in the test of the check.
We should also update RSpec to mention the following exceptions:
1. No issue raised on non-public default constructors
2. No issue raised on non-default constructors

johann-beleites-sonarsource:
The corresponding RSPEC PR should be merged with this PR: https://github.com/SonarSource/rspec/pull/1337

java-checks-test-sources/src/main/java/checks/EmptyMethodsCheckNoSemantics.java
```diff
@@ -0,0 +1,194 @@
...
+    class D {
+      // Compliant
+      public D() {
+        // usefull comment
+      }
+    }
+
+    class E {
+      // Compliant - not public
+      E() {
```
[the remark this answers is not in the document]
johann-beleites-sonarsource: Nice find! I agree this is inconsistent. I don't think it fits into the scope of this issue/PR, so I've created https://sonarsource.atlassian.net/browse/SONARJAVA-4338, which I'll fix in a separate PR.

java-checks-test-sources/src/main/java/checks/EmptyMethodsCheckNoSemantics.java
johann-beleites-sonarsource: See reply on other comment

dorian-burihabwa-sonarsource (approved):
Thank you for the changes. LGTM 👍🏿

PR #4167, opened 2022-10-12
SONARJAVA-4176 NPE in JSymbol.typeOwner

alban-auzeill:


dorian-burihabwa-sonarsource (approved):
The change looks good to me and is in line with the ticket.
But I am wondering if we should not include a test on source code where `getPackage` returns null. From the documentation, calling owner on an array should go through that path

dorian-burihabwa-sonarsource (approved):
Thanks for adding the extra test. LGTM!

PR #4169, opened 2022-10-12
SONARJAVA-4196 Cover methods `start` and `end` of `java.util.regex.Matcher` in rule S5860

Swalkyn:


its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json
```diff
@@ -2867,4 +2867,4 @@
     "falseNegatives": 32,
     "falsePositives": 0
   }
-]
\ No newline at end of file
+]
```
dorian-burihabwa-sonarsource, line 2870: No need to change the file
```suggestion
]

```

java-checks/src/main/java/org/sonar/java/checks/regex/UnusedGroupNamesCheck.java
```diff
@@ -92,7 +94,7 @@ protected void checkRegex(RegexParseResult regexForLiterals, ExpressionTree meth
 
   private void checkGroupUsage(MethodInvocationTree mit, KnownGroupsCollector knownGroups) {
     String methodName = ExpressionUtils.methodName(mit).name();
-    if ("group".equals(methodName)) {
+    if (Arrays.asList(NAMES_OF_METHODS_WITH_GROUP_ARGUMENT).contains(methodName)) {
```
dorian-burihabwa-sonarsource, line 97: We should duplicate the `NAMES_OF_METHODS_WITH_GROUP_ARGUMENT` as a list in a static field to avoid the unnecessary operation of initializing a new list on every call to the `checkGroupUsage`

dorian-burihabwa-sonarsource (changes requested):
The change looks good but we should 2 things before merging into master.
1. autoscan-diff-rules should not be modified just for the end line
2. the names of methods to check should be in a list, initialized once, even if it means duplicating the information

its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json
Swalkyn: Yes, my mistake!

java-checks/src/main/java/org/sonar/java/checks/regex/UnusedGroupNamesCheck.java
Swalkyn: Good point, it's fixed

dorian-burihabwa-sonarsource (approved):
Thanks for applying the changes. LGTM 👍🏿

PR #4171, opened 2022-10-12
SONARJAVA-4241 S1125: erroneous quick fix suggestion when negating a binary operation

dorian-burihabwa-sonarsource:


johann-beleites-sonarsource (approved):
LGTM!

PR #4176, opened 2022-10-13
Improve autoscan IT that was always failing each time we add a unit test

alban-auzeill:


johann-beleites-sonarsource (approved):
Nice compromise between keeping some info about TPs but not to the degree that causes too much friction! LGTM!

PR #4177, opened 2022-10-14
Export test classpath and binaries via java resource locator

gtoison:
Hello @johann-beleites-sonarsource  ,
This is a follow up to the discussion: https://community.sonarsource.com/t/locating-compiled-class-files/69145/1

Exposing the test classpath and binaries of the project under analysis would help implementing https://github.com/spotbugs/sonar-findbugs/issues/18

Thanks in advance for looking into it!

### DIFF
--- a/java-frontend/src/main/java/org/sonar/plugins/java/api/JavaResourceLocator.java
+++ b/java-frontend/src/main/java/org/sonar/plugins/java/api/JavaResourceLocator.java
@@ -53,29 +53,27 @@ public interface JavaResourceLocator extends JavaFileScanner {
   /**
    * The folders containing the binary .class files.
    * @return a list of folders.
-   * @since 9.7
+   * @since SonarQube 9.8
    */
   Collection<File> binaryDirs();
 
   /**
    * The folders containing the binary .class files corresponding to the tests.
    * @return a list of folders.
-   * @since 9.7
+   * @since SonarQube 9.8
    */
   Collection<File> testBinaryDirs();
 
   /**
    * Classpath configured for the project.
-   * This classpath method is used by the findbugs plugin to configure the analysis.
    * @return the list of jar and class files constituting the classpath of the analyzed project.
    */
   Collection<File> classpath();
 
   /**
    * Classpath configured for the project tests.
-   * This classpath method is used by the findbugs plugin to configure the analysis.
    * @return the list of jar and class files constituting the classpath of the analyzed project.
-   * @since 9.7
+   * @since SonarQube 9.8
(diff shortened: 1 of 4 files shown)


gtoison:
The build failed with what seems to be a compilation error on the test classes but the logs are truncated and I cannot see what might be wrong

java-frontend/src/main/java/org/sonar/plugins/java/api/JavaResourceLocator.java
```diff
@@ -53,14 +53,29 @@ public interface JavaResourceLocator extends JavaFileScanner {
...
   /**
    * Classpath configured for the project.
    * This classpath method is used by the findbugs plugin to configure the analysis.
    * @return the list of jar and class files constituting the classpath of the analyzed project.
    */
   Collection<File> classpath();
 
+  /**
+   * Classpath configured for the project tests.
+   * This classpath method is used by the findbugs plugin to configure the analysis.
```
dorian-burihabwa-sonarsource, line 76: I would avoid explicitly mentioning dependent projects that we don't maintain. I will create a JIRA ticket with a summary mentioning findbugs and we can mention the ticket in the final merge commit.
```suggestion
```

dorian-burihabwa-sonarsource (commented):
I think the change makes sense but we need to:
* Create a ticket to be able to track the change
* Document the introduction in the changelog

java-frontend/src/main/java/org/sonar/plugins/java/api/JavaResourceLocator.java
gtoison: Agreed, I've removed that mention to findbugs as well as the one on `classpath()`
I suppose it dates from the time when that plugin used to be managed by SonarSource

I've also adjusted the `@Since` to the next version of SonarQube, presumably 9.8 ?

For the ticket and the changelog, is that something I should be taking care of?

dorian-burihabwa-sonarsource: No worries, I will take care of the ticket. Thank you for the  quick changes

gtoison:
I think the failed build is not due to the changes in that PR, right?

Wohops:
@gtoison we changed some parts of our CI over the last days/hours. Can you maybe rebase your branch on top of our master? I hope it will fix the build.

dorian-burihabwa-sonarsource:
Closing this PR as the change has been included and documented as part of SONARJAVA-4348 (#4193).

PR #4178, opened 2022-10-14
SONARJAVA-4325 Change message suggestion for S3878 when method argume…

margarita-nedzelska-sonarsource:
…nt type is not Object


alban-auzeill (approved):

alban-auzeill:
This PR fixes the message problem. So you can merge it as it is.
But I see another issue with this rule related to the following false-positive:
```java
  doTrickyThing(new String[]{"hello", "world"}); // false-positive with message {{Remove this array creation and simply pass the elements.}}


  public void doTrickyThing(String[]... args) {
  }
```
I let you decide if you want that I create a jira ticket about this FP and you merge this PR. Or if you want to use this PR to also remove the false-positive.

PR #4182, opened 2022-10-19
SONARJAVA-3896 FP S3329 should not raise when the IV is not defined

alban-auzeill:


java-checks-test-sources/src/main/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -180,23 +181,23 @@ static void decryptImpl4(SecretKeySpec ks) throws Exception {
...
+    new IvParameterSpec("111".getBytes("UTF-8")); // Noncompliant - not used...
     Cipher
       .getInstance(OPERATION_MODE, "BC")
       .init(Cipher.DECRYPT_MODE, ks, iv);
   }
 
-  static void decryptImpl6(byte[] biv, SecretKeySpec ks, IvParameterSpec iv) throws Exception {
-    IvParameterSpec iv2 = new IvParameterSpec(biv); // Noncompliant - not used...
+  static void decryptImpl6(SecretKeySpec ks, IvParameterSpec iv) throws Exception {
+    IvParameterSpec iv2 = new IvParameterSpec("111".getBytes("UTF-8")); // Noncompliant - not used...
```
chrislain-razafimahefa-sonarsource, line 192: Is it of any value to say that it is 'not used'?
In any case, I think we can also get rid of the '...', I don't see the point of having them.

chrislain-razafimahefa-sonarsource (changes requested):
Just a few suggestions that you might want to consider.

chrislain-razafimahefa-sonarsource (approved):
Sounds good. Well done!

PR #4189, opened 2022-10-20
Added some exceptions to rule S107 to exclude more annotated methods and implicit @Autowired constructors

leonardo-pilastri-sonarsource:


java-checks-test-sources/src/main/java/checks/TooManyParametersCheck.java
```diff
@@ -103,3 +119,96 @@ public void foo4(String p1, String p2, String p3, String p4, String p5, String p
...
+  public void trace(String p1, String p2, String p3, String p4, String p5, String p6, String p7, String p8) {} // Compliant
+  
+}
+
+/* 
+ * We can maybe have a discussion on records' constructors, if they should be checked against the rule,
+ * but since they are used to store articulate data, it might not be a good idea
+ * The one note to keep in mind is that it seems like a record could be used as a Spring @Component
+ */
+record Record1(String p1, String p2, String p3, String p4, String p5, String p6, String p7, String p8) {} // Compliant
```
alban-auzeill, line 213: Once we understand this rule, it's better to take a decision about records now and not trigger a discussion for later.
If someone comes back to this rule, it may be in 2 years, perhaps not us, and it will cost some effort to understand the rule again.
The two of us are enough to make a decision.
I vote to have a specific exclusion about records because records are data classes where all the fields are always set through the constructor parameters, it's the only solution. And I have the same point of view for annotations. So I would add those exceptions:
```java
    // This rule has the following exceptions: RECORD, ANNOTATION_TYPE
    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.ENUM);
  }
``` 
And if we make an exception about records, we should update the RSPEC, and just having one compliant test for records is enough, but it should also be with an additional constructor, not only the compact constructor, here it's cheating.
But if we don't exclude records, as a general rule, then it's better to always have two kinds of tests: `Compliant` and `Noncompliant`.
Same for enum, interface, and annotation, if we decide to raise issues on them, we should have `Noncompliant` tests.

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/TooManyParametersCheck.java
```diff
@@ -66,33 +63,67 @@ public class TooManyParametersCheck extends BaseTreeVisitor implements JavaFileS
...
+    "io.micronaut.http.annotation.Put",
+    "io.micronaut.http.annotation.Delete",
+    "io.micronaut.http.annotation.Options",
+    "io.micronaut.http.annotation.Patch",
+    "io.micronaut.http.annotation.Head",
+    "io.micronaut.http.annotation.Trace"
   );
+  
+  //if a class is annotated as one of these types, its constructor should be ignored
+  private static final List<String> CLASS_CONSTRUCTOR_WHITELIST = Arrays.asList(
```
alban-auzeill, line 79: When we modify the existing code, it's a good opportunity to replace `loaded terms` like `blacklist/whitelist`, `master/slave`. So here, I propose to replace `WHITE_LIST` with `METHOD_ANNOTATION_EXCEPTIONS` and `CLASS_CONSTRUCTOR_WHITELIST` with `CLASS_ANNOTATION_IMPLYING_CONSTRUCTOR_EXCEPTIONS` ... or something smaller ...

java-checks/src/main/java/org/sonar/java/checks/TooManyParametersCheck.java
```diff
@@ -106,4 +137,16 @@ private static boolean usesAuthorizedAnnotation(MethodTree method) {
...
   }
 
+  //As of Spring 4.3, classes (@Component, @Service, etc..) with a single constructor can omit the @Autowired annotation.
+  private static boolean classUsesAuthorizedAnnotation(ClassTree methodParentClass) {
+    SymbolMetadata parentClassMetadata = methodParentClass.symbol().metadata();
+    //if the parent class is a Spring component or has unknown annotations
+    if(hasUnknownAnnotation(parentClassMetadata) || CLASS_CONSTRUCTOR_WHITELIST.stream().anyMatch(parentClassMetadata::isAnnotatedWith)) {  
+      long numberOfConstructors = methodParentClass.members().stream().filter(member -> member.is(Tree.Kind.CONSTRUCTOR)).count();    
+      //if it only has 1 constructor, @Autowired could be implicit, and it's an exception to the rule
+      if(numberOfConstructors == 1) return true;
```
alban-auzeill, line 147: It seems you don't have the proper formatting settings in your IDE, or you don't format on `save`.
I was expecting comments with a leading space `// if` and `if` with `{ }`

alban-auzeill (changes requested):

alban-auzeill (changes requested):

java-checks-test-sources/src/main/java/checks/TooManyParametersCheck.java
leonardo-pilastri-sonarsource: Yes I agree that we should ignore them as exceptions to the rule. I will add compliant tests for both records and annotations

java-checks/src/main/java/org/sonar/java/checks/TooManyParametersCheck.java
leonardo-pilastri-sonarsource: No I don't have any automatic formatting on ctrl+s, I will see if I can get to set that up

java-checks-test-sources/src/main/java/checks/TooManyParametersCheck.java
leonardo-pilastri-sonarsource: I think annotations cannot really have parameters in their methods, nor they can have a constructor. Am I right? If that is the case I can skip tests for them I think

java-checks/src/main/java/org/sonar/java/checks/TooManyParametersCheck.java
```diff
@@ -66,33 +63,60 @@ public class TooManyParametersCheck extends BaseTreeVisitor implements JavaFileS
...
 
   @Override
-  public void visitMethod(MethodTree tree) {
-    super.visitMethod(tree);
-    if (isOverriding(tree) || usesAuthorizedAnnotation(tree)) {
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+
+    // this is true only if the class has unknown annotations, or if it has a single constructor and
+    // it is annotated with one of CLASS_CONSTRUCTOR_WHITELIST
```
chrislain-razafimahefa-sonarsource, line 95: `CLASS_ANNOTATION_CONSTRUCTOR_EXCEPTIONS` instead of `CLASS_CONSTRUCTOR_WHITELIST`

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S107.html
```diff
@@ -21,6 +21,11 @@ <h2>Exceptions</h2>
   <li> Bean constructor injection with <code>@org.springframework.beans.factory.annotation.Autowired</code> </li>
   <li> CDI constructor injection with <code>@javax.inject.Inject</code> </li>
   <li> <code>@com.fasterxml.jackson.annotation.JsonCreator</code> </li>
+  <li> <code>@org.springframework.context.annotation.Bean</code> </li>
+  <li> Micronaut’s annotations like <code>@io.micronaut.http.annotation.Get</code> </li>
+  <li> Single constructors of classes annotated as Spring components ( <code>@org.springframework.stereotype.Component</code> ) </li>
```
chrislain-razafimahefa-sonarsource, line 26: Don't forget to update this with the latest from RSPEC.

chrislain-razafimahefa-sonarsource (changes requested):
Almost there.

chrislain-razafimahefa-sonarsource (approved):
Looks good! Well done!

java-checks/src/main/java/org/sonar/java/checks/TooManyParametersCheck.java
leonardo-pilastri-sonarsource: I updated the names of those static final lists accordingly

PR #4190, opened 2022-10-21
SONARJAVA-4335 S3776 should Ignore equals() and hashCode() methods similarly to S1541

chrislain-razafimahefa-sonarsource:
Fix S3776 and update S1541.
Also update the rule description (https://github.com/SonarSource/rspec/pull/1361).

### DIFF
--- a/its/ruling/src/test/resources/commons-beanutils/java-S3776.json
+++ b/its/ruling/src/test/resources/commons-beanutils/java-S3776.json
@@ -54,9 +54,6 @@
 1743,
 2095,
 ],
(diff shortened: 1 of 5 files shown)


leonardo-pilastri-sonarsource (commented):
Since we want to use these two MethodMatchers in both classes ( MethodComplexityCheck and CognitiveComplexityCheck ) and possibly in the future more, could it be a good idea to centralize these matchers somewhere else?

leonardo-pilastri-sonarsource (approved):

leonardo-pilastri-sonarsource (approved):

PR #4191, opened 2022-10-21
SONARJAVA-4282 Exclude "com.sun.xml.ws" package from S1191 by default

dorian-burihabwa-sonarsource:


chrislain-razafimahefa-sonarsource (approved):
Good job!

PR #4192, opened 2022-10-21
SONARJAVA-4251  FIX FN S2252(ForLoopFalseConditionCheck) does not support constants

margarita-nedzelska-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4194, opened 2022-10-24
SONARJAVA-4348 Export test classpath and binaries in the API

dorian-burihabwa-sonarsource:
Originally proposed in #4177 by @gtoison


johann-beleites-sonarsource (approved):
LGTM!

PR #4203, opened 2022-10-26
SONARJAVA-4342 Nullness annotation on interface methods should be inherited in implementation methods

chrislain-razafimahefa-sonarsource:


Wohops (approved):
LGTM! Nice work!

PR #4213, opened 2022-11-09
SONARJAVA-4268 Rule S5542: Detect CBC mode when used with padding

irina-batinic-sonarsource:


java-checks-test-sources/src/main/java/checks/security/EncryptionAlgorithmCheck.java
```diff
@@ -13,41 +13,41 @@ abstract class EncryptionAlgorithmCheck {
...
     try {
       // First case
-      Cipher.getInstance("AES"); // Noncompliant [[sc=26;ec=31]] {{Use secure mode and padding scheme.}}
+      Cipher.getInstance("AES"); // Noncompliant [[sc=26;ec=31]] {{Use a secure padding scheme.}}
 
-      Cipher.getInstance("AES/ECB/NoPadding"); // Noncompliant
-      Cipher.getInstance("AES" + "/ECB/NoPadding"); // Noncompliant
-      Cipher.getInstance("AES/ECB/NoPadding", getProvider()); // Noncompliant
-      Cipher.getInstance("AES/ECB/NoPadding", "someProvider"); // Noncompliant
+      Cipher.getInstance("AES/ECB/NoPadding"); // Noncompliant {{Use a secure cipher mode.}}
```
margarita-nedzelska-sonarsource, line 26: It's not mandatory to check the message every time. You can test put just `// Noncompliant`

java-checks/src/main/java/org/sonar/java/checks/security/EncryptionAlgorithmCheck.java
```diff
@@ -88,15 +89,20 @@ private static boolean isInsecureAlgorithm(String algorithmName) {
       boolean isRSA = "RSA".equalsIgnoreCase(algorithm);
 
       if ("ECB".equalsIgnoreCase(mode) && !isRSA) {
-        return true;
+        return Optional.of("Use a secure cipher mode.");
       }
-      if ("CBC".equalsIgnoreCase(mode)) {
-        return false;
+
+      if ("CBC".equalsIgnoreCase(mode) && !padding.equalsIgnoreCase("NoPadding")) {
```
margarita-nedzelska-sonarsource, line 95: It's better to be consistent. A string literal should always be on the left or on the right side. Actually, our activated rule insists on the left, so let's keep it that way.

margarita-nedzelska-sonarsource (commented):
Nice work!
Just 2 Minor comments

PR #4214, opened 2022-11-09
Remove usages of java.util.Arrays in favor of Java 9+ collections methods

Wohops:


Wohops:
Discarded. Collections returned by List.of() and so on does not have the same behavior.

PR #4215, opened 2022-11-11
Remove unused .github/workflows/nightly.yml

Godin:
CirrusCI provides cron for quite some time already and AFAICS cron is configured in CirrusCI for this repository 😉 

Moreover as you can see in https://github.com/SonarSource/sonar-java/actions/runs/3441794235/jobs/5741716852#step:4:11 this GitHub workflow actually fails with the message

> {"errors":[{"message":"Variable 'input' has an invalid value: Field 'repositoryId' has coerced Null value for NonNull type 'ID!'","locations":[{"line":1,"column":36}],"extensions":{"classification":"ValidationError"}}],"data":null}

even if after it sends to the Slack message 

> Nightly build successful

🤣


Wohops (approved):
LGTM

PR #4218, opened 2022-11-29
SONARJAVA-4368 Support of Java 19

dorian-burihabwa-sonarsource:


java-frontend/src/main/java/org/sonar/java/model/pattern/GuardedPatternTreeImpl.java
```diff
@@ -57,7 +57,7 @@ public PatternTree pattern() {
   }
 
   @Override
-  public SyntaxToken andOperator() {
+  public SyntaxToken whenOperator() {
```
Wohops, line 60: The corresponding field should be renamed as well (as well as the constructor parameter)

java-frontend/src/main/java/org/sonar/java/model/expression/InstanceOfTreeImpl.java
```diff
@@ -42,21 +43,28 @@ public class InstanceOfTreeImpl extends AssessableExpressionTree implements Inst
   private final TypeTree type;
   @Nullable
   private final VariableTree variable;
+  @Nullable
+  private final PatternTree pattern;
```
Wohops, line 47: I'm not sure we need to keep the `VariableTree` here. What about keeping only the pattern?

For me, the case `variable != null`  corresponds to the situation where the use is a  `TypePattern`.

Dropping the variable would prevent the introduction of yet another constructor.

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/PatternInstanceOfTree.java
```diff
@@ -39,6 +39,13 @@ public interface PatternInstanceOfTree extends ExpressionTree {
 
   SyntaxToken instanceofKeyword();
 
+  @Deprecated
+  /**
+   * @deprecated Use {@link PatternInstanceOfTree#pattern()}
+   * @since 7.16
+   */
```
Wohops, line 46: I would place the `@Deprecated` annotation after the javadoc (convention).
Also, I would add the `@Nullable` annotation, since when the pattern won't always be TypePattern, you might not always have variables inside.

java-frontend/src/main/java/org/sonar/java/model/expression/InstanceOfTreeImpl.java
```diff
@@ -90,6 +98,11 @@ public VariableTree variable() {
     return variable;
```
Wohops, line 98: For me, this should rather be something like:
```java
return pattern != null && pattern.is(Tree.Kind.TYPE_PATTERN) ? ((TypePatternTree) pattern).variable() : null;
```

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/RecordPatternTree.java
```diff
@@ -0,0 +1,29 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.plugins.java.api.tree;
+
+import java.util.List;
+
+public interface RecordPatternTree extends PatternTree {
+  TypeTree type();
+  SyntaxToken openingParenthesis();
```
Wohops, line 26: Not sure about the parentheses being needed here... Maybe for completeness.
However, if you want to keep them, what about using an `org.sonar.plugins.java.api.tree.ListTree<PatternTree>` instead? This would also encode the separators.

java-frontend/src/test/java/org/sonar/java/model/JavaVersionImplTest.java
```diff
@@ -85,9 +85,9 @@ void compatibilityMesssages() {
 
   @Test
   void test_effective_java_version() {
-    assertThat(new JavaVersionImpl().effectiveJavaVersionAsString()).isEqualTo("18");
+    assertThat(new JavaVersionImpl().effectiveJavaVersionAsString()).isEqualTo("19");
```
Wohops, line 88: There is a parametrized tests a few lines above, which test all the versions we support. What about adding `19` as well?

java-frontend/src/test/java/org/sonar/java/model/expression/InstanceOfTreeImplTest.java
```diff
@@ -78,23 +78,6 @@ void test_GuardedPatternInstanceOfTree() {
 
   }
 
-  @Test
```
Wohops, line 81: I'm ok with dropping this one, but what about keeping one with parentheses?
```java
ExpressionTree condition = ifCondition("o instanceof (String s) && s.length() > 10");
```

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/Tree.java
```diff
@@ -741,6 +741,14 @@ enum Kind implements GrammarRuleKey {
...
     DEFAULT_PATTERN(DefaultPatternTree.class),
 
+    /**
+     * {@link RecordPatternTree}
+     *
+     * @since Java 19
+     * @deprecated Preview feature
+     */
+    @Deprecated(since = "7.16", forRemoval = false)
+    RECORD_PATTERN(RecordPatternTree.class),
```
Wohops, line 751: Since you added a new tree, you need to also update the `baseTreeVisitor` to give an access to it.
Also, don't forget to modify `visitPatternInstanceOf` to replace the use of `scan(tree.variable());` by `scan(tree.pattern());`

Wohops (commented):
That's a good start! Good job!

java-frontend/src/test/java/org/sonar/java/model/expression/InstanceOfTreeImplTest.java
dorian-burihabwa-sonarsource: Parenthesis around `String s`

java-frontend/src/main/java/org/sonar/java/model/JParser.java
```diff
@@ -1433,11 +1434,24 @@ private PatternTree convertPattern(Pattern p) {
...
+        RecordPattern recordPattern = (RecordPattern) p;
+        List<PatternTree> nestedPatterns = recordPattern.patterns().stream()
+          .map(this::convertPattern)
+          .collect(Collectors.toList());
+
+        SimpleName patternName = recordPattern.getPatternName();
+        return new RecordPatternTreeImpl(
+          convertType(recordPattern.getPatternType()),
+          nestedPatterns,
+          convertSimpleName(patternName)
```
Wohops, line 1447: I would inline the `patternName` here.

java-frontend/src/main/java/org/sonar/java/model/pattern/RecordPatternTreeImpl.java
```diff
@@ -0,0 +1,81 @@
...
+import org.sonar.plugins.java.api.tree.TreeVisitor;
+import org.sonar.plugins.java.api.tree.TypeTree;
+import org.sonarsource.analyzer.commons.collections.ListUtils;
+
+public class RecordPatternTreeImpl extends AbstractPatternTree implements RecordPatternTree {
+
+  private final TypeTree type;
+  private final List<PatternTree> patterns;
+
+  private final IdentifierTree name;
```
Wohops, line 38: I would had the `@Nullable` annotation also here.

java-frontend/src/main/java/org/sonar/java/model/pattern/RecordPatternTreeImpl.java
```diff
@@ -0,0 +1,81 @@
...
+  public TypeTree type() {
+    return type;
+  }
+
+  @Override
+  public List<PatternTree> patterns() {
+    return patterns;
+  }
+
+  @Nullable
```
Wohops, line 67: I think I would rather use `@CheckForNull`, I hardly see how we could predict cases where it's going to be null or not.

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/BaseTreeVisitor.java
```diff
@@ -465,4 +465,10 @@ public void visitGuardedPattern(GuardedPatternTree tree) {
     scan(tree.pattern());
     scan(tree.expression());
   }
+
+  @Override
+  public void visitRecordPattern(RecordPatternTree tree) {
+    scan(tree.type());
+    scan(tree.patterns());
```
Wohops, line 472: The name is missing here. Normally it can deal with nullness.
```
scan(tree.name());
```

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/RecordPatternTree.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.List;
+import javax.annotation.Nullable;
+
+public interface RecordPatternTree extends PatternTree {
+  TypeTree type();
+
+  List<PatternTree> patterns();
+
+  @Nullable
```
Wohops, line 30: I would prefer `@CheckForNull` here. We clearly will have to check systematically.

java-frontend/src/main/java/org/sonar/plugins/java/api/tree/PatternInstanceOfTree.java
```diff
@@ -39,6 +40,15 @@ public interface PatternInstanceOfTree extends ExpressionTree {
 
   SyntaxToken instanceofKeyword();
 
+  /**
+   * @since 7.16
+   * @deprecated Use {@link PatternInstanceOfTree#pattern()}
+   * @return null for all patterns that are not {@link TypePatternTree}.
+   */
+  @Deprecated
```
Wohops, line 48: What about marking it for removal?
```
@Deprecated(since = "7.16", forRemoval = true)
```

Wohops (commented):
Almost there!

chrislain-razafimahefa-sonarsource (changes requested):
Just one English issue.

dorian-burihabwa-sonarsource (commented):
Reviewed everything but the symbolic execution code

sonar-java-plugin/pom.xml
```diff
@@ -143,7 +143,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>18000000</maxsize>
+                  <maxsize>18500000</maxsize>
```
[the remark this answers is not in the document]
Wohops: Yep. Unfortunately, the new libraries are bigger than expected.

java-symbolic-execution/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -960,10 +959,14 @@ private SymbolicValue executeVariableSpecialParent(VariableTree variableTree, Sy
...
-    } else if (parent.is(Tree.Kind.PATTERN_INSTANCE_OF)) {
-      // The new variable created from a pattern instance of share the same symbolic value as the current object.
+    } else if (parent.is(Tree.Kind.TYPE_PATTERN) && parent.parent().is(Tree.Kind.PATTERN_INSTANCE_OF)) {
+      // FIXME only works with the TYPE_PATTERN inside an pattern instanceof
+      // handling of all patterns as introduced in JDK17+ should also be supported
+      // skip the unused SV corresponding to the type pattern
+      programState = programState.unstackValue(1).state;
       SymbolicValue peekValue = programState.peekValue();
       if (peekValue != null) {
+        // The new variable created from a pattern instance of share the same symbolic value as the current object.
```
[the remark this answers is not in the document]
Wohops: fixed

java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java
```diff
@@ -47,6 +46,7 @@
 import org.sonar.plugins.java.api.tree.SyntaxToken;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.VariableTree;
+import org.sonarsource.analyzer.commons.collections.ListUtils;
```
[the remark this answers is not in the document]
Wohops: nop it is used. (Line 227)

sonar-java-plugin/src/main/resources/static/documentation.md
```diff
@@ -139,6 +139,17 @@ The tutorial [Writing Custom Java Rules 101](https://redirect.sonarsource.com/do
...
+* New method: `RecordPatternTree#type()`. Use this method to get the reference type in the record pattern (Preview feature in Java 19).
+* New method: `RecordPatternTree#patterns()`. Use this method to get the patterns nested in the record pattern (Preview feature in Java 19).
+* New method: `RecordPatternTree#name()`. Use this method get the optional record pattern identifier (Preview feature in Java 19).
+* Dropped method: `GuardedPatternTree#andOperator()`. Use `GuardedPatternTree#whenOperator()` instead.
+* New method: `GuardedPatternTree#whenOperator()`. Use this method to get the `when` operator prefixing the guard (Preview feature in Java 19).
+* Deprecated method: `PatternInstanceOfTree#variable()`. Use `PatternInstanceOfTree#pattern` instead.
+* New method: `PatternInstanceOfTree#pattern()`. Use this method to get the pattern in an `instanceof` expression. When the pattern is a `TypePatternTree`, the variable can then be extracted using `TypePatternTree#patternVariable`.
+* New method: `TreeVisitor#visitRecordPattern()`. Use this method to traverse a `RecordPatternTree`.
```
chrislain-razafimahefa-sonarsource, line 151:
```suggestion
* New method: `TreeVisitor#visitRecordPattern()`. Use this method to traverse a `RecordPatternTree` (Preview feature in Java 19).
```

dorian-burihabwa-sonarsource:
The PR is now closed because the partial support of Java 19 introduces bugs that make the upgrade too unstable.

PR #4220, opened 2022-12-04
SONARJAVA-4414 (S5413): SuspiciousListRemove: control flow with "return" is compliant

julianladisch:
Leaving a loop with `return` is the same as leaving the loop with `break`.


johann-beleites-sonarsource:
Tracked by [SONARJAVA-4414](https://sonarsource.atlassian.net/browse/SONARJAVA-4414)

erwan-serandour-sonarsource:
Merged to master.

PR #4222, opened 2022-12-05
SONARJAVA-4353 Fixed FP on switch covering all cases.

leonardo-pilastri-sonarsource:


Wohops (changes requested):
For me the fix is not enough to fix all problems introduced by expressions. For instance, the following switch will also raise a FP, while it is compliant:

```java
    switch (myEnum) { // Compliant
      case A:
        break;
      case B, default:
        break;
    }

    switch (myEnum) { // Noncompliant, but the message should be checked
      case A, default:
        break;
      case B:
        break;
    }
```

java-checks/src/main/java/org/sonar/java/checks/SwitchLastCaseIsDefaultCheck.java
```diff
@@ -57,7 +58,14 @@ private static boolean getDefaultLabel(SwitchStatementTree switchStatementTree)
...
   private static boolean isDefault(CaseLabelTree caseLabelTree) {
-    return JavaKeyword.DEFAULT.getValue().equals(caseLabelTree.caseOrDefaultKeyword().text());
+    if (equalsDefaultKeyword(caseLabelTree.caseOrDefaultKeyword().text())) {
+      return true;
+    }
+    return caseLabelTree.expressions().stream().anyMatch(expr -> equalsDefaultKeyword(expr.firstToken().text()));
+  }
+
+  private static boolean equalsDefaultKeyword(String text) {
+    return JavaKeyword.DEFAULT.getValue().equals(text);
```
Wohops, line 68: What about extracting `JavaKeyword.DEFAULT.getValue()` in a dedicated constant, so it is not recalculated all the time?

java-checks/src/main/java/org/sonar/java/checks/SwitchLastCaseIsDefaultCheck.java
```diff
@@ -57,7 +58,14 @@ private static boolean getDefaultLabel(SwitchStatementTree switchStatementTree)
   }
 
   private static boolean isDefault(CaseLabelTree caseLabelTree) {
-    return JavaKeyword.DEFAULT.getValue().equals(caseLabelTree.caseOrDefaultKeyword().text());
+    if (equalsDefaultKeyword(caseLabelTree.caseOrDefaultKeyword().text())) {
+      return true;
+    }
+    return caseLabelTree.expressions().stream().anyMatch(expr -> equalsDefaultKeyword(expr.firstToken().text()));
```
Wohops, line 64: To be honest, I preferred the comparison with the `DEFAULT_PATTERN` tree kind.
It is indeed deprecated, but only because this is a preview feature of the JDK not yet officialized. In the meantime, we still want to be able to parse this code, so it's valid. We will remove the deprecation as soon as the feature is final, and so the issue will disappear. I'm OK living with the issue.

Wohops (changes requested):

Wohops (approved):
LGTM!

PR #4226, opened 2022-12-07
SONARJAVA-4255 FP S1185(MethodOnlyCallsSuperCheck) with different modifiers

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/MethodOnlyCallsSuperCheck.java
```diff
@@ -67,14 +67,16 @@ public void visitNode(Tree tree) {
...
       reportIssue(methodTree.simpleName(), "Remove this method to simply inherit it.");
     }
   }
 
-  private static boolean isFinal(MethodTree methodTree) {
-    return ModifiersUtils.hasModifier(methodTree.modifiers(), Modifier.FINAL);
+  private static boolean isFinalOrSynchronizedOrStrictFP(MethodTree methodTree) {
+    return methodTree.modifiers().modifiers().stream()
+      .map(ModifierKeywordTree::modifier)
+      .anyMatch(m -> m == Modifier.FINAL || m == Modifier.SYNCHRONIZED || m == Modifier.STRICTFP);
```
leonardo-pilastri-sonarsource, line 79: Since there was a util method call here, I think it would be nice if your extended logic was included in the `ModifiersUtils` class, for others to use in future!
So here you could still call the utils class like this:
`return ModifiersUtils.hasAnyModifier(methodTree.modifiers(), Modifier.FINAL, Modifier.SYNCHRONIZED , Modifier.STRICTFP);`

And an implementation in `ModifiersUtils` might be:

`public static boolean hasAnyModifier(ModifiersTree modifiers, Modifier... expectedModifiers) {
    return Stream.of(expectedModifiers).anyMatch(expectedModifier -> hasModifier(modifiers, expectedModifier));
  }`


(Sorry for the bad review format and style, I don't know how to do it :( )

leonardo-pilastri-sonarsource (changes requested):
I think we might want to include the multiple modifier check in the util class

leonardo-pilastri-sonarsource (approved):
Nice job!

PR #4227, opened 2022-12-08
SONARJAVA-4365 S5786 should not report issues on classes defining publicly visible constants

alban-auzeill:


leonardo-pilastri-sonarsource (approved):
LGTM

PR #4243, opened 2022-12-14
SONARJAVA-4363 Update S2272: Check if methods called by `next()` can throw NoSuchElementException

Swalkyn:
With this change, method bodies are recursively visited to find a throw statement for `NoSuchElementException`. We make sure not to visit the same method twice and to limit to the total number of method visited.

Additionally, exceptions that are subtypes of `NoSuchElementException` are now also taken into account.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/plugins/java/api/tree/MethodInvocationTree.java
+++ b/java-frontend/src/main/java/org/sonar/plugins/java/api/tree/MethodInvocationTree.java
@@ -49,6 +49,6 @@ public interface MethodInvocationTree extends ExpressionTree {
 
   Arguments arguments();
 
-  Symbol symbol();
+  Symbol.MethodSymbol symbol();
 
 }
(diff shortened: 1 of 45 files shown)


java-frontend/src/main/java/org/sonar/plugins/java/api/tree/MethodInvocationTree.java
```diff
@@ -49,6 +49,6 @@ public interface MethodInvocationTree extends ExpressionTree {
 
   Arguments arguments();
 
-  Symbol symbol();
+  Symbol.MethodSymbol symbol();
```
alban-auzeill, line 52: This change generates compilation warnings about useless casts to MethodSymbol.
Like in `java-checks/src/main/java/org/sonar/java/checks/aws/AwsConsumerBuilderUsageCheck.java`:
```java
        MethodSymbol methodSymbol = (MethodSymbol) mit.symbol();
```
We should remove those casts.
And to be able to remove the maximum of useless casts, we could also change the signature of NewClassTree:
```java
Symbol.MethodSymbol constructorSymbol();
```

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #4249, opened 2022-12-15
SONARJAVA-4371 Fix memory leaks in SE rules S2189, S2259, S2637 & S3516

Wohops:


chrislain-razafimahefa-sonarsource (approved):
Well done! LGTM.

PR #4250, opened 2022-12-16
SONARJAVA-4127 Fix handling of anonymous exceptions in computation of method behaviors

Wohops:


java-symbolic-execution/src/test/java/org/sonar/java/se/xproc/ExceptionalYieldTest.java
```diff
@@ -138,11 +134,37 @@ void exceptional_yields_void_method() {
...
+      new NullDereferenceCheck());
+
+    MethodBehavior mb = getMethodBehavior(sevAndSemantic.a, "foo");
+    List<MethodYield> yields = mb.yields();
+    assertThat(yields).hasSize(3);
+
+    Sema semanticModel = sevAndSemantic.b;
+    List<ExceptionalYield> exceptionalYields = mb.exceptionalPathYields().collect(Collectors.toList());
+    assertThat(exceptionalYields).hasSize(2);
+    assertThat(exceptionalYields.stream().filter(y -> y.exceptionType(semanticModel).isUnknown()).count()).isEqualTo(1);
```
chrislain-razafimahefa-sonarsource, line 157: I wonder if we really need to test all these here.
I am thinking that the priority is to test that the anonymous `RuntimeException` is handled well. I.e., that it is returned in the `exceptionalPathYields` list and that its `parameter constraint` is correct.

I agree it is a bit of nitpicking so I would not mind if you leave it as is.

chrislain-razafimahefa-sonarsource (approved):
Very nice! I approved it but I have a suggestion you might want to consider.

PR #4252, opened 2022-12-20
SONARJAVA-4347 Added exception for S1144 on methods with annotated params

leonardo-pilastri-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -165,7 +168,14 @@ private static boolean isUnusedPrivate(Symbol symbol) {
   }
 
   private static boolean hasNoAnnotation(MethodTree methodTree) {
-    return methodTree.modifiers().annotations().isEmpty();
+    return methodTree.modifiers().annotations().isEmpty() && methodTree.parameters().stream().allMatch(UnusedPrivateMethodCheck::hasNoAnnotation);
```
Wohops, line 171: Okay I think this is not enough here if we want to avoid FPs in case of misconfigured projects. I would prefer that we try to avoid having a new rule generating potential FP, so we would privilege FN. It means that we need to be able to react correctly in absence of semantic... and so we have to rely on the tokens only.

Here is the fix I foresee instead:

```java
  private static boolean hasNoAnnotation(MethodTree methodTree) {
    return methodTree.modifiers().annotations().isEmpty()
      && methodTree.parameters().stream().noneMatch(UnusedPrivateMethodCheck::hasNoAllowedAnnotation);
  }

  private static boolean hasNoAllowedAnnotation(VariableTree variableTree) {
    List<AnnotationTree> annotations = variableTree.modifiers().annotations();
    return !annotations.isEmpty()
      && annotations.stream().anyMatch(UnusedPrivateMethodCheck::isAllowedAnnnotation);
  }

  private static boolean isAllowedAnnnotation(AnnotationTree annotation) {
    Type annotationSymbolType = annotation.symbolType();
    if (annotationSymbolType.is(PARAM_ANNOTATION_EXCEPTION)) {
      // semantic-based
      return true;
    }
    if (annotationSymbolType.isUnknown()) {
      // absence of semantic - rely on tokens to build the name
      TypeTree annotationType = annotation.annotationType();
      if (annotationType.is(Tree.Kind.IDENTIFIER)) {
        // might lead to FN in case of wrongly configured project - acceptable
        return "Observes".equals(((IdentifierTree) annotationType).name());
      }
      if (annotationType.is(Tree.Kind.MEMBER_SELECT)) {
        return PARAM_ANNOTATION_EXCEPTION.equals(ExpressionsHelper.concatenate((MemberSelectExpressionTree) annotationType));
      }
    }
    return false;
  }
```

java-checks-test-sources/src/main/java/checks/UnusedPrivateMethod.java
```diff
@@ -11,10 +11,15 @@
...
 import java.util.stream.IntStream;
+import javax.enterprise.event.Observes;
 
 import static java.util.Collections.emptySet;
 
 class UnusedPrivateMethodCheck {
+  
+  private void initNc(@AnotherAnnotation Object object) {} // Noncompliant
+  private void init(@Observes Object object) {} //Compliant, javax.enterprise.event.Observes is an exception to the rule
+  private void init(@Observes Object object, String test) {}
```
Wohops, line 22: I would add one with the fully qualified annotation and one with a custom annotation: 
```java
private void foo(@javax.enterprise.event.Observes Object object) {} // Compliant

// custom annotation
@interface @Observes {}

private void bar(@checks.UnusedPrivateMethodCheck.Observes Object object) {} // Noncompliant - will be a FN in absence of semantic
```

its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json
```diff
@@ -267,7 +267,7 @@
     "ruleKey": "S1144",
     "hasTruePositives": true,
     "falseNegatives": 1,
-    "falsePositives": 0
+    "falsePositives": 2
```
Wohops, line 270: let's try to avoid this! see below!

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1144.html
```diff
@@ -32,5 +32,9 @@ <h2>Compliant Solution</h2>
 }
 </pre>
 <h2>Exceptions</h2>
-<p>This rule doesn’t raise any issue on annotated methods.</p>
+<p>This rule doesn’t raise issues for:</p>
```
Wohops, line 35: :+1:

Wohops (changes requested):
I think we can avoid introducing FPs here.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -90,22 +96,15 @@ private void reportUnusedPrivateMethods() {
...
-        break;
-      case METHOD_INVOCATION:
-        checkIfUnknown((MethodInvocationTree) tree);
-        break;
-      case METHOD_REFERENCE:
-        checkIfUnknown((MethodReferenceTree) tree);
-        break;
-      default:
-        throw new IllegalStateException("Unexpected subscribed tree.");
+    Kind kind = tree.kind();
```
Wohops, line 99: to be honest I feel it was much nicer and easier to read before... I would not challenge this and be OK with the line not being covered. Can you revert the change?
You can also decide to change the last default case to not throw and just do nothing instead... will still remove one line of code.

```java
     default:
        // do nothing
    }

File: java-checks/src/test/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheckTest.java [hunk 6]
```diff
@@ -40,7 +40,15 @@ void test_non_compiling() {
...
       .onFile(TestUtils.nonCompilingTestSourcesPath("checks/UnusedPrivateMethod.java"))
       .withCheck(new UnusedPrivateMethodCheck())
-      .verifyNoIssues();
+      .verifyIssues();
+  }
+  
+  @Test
+  void test_non_compiling_unknown() {
+    CheckVerifier.newVerifier()
+      .onFile(TestUtils.nonCompilingTestSourcesPath("checks/unused/UnusedPrivateMethod.java"))
```
Wohops, line 49: Can you rename the file so it does not have the same as the other before? For instance
`UnusedPrivateMethodWithUnknownResolution.java`

.gitattributes
```diff
@@ -5,3 +5,4 @@ java-checks/src/main/java/org/sonar/java/checks/helpers/ExpressionEvaluator.java
 java-checks-test-sources/src/main/java/checks/CounterModeIVShouldNotBeReusedCheck.java eol=lf
 java-checks/src/main/java/org/sonar/java/checks/helpers/HardcodedStringExpressionChecker.java eol=lf
 its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json eol=lf
+java-checks-test-sources/src/main/files/non-compiling/checks/unused/UnusedPrivateMethod.java eol=lf
```
Wohops, line 8: Really not sure about this... Did you configure your git correctly to handle cr-lf appropriately?

https://github.com/SonarSource/sonar-developer-toolset#git

Wohops (changes requested):

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -105,7 +111,6 @@ public void visitNode(Tree tree) {
         checkIfUnknown((MethodReferenceTree) tree);
         break;
       default:
-        throw new IllegalStateException("Unexpected subscribed tree.");
```
Wohops, line 108: I would add a comment here if you remove the exception, such as `// not registered - can not happen`

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -43,7 +45,9 @@
 import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.ParameterizedTypeTree;
 import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.Tree.Kind;
```
Wohops, line 48: This is an unused import (see SQ analysis). Note that you can configure eclipse to remove them.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -165,7 +170,29 @@ private static boolean isUnusedPrivate(Symbol symbol) {
...
+    Type annotationSymbolType = annotation.symbolType();
+    if (annotationSymbolType.is(PARAM_ANNOTATION_EXCEPTION)) {
+      return true;
+    }
+    if (annotationSymbolType.isUnknown()) {
+      TypeTree annotationType = annotation.annotationType();
+      if (annotationType.is(Tree.Kind.IDENTIFIER)) {
+        return "Observes".equals(((IdentifierTree) annotationType).name());
+      }
+      if (annotationType.is(Tree.Kind.MEMBER_SELECT)) {
```
Wohops, line 191: Not sure it is possible to cover the `else` here, so OK if not covered.

java-checks-test-sources/src/main/files/non-compiling/checks/unused/UnusedPrivateMethod.java
```diff
@@ -0,0 +1,8 @@
+package checks.unused;
```
Wohops, line 1: I'm really not sure you are using this file at the end. Can you remove it?

Wohops (changes requested):

Wohops (approved):
LGTM!

PR #4254, opened 2022-12-20
SONARJAVA-4372 FP in rule S6204 when Collections.shuffle() is used as a mutator

chrislain-razafimahefa-sonarsource:


Swalkyn (approved):
LGTM

PR #4256, opened 2022-12-22
SONARJAVA-4279 S1612 message should not be "!unknown!::new" for constructor with incomplete semantics

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/ReplaceLambdaByMethodRefCheck.java
```diff
@@ -262,13 +262,10 @@ private static Optional<String> getMethodInvocationOrNewClass(@Nullable Tree tre
   private static Optional<String> getNewClass(NewClassTree newClassTree, List<VariableTree> parameters) {
     if (newClassTree.classBody() == null && matchingParameters(parameters, newClassTree.arguments())) {
       TypeTree identifier = newClassTree.identifier();
-      String className;
-      if (identifier.is(Tree.Kind.MEMBER_SELECT)) {
-        className = ExpressionsHelper.concatenate((MemberSelectExpressionTree) identifier);
-      } else {
-        className = identifier.symbolType().name();
+      if (identifier.is(Tree.Kind.MEMBER_SELECT, Tree.Kind.IDENTIFIER)) {
```
Wohops, line 265: I can cover this... and this is a False Negative. Not sure if you want to fix it or create a ticket.
```java
  void foo() {
    Stream.of(Integer.valueOf(42)).forEach(x -> new java.util.ArrayList<String>(x)); // FN
    Stream.of(Integer.valueOf(42)).forEach(java.util.ArrayList<String>::new); // Compile!
  }
```

Wohops (changes requested):

Wohops (approved):

PR #4258, opened 2022-12-22
SONARJAVA-4370 Memory leak in rule S5852 RedosCheck because regexCrea…

irina-batinic-sonarsource:


margarita-nedzelska-sonarsource (approved):
LGTM!

PR #4260, opened 2022-12-22
SONARJAVA-4363 Fix FNs on S2272 when NoSuchElementException is thrown indirectly

Swalkyn:
Method calls are visited recursively to check if they might throw NoSuchElementException.
Methods outside the current file and try-catch statement are not handled.


irina-batinic-sonarsource (approved):
Good job!

PR #4262, opened 2022-12-27
SONARJAVA-4331 Fix S1213 ignoring record classes.

leonardo-pilastri-sonarsource:


Wohops:
Changing approach in #4266

PR #4264, opened 2022-12-30
Use a container image with java 17 for ITs

christophe-zurn-sonarsource:


Wohops (approved):
LGTM!
Thanks @christophe-zurn-sonarsource !

PR #4274, opened 2023-01-12
SONARJAVA-4394 Rule S1849: Refactor HasNextCallingNext rule to not use non-static inner class

kaufco:


margarita-nedzelska-sonarsource (approved):
LGTM!

PR #93, opened 2015-03-12
SONARJAVA-928 RSPEC-2259 Null pointers should not be dereferenced

merciesa:

PR #4276, opened 2023-01-16
SONARJAVA-4395 Rule S1114: In ObjectFinalizeOverridenCallsSuperFinalizeCheck lastStatementTree field is not always cleaned

kaufco:


margarita-nedzelska-sonarsource (commented):
Looks good,
Just check the coverage, please.

java-checks/src/main/java/org/sonar/java/checks/ObjectFinalizeOverridenCallsSuperFinalizeCheck.java
```diff
@@ -47,76 +47,91 @@ public class ObjectFinalizeOverridenCallsSuperFinalizeCheck extends IssuableSubs
...
 
   @Override
   public void visitNode(Tree tree) {
-    if (tree.is(Tree.Kind.METHOD_INVOCATION)) {
-      MethodInvocationTree methodInvocationTree = (MethodInvocationTree) tree;
-      if (methodInvocationTree.methodSelect().is(Tree.Kind.MEMBER_SELECT)) {
-        MemberSelectExpressionTree mset = (MemberSelectExpressionTree) methodInvocationTree.methodSelect();
-        if (FINALIZE.equals(mset.identifier().name()) && mset.expression().is(Tree.Kind.IDENTIFIER) && "super".equals(((IdentifierTree) mset.expression()).name())) {
-          lastStatementTree = methodInvocationTree;
+    final MethodTree methodTree = (MethodTree) tree;
```
Wohops, line 57: I know this is redundant with the initiative you started about the style, however we have not been using the `final` keyword anywhere so far anywhere in the project. Can you remove them?

java-checks/src/main/java/org/sonar/java/checks/ObjectFinalizeOverridenCallsSuperFinalizeCheck.java
```diff
@@ -47,76 +47,91 @@ public class ObjectFinalizeOverridenCallsSuperFinalizeCheck extends IssuableSubs
...
-        Symbol.TypeSymbol currentClass = superClassType.symbol();
-        if (currentClass.lookupSymbols(FINALIZE).stream().anyMatch(FINALIZE_MATCHER::matches)) {
-          return true;
+  private static class FindLastSuperFinalizeInvocationVisitor extends BaseTreeVisitor {
+
+    @Nullable
+    private MethodInvocationTree lastSuperFinalizeInvocation;
+
+    @Nullable
+    MethodInvocationTree getLastSuperFinalizeInvocation() {
```
Wohops, line 123: The class is static and private, I think it is safe to use directly `return visitor.lastSuperFinalizeInvocation` on line #93 and drop the getter.

java-checks/src/main/java/org/sonar/java/checks/ObjectFinalizeOverridenCallsSuperFinalizeCheck.java
```diff
@@ -47,76 +47,91 @@ public class ObjectFinalizeOverridenCallsSuperFinalizeCheck extends IssuableSubs
...
   }
 
-  private static boolean doesOverrideFinalize(Symbol classSymbol) {
-    if (classSymbol.isTypeSymbol()) {
-      Type superClassType = ((Symbol.TypeSymbol) classSymbol).superClass();
-      while (superClassType != null && !superClassType.is("java.lang.Object")) {
-        Symbol.TypeSymbol currentClass = superClassType.symbol();
-        if (currentClass.lookupSymbols(FINALIZE).stream().anyMatch(FINALIZE_MATCHER::matches)) {
-          return true;
+  private static class FindLastSuperFinalizeInvocationVisitor extends BaseTreeVisitor {
```
Wohops, line 117: I believe this is not enough. You need to avoid exploring inner classes or anonymous classes in your base tree visitor. The following case will trigger both FNs and FPs:

```Java
class FpAndFn extends S1114_Class1 {
  @Override
  protected void finalize() throws Throwable {  // False Negative: there is no call to finalize here
    class InnerClass {
      void foo() throws Throwable {
        super.finalize(); // False positive: this call is NOT inside a finalize method();
      }
    }
  }
}
```

Wohops (changes requested):
Since you dropped completely the `lastStatementTree` field, I believe the commit message (and JIRA ticket) should be reworded to reflect the change.

java-checks/src/main/java/org/sonar/java/checks/ObjectFinalizeOverridenCallsSuperFinalizeCheck.java
```diff
@@ -47,76 +49,102 @@ public class ObjectFinalizeOverridenCallsSuperFinalizeCheck extends IssuableSubs
...
-          reportIssue(methodTree.simpleName(), "Add a call to super.finalize() at the end of this Object.finalize() implementation.");
-        } else if (!isLastStatement(methodTree, lastStatementTree)) {
-          reportIssue(lastStatementTree, "Move this super.finalize() call to the end of this Object.finalize() implementation.");
+  private static boolean isFinalizeOverriddenMethod(MethodTree methodTree) {
+    return FINALIZE_MATCHER.matches(methodTree) && doesOverrideFinalize(methodTree.symbol().owner());
+  }
+
+  private static boolean doesOverrideFinalize(@Nullable Symbol classSymbol) {
+    if ((classSymbol != null) && (classSymbol.isTypeSymbol())) {
+      Type superClassType = ((Symbol.TypeSymbol) classSymbol).superClass();
```
dorian-burihabwa-sonarsource, line 80: As this method is only called using owners of methods, and because Java does not allow methods to be defined outside of class-like constructs, we can safely assume that the owner is never and is always a type symbol.
```suggestion
  private static boolean doesOverrideFinalize(Symbol.TypeSymbol superClassType) {
```

To validate this, you can add a new test by copying part of your test cases to a file in the `src/main/files/non-compiling/checks` and the compiler should still be able to get a really symbol and resolve the method owner as class-like thing.

dorian-burihabwa-sonarsource: In general, I would recommend avoiding extra parentheses that don't enforce any additional precedence in if conditions by breaking the condition on multiple lines or extracting the test in a separate method

dorian-burihabwa-sonarsource (changes requested):
Nice work on supporting the edge case. The PR just needs a little bit of polish to get past the finish line

java-checks/src/main/java/org/sonar/java/checks/ObjectFinalizeOverridenCallsSuperFinalizeCheck.java
dorian-burihabwa-sonarsource: I would agree that the getter can be dropped here in favor of a public field. No need to protect against misusage that is unlikely to happen on a private static class

dorian-burihabwa-sonarsource (changes requested):
Looks good just a tiny method to remove to simplify the code

java-checks/src/main/java/org/sonar/java/checks/ObjectFinalizeOverridenCallsSuperFinalizeCheck.java
kaufco: Done.

dorian-burihabwa-sonarsource (approved):
LGTM Thanks for taking the changes into consideration

Wohops (approved):

Wohops:
LGTM! Nice work!

PR #4277, opened 2023-01-17
[No-jira] Extract autoscan test to its own module

dorian-burihabwa-sonarsource:
Extract the autoscan test to its own module in order to shorten the feedback loop and differentiate between failures on changes due to changing issue locations and issue numbers.


Wohops (dismissed):
While I agree with the general idea of splitting the ruling execution in two, I would favor a more radical approach, and extract the AutoScan test into its own module, so we can safely execute each of them independently. I fear that with the current approach, we might introduce a hole and forget about the two classes being the only one executed.

.cirrus.yml
```diff
@@ -182,9 +182,47 @@ ruling_win_task:
...
+    - mvn clean compile --projects java-checks-test-sources
+    - cd its/autoscan
+    - mvn package -Pit-autoscan -Dsonar.runtimeVersion=LATEST_RELEASE[9.9] -Dmaven.test.redirectTestOutputToFile=false -B -e -V -Dparallel=methods -DuseUnlimitedThreads=true
+  cleanup_before_cache_script: cleanup_maven_repository
   on_failure:
     actual_artifacts:
-      path: "${CIRRUS_WORKING_DIR}/its/ruling/target/actual/**/*"
+      path: "${CIRRUS_WORKING_DIR}/its/autoscan/target/actual/**/*"
+
+autoscan_win_task:
```
johann-beleites-sonarsource, line 210: Do we need the autoscan test on Windows? We have the ruling running on Windows already and e.g. plugin is also not being run on Windows, so I wonder if autoscan would be enough on Linux.

johann-beleites-sonarsource (approved):
Seems like a nice change! A question about running this on Windows, otherwise LGTM!

.cirrus.yml
dorian-burihabwa-sonarsource: Indeed, I am not sure if expect a different behavior for the 2 platforms. I will remove it for now

PR #4278, opened 2023-01-23
SONARJAVA-4315 Add quick fixes for S1066 (CollapsibleIfCandidateCheck)

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/CollapsibleIfCandidateCheck.java
```diff
@@ -85,4 +94,20 @@ private static boolean hasBodySingleIfStatement(StatementTree thenStatement) {
...
+    if (containingStatement.is(Tree.Kind.BLOCK)) {
+      StatementTree thenStatement = ifStatement.thenStatement();
+      if (thenStatement.is(Tree.Kind.BLOCK)) {
+        SyntaxToken closingBrace = ((BlockTree) thenStatement).closeBraceToken();
+        quickFixBuilder.addTextEdit(JavaTextEdit.removeTree(closingBrace));
+      } else {
+        quickFixBuilder.addTextEdit(JavaTextEdit.insertBeforeTree(ifStatement.thenStatement(), "{"));
+      }
+    }
+    quickFixBuilder.addTextEdit(JavaTextEdit.replaceBetweenTree(outerIf.closeParenToken(), ifStatement.openParenToken(), " && ("));
```
leonardo-pilastri-sonarsource, line 110: By ending the `.replaceBetweenTree` on the `ifStatement.openParenToken()` you are removing the starting open parenthesis of the inner if, and then replacing it back with another `(` , taken from the string you're passing `" && ("` .
Is this on purpose? Does it make it better or worse to go instead with:
`quickFixBuilder.addTextEdit(JavaTextEdit.replaceBetweenTree(outerIf.closeParenToken(), ifStatement.ifKeyword(), " && "));`
Apparently it should work the same to me

leonardo-pilastri-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/CollapsibleIfCandidateCheck.java
dorian-burihabwa-sonarsource: You are right it probably makes more sense to do it the way you proposed. Chaning it now

leonardo-pilastri-sonarsource (approved):
Looks good!

PR #4281, opened 2023-01-24
SONARJAVA-3938  Add quick fixes for S4719 (StandardCharsetsConstantsCheck)

chrislain-razafimahefa-sonarsource:


chrislain-razafimahefa-sonarsource:
I skipped handling the following special case because it is more involved:
`org.apache.commons.io.IOUtils.toString(bytes, "UTF-8"); `

java-checks/src/test/java/org/sonar/java/checks/StandardCharsetsConstantsCheckTest.java
```diff
@@ -21,36 +21,42 @@
...
       .withJavaVersion(7)
+      .withQuickFixes()
       .verifyIssues();
-    CheckVerifier.newVerifier()
-      .onFile(testSourcesPath("checks/StandardCharsetsConstantsCheck_java8.java"))
+    InternalCheckVerifier.newInstance()
+      .onFile(mainCodeSourcesPath("checks/StandardCharsetsConstantsCheck_java8.java"))
       .withCheck(new StandardCharsetsConstantsCheck())
       .withJavaVersion(8)
+      .withQuickFixes()
```
dorian-burihabwa-sonarsource, line 47: I am not sure anymore: Is it necessary to add `withQuickFixes` here because the check now supports quick fixes even though we did not modify the test file?

java-checks/src/main/java/org/sonar/java/checks/StandardCharsetsConstantsCheck.java
```diff
@@ -227,13 +234,35 @@ public void visitNode(Tree tree) {
...
+
+  private void reportQuickfixOnMemberSelect(IdentifierTree identifierTree, String identifier) {
+    QuickFixHelper.newIssue(context)
+      .forRule(this)
+      .onTree(identifierTree)
+      .withMessage("Replace \"com.google.common.base.Charsets." + identifier + "\" with \"StandardCharsets." + identifier + "\".")
+      .withQuickFixes(() -> quickFixesOnMemberSelect(identifierTree))
+      .report();
+  }
+
```
dorian-burihabwa-sonarsource, line 258: How about moving this block below the place with the newer additional methods that generate quickfixes?

java-checks/src/main/java/org/sonar/java/checks/StandardCharsetsConstantsCheck.java
```diff
@@ -227,13 +234,35 @@ public void visitNode(Tree tree) {
     }
   }
 
+  List<JavaQuickFix> quickFixesOnMemberSelect(IdentifierTree identifierTree) {
+    Tree parent = identifierTree.parent();
+    if (parent.is(Tree.Kind.MEMBER_SELECT)) {
+      MemberSelectExpressionTree parentMemberSelect = (MemberSelectExpressionTree) parent;
+      return List.of(
+        JavaQuickFix.newQuickFix(REPLACE_WITH_STANDARD_CHARSETS + identifierTree.name() + "\"")
+          .addTextEdit(JavaTextEdit.replaceTree(parentMemberSelect.expression(), "java.nio.charset.StandardCharsets"))
```
dorian-burihabwa-sonarsource, line 243: Using the [ImportSupplier](https://github.com/SonarSource/sonar-java/blob/4892839eea5c24a6ad45c4dcb8d7d9561377680e/java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java#L246), we should be able to add insert `StandardCharsets` with a minimal amount of text.
See this [check](https://github.com/SonarSource/sonar-java/blob/4892839eea5c24a6ad45c4dcb8d7d9561377680e/java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java#L157) for an example.

dorian-burihabwa-sonarsource (changes requested):

dorian-burihabwa-sonarsource (changes requested):
We are functionnly ther but for code clarity I think it would be best to move all the quickfix methods below the `checkCall` method so that levels of abstraction are easier to discern and reading the code does not require jumping up and down in the code base

java-checks/src/main/java/org/sonar/java/checks/StandardCharsetsConstantsCheck.java
```diff
@@ -252,11 +264,93 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
...
+      .forRule(this)
+      .onTree(callExpression)
+      .withMessage(String.format("Replace %s() call with StandardCharsets.%s", methodRef, constantName))
+      .withQuickFix(() -> quickfixOnCharsetCall(callExpression, constantName))
+      .report();
+  }
+
+  private void reportDefaultQuickfix(ExpressionTree charsetNameArgument, String constantName) {
+    List<JavaTextEdit> edits = new ArrayList<>();
+    edits.add(JavaTextEdit.replaceTree(charsetNameArgument, "StandardCharsets." + constantName));
```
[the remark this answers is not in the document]
chrislain-razafimahefa-sonarsource: This change cannot be applied because `edits` needs to be a mutable list.

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4282, opened 2023-01-24
SONARJAVA-4319 Add quick fixes for S2116 (ArrayHashCodeAndToStringCheck)

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ArrayHashCodeAndToStringCheck.java
```diff
@@ -41,6 +46,27 @@ protected MethodMatchers getMethodInvocationMatchers() {
...
+      .withMessage("Use \"Arrays." + methodName + "(array)\" instead.")
+      .withQuickFixes(() -> getQuickFix(mit))
+      .report();
   }
+
+  private static List<JavaQuickFix> getQuickFix(MethodInvocationTree tree) {
+    String methodName = ExpressionUtils.methodName(tree).name();
+    if ("toString".equals(methodName)) {
+      JavaQuickFix toStringQuickFix = JavaQuickFix.newQuickFix("Use \"Arrays.toString(array)\" instead.")
+        .addTextEdit(JavaTextEdit.replaceTree(tree, "Arrays.toString(" + tree.firstToken().text() + ")"))
```
Wohops, line 61: I feel this is not enough to build something that compiles. I see 2 issues:

1. The tree is not necessarily an identifier, which can lead to very different fixes:
```java
int[][] array = new int[42][42];
array[0].toString(); // you expect "Arrays.toString(array[0])", but would get  "Arrays.toString(array)"

this.foo.bar.myArray.toString();  // same problem, you would end up with "Arrays.toString(this)", which is wrong
```
2. If you don't import "java.util.Arrays", it won't compile. You need to check that it's already there, and if not, add it to the imports. Have a look at the [`QuickFixHelper.ImportSupplier`](https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/java/org/sonar/java/checks/helpers/QuickFixHelper.java#L232-L246), which might help you understand how to inject such import. The rule `org.sonar.java.checks.CollectionImplementationReferencedCheck` already uses it and might help you as well.

java-checks/src/main/java/org/sonar/java/checks/ArrayHashCodeAndToStringCheck.java
```diff
@@ -41,6 +46,27 @@ protected MethodMatchers getMethodInvocationMatchers() {
...
+        .build();
+      return Collections.singletonList(toStringQuickFix);
+    } else {
+      JavaQuickFix hashCodeQuickFix = JavaQuickFix.newQuickFix("Use \"Arrays.hashCode(array)\" instead.")
```
Wohops, line 65: The hashcode will suffer from the same problem than toString();

java-checks/src/main/java/org/sonar/java/checks/ArrayHashCodeAndToStringCheck.java
```diff
@@ -41,6 +46,27 @@ protected MethodMatchers getMethodInvocationMatchers() {
   @Override
   protected void onMethodInvocationFound(MethodInvocationTree mit) {
     String methodName = mit.methodSymbol().name();
-    reportIssue(ExpressionUtils.methodName(mit), "Use \"Arrays." + methodName + "(array)\" instead.");
+    QuickFixHelper.newIssue(context)
+      .forRule(this)
+      .onTree(ExpressionUtils.methodName(mit))
+      .withMessage("Use \"Arrays." + methodName + "(array)\" instead.")
+      .withQuickFixes(() -> getQuickFix(mit))
```
Wohops, line 53: The method name has already been computed, you might want to re-use it later in your quickfixes builder

Wohops (changes requested):
Nice first attempt! Some stuff to rework to make it work in (almost?) all situations.

Be mindful that you can have a ton of possible patterns before a `.toString()`, for instance, parenthesis, member-select, another array select, etc. You might want to consider using the quickfix builder to remove some parts before/after rather than replacing everything.

java-checks/src/main/java/org/sonar/java/checks/ArrayHashCodeAndToStringCheck.java
```diff
@@ -40,7 +52,41 @@ protected MethodMatchers getMethodInvocationMatchers() {
...
+    } else {
+      edits.add(JavaTextEdit.replaceTree(tree, "Arrays.hashCode(" + methodCallee + ")"));
+      return JavaQuickFix.newQuickFix("Use \"Arrays.hashCode(array)\" instead.")
+        .addTextEdits(edits)
+        .build();
+    }
+  }
+
+  private QuickFixHelper.ImportSupplier getImportSupplier() {
+    if (importSupplier == null) {
```
Wohops, line 86: I believe that we have a problem here.

Since you create an `ImportSupplier` based on the context, it means that there is a direct link between the current file and the rule. When changing to another file, the context will change but the `ImportSupplier` will stay the same. I believe this might (will?) lead to errors when generating the quickfix.

If you look at other rules using an import supplier, they reset to null the supplier after each file.

Wohops (changes requested):

Wohops (approved):

Wohops:
LGTM! Nice work!

PR #4283, opened 2023-01-24
SONARJAVA-3924 Add quickfix for S2129

leonardo-pilastri-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/StringPrimitiveConstructorCheck.java
```diff
@@ -82,4 +97,88 @@ private static boolean isBigIntegerPotentiallyBiggerThanLong(NewClassTree newCla
...
+      message = "Replace this \"String\" constructor with an empty string \"\"";
+      textEdit = JavaTextEdit.replaceTree(newClassTree, "\"\"");
+    } else if (BIG_INT_MATCHER.matches(newClassTree)) {
+      String arg = getFirstArgumentAsString(newClassTree).replace("\"", "");
+      try {
+        Integer.valueOf(arg);
+      } catch (NumberFormatException e) {
+        arg += "L";
+      }
+      String replacement = String.format("BigInteger.valueOf(%s)", arg);
```
dorian-burihabwa-sonarsource, line 123: Because big integers can be as big as the biggest integer array that Java can support, the best you can do if you want to validate that number is to make sure the string matches `^(\+|\-)?[:digit:]+`

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/StringPrimitiveConstructorCheck.java
```diff
@@ -20,51 +20,85 @@
...
   private static final BigInteger MIN_BIG_INTEGER_VALUE = BigInteger.valueOf(Long.MIN_VALUE);
   private static final BigInteger MAX_BIG_INTEGER_VALUE = BigInteger.valueOf(Long.MAX_VALUE);
 
+  private static final MethodMatchers EMPTY_STRING_MATCHER = MethodMatchers.create().ofTypes(STRING).constructor().addWithoutParametersMatcher().build();
+  private static final MethodMatchers BIG_INT_MATCHER = MethodMatchers.create().ofTypes("java.math.BigInteger").constructor().addParametersMatcher(STRING).build();
+
+  private static final MethodMatchers matchers = MethodMatchers.or(
+    EMPTY_STRING_MATCHER,
+    BIG_INT_MATCHER,
+    MethodMatchers.create().ofTypes(STRING).constructor().addParametersMatcher(STRING).build(),
```
Wohops, line 57: For the sack of readability, I believe it would be nice to extract this code into some static helper method within the class, for instance:
```java
private static MethodMatchers constructor(String type, String arg0) {
  return MethodMatchers.create()
    .ofTypes(type)
    .constructor()
    .addParametersMatcher(arg0)
    .build();
}
```
Which would allow you to write then:
```java
private static final MethodMatchers matchers = MethodMatchers.or(
    EMPTY_STRING_MATCHER,
    BIG_INT_MATCHER,
    constructor(STRING, STRING),
    constructor("java.lang.Byte", "byte"),
    constructor("java.lang.Character", "char"),
    ...

);
```

java-checks/src/main/java/org/sonar/java/checks/StringPrimitiveConstructorCheck.java
```diff
@@ -20,51 +20,85 @@
...
-    if(isBigIntegerPotentiallyBiggerThanLong(newClassTree)) {
+    if (isBigIntegerPotentiallyBiggerThanLong(newClassTree)) {
       return;
     }
-    reportIssue(newClassTree.identifier(), "Remove this \"" + newClassTree.symbolType().name() + "\" constructor");
+    if(matchers.matches(newClassTree)) {
+      QuickFixHelper.newIssue(context)
+        .forRule(this)
+        .onTree(newClassTree.identifier())
+        .withMessage("Remove this \"" + newClassTree.symbolType().name() + "\" constructor")
```
Wohops, line 98: `withMessage()` allows you to directly pass parameters. So you can clean it a lot by using the following:
```java
.withMessage("Remove this \"%s\" constructor.", newClassTree.symbolType().name())
```
And you can also extract the issue message constant then.

java-checks/src/main/java/org/sonar/java/checks/StringPrimitiveConstructorCheck.java
```diff
@@ -82,4 +116,44 @@ private static boolean isBigIntegerPotentiallyBiggerThanLong(NewClassTree newCla
       return true;
     }
   }
+
+  private JavaQuickFix computeQuickFix(NewClassTree newClassTree) {
+    String message = "";
+    JavaTextEdit textEdit = null;
+    String className = QuickFixHelper.contentForTree(newClassTree.identifier(), context);
```
Wohops, line 123: if you just search for the className, you can probably just use the symbol here:
```java
String className = newClassTree.symbolType().name();
```

java-checks/src/main/java/org/sonar/java/checks/StringPrimitiveConstructorCheck.java
```diff
@@ -82,4 +116,44 @@ private static boolean isBigIntegerPotentiallyBiggerThanLong(NewClassTree newCla
...
+    return JavaQuickFix.newQuickFix(message).addTextEdit(textEdit).build();
+  }
+
+  private String formatQuickFixMessage(String constructor, String replacement) {
+    return String.format(QUICK_FIX_MESSAGE, constructor, replacement);
+  }
+
+  private String getFirstArgumentAsString(NewClassTree newClassTree) {
+    ExpressionTree expr = newClassTree.arguments().get(0);
+    if (expr instanceof LiteralTreeImpl) {
```
Wohops, line 146: Why use implementation class here?! You can use the interface! (`LiteralTree` instead of `LiteralTreeImpl`)

java-checks/src/main/java/org/sonar/java/checks/StringPrimitiveConstructorCheck.java
```diff
@@ -82,4 +116,44 @@ private static boolean isBigIntegerPotentiallyBiggerThanLong(NewClassTree newCla
...
+      LiteralTreeImpl arg = (LiteralTreeImpl) expr;
+      return arg.value();
+    } else if (expr instanceof IdentifierTreeImpl) {
+      IdentifierTreeImpl arg = (IdentifierTreeImpl) expr;
+      return arg.name();
+    } else if (expr instanceof TypeCastExpressionTreeImpl){
```
Wohops, line 152: why not `expr.is(Tree.Kind.TYPE_CAST)`?

java-checks/src/main/java/org/sonar/java/checks/StringPrimitiveConstructorCheck.java
```diff
@@ -82,4 +116,44 @@ private static boolean isBigIntegerPotentiallyBiggerThanLong(NewClassTree newCla
...
+  private String formatQuickFixMessage(String constructor, String replacement) {
+    return String.format(QUICK_FIX_MESSAGE, constructor, replacement);
+  }
+
+  private String getFirstArgumentAsString(NewClassTree newClassTree) {
+    ExpressionTree expr = newClassTree.arguments().get(0);
+    if (expr instanceof LiteralTreeImpl) {
+      LiteralTreeImpl arg = (LiteralTreeImpl) expr;
+      return arg.value();
+    } else if (expr instanceof IdentifierTreeImpl) {
```
Wohops, line 149: why not `expr.is(Tree.Kind.IDENTIFIER)`?

java-checks/src/main/java/org/sonar/java/checks/StringPrimitiveConstructorCheck.java
```diff
@@ -82,4 +116,44 @@ private static boolean isBigIntegerPotentiallyBiggerThanLong(NewClassTree newCla
...
+      IdentifierTreeImpl arg = (IdentifierTreeImpl) expr;
+      return arg.name();
+    } else if (expr instanceof TypeCastExpressionTreeImpl){
+      return QuickFixHelper.contentForTree(expr, context);
+    }else {
+      return "";
```
Wohops, line 155: I'm really not sure about this.
From what I understand, this means you will probably end up with an empty replacement, and so it's not equivalent. Are you sure?

Since it is not covered, you have no idea what will happen here. This is not tested

Wohops (changes requested):
Good starts. I let you handle my comments. Also, be careful with coverage. I feel that the only line not covered is going to cause problems. You also have an open issue on Next. Please fix it/them before assigning PR for reviews, thanks!

Wohops (approved):
LGTM! Nice work!

PR #4284, opened 2023-01-24
SONARJAVA-4314 Add quick fixes for S1450 (PrivateFieldUsedLocallyCheck)

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/PrivateFieldUsedLocallyCheck.java
```diff
@@ -49,6 +56,7 @@
 public class PrivateFieldUsedLocallyCheck extends IssuableSubscriptionVisitor {
 
   private static final String MESSAGE = "Remove the \"%s\" field and declare it as a local variable in the relevant methods.";
+  private static final String QUICK_FIX_MESSAGE = "Move declaration to the relevant method";
```
chrislain-razafimahefa-sonarsource, line 59: Not sure it is better but as I user I think I would have preferred a message like:
'Move field to the method where it is used'.

java-checks/src/main/java/org/sonar/java/checks/PrivateFieldUsedLocallyCheck.java
```diff
@@ -77,10 +89,69 @@ private void checkPrivateField(Symbol privateFieldSymbol, TypeSymbol classSymbol
...
+      JavaQuickFix.newQuickFix(QUICK_FIX_MESSAGE)
+        .addTextEdits(editUsagesWithThis(symbol))
+        .addTextEdit(JavaTextEdit.insertAfterTree(openingBrace, newDeclaration))
+        .addTextEdit(JavaTextEdit.removeTree(declaration))
+        .build()
+    );
+  }
+
+  /*
+   * Compares the name of the field against the local variables and parameters in the method to ensure that moving the
```
chrislain-razafimahefa-sonarsource, line 125:
```suggestion
   * Compares the field name against local variables and parameters in the method to ensure that moving the
```

java-checks/src/main/java/org/sonar/java/checks/PrivateFieldUsedLocallyCheck.java
```diff
@@ -77,10 +89,69 @@ private void checkPrivateField(Symbol privateFieldSymbol, TypeSymbol classSymbol
...
+   * field declaration to the method would not create any clash.
+   */
+  private static boolean relocationWouldClashWithLocalVariables(Symbol.VariableSymbol symbol, MethodTree method) {
```
chrislain-razafimahefa-sonarsource, line 128: Rename method to 'wouldRelocationClashWithLocalVariables'.

chrislain-razafimahefa-sonarsource (changes requested):
Well done. Especially that this one was not a straightforward one.
I have a few suggestions for you to consider.

java-checks-test-sources/src/main/java/checks/PrivateFieldUsedLocallyCheck.java
```diff
@@ -0,0 +1,70 @@
...
+  class LocalVariableAmbiguity {
+    private int localVariable; // Noncompliant [[sc=17;ec=30;quickfixes=]]
+
+    int useLocally(int parameter) {
+      int localVariable = 12;
+      this.localVariable = localVariable;
+      return this.localVariable * parameter;
+    }
+  }
+
```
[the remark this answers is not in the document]
chrislain-razafimahefa-sonarsource: Here are example tests I would add:

```
class A {
    private int arg;

    A(int b) {
        arg = b;
        System.out.println(arg);
    }

    A(int arg) {
        this.arg = arg;
    }

  void aMethod(int arg) {
    this.arg = arg;
    System.out.println(arg);
  }
}
```

java-checks/src/main/java/org/sonar/java/checks/PrivateFieldUsedLocallyCheck.java
dorian-burihabwa-sonarsource: How about `Move this field to the only method where it is used` ?

dorian-burihabwa-sonarsource: Knowing that the message from IntelliJ is `Convert this field to a local variable in method '<method name>'`

chrislain-razafimahefa-sonarsource: `Move this field to the only method where it is used` is even better. It gives more clue to the user.

chrislain-razafimahefa-sonarsource (approved):
There is just the message change left but otherwise it is all good. Well done!

PR #4286, opened 2023-01-25
SONARJAVA-4317 Add quickfix for S2147

leonardo-pilastri-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/CombineCatchCheck.java
```diff
@@ -47,23 +51,41 @@ public void visitNode(Tree tree) {
...
       catches.add(catchTree);
     }
   }
 
-  private void reportIssue(CatchTree catchTree, CatchTree catchTreeToBeCompared) {
-    String message = "Combine this catch with the one at line " + LineUtils.startLine(catchTreeToBeCompared.catchKeyword())
-      + ", which has the same body." + context.getJavaVersion().java7CompatibilityMessage();
+  private void reportIssueWithQuickFix(CatchTree catchTree, CatchTree catchTreeToBeCompared) {
+    String quickFixMessage = "Combine this catch with the one at line " + LineUtils.startLine(catchTreeToBeCompared.catchKeyword());
+    String issueMessage = quickFixMessage + ", which has the same body." + context.getJavaVersion().java7CompatibilityMessage();
```
dorian-burihabwa-sonarsource, line 64: We should be careful about not having long messages since they have to appear in the contextual menu

java-checks/src/main/java/org/sonar/java/checks/CombineCatchCheck.java
```diff
@@ -47,23 +51,41 @@ public void visitNode(Tree tree) {
...
+
+  private JavaQuickFix computeQuickFix(CatchTree catchTree, CatchTree catchTreeToBeCompared, String qfMessage) {
+    var builder = JavaQuickFix.newQuickFix(qfMessage);
+    builder.addTextEdit(JavaTextEdit.removeTree(catchTree));
+    SyntaxToken openParent = catchTreeToBeCompared.openParenToken();
+    builder.addTextEdit(
+      JavaTextEdit.insertAfterTree(
+        openParent,
+        QuickFixHelper.contentForTree(catchTree.parameter().type(), context) + " | "));
+    return builder.build();
```
dorian-burihabwa-sonarsource, line 88: We need to check deeper what kind of type is handled by the catch clause as some combinations lead to compilation breaking.
```java
 void writeStuff(String message) {
    try (OutputStream out = new FileOutputStream("/tmp/out.txt")) {
      out.write(message.getBytes(StandardCharsets.UTF_8));
    } catch (FileNotFoundException e) {
      throw new RuntimeException(e);
    } catch (IOException e) {
      throw new RuntimeException(e);
    }
  }
```
In this case, because `FileNotFoundException` is a sub type of `IOException`, the masking breaks compilation

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/CombineCatchCheck.java
leonardo-pilastri-sonarsource: The `quickFixMessage` is the one going in the context menu, and is a portion of the original issue message, do you feel it could be too long? To me, we might have issues only with really big line numbers but I don't know if it makes sense to consider that case

dorian-burihabwa-sonarsource: All good, it looks like the message is not too long
Screenshot from 2023-01-30 14-26-02

dorian-burihabwa-sonarsource (changes requested):
We are almost there but I feel like we can improve further the way that we remove super types and duplicates

dorian-burihabwa-sonarsource (approved):

PR #4287, opened 2023-01-26
SONARJAVA-4408 Ruling tests fail for external contributors due to non-public dependencies

kaufco:


README.md
```diff
@@ -99,9 +99,11 @@ The "Sanity Test" is a test which runs all checks against all the test sources f
 
 The "Plugin Test" is an integration test suite which verifies plugin features such as metric calculation, coverage etc. To launch it:
 
-    mvn clean install -Pit-plugin
+    mvn clean install -Pit-plugin -DskipTests=false -DcommunityEditionTestsOnly=true
```
alban-auzeill, line 102: It should work with `-Pit-plugin` and without `-DskipTests=false` because it's what the profile `it-plugin` is supposed to do in `its/plugin/pom.xml`:
```xml
    <profile>
      <id>it-plugin</id>
      <properties>
        <skipTests>false</skipTests>
      </properties>
    </profile>
```

README.md
```diff
@@ -99,9 +99,11 @@ The "Sanity Test" is a test which runs all checks against all the test sources f
...
 
-Beware that some tests, specifically the `JspTest`, may fail due a missing or unauthorized Github token.
+Note for internal contributors: in order to execute also the tests that depend on the SonarQube Enterprise Edition, use:
+
+    mvn clean install -Pit-plugin -DskipTests=false
```
alban-auzeill, line 106: same here, remove `-DskipTests=false`

README.md
```diff
@@ -116,9 +118,13 @@ Failing to do so will produce inconsistencies with the expected results.
 
 From the `its/ruling` folder, launch the ruling tests:
 
-    mvn clean install -DskipTests=false
+    mvn clean install -DskipTests=false -DcommunityEditionTestsOnly=true
```
alban-auzeill, line 121: Here we should replace `-DskipTests=false` with `-Pit-ruling`, because it is the way `.cirrus.yml` execute its `ruling_task:`

its/ruling/src/test/java/org/sonar/java/it/JavaRulingTest.java
```diff
@@ -38,15 +39,10 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
alban-auzeill, line 42: We don't use import '*', in the https://github.com/SonarSource/sonar-developer-toolset there's a configuration saying:
```
Class count to use import with '*'" -> 999
```

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #4288, opened 2023-01-30
SONARJAVA-3995 FP S3400 when method can be overridden

alban-auzeill:


kaufco:
Rejected, because Ruling fails. its/ruling/src/test/resources/**/java-S3400.json need to be adapted to the change in rule behavior.

kaufco (changes requested):
Ruling fails, needs to be fixed before approval: its/ruling/src/test/resources/**/java-S3400.json files must be adapted to change in rule behavior. Also, these files are missing in the repo at the moment.

kaufco:
Looks good!

kaufco (approved):
Looks good!

PR #4289, opened 2023-01-30
SONARJAVA-4315 Add missing closing parenthesis to quick fix

dorian-burihabwa-sonarsource:


leonardo-pilastri-sonarsource (approved):
Looks like working fine now!

PR #4290, opened 2023-01-30
SONARJAVA-4415 Add parameter to ignore particular annotations in S1068

bvfalcon:
Parameter `ignoreAnnotations` (default value "") can contain a list of annotation names (example: "javax.inject.Inject,jakarta.inject.Inject") which would be ignored by Analyzer. In other words, annotations are listed in `ignoreAnnotations` have no impact on analysis process. In other case they cause false-negative result of rule evaluation


kaufco:
Review is continued in [PR 4637](https://github.com/SonarSource/sonar-java/pull/4637) (mirrors this PR to trigger CI and add further commits).

PR #4291, opened 2023-01-31
SONARJAVA-4321 Add quickfix for S2225 (ToStringReturningNullCheck)

leonardo-pilastri-sonarsource:


irina-batinic-sonarsource (changes requested):
I would add some more test examples, for example:
public String toString () {
```
  if (this.collection.isEmpty()) {
    return null; // Noncompliant
  } else {
    // ...
  }
```
and assure that quick fixes only work with `toString` methods.

irina-batinic-sonarsource (approved):
LGTM!

PR #4294, opened 2023-02-01
SONARJAVA-4317 Fix quickfix S2147 for multiple caught exceptions subclasses

leonardo-pilastri-sonarsource:


dorian-burihabwa-sonarsource (approved):
Great Job!

PR #4295, opened 2023-02-01
SONARJAVA-4403 Add exception to S3553

johann-beleites-sonarsource:
Don't flag Optional uses as parameter types when the method is overriding

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/OptionalAsParameterCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/OptionalAsParameterCheck.java
@@ -48,11 +48,17 @@ public class OptionalAsParameterCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public void visitNode(Tree tree) {
-    for (VariableTree parameter : ((MethodTree) tree).parameters()) {
-      TypeTree typeTree = parameter.type();
-      Optional<String> msg = expectedTypeInsteadOfOptional(typeTree.symbolType());
-      if (msg.isPresent()) {
-        reportIssue(typeTree, msg.get());
+    final var methodTree = (MethodTree) tree;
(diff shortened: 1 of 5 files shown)


java-checks/src/main/java/org/sonar/java/checks/OptionalAsParameterCheck.java
```diff
@@ -48,11 +48,17 @@ public List<Tree.Kind> nodesToVisit() {
...
   public void visitNode(Tree tree) {
-    for (VariableTree parameter : ((MethodTree) tree).parameters()) {
-      TypeTree typeTree = parameter.type();
-      Optional<String> msg = expectedTypeInsteadOfOptional(typeTree.symbolType());
-      if (msg.isPresent()) {
-        reportIssue(typeTree, msg.get());
+    final var methodTree = (MethodTree) tree;
+
+    // If the method is overriding something, the user has no control over the parameter type here, so we should not raise an issue.
+    if (methodTree.isOverriding() == Boolean.FALSE) {
```
alban-auzeill, line 54: Interview question: What is the result of this java expression?
```java
(new Boolean(false) == Boolean.FALSE)
```
So here, because we are not `kotlin` developers :stuck_out_tongue: but `java` developers :grin:, we should use:
```java
Boolean.FALSE.equals(methodTree.isOverriding())
```

Except for this, I approve this PR, once fixed you can merge.

alban-auzeill (approved):

java-checks/src/main/java/org/sonar/java/checks/OptionalAsParameterCheck.java
johann-beleites-sonarsource: :eyes:

PR #4296, opened 2023-02-01
SONARJAVA-4350 Improve the suggested quick fix for S1068 when there are some writes to the variable

dorian-burihabwa-sonarsource:


java-checks-test-sources/src/main/java/checks/unused/UnusedPrivateFieldCheckWithQuickfixes.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  private class QuickFixesRemovesAsssignmentFromLoopInitializer {
+    // fix@qf3 {{Remove this unused private field}}
+    // edit@qf3 [[sl=+3;el=+3;sc=12;ec=26]] {{}}
+    // edit@qf3 [[sc=5;ec=19]] {{}}
+    private int x; // Noncompliant [[sc=17;ec=18;quickfixes=qf3]]
+
+    QuickFixesRemovesAsssignmentFromLoopInitializer(int x, int y) {
+      for (this.x = x + y; y < 0; y++) {
```
chrislain-razafimahefa-sonarsource, line 38: What happens if I have this?
`for (this.x = x + y++; y < 0; y++)`
Now there is the `y++` side-effect.

chrislain-razafimahefa-sonarsource (changes requested):
I tested the concern I mention in this review and unless I am wrong, I think it needs to be handled first.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateFieldCheck.java
```diff
@@ -242,15 +243,20 @@ private static List<JavaTextEdit> computeExpressionCaptures(List<AssignmentExpre
...
+      name = identifier.name() + index;
+    } else {
+      identifier = ((MemberSelectExpressionTree) variable).identifier();
+      name = identifier.name() +  index;
     }
     name = Character.toUpperCase(name.charAt(0)) + name.substring(1);
-    return String.format("var valueFormerlyAssignedTo%s = ", name);
+    TypeTree typeInDeclaration = ((VariableTree) identifier.symbol().declaration()).type();
+    String type = QuickFixHelper.contentForTree(typeInDeclaration, context);
+    return String.format("%s valueFormerlyAssignedTo%s = ", type, name);
```
chrislain-razafimahefa-sonarsource, line 259: I think this approach will fail on the following example:
```
private class QuickfixAndCall {
    private int x;

    QuickfixAndCall(int y) {
      System.out.println(x = y + 1);
    }
}
```

The solution might be to declare the new variable one line before its usage?

chrislain-razafimahefa-sonarsource (changes requested):
I think there is still an issue.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateFieldCheck.java
chrislain-razafimahefa-sonarsource: Actually the rule do not trigger here, so it is all good.

chrislain-razafimahefa-sonarsource (approved):
All good! Well done. It was a tricky one.

PR #4298, opened 2023-02-02
SONARJAVA-4311 Add quick fixes for S1217 (ThreadRunCheck)

chrislain-razafimahefa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ThreadRunCheck.java
```diff
@@ -50,7 +54,19 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
     if (parent != null && THREAD_RUN_METHOD_MATCHER.matches((MethodTree) parent)) {
       return;
     }
-    reportIssue(ExpressionUtils.methodName(mit), "Call the method Thread.start() to execute the content of the run() method in a dedicated thread.");
+
+    IdentifierTree methodName = ExpressionUtils.methodName(mit);
```
leonardo-pilastri-sonarsource, line 58: I don't see a reason to be specific about the type of `methodName`, so maybe we could store it as a simple `Tree` and also change the `computeQuickFix` to take a `Tree` as parameter
```suggestion
    Tree methodName = ExpressionUtils.methodName(mit);
```

java-checks/src/main/java/org/sonar/java/checks/ThreadRunCheck.java
```diff
@@ -50,7 +54,19 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
...
+    QuickFixHelper.newIssue(context)
+      .forRule(this)
+      .onTree(methodName)
+      .withMessage("Call the method Thread.start() to execute the content of the run() method in a dedicated thread.")
+      .withQuickFix(() -> computeQuickFix(methodName))
+      .report();
   }
 
+  private static JavaQuickFix computeQuickFix(IdentifierTree methodName) {
```
leonardo-pilastri-sonarsource, line 67:
```suggestion
  private static JavaQuickFix computeQuickFix(Tree methodName) {
```

java-checks/src/main/java/org/sonar/java/checks/ThreadRunCheck.java
```diff
@@ -50,7 +54,19 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
...
+    return JavaQuickFix.newQuickFix("Call method Thread.start()")
```
leonardo-pilastri-sonarsource, line 68: While it's true that the raised issue message is `"Call the method Thread.start() to execute..."` , I feel like the quick fix message could describe a little bit better what it will actually do on the code when applied, maybe something along the lines of:
```suggestion
    return JavaQuickFix.newQuickFix("Replace the .run() call with .start() call")
```
Or something like that, maybe you can come up with something else that sounds better!

leonardo-pilastri-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/ThreadRunCheck.java
chrislain-razafimahefa-sonarsource: But what does it buy to be more generic here?
Whereas if we keep `IdentifierTree`, assuming that in the future there is a change and the change requires a call (from `methodName) to a method defined in `IdentifierTree`, we would not have to cast.
What do you think?
In general, I think it is better to be more specific.

java-checks/src/main/java/org/sonar/java/checks/ThreadRunCheck.java
chrislain-razafimahefa-sonarsource: You are right.
What about 'Replace run() with start()'?

java-checks/src/main/java/org/sonar/java/checks/ThreadRunCheck.java
leonardo-pilastri-sonarsource: Mh, yeah it's a good point, since `.methodName` returns a `IdentifierTree` and it doesn't have any cost to just keep it as it is, being specific doesn't hurt and it could be useful in the future, I agree

java-checks/src/main/java/org/sonar/java/checks/ThreadRunCheck.java
leonardo-pilastri-sonarsource: Yeah sounds good to me, after this I feel we're good to go! Nice job!

leonardo-pilastri-sonarsource (approved):

PR #4299, opened 2023-02-02
SONARJAVA-3959 Add quick fixes for S1656 (SelfAssignementCheck)

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/SelfAssignementCheck.java
```diff
@@ -54,28 +69,77 @@ public void visitNode(Tree tree) {
...
+
+      if (isParameter) {
+        List<String> memberNames = classParent.members().stream()
+          .filter(m -> m.is(Tree.Kind.VARIABLE))
+          .map(VariableTree.class::cast)
+          .map(m -> m.simpleName().name())
+          .collect(Collectors.toList());
+
+        if (memberNames.contains(name)) {
+          return JavaQuickFix.newQuickFix("Correct this useless self-assignment")
```
dorian-burihabwa-sonarsource, line 100: I think the quick fix message is a bit contradictory: we advise correcting the statement but also this is a useless statement.
How about changing the message?
```suggestion
          return JavaQuickFix.newQuickFix("Disambiguate this self-assignment")
```

java-frontend/src/test/java/org/sonar/java/model/ExpressionUtilsTest.java
```diff
@@ -284,6 +285,23 @@ void enclosing_method_test() {
...
+    File file = new File("src/test/files/model/ExpressionEnclosingMethodTest.java");
+    CompilationUnitTree tree = JParserTestUtils.parse(file);
+    FindAssignment findAssignment = new FindAssignment();
+    tree.accept(findAssignment);
+    findAssignment.assignments.forEach(a -> {
+      ClassTree parent = (ClassTree) ExpressionUtils.getParentOfType(a, Tree.Kind.CLASS, Tree.Kind.INTERFACE);
+      if (parent.is(Tree.Kind.CLASS)) {
+        assertThat(parent.simpleName().name()).isEqualTo("Clazz");
+      }
+      if (parent.is(Tree.Kind.INTERFACE)) {
```
dorian-burihabwa-sonarsource, line 299: Can a tree be a class and interface? Otherwise, I would join these two ifs as mutually exclusive branches of the same tree.
```suggestion
      } else if (parent.is(Tree.Kind.INTERFACE)) {
```

dorian-burihabwa-sonarsource (changes requested):
The changes look pretty good but I think we should update the quick fix message to be more aligned with its effective effect

java-frontend/src/test/java/org/sonar/java/model/ExpressionUtilsTest.java
```diff
@@ -284,6 +285,22 @@ void enclosing_method_test() {
...
+    FindAssignment findAssignment = new FindAssignment();
+    tree.accept(findAssignment);
+    findAssignment.assignments.forEach(a -> {
+      ClassTree parent = (ClassTree) ExpressionUtils.getParentOfType(a, Tree.Kind.CLASS, Tree.Kind.INTERFACE);
+      if (parent.is(Tree.Kind.CLASS)) {
+        assertThat(parent.simpleName().name()).isEqualTo("Clazz");
+      } else if (parent.is(Tree.Kind.INTERFACE)) {
+        assertThat(parent.simpleName().name()).isEqualTo("I1");
+      }
+    });
```
dorian-burihabwa-sonarsource, line 301: Because the method `getParentOfType` is added to the frontend module API, we should cover all branches here.
The current test only validates use cases that are related to S1656 but we should also test other usages of the API.
```suggestion
    });
   assertThat(ExpressionUtils.getParentOfType(tree, Tree.Kind.COMPILATION_UNIT)).isNull();
```

java-checks/src/main/java/org/sonar/java/checks/SelfAssignementCheck.java
```diff
@@ -54,28 +69,77 @@ public void visitNode(Tree tree) {
...
       updateWarnings(node);
     }
   }
 
+  private static JavaQuickFix getQuickFix(AssignmentExpressionTree tree) {
+    ClassTree classParent = (ClassTree) ExpressionUtils.getParentOfType(tree, Tree.Kind.CLASS);
+    MethodTree methodParent = (MethodTree) ExpressionUtils.getParentOfType(tree, Tree.Kind.METHOD, Tree.Kind.CONSTRUCTOR);
+    String name = getName(tree.variable());
+
+    if (methodParent != null && classParent != null) {
```
dorian-burihabwa-sonarsource, line 87: This line is not fully covered in the tests and I am not sure we need to test if an assignment has a class parent in its hierarchy because a method can only exist in a class. Maybe we can simplify, and remove part of the condition?
```suggestion
    if (methodParent != null) {
```

java-checks/src/main/java/org/sonar/java/checks/SelfAssignementCheck.java
```diff
@@ -54,28 +69,77 @@ public void visitNode(Tree tree) {
...
+        .anyMatch(p -> p.equals(name));
+
+      if (isParameter) {
+        List<String> memberNames = classParent.members().stream()
+          .filter(m -> m.is(Tree.Kind.VARIABLE))
+          .map(VariableTree.class::cast)
+          .map(m -> m.simpleName().name())
+          .collect(Collectors.toList());
+
+        if (memberNames.contains(name)) {
```
dorian-burihabwa-sonarsource, line 99: Do we test a case where we have a class with no field and reassignment of a parameter/variable to itself? I think this would cover the remaining branch

dorian-burihabwa-sonarsource (changes requested):
Thanks for applying the changes from the last review, let's get coverage to 100% before we merge this one into master

dorian-burihabwa-sonarsource (approved):
Great job on the corner cases and coverage 👍🏿

PR #4301, opened 2023-02-02
SONARJAVA-4187 S3329 should not report FP when the semantic is incomplete

kaufco:


java-checks-test-sources/src/main/files/non-compiling/checks/security/CipherBlockChainingCheck.java
```diff
@@ -0,0 +1,33 @@
...
+    IvParameterSpec iv = new IvParameterSpec(biv); // Noncompliant
+  }
+
+  void unknownFunctionFromRandomIv() throws UnsupportedEncodingException {
+    byte[] iv = random.unknown(16);
+    IvParameterSpec ivParameterSpec = new IvParameterSpec(iv); // Compliant, because function cannot be resolved
+  }
+
+  void unknownFunctionIv() throws UnsupportedEncodingException {
+    byte[] iv = unknown(16); // "iv" is random thanks to SecureRandom
```
johann-beleites-sonarsource, line 30: The comment doesn't make sense to me - a remainder of copy+pasting test code from elsewhere?

johann-beleites-sonarsource (approved):
One small comment about a comment, then LGTM!

java-checks-test-sources/src/main/files/non-compiling/checks/security/CipherBlockChainingCheck.java
kaufco: Yes, you're right. That comes from a copy and past of the compiling tests.

PR #4303, opened 2023-02-03
SONARJAVA-2573 Methods 'add', 'remove' and some other should not be called on fixed-length collections

bvfalcon:


johann-beleites-sonarsource:
Tracked in [SONARJAVA-4413](https://sonarsource.atlassian.net/browse/SONARJAVA-4413)

kaufco (changes requested):
Hi @bvfalcon , the current implementation creates a false positive for this minimal example:

```
  void foo() {
      List<String> list = new ArrayList<>();
      list.add("42");
      list = Collections.emptyList();
  }
```

It seems that the implementation doesn't check whether the type assignment was before or after the actual method call. However, it is not that easy as looking for the line number because we also could have loops (in fact, that's what's happening in `org/apache/commons/beanutils2/converters/ArrayConverter.java` in the ruling tests, where S4265 generates an FP on line 468) and would need a complete data flow analysis on the scope of the variable for an exact implementation.

I would suggest an approximate implementation (without the data flow analysis) that just checks all type assignments to a variable, and when they contain at least one type that supports add, remove etc. (like `ArrayList` in my example), the rule should not report on the method calls.

PR #4312, opened 2023-02-14
SONARJAVA-4254 Fix FP S4684 when @Entity are not converter from json automatically

margarita-nedzelska-sonarsource:


kaufco (approved):
LGTM!

PR #4317, opened 2023-02-22
Irina/sonarjava 4396

irina-batinic-sonarsource:


java-checks-testkit/src/test/java/org/sonar/java/checks/verifier/internal/InternalCheckVerifierTest.java
```diff
@@ -967,7 +967,7 @@ void withCache_effectively_sets_the_caches_for_scanWithoutParsing() {
       .withCheck(check)
       .verifyNoIssues();
 
-    verify(check, times(1)).scanWithoutParsing(argThat(context -> equivalent(cacheContext, context.getCacheContext())));
+    verify(check, times(0)).scanWithoutParsing(argThat(context -> equivalent(cacheContext, context.getCacheContext())));
```
dorian-burihabwa-sonarsource, line 970: let's have a look at this change

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -476,9 +481,19 @@ public boolean fileCanBeSkipped(InputFile inputFile) {
...
+      contentHashCache.writeToCache(inputFile);
+      return false;
+    }
+    if (!inputFile.status().equals(InputFile.Status.SAME)) {
+      LOG.trace("File status is: " + inputFile.status() + ". File can be skipped.");
+      contentHashCache.writeToCache(inputFile);
       return false;
     }
-    return canSkipInContext && inputFile.status() == InputFile.Status.SAME;
+    return contentHashCache.hasSameHashCached(inputFile);
```
dorian-burihabwa-sonarsource, line 496: It looks like we are taking a shortcut by checking the status before checking the cache. My understanding is that we should check the cache, and if the cache is disabled, check the status. WDYT?

java-frontend/src/main/java/org/sonar/java/caching/ContentHashCache.java
```diff
@@ -0,0 +1,117 @@
...
+    if (writeCache != null) {
+      LOG.trace(String.format("Writing to the cache for file %s", inputFile.key()));
+      String cacheKey = getCacheKey(inputFile);
+      try {
+        writeCache.write(cacheKey, FileHashingUtils.inputFileContentHash(inputFile));
+        return true;
+      } catch (IllegalArgumentException e) {
+        LOG.error(String.format("Tried to write multiple times to cache key '%s'. Ignoring writes after the first.", cacheKey));
+      } catch (IOException | NoSuchAlgorithmException e) {
+        LOG.error(String.format(HASH_COMPUTE_FAIL_MSG, inputFile.key()));
```
dorian-burihabwa-sonarsource, line 101: We should change the level of these logs to warning because while it may be interesting to investigate for us, failure to write to the cache should not affect analysis correctness

java-frontend/src/main/java/org/sonar/java/caching/ContentHashCache.java
```diff
@@ -0,0 +1,117 @@
...
+
+    if (enabled) {
+      readCache = context.previousCache();
+      writeCache = context.nextCache();
+    }
+  }
+
+  public boolean hasSameHashCached(InputFile inputFile) {
+    if (!enabled) {
+      LOG.trace("Cannot read from cache when disabled");
```
dorian-burihabwa-sonarsource, line 55: Let's make the log message more specific here
```suggestion
      LOG.trace(() -> "Cannot read cached hashes when the cache is disabled (" + inputFile.key() + ").");
```

java-frontend/src/main/java/org/sonar/java/caching/ContentHashCache.java
```diff
@@ -0,0 +1,117 @@
...
+      writeToCache(inputFile);
+    } catch (IOException | NoSuchAlgorithmException e) {
+      LOG.trace(String.format(HASH_COMPUTE_FAIL_MSG, inputFile.key()));
+    }
+    return false;
+  }
+
+  public boolean contains(InputFile inputFile) {
+    if (!enabled) {
+      LOG.trace("Cannot read from cache when disabled");
```
dorian-burihabwa-sonarsource, line 81:
```suggestion
      LOG.trace(() -> "Cannot lookup cached hashes when the cache is disabled (" + inputFile.key() + ").");
```

java-frontend/src/main/java/org/sonar/java/caching/ContentHashCache.java
```diff
@@ -0,0 +1,117 @@
...
+      } catch (IOException | NoSuchAlgorithmException e) {
+        LOG.error(String.format(HASH_COMPUTE_FAIL_MSG, inputFile.key()));
+      }
+    }
+    return false;
+  }
+
+  private void copyFromPrevious(InputFile inputFile) {
+    if (writeCache != null) {
+      LOG.trace("Coping cache from previous for file {}", inputFile.key());
```
dorian-burihabwa-sonarsource, line 109:
```suggestion
      LOG.trace("Copying cache from previous for file {}", inputFile.key());
```

java-frontend/src/main/java/org/sonar/java/caching/ContentHashCache.java
```diff
@@ -0,0 +1,117 @@
...
+    if (!enabled) {
+      LOG.trace("Cannot read from cache when disabled");
+      return false;
+    }
+    return readCache.contains(getCacheKey(inputFile));
+  }
+
+  public boolean writeToCache(InputFile inputFile) {
+    if (!enabled) {
+      LOG.trace("Cannot write on cache when disabled");
```
dorian-burihabwa-sonarsource, line 89:
```suggestion
      LOG.trace(() -> "Cannot write hashes to the cache when the cache is disabled (" + inputFile.key() + ").");
```

java-frontend/src/test/java/org/sonar/java/SonarComponentsTest.java
```diff
@@ -728,52 +729,44 @@ void fileCanBeSkipped_returns_false_when_the_file_is_a_generated_file() throws A
...
     when(sonarComponents.canSkipUnchangedFiles()).thenReturn(false);
     InputFile inputFile = mock(InputFile.class);
 
     assertThat(sonarComponents.fileCanBeSkipped(inputFile)).isFalse();
-    verify(inputFile, times(0)).status();
   }
 
   @Test
-  void fileCanBeSkipped_always_returns_true_when_skipUnchangedFiles_is_true_and_file_status_is_same() throws ApiMismatchException {
+  void fileCanBeSkipped_returns_false_when_inputFileStatusIsDifferentFromSame() throws ApiMismatchException {
```
dorian-burihabwa-sonarsource, line 754: Let's see if we need to adapt the logic of this test by first disabling the cache for `fileCanBeSkipped`to fallback on the input file status

dorian-burihabwa-sonarsource (changes requested):
Thanks for the great work, I think there is some fine-tuning needed so that we align with the other analyzers but we are almost there

java-frontend/src/main/java/org/sonar/java/caching/ContentHashCache.java
```diff
@@ -0,0 +1,122 @@
...
+      readCache = context.previousCache();
+      writeCache = context.nextCache();
+    }
+  }
+
+  public boolean hasSameHashCached(InputFile inputFile) {
+    if (!enabled) {
+      if (!inputFile.status().equals(InputFile.Status.SAME)) {
+        LOG.trace(() -> "Cache is disabled. File status is: " + inputFile.status() + ". File can be skipped.");
+        writeToCache(inputFile);
```
dorian-burihabwa-sonarsource, line 57: Let's remove this call to writeToCache

java-frontend/src/main/java/org/sonar/java/caching/ContentHashCache.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+    return readCache.contains(getCacheKey(inputFile));
+  }
+
+  public boolean writeToCache(InputFile inputFile) {
+    if (!enabled) {
+      LOG.trace(() -> "Cannot write hashes to the cache when the cache is disabled (" + inputFile.key() + ").");
+      return false;
+    }
+    if (writeCache != null) {
```
dorian-burihabwa-sonarsource, line 97: In practice, I think it is enough to check that the cache is enabled. Even if the cache is disabled, the documentation tells us that we should [not receive null from the sensorContext](https://github.com/SonarSource/sonar-plugin-api/blob/38fbf9cadc373fd5853e20e23dcc8cd6c3338123/plugin-api/src/main/java/org/sonar/api/batch/sensor/SensorContext.java#L241).

java-frontend/src/test/java/org/sonar/java/caching/ContentHashCacheTest.java
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  @Test
+  void contains_returns_false_whenCacheIsDisabled() {
+    ContentHashCache contentHashCache = new ContentHashCache(getSensorContextTesterWithEmptyCache(false));
+    Assertions.assertFalse(contentHashCache.contains(inputFile));
+  }
+
+  @Test
+  void writeToCache_returns_false_whenWriteCacheIsNull() {
```
dorian-burihabwa-sonarsource, line 141: If we remove the test in the main code, let's remove this test

dorian-burihabwa-sonarsource (changes requested):
Nice job on the fixes, I think that should be the last round of ping-pong. Feel free to test the logs in the ContehHashCacheTest tests if you feel it makes sense

java-frontend/src/main/java/org/sonar/java/caching/ContentHashCache.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  public boolean hasSameHashCached(InputFile inputFile) {
+    if (!enabled) {
+      if (inputFile.status() != InputFile.Status.SAME) {
+        LOG.trace(() -> "Cache is disabled. File status is: " + inputFile.status() + ". File can be skipped.");
+        return false;
+      }
+      LOG.trace(() -> "Cannot read cached hashes when the cache is disabled (" + inputFile.key() + ").");
+      return false;
```
dorian-burihabwa-sonarsource: Should we flip the logic so that we return `true` when the status is same?

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4318, opened 2023-02-22
Sample of code reproducing a crash in the division by zero check

gtoison:
This PR is a code sample causing the division by zero check to fail.
Follow up of the discussion started here: https://community.sonarsource.com/t/divisionbyzerocheck-error-with-and-switch-expression/82974

Maybe this comment in `ExplodedGraphWalker` gives a clue to the problem?
[      // FIXME this logic do not work with patterns from JDK17, as we need to consume the pattern adequately.
](https://github.com/SonarSource/sonar-java/blob/ef0fb9d219e24e8b158611657076eee5ec7aa936/java-symbolic-execution/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java#L533)


johann-beleites-sonarsource:
Fixed in #4380 (and #4386). Thanks for providing the reproducer here!

gtoison:
Thanks a lot for the fix, looking forward to the release!

PR #4319, opened 2023-02-23
Fix a unit test not passing on OS with French locale

henryju:


henryju:
Sorry, seems to have been fixed in master already

PR #4324, opened 2023-02-28
Make CognitiveComplexityMethodCheck more flexible

renoth:
- Add configuration for CognitiveComplexityMethodCheck to make ignored methods configurable (in addition to equals and hashCode)
- Add testcases
- Make MethodTreeUtils#isNamed public for use in CognitiveComplexityMethodCheck

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CognitiveComplexityMethodCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CognitiveComplexityMethodCheck.java
@@ -19,6 +19,9 @@
  */
 package org.sonar.java.checks;
 
+import java.util.Arrays;
+import java.util.List;
+
 import org.sonar.check.Rule;
 import org.sonar.check.RuleProperty;
 import org.sonar.java.ast.visitors.CognitiveComplexityVisitor;
@@ -26,9 +29,6 @@ import org.sonar.java.checks.helpers.MethodTreeUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.Tree;
-
-import java.util.Arrays;
-import java.util.List;
 
 @Rule(key = "S3776")
 public class CognitiveComplexityMethodCheck  extends IssuableSubscriptionVisitor {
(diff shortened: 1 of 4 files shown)


renoth:
Hello. any plans of merging or at least rejecting?

java-checks/src/main/java/org/sonar/java/checks/CognitiveComplexityMethodCheck.java
```diff
@@ -64,7 +70,15 @@ public void setMax(int max) {
...
+  public void setIgnoredMethods(String ignoredMethods) {
+    this.ignoredMethods = ignoredMethods;
+  }
+
+  private boolean isExcluded(MethodTree methodTree) {
+    return MethodTreeUtils.isEqualsMethod(methodTree) ||
+      MethodTreeUtils.isHashCodeMethod(methodTree) ||
+      Arrays.stream(ignoredMethods.split(";"))
+        .map(ignoredMethodName -> MethodTreeUtils.isNamed(methodTree, ignoredMethodName))
+        .reduce(Boolean.FALSE, (base, acc) -> base || acc);
```
kaufco, line 82: This should not be done every time the `isExcluded` method is called. Instead, a set `ignored` of ignored methods should be created from the `ignoredMethods` parameter only once, and `isExcluded` should then call `ignored.contains(methodName)`.

kaufco:
Hi @renoth , thanks for your contribution! 

Your suggestion (as implemented now) filters out any method of a given name, regardless of its defining class. So let's say you do not want to check `Foo.bar` because it is a complex method by intention; adding `bar` to the `ignoredMethods` parameter of the check would also skip `Box.bar` (as well as any other `bar` in the codebase of the project). From my point of view, that's not what is wanted and bears the risk of "accidentally" hiding methods from the check.

An alternative would be to have fully qualified method names for the `ignoredMethods` parameter (such as `"myPackage.Foo.bar"`), but then the idea of such a parameter would be pointless because you can already suppress the check with an annotation in the code or mark it as _won't fix_ in SQ.

renoth:
It was fully intentional to only consider the method name because of the following scenario:

Suppose you have a framework that uses generated methods that most of the time have high complexity. Instead of generating the SupressWarnings annotation in to every method you just give the name of the method to the rule and be done with it. You know that there can be no false positives because the name of the method is very unique.

I see your point though, although someone configuring the check parameters would most likely be knowing what he is doing and not enter possibly colliding method names.

PR #4329, opened 2023-03-10
Restrict ITs to run on 9.9 LTS

dorian-burihabwa-sonarsource:
Since the artifacts produced on this branch are only meant to run on SQ 9.9, we can limit cycle waste by avoiding to test against unrelated versions.


Wohops (approved):
LGTM!

PR #4330, opened 2023-03-13
SONARJAVA-4277 Rework issue message and use highlighting

Wohops:


java-checks/src/main/java/org/sonar/java/checks/MethodWithExcessiveReturnsCheck.java
```diff
@@ -64,6 +67,14 @@ public void visitNode(Tree tree) {
     }
   }
 
+  static void foo(@Nullable Object o) {
```
Wohops, line 70: I disagree. Please remove this.

leonardo-pilastri-sonarsource (approved):
Looks good! Nice job 👍

PR #4335, opened 2023-03-24
SONARJAVA-4434 New rule S6541

leonardo-pilastri-sonarsource:


java-frontend/src/main/java/org/sonar/java/metrics/MetricsScannerContext.java
```diff
@@ -0,0 +1,108 @@
...
+import org.sonar.java.ast.visitors.StatementVisitor;
+import org.sonar.plugins.java.api.tree.CompilationUnitTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+public interface MetricsScannerContext {
+
+  MetricsComputer getMetricsComputer();
+
+  class MetricsComputer {
```
dorian-burihabwa-sonarsource, line 41: I think we should extract this class to its own file

java-frontend/src/test/java/org/sonar/java/metrics/MetricsComputerTest.java
```diff
@@ -0,0 +1,99 @@
...
+    var methodNestingLevel = (Map<Integer, ?>) getValue("methodNestingLevel");
+    assertThat(methodNestingLevel).containsKey(methodTree.hashCode());
+    
+  }
+
+  Object getValue(String fname) throws NoSuchFieldException, SecurityException, IllegalArgumentException, IllegalAccessException {
+    Field field = mc.getClass().getDeclaredField(fname);
+    field.setAccessible(true);
+    return field.get(mc);
+  }
```
dorian-burihabwa-sonarsource, line 97: Maybe we would be better off by using a package-visible getter annotated with VisibleForTesting rather than playing with reflection

java-frontend/src/main/java/org/sonar/java/metrics/MetricsScannerContext.java
```diff
@@ -0,0 +1,108 @@
...
+    private Map<Integer, List<Tree>> methodComplexityNodes = new HashMap<>();
+    private Map<Integer, CognitiveComplexityVisitor.Result> methodComplexity = new HashMap<>();
+    private Map<Integer, Integer> compilationUnityComplexity = new HashMap<>();
+    private Map<Integer, Integer> methodNumberOfAccessedVariables = new HashMap<>();
+    private Map<Integer, Integer> treeLinesOfCode = new HashMap<>();
+    private Map<Integer, Integer> treeNumberOfStatements = new HashMap<>();
+    private Map<Integer, Integer> treeNumberOfCommentedLines = new HashMap<>();
+    private Map<Integer, Set<Integer>> treeNoSonarLines = new HashMap<>();
+    private Map<Integer, Integer> methodNestingLevel = new HashMap<>();
```
dorian-burihabwa-sonarsource, line 50: These maps could final

java-frontend/src/test/java/org/sonar/java/metrics/MetricsComputerTest.java
```diff
@@ -0,0 +1,99 @@
...
+        "      for(int i=0; i<3; i++) { if(i==2){ return null; } }" +
+        "      for(int i : new int[]{1,2}) { if(i==2){ return null; } }" +
+        "      return new Object();   " +
+        "    };" +
+        "  } " +
+        "}");
+    MethodTree methodTree = (MethodTree) ((ClassTree) cut.types().get(0)).members().get(0);
+    mc.getComplexityNodes(methodTree).size();
+    var complexityNodes = (Map<Integer, ?>) getValue("methodComplexityNodes");
+    assertThat(complexityNodes).containsKey(methodTree.hashCode());
```
dorian-burihabwa-sonarsource, line 57: I think it would be interesting to check that the `MetricsComputer` map does not contain information about the tree before calling it to ensure that data is created on the fly

dorian-burihabwa-sonarsource (changes requested):
Good work, I think we should pay more attention to the documentation of new utility classes and their usage but we are on the right track

java-checks/src/test/java/org/sonar/java/checks/BrainMethodCheckTest.java
dorian-burihabwa-sonarsource, line 1: We should add a test with the default value because if we cannot generate an example that triggers the rule, it means that we did not pick the right paramters

java-checks-test-sources/src/main/java/checks/BrainMethodCheck.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  void empty() {
+  }
+
+  public static native void alert(String msg) /* not JSNI comment */ /*-{
+  for (i=0;i<=5;i++) {
+  $wnd.alert(msg);
+}
+}-*/;
```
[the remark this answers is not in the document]
leonardo-pilastri-sonarsource: Not being familiar with `native`, I thought the commented part was necessary, removed!

java-checks/src/test/java/org/sonar/java/checks/BrainMethodCheckTest.java
```diff
@@ -0,0 +1,50 @@
...
+  @Test
+  void test() {
+    InternalCheckVerifier.newInstance()
+      .onFile(mainCodeSourcesPath("checks/BrainMethodCheck.java"))
+      .withChecks(new BrainMethodCheck())
+      .verifyIssues();
+  }
+  
+  @Test
+  void testLowerThresholds() {
```
dorian-burihabwa-sonarsource, line 39: Ideally, we should test a case where the values chosen for the parameters do not raise any issue

dorian-burihabwa-sonarsource (approved):

dorian-burihabwa-sonarsource (approved):

PR #4336, opened 2023-03-27
SONARJAVA-4433 Add check to count class coupling by counting imports

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/design/AbstractCouplingChecker.java
```diff
@@ -50,48 +42,20 @@
...
-  @RuleProperty(
-    key = "max",
-    description = "Maximum number of classes a single class is allowed to depend upon",
-    defaultValue = "" + DEFAULT_MAX)
-  public int max = DEFAULT_MAX;
-
-  private final Deque<Set<String>> nesting = new LinkedList<>();
-  private Set<String> types;
-  private JavaFileScannerContext context;
+  abstract void checkTypes(@Nullable Tree type);
```
dorian-burihabwa-sonarsource, line 51: Because checkTypes is invoked in the `visitVariable` method, we should document what is the expected behavior for an implementation of the method

java-checks/src/main/java/org/sonar/java/checks/design/ClassImportCouplingCheck.java
```diff
@@ -0,0 +1,152 @@
...
+      ExpressionTree expr = (ExpressionTree) type;
+      while (expr.is(Tree.Kind.MEMBER_SELECT)) {
+        MemberSelectExpressionTree mse = (MemberSelectExpressionTree) expr;
+        IdentifierTreeImpl identifierTree = (IdentifierTreeImpl) mse.identifier();
+        String name = identifierTree.binding.getName();
+        if (name.contains(packageName)) {
+          fullyQualifiedNameComponents.push(mse.identifier().name());
+        }
+        expr = mse.expression();
+      }
```
dorian-burihabwa-sonarsource, line 117: Just to be sure I understand this block: If the class we are visiting depends upon a type from a subpackage, we only add the name of the identifier?
```java
package com.acme;

class MyType {
  void doNothing() {
    com.acme.sub.package.OtherType unused = null;
  }  
}
```
In the case above, is `OtherType` or ` com.acme.sub.package.OtherType` added to `types`?

java-checks/src/main/java/org/sonar/java/checks/design/ClassImportCouplingCheck.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  private void addToTypes(IdentifierTree type, IPackageBinding packageBinding) {
+    if (packageBinding != null && packageBinding.getName().contains(packageName)) {
+      types.add(type.name());
+    }
+  }
+
+  private static String helper(Tree tree) {
```
dorian-burihabwa-sonarsource, line 135: We should rename this method to be more explicit in what it does

java-checks/src/test/java/org/sonar/java/checks/design/ClassImportCouplingCheckTest.java
dorian-burihabwa-sonarsource, line 1: We should add a test on the same file with a different value for the `couplingThreshold` parameter to see if it raises/does not raise on the expected value

java-checks/src/main/java/org/sonar/java/checks/design/AbstractCouplingChecker.java
dorian-burihabwa-sonarsource: Should the return type be changed to `Set<String>` and return the new types added to the field?

java-checks/src/main/java/org/sonar/java/checks/helpers/ClassPattternsUtils.java
```diff
@@ -0,0 +1,101 @@
...
+  private ClassPattternsUtils() {
+  }
+
+  public static boolean isPrivateInnerClass(ClassTree classTree) {
+    return !classTree.symbol().owner().isPackageSymbol() &&
+      ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.PRIVATE);
+  }
+
+  public static boolean isUtilityClass(ClassTree classTree) {
+    return hasOnlyStaticMembers(classTree) && !anonymousClass(classTree) && !extendsAnotherClassOrImplementsSerializable(classTree)
```
dorian-burihabwa-sonarsource, line 42: We can probably swap the first 2 tests since the first does not need to iterate through all the members
```suggestion
    return !anonymousClass(classTree) && hasOnlyStaticMembers(classTree) && !extendsAnotherClassOrImplementsSerializable(classTree)
```

dorian-burihabwa-sonarsource (changes requested):
Great work! Not much comments to add her

java-checks/src/main/java/org/sonar/java/checks/helpers/ClassPatternsUtils.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.List;
+import org.sonar.java.model.ModifiersUtils;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Modifier;
+import org.sonar.plugins.java.api.tree.ModifiersTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+public class ClassPatternsUtils {
```
dorian-burihabwa-sonarsource, line 31: Very minor issue: Some of the utility names methods below start with an action word, while others do not. Would it make sense to have something more consistent?

dorian-burihabwa-sonarsource (approved):
Let's rebase the branch on the [MMF branch](https://github.com/SonarSource/sonar-java/tree/MMF-3102) and we should be good to go

PR #4340, opened 2023-04-03
MMF-3102 Design rules for Java

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/BrainMethodCheck.java
```diff
@@ -97,4 +112,31 @@ private static boolean isExcluded(MethodTree methodTree) {
...
   }
 
+  @Override
+  public void endOfAnalysis(ModuleScannerContext context) {
+    if (issuesFound.size() > numberOfIssuesToReport) {
+      numberOfIssuesToReport += issuesFound.size() / 10;
+      issuesFound.sort((a, b) -> b.brainScore - a.brainScore);
+    } else {
+      numberOfIssuesToReport = issuesFound.size();
+    }
```
dorian-burihabwa-sonarsource, line 122: It looks like we are choosing to pick all issues up to 10 or the top 10% issues when the total exceeds 10.
That means if your project has 500 issues, you will report 50 of them which can still be overwhelming.
I feel like we should flip this around so that we have at most 10 issues even if these 10 do not cover all of the top 10%.

```
var numberOfIssuesToReport = Math.min(numberOfIssuesToReport, issuesFound.size());
issuesFound.stream()
  .sorted((a, b) -> b.brainScore - a.brainScore)
  .limit(numberOfIssuesToReport)
  .forEach(issue -> defaultContext.reportIssue(issueFound.analyzerMessage));
```

dorian-burihabwa-sonarsource (commented):
LGTM

leonardo-pilastri-sonarsource (approved):

PR #4344, opened 2023-04-05
Irina/sonarjava 4433

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/design/ClassImportCouplingCheck.java
```diff
@@ -73,34 +74,40 @@ public void visitClass(ClassTree tree) {
...
 
     if (imports == null) {
+      String fileProjectName = context.getProject().key();
+
       imports = compilationUnitTree.imports().stream()
         .filter(i -> !i.is(Tree.Kind.EMPTY_STATEMENT))
         .map(ImportTree.class::cast)
         .map(ImportTree::qualifiedIdentifier)
-        .map(i -> ExpressionsHelper.concatenate(((ExpressionTree) i)))
+        .filter(i -> ExpressionsHelper.concatenate(((ExpressionTree) i)).startsWith(fileProjectName))
```
dorian-burihabwa-sonarsource, line 83: I am not sure how to unit test this because project keys tend to always have the same value but we will have to see it live on peach to be sure

dorian-burihabwa-sonarsource (approved):
LGTM! Let's just update the ruling before merging

PR #4359, opened 2023-04-20
Add .mailmap file to sort out contributors with multiple emails

Wohops:
Set up a `.mailmap` document to clean up the contributor list using various emails when committing (including internal developers from Sonar).

You can compare the differences in results when executing the following command line, before and after the commit.
```
git shortlog -sne
```


leonardo-pilastri-sonarsource (approved):
Much neater now!

PR #4360, opened 2023-04-20
SONARJAVA-4369 Create new check for rule S6485

leonardo-pilastri-sonarsource:


java-checks/src/test/java/org/sonar/java/checks/KnownCapacityHashMapsCheckTest.java
```diff
@@ -0,0 +1,52 @@
...
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/KnownCapacityHashMapsCheckSample.java"))
+      .withCheck(new KnownCapacityHashMapsCheck())
+      .withJavaVersion(18)
+      .verifyNoIssues();
+    
+    CheckVerifier.newVerifier()
+    .onFile(mainCodeSourcesPath("checks/KnownCapacityHashMapsCheckSample.java"))
+    .withCheck(new KnownCapacityHashMapsCheck())
+    .verifyNoIssues();
```
dorian-burihabwa-sonarsource, line 49: We should document the default expected Java version here because when we look back in 2 years, we will need to dig deeper into the code to understand what the expectations were

java-checks/src/main/java/org/sonar/java/checks/KnownCapacityHashMapsCheck.java
```diff
@@ -0,0 +1,58 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+
+    if (!context.getJavaVersion().isJava19Compatible()) {
+      return;
+    }
+
+    NewClassTree newClassTree = (NewClassTree) tree;
+    if (MATCHERS.matches(newClassTree)) {
+      reportIssue(newClassTree, "Replace this call to the constructor with the better suited static method.");
```
dorian-burihabwa-sonarsource, line 54: Let's try and add a more specific issue message to set developers on the right path to a fix

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/KnownCapacityHashBasedCollectionCheck.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Map;
+import org.sonar.check.Rule;
+import org.sonar.java.JavaVersionAwareVisitor;
+import org.sonar.java.checks.methods.AbstractMethodDetection;
+import org.sonar.plugins.java.api.JavaVersion;
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
+import org.sonar.plugins.java.api.tree.NewClassTree;
+
+@Rule(key = "S6485")
+public class KnownCapacityHashBasedCollectionCheck extends AbstractMethodDetection implements JavaVersionAwareVisitor {
```
dorian-burihabwa-sonarsource, line 31: I think that because we are only interested in constructors we can override `nodesToVisit` to focus on these calls
```java
@Override
public List<Kind> nodesToVisit() {
  return Collections.singletonList(Kind.NEW_CLASS);
}
```

dorian-burihabwa-sonarsource (changes requested):
Thanks for the changes, It looks a lot better already! I think we can make the tests a little stronger by checking the exact location of the issues and focusing only on constructors

dorian-burihabwa-sonarsource (approved):

dorian-burihabwa-sonarsource (approved):

PR #4364, opened 2023-04-25
SONARJAVA-4464 Add new "sonar.java.enablePreview" analysis parameter

leonardo-pilastri-sonarsource:


dorian-burihabwa-sonarsource (changes requested):
Just a few small changes and we should be good to merge 👍🏿

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/CheckVerifier.java
```diff
@@ -114,13 +114,28 @@ static CheckVerifier newVerifier() {
...
+   * test files target explicitly a given version (eg. java 7), where a particular syntax/API has been introduced.
+   *
+   * @param javaVersionAsInt defines the java language syntax version to be considered during verification, provided as an integer.
+   * For instance, for Java 1.7, use '7'. For Java 12, simply '12'.
+   * 
+   * @param enablePreviewFeatures defines if preview features from the specified java version should be enabled or not
+   *
+   * @return the verifier configured to consider the provided test file(s) as following the syntax of the given java version
+   */
+  CheckVerifier withJavaVersion(int javaVersionAsInt, boolean enablePreviewFeatures);
```
Wohops, line 138: I'm not sure this is a good approach.
1. ECJ supports preview features only with its latest version of the language. So for us, we have the same behavior: the preview feature is supported when version == latest java version supported. In our case, 19. So allowing people to enable it for other than the latest is useless. It should not work. Did you test this scenario?
2. I believe it's not great to have a "flag" to enable the preview features. I would prefer two different methods. One without (implicitly), and one with the preview feature enabled.
3. Since only the latest version supports preview features, what about a new method `withPreviewFeatureEnabled()` without parameters? This would either force the latest version or check that the latest has been configured and crash if not, and eventually enable preview features.

java-frontend/src/main/java/org/sonar/java/model/JavaVersionImpl.java
```diff
@@ -42,19 +44,26 @@ public class JavaVersionImpl implements JavaVersion {
   public static final int MAX_SUPPORTED = JAVA_19;
 
   private final int javaVersion;
+  private final boolean previewFeaturesEnabled;
 
   public JavaVersionImpl() {
     this.javaVersion = -1;
+    this.previewFeaturesEnabled = false;
```
Wohops, line 51: what about calling the constructor instead?
```
this(-1, false);
```

java-frontend/src/main/java/org/sonar/java/model/JavaVersionImpl.java
```diff
@@ -42,19 +44,26 @@ public class JavaVersionImpl implements JavaVersion {
...
   }
 
   public JavaVersionImpl(int javaVersion) {
     this.javaVersion = javaVersion;
+    this.previewFeaturesEnabled = false;
```
Wohops, line 56: What about calling the constructor instead?
```
this(javaVersion, false);
```

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaSensor.java
```diff
@@ -184,10 +185,13 @@ private Iterable<InputFile> javaFiles(InputFile.Type type) {
   }
 
   private JavaVersion getJavaVersion() {
-    JavaVersion javaVersion = settings.get(JavaVersion.SOURCE_VERSION)
-      .map(JavaVersionImpl::fromString)
-      .orElse(new JavaVersionImpl());
-    LOG.info("Configured Java source version (" + JavaVersion.SOURCE_VERSION + "): " + javaVersion);
+    Map<String, String> javaVersionProperties = Map.of(
+      JavaVersion.ENABLE_PREVIEW, settings.get(JavaVersion.ENABLE_PREVIEW).orElseGet(() -> ""),
```
Wohops, line 189: there is much simpler to do here: 
```java
settings.getBoolean(JavaVersion.ENABLE_PREVIEW).orElse(false)
```

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaSensor.java
```diff
@@ -184,10 +185,13 @@ private Iterable<InputFile> javaFiles(InputFile.Type type) {
...
+      JavaVersion.SOURCE_VERSION, settings.get(JavaVersion.SOURCE_VERSION).orElseGet(() -> ""));
+    
+    JavaVersion javaVersion = JavaVersionImpl.fromMap(javaVersionProperties);
```
Wohops, line 192: Not sure this is the best way... For me, a constructor with both parameters doing the validation would be fine.

java-frontend/src/main/java/org/sonar/java/model/JavaVersionImpl.java
```diff
@@ -42,19 +44,26 @@ public class JavaVersionImpl implements JavaVersion {
...
+  }
+
+  public JavaVersionImpl(int javaVersion, boolean previewFeaturesEnabled) {
+    this.javaVersion = javaVersion;
+    this.previewFeaturesEnabled = previewFeaturesEnabled;
```
Wohops, line 61: This should only be possible on the latest supported version. In our case, if javaVersion >= 19.

Wohops (commented):
@leonardo-pilastri-sonarsource @dorian-burihabwa-sonarsource I did a second pass on your PR since I believe there is some misunderstanding on what is achievable with the parser. Can I let you have a look?

PR #4367, opened 2023-04-28
SONARJAVA-4464 Restrict the use of the sonar.java.enablePreview flag to the greatest supported Java version

dorian-burihabwa-sonarsource:


dorian-burihabwa-sonarsource:
This change comes as a result of the feedback on  #4364 (see [latest comments](https://github.com/SonarSource/sonar-java/pull/4364#pullrequestreview-1404433208)) and prevents users and developers from setting `sonar.java.enablePreview` flag to `true` when the java version number is lower than the maximum supported version

sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaSensor.java
```diff
@@ -185,11 +184,15 @@ private Iterable<InputFile> javaFiles(InputFile.Type type) {
   }
 
   private JavaVersion getJavaVersion() {
```
alban-auzeill, line 186: When `settings.get(JavaVersion.ENABLE_PREVIEW)` is empty we should return `new JavaVersionImpl()` without logs

java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/internal/InternalCheckVerifier.java
```diff
@@ -153,13 +153,19 @@ public InternalCheckVerifier withQuickFixes() {
...
-    return this;
+    return withJavaVersion(javaVersionAsInt, false);
   }
   
   @Override
   public InternalCheckVerifier withJavaVersion(int javaVersionAsInt, boolean enablePreviewFeatures) {
     requiresNull(javaVersion, "java version");
+    if (enablePreviewFeatures && javaVersionAsInt < JavaVersionImpl.MAX_SUPPORTED) {
+      var message = String.format(
+        "Preview features can only be enabled for the latest supported Java version (%d)",
```
alban-auzeill, line 164: `for the latest supported Java version (%d)` => `when the version >= latest supported Java version (>= %d)`

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #4372, opened 2023-05-02
Prepare for next development iteration 7.20.0-SNAPSHOT

alban-auzeill:


dorian-burihabwa-sonarsource (approved):
Thanks for the update 👍🏿

PR #4374, opened 2023-05-03
SONARJAVA-4286 Add support for ZERO and NON_ZERO constraints for MethodBehaviors

leonardo-pilastri-sonarsource:


dorian-burihabwa-sonarsource (changes requested):
Let's make sure we don't need to test the flows better

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4375, opened 2023-05-05
SONARJAVA-4423 Update rules metadata for symbolic-execution tag, add documentation and info log for disabled SE engine

leonardo-pilastri-sonarsource:


sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaSensorTest.java
```diff
@@ -277,6 +277,15 @@ void does_not_insert_SymbolicExecutionVisitor() throws IOException {
         "ParameterReassignedToCheck"
       );
   }
+  
+  @Test
+  void info_log_when_no_SE_rules_enabled() throws IOException {
+    Loggers.get(PerformanceMeasure.class).setLevel(LoggerLevel.INFO);
```
johann-beleites-sonarsource, line 283: `INFO` is the default, no need to set it explicitly. (And in cases where we do want to set it explicitly, why not use `logTester.setLevel()`?)

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaSensorTest.java
```diff
@@ -277,6 +277,15 @@ void does_not_insert_SymbolicExecutionVisitor() throws IOException {
...
+    assertJasperIsInvoked(new MapSettings());
+    String debugLogs = String.join("\n", logTester.logs(LoggerLevel.INFO));
+    assertThat(debugLogs).contains("No rules with 'symbolic-execution' tag were enabled,"
+      + " the Symbolic Execution Engine will not run during the analysis.");
```
johann-beleites-sonarsource, line 287: No need to get and join all log lines. You can simply do `assertThat(logTester.logs(LoggerLevel.INFO)).contains("My log message");`

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaSensorTest.java
```diff
@@ -277,6 +277,15 @@ void does_not_insert_SymbolicExecutionVisitor() throws IOException {
         "ParameterReassignedToCheck"
       );
   }
+
```
johann-beleites-sonarsource, line 280: I believe there is a little left over whitespace here.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S2589.json
```diff
@@ -7,9 +7,7 @@
     "constantCost": "10min"
   },
   "tags": [
-    "cwe",
-    "suspicious",
-    "redundant"
```
johann-beleites-sonarsource, line 12: Why are we removing these tags?

johann-beleites-sonarsource (approved):
LGTM!

PR #4376, opened 2023-05-08
BUILD-2700 Update gh-action_dogfood_merge version

sebastienvermeille:
# BUILD-2700 Update gh-action_dogfood_merge version

## Changes
* Use gh-action_dogfood_merge tag `v1` (stable) instead of master
  That way it no longer relies on master (unsafe) (At the moment they are in sync but for future changes, it is better that way)


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4377, opened 2023-05-08
SONARJAVA-4097 Stop method SE at the first unknown method call symbol

alban-auzeill:


leonardo-pilastri-sonarsource (changes requested):

leonardo-pilastri-sonarsource (approved):
Nice work!

PR #4379, opened 2023-05-10
Download specific JDK to build java-checks-test-sources as part of AutoScan test

dorian-burihabwa-sonarsource:


dorian-burihabwa-sonarsource:
This topic will be revisited with a more sustainable approach in the future

PR #4380, opened 2023-05-10
SONARJAVA-4420 Fix IndexOutOfBoundsException on switch expression

johann-beleites-sonarsource:
Reduced approximation involved in cleaning up (clearing the stack). Instead, we now try to take off only what we should from the stack. Also, exceptions that are the cause for a path to be walked are not stored on the stack anymore but instead have a dedicated field in the ProgramState.


java-symbolic-execution/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
```diff
@@ -910,7 +908,12 @@ private void enqueueExceptionalPaths(ProgramState ps, Symbol methodSymbol, @Null
         .orElse(exitBlock);
       enqueue(new ProgramPoint(methodExit), ps, true, methodYield);
     } else {
-      otherBlocks.forEach(b -> enqueue(new ProgramPoint(b), ps, true, methodYield));
+      var exception = ps.peekValue();
```
alban-auzeill, line 911: Why do we peek a value from the stack instead of using the above `exceptionSV` variable?
And why not changing the way we initialize the above `exceptionSV` by using `.unstackValue(1)` instead of `peekVallue()`, and then always initialize `withEntryException` at the beginning of this method to benefit to following program states?

I would also not only set `withEntryException` in this `enqueueExceptionalPaths` method but also in `ExplodedGraphWalker#handleBlockExit` / `case THROW_STATEMENT:`

alban-auzeill (changes requested):

java-symbolic-execution/src/main/java/org/sonar/java/se/ExplodedGraphWalker.java
johann-beleites-sonarsource: So I improved it a little. Unfortunately, I can't change how exceptions are handled in more parts of the engine without causing issues with flows. My proposal here would be to leave it as it is for now and in future do a sprint on proper stack and exception handling, where we can git rid of some of the hacks that are currently deeply-embedded in the engine.

alban-auzeill (approved):

PR #4381, opened 2023-05-10
SONARJAVA-2126 Unboxing of NULL primitive wrapper raises NPE and should be detected by S2259

dorian-burihabwa-sonarsource:


java-symbolic-execution/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -64,6 +70,16 @@ public class NullDereferenceCheck extends SECheck {
...
     .addParametersMatcher("java.util.function.Supplier").build();
 
+  private static final Set<String> BOXED_TYPES = Set.of(
+    "java.lang.Boolean",
+    "java.lang.Byte",
+    "java.lang.Character",
+    "java.lang.Float",
+    "java.lang.Double",
+    "java.lang.Integer",
+    "java.lang.Long"
```
leonardo-pilastri-sonarsource, line 80: I think we missed `"java.lang.Short"` here.
```
  boolean shortComparison() {
    Short l = null;
    short i = 1;
    return i == l; // FN
  }
```

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/NullDereferenceCheck.java
```diff
@@ -120,12 +136,117 @@ public ProgramState checkPreStatement(CheckerContext context, Tree syntaxNode) {
...
+        if (syntaxNode instanceof AssignmentExpressionTree) {
+          inspectAssignment(context, (AssignmentExpressionTree) syntaxNode, peekValue);
+        } else if (syntaxNode instanceof BinaryExpressionTree) {
+          inspectBinaryExpression(context, (BinaryExpressionTree) syntaxNode, peekValue);
+        } else if (syntaxNode instanceof UnaryExpressionTree) {
+          var unaryOperation = (UnaryExpressionTree) syntaxNode;
+          extractUnboxingExpression(unaryOperation).ifPresent(expression ->
+            checkConstraint(context, expression, peekValue)
+          );
+        }
```
leonardo-pilastri-sonarsource, line 155: I am not sure if this might be out of scope for this PR, but we're not detecting this case:
```
  int illegalReturn() {
    Integer l = null;
    return l;
  }
```
We might be able to catch it by adding here the switch case and its relative logic for `Tree.Kind.RETURN_STATEMENT` .
WDYT?

leonardo-pilastri-sonarsource (changes requested):
Very nice job! Just a couple hints for improvement

leonardo-pilastri-sonarsource (approved):
Looks good!

PR #4382, opened 2023-05-10
SONARJAVA-4475 Add visitIdentifier to DivisionByZeroCheck.PostStatementVisitor to evaluate constants outside method's scope

leonardo-pilastri-sonarsource:


dorian-burihabwa-sonarsource (changes requested):
The change look good overall. Following our discussion, let's try to:
- See if using `isEffectivelyFinal` is more interesting than relying on an explicit `final` declaration
- Add the non-compiling test case to a test source file related to S2589
- Format the code of the check so that spaces are consistent
- See if we can remove the mocked test by testing a non-compiling bit of code
```java
static final int constant = 0;
  static void alwaysTrue() {
    //Fixed, now ZERO constraint correctly applies to identifiers out of method scope
    if (unknown == 0) { System.out.println("Always true"); } // Compliant
  }
```

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4385, opened 2023-05-22
SONARJAVA-4442 S3518 DivisionByZeroCheck reports misleading secondary locations

alban-auzeill:


java-symbolic-execution/src/main/java/org/sonar/java/se/FlowComputation.java
```diff
@@ -68,12 +66,26 @@
...
+    Tree.Kind.PLUS,
+    Tree.Kind.MINUS,
+    Tree.Kind.MULTIPLY,
+    Tree.Kind.DIVIDE,
+    Tree.Kind.REMAINDER,
+    Tree.Kind.PLUS_ASSIGNMENT,
+    Tree.Kind.MINUS_ASSIGNMENT,
+    Tree.Kind.MULTIPLY_ASSIGNMENT,
+    Tree.Kind.DIVIDE_ASSIGNMENT,
+    Tree.Kind.REMAINDER_ASSIGNMENT);
```
johann-beleites-sonarsource, line 88: Do we have tests that cover the X_ASSIGNMENT types? Do they also need to be handled in the `ExplodedGraphWalker`?

johann-beleites-sonarsource (approved):
LGTM!

PR #4390, opened 2023-05-25
SONARJAVA-2126 Fix NPEs when looking up symbolic values of volatile fields

dorian-burihabwa-sonarsource:


johann-beleites-sonarsource (approved):
LGTM!

PR #4395, opened 2023-06-08
mg/SONARJAVA-4506

Wohops:
Because the rule has a non-numeric key, it is not automatically updated by rule-api. In order to update the rule, you must first generate the rule metadata and then overwrite the data to have it in the file that matches the non-numeric key.

    java -jar ../rule-api-....jar generate -rule S1291
    mv java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1291.html java-checks/src/main/resources/org/sonar/l10n/java/rules/java/NoSonar.html
    mv java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S1291.json java-checks/src/main/resources/org/sonar/l10n/java/rules/java/NoSonar.json


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4397, opened 2023-06-12
Upgrade JaCoCo to 0.8.10

Godin:
JaCoCo 0.8.7 supports only class file versions up to Java 17, whereas JaCoCo 0.8.7 supports up to Java 20,
so after this change following messages won't appear in CI that uses Java 19

    java.lang.instrument.IllegalClassFormatException: Error while instrumenting javax/lang/model/SourceVersion.
        at org.jacoco.agent.rt.internal_3570298.CoverageTransformer.transform(CoverageTransformer.java:94)
    ...
    Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 63
        at org.jacoco.agent.rt.internal_3570298.asm.ClassReader.<init>(ClassReader.java:196)
    ...

Seems that this was overlooked in
68b8905d6ed9dd64212d1ea59d35500a160b7b68


chrislain-razafimahefa-sonarsource (approved):
LGTM!

PR #4398, opened 2023-06-12
SONARJAVA-4518 Introduce a new IssueFilter targeting Spring

Wohops:


java-checks/src/main/java/org/sonar/java/checks/MethodOnlyCallsSuperCheck.java
```diff
@@ -46,8 +46,7 @@
 @Rule(key = "S1185")
 public class MethodOnlyCallsSuperCheck extends IssuableSubscriptionVisitor {
 
-  private static final List<String> TRANSACTIONAL_ANNOTATIONS = Arrays.asList("javax.transaction.Transactional",
-    "org.springframework.transaction.annotation.Transactional");
+  private static final List<String> TRANSACTIONAL_ANNOTATIONS = Arrays.asList("javax.transaction.Transactional");
```
leonardo-pilastri-sonarsource, line 49:
```suggestion
  
```
We could remove this constant and instead at line 169 have a simple 
`return metadata.isAnnotatedWith("javax.transaction.Transactional");`

java-checks/src/main/java/org/sonar/java/filters/SpringFilter.java
```diff
@@ -0,0 +1,139 @@
...
+      // missing semantic
+      || hasUnknownAnnotationWithName(tree.modifiers(), "Transactional");
+  }
+
+  private static boolean hasUnknownAnnotationWithName(ModifiersTree modifiers, String annotation) {
+    // Token based check in case of missing semantic
+    List<AnnotationTree> annotations = modifiers.annotations();
+    if (annotations.isEmpty()) {
+      return false;
+    }
```
leonardo-pilastri-sonarsource, line 120:
```suggestion
```
Semantically we could remove this condition, and the final `anyMatch` would return false for empty lists of annotations, but I'm wondering if you put the condition to prevent the creation of `Stream` objects for every method with no annotations, so I don't have a strong opinion here

java-checks/src/main/java/org/sonar/java/filters/SpringFilter.java
```diff
@@ -0,0 +1,139 @@
...
+      .anyMatch(annotationName -> annotationName.endsWith(annotation));
+  }
+
+  private void excludeLinesIfTrue(boolean shouldExclude, Tree tree, Class<? extends JavaCheck> rule) {
+    if (shouldExclude) {
+      excludeLines(tree, rule);
+    } else {
+      acceptLines(tree, rule);
+    }
+  }
```
leonardo-pilastri-sonarsource, line 137: I noticed that `LombokFilter` has the same method, plus one that takes more than one `JavaCheck` as varargs.
I suggest we move both methods in `LombokFilter` inside `BaseTreeVisitorIssueFilter`, as they can be shared between `SpringFilter` and `LombokFilter` for now, and could also benefit future filters.

So I would remove these two methods from `LombokFilter`:
```
  @SafeVarargs
  private final void excludeLinesIfTrue(boolean shouldExclude, Tree tree, Class<? extends JavaCheck>... rules) {
    for (Class<? extends JavaCheck> rule : rules) {
      excludeLinesIfTrue(shouldExclude, tree, rule);
    }
  }

  private void excludeLinesIfTrue(boolean shouldExclude, Tree tree, Class<? extends JavaCheck> rule) {
    if (shouldExclude) {
      excludeLines(tree, rule);
    } else {
      acceptLines(tree, rule);
    }
  }
```
move them inside `BaseTreeVisitorIssueFilter` and change their visibility to `protected`. Then also remove the `excludeLinesIfTrue` from `SpringFilter`

leonardo-pilastri-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/filters/SpringFilter.java
Wohops: Indeed, I wanted to avoid using streams, as I believe that a large majority of methods are not annotated. I'll add a comment.

java-checks/src/main/java/org/sonar/java/filters/SpringFilter.java
Wohops: good point.

PR #4399, opened 2023-06-13
SONARJAVA-4476 FP for S2629 on record getter methods

ADarko22:


alban-auzeill (approved):

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -288,6 +287,8 @@ private static class StringExpressionVisitor extends BaseTreeVisitor {
 
     @Override
     public void visitMethodInvocation(MethodInvocationTree tree) {
+      // assuming that tree.methodSymbol().returnType().isUnknown() is always false,
+      // because findStringArg() implicitly filters out unknown types due to symbolType() being null
       if (!isGetter(tree) && !isAnnotationMethod(tree)) {
```
ADarko22, line 292: I removed the check we had in the previous PR:
```java
tree.methodSymbol().returnType().isUnknown() 
``` 
because it couldn't be covered by any test since unknown types generate a `MethodInvocationTree` where the arguments have a null `SymbolType`: those trees are filtered out by `findStringArg(MethodInvocationTree)`; the new logic never gets executed in such cases.
@alban-auzeill

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -296,7 +297,18 @@ public void visitMethodInvocation(MethodInvocationTree tree) {
...
     private static boolean isGetter(MethodInvocationTree tree) {
       String methodName = tree.methodSymbol().name();
-      return methodName != null && (methodName.startsWith("get") || methodName.startsWith("is"));
+      return methodName.startsWith("get") || methodName.startsWith("is") || isGetterMatchingFieldNameAndType(tree.methodSymbol());
+    }
+
+    private static boolean isGetterMatchingFieldNameAndType(Symbol.MethodSymbol methodSymbol) {
+      Symbol owner = methodSymbol.owner();
+      Type getterReturnType = methodSymbol.returnType().type();
+      return methodSymbol.parameterTypes().isEmpty()
```
ADarko22, line 306: The `methodSymbol.parameterTypes().isEmpty()` is used to check that the candidate getter has no arguments, anyway such a check is not performed in the existing logic. Should we enforce it, and clean the possible hits on the ITs, or should we leave the existing logic and accept those FNs?
@alban-auzeill

its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json
```diff
@@ -1334,7 +1334,7 @@
   {
     "ruleKey": "S2629",
     "hasTruePositives": true,
-    "falseNegatives": 27,
+    "falseNegatives": 58,
```
ADarko22, line 1337: After splitting the `LazyArgEvaluationCheck.java` resource in compiling and not-compiling code, and moving it to `java-checks-tests-sources` the compiling resource is analyzed by autoscan and more FN are found!
Namely, all the Noncompliant code is reported; also other rules are hitting on `LazyArgEvaluationCheck.java` resource

PR #4400, opened 2023-06-13
Fix FP on S1149 on overriding methods body

leonardo-pilastri-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -130,18 +131,34 @@ public void visitClass(ClassTree tree) {
     @Override
     public void visitMethod(MethodTree tree) {
       TypeTree returnTypeTree = tree.returnType();
-      if (isNotOverriding(tree) || returnTypeTree == null) {
+      populateOverridingMethodExclusions(tree, returnTypeTree);
+      if (!isOverriding(tree) || returnTypeTree == null) {
```
dorian-burihabwa-sonarsource, line 135: Why is data implicitly pushed to the stack but explicitly popped?
```suggestion
      overridingMethodTypes.push(collectTypesFromOverridingMethodSignature(tree, returnTypeTree));
      if (!isOverriding(tree) || returnTypeTree == null) {
```

java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -130,18 +131,34 @@ public void visitClass(ClassTree tree) {
...
+        if(returnType != null) {
+          methodTypes.add(returnType.symbolType().fullyQualifiedName());          
+        }
+        methodTree.parameters().stream()
+          .filter(param -> param.type() != null)
+          .map(param -> param.type().symbolType().fullyQualifiedName())
+          .forEach(methodTypes::add);
+      }
+      overridingMethodTypes.push(methodTypes);
+    }
```
dorian-burihabwa-sonarsource, line 157:
```suggestion
    HashSet<String>() collectTypesFromOverridingMethodSignature(MethodTree methodTree, TypeTree returnType) {
      var methodTypes = new HashSet<String>();
      if (isOverriding(methodTree)) {
        if (returnType != null) {
          methodTypes.add(returnType.symbolType().fullyQualifiedName());          
        }
        methodTree.parameters().stream()
          .filter(param -> param.type() != null)
          .map(param -> param.type().symbolType().fullyQualifiedName())
          .forEach(methodTypes::add);
      }
      return methodTypes;
    }
```

java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -130,18 +131,34 @@ public void visitClass(ClassTree tree) {
...
           reportIssueOnDeprecatedType(returnTypeTree, returnTypeTree.symbolType());
         }
         scan(tree.parameters());
       }
       scan(tree.block());
+      overridingMethodTypes.pop();
     }
 
-    private boolean isNotOverriding(MethodTree tree) {
+    void populateOverridingMethodExclusions(MethodTree methodTree, TypeTree returnType) {
```
dorian-burihabwa-sonarsource, line 145:
```suggestion
    void populateOverridingMethodExclusions(MethodTree methodTree, @Nullable TypeTree returnType) {
```

java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -130,18 +131,34 @@ public void visitClass(ClassTree tree) {
...
+      var methodTypes = new HashSet<String>();
+      if (isOverriding(methodTree)) {
+        if(returnType != null) {
+          methodTypes.add(returnType.symbolType().fullyQualifiedName());          
+        }
+        methodTree.parameters().stream()
+          .filter(param -> param.type() != null)
```
dorian-burihabwa-sonarsource, line 152: Do we know in which case a parameter, which is `VariableTree`, return null on `type()`?

java-checks/src/main/java/org/sonar/java/checks/SynchronizedClassUsageCheck.java
```diff
@@ -130,18 +131,34 @@ public void visitClass(ClassTree tree) {
...
+      overridingMethodTypes.pop();
     }
 
-    private boolean isNotOverriding(MethodTree tree) {
+    void populateOverridingMethodExclusions(MethodTree methodTree, TypeTree returnType) {
+      var methodTypes = new HashSet<String>();
+      if (isOverriding(methodTree)) {
+        if(returnType != null) {
+          methodTypes.add(returnType.symbolType().fullyQualifiedName());          
+        }
```
dorian-burihabwa-sonarsource, line 150: Actually, we can probably cut some branches in this method by running some of the tests in this method in `visitMethod`.
Because we only collect types for overriding methods and we test it twice between the 2 methods.
But also because, according to the Javadoc, `returntype` can only be null for constructors and constructors can be overloaded but not overriden

dorian-burihabwa-sonarsource (changes requested):

dorian-burihabwa-sonarsource (approved):

PR #4403, opened 2023-06-14
SONARJAVA-4497 FP on S2142 when using union types and rethrowing

ADarko22:


java-checks-test-sources/src/main/java/checks/InterruptedExceptionCheck.java
```diff
@@ -450,13 +450,27 @@ public void throwInterruptedExceptionWithinThrowStatement() throws InterruptedEx
     }
   }
 
+
+  public void catchUnionType1() throws IOException, InterruptedException {
+    try {
+      getRuntimeExceptionThrowsInterruptedException();
+      getRuntimeExceptionThrowsIOException();
+    } catch (InterruptedException | IOException e) { // Compliant - FP from S2142
```
alban-auzeill, line 458:
```suggestion
    } catch (InterruptedException | IOException e) { // Compliant - FP from SONARJAVA-4497
```

alban-auzeill (approved):

PR #4410, opened 2023-06-21
Angelo/update jdt

ADarko22:


java-checks/src/test/java/org/sonar/java/checks/InstanceOfPatternMatchingCheckTest.java
```diff
@@ -30,21 +30,12 @@ class InstanceOfPatternMatchingCheckTest {
...
   void test() {
     CheckVerifier.newVerifier()
+      .withJavaVersion(20, true)
       .onFile(testSourcesPath("checks/InstanceOfPatternMatching.java"))
       .withCheck(new InstanceOfPatternMatchingCheck())
-      .withJavaVersion(16)
       .verifyIssues();
   }
 
-  @Test
```
ADarko22, line 39: it seems that these features require Preview enabled, so I don't think it works on previous, thus I'm removing the test

PR #4411, opened 2023-06-23
SONARJAVA-4526  Update Java parser version to ECJ 3.34.0

alban-auzeill:


java-checks/src/main/java/org/sonar/java/checks/KnownCapacityHashBasedCollectionCheck.java
```diff
@@ -61,7 +61,7 @@ protected MethodMatchers getMethodInvocationMatchers() {
 
   @Override
   public boolean isCompatibleWithJavaVersion(JavaVersion version) {
-    return version.isJava19Compatible();
+    return version.isJava20Compatible();
```
alban-auzeill, line 64: We should not change the above line. This rule tries to enforce new Java methods introduced in Java 19, not 20.
Like [HashMap#newHashMap(int)](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/HashMap.html#newHashMap(int))

sonar-java-plugin/pom.xml
```diff
@@ -166,7 +166,7 @@
             <configuration>
               <shadedArtifactAttached>false</shadedArtifactAttached>
               <createDependencyReducedPom>false</createDependencyReducedPom>
-              <minimizeJar>true</minimizeJar>
+              <minimizeJar>false</minimizeJar>
```
ADarko22, line 169: we should revert the flag to true and clean the filters: find the dependency with the Class that is not found and remove all the others that are not necessary but included transitively

ADarko22 (changes requested):

alban-auzeill:
We are closing this PR without merging it because:
* ECJ 3.34 still has a bug and fails to parse some "when" identifiers. It was the main motivation to update.
* ECJ 3.34 now requires Java 17. It contradicts our scanner context requirements (Java 11) and during this sprint, we learned that there are no easy and reliable solutions to convert this dependency to target Java 11.

For more details see [SONARJAVA-4526](https://sonarsource.atlassian.net/browse/SONARJAVA-4526)

alban-auzeill:
We are closing this PR without merging it because:
* ECJ 3.34 still has a bug and fails to parse some "when" identifiers. It was the main motivation to update.
* ECJ 3.34 now requires Java 17. It contradicts our scanner context requirements (Java 11) and during this sprint, we learned that there are no easy and reliable solutions to convert this dependency to target Java 11.

For more details see [SONARJAVA-4526](https://sonarsource.atlassian.net/browse/SONARJAVA-4526)

PR #4416, opened 2023-07-17
SONARJAVA-4444 Upgrade sonar-plugin-api to 10.0.0.695

alban-auzeill:
⚠️ Warning: We should keep commits separated by merging using `Rebase` instead of `Squash`.


java-frontend/src/main/java/org/sonar/java/model/JParser.java
irina-batinic-sonarsource, line 1: Should we keep imports like this?

irina-batinic-sonarsource (approved):
LGTM!

PR #4429, opened 2023-08-01
SONARJAVA-4007, SONARJAVA-4401, SONARJAVA-1186 FPs in S1120

kaufco:


irina-batinic-sonarsource (commented):
Hey,
What about the cases:
```
case ORANGE-> System.out.println("orange");
case LIME->System.out.println("lime");
```
Do we think of them as Noncompliant?

java-checks/src/test/files/checks/IndentationCheck_default.java
```diff
@@ -1,5 +1,176 @@
...
+        {
+          } // Noncompliant
+      case BEIGE:
+          { // Noncompliant
+          }
+    }
+  }
+
+  public static void testBlockIndentationArrowCases(Color color) {
+    switch(kind) {
```
irina-batinic-sonarsource, line 102: Same question as previous one.

kaufco:
You're right, that's not compliant formatting, but not a matter of this rule. This rule checks only the indentation:

- when a block or statement is on a new line: whether the line indentation matches
- when the first statement of a new block does not break lines: whether the distance to the previous token on the same line matches

I think we also have no other rule that checks for a space between the case expression and `->` since these formatting rules are comparatively old.

java-checks/src/test/files/checks/IndentationCheck_default.java
```diff
@@ -1,5 +1,176 @@
...
 
+class TestSwitch {
+  enum Color {
+    RED, GREEN, BLUE, CYAN, MAGENTA, YELLOW, BLACK, WHITE,
+    ORANGE, BROWN, LIME, PURPLE, GREY, CRIMSON, NAVY, OLIVE,
+    FUCHSIA, MAROON, AQUA, TEAL, SILVER, AZURE, BEIGE
+  }
+
+  public static void testArrowCases(Color color) {
+    switch(kind) {
```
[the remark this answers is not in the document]
kaufco: Fixed.

java-checks/src/test/files/checks/IndentationCheck_default.java
kaufco: Fixed.

PR #4430, opened 2023-08-02
SONARJAVA-4519 FP on S100 when using OSGI annotation `AttributeDefinition`

kaufco:


java-checks/src/main/java/org/sonar/java/checks/naming/BadMethodNameCheck.java
```diff
@@ -57,16 +57,18 @@ public void setContext(JavaFileScannerContext context) {
...
     }
   }
 
-  private static boolean isNotOverriden(MethodTree methodTree) {
-    return Boolean.FALSE.equals(methodTree.isOverriding());
+  private boolean isExcluded(MethodTree methodTree) {
+    return !Boolean.FALSE.equals(methodTree.isOverriding()) || methodTree.modifiers().annotations().stream().anyMatch(
+      // use simple name, so we have no FP on missing semantics
+      it -> "AttributeDefinition".equals(it.annotationType().symbolType().name())
+    );
```
dorian-burihabwa-sonarsource, line 72: How about extracting the annotation check to its own method for readability?
```suggestion
  private static boolean isExcluded(MethodTree methodTree) {
    return !Boolean.FALSE.equals(methodTree.isOverriding()) ||
      isExcludedByAnnotation(methodTree);
  }
  
  private static boolean isExcludedByAnnotation(methodTree) {
    return methodTree.modifiers().annotations().stream().anyMatch(
      // use simple name, so we have no FP on missing semantics
      it -> "AttributeDefinition".equals(it.annotationType().symbolType().name())
    );
```

dorian-burihabwa-sonarsource (changes requested):
Looks good, let's see if we refactor for readability and move the test file to its own compiling file

java-checks/src/test/files/checks/naming/BadMethodName.java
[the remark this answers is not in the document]
kaufco: I thought about it but put it into the existing (non-compiling) file intentionally because otherwise, we would have to add the OSGI framework to the java-checks-test-sources dependencies. I think it's not worth it, even more so when the non-compiling file already exists. 

Plus: the more important test case is that it's not generating an FP when there is no semantics. So we would need that non-compiling test even if we'd added another compiling one.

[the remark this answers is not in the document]
dorian-burihabwa-sonarsource: I get your point but it is also easier to test a compiling file twice, with and without semantics, than maintaining a non-compiling file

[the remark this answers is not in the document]
kaufco: Fixed.

java-checks/src/test/java/org/sonar/java/checks/naming/BadMethodNameCheckTest.java
```diff
@@ -21,23 +21,33 @@
...
+  void testWithoutSemantic() {
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/BadMethodName.java"))
+      .withCheck(new BadMethodNameCheck())
+      .withoutSemantic()
+      .verifyIssues();
+  }
+
   @Test
   void test2() {
```
dorian-burihabwa-sonarsource, line 46: Can we rename this test to be more indicative of what they are testing? Or perhaps split the 2 tests in their own test case? I see one test raising issues and the other none. But what is the implication of one failing and not the other?

java-checks/src/main/java/org/sonar/java/checks/naming/BadMethodNameCheck.java
```diff
@@ -57,16 +57,22 @@ public void setContext(JavaFileScannerContext context) {
...
     MethodTree methodTree = (MethodTree) tree;
-    if (isNotOverriden(methodTree) && !pattern.matcher(methodTree.simpleName().name()).matches()) {
+    if (!isExcluded(methodTree) && !pattern.matcher(methodTree.simpleName().name()).matches()) {
       reportIssue(methodTree.simpleName(), "Rename this method name to match the regular expression '" + format + "'.");
     }
   }
 
-  private static boolean isNotOverriden(MethodTree methodTree) {
-    return Boolean.FALSE.equals(methodTree.isOverriding());
+  private boolean isExcluded(MethodTree methodTree) {
```
dorian-burihabwa-sonarsource, line 68: This method can be `static`
```suggestion
  private static boolean isExcluded(MethodTree methodTree) {
```

java-checks/src/main/java/org/sonar/java/checks/naming/BadMethodNameCheck.java
```diff
@@ -57,16 +57,22 @@ public void setContext(JavaFileScannerContext context) {
...
+    return !Boolean.FALSE.equals(methodTree.isOverriding()) || isExcludedByAnnotation(methodTree);
+  }
+
+  private boolean isExcludedByAnnotation(MethodTree methodTree) {
```
dorian-burihabwa-sonarsource, line 72: This method can be `static`
```suggestion
  private static boolean isExcludedByAnnotation(MethodTree methodTree) {
```

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/naming/BadMethodNameCheck.java
kaufco: Fixed.

java-checks/src/test/java/org/sonar/java/checks/naming/BadMethodNameCheckTest.java
kaufco: Fixed. Changed the name to be more descriptive.

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4431, opened 2023-08-02
SONARJAVA-4411 Fix FP when URI is a pattern

irina-batinic-sonarsource:


kaufco (changes requested):
LGTM, but let's see if we can increase the test coverage. An additional test string that contains `$` but not `*` should do it (due to boolean shortcut in line 139)

kaufco (approved):
Perfect!

PR #4432, opened 2023-08-02
SONARJAVA-4556 Add a language to the PMD Sensor descriptor

kaufco:


external-reports/src/test/java/org/sonar/java/externalreport/PmdSensorTest.java
```diff
@@ -62,6 +62,7 @@ void test_descriptor() {
     DefaultSensorDescriptor sensorDescriptor = new DefaultSensorDescriptor();
     sensor.describe(sensorDescriptor);
     assertThat(sensorDescriptor.name()).isEqualTo("Import of PMD issues");
+    assertThat(sensorDescriptor.languages()).isEqualTo(List.of("java"));
```
dorian-burihabwa-sonarsource, line 65: Because we set the only value to be java, but we don't know what kind of data structure holds the information, it might be safer to remove the list type from the test
```suggestion
    assertThat(sensorDescriptor.languages()).containsOnly("java");
```

dorian-burihabwa-sonarsource (approved):
The changes look good to me. I would make a small suggestion for the test to make sure that it does not suffer from future implementation changes

PR #4433, opened 2023-08-02
SONARJAVA-4467 Add method to check if there is a comment above the me…

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/FieldModifierCheck.java
```diff
@@ -44,11 +45,23 @@ public void visitNode(Tree tree) {
...
+        if (!hasModifierComment(variableTree)) {
+          IdentifierTree simpleName = variableTree.simpleName();
+          reportIssue(simpleName, "Explicitly declare the visibility for \"" + simpleName.name() + "\".");
+        }
       });
   }
 
+  private static boolean hasModifierComment(VariableTree variableTree) {
+    for (SyntaxTrivia syntaxTrivia : variableTree.type().lastToken().trivias()) {
+      if (syntaxTrivia.comment().contains("modifier")) {
```
kaufco, line 58: 1. Maybe we should ignore upper / lower case? Because "Modifier" might be the first word of a sentence.

```suggestion
      if (syntaxTrivia.comment().toLowerCase().contains("modifier")) {
```

2. The current code also hits if "modifier" is not a separate word, like in "FooBarmodifierxyz". I don't see it as a problem though, we can just exclude these cases, too.

java-checks/src/test/files/checks/FieldModifier.java
```diff
@@ -1,4 +1,17 @@
...
+  int a; // Noncompliant
+
+  // modifier
+  // comment
+  int c; // Compliant
+
+  // comment2
+  // comment
+  int d; // Noncompliant
+
```
kaufco, line 14: 1. We could add one more test case to show that things like `// FooBarmodifierxyz` are also compliant
2. If you chose to merge the above suggestion about ignoring case sensitivity, we also should add a testcase like `// Modifier for visibility intentionally ommitted.` (i.e., containing some uppercase chars).

kaufco (changes requested):
LGTM, just one remark about case sensitivity and test cases.

kaufco (approved):
Looks good!

PR #4438, opened 2023-08-03
SONARJAVA-4559 Provide a list of impacted files when displaying DEBUG log of unresolved imports/types

dorian-burihabwa-sonarsource:


java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -105,7 +110,7 @@ public class SonarComponents {
 
   private final ClasspathForMain javaClasspath;
   private final ClasspathForTest javaTestClasspath;
-  private final Set<JProblem> undefinedTypes = new HashSet<>();
+  private final HashMap<String, Set<JProblem>> undefinedTypes = new HashMap<>();
```
kaufco, line 113: Map would be sufficient.

kaufco (approved):
LGTM, just one minor thing about the map.

PR #4441, opened 2023-08-07
Revert SONARJAVA-4559 faulty implementation

dorian-burihabwa-sonarsource:
The current implementation does not provide the best approach to log common issues between files and currently only logs the filename instead of the full path.

This reverts commit d165c98dcb4db03c450ca1ffcf416e00853f29ea.


irina-batinic-sonarsource (approved):
LGTM!

PR #4443, opened 2023-08-07
SONARJAVA-4552 Update ruling tests results because S6374 is deprecate…

irina-batinic-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM

PR #4444, opened 2023-08-07
Revert "SONARJAVA-4552 Update ruling tests results because S6374 is deprecated and removed from default quality profile (#4443)"

Wohops:
This reverts commit 2b888790b0d625f2ff7f308c4fca1c5d519ecb02.


dorian-burihabwa-sonarsource (approved):
Thank you for the fix

PR #4448, opened 2023-08-14
SONARJAVA-4574 update issue message for rule S106

erwan-serandour-sonarsource:
Now, we log differently System.err and System.out.


java-checks/src/main/java/org/sonar/java/checks/SystemOutOrErrUsageCheck.java
```diff
@@ -55,7 +58,11 @@ private static boolean isSystem(ExpressionTree expression) {
     return identifierTree != null && "System".equals(identifierTree.name());
   }
 
-  private static boolean isOutOrErr(MemberSelectExpressionTree mset) {
-    return "out".equals(mset.identifier().name()) || "err".equals(mset.identifier().name());
+  private static boolean isOut(MemberSelectExpressionTree mset) {
+    return "out".equals(mset.identifier().name());
```
dorian-burihabwa-sonarsource, line 62: Why don't we get the value of `identifier.name()` in visitNode rather than getting its value twice ?

dorian-burihabwa-sonarsource (changes requested):

PR #4450, opened 2023-08-14
SONARJAVA-4503 rule S2140 raises issue when casting random floating point number to long

erwan-serandour-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/RandomFloatToIntCheck.java
```diff
@@ -63,19 +63,26 @@ public List<Tree.Kind> nodesToVisit() {
   @Override
   public void visitNode(Tree tree) {
     TypeCastTree castTree = (TypeCastTree) tree;
-    if(castTree.type().symbolType().is("int")) {
-      castTree.expression().accept(new RandomDoubleVisitor());
+    if (castTree.type().symbolType().is("int")) {
+      castTree.expression().accept(new RandomDoubleVisitor("nextInt()"));
+    } else if (castTree.type().symbolType().is("long")) {
+      castTree.expression().accept(new RandomDoubleVisitor("nextLong()"));
```
dorian-burihabwa-sonarsource, line 69: If we are going to check the symbol  1 to 2 times, we might as well extract it as a variable
```suggestion
    SymbolType symbolType = castTree.type().symbolType().is("int");
    if (symbolType.is("int")) {
      castTree.expression().accept(new RandomDoubleVisitor("nextInt()"));
    } else if (symbolType.is("long")) {
      castTree.expression().accept(new RandomDoubleVisitor("nextLong()"));
```

java-checks/src/main/java/org/sonar/java/checks/RandomFloatToIntCheck.java
```diff
@@ -63,19 +63,26 @@ public List<Tree.Kind> nodesToVisit() {
...
+    private final String methodToCall;
+
+    public RandomDoubleVisitor(String methodToCall){
+      this.methodToCall = methodToCall;
+    }
 
     @Override
     public void visitMethodInvocation(MethodInvocationTree tree) {
       if (MATH_RANDOM_METHOD_MATCHER.matches(tree)) {
-        reportIssue(tree.methodSelect(), "Use \"java.util.Random.nextInt()\" instead.");
```
dorian-burihabwa-sonarsource, line 76: Let's reformat this line to add spaces between `+` and operands to make the concatenation more readable

dorian-burihabwa-sonarsource (changes requested):
Thanks for the change, the fix seems to work as intended. Let's take some time to apply some light touches to clean the code up

dorian-burihabwa-sonarsource (approved):
Thanks for the changes. LGTM 👍🏿

PR #4451, opened 2023-08-15
SONARJAVA-4509 update reporting message for rule S2438 to be more actionable

erwan-serandour-sonarsource:


java-checks-test-sources/src/main/java/checks/ThreadAsRunnableArgumentCheck.java
```diff
@@ -20,14 +21,14 @@ public void run() {
...
-    new A(0);
+    new ThreadAsRunnableArgumentCheck (0);
 
     MyClass m = new MyClass(myThread); // Noncompliant
     m.foo(myThread); // Noncompliant
     m = new MyClass(0, new MyThread()); // Noncompliant
     // Noncompliant@+1
     m = new MyClass(0, myThread, r, new MyThread()); // Noncompliant because of arg1 and arg3
-    m = new MyClass(0, new Thread[] {myThread, new MyThread()}); // Noncompliant {{"Argument 2" is a "Thread[]".}}
+    m = new MyClass(0, new Thread[] {myThread, new MyThread()}); // Noncompliant {{Replace Thread "Argument 2" with an instance of Runnable[].}}
```
dorian-burihabwa-sonarsource, line 31: From the issue message, we can tell that we need to replace argument 2 with an array of runnables but it is not clear that the original type was an array of thread and not just a single thread.
```suggestion
    m = new MyClass(0, new Thread[] {myThread, new MyThread()}); // Noncompliant {{Replace Thread[] "Argument 2" with an instance of Runnable[].}}
```

dorian-burihabwa-sonarsource: We should also change the capitalization of `Argument X` to be lowercase to `argument X` because it is no longer the beginning of the sentence.
```suggestion
    m = new MyClass(0, new Thread[] {myThread, new MyThread()}); // Noncompliant {{Replace Thread "argument 2" with an instance of Runnable[].}}
```

dorian-burihabwa-sonarsource (changes requested):
Thank you for migrating the file. We can still do a few tweaks to improve the PR and get an improved issue message

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4452, opened 2023-08-16
SONARJAVA-4559 Provide a list of impacted files when displaying DEBUG log of unresolved imports/types

dorian-burihabwa-sonarsource:


java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -494,67 +493,75 @@ public InputComponent project() {
...
+      return;
     }
 
-    final String delimiter = System.lineSeparator() + "- ";
-    final String prefix = debugMessage + delimiter;
-    final String suffix = moreThanMax ? (delimiter + "...") : "";
-
     LOG.warn(warningMessage);
     if (LOG.isDebugEnabled()) {
+      final boolean moreThanMax = messagesList.size() > maxProblems;
```
Wohops, line 556: not sure about the `final` modifier here. Either we use it everywhere, or not. In practice, we almost never use it in our projects, so I would drop it.

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -494,67 +493,75 @@ public InputComponent project() {
...
 
-  private static void logParserMessages(Stream<JProblem> messages, int maxLines, String warningMessage, String debugMessage) {
-    final List<String> messagesList = messages
-      .map(Object::toString)
-      .sorted()
+  private static void logParserMessages(Stream<Map.Entry<JProblem, List<String>>> messages, int maxProblems, String warningMessage, String debugMessage) {
+    final String problemDelimiter = System.lineSeparator() + "- ";
+    final List<List<String>> messagesList = messages
+      .sorted(Comparator.comparing(entry -> entry.getKey().toString()))
+      .limit(maxProblems + 1L)
```
Wohops, line 540: Why `+1` here? On line 561 we don't do it.

dorian-burihabwa-sonarsource: The `+1L` here allows us to prepare messages for up to `maxProblems` JProblems.
Because we do not know how many elements are contained in the stream but we know that we want at most `maxProblems`, this trick allows us to apply the mapping function unnecessarily to at most 1 element.
Knowing that we have one extra element allows us to decide whether we want to mention in the debug message that we are truncating the logs.

Wohops: The extra comment is OK for me! Thanks for the rationale.

erwan-serandour-sonarsource (approved):
LGTM

PR #4453, opened 2023-08-16
SONARJAVA-4478 S1948 does not raise on fields annotated with `@Resource`

dorian-burihabwa-sonarsource:


erwan-serandour-sonarsource (dismissed):
LGTM

java-checks/src/main/java/org/sonar/java/checks/serialization/SerializableFieldInSerializableClassCheck.java
```diff
@@ -157,7 +157,7 @@ private static boolean isTransientSerializableOrInjected(VariableTree member) {
       return true;
     }
     SymbolMetadata metadata = member.symbol().metadata();
-    return isAnnotatedWith(metadata, JAVAX_INJECT, "javax.ejb.EJB", "org.apache.wicket.spring.injection.annot.SpringBean");
+    return isAnnotatedWith(metadata, JAVAX_INJECT, "javax.ejb.EJB", "org.apache.wicket.spring.injection.annot.SpringBean", "javax.annotation.Resource", "jakarta.annotation.Resource");
```
erwan-serandour-sonarsource, line 160: this line looks too long to me. You could extract it in an array.
```suggestion
    return isAnnotatedWith(EXCLUDED_ANNOTATIONS);
```

erwan-serandour-sonarsource (changes requested):

erwan-serandour-sonarsource (approved):

PR #4454, opened 2023-08-16
SONARJAVA-4563  S1258 does not raise on fields annotated with @Builder

erwan-serandour-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4460, opened 2023-08-21
SONARJAVA-4592 Fix FPs with non-zero doubles/floats being wrongly mapped to zero

Wohops:


java-checks-test-sources/src/main/java/symbolicexecution/checks/DivisionByZeroCheckPrimitives.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public int notZero(byte b)      { return b / BYTE_NOT_ZERO; }
+  public int notZero(short s)     { return s / SHORT_NOT_ZERO; }
+  public int notZero(int i)       { return i / INT_NOT_ZERO; }
+  public long notZero(long l)     { return l / LONG_NOT_ZERO; }
+  public double notZero(double d) { return d / DOUBLE_NOT_ZERO; }
+  public float notZero(float f)   { return f / FLOAT_NOT_ZERO; }
+  public int notZero(char c)      { return c / CHAR_NOT_ZERO; }
+  public BigInteger notZero(BigInteger bi) { return bi.divide(BigInteger.ONE); }
+  public BigDecimal notZero(BigDecimal bd) { return bd.divide(BigDecimal.ONE); }
```
johann-beleites-sonarsource, line 49: Below we are also testing both the `BigInteger.ZERO` and `BigInteger.valueOf(LONG_ZERO)` divisor cases − what do you think of including a test case along the lines of:
```
  public BigDecimal notZero2(BigDecimal bd) { return bd.divide(BigDecimal.valueOf(0.3048f)); }
```

johann-beleites-sonarsource (approved):
Looks good, one minor suggestion, otherwise LGTM!

java-checks-test-sources/src/main/java/symbolicexecution/checks/DivisionByZeroCheckPrimitives.java
Wohops: fair point!

PR #4461, opened 2023-08-22
SONARJAVA-4462 Add quickfix  support for S6485

dorian-burihabwa-sonarsource:


java-checks/src/test/java/org/sonar/java/checks/KnownCapacityHashBasedCollectionCheckTest.java
```diff
@@ -20,18 +20,19 @@
...
+import org.sonar.java.checks.verifier.internal.InternalCheckVerifier;
 
 import static org.sonar.java.checks.verifier.TestUtils.mainCodeSourcesPath;
 
 class KnownCapacityHashBasedCollectionCheckTest {
 
   @Test
   void test() {
-    CheckVerifier.newVerifier()
+    InternalCheckVerifier.newInstance()
```
leonardo-pilastri-sonarsource, line 31: I realize now that it is misleading having 2 different types of verifier, and one of them as a `.withQuickFixes` method, but actually the `CheckVerifier.newVerifier()` that was there before was already checking quickfixes since we introduced the analyzer-commons-api underneath, so you can undo these changes here

java-checks/src/main/java/org/sonar/java/checks/KnownCapacityHashBasedCollectionCheck.java
```diff
@@ -47,7 +50,12 @@ public List<Kind> nodesToVisit() {
 
   @Override
   protected void onConstructorFound(NewClassTree newClassTree) {
-    reportIssue(newClassTree, getIssueMessage(newClassTree));
+    QuickFixHelper.newIssue(context)
+      .forRule(this)
+      .onTree(newClassTree)
+      .withMessage(getIssueMessage(newClassTree))
+      .withQuickFixes(() -> computeQuickFix(newClassTree))
```
leonardo-pilastri-sonarsource, line 57: Here we can use the singular api `withQuickFix` since we always provide only one

java-checks/src/main/java/org/sonar/java/checks/KnownCapacityHashBasedCollectionCheck.java
```diff
@@ -69,4 +77,13 @@ private static String getIssueMessage(NewClassTree newClassTree) {
...
   }
 
+  private static List<JavaQuickFix> computeQuickFix(NewClassTree newClassTree) {
+    String replacementMethod = TYPES_TO_METHODS.get(newClassTree.symbolType().name()).replace("(int numMappings)", "");
+    JavaTextEdit edit = JavaTextEdit.replaceBetweenTree(newClassTree.firstToken(), newClassTree.identifier().lastToken(), replacementMethod);
+    JavaQuickFix quickFix = JavaQuickFix.newQuickFix("Replace with \"" + replacementMethod + "\".")
+      .addTextEdit(edit)
+      .build();
+    return List.of(quickFix);
+  }
```
leonardo-pilastri-sonarsource, line 87: This can be simplified to return 1 quickfix

```suggestion
  private static JavaQuickFix computeQuickFix(NewClassTree newClassTree) {
    String replacementMethod = TYPES_TO_METHODS.get(newClassTree.symbolType().name()).replace("(int numMappings)", "");
    JavaTextEdit edit = JavaTextEdit.replaceBetweenTree(newClassTree.firstToken(), newClassTree.identifier().lastToken(), replacementMethod);
    return JavaQuickFix.newQuickFix("Replace with \"" + replacementMethod + "\".")
      .addTextEdit(edit)
      .build();
  }
```

leonardo-pilastri-sonarsource (changes requested):

leonardo-pilastri-sonarsource (approved):

PR #4462, opened 2023-08-22
S5689: Update issue message for clarity

dorian-burihabwa-sonarsource:


dorian-burihabwa-sonarsource:
Already fixed by #4462

PR #4474, opened 2023-10-04
Sonarjava 4612/support jakarta packages

ADarko22:


its/ruling/src/test/resources/autoscan/autoscan-diff-by-rules.json
```diff
@@ -2421,7 +2421,7 @@
     "ruleKey": "S5854",
     "hasTruePositives": true,
     "falseNegatives": 3,
-    "falsePositives": 0
+    "falsePositives": 1
```
ADarko22, line 2424: I don't know where this is coming from. Is there a way to check this?

leonardo-pilastri-sonarsource (approved):
Looks good!

PR #4475, opened 2023-10-05
SONARJAVA-4388 Supported jakarta package in CookieHttpOnlyCheck

ADarko22:


java-symbolic-execution/src/main/java/org/sonar/java/se/checks/NonNullSetToNullCheck.java
```diff
@@ -166,7 +169,9 @@ private void checkVariable(CheckerContext context, MethodTree tree, final Symbol
 
   private static boolean isJavaxValidationConstraint(Symbol symbol) {
```
leonardo-pilastri-sonarsource, line 170: Maybe it's worth renaming the method  to reflect that it also checks for jakarta constraints, something like `isJavaxOrJakartaValidationConstraint`

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/NonNullSetToNullCheck.java
```diff
@@ -166,7 +169,9 @@ private void checkVariable(CheckerContext context, MethodTree tree, final Symbol
...
     SymbolMetadata.AnnotationInstance annotation = symbol.metadata().nullabilityData().annotation();
-    return annotation != null && annotation.symbol().type().fullyQualifiedName().startsWith("javax.validation.constraints.");
+    return annotation != null
+      && (annotation.symbol().type().fullyQualifiedName().startsWith("javax.validation.constraints.")
+      || annotation.symbol().type().fullyQualifiedName().startsWith("jakarta.validation.constraints."));
```
leonardo-pilastri-sonarsource, line 174: I feel like it is starting to become less readable now with 2 different packages, what do you think of something like:

```suggestion
    if (annotation != null) {
      String fullyQualifiedName = annotation.symbol().type().fullyQualifiedName();
      return fullyQualifiedName.matches("^(javax|jakarta)\\.validation\\.constraints\\..*");
    }
    return false;
```

java-checks/src/main/java/org/sonar/java/checks/security/CookieHttpOnlyCheck.java
```diff
@@ -71,20 +71,25 @@ public class CookieHttpOnlyCheck extends IssuableSubscriptionVisitor {
...
     private static final String JAX_RS_NEW_COOKIE = "javax.ws.rs.core.NewCookie";
+    private static final String JAKARTA_RS_NEW_COOKIE = "jakarta.ws.rs.core.NewCookie";
     private static final String SHIRO_COOKIE = "org.apache.shiro.web.servlet.SimpleCookie";
     private static final String PLAY_COOKIE = "play.mvc.Http$Cookie";
     private static final String PLAY_COOKIE_BUILDER = "play.mvc.Http$CookieBuilder";
   }
 
   private static final Set<String> SETTER_NAMES = SetUtils.immutableSetOf("setHttpOnly", "withHttpOnly");
 
-  private static final Set<String> CLASSES = SetUtils.immutableSetOf(
```
leonardo-pilastri-sonarsource, line 84: Is there a specific reason to change this? I think it's good to enforce that the set is not changed by the logic of the rule at any point

leonardo-pilastri-sonarsource (changes requested):
Everything looks good, just want to check if we can make a couple of things clearer!

java-symbolic-execution/src/main/java/org/sonar/java/se/checks/NonNullSetToNullCheck.java
ADarko22: I like your simplification! I'll opt for a mixed solution:
`
    return annotation != null
      && annotation.symbol().type().fullyQualifiedName().matches("^(javax|jakarta)\\.validation\\.constraints\\..*"); 
`

java-checks/src/main/java/org/sonar/java/checks/security/CookieHttpOnlyCheck.java
ADarko22: `Set.of` is more efficient and it returns an immutable collection, it is also from `java.util` that's more convenient :)

leonardo-pilastri-sonarsource (approved):
Nice job!

PR #4482, opened 2023-10-16
SONARJAVA-4648 Implement S6810: Async methods should return void or Future

johann-beleites-sonarsource:


dorian-burihabwa-sonarsource (changes requested):
The implementation looks solid to me but we can add an extra test case before merging

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4483, opened 2023-10-16
SONARJAVA-4649 Implement S6813: Avoid field dependency injection

johann-beleites-sonarsource:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6813.json
```diff
@@ -0,0 +1,22 @@
...
+  "tags": [],
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-6813",
+  "sqKey": "S6813",
+  "scope": "All",
+  "quickfix": "unknown",
+  "code": {
+    "impacts": {
+      "MAINTAINABILITY": "MEDIUM",
+      "RELIABILITY": "MEDIUM"
```
irina-batinic-sonarsource, line 18: Do we need the `SECURITY` code impact set here, or is it okay to leave it out?

irina-batinic-sonarsource (approved):
LGTM!

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6813.json
johann-beleites-sonarsource: As far as I understand it, we only need the impacts that a rule really has. So it is okay to remove the `SECURITY` impact, if the issue doesn't touch security. At least many other migrated rules in RSPEC only have 1 of these defined.

PR #4486, opened 2023-10-16
SONARJAVA-4653 S6814 Optional REST parameters should have an object type

dorian-burihabwa-sonarsource:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6814.json
```diff
@@ -0,0 +1,23 @@
...
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-6814",
+  "sqKey": "S6814",
+  "scope": "All",
+  "quickfix": "unknown",
+  "code": {
+    "impacts": {
+      "MAINTAINABILITY": "HIGH",
+      "RELIABILITY": "MEDIUM",
+      "SECURITY": "LOW"
```
johann-beleites-sonarsource, line 19: Do we expect that this has any impact at all on e.g. security? If yes, ignore this comment, if now, should we remove the security key?

java-checks/src/main/java/org/sonar/java/checks/spring/OptionalRestParametersShouldBeObjectsCheck.java
johann-beleites-sonarsource, line 1: I might not understand some of the Spring specificities well enough here, but doesn't this also apply to [request parameters](https://www.baeldung.com/spring-request-param)?

johann-beleites-sonarsource (commented):
The implementation looks good! I only have a couple of comments on the RSPEC (see the RSPEC PR) and a question about the scope of the rule.

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6814.json
dorian-burihabwa-sonarsource: Indeed, I will remove that

java-checks/src/main/java/org/sonar/java/checks/spring/OptionalRestParametersShouldBeObjectsCheck.java
dorian-burihabwa-sonarsource: Thanks, the annotation behavior seems similar enough that I should be able to add support pretty easily

johann-beleites-sonarsource (approved):
LGTM!

PR #4487, opened 2023-10-17
SONARJAVA-4645 Add implementation of S6806

irina-batinic-sonarsource:


java-checks-test-sources/src/main/java/checks/spring/ModelAttributeNamingConventionForSpELCheck.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.Map;
+
+class ModelAttributeNamingConventionForSpELCheck {
+
+  public void foo(org.springframework.ui.Model model) {
+    model.addAllAttributes(Map.of("m", 42, "a", 22)); // Compliant
+    model.addAllAttributes(getMap()); // Compliant
+
+    model.addAllAttributes(Map.of(" m", 42, " a", 22)); // Noncompliant [[sc=35;ec=39]] {{Attribute names must begin with a letter (a-z, A-Z), underscore (_), or dollar sign ($)
```
dorian-burihabwa-sonarsource, line 11: Do we only raise a single issue on this line? Or should we raise on the two problematic elements within the line?

java-checks-test-sources/src/main/java/checks/spring/ModelAttributeNamingConventionForSpELCheck.java
```diff
@@ -0,0 +1,30 @@
...
+    model.addAttribute("", 5); // Noncompliant
+    model.addAttribute(" a", ""); // Noncompliant [[sc=24;ec=28]] {{Attribute names must begin with a letter (a-z, A-Z), underscore (_), or dollar sign ($) and can be
+                                  // followed by letters, digits, underscores, or dollar signs.}}
+    model.addAttribute("a-b", ""); // Noncompliant
+    model.addAttribute("1c", 42); // Noncompliant
+
+    model.addAttribute("a", 100); // Compliant
+    model.addAttribute("b", 42); // Compliant
+    model.addAttribute("_c", 7); // Compliant
+    model.addAttribute("$d", 8); // Compliant
```
dorian-burihabwa-sonarsource, line 23: Should we add tests with constants in addition to literals?
```suggestion
    model.addAttribute("$d", 8); // Compliant
    model.addAttribute(File.separator, 42); // Noncompliant should be forward or backward slash depending on the system but should be illegal anyway
    model.addAttribute(MY_LEGAL_CONSTANT, 0); // Compliant
    }
    private static final String MY_LEGAL_CONSTANT = "aLegalName";
```

java-checks/src/main/java/org/sonar/java/checks/spring/ModelAttributeNamingConventionForSpELCheck.java
```diff
@@ -0,0 +1,93 @@
...
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    ExpressionTree argumentTree = mit.arguments().get(0);
+    checkStringLiteralAndReport(argumentTree);
+
+    if (argumentTree.is(Tree.Kind.METHOD_INVOCATION)) {
+      MethodInvocationTree methodInvocationTree = (MethodInvocationTree) argumentTree;
+      if (MAP_OF.matches(methodInvocationTree)) {
+        for (int i = 0; i < methodInvocationTree.arguments().size(); i += 2) {
+          ExpressionTree tree = methodInvocationTree.arguments().get(i);
+          if (checkStringLiteralAndReport(tree)) {
```
dorian-burihabwa-sonarsource, line 71: Some nitpicking here, feel free to disregard
```suggestion
          ExpressionTree key = methodInvocationTree.arguments().get(i);
          if (checkStringLiteralAndReport(key)) {
```

java-checks/src/main/java/org/sonar/java/checks/spring/ModelAttributeNamingConventionForSpELCheck.java
```diff
@@ -0,0 +1,93 @@
...
+            break;
+          }
+        }
+      }
+    }
```
dorian-burihabwa-sonarsource, line 76: Should we extract this block to its own method?

java-checks/src/main/java/org/sonar/java/checks/spring/ModelAttributeNamingConventionForSpELCheck.java
```diff
@@ -0,0 +1,93 @@
...
+    .ofTypes("org.springframework.ui.Model")
+    .names("addAllAttributes")
+    .addParametersMatcher("java.util.Map")
+    .build();
+
+  private static final MethodMatchers MAP_OF = MethodMatchers.create()
+    .ofTypes("java.util.Map")
+    .names("of")
+    .withAnyParameters()
+    .build();
```
dorian-burihabwa-sonarsource, line 52: For future work (as part of another ticket), we could also look into supporting `Map.ofEntries` that is a more verbose way of doing the same job as `Map.of`

java-checks/src/main/java/org/sonar/java/checks/spring/ModelAttributeNamingConventionForSpELCheck.java
```diff
@@ -0,0 +1,93 @@
...
+
+  Pattern pattern = Pattern.compile("^[a-zA-Z_$][a-zA-Z0-9_$]*$");
```
dorian-burihabwa-sonarsource, line 54: Can we make this pattern a static field?

java-checks/src/main/java/org/sonar/java/checks/spring/ModelAttributeNamingConventionForSpELCheck.java
```diff
@@ -0,0 +1,93 @@
...
+          ExpressionTree tree = methodInvocationTree.arguments().get(i);
+          if (checkStringLiteralAndReport(tree)) {
+            break;
+          }
+        }
+      }
+    }
+  }
+
+  private boolean checkStringLiteralAndReport(ExpressionTree argumentTree) {
```
dorian-burihabwa-sonarsource, line 79: Can we make this method static? We might need to change the `pattern` field` first

java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6806.json
```diff
@@ -0,0 +1,23 @@
...
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-6806",
+  "sqKey": "S6806",
+  "scope": "Main",
+  "quickfix": "unknown",
+  "code": {
+    "impacts": {
+      "MAINTAINABILITY": "MEDIUM",
+      "RELIABILITY": "MEDIUM",
+      "SECURITY": "LOW"
```
dorian-burihabwa-sonarsource, line 19: I mentioned this on the RSPEC PR but I am not sure if this rule should add to the security impact since this is a self-inflicted pain and does not seem to be influenceable by user input

dorian-burihabwa-sonarsource (changes requested):
I think we can make that rule a lot more powerful with the support of constants (ie not literals) but let me know if you want to discuss this further

java-checks/src/main/java/org/sonar/java/checks/spring/ModelAttributeNamingConventionForSpELCheck.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private void checkMemberSelect(MemberSelectExpressionTree memberSelectExpressionTree) {
+    checkIdentifier(memberSelectExpressionTree.identifier());
+  }
+
+  private void checkMethodInvocation(MethodInvocationTree methodInvocationTree) {
+    if (MAP_OF.matches(methodInvocationTree)) {
+      for (int i = 0; i < methodInvocationTree.arguments().size(); i += 2) {
+        ExpressionTree key = methodInvocationTree.arguments().get(i);
+        if (checkStringLiteralAndReport(key, methodInvocationTree)) {
```
dorian-burihabwa-sonarsource, line 107: Thanks for the update! Should we go further and check for keys that are identifiers, as well? It looks to me like we could apply the same logic we use `onMethodInvocationFound` here to complete the loop. WDYT?

java-checks-test-sources/src/main/java/checks/spring/ModelAttributeNamingConventionForSpELCheck.java
```diff
@@ -0,0 +1,47 @@
...
+    model.addAttribute(File.separator, 42); // Compliant - can not resolve
+    model.addAttribute(MY_LEGAL_CONSTANT, 0); // Compliant
+    model.addAttribute(MY_ILLEGAL_CONSTANT, 0); // Noncompliant
+
+    model.addAllAttributes(Map.of("m", 42, "a", 22)); // Compliant
+    model.addAllAttributes(getMap()); // Compliant
+
+    model.addAllAttributes(Map.of(" m", 42, " a", 22)); // Noncompliant [[sc=28;ec=54]] {{Attribute names must begin with a letter (a-z, A-Z), underscore (_), or dollar sign ($)
+                                                        // and can be followed by letters, digits, underscores, or dollar signs.}}
+
```
dorian-burihabwa-sonarsource, line 23: To go further into the stupidly complicated what about supporting:
```suggestion

model.addAllAttributes(Map.of(MY_ILLEGAL_CONSTANT, 42)); // Noncompliant
```

dorian-burihabwa-sonarsource (changes requested):
Thanks for the changes! What do you think about using the same recursive logic when checking for a method invocation?

dorian-burihabwa-sonarsource (approved):

PR #4488, opened 2023-10-17
[No Jira] Improve AutoScanTest assertions by using soft assertions

johann-beleites-sonarsource:
This will allow us to see all failing assertions instead of only the first one.


dorian-burihabwa-sonarsource (approved):
LGTM! I can see multiple assertions failing at once

PR #4490, opened 2023-10-17
Send build-scans from CI to DevVelocity

leveretka:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 Let's just make sure that lacking a key does pollute the logs uncessarily

PR #4500, opened 2023-10-23
SONARJAVA-4664 Disable deployment of java-check-test-sources artifacts

dorian-burihabwa-sonarsource:
Can be tested by adding a local repository to the top-level pom and running `mvn clean deploy`.

An example of a local repository in your pom.xml
```xml
<distributionManagement>
    <repository>
        <id>local-folder</id>
        <name>Local folder to test deployment configuration</name>
        <url>file:./local-deployment</url>
    </repository>
</distributionManagement>
```

### DIFF
--- a/java-checks-test-sources/pom.xml
+++ b/java-checks-test-sources/pom.xml
@@ -951,12 +951,13 @@
           <argLine>--enable-preview</argLine>
         </configuration>
       </plugin>
-      <!--
-      We don't want to publish any artifacts from this module in artifactory or maven central.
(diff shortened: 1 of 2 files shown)


leonardo-pilastri-sonarsource (approved):

tomverin:
I still see a pom deployed in artifactory https://repox.jfrog.io/ui/builds/sonar-java/33475/1698051248531/published/org.sonarsource.java:java-checks-test-sources:7.27.0.33475?buildRepo=artifactory-build-info
so it seems artifactory plugin does not follow this configuration

PR #4504, opened 2023-10-24
[NO JIRA] Improve caching test output

leveretka:
* Disable tests execution in the random order 
* Remove "Build-Time" from jar Manifest (except plugin)


alban-auzeill (changes requested):

pom.xml
```diff
@@ -336,6 +330,30 @@
...
+        <configuration>
+          <archive>
+            <addMavenDescriptor>false</addMavenDescriptor>
+            <manifestEntries>
+              <!-- ${buildNumber} is the SCM revision generated by the buildnumber-maven-plugin -->
+              <Implementation-Build>${buildNumber}</Implementation-Build>
+              <!-- This is used to get plugin version at runtime using getPackage().getImplementationVersion() -->
+              <Implementation-Version>${project.version}</Implementation-Version>
+              <Build-Time>_</Build-Time>
+            </manifestEntries>
```
alban-auzeill, line 353: The parent configuration already defines `Implementation-Build` and `Implementation-Version`, we don't need to override them with the same values. Here we only need to override `Build-Time` and a comment to explain why:
```xml
<manifestEntries>
  <!--
    Develocity (gradle-enterprise-maven-extension) can reduce the build duration using its cache only for deterministic build
    'Build-Time', defined in the parent pom, changes at each build and prevents us from benefiting from the cache
    We didn't find a solution to unset 'Build-Time', so as a workaround, we provide a constant value '_'
  -->
  <Build-Time>_</Build-Time>
</manifestEntries>
```
With this configuration I still have the `Implementation-Build` and `Implementation-Version`, for example:
```shell
$ unzip -q -c "java-symbolic-execution/target/java-symbolic-execution-7.29.0-SNAPSHOT.jar" META-INF/MANIFEST.MF
Manifest-Version: 1.0
Created-By: Maven JAR Plugin 3.3.0
Build-Jdk-Spec: 17
Build-Time: _
Implementation-Build: 4b301f52d101278e3723f27b34bcaec5d82c39ed
Implementation-Version: 7.29.0-SNAPSHOT
Version: 7.29.0-SNAPSHOT
```

And one more thing. The parent pom does not force the activation of `maven-jar-plugin` by defining its configuration in `project/build/pluginManagement/plugins` and not in `project/build/plugins`. We should do the same and move this `<plugin>...</plugin>` above in `project/build/pluginManagement/plugins`.

java-checks/pom.xml
```diff
@@ -128,6 +128,7 @@
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
+              <runOrder>alphabetical</runOrder>
```
alban-auzeill, line 131: The `<runOrder>alphabetical</runOrder>` is already set on `maven-surefire-plugin` of its parent pom and it is inherited. So we don't need to repeat the configuration here.
After removing this line I still have:
```shell
$ cd java-checks
$ mvn -Psanity help:effective-pom | grep runOrder   
              <runOrder>alphabetical</runOrder>
          <runOrder>alphabetical</runOrder>
```

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #4508, opened 2023-10-27
SONARJAVA-4676 S6829: @Autowired should be used when multiple constru…

irina-batinic-sonarsource:


java-checks/src/main/resources/org/sonar/l10n/java/rules/java/S6829.json
```diff
@@ -0,0 +1,24 @@
+{
+  "title": "@Autowired should be used when multiple constructors are provided",
```
dorian-burihabwa-sonarsource, line 2: Do we have a convention on "quoting" annotations in rule titles?

java-checks-test-sources/src/main/java/checks/spring/AutowiredOnConstructorWhenMultipleConstructorsCheck.java
dorian-burihabwa-sonarsource, line 1: Should we nest all these example classes within another class? It might be nothing but I worry that every time we had a new top-level public class to `java-check-test-sources` we risk polluting the semantics of the whole module and encounter weird bugs when analyzing

java-checks/src/main/java/org/sonar/java/checks/spring/AutowiredOnConstructorWhenMultipleConstructorsCheck.java
dorian-burihabwa-sonarsource, line 1: Should we raise in cases where a field is marked as autowired? Because we also have a rule that incites developers to replace their autowired fields with autowired constructors.

dorian-burihabwa-sonarsource (changes requested):
The implementation looks pretty solid to me. There are a couple of tweaks to do in the spec and test sample but there is also a question to answer about raising when some fields are marked as autowired

java-checks/src/main/java/org/sonar/java/checks/spring/AutowiredOnConstructorWhenMultipleConstructorsCheck.java
irina-batinic-sonarsource: IMO, it is better to separate those.
Keep this one concise and only check constructors.
But, if you want, we can discuss this over the call.

dorian-burihabwa-sonarsource: No, you are right. Let's keep the 2 rules independent of one another

dorian-burihabwa-sonarsource (approved):
Thanks for the changes, LGTM 👍🏿

PR #4511, opened 2023-10-30
SONARJAVA-4681: Implement rule S6832: Non-singleton Spring beans should not be injected in a Singleton bean

ADarko22:


java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
```diff
@@ -0,0 +1,171 @@
...
+  private static boolean isNotAssignmentToSingletonValue(AssignmentExpressionTree assignmentExpressionTree) {
+    var expression = assignmentExpressionTree.expression();
+
+    if (expression.is(Tree.Kind.STRING_LITERAL)) {
+      return isNotSingletonLiteral(((LiteralTree) expression).value());
+    }
+
+    var variable = (IdentifierTree) assignmentExpressionTree.variable();
+
+    return ("value".equals(variable.name()) || "scopeName".equals(variable.name()))
```
ADarko22, line 108: I can't figure out how to improve the coverage here :(

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
```diff
@@ -0,0 +1,171 @@
...
+
+  private  void analyzeInjectedBeans(Stream<VariableTree> injectedBeans) {
+    injectedBeans
+      .filter(NonSingletonAutowiredInSingletonCheck::hasNotSingletonBeanType)
+      .forEach(variable -> reportIssue(variable.type(), "Singleton beans should not auto-wire non-Singleton beans."));
+  }
+
+  private static boolean hasNotSingletonBeanType(VariableTree classField) {
+    var typeSymbolDeclaration = classField.type().symbolType().symbol().declaration();
+    return typeSymbolDeclaration != null && hasNotSingletonScopeAnnotation(typeSymbolDeclaration.modifiers().annotations());
```
ADarko22, line 124: not testing when the declaration is null

java-checks-test-sources/src/main/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+
+  @Scope("prototype")
+  public class PrototypeBean1 {
+  }
+
+  public class SingletonBean1 {
+    @Autowired
+    private final PrototypeBean1 prototypeBean1; // Noncompliant, [[sc=19;ec=33]] {{Singleton beans should not auto-wire non-Singleton beans.}}
```
johann-beleites-sonarsource, line 32: It's enough to check the range for field injection once, it makes the tests easier to maintain if we don't check every time. It definitely does make sense to check the range again for other types of injection, like constructor injection. Again, only once though.

Same for the issue message, it's enough to only verify it once :).

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
```diff
@@ -0,0 +1,171 @@
...
+import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = "S6832")
+public class NonSingletonAutowiredInSingletonCheck extends IssuableSubscriptionVisitor {
+  private static final String SCOPED_ANNOTATION = "org.springframework.context.annotation.Scope";
+  private static final String AUTOWIRED_ANNOTATION = "org.springframework.beans.factory.annotation.Autowired";
```
johann-beleites-sonarsource, line 41: So fun fact, Spring supports the `@javax.inject.Inject` annotation as well :upside_down_face: . I would suggest we also support that one (it behaves the same as `@Autowired`).

java-checks-test-sources/src/main/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
johann-beleites-sonarsource, line 1: What about the other types of scopes that aren't singleton or prototype? It would be nice to have a test for each (it doesn't have to be as exhaustive as for prototype, just a single one would be enough).

johann-beleites-sonarsource, line 1: [Perhaps a little personal preference], WDYT of renaming the file to NonSingletonAutowiredInSingletonCheck**Sample**.java? I find that improves the ease of navigating files enormously.

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
```diff
@@ -0,0 +1,171 @@
...
+  private static Stream<VariableTree> parametersFromAutowiredConstructors(ClassTree classTree) {
+    return classTree.members()
+      .stream()
+      .filter(member -> member.is(Tree.Kind.CONSTRUCTOR))
+      .map(MethodTree.class::cast)
+      .filter(NonSingletonAutowiredInSingletonCheck::isAutowiredConstructor)
+      .flatMap(methodTree -> methodTree.parameters().stream());
+  }
+
+  private static Stream<VariableTree> autowiredParametersFromMethodsAndConstructors(ClassTree classTree) {
```
johann-beleites-sonarsource, line 153: What about methods like the following in a controller (random example):
```
  @GetMapping("/hello")
  public Map<String, String> hello(@Autowired PrototypeBean1 fooBean) {
    return Map.of();
  }
```
Shouldn't this be fine? Currently the rule raises an issue here, I think this is an FP. (Happy to discuss!)

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
johann-beleites-sonarsource, line 1: Let's discuss this some more! I think we may be able to optimize here by changing the approach a bit.

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
johann-beleites-sonarsource: Consider static imports, which mean it's not a String literal or a member select:
```
import static org.springframework.context.annotation.ScopedProxyMode.TARGET_CLASS;
```

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
ADarko22: So, the current implementation will raise an issue. The @Autowired is allowed, even if it should be logically incorrect and thus a potential new rule!

My understanding is that the `@Autowired` is basically used by the ApplicationContext to instantiate objects in those entry points. Your example requires in a way "explicit" invocation from the user, so the `@Autowired` doesn't make any sense. 

Anyway, good point and it is worth more investigation!

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
```diff
@@ -0,0 +1,157 @@
...
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = "S6832")
+public class NonSingletonAutowiredInSingletonCheck extends IssuableSubscriptionVisitor {
+  private static final String SCOPED_ANNOTATION = "org.springframework.context.annotation.Scope";
+  private static final String AUTOWIRED_ANNOTATION = "org.springframework.beans.factory.annotation.Autowired";
+  private static final String JAVAX_INJECT_ANNOTATION = "javax.inject.Inject";
+  private static final String JAKARTA_INJECT_ANNOTATION = "jakarta.inject.Inject";
```
johann-beleites-sonarsource, line 45: Indeed, good catch remembering Jakarta!

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * This rule reports an issue when a Singleton Bean auto-wires a non-Singleton Bean via constructor, field or method parameter.
+   */
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree.is(Tree.Kind.ANNOTATION)) {
+      var annotationTree = (AnnotationTree) tree;
+
+      if (isAutoWiringAnnotation(annotationTree) && isAnnotationOnKind(annotationTree, Tree.Kind.VARIABLE)) {
```
johann-beleites-sonarsource, line 62: I think we can make the conditions here a little easier to understand. You could extract `isAutoWiringAnnotation` to earlier in the check, maybe just return if it is false.
Instead of calling `isAnnotationOnKind` 3 times, you make the method return the annotated node instead, allowing you to reduce the conditions here to `annotatedNode.is(...)`. Another advantage of this is you don't have to get the annotation symbol again inside of the logic.

johann-beleites-sonarsource (commented):
As discussed, let's not raise issues on methods (except setters, identifiable by `void` return, `setX...` name and exactly 1 parameter). I'll review the rest tomorrow :).

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
ADarko22: let's discuss how to improve this

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
```diff
@@ -0,0 +1,208 @@
...
+
+  private void analyzeAnnotatedFieldOrParameter(VariableTree annotatedVar) {
+    if (isClassField(annotatedVar)) {
+      getEnclosingClass(annotatedVar.symbol().enclosingClass())
+        .ifPresent(enclosingClassTree -> reportIfNonSingletonInSingleton(enclosingClassTree, annotatedVar, "autowired field"));
+    }
+
+    if (isSetterParameter(annotatedVar) || isConstructorParameter(annotatedVar)) {
+      getEnclosingClass(annotatedVar.symbol().enclosingClass())
+        .ifPresent(enclosingClassTree -> reportIfNonSingletonInSingleton(enclosingClassTree, annotatedVar, "autowired parameter"));
```
johann-beleites-sonarsource, line 98: The only difference seems to be the message to me - can we save a couple of lines of code by only setting a different message in the if/else getting the class + reporting only once?

johann-beleites-sonarsource (approved):
Just a couple of final suggestions, then LGTM! Thanks for your patience & effort during the implementation of this rule! It's clearly a rule that requires a bit more logic and hence work to cover all cases.

java-checks/src/main/java/org/sonar/java/checks/helpers/MethodTreeUtils.java
```diff
@@ -128,14 +137,12 @@ public static Optional<MethodInvocationTree> consecutiveMethodInvocation(Tree tr
 
   public static Optional<MethodInvocationTree> subsequentMethodInvocation(Tree tree, MethodMatchers methodMatchers) {
     return consecutiveMethodInvocation(tree)
-      .map(consecutiveMethod ->
-        methodMatchers.matches(consecutiveMethod) ?
-          consecutiveMethod : subsequentMethodInvocation(consecutiveMethod, methodMatchers).orElse(null));
+      .map(consecutiveMethod -> methodMatchers.matches(consecutiveMethod) ? consecutiveMethod : subsequentMethodInvocation(consecutiveMethod, methodMatchers).orElse(null));
```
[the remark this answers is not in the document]
ADarko22: Oh, I was trying the auto-reformatting configuration on Intellij upon commits, so it probably applied the formatting to the existing code.

PR #4513, opened 2023-10-30
Enable GE remote cache in CI

leveretka:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 Let's see what we can save here

PR #4516, opened 2023-10-31
SONARJAVA-4677 Implement S6830: Bean names should adhere to the namin…

johann-beleites-sonarsource:
…g conventions


java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeanNamingConventionCheck.java
```diff
@@ -0,0 +1,90 @@
...
+    "org.springframework.context.annotation.Bean",
+    "org.springframework.context.annotation.Configuration",
+    "org.springframework.stereotype.Controller",
+    "org.springframework.stereotype.Component",
+    "org.springframework.stereotype.Repository",
+    "org.springframework.stereotype.Service",
+    "org.springframework.web.bind.annotation.RestController"
+  );
+
+  Pattern namingConvention = Pattern.compile("^[a-z][a-zA-Z0-9]*$");
```
irina-batinic-sonarsource, line 49:
```suggestion
  private Pattern namingConvention = Pattern.compile("^[a-z][a-zA-Z0-9]*$");
```

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeanNamingConventionCheck.java
```diff
@@ -0,0 +1,90 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    var annotation = (AnnotationTree) tree;
+    annotationsToCheck.stream().filter(a -> annotation.symbolType().is(a)).findFirst()
+      .map(a -> getNoncompliantNameArgument(annotation))
+      .ifPresent(n -> reportIssue(n, "Rename this bean to match the regular expression '" + namingConvention.pattern() + "'."));
+  }
+
+  @CheckForNull
+  private ExpressionTree getNoncompliantNameArgument(AnnotationTree annotation) {
```
irina-batinic-sonarsource, line 65: I would refactor this method a bit, but it is up to you.
I suggest splitting it into two methods. Move the logic from the `.map()` function into a separate method.

johann-beleites-sonarsource: Good idea. Also because it helped me find a bug that was still there!

irina-batinic-sonarsource (approved):
LGTM! Let's fix the quality gate.

PR #4517, opened 2023-10-31
SONARJAVA-4682 S6816: Nullable injected fields and parameters should provide a default value

dorian-burihabwa-sonarsource:


java-checks-test-sources/src/main/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Nullable
+  @Value(value = "${my.property}") // Noncompliant [[sc=3;ec=35;quickfixes=argumentfix]]
+  private String myPropertyToFixOnNamedArgument;
+  // fix@argumentfix {{Set null as default value}}
+  // edit@argumentfix [[sc=18;ec=34]] {{"${my.property:#{null}}"}}
+
+  private static final String WEIRD = "${my.property:#{null}}";
+  @Nullable
+  @Value(value = WEIRD) // FN compliant
```
johann-beleites-sonarsource, line 33: Why is this a FN? Isn't this just compliant? Or should the `WEIRD` field not contain a default value? That would be a good test to have and we should be able to detect it.

java-checks/src/main/java/org/sonar/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
```diff
@@ -0,0 +1,150 @@
...
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = "S6816")
+public class NullableInjectedFieldsHaveDefaultValueCheck extends IssuableSubscriptionVisitor {
+
+  private static final String VALUE_ANNOTATION = "org.springframework.beans.factory.annotation.Value";
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.CLASS);
```
johann-beleites-sonarsource, line 46: The `@Value` annotation can be used for e.g. parameters as well. WDYT of visiting `Kind.ANNOTATION` directly to avoid the need to dig deeper in the AST manually?

java-checks-test-sources/src/main/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
johann-beleites-sonarsource, line 1: `@Value` can also be applied e.g. to parameters

java-checks/src/main/java/org/sonar/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
```diff
@@ -0,0 +1,150 @@
...
+  private static Optional<String> extractLiteralValue(ExpressionTree annotationArgument) {
+    return extractLiteralTree(annotationArgument)
+      .flatMap(literal -> literal.asConstant(String.class));
+  }
+
+  private static Optional<LiteralTree> extractLiteralTree(ExpressionTree annotationArgument) {
+    if (annotationArgument.is(Tree.Kind.STRING_LITERAL)) {
+      return Optional.of((LiteralTree) annotationArgument);
+    }
+    // If the argument is not a literal, then it must be an assignment
```
johann-beleites-sonarsource, line 133: It could also be a variable (the `WEIRD` case you have in the sample file). And we have some logic already to deal with that - you can use `ExpressionHelper#getConstantValueAsString`. If it isn't an assignment you can pass the argument directly to that method, otherwise the assignment value.

java-checks/src/main/java/org/sonar/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
```diff
@@ -0,0 +1,150 @@
...
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree component = (ClassTree) tree;
+    component.members().stream()
+      .map(NullableInjectedFieldsHaveDefaultValueCheck::mapToTreesOfInterest)
+      .filter(Optional::isPresent)
+      .map(Optional::get)
+      .forEach(trees ->
+        QuickFixHelper.newIssue(context)
```
johann-beleites-sonarsource, line 57: Nice!

java-checks/src/main/java/org/sonar/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
```diff
@@ -0,0 +1,150 @@
...
+    if (instance == null) {
+      return Optional.empty();
+    }
+    return Optional.ofNullable(field.symbol().metadata().findAnnotationTree(instance));
+  }
+
+  private static boolean hasDefaultValue(AnnotationTree annotation) {
+    ExpressionTree expression = annotation.arguments().get(0);
+    return extractLiteralValue(expression)
+      .map(argument -> argument.contains(":"))
```
johann-beleites-sonarsource, line 120: I don't think this is quite enough to avoid FPs. Notably, we should probably ignore any real expressions, so things starting with `#`. The following is valid and currently raises issues:
```
@Value("#{\"foo\"}")
```
even more relevant:
```
@Value("#{null}")
```

java-checks/src/main/java/org/sonar/java/checks/spring/ValueAnnotationShouldInjectPropertyOrSpELCheck.java
```diff
@@ -61,15 +64,24 @@ public void visitNode(Tree tree) {
...
 
+  private static Optional<String> extractExpression(ExpressionTree annotationArgument) {
+    if (annotationArgument.is(Tree.Kind.STRING_LITERAL)) {
+      LiteralTree literal = (LiteralTree) annotationArgument;
+      return Optional.of(literal.value());
+    }
+    var assignment = ((AssignmentExpressionTree) annotationArgument);
+    return assignment.expression().asConstant(String.class);
+  }
+
```
johann-beleites-sonarsource, line 84: Here as well, we can use the `ExpressionHelper#getConstantValueAsString` method

java-checks-test-sources/src/main/java/checks/spring/ValueAnnotationShouldInjectPropertyOrSpELCheckSample.java
```diff
@@ -20,6 +27,7 @@ class ValueAnnotationShouldInjectPropertyOrSpELCheckSample {
   @Value("Hello, world!") // Noncompliant, this use of @Value is redundant
   String greetingA;
 
+
```
johann-beleites-sonarsource, line 30:
```suggestion
```

johann-beleites-sonarsource (changes requested):
I think changing the approach to visit `Kind.ANNOTATION` directly might make the rule a bit simpler and more complete. Happy to discuss! Some other smaller comments.

java-checks-test-sources/src/main/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
dorian-burihabwa-sonarsource: My mistake, I modified the example at some point and made it compliant when it wasn't. I will double the example with a non-compliant one now

java-checks/src/main/java/org/sonar/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
dorian-burihabwa-sonarsource: That could work. However, I think that starting at the level of the class, we can maintain the same approach of pruning the members or interest and keep the exact same reporting and quick-fix logic. WDYT?

java-checks/src/main/java/org/sonar/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
dorian-burihabwa-sonarsource: Indeed, the logic is too naive (so is the one for the quick-fix). I will have another try at hardening it

java-checks/src/main/java/org/sonar/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
johann-beleites-sonarsource: Hmm when I set a breakpoint on line 159 when visiting sample line 41, I can call:
```
ExpressionsHelper.getConstantValueAsString(expression).value()
```
And it will return `${my.constant:#{null}}`.
I think the problem here may be that the line 41 should read:
```
  @Value(value = NON_COMPLIANT_IN_A_CONSTANT) // Compliant FN
```
but currently reads:
```
  @Value(value = DEFINED_IN_A_CONSTANT) // Compliant FN
```

In any case, the other problem remains that the annotation argument could be a variable (this case is currently missing in the sample file). If I add e.g. `@Value(DEFINED_IN_A_CONSTANT)` somewhere, it will cause a `ClassCastException` on line 155.

java-checks/src/main/java/org/sonar/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Maps a variable to an {@link AnnotationsOfInterest} if it is annotated as nullable and injected with a Spring value annotation missing a default value.
+   *
+   * @param variable A class member (eg: method, field, ...)
+   * @return An Optional with a {@link AnnotationsOfInterest} if the member matches. Optional.empty() otherwise.
+   */
+  private static Optional<AnnotationsOfInterest> mapToAnnotationsOfInterest(VariableTree variable) {
+    Optional<AnnotationTree> valueAnnotation = getValueAnnotationWithoutDefault(variable);
+    if (valueAnnotation.isEmpty()) {
+      return Optional.empty();
```
johann-beleites-sonarsource, line 105: Minor: any reason not to just return `valueAnnotation` itself when it's empty?

johann-beleites-sonarsource (commented):
I found a couple of edge cases that aren't quite handled yet. I think it shows that this rule is somewhat on the complicated side, but I think it's going to be a nice one!

java-checks/src/main/java/org/sonar/java/checks/spring/NullableInjectedFieldsHaveDefaultValueCheck.java
```diff
@@ -0,0 +1,226 @@
...
+    } else {
+      valueAnnotation = valueAnnotationOnParent;
+    }
+    Optional<AnnotationTree> nullableAnnotation = getNullableAnnotation(variable);
+    return nullableAnnotation.map(annotationTree -> new AnnotationsOfInterest(valueAnnotation, annotationTree));
+  }
+
+  @Nullable
+  private static AnnotationTree extractValueAnnotationOnSetter(MethodTree method) {
+    if (method.simpleName().name().startsWith("set") && method.parameters().size() == 1) {
```
johann-beleites-sonarsource, line 163: As it so happens, @ADarko22 just introduced a helper to check if a method is a setter this morning `MethodTreeUtils#isSetterMethod`: https://github.com/SonarSource/sonar-java/pull/4511/files#diff-96cf5c4851ac66958b366adab42fe06f493aea83a9c7be579c15701839748b16R83

johann-beleites-sonarsource (approved):
Quite a few edge cases to consider in this rule! One minor comment due to new changes introduced to master this morning. You were probably planning to do so anyway, it would be nice to merge this as 2 commits, one for SONARJAVA-4682, one for SONARJAVA-4691. LGTM!

PR #4518, opened 2023-10-31
SONARJAVA-4678: mplement S6831: @Qualifier should not be used on @Bean methods

ADarko22:


java-checks/src/main/java/org/sonar/java/checks/spring/AvoidQualifierOnBeanMethodsCheck.java
```diff
@@ -0,0 +1,72 @@
...
+    var methodTree = (MethodTree) tree;
+
+    var beanAnnotation = getAnnotation(methodTree, BEAN_ANNOTATION);
+    var qualifierAnnotation = getAnnotation(methodTree, QUALIFIER_ANNOTATION);
+
+    if(beanAnnotation != null && qualifierAnnotation != null) {
+      QuickFixHelper.newIssue(context)
+       .forRule(this)
+        .onTree(qualifierAnnotation)
+        .withMessage("Remove this redundant \"@Qualifier\" annotation")
```
dorian-burihabwa-sonarsource, line 56: Let's add a `.` to finish the issue message
```suggestion
        .withMessage("Remove this redundant \"@Qualifier\" annotation.")
```

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/spring/AvoidQualifierOnBeanMethodsCheck.java
```diff
@@ -0,0 +1,114 @@
...
+      .orElse(null);
+  }
+
+  private static List<JavaQuickFix> getQuickFix(MethodTree methodTree, AnnotationTree qualifierAnnotation) {
+    List<JavaQuickFix> quickFixes = new LinkedList<>();
+
+    // quick fix only for @Qualifier annotations without arguments or with argument that matches the method name
+    if(isFixable(methodTree, qualifierAnnotation)) {
+      var quickFix = JavaQuickFix.newQuickFix("Remove \"@Qualifier\"")
+        .addTextEdit(JavaTextEdit.removeTree(qualifierAnnotation))
```
ADarko22, line 80: quick fix tested locally and works; anyway comments on the same line are not remove (i.e. the // Noncompliant int he sample)

java-checks/src/main/java/org/sonar/java/checks/spring/AvoidQualifierOnBeanMethodsCheck.java
```diff
@@ -0,0 +1,117 @@
...
+    var qualifierAnnotationValue = "";
+
+    if(argument.is(Tree.Kind.ASSIGNMENT)) {
+      qualifierAnnotationValue = ((LiteralTreeImpl) ((AssignmentExpressionTreeImpl) argument).expression()).value().replace("\"", "");
+    }
+    else if(argument.is(Tree.Kind.STRING_LITERAL)) {
+      qualifierAnnotationValue = ((LiteralTreeImpl) argument).token().text().replace("\"", "");
+    }
+
+    return qualifierAnnotationValue;
```
dorian-burihabwa-sonarsource, line 113: I understand checking the types twice just to be sure but in practice I am not sure if we are likely to see other types of arguments. Have you encountered any other while testing?

java-checks/src/main/java/org/sonar/java/checks/spring/AvoidQualifierOnBeanMethodsCheck.java
```diff
@@ -0,0 +1,117 @@
...
+    var methodName = methodTree.simpleName().name();
+    return getQualifierAnnotationValue(arguments).equals(methodName);
+  }
+
+  private static String getQualifierAnnotationValue(Arguments arguments) {
+    var argument = arguments.get(0);
+    var qualifierAnnotationValue = "";
+
+    if(argument.is(Tree.Kind.ASSIGNMENT)) {
+      qualifierAnnotationValue = ((LiteralTreeImpl) ((AssignmentExpressionTreeImpl) argument).expression()).value().replace("\"", "");
```
dorian-burihabwa-sonarsource, line 107: Is it possible to set the value of the assignment expression using a constant?
```java
class A {
  static final String CONSTANT = "foo";
  @Bean  
  @Qualifier(value = CONSTANT)
  public String foo() { return "Hello"; }
}
```
I know that it is possible for the `@Variable` annotation. We should just make sure it is safe for our code to assume that we always have a literal

dorian-burihabwa-sonarsource (changes requested):
Looking nice on the QF front, let's make sure we cover all our angles and fix the open issues

dorian-burihabwa-sonarsource (approved):

PR #4520, opened 2023-11-01
[No Jira] Improve Autoscan testing

johann-beleites-sonarsource:


its/ruling/src/test/java/org/sonar/java/it/AutoScanTest.java
```diff
@@ -167,28 +170,49 @@ public void javaCheckTestSources() throws Exception {
...
+      path -> path.getFileName().toString().startsWith("diff_") && path.toString().endsWith(".json"))) {
+      dirStream.forEach(knownDiffFiles::add);
+    }
+    var knownDiffs = new HashMap<String, IssueDiff>();
+    for (var diffFile : knownDiffFiles) {
+      IssueDiff diff = GSON.fromJson(Files.readString(diffFile), GSON_ISSUE_DIFF_TYPE);
+      knownDiffs.put(diff.ruleKey, diff);
+    }
+
+    // store new unexpected diffs in JSON files - serializable
```
johann-beleites-sonarsource, line 185: For easier debugging, we only output diffs that are different from the expected. There is no value in outputting diffs that are expected that way.

its/ruling/src/test/java/org/sonar/java/it/AutoScanTest.java
```diff
@@ -167,28 +170,49 @@ public void javaCheckTestSources() throws Exception {
...
      */
+
+    // The expected number of differences is the sum of FPs and FNs from the known differences.
+    // We calculate this value based on the known diffs to avoid a single value in the tests that is affected by all rules (which would
+    // inevitably lead to merge conflicts when people are working on rules in parallel).
+    var expectedDiffs = knownDiffs.stream().map(diff -> diff.falseNegatives + diff.falsePositives).reduce(Integer::sum).orElse(0);
+
     String differences = Files.readString(pathFor(TARGET_ACTUAL + PROJECT_KEY + "-no-binaries_differences"));
-    softly.assertThat(differences).isEqualTo("Issues differences: 3473");
+    softly.assertThat(differences).isEqualTo("Issues differences: " + expectedDiffs);
```
johann-beleites-sonarsource, line 215: I'm fine with removing this assertion entirely (I don't see much value here that goes beyond the comparison of the actual and new diffs collections), however the modification is okay as well, as it doesn't need to be edited when diffs change.

its/ruling/src/test/java/org/sonar/java/it/AutoScanTest.java
```diff
@@ -167,28 +170,49 @@ public void javaCheckTestSources() throws Exception {
...
-    IssueDiff knownTotal = IssueDiff.total(knownDiffs);
+    IssueDiff knownTotal = IssueDiff.total(knownDiffs.values());
 
     SoftAssertions softly = new SoftAssertions();
-    softly.assertThat(newDiffs).containsExactlyInAnyOrderElementsOf(knownDiffs);
+    softly.assertThat(newDiffs).containsExactlyInAnyOrderElementsOf(knownDiffs.values());
     softly.assertThat(newTotal).isEqualTo(knownTotal);
     softly.assertThat(rulesCausingFPs).hasSize(6);
     softly.assertThat(rulesNotReporting).hasSize(7);
-    softly.assertThat(rulesSilenced).hasSize(74);
```
johann-beleites-sonarsource, line 183: The vast majority of rules relying on semantics will be silenced by the absence of binaries. Nowadays, it seems to me that most of our rules that we write rely on semantics. I don't think having this check is very valuable in addition to the diff data, that already tells us if a rule is silenced.

its/ruling/src/test/java/org/sonar/java/it/AutoScanTest.java
```diff
@@ -167,28 +170,49 @@ public void javaCheckTestSources() throws Exception {
...
-    List<IssueDiff> knownDiffs = GSON.fromJson(Files.readString(pathFor("src/test/resources/autoscan/" + DIFF_FILE + ".json")), GSON_LIST_ISSUE_DIFF_TYPE);
-    IssueDiff knownTotal = IssueDiff.total(knownDiffs);
+    IssueDiff knownTotal = IssueDiff.total(knownDiffs.values());
 
     SoftAssertions softly = new SoftAssertions();
-    softly.assertThat(newDiffs).containsExactlyInAnyOrderElementsOf(knownDiffs);
+    softly.assertThat(newDiffs).containsExactlyInAnyOrderElementsOf(knownDiffs.values());
     softly.assertThat(newTotal).isEqualTo(knownTotal);
     softly.assertThat(rulesCausingFPs).hasSize(6);
     softly.assertThat(rulesNotReporting).hasSize(7);
```
johann-beleites-sonarsource, line 201: I think it's okay to leave these two assertions because this should actually be unexpected in most cases and require some additional verification. It's changed very rarely.

leveretka (approved):
Looks good to me. Thanks for the effort.
Let's merge this PR after the release

PR #4521, opened 2023-11-02
SONARJAVA-4679 S6833: @Controller should be replaced with @RestContro…

irina-batinic-sonarsource:


java-checks-test-sources/src/main/java/checks/spring/ControllerWithRestControllerReplacementCheck.java
```diff
@@ -0,0 +1,79 @@
+package checks.spring;
+
+import javax.annotation.Nullable;
+import org.springframework.stereotype.Controller;
+import org.springframework.stereotype.Service;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+public class ControllerWithRestControllerReplacementCheck {
+
+  @Controller // Noncompliant [[sc=3;ec=14;secondary=14;quickfixes=qf1]] {{Replace the "@Controller" annotation by "@RestController" and remove all
```
dorian-burihabwa-sonarsource, line 10: We are able to offer the removal of the `@ResponseBody` annotations as part of the quickfix, but to complete the operation we should also replace the `@Controller` annotation, shouldn't we?

java-checks/src/main/java/org/sonar/java/checks/spring/ControllerWithRestControllerReplacementCheck.java
```diff
@@ -0,0 +1,83 @@
...
+
+    if (!secondaryLocations.isEmpty()) {
+      QuickFixHelper.newIssue(context)
+        .forRule(this)
+        .onTree(annotation.get())
+        .withMessage("Replace the \"@Controller\" annotation by \"@RestController\" and remove all \"@ResponseBody\" annotations.")
+        .withSecondaries(secondaryLocations)
+        .withQuickFixes(() -> List.of(JavaQuickFix.newQuickFix("Remove \"@ResponseBody\" annotations.").addTextEdits(edits).build()))
+        .report();
+    }
```
dorian-burihabwa-sonarsource, line 80: Nitpicking here: I would rather early return like we do earlier in the method. Somehow `!collection.isEmpty` always feels clunky.
```suggestion
    if (secondaryLocations.isEmpty()) {
      return;
    }
    QuickFixHelper.newIssue(context)
      .forRule(this)
      .onTree(annotation.get())
      .withMessage("Replace the \"@Controller\" annotation by \"@RestController\" and remove all \"@ResponseBody\" annotations.")
      .withSecondaries(secondaryLocations)
      .withQuickFixes(() -> List.of(JavaQuickFix.newQuickFix("Remove \"@ResponseBody\" annotations.").addTextEdits(edits).build()))
      .report(); 
```

dorian-burihabwa-sonarsource (changes requested):
The rule looks good but it seems that the quickfix implementation is incomplete

java-checks-test-sources/src/main/java/checks/spring/ControllerWithRestControllerReplacementCheck.java
irina-batinic-sonarsource: Yes! It is done.

dorian-burihabwa-sonarsource (approved):

PR #4524, opened 2023-11-06
SONARJAVA-4681: Fix bug due to wrong semantics usage on S6832

ADarko22:


johann-beleites-sonarsource (commented):
Let's just add a regression test where the scoped bean is defined in another file, otherwise looks good!

java-checks/src/main/java/org/sonar/java/checks/spring/NonSingletonAutowiredInSingletonCheck.java
```diff
@@ -177,6 +177,7 @@ private static boolean isNotSingletonScopeAnnotation(SymbolMetadata.AnnotationIn
 
   private static boolean isNotSingletonAnnotationValue(SymbolMetadata.AnnotationValue annotationValue) {
     return ("value".equals(annotationValue.name()) || "scopeName".equals(annotationValue.name()))
-      && annotationValue.value() instanceof String && !"singleton".equalsIgnoreCase((String) annotationValue.value());
+      // both "value" and "scopeName" in @Scope annotation have String type
```
ADarko22, line 180: no need for the type check :)

johann-beleites-sonarsource (approved):
LGTM!

PR #4533, opened 2023-11-13
SONARJAVA-4617 S1226 - Fix CFG construction inside "return switch" expression for yield without break cases

irina-batinic-sonarsource:


java-checks/src/test/files/checks/ParameterReassignedToCheck.java
dorian-burihabwa-sonarsource, line 1: Can we move this file to `java-checks-test-sources` ? If it is not possible to place it with the compiling files at least move it with the non-compiling ones.

java-checks/src/test/files/checks/ParameterReassignedToCheck.java
```diff
@@ -1,6 +1,19 @@
...
+      case 1 -> {
+        try {
+          yield "";
+        } catch (Exception e) {
+          yield "Exc";
+        }
+      }
+      default -> null;
+    };
+  }
```
dorian-burihabwa-sonarsource, line 16: This is a nice example to show that we do not break on switch expressions that yield. But it if feels gratuitous without a compliant/noncompliant comment in the sample for a specific rule. Can we replace this with/add a non-compliant example?

java-frontend/src/main/java/org/sonar/java/cfg/CFG.java
```diff
@@ -923,7 +923,7 @@ private void buildBreakStatement(BreakStatementTree tree) {
   }
 
   private void buildYieldStatement(YieldStatementTree tree) {
-    currentBlock = createUnconditionalJump(tree, breakTargets.getLast(), currentBlock);
+    currentBlock = createUnconditionalJump(tree, breakTargets.isEmpty() ? null : breakTargets.getLast(), currentBlock);
```
dorian-burihabwa-sonarsource, line 926: Finding the right place to test the CFG is a bit complicated but I think a good candidate would be the tests for `MethodTreeImpl` that has a CFG method. With the new fix, we should not fail to generate a CFG for the example method that you added in `java-checks/src/test/files/checks/ParameterReassignedToCheck.java` . In particular, the [compute_cfg test](https://github.com/SonarSource/sonar-java/blob/71f5cb1008da6e42e70412012ec95f3898776209/java-frontend/src/test/java/org/sonar/java/model/declaration/MethodTreeImplTest.java#L198) looks like a good place to add test cases.

dorian-burihabwa-sonarsource (changes requested):
Nice changes but I think we should double-up in the way we test so that we can detect regressions both in the frontend and in the checks

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 Thank you for investigating the issue, too bad the SE interactions do not make it possible to extend the rule further 🤷🏿‍♂️

PR #4534, opened 2023-11-14
SONARJAVA-4447 S2185: Rephrase issue message to not use "silly"

kaufco:


java-checks/src/test/files/checks/ConstantMath.java
dorian-burihabwa-sonarsource, line 1: While we are fixing the issue message, we should migrate the test file to `java-checks-test-sources` (in the compiling set of files if possible but in the non-compiling one at least if too difficult)

dorian-burihabwa-sonarsource (changes requested):
LGTM overall, let's move over the test sample to the test sources module before merging

dorian-burihabwa-sonarsource (approved):
Thank you for the changes! LGTM 👍🏿

PR #4536, opened 2023-11-14
BUILD-4131 Use GitHub token from Vault instead of sonartech api token

sebastienvermeille:
# BUILD-4131 Use GitHub token from Vault instead of GitHub

## Changes
* Retrieve Github token from Hashicorp Vault instead of the one stored in Github secrets
  That way it no longer depends on a technical user account and can be rotated easily

## Depends on
* https://github.com/SonarSource/re-terraform-aws-vault/pull/823

## Checks


matemoln (changes requested):
`contents: write` permission is not needed

PR #4537, opened 2023-11-14
SONARJAVA-4169: S3553 should not report an issue for spring mvc optio…

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/OptionalAsParameterCheck.java
```diff
@@ -54,6 +53,11 @@ public void visitNode(Tree tree) {
     if (Boolean.FALSE.equals(methodTree.isOverriding())) {
 
       for (VariableTree parameter : methodTree.parameters()) {
+        boolean hasRequestParamAnnotation = parameter.symbol().metadata().isAnnotatedWith("org.springframework.web.bind.annotation.RequestParam");
```
dorian-burihabwa-sonarsource, line 56: According to [S6814](https://sonarsource.github.io/rspec/#/rspec/S6814), we should also support the `@PathVariable` annotation as it should also accept Optional-wrapped parameters

dorian-burihabwa-sonarsource (changes requested):
LGTM but I think we also need to support the `@PathVariable` annotation to be consistent with S6814

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 Thank you for the changes!

PR #4538, opened 2023-11-15
SONARJAVA-4502 Add quickfix for S1153

kaufco:


java-checks/src/main/java/org/sonar/java/checks/ConcatenationWithStringValueOfCheck.java
```diff
@@ -69,13 +71,31 @@ public void visitBinaryExpression(BinaryExpressionTree tree) {
...
 
+  private JavaQuickFix createQuickFix(MethodInvocationTree invocationTree) {
+    ExpressionTree argumentTree = invocationTree.arguments().get(0);
+    String replacement = QuickFixHelper.contentForTree(argumentTree, context);
+
+    if (argumentTree instanceof BinaryExpressionTree && !argumentTree.symbolType().is("java.lang.String")) {
+      replacement = "(" + replacement + ")";
+    }
+
+    return JavaQuickFix.newQuickFix("Replace String.valueOf() with its argument")
```
dorian-burihabwa-sonarsource, line 94: How about "Remove redundant String.valueOf wrapping"
```suggestion
    return JavaQuickFix.newQuickFix("Remove redundant String.valueOf wrapping")
```

dorian-burihabwa-sonarsource (approved):
LGTM but I wonder we could not improve the QF message

PR #4547, opened 2023-11-16
Fix Mend build, it uses the non -recursive maven option '-pl'

alban-auzeill:


alban-auzeill:
Wrong branch name, should start with 'mend-'

PR #4551, opened 2023-11-17
SONARJAVA-4576 Update API with methods from JUtils

leonardo-pilastri-sonarsource:


dorian-burihabwa-sonarsource (changes requested):
I think the changes look pretty but there are a couple of things we can do to improve before merging.

1. The checks that have been modified to call the moved away from JUtils still import JUtils and might create a long list of issues when merged into master
2. The JIRA ticket does not document the move of the isIntersectionType

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 Thank you for the changes!

PR #4558, opened 2023-11-20
SONARJAVA-4590 S1319: Method parameter uses method not available in interface

kaufco:


java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java
```diff
@@ -177,4 +330,44 @@ private static String toSimpleName(String fullyQualifiedName) {
   private static boolean isPublic(ModifiersTree modifiers) {
     return ModifiersUtils.hasModifier(modifiers, Modifier.PUBLIC);
   }
+
+  private static <T> Set<T> union(Set<T> a, Set<T> b) {
```
dorian-burihabwa-sonarsource, line 334: Since this method is only used to join set of strings, we can probably drop the type parameter and just use `String` as the type. WDYT?

java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java
```diff
@@ -109,35 +243,54 @@ public void visitVariable(VariableTree tree) {
...
-      checkIfAllowed(tree.returnType(), "The return type of this method");
-      for (VariableTree variableTree : tree.parameters()) {
-        checkIfAllowed(variableTree.type(), String.format("The type of \"%s\"", variableTree.simpleName()));
-      }
+    if (!isPublic(tree.modifiers()) || !Boolean.FALSE.equals(tree.isOverriding())) {
+      return;
+    }
+
+    checkIfAllowed(tree.returnType(), "The return type of this method");
+    var candidateParameters = tree.parameters().stream()
```
dorian-burihabwa-sonarsource, line 251: I think it would be better to have the type of the resulting collection for clarity
```suggestion
    List<VariableTree> candidateParameters = tree.parameters().stream()
```

java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java
```diff
@@ -177,4 +330,44 @@ private static String toSimpleName(String fullyQualifiedName) {
...
+    set.addAll(a);
+    set.addAll(b);
+    return set;
+  }
+
+  private static class MethodBodyVisitor extends BaseTreeVisitor {
+
+    private final Map<String, VariableTree> candidateParametersByName = new HashMap<>();
+
+    public final Set<VariableTree> excludeParameters = new HashSet<>();
```
dorian-burihabwa-sonarsource, line 345: How about renaming this field to `excludedParameters` to make it more clear in usage that it is a variable and not a method?

java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java
```diff
@@ -20,39 +20,45 @@
...
   private static final String QUEUE = "java.util.Queue";
   private static final String SET = "java.util.Set";
-  private static final String SORTED_MAP = "java.util.SortedMap";
   private static final String SORTED_SET = "java.util.SortedSet";
+  private static final String MAP = "java.util.Map";
+  private static final String CONCURRENT_MAP = "java.util.concurrent.ConcurrentMap";
+  private static final String SORTED_MAP = "java.util.SortedMap";
 
-  private static final Map<String, String> MAPPING = MapBuilder.<String, String> newMap()
+  private static final Map<String, String> SUGGESTED_INTERFACE_BY_COLLECTION_CLASS = MapBuilder.<String, String>newMap()
```
dorian-burihabwa-sonarsource, line 61: We will probably need to rework some of this mapping strategy with the introduction of SequencedCollection, SequencedSet and and SequencedMap but let's keep this for when we can actually test against Java 21 samples

java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java
```diff
@@ -88,6 +94,134 @@ public class CollectionImplementationReferencedCheck extends BaseTreeVisitor imp
...
+  private static final Map<String, Set<String>> METHODS_BY_INTERFACE = MapBuilder.<String, Set<String>>newMap()
+    .put(LIST, LIST_METHODS)
+    .put(QUEUE, QUEUE_METHODS)
+    .put(DEQUE, DEQUE_METHODS)
+    .put(SET, SET_METHODS)
+    .put(SORTED_SET, SORTED_SET_METHODS)
+    .put(MAP, MAP_METHODS)
+    .put(CONCURRENT_MAP, CONCURRENT_MAP_METHODS)
+    .put(SORTED_MAP, SORTED_MAP_METHODS)
+    .build();
```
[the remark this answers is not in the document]
kaufco: Fixed.

java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java
```diff
@@ -109,35 +243,54 @@ public void visitVariable(VariableTree tree) {
   @Override
   public void visitMethod(MethodTree tree) {
     super.visitMethod(tree);
-    if (isPublic(tree.modifiers()) && Boolean.FALSE.equals(tree.isOverriding())) {
-      checkIfAllowed(tree.returnType(), "The return type of this method");
-      for (VariableTree variableTree : tree.parameters()) {
-        checkIfAllowed(variableTree.type(), String.format("The type of \"%s\"", variableTree.simpleName()));
-      }
+    if (!isPublic(tree.modifiers()) || !Boolean.FALSE.equals(tree.isOverriding())) {
```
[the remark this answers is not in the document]
kaufco: Fixed.

java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java
kaufco: Replaced it with `SetUtils.concat` from `analyzer-commons`.

java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java
kaufco: Fixed.

java-checks/src/main/java/org/sonar/java/checks/CollectionImplementationReferencedCheck.java
kaufco: Fixed.

PR #4562, opened 2023-11-21
SONARJAVA-4513 FN S2060 (ExternalizableClassConstructorCheck) if no-argument constructor is not public

kaufco:


leonardo-pilastri-sonarsource (approved):
Very nice job! Looks good to me!

PR #4574, opened 2023-11-23
[NO JIRA] Improve CI performance, split compile, test and analysis

alban-auzeill:
The goal is to experiment with a new build method that separates the build from the tests and analysis. This will allow us to start the integration tests just after the build and not wait for the analysis to be completed.


leveretka (approved):
LGTM!

PR #4583, opened 2023-11-24
[NO JIRA] Remove 'Plugin-RequirePlugins' from hash calculations for t…

leveretka:
…he build cache


leveretka:
This should speed up the plugin ITs as they will use remote cache. Since the custom rules plugin is now a direct dependency, we now need to take into account another manifest property that contains versions. This is 'Plugin-RequirePlugins'

PR #4591, opened 2023-11-29
SONARJAVA-4143 update S3749 to raise issues when a class is annotated with `@RestControler`

erwan-serandour-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4596, opened 2023-11-30
SONARJAVA-4722

kaufco:


java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,246 @@
...
+    var startIndex = 0;
+    var endIndex = placeholder.indexOf(':');
+
+    while (endIndex != -1) {
+      var segment = placeholder.substring(startIndex, endIndex);
+      if (!isValidPropertyPlaceholderSegment(segment, rangeStart + startIndex)) {
+        return false;
+      }
+      startIndex = endIndex + 1;
+      endIndex = placeholder.indexOf(':', startIndex);
```
dorian-burihabwa-sonarsource, line 195: My understanding is that the column token marks an alternative in SpEL syntax. In practice, is there a limit to the number of alternatives we can encounter?

java-checks/pom.xml
```diff
@@ -18,6 +18,11 @@
       <artifactId>sonar-plugin-api</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+    <groupId>org.springframework</groupId>
+      <artifactId>spring-expression</artifactId>
+      <version>6.1.1</version>
+    </dependency>
```
dorian-burihabwa-sonarsource, line 25: We will want to move this dependency version in the dependency management of the top-level pom

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,246 @@
...
+import org.springframework.expression.ParseException;
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+
+@Rule(key = "S6857")
+public class SpelExpressionCheck extends IssuableSubscriptionVisitor {
+
+  private static final String SPRING_PREFIX = "org.springframework";
+
+  private static final Pattern PROPERTY_PLACEHOLDER_PATTERN = Pattern.compile(
+    "[a-zA-Z_]\\w*+(\\[\\d++])*+(\\.[a-zA-Z_]\\w*+(\\[\\d++])*+)*+"
```
dorian-burihabwa-sonarsource, line 48: Do we have a reference for this pattern? Isnt' a placeholder pattern any pattern that respects Java's identifier syntax?

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,246 @@
...
+  private static boolean isValidPropertyPlaceholderSegment(String segment, int rangeStart) {
+    var stripped = segment.stripLeading();
+    rangeStart += segment.length() - stripped.length();
+    stripped = stripped.stripTrailing();
+
+    if (stripped.startsWith("#{")) {
+      parseDelimitersAndContents(stripped, 1, rangeStart + 2, SpelExpressionCheck::checkValidSpelExpression);
+      return true;
+    } else {
+      return PROPERTY_PLACEHOLDER_PATTERN.matcher(stripped).matches();
```
dorian-burihabwa-sonarsource, line 210: Can't we just use isValidSpelExpression on placeholders?

dorian-burihabwa-sonarsource (changes requested):
I have done a pretty shallow review but I will continue on tomorrow. There are still a few questions that we could try to answer:
* Can we just give placeholders for the expression parser to resolve?
* If we are now importing the parsing from spring-expression, is there a utility that would allow us to extract the expression from the brace delimiters?
* Can we extract the parsing logic from the rule itself?

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+  }
+
+  private static boolean isSpringAnnotation(AnnotationTree annotation) {
+    return annotation.symbolType().fullyQualifiedName().startsWith(SPRING_PREFIX);
+  }
+
+  private void checkSpringAnnotationArguments(AnnotationTree annotation) {
+    annotation.arguments().stream()
+      .filter(argument -> argument.is(Tree.Kind.STRING_LITERAL))
```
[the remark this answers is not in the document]
dorian-burihabwa-sonarsource: There is also another limitation here which is that a SpringExpression might be a resolvable constant rather than a literal.
```java
// Defined in the same class or elsewhere
public static final String MY_CONSTANT_EXPRESSION = "${1 + 2 + 3}"
// ...
@Value(MY_CONSTANT_EXPRESSION)
String myProperty;
```

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,246 @@
...
+      tokenStart.columnOffset() + error.rangeStart,
+      tokenStart.line(),
+      tokenStart.columnOffset() + error.rangeEnd
+    );
+
+    var analyzerMessage = new AnalyzerMessage(this, context.getInputFile(), textSpan, error.getMessage(), 0);
+    ((DefaultJavaFileScannerContext) context).reportIssue(analyzerMessage);
+  }
+
+  private static void parseStringContents(String content) {
```
dorian-burihabwa-sonarsource, line 112: We should add the exception to the signature since the this is the last level it can bubble up from
```suggestion
  private static void parseStringContents(String content) throws SyntaxError {
```

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,246 @@
...
+          i = parseDelimitersAndContents(content, i + 1, i + 1, SpelExpressionCheck::checkValidSpelExpression);
+          break;
+        default:
+          i++;
+          break;
+      }
+    }
+  }
+
+  private static int parseDelimitersAndContents(
```
dorian-burihabwa-sonarsource, line 130: With 2 parameters of the same type with very similar names, you probably want to add some Javadoc to explain why `startIndex` is sometimes different from `rangeStart`

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,246 @@
...
+        case '}':
+          openCount--;
+          if (openCount == 0) {
+            return i;
+          }
+          break;
+        default:
+          break;
+      }
+    }
```
dorian-burihabwa-sonarsource, line 170: This could be a for loop as well
```suggestion
    var i = startIndex + 1;

    for (int opencount = 1; i < value.length(); i++) {
      var c = value.charAt(i);
      if (c == '{') {
          openCount++;
      } else if (c == '}') {
          openCount--;
          if (openCount == 0) {
            return i;
          }
       }
    }
```

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
dorian-burihabwa-sonarsource: This regex is about identifiers and member selects, let's see if we can find a reference regex for those cases

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,246 @@
...
+public class SpelExpressionCheck extends IssuableSubscriptionVisitor {
+
+  private static final String SPRING_PREFIX = "org.springframework";
+
+  private static final Pattern PROPERTY_PLACEHOLDER_PATTERN = Pattern.compile(
+    "[a-zA-Z_]\\w*+(\\[\\d++])*+(\\.[a-zA-Z_]\\w*+(\\[\\d++])*+)*+"
+  );
+
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.ANNOTATION_TYPE);
```
dorian-burihabwa-sonarsource, line 52: Investigate why annotation declarations need to be explored for semantics to be built properly

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
dorian-burihabwa-sonarsource: Is there a relation between startIndex and rangeStart?

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
kaufco: Could not reproduce that this is required to build up the semantics for Annotations anymore. Whatever strange effect made it appear so during the dev phase: the tests are still working when I remove it. Fixed.

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
kaufco: Fixed.

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
kaufco: Added a doc for the method + gave the parameter a more proper name (also in the other methods). Fixed.

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
kaufco: Sadly, no :-/

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
kaufco: See comment on same regex below. After doing a quick research, I found nothing that implements the grammar as which is now documented in the comment, along with some examples. Fixed.

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
kaufco: Probably not: I have not found a single example that has more than one `:` in a property placeholder. However, I have also not found an explicit specification of what is allowed and what not, or even a grammar for property placeholders. Best doc I have found so far is the API spec of the Spring PropertyPlaceholderConfigurer.

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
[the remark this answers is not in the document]
kaufco: OMG. This has implications for reporting exact error position inside the string literal again. Fixed.

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,246 @@
...
+    var endIndex = parseDelimiterBraces(value, startIndex, rangeStart);
+    if (endIndex == startIndex) {
+      return endIndex;
+    }
+    var contents = value.substring(startIndex + 1, endIndex - 1);
+    parseContents.accept(contents, rangeStart);
+    return endIndex;
+  }
+
+  private static int parseDelimiterBraces(String value, int startIndex, int rangeStart) throws SyntaxError {
```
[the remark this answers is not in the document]
kaufco: I think the current name fits better, bacause:
1. it consomues the input (because it returns next char input) and also can throw a SyntaxError. Hence I think `parse` fits better here.
2. it also checks for the starting brace, not just the closing one, hence `DelimiterBraces`

java-checks/pom.xml
kaufco: Fixed.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S6857.json
```diff
@@ -0,0 +1,23 @@
+{
+  "title": "SpEL expression should have valid syntax",
```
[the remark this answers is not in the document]
kaufco: Fixed.

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,301 @@
...
+   *
+   * <pre>
+   * ('{' contents '}')?
+   * </pre>
+   * <p>
+   * Where correct bracing is checked and then <code>contents</code> is parsed using the given <code>parseContents</code> function.
+   *
+   * @param value         string containing the character sequence to parse
+   * @param startIndex    start index of the character sequence in <code>value</code>
+   * @param startColumn   column offset between the original string literal and <code>value</code> (needed for reporting)
```
dorian-burihabwa-sonarsource, line 187: Thank you for the change, my point was also about understanding why these two values would be different.
I was expecting something closer to
```suggestion
   * @param startIndex       index of the opening delimiter we start from in the <code>value</code> value
   * @param startColumn   offset with the position of  <code>value</code> within a potentially longer original string (used for reporting)
```

dorian-burihabwa-sonarsource: Let's be careful when mixing index and column terms because index is generic and 0-based, while column could be considered analyzer jargon and lead to confusion

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
dorian-burihabwa-sonarsource: Indeed, from this [outdated grammar](https://github.com/spring-projects/spring-framework/blob/2a268f2250a1016c0c1444a9a4a467c8f2b71d32/spring-expression/src/main/java/org/springframework/expression/spel/generated/SpringExpressions.g#L47C1-L47C1), I am not sure that there is a limit on the number of alternatives

dorian-burihabwa-sonarsource (commented):
There are still a few changes that could be made here but I am going to be able to go for a second round of review today. If there is no hurry, I will have a look next week. Otherwise feel free to submit it for review to someone else

java-checks/pom.xml
```diff
@@ -18,6 +18,11 @@
       <artifactId>sonar-plugin-api</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+    <groupId>org.springframework</groupId>
+      <artifactId>spring-expression</artifactId>
+      <version>${spring.expressions.version}</version>
+    </dependency>
```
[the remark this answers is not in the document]
kaufco: Fixed.

java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -0,0 +1,301 @@
...
+public class SpelExpressionCheck extends IssuableSubscriptionVisitor {
+
+  private static final String SPRING_PREFIX = "org.springframework";
+
+  /**
+   * Regular expression for a property placeholder segment that is not a SpEL expression.
+   * It implements the following grammar with possessive quantifiers:
+   * <p>
+   * ```
+   * PropertyPlaceholder ::= Identifier IndexedAccess* ("." Identifier IndexedAccess*)*
```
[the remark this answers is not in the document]
kaufco: Changed.

pom.xml
```diff
@@ -91,6 +91,7 @@
     <sslr.version>1.24.0.633</sslr.version>
     <argLine>-Xmx512m</argLine>
     <gitRepositoryName>sonar-java</gitRepositoryName>
+
```
dorian-burihabwa-sonarsource, line 94: Let's remove this line before merging
```suggestion
```

dorian-burihabwa-sonarsource (approved):
LGTM! Thank you for the changes!

PR #4597, opened 2023-11-30
SONARJAVA-4685 S6838 Bean methods should not be invoked directly when `proxyBeanMethods` is set to false

dorian-burihabwa-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/spring/BeanMethodOfNonProxiedSingletonInvocationCheck.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private static Optional<AnnotationTree> getConfigurationAnnotation(ClassTree tree) {
+    SymbolMetadata metadata = tree.symbol().metadata();
+    for (SymbolMetadata.AnnotationInstance instance : metadata.annotations()) {
+      if (instance.symbol().type().is(CONFIGURATION_ANNOTATION)) {
+        return Optional.ofNullable(metadata.findAnnotationTree(instance));
+      }
+    }
+    return Optional.empty();
```
kaufco, line 69: What about `metadata.annotations().stream().findFirst(...)` instead? I like the more functional approach with streams, and even for a consistent coding style because you use streams in two other functions (`visitNode()`, `hasProxyBeanMethodsDisabled()`) as well. Bonus: `findFirst()` even returns `Optional.empty()` already when nothing is found.

java-checks/src/main/java/org/sonar/java/checks/spring/BeanMethodOfNonProxiedSingletonInvocationCheck.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    boolean isTargetClass = getConfigurationAnnotation((ClassTree) tree)
+      .map(BeanMethodOfNonProxiedSingletonInvocationCheck::hasProxyBeanMethodsDisabled)
+      .orElse(Boolean.FALSE);
+    if (!isTargetClass) {
+      return;
+    }
```
kaufco, line 56: Maybe this version is a bit more readable? As said, I like the functional style, but maybe it's overdone here.

```suggestion
    var annotation = getConfigurationAnnotation((ClassTree) tree);
    if (annotation.isEmpty() || !hasProxyBeanMethodsDisabled(annotation.get())) {
      return;
    }
```

java-checks/src/main/java/org/sonar/java/checks/spring/BeanMethodOfNonProxiedSingletonInvocationCheck.java
```diff
@@ -0,0 +1,121 @@
...
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree tree) {
+      super.visitMethodInvocation(tree);
+      MethodTree declaration = tree.methodSymbol().declaration();
+      if (declaration == null) {
+        return;
+      }
+      if (returnsAPrototypeBean(declaration)) {
+        return;
+      }
```
kaufco, line 102: Maybe combine this into one condition?

```suggestion
      if (declaration == null || returnsAPrototypeBean(declaration)) {
        return;
      }
```

java-checks/src/main/java/org/sonar/java/checks/spring/BeanMethodOfNonProxiedSingletonInvocationCheck.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private static boolean returnsAPrototypeBean(MethodTree method) {
+      List<SymbolMetadata.AnnotationValue> annotationValues = method.symbol().metadata().valuesForAnnotation(SCOPE_ANNOTATION);
+      if (annotationValues == null) {
+        return false;
+      }
+      return annotationValues.stream()
+        .filter(argument -> List.of("value", "scopeName").contains(argument.name()))
+        .map(SymbolMetadata.AnnotationValue::value)
+        .anyMatch("Prototype"::equals);
```
kaufco, line 117: _Could_ also be combined in one `return`, but maybe your version with two returns is more readable:

```suggestion
      return annotationValues != null && annotationValues.stream()
        .filter(argument -> List.of("value", "scopeName").contains(argument.name()))
        .map(SymbolMetadata.AnnotationValue::value)
        .anyMatch("Prototype"::equals);
```

java-checks/src/main/java/org/sonar/java/checks/spring/SpringBeansShouldBeAccessibleCheck.java
```diff
@@ -163,7 +163,7 @@ private static void writeToCache(InputFileScannerContext context, Collection<Str
     }
   }
 
-  private static Optional<List<String>> readFromCache(InputFileScannerContext context) {
+  private static Optional<List<String>>   readFromCache(InputFileScannerContext context) {
```
kaufco, line 166: Bad formatting: 3 spaces instead of one.

```suggestion
  private static Optional<List<String>> readFromCache(InputFileScannerContext context) {
```

sonar-java-plugin/src/main/java/org/sonar/plugins/java/CheckList.java
```diff
@@ -24,6 +24,7 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.sonar.java.checks.spring.BeanMethodOfNonProxiedSingletonInvocationCheck;
```
kaufco, line 28: Bad sorting of `import` statements: should be moved to the block with the other spring checks imports.

kaufco (changes requested):
Implementation idea / algorithm + completeness of the tests looks good. I made some suggestions for code improvements.

java-checks/src/main/java/org/sonar/java/checks/spring/BeanMethodOfNonProxiedSingletonInvocationCheck.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+    var visitor = new NonProxiedMethodInvocationVisitor((ClassTree) tree);
+    tree.accept(visitor);
+    visitor.locations.forEach(invocation -> reportIssue(invocation, "Replace this bean method invocation " + "with a dependency injection."));
+  }
+
+  private static Optional<AnnotationTree> getConfigurationAnnotation(ClassTree tree) {
+    SymbolMetadata metadata = tree.symbol().metadata();
+    for (SymbolMetadata.AnnotationInstance instance : metadata.annotations()) {
+      if (instance.symbol().type().is(CONFIGURATION_ANNOTATION)) {
```
kaufco, line 65: We need one more test case to get full test coverage in this line: a class that is annotated with something else than `@Configuration`. However, when transformed into a stream (see other ther comment), this is not necessary anymore to get the full coverage, but maybe we should still add it.

kaufco (changes requested):

java-checks/src/main/java/org/sonar/java/checks/spring/BeanMethodOfNonProxiedSingletonInvocationCheck.java
```diff
@@ -0,0 +1,115 @@
...
+    if (configurationAnnotation.isEmpty() || !hasProxyBeanMethodsDisabled(configurationAnnotation.get())) {
+      return;
+    }
+    var visitor = new NonProxiedMethodInvocationVisitor((ClassTree) tree);
+    tree.accept(visitor);
+    visitor.locations.forEach(invocation -> reportIssue(invocation, "Replace this bean method invocation with a dependency injection."));
+  }
+
+  private static Optional<AnnotationTree> getConfigurationAnnotation(ClassTree tree) {
+    SymbolMetadata metadata = tree.symbol().metadata();
```
erwan-serandour-sonarsource, line 61:
```suggestion
    return tree.modifiers().annotations().stream()
      .filter(annotation -> annotation.annotationType().symbolType().is(CONFIGURATION_ANNOTATION))
      .findFirst();
```
We can do a bit shorter.

erwan-serandour-sonarsource (approved):
LGTM

kaufco (approved):
LGTM!

PR #4599, opened 2023-12-01
SONARJAVA-4727 S6862: Beans in "@Configuration" class should have different names

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ConfigurationBeanNamesCheck.java
```diff
@@ -0,0 +1,73 @@
...
+    if (!isConfigurationClass(classTree)) {
+      return;
+    }
+
+    var beanMethods = getBeanMethods(classTree);
+    var reportedElements = new HashSet<>();
+
+    for (int i = 0; i < beanMethods.size(); i++) {
+      String name = beanMethods.get(i).simpleName().name();
+      for (int j = i + 1; j < beanMethods.size(); j++) {
```
kaufco, line 51: Nested loop -> O(n²) complexity, and scanning a loop iteratively for the presence of an element (inner loop) is not a good practice. I would prefer the use of one loop and a HashSet instead:
```
var foundNames = new HashSet<String>();
for (MethodTree bean: beanMethods) {
  if (!foundNames.add(bean.simpleName().name()) {
    // report the issue here
  }
}
```

kaufco (changes requested):

java-checks-test-sources/default/src/main/java/checks/ConfigurationBeanNamesCheckSample.java
```diff
@@ -0,0 +1,80 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.core.userdetails.User;
+
+public class ConfigurationBeanNamesCheckSample {
+
+  @Configuration
+  class Config1 {
+    @Bean
+    public User user() {
```
kaufco, line 13: I think the usage of `User` instead of a simple dummy class (like `foo`, or an own `class User{}`) overly complicates the test, because:
1. lots of boilerplate code for the constructor call, but the check does not even enter method bodies, so it's not necessary for the test
2. it creates an otherwise unnecessary import

kaufco (approved):
LGTM! Just two minor remarks for the test.

PR #4606, opened 2023-12-06
SONARJAVA-4723 in S6856, filter path variables used by Model Attribute when we extracting path variable from the url

erwan-serandour-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/PathVariableAnnotationShouldBePresentIfPathVariableIsUsedCheck.java
```diff
@@ -51,18 +53,30 @@ public class PathVariableAnnotationShouldBePresentIfPathVariableIsUsedCheck exte
...
+      .map(variable -> pathVariableName(variable))
+      .flatMap(Optional::stream)
+      .collect(Collectors.toSet());
+
+    methods.forEach(method -> MAPPING_ANNOTATIONS
+      .forEach(annotation -> reportIssueOnParameters(method, annotation, modelAttributePathVariable)));
   }
 
-  private void reportIssueOnParameters(MethodTree method, String annotation) {
+  private void reportIssueOnParameters(MethodTree method, String annotation, Set<String> modelAttributePathVariable) {
```
kaufco, line 79: This method does too many things at once. I would split it into two or three methods. Also about the naming: I think we have the implicit convention that a method should be called `reportSomething` only when it _definitely_ reports something (like: all paths in the method lead to it). If it may report or not, we should call it `checkSomething` (or something similar that fits the purpose of the method).

About splitting: I would split it into `containsMapParameter`, sth. like `findMissingPathVariables` and then the actual reporting:

```java
private void checkParameters(...) {
  if (!containsMapParameter) {
    return;
  }
  findMissingPathVariables(...).forEach(MissingVariableAnnotationCheck::reportMissingPathVariable);
}
```

kaufco (changes requested):
Looks good, just some remarks.

java-checks/src/main/java/org/sonar/java/checks/PathVariableAnnotationShouldBePresentIfPathVariableIsUsedCheck.java
```diff
@@ -41,6 +42,7 @@
 @Rule(key = "S6856")
 public class PathVariableAnnotationShouldBePresentIfPathVariableIsUsedCheck extends IssuableSubscriptionVisitor {
   private static final String PATH_VARIABLE_ANNOTATION = "org.springframework.web.bind.annotation.PathVariable";
+  private static final String MODEL_ATTRIBUTE_ANNOTATION = "org.springframework.web.bind.annotation.ModelAttribute";
   private static final Pattern EXTRACT_PATH_VARIABLE = Pattern.compile("([^:}/]*)(:.*)?\\}.*");
   private static final Predicate<String> CONTAINS_PLACEHOLDER = Pattern.compile("\\$\\{.*\\}").asPredicate();
```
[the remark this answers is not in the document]
erwan-serandour-sonarsource: :+1:

java-checks/src/main/java/org/sonar/java/checks/PathVariableAnnotationShouldBePresentIfPathVariableIsUsedCheck.java
```diff
@@ -51,18 +53,30 @@ public class PathVariableAnnotationShouldBePresentIfPathVariableIsUsedCheck exte
...
-      .forEach(annotation -> reportIssueOnParameters(method, annotation));
+    List<MethodTree> methods = clazzTree.members().stream()
+      .filter(member -> member.is(Tree.Kind.METHOD))
+      .map(MethodTree.class::cast)
+      .collect(Collectors.toList());
+
+    Set<String> modelAttributePathVariable = methods.stream()
+      .filter(method -> method.symbol().metadata().isAnnotatedWith(MODEL_ATTRIBUTE_ANNOTATION))
+      .flatMap(method -> method.parameters().stream())
+      .map(variable -> pathVariableName(variable))
```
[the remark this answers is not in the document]
erwan-serandour-sonarsource: :+1:

java-checks/src/main/java/org/sonar/java/checks/MissingPathVariableAnnotationCheck.java
```diff
@@ -68,56 +68,62 @@ public void visitNode(Tree tree) {
...
 
   private static ExpressionTree annotation(MethodTree method, String name) {
     return method.modifiers().annotations().stream()
       .filter(annotation -> annotation.symbolType().is(name))
-      .findFirst()
-      // it will never be null because we are filtering on the annotation before.
-      .orElse(null);
+      .collect(Collectors.toList())
+      // it will never be empty because we are filtering on the annotation before.
+      .get(0);
```
kaufco, line 126: Should be changed into (as discussed in slack):

```
      .findFirst()
      // it will never be null because we are filtering on the annotation before.
      .orElseThrow();
```

kaufco (approved):
LGTM! Just one remark about unnecessary list built-up

java-checks-test-sources/default/src/main/java/checks/MissingPathVariableAnnotationCheckSample.java
```diff
@@ -187,4 +181,54 @@ public String getPlaceHolder(String id) { // compliant, we don't consider this c
...
+    }
+
+    @GetMapping("/{id}/{name}")
+    public String get() { // compliant, @ModelAttribute is always called before @GetMapping to generate the model. In our case model attribute
+      // consume the id and name path variables
+      return "Hello World";
+    }
+
+    @GetMapping("/{id}/{name}/{age}")
+    public String get2(@PathVariable String age) { // compliant
```
kaufco, line 197: Should be same line as Noncompliant, and also (although it has no effect on the test itself), its `// Compliant` (with capital C). Happens multiple times in this file.

```suggestion
    @GetMapping("/{id}/{name}/{age}") // Compliant
    public String get2(@PathVariable String age) { 
```

kaufco (approved):
Nope, it's 2, not 1 like I claimed earlier. I forgot about this one in the test samples. I still approved, sou you can merge after doing the change.

PR #4607, opened 2023-12-07
SONARJAVA-4427  add andExpect to assertions methods patterns to support Spring test methods

erwan-serandour-sonarsource:


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 Just make sure that we do need to make the test sample abstract

PR #4609, opened 2023-12-11
SONARJAVA-4732 S6863: Set appropriate Status Codes on HTTP responses

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -0,0 +1,150 @@
...
+
+      if (statusMethodMatchers.matches(methodInvocationTree)) {
+        checkTryCatchForStatus(methodInvocationTree);
+      }
+
+      MethodMatchers okMethodMatchersWithParam = MethodMatchers.create()
+        .ofTypes(RESPONSE_ENTITY)
+        .names("ok")
+        .withAnyParameters()
+        .build();
```
dorian-burihabwa-sonarsource, line 74: Let's move all method matchers outside of this method for clarity

java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -0,0 +1,150 @@
...
+
+      if (catchParent != null && !isError) {
+        reportIssue(methodInvocationTree,
+          "Use the \"ResponseEntity.badRequest()\" or \"ResponseEntity.notFound()\" method" +
+            "or set the status to \"HttpStatus.INTERNAL_SERVER_ERROR\" or \"HttpStatus.NOT_FOUND\".");
+      }
+      return catchParent;
+    }
+
+    private void checkTryWithOkStatus(MethodInvocationTree methodInvocationTree, boolean isOk) {
```
dorian-burihabwa-sonarsource, line 136: I am a bit confused as to why two methods with similar signatures `check[Catch|Try]with[Error|Ok]Status` sometimes return a Tree sometimes. Can we align on that?

java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+      }
+
+      super.visitMethodInvocation(methodInvocationTree);
+    }
+
+    private void checkTryCatchForStatus(MethodInvocationTree methodInvocationTree) {
+      Tree catchParent = ExpressionUtils.getParentOfType(methodInvocationTree, Tree.Kind.CATCH);
+      boolean isError = methodInvocationTree.arguments().stream()
+        .map(MemberSelectExpressionTree.class::cast)
```
dorian-burihabwa-sonarsource, line 104: Is it possible to statically import the single name `NOT_FOUND` and break the cast here?

dorian-burihabwa-sonarsource (changes requested):
Made a first pass, let's discuss if we can make our check a little more tolerant on the codes accepted and move the method matchers and other constants outside of the methods they are used in

java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -0,0 +1,222 @@
...
+
+@Rule(key = "S6863")
+public class StatusCodesOnResponseCheck extends IssuableSubscriptionVisitor {
+
+  public static final String RESPONSE_ENTITY = "org.springframework.http.ResponseEntity";
+  public static final String OK_ISSUE_MESSAGE = "Use the \"ResponseEntity.ok()\" method or set the status to \"HttpStatus.OK\".";
+  public static final String ERROR_ISSUE_MESSAGE = "Use the \"ResponseEntity.badRequest()\" or \"ResponseEntity.notFound()\" method" +
+    "or set the status to \"HttpStatus.INTERNAL_SERVER_ERROR\" or \"HttpStatus.NOT_FOUND\".";
+
+  private final List<String> okCodes = List.of("CONTINUE",
```
dorian-burihabwa-sonarsource, line 44: We should add a link to the documentation we extracted these codes. This will help us in the future when dealing with FPs/FNs to ensure that we are using a list that is up-to-date (or needs to be updated).

dorian-burihabwa-sonarsource: Complementary question: Is the list sorted in any way? Alphabetically? According to the error code?

java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -0,0 +1,222 @@
...
+      return;
+    }
+
+    MethodInvocationVisitor methodInvocationVisitor = new MethodInvocationVisitor();
+    classTree.accept(methodInvocationVisitor);
+  }
+
+  private class MethodInvocationVisitor extends BaseTreeVisitor {
+
+    MethodMatchers statusMethodMatchers = MethodMatchers.create()
```
dorian-burihabwa-sonarsource, line 124: We should mark this matcher and the following ones as `static final`. Perhaps `private` as well?

java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -0,0 +1,222 @@
...
+
+      super.visitMethodInvocation(methodInvocationTree);
+    }
+
+    private void checkTryCatch(MethodInvocationTree methodInvocationTree) {
+      Tree catchParent = ExpressionUtils.getParentOfType(methodInvocationTree, Tree.Kind.CATCH);
+      boolean isError = isCodeInList(methodInvocationTree, errorCodes);
+
+      if (catchParent != null && !isError) {
+        reportIssue(methodInvocationTree, ERROR_ISSUE_MESSAGE);
```
dorian-burihabwa-sonarsource, line 171: Is there any reason to continue with the rest of the method if we report here? Is there a chance we can report here and in a try as well below?
```suggestion
        reportIssue(methodInvocationTree, ERROR_ISSUE_MESSAGE);
        return;
```

java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -0,0 +1,222 @@
...
+      for (ExpressionTree arg : methodInvocationTree.arguments()) {
+        if (arg.is(Tree.Kind.MEMBER_SELECT)) {
+          MemberSelectExpressionTree memberSelectExpressionTree = (MemberSelectExpressionTree) arg;
+          return codes.contains(memberSelectExpressionTree.identifier().name());
+        } else if (arg.is(Tree.Kind.IDENTIFIER)) {
+          IdentifierTree identifierTree = (IdentifierTree) arg;
+          return codes.contains(identifierTree.name());
+        }
+      }
+      return true;
```
dorian-burihabwa-sonarsource, line 194: My understanding is that the methods invoked here take up to 1 argument (since we systematically return after testing the first argument). Can we replace this `for` loop with a single `if` statement?

dorian-burihabwa-sonarsource (changes requested):

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #4610, opened 2023-12-11
SONARJAVA-4685 Remove FP on invocation of non-bean methods

dorian-burihabwa-sonarsource:
Issues should only be raise when the method would be proxied if `proxyBeanMethods` was set to `true`. While this is the case for `@Bean` annotated methods, it is not the case for other methods.


erwan-serandour-sonarsource (approved):
LGTM

PR #4611, opened 2023-12-13
SONARJAVA-4722: Fixed FPs, "-" allowed in property placeholders

kaufco:


java-checks/src/main/java/org/sonar/java/checks/spring/SpelExpressionCheck.java
```diff
@@ -55,7 +55,7 @@ public class SpelExpressionCheck extends IssuableSubscriptionVisitor {
    * <p>
    * <pre>
    * PropertyPlaceholder ::= Identifier IndexExpression* ("." Identifier IndexExpression*)*
-   * Identifier          ::= [a-zA-Z_][a-zA-Z0-9_]*
+   * Identifier          ::= [a-zA-Z0-9_-]+
```
dorian-burihabwa-sonarsource, line 58: So we can also start with a hyphen `-`? It would be nice to have an example that matches this.

dorian-burihabwa-sonarsource (changes requested):
Thanks for the changes, the commits need to be reworked into a set of 2 commits that will then be rebased on to master
1. Use a version of spring-expression that is Java 11 compatible
2. Fix False Positives on property placeholders containing hyphens `-`

dorian-burihabwa-sonarsource (approved):
LGTM Thank you for the changes!

PR #4613, opened 2023-12-13
SONARJAVA-4414 (S5413): SuspiciousListRemove: control flow with "return" is compliant

erwan-serandour-sonarsource:


dorian-burihabwa-sonarsource (approved):
Please make sure to add a test on the issue location and an acknowledgement of julianladisch's work before merging

PR #4615, opened 2023-12-14
SONARJAVA-4732 Fix FPs S6863: Set appropriate Status Codes on HTTP responses

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -217,12 +216,21 @@ private Tree checkCatch(MethodInvocationTree methodInvocationTree, boolean isErr
     private Tree checkTry(MethodInvocationTree methodInvocationTree, boolean isOk) {
       Tree tryParent = ExpressionUtils.getParentOfType(methodInvocationTree, Tree.Kind.TRY_STATEMENT);
 
-      if (tryParent != null && !isOk) {
-        reportIssue(methodInvocationTree, ISSUE_MESSAGE);
+      if (tryParent != null) {
+        Tree ifParent = ExpressionUtils.getParentOfType(methodInvocationTree, Tree.Kind.IF_STATEMENT);
```
dorian-burihabwa-sonarsource, line 220: Shouldn't we look for the `ifParent` of the the `tryParent`?
```suggestion
        Tree ifParent = ExpressionUtils.getParentOfType(tryParent, Tree.Kind.IF_STATEMENT);
```

java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -217,12 +216,21 @@ private Tree checkCatch(MethodInvocationTree methodInvocationTree, boolean isErr
...
+        if (ifParent == null) {
+          if (!isOk) {
+            reportIssue(methodInvocationTree, ISSUE_MESSAGE);
+          }
+        } else {
+          tryParent = ExpressionUtils.getParentOfType(ifParent, Tree.Kind.TRY_STATEMENT);
```
dorian-burihabwa-sonarsource, line 226: Why do we look up for a parent try if we could not find any in the first place?

java-checks-test-sources/default/src/main/java/checks/spring/StatusCodesOnResponseCheckSample.java
```diff
@@ -24,6 +24,52 @@ public ResponseEntity<User> getBadReqUser() {
       return ResponseEntity.badRequest().build(); // Compliant
     }
 
+    public ResponseEntity<User> foo() {
+      User user = getUserObject();
+      if (user == null) {
+        return ResponseEntity.notFound().build(); // Compliant
```
dorian-burihabwa-sonarsource, line 30: I think we should add a comment here to explain that we cannot raise on invocation inside of if statements alone

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/spring/StatusCodesOnResponseCheck.java
```diff
@@ -207,22 +207,50 @@ private boolean isCodeInList(MethodInvocationTree methodInvocationTree, List<Str
...
       }
       return catchParent;
     }
 
     private Tree checkTry(MethodInvocationTree methodInvocationTree, boolean isOk) {
-      Tree tryParent = ExpressionUtils.getParentOfType(methodInvocationTree, Tree.Kind.TRY_STATEMENT);
-
-      if (tryParent != null && !isOk) {
+      Parents parents = getParentsTryAndIf(methodInvocationTree);
+      if (!parents.foundIfWithinTry() && parents.foundTryStatement && !isOk) {
```
dorian-burihabwa-sonarsource, line 218: Nice changes!

java-checks-test-sources/default/src/main/java/checks/spring/StatusCodesOnResponseCheckSample.java
dorian-burihabwa-sonarsource: Let's describe in the comments why we do not raise on invocations in `if` statements

dorian-burihabwa-sonarsource (approved):
LGTM, let's document the if case  in the sample to make it easier to understand in the future

PR #4623, opened 2023-12-19
SONARJAVA-4638 Improve Jakarta support

johann-beleites-sonarsource:
Add support for the Jakarta namespace for the following checks:

* RegexComplexityCheck
* ReluctantQuantifierWithEmptyContinuationCheck
* SecureCookieCheck
* SerializableObjectInSessionCheck
* ServletInstanceFieldCheck

### DIFF
--- a/sonar-java-plugin/src/test/java/org/sonar/plugins/java/SanityTest.java
+++ b/sonar-java-plugin/src/test/java/org/sonar/plugins/java/SanityTest.java
@@ -32,6 +32,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.assertj.core.api.SoftAssertions;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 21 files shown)


sonar-java-plugin/src/test/java/org/sonar/plugins/java/SanityTest.java
```diff
@@ -109,7 +110,7 @@ class SanityTest {
    * It does not prevent other rules to fail if similar construct of the language, but not yet encountered.
    */
   @Test
-  @EnabledIfSystemProperty(named = "force.sanity.test", matches = "true")
+  //@EnabledIfSystemProperty(named = "force.sanity.test", matches = "true")
```
ADarko22, line 113: Is this commented code intentional? Better to revert or remove in case it is not needed

ADarko22 (changes requested):
Excellent job covering all these checks and fixing the SanityTest.

There is only a commented-out line in the SaityTest that probably needs to be uncommented; please have a look.

ADarko22 (approved):

PR #4626, opened 2024-01-02
SONARJAVA-4745: Fixed license headers and depend on parent POM

kaufco:


pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.sonarsource.parent</groupId>
     <artifactId>parent</artifactId>
-    <version>70.0.0.270</version>
+    <version>70.0.0.1287</version>
```
johann-beleites-sonarsource, line 8: This is not actually a released version of the parent, it is only a dev build - we can't depend on it. If in doubt, you can [check maven central](https://mvnrepository.com/artifact/org.sonarsource.parent/parent) for open-source projects to see what has been released.

johann-beleites-sonarsource (changes requested):

pom.xml
kaufco: Fixed.

johann-beleites-sonarsource (approved):

PR #4634, opened 2024-01-18
SONARJAVA-4754 Use 'leaveFile()' instead of 'scanFile()' in java:S1849

Wohops:


java-checks/src/main/java/org/sonar/java/checks/HasNextCallingNextCheck.java
```diff
@@ -54,7 +54,7 @@ public void visitNode(Tree tree) {
   }
 
   @Override
-  public void scanFile(JavaFileScannerContext context) {
+  public void leaveFile(JavaFileScannerContext context) {
```
alban-auzeill, line 57: Cleaning the state in `leaveFile` is not guaranty, to ensure we don't face FP or FN because of this, we also need to clean the sate in `setContext`:
```java
  @Override
    super.setContext(context);
    hasNextBodyVisitor.clearHasNextOwner();
  }
```

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/HasNextCallingNextCheck.java
Wohops: good point

alban-auzeill (approved):

PR #4635, opened 2024-01-19
SONARJAVA-4413: add Vert.x assertion methods for JUnit 5

ADarko22:


java-checks/src/main/java/org/sonar/java/checks/helpers/UnitTestUtils.java
```diff
@@ -205,4 +213,16 @@ private static boolean hasNestedClass(ClassTree classTree) {
       .map(ClassTree.class::cast)
       .anyMatch(UnitTestUtils::hasNestedAnnotation);
   }
+
+  public static boolean matchesAssertionMethodPattern(IdentifierTree method, Symbol methodSymbol) {
```
leonardo-pilastri-sonarsource, line 217: I think this method could have a more specific name since it belongs to a class that does not strictly imply we are working with methods. Also, as we only make use of the `method.name()`, we can make the method take as the first parameter directly the `String methodName`. This could also be useful for future uses where we might not have an IdentifierTree to compare against.
```suggestion
  public static boolean methodNameMatchesAssertionMethodPattern(String methodName, Symbol methodSymbol) {
```

leonardo-pilastri-sonarsource (changes requested):
Great work! Just a small change about a method naming and then it looks good to me!

leonardo-pilastri-sonarsource (approved):

PR #4638, opened 2024-01-22
SONARJAVA-4759 Deprecate methods scanFile() and scanTree() and prevent implementation

Wohops:


java-frontend/src/test/java/org/sonar/plugins/java/api/IssuableSubscriptionVisitorTest.java
```diff
@@ -58,6 +58,18 @@ void check_issuable_subscription_visitor_does_not_visit_tree_on_its_own() {
...
+  @Test
+  void issuable_subscription_visitor_does_not_visit_file_on_its_own() {
+    CustomRule visitor = new CustomRule();
+    JavaFileScannerContext context = Mockito.mock(JavaFileScannerContext.class);
+    try {
+      visitor.scanFile(context);
+      fail("Analysis should have failed");
+    } catch (UnsupportedOperationException e) {
+      assertThat(e).hasMessage("IssuableSubscriptionVisitor should not drive visit of file. Use leaveFile() instead.");
+    }
```
alban-auzeill, line 70: Using `import static org.assertj.core.api.Assertions.assertThatThrownBy; `, I would write:
```suggestion
    assertThatThrownBy(() -> visitor.scanFile(context))
      .isInstanceOf(UnsupportedOperationException.class)
      .hasMessage("IssuableSubscriptionVisitor should not drive visit of file. Use leaveFile() instead.");
```

alban-auzeill (approved):

java-frontend/src/test/java/org/sonar/plugins/java/api/IssuableSubscriptionVisitorTest.java
Wohops: I wonder if we don't already have a rule saying to do that...

PR #4646, opened 2024-01-29
java-custom-rules-example should only be downloaded when there are Ja…

agigleux:
Update the java-custom-rules-example to illustrate the use of the <requiredForLanguages> property introduced with SonarQube 10.4

With this property, the custom plugin is only downloaded from the server when there are files matching the language of the plugin: Java files. It's a mandatory property to benefit from this enhancement globally at the server level. Without it, if there are Java files in the project to scan and this property is not set, the server will crash saying that there is a plugin not compatible. This is due to the fact that the custom Java plugins have a strong dependency on the parent analyzer (SonarJava) providing the APIs.


johann-beleites-sonarsource:
Note that this change is already present in #4644

agigleux:
Perfect. I guess I can just close this PR and wait for https://github.com/SonarSource/sonar-java/pull/4644 to be merged

PR #4647, opened 2024-01-30
SONARJAVA-4750 Use "Eclipse Public License - v 2.0" in the jdt module to allow JDT changes

alban-auzeill:


alban-auzeill:
Broken PR, recreate a new one

PR #4651, opened 2024-02-12
SONARJAVA-4839 Add S6904: Avoid using FetchType.EAGER

johann-beleites-sonarsource:


java-checks-test-sources/default/src/main/java/checks/JpaEagerFetchTypeCheckSample.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @OneToOne // Compliant
+  private String baz3;
+
+  @CustomAnnotation(fetch = CustomFetchType.EAGER) // Compliant
+  private String custom;
+
+  @CustomAnnotation(EAGER) // Compliant
+  private String customNoAssignment;
+}
```
kaufco, line 64: If we add two more test cases, this reveals an inconsistency: either we let the rule report also on custom annotations. Then the case can happen that a `FetchType` argument is not an assignment and line 62 in your samples should be `// Noncompliant` instead of `// Compliant`. 

_OR:_ we might choose to ignore custom annotations. Then sample line 62 is OK, but my added sample `@CustomAnnotationSwapped(fetch = EAGER) // Noncompliant` should then also be compliant instead.

Since we do not want to manage a list of all non-custom annotations that take a `FetchType` as an argument, I would suggest the implementation should not check for assignment arguments only.

```Java
  @Target(ElementType.FIELD)
  private @interface CustomAnnotationSwapped {
    CustomFetchType value() default CustomFetchType.LAZY;
    FetchType fetch() default FetchType.LAZY;
  }
  
  @CustomAnnotationSwapped(fetch = EAGER) // Noncompliant
  private String customSwapped;

  @CustomAnnotationSwapped(CustomFetchType.EAGER) // Compliant
  private String customNoAssignmentSwapped;
```

java-checks/src/main/java/org/sonar/java/checks/JpaEagerFetchTypeCheck.java
```diff
@@ -0,0 +1,80 @@
...
+        return null;
+      }
+
+      if ("EAGER".equals(fetchType.name()) &&
+        FETCH_TYPE_ENUMS.contains(assignedExpr.symbolType().fullyQualifiedName())) {
+        return fetchType;
+      }
+    }
+
+    return null;
```
kaufco, line 78: I think the implementation should not assume that a `FetchType` annotation argument is always an assignment argument, since you can also have custom annotations that have `FetchType value()` (for an example test case see the other comment).

kaufco (changes requested):
Nice implementation! I think one small addition is needed. For consistency reasons, the implementation should:
- also include non-assignment arguments (a good option), _or_
- ignore custom annotations at all (not a good option), _or_
- explicitly check that the parameter name is `fetch` and also mention that in the RSPEC (also a good option).

kaufco (approved):
Lgtm!

PR #4652, opened 2024-02-12
SONARJAVA-4841 Add S6898:  Avoid high frame rates on mobile

johann-beleites-sonarsource:


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/Sonar_way_profile.json
```diff
@@ -499,6 +499,7 @@
     "S6857",
     "S6862",
     "S6863",
+    "S6898",
```
ADarko22, line 502: should remove it from sonar way now or on a following effort?

ADarko22 (approved):
Look good! Nice the corner case that tests that non-final non-static fields can lead to FN!

Small comment about removing it from Sonar Way

PR #4654, opened 2024-02-13
SONARJAVA-4833 - Implement S6889: Proper Sensor Resource Management

ADarko22:


java-checks/src/main/java/org/sonar/java/checks/ReleaseSensorsCheck.java
```diff
@@ -0,0 +1,115 @@
...
+        return methodMatchers.matches((MethodInvocationTree) tree);
+      case NEW_CLASS:
+        return methodMatchers.matches((NewClassTree) tree);
+      default:
+        return false;
+    }
+  }
+
+  @Override
+  public void leaveFile(JavaFileScannerContext context) {
```
johann-beleites-sonarsource, line 108: How about initializing in `setContext` instead and only clearing in `leaveFile`? This would avoid errors when rule execution runs into a problem mid-file, exiting with an exception and hence not executing `leaveFile`.

johann-beleites-sonarsource (changes requested):
Nice work, a couple of mostly small suggestions

java-checks/src/main/java/org/sonar/java/checks/ReleaseSensorsCheck.java
```diff
@@ -0,0 +1,115 @@
...
+ * <p>
+ * The implemented logic checks that after acquiring a sensor there is at least one release invocation within the same file.
+ */
+@Rule(key = "S6889")
+public class ReleaseSensorsCheck extends IssuableSubscriptionVisitor {
+
+  enum AcquireReleaseSensor {
+    LOCATION_MANAGER("android.location.LocationManager", "requestLocationUpdates", "removeUpdates"),
+    SENSOR_MANAGER("android.hardware.SensorManager", "registerListener", "unregisterListener"),
+    CAMERA("android.hardware.Camera", "open", RELEASE),
```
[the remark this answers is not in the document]
ADarko22: nice catch!
I had a look and in order to cover it we need a slightly different logic. Let's discuss it since the acquire is done on `CameraManager.openCamera()` and the release should be done invoking `CameraDevice.close()` in the callback provided to `CameraManager.openCamera()`

java-checks/src/main/java/org/sonar/java/checks/ReleaseSensorsCheck.java
ADarko22: It's a very good point! I'll keep it in mind for the future :)

java-checks/src/main/java/org/sonar/java/checks/ReleaseSensorsCheck.java
```diff
@@ -55,17 +60,34 @@ enum AcquireReleaseSensor {
...
+    }
   }
 
   private static class AcquireReleaseStatus {
     List<Tree> acquireInvocations = new LinkedList<>();
     boolean released = false;
   }
 
+  public static final String ACQUIRE = "acquire";
   public static final String RELEASE = "release";
```
johann-beleites-sonarsource, line 76: Since we only use them here and wouldn't expect other rules to use constants from this rule, I would make them private.

java-checks/src/main/java/org/sonar/java/checks/ReleaseSensorsCheck.java
```diff
@@ -55,17 +60,34 @@ enum AcquireReleaseSensor {
...
+    initStatuses();
+  }
+
+  @Override
+  public void leaveFile(JavaFileScannerContext context) {
+    Arrays.stream(statuses)
+      .filter(status -> !status.released)
+      .forEach(status -> status.acquireInvocations.forEach(mit -> reportIssue(mit, "Make sure to release this sensor after use.")));
+
     initStatuses();
```
johann-beleites-sonarsource, line 91: Since we do the init work in the `setContext` now, we can simply set the field to `null` here - otherwise, we will initialize the field twice in most cases.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S6889.html
```diff
@@ -9,11 +9,16 @@ <h2>Why is this an issue?</h2>
     <ul>
```
johann-beleites-sonarsource, line 9: Actually, I think this list is not complete - `VirtualDisplay` is missing, for instance

johann-beleites-sonarsource (commented):
Just a couple of minor suggestions

java-checks/src/main/java/org/sonar/java/checks/ReleaseSensorsCheck.java
ADarko22: good one! The auto-extract is tricky XD

johann-beleites-sonarsource (approved):

PR #4655, opened 2024-02-13
SONARJAVA-4840 - implement S6905: SQL queries should retrieve only necessary fields

ADarko22:


java-checks/src/main/java/org/sonar/java/checks/QueryOnlyRequiredFieldsCheck.java
```diff
@@ -0,0 +1,99 @@
...
+
+  public static final String JAVA_LANG_STRING = "java.lang.String";
+
+  private static final MethodMatchers SQL_QUERY_METHODS = MethodMatchers.create()
+    .ofTypes("java.sql.Connection", "java.sql.Statement", "java.sql.PreparedStatement", "java.sql.CallableStatement")
+    .names("prepareStatement", "prepareCall", "execute", "executeQuery")
+    .addParametersMatcher(JAVA_LANG_STRING)
+    .addParametersMatcher(JAVA_LANG_STRING, "int")
+    .addParametersMatcher(JAVA_LANG_STRING, "int[]")
+    .addParametersMatcher(JAVA_LANG_STRING, "java.lang.String[]")
```
kaufco, line 57: We are not interested in the actual signature, only in the fact that the first argument is a string, so this replacement would be more performant:

```suggestion
    .addParametersMatcher(types -> !types.isEmpty() && types.get(0).is("java.lang.String"))
```

java-checks/src/main/java/org/sonar/java/checks/QueryOnlyRequiredFieldsCheck.java
```diff
@@ -0,0 +1,99 @@
...
+    return List.of(Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodInvocationTree mit = (MethodInvocationTree) tree;
+
+    if (SQL_QUERY_METHODS.matches(mit)) {
+      mit.arguments().stream()
+        .map(QueryOnlyRequiredFieldsCheck::extractQuery)
```
kaufco, line 73: I'd suggest to refactor this in favor of `ExpressionUtils::resolveAsConstant` over own implementation `(QueryOnlyRequiredFieldsCheck::extractQuery` (see comment below).

kaufco (changes requested):
The biggest thing we need to think about IMO is the constant resolving and that we don't want a custom implementation in each rule. This rule is a good showcase for that. I'd rather suggest to use `ExpressionUtils::resolveAsConstant` and extend it if needed (in this case: by re thesolving of local variables), from which all constant-resolving rules would profit at once. 

Maybe we should discuss this also in a greater round with the people involved in this sprint.

java-checks/src/main/java/org/sonar/java/checks/QueryOnlyRequiredFieldsCheck.java
ADarko22: very nice suggestion!

java-checks/src/main/java/org/sonar/java/checks/QueryOnlyRequiredFieldsCheck.java
```diff
@@ -0,0 +1,99 @@
...
+        .filter(Objects::nonNull)
+        .filter(arg -> SELECT_FROM_REGEXP.test(arg.value()))
+        .forEach(arg -> reportIssue(arg, "Don't use the query \"SELECT *\"."));
+    }
+  }
+
+  private static LiteralTree extractQuery(ExpressionTree expression) {
```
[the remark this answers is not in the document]
ADarko22: Agreed to keep it for now, and re-discuss when DBD will be available for re-implementing these rules

java-checks/src/main/java/org/sonar/java/checks/QueryOnlyRequiredFieldsCheck.java
ADarko22: Discussed

java-checks-test-sources/default/src/main/java/checks/QueryOnlyRequiredFieldsCheckSample.java
```diff
@@ -0,0 +1,53 @@
...
+    statement.execute("SELeCT * FROM table", new String[0]); // Noncompliant
+    statement.executeQuery("select * FROM table"); // Noncompliant
+
+    String noSqlCompiliant = "SELECTABLE * FROM table";
+    statement.execute(noSqlCompiliant);
+    statement.execute("SELECT* FROM table");
+    statement.execute("SELECT *FROM table");
+
+    String requestNonCompiliant2 = requestNonCompiliant;
+    preparedStatement.executeQuery(requestNonCompiliant2); // False Negative
```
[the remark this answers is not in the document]
kaufco: As discussed: we keep it because we don't use `ExpressionUtils.resolveAsConstant`, but have an own implementation in this rule for which it makes sense to test it.

kaufco (changes requested):
Nice work! Autoscan is failing though (I think it didn't do that yesterday), could you please have a look? 

Test coverage: the `isEmpty()` case in the line below is not covered, but that's OK. We shouldn't use `types.iterator().next()` just to increase the coverage, because we want the method matcher performant and should therefore avoid extra object creation here.
```
.addParametersMatcher(types -> !types.isEmpty() && types.get(0).is("java.lang.String"))
```

kaufco (approved):
Lgtm!

PR #4656, opened 2024-02-13
SONARJAVA-4831 - Implement S6891: Avoid exact alarms

kaufco:


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/Sonar_way_profile.json
```diff
@@ -498,6 +498,7 @@
     "S6856",
     "S6857",
     "S6862",
-    "S6863"
+    "S6863",
+    "S6891"
```
johann-beleites-sonarsource, line 502: Do we want this in Sonar way after all?

java-frontend/src/main/java/org/sonar/java/matcher/MethodMatchersBuilder.java
```diff
@@ -255,7 +255,7 @@ private boolean isSearchedMethod(Symbol.MethodSymbol symbol, @Nullable Type call
       typePredicate.test(type);
   }
 
-  private static IdentifierTree getIdentifier(MethodInvocationTree mit) {
+  public static IdentifierTree getIdentifier(MethodInvocationTree mit) {
```
johann-beleites-sonarsource, line 258: You're looking for `ExpressionUtils.methodName()` - this can stay private.

sonarpedia.json
```diff
@@ -3,7 +3,7 @@
   "languages": [
     "JAVA"
   ],
-  "latest-update": "2023-12-18T09:12:31.787089Z",
+  "latest-update": "2024-02-13T17:35:17.884561921Z",
```
johann-beleites-sonarsource, line 6: Is it possible you updated the metadata at some point? I think this is only updated when running a metadata update, which isn't consistent with the changes in this PR.

java-checks/src/main/java/org/sonar/java/checks/sustainability/AndroidExactAlarmCheck.java
```diff
@@ -0,0 +1,80 @@
...
+    switch (MethodMatchersBuilder.getIdentifier(tree).name()) {
+      case SET_EXACT_NAME:
+      case SET_EXACT_AND_ALLOW_WHILE_IDLE_NAME:
+        onSetExactInvocation(tree);
+        break;
+      case SET_WINDOW_NAME:
+        onSetWindowInvocation(tree);
+        break;
+      default:
+        throw new IllegalArgumentException();
```
johann-beleites-sonarsource, line 60: How about adding an error message here? Alternatively logging an error instead of throwing an exception?

java-checks/src/main/java/org/sonar/java/checks/sustainability/AndroidExactAlarmCheck.java
```diff
@@ -0,0 +1,80 @@
...
+    reportIssue(identifier, String.format("Use \"%s\" instead of \"%s\".", replacementName, methodName));
+  }
+
+  private void onSetWindowInvocation(MethodInvocationTree tree) {
+    if (tree.arguments().size() < 3) return;
+    var windowLengthMillisArg = tree.arguments().get(2);
+    var windowLengthMillis = ExpressionUtils.resolveAsConstant(windowLengthMillisArg);
+
+    if (windowLengthMillis instanceof Number && ((Number) windowLengthMillis).longValue() < SUGGESTED_MIN_LENGTH_MILLIS) {
+      reportIssue(windowLengthMillisArg, "Don't use alarm windows below 10 minutes.");
```
johann-beleites-sonarsource, line 77: I would suggest making this statement positive, so something along the lines of "Use alarm windows of 10 minutes or more instead."

java-checks/src/test/java/org/sonar/java/checks/sustainability/AndroidExactAlarmCheckTest.java
```diff
@@ -0,0 +1,46 @@
...
+  import org.sonar.java.checks.verifier.CheckVerifier;
+import org.sonar.plugins.java.api.InputFileScannerContext;
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
+
+import static org.sonar.java.checks.verifier.TestUtils.mainCodeSourcesPath;
+
+class AndroidExactAlarmCheckTest {
+
+  @Test
+  void test() {
```
johann-beleites-sonarsource, line 40: Since we are cheating in our test sources for Android rules, it may be worth adding an extra test to test that nothing is raised without semantics. It should work out of the box, you can use the same file you have anyway. Not a huge point, happy to discuss.

java-checks-test-sources/default/src/main/java/checks/sustainability/AndroidExactAlarmCheckSample.java
```diff
@@ -0,0 +1,97 @@
...
+    var noAlarmManager = new NoAlarmManager();
+    noAlarmManager.setWindow(0, triggerTime, windowLengthMillisTooShort, "", listener, targetHandler); // Compliant
+  }
+
+  public void setWindow(AlarmManager alarmManager, long triggerTime, Executor executor, AlarmManager.OnAlarmListener listener) {
+    alarmManager.setWindow(0, triggerTime, windowLengthMillisTooShort, "", executor, listener); // Noncompliant
+    alarmManager.setWindow(0, triggerTime, windowLengthMillisLongEnough, "", executor, listener); // Compliant
+
+    var noAlarmManager = new NoAlarmManager();
+    noAlarmManager.setWindow(0, triggerTime, windowLengthMillisTooShort, "", executor, listener); // Cmpliant
```
johann-beleites-sonarsource, line 64:
```suggestion
    noAlarmManager.setWindow(0, triggerTime, windowLengthMillisTooShort, "", executor, listener); // Compliant
```

java-checks-test-sources/default/src/main/java/checks/sustainability/AndroidExactAlarmCheckSample.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  public void setWindow(AlarmManager alarmManager, long triggerTime, PendingIntent pendingIntent) {
+    alarmManager.setWindow(0, triggerTime, 300000, pendingIntent); // Noncompliant [[sc=44;ec=50]] {{Don't use alarm windows below 10 minutes.}}
+    alarmManager.setWindow(0, triggerTime, 600000, pendingIntent); // Compliant
+    alarmManager.setWindow(0, triggerTime, 900000, pendingIntent); // Compliant
+    alarmManager.setWindow(0, triggerTime, 300000L, pendingIntent); // Noncompliant
+    alarmManager.setWindow(0, triggerTime, 600000L, pendingIntent); // Compliant
+    alarmManager.setWindow(0, triggerTime, 900000L, pendingIntent); // Compliant
+    alarmManager.setWindow(0, triggerTime, windowLengthMillisTooShort, pendingIntent); // Noncompliant [[sc=44;ec=70]] {{Don't use alarm windows below 10 minutes.}}
```
johann-beleites-sonarsource, line 41:
```suggestion
    alarmManager.setWindow(0, triggerTime, windowLengthMillisTooShort, pendingIntent); // Noncompliant
```

johann-beleites-sonarsource (changes requested):
Nice, a few mostly minor suggestions

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/Sonar_way_profile.json
kaufco: Yes, you're right, we don't want it. Fixed it in this file and also in the RSPEC.

sonarpedia.json
kaufco: Yes, accidentally, instead of generating the metadata for S6891 only. Reverted it.

java-checks/src/main/java/org/sonar/java/checks/sustainability/AndroidExactAlarmCheck.java
kaufco: I don't think it's needed: this is just to make the `switch` clause exhaustive. It will never happen because the `MethodMatchers` have already made sure it's one of those three functions. If it still happens, this is a programming mistake (e.g., one more matching function added to the `MethodMatchers`, but not to this switch statement). I think, for this purpose, no logging and just an exception (without a message) to disrupt the control flow is OK.

java-checks/src/test/java/org/sonar/java/checks/sustainability/AndroidExactAlarmCheckTest.java
kaufco: Added check for android context in report function + second test `withAndroidContext(false)`.

sonarpedia.json
johann-beleites-sonarsource, line 1: Just to avoid noise, maybe remove this file entirely from the PR - the IDE seems to have auto-added a new line at the end here.

johann-beleites-sonarsource (approved):
Looks good!

PR #4657, opened 2024-02-15
SONARJAVA-4855 Implement S6926: Bluetooth should be configured to use low power

ADarko22:


java-checks/src/main/java/org/sonar/java/checks/BluetoothLowPowerModeCheck.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodInvocationTree mit = (MethodInvocationTree) tree;
+
+    if ((REQUEST_CONNECTION_PRIORITY.matches(mit) && firstArgumentNotEqualsTo(mit, CONNECTION_PRIORITY_LOW_POWER))
+      || (SET_ADVERTISE_MODE.matches(mit) && firstArgumentNotEqualsTo(mit, ADVERTISE_MODE_LOW_POWER))) {
+
+      reportIssue(mit.methodSelect(), "Use the low power mode for this Bluetooth operation.");
```
ADarko22, line 61: Quick fixes can be implemented relatively easily, would they be valuable for this rule?
In general the rule may be prone to FPs.

johann-beleites-sonarsource: Hmm fair question. If the rule is prone to FPs, I think it is fine to not spend time on it at this point and instead focus on implementing other, maybe more valuable, rules instead. If you think it is valuable and quickly done, feel free to add it. Given our doubts about some of these rules, I personally would probably not add it right now.

java-checks/src/main/java/org/sonar/java/checks/BluetoothLowPowerModeCheck.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Optional;
+import org.sonar.check.Rule;
+import org.sonar.java.model.ExpressionUtils;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S6926")
+public class BluetoothLowPowerModeCheck extends IssuableSubscriptionVisitor {
```
johann-beleites-sonarsource, line 32: Did you know that for rules that only visit function calls, you can use the more specific `AbstractMethodDetection` to extend from?

java-checks/src/main/java/org/sonar/java/checks/BluetoothLowPowerModeCheck.java
```diff
@@ -0,0 +1,73 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodInvocationTree mit = (MethodInvocationTree) tree;
+
+    if ((REQUEST_CONNECTION_PRIORITY.matches(mit) && firstArgumentNotEqualsTo(mit, CONNECTION_PRIORITY_LOW_POWER))
+      || (SET_ADVERTISE_MODE.matches(mit) && firstArgumentNotEqualsTo(mit, ADVERTISE_MODE_LOW_POWER))) {
```
johann-beleites-sonarsource, line 59: WDYT of only comparing the method name here (`ExpressionUtils.methodName()`) instead of matching the entire method matcher again?

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S6926.json
```diff
@@ -0,0 +1,24 @@
...
+    "sustainability"
+  ],
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-6926",
+  "sqKey": "S6926",
+  "scope": "Main",
+  "quickfix": "unknown",
+  "code": {
+    "impacts": {
+      "MAINTAINABILITY": "LOW"
```
johann-beleites-sonarsource, line 20: Is it really maintainability that is impacted? Shouldn't it be sustainability?

java-checks-test-sources/default/src/main/java/checks/BluetoothLowPowerModeCheckSample.java
```diff
@@ -0,0 +1,30 @@
...
+import android.bluetooth.le.AdvertiseSettings;
+
+public class BluetoothLowPowerModeCheckSample {
+
+  private static int connectionPriorityHigh = 1;
+  private int advertiseModeBalanced = 1;
+
+
+  public void nonCompliant(android.bluetooth.BluetoothGatt gatt, android.bluetooth.le.AdvertiseSettings.Builder builder) {
+    gatt.requestConnectionPriority(BluetoothGatt.CONNECTION_PRIORITY_HIGH); // Noncompliant {{Use the low power mode for this Bluetooth operation.}}
```
johann-beleites-sonarsource, line 13: Perhaps add `[[sc=X;ec=Y]]` in one test case

johann-beleites-sonarsource (changes requested):
Nice, a few mostly minor suggestions

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S6926.json
ADarko22: Since there is no sustainability impact defined, it was agreed with Alex to set maintainability=LOW and have a sustainability tag.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/Sonar_way_profile.json
```diff
@@ -498,6 +498,9 @@
     "S6856",
     "S6857",
     "S6862",
-    "S6863"
+    "S6863",
+    "S6863",
+    "S6889",
+    "S6904"
```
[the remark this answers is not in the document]
ADarko22: I'll remove them even if these rules are valid also outside the sustainability domain :)

java-checks/src/main/java/org/sonar/java/checks/BluetoothLowPowerModeCheck.java
ADarko22: Thanks!

java-checks/src/main/java/org/sonar/java/checks/BluetoothLowPowerModeCheck.java
ADarko22: Makes sense :)

PR #4662, opened 2024-02-16
Fix Quality Flaws S6204: Replace '.collect(Collectors.toList())' with  '.toList()'

alban-auzeill:


Wohops (approved):
I did another pass to remove all unused imports, plus added a few changes to explicitly use a mutable collection.

PR #4672, opened 2024-02-19
SONARJAVA-4854 Implement S6923: Motion Sensor should not use gyroscope

ADarko22:

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/UseMotionSensorWithoutGyroscopeCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/UseMotionSensorWithoutGyroscopeCheckSample.java
@@ -1,0 +1,23 @@
+package checks;
+
+import android.hardware.Sensor;
+import android.hardware.SensorManager;
+
+public class UseMotionSensorWithoutGyroscopeCheckSample {
+
+  void compliant(SensorManager sensorManager) {
+    sensorManager.getDefaultSensor(android.hardware.Sensor.TYPE_ROTATION_VECTOR); // Noncompliant {{Replace `TYPE_ROTATION_VECTOR` (11) with `TYPE_GEOMAGNETIC_ROTATION_VECTOR` (20) to optimize battery life.}}
+    sensorManager.getDefaultSensor(11); // Noncompliant [[sc=5;ec=39]]
+    sensorManager.getDefaultSensor(android.hardware.Sensor.TYPE_ROTATION_VECTOR, true); // Noncompliant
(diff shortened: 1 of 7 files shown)


java-checks-test-sources/default/src/main/java/checks/UseMotionSensorWithoutGyroscopeCheckSample.java
```diff
@@ -0,0 +1,23 @@
+package checks;
+
+import android.hardware.Sensor;
+import android.hardware.SensorManager;
+
+public class UseMotionSensorWithoutGyroscopeCheckSample {
+
+  void compliant(SensorManager sensorManager) {
+    sensorManager.getDefaultSensor(android.hardware.Sensor.TYPE_ROTATION_VECTOR); // Noncompliant {{Replace `TYPE_ROTATION_VECTOR` (11) with `TYPE_GEOMAGNETIC_ROTATION_VECTOR` (20) to optimize battery life.}}
+    sensorManager.getDefaultSensor(11); // Noncompliant [[sc=5;ec=39]]
```
johann-beleites-sonarsource, line 10: Out of curiosity - why do you test the `[[sc=X;ec=X]]` and `{{Message}}` with two separate test cases?

johann-beleites-sonarsource (approved):
Nice!

java-checks/src/main/java/org/sonar/java/checks/UseMotionSensorWithoutGyroscopeCheck.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  protected MethodMatchers getMethodInvocationMatchers() {
+    return GET_DEFAULT_SENSOR_MATCHER;
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    var sensorType = (Integer) ExpressionUtils.resolveAsConstant(mit.arguments().get(0));
+    if (sensorType!= null && sensorType == TYPE_ROTATION_VECTOR) {
```
ADarko22, line 46: the null check can be tested by adding a fake `getDefaultSensor()` with no arg, in the mocked `android.hardware.SensorManager`,  to make the test coverage happy, but it would be fictitious since such method doesn't exist. Is it fine to have <95% coverage in this case?

java-checks-test-sources/default/src/main/java/checks/UseMotionSensorWithoutGyroscopeCheckSample.java
ADarko22: No specific reason, it may feel more readable since the comment tends to be long :)

PR #4676, opened 2024-02-20
SONARJAVA-4831 Fix QG issues due to move from Java 11 to Java 17

kaufco:


java-checks/src/main/java/org/sonar/java/checks/sustainability/AndroidExactAlarmCheck.java
```diff
@@ -49,15 +49,9 @@ protected MethodMatchers getMethodInvocationMatchers() {
   @Override
   protected void onMethodInvocationFound(MethodInvocationTree tree) {
     switch (ExpressionUtils.methodName(tree).name()) {
```
Wohops, line 51: I think I would prefer to go to something even simpler in our case:
```java
if (SET_WINDOW_NAME.equals(tree.methodSymbol().name())) {
  onSetWindowInvocation(tree);
} else {
  onSetExactInvocation(tree);
}
```
1. this prevents the uncovered line
2. it does not require `ExpressionUtils` and relies directly on semantic

kaufco: Yes, indeed. Fixed.

java-checks/src/main/java/org/sonar/java/checks/sustainability/AndroidExactAlarmCheck.java
```diff
@@ -73,7 +67,7 @@ private void onSetWindowInvocation(MethodInvocationTree tree) {
     var windowLengthMillisArg = tree.arguments().get(2);
     var windowLengthMillis = ExpressionUtils.resolveAsConstant(windowLengthMillisArg);
```
Wohops, line 68: Sorry, I just realized that all this can also be rewritten in a much simpler way:
```java
windowLengthMillisArg
  .asConstant(Long.class)
  .filter(value -> value < SUGGESTED_MIN_LENGTH_MILLIS)
  .ifPresent(value -> reportIfInAndroidContext(windowLengthMillisArg, "Use alarm windows of 10 minutes or more instead."));
```

Wohops (approved):

PR #4680, opened 2024-02-21
SONARJAVA-4851 Implement S6912: Use batch Processing in JDBC

ADarko22:


java-checks/src/main/java/org/sonar/java/checks/BatchSQLStatementsCheck.java
```diff
@@ -0,0 +1,121 @@
...
+    var invocationsCollector = new StatementExecuteInvocationsCollector();
+    methodTree.accept(invocationsCollector);
+
+    var inLambdaInvocationsCollector = new InLambdaStatementExecuteInvocationsCollector();
+    methodTree.accept(inLambdaInvocationsCollector);
+
+    if (invocationsCollector.shouldReport() || inLambdaInvocationsCollector.shouldReport()) {
+      invocationsCollector.invocations().forEach(invocation -> reportIssue(invocation, MESSAGE));
+      inLambdaInvocationsCollector.invocations().forEach(invocation -> reportIssue(invocation, MESSAGE));
+    }
```
johann-beleites-sonarsource, line 67: These seem like two disjoint pieces of logic to me: 1 for the `invocationsCollector`, one for the `inLambdaInvocationsCollector`. I think it would be a little more readable to have them in two separate if-statements.

java-checks/src/main/java/org/sonar/java/checks/BatchSQLStatementsCheck.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  private static class StatementExecuteInvocationsCollector extends BaseTreeVisitor {
+    private static final Set<Tree.Kind> LOOP_KINDS = EnumSet.of(
+      Tree.Kind.FOR_STATEMENT,
+      Tree.Kind.FOR_EACH_STATEMENT,
+      Tree.Kind.WHILE_STATEMENT,
+      Tree.Kind.DO_STATEMENT);
+    private final List<MethodInvocationTree> invocations = new LinkedList<>();
```
johann-beleites-sonarsource, line 76: Why a LinkedList here? In most scenarios an ArrayList turns out to be more efficient.

java-checks/src/main/java/org/sonar/java/checks/BatchSQLStatementsCheck.java
```diff
@@ -0,0 +1,121 @@
...
+
+    var invocationsCollector = new StatementExecuteInvocationsCollector();
+    methodTree.accept(invocationsCollector);
+
+    var inLambdaInvocationsCollector = new InLambdaStatementExecuteInvocationsCollector();
+    methodTree.accept(inLambdaInvocationsCollector);
+
+    if (invocationsCollector.shouldReport() || inLambdaInvocationsCollector.shouldReport()) {
+      invocationsCollector.invocations().forEach(invocation -> reportIssue(invocation, MESSAGE));
+      inLambdaInvocationsCollector.invocations().forEach(invocation -> reportIssue(invocation, MESSAGE));
```
johann-beleites-sonarsource, line 66: WDYT of using secondaries instead of reporting a separate issue on every single invocation?

java-checks/src/main/java/org/sonar/java/checks/BatchSQLStatementsCheck.java
johann-beleites-sonarsource, line 1: Looking at the ITs, it looks like we are seeing some FPs, e.g. `JDBCSessionDataStore.java` lines 818 and 852. You may not always be able to execute a second query without getting results from a first. I realize that this may be very difficult to detect. If you have a good idea, I'm happy to talk about it. Alternatively, I think it would also be okay to limit our scope to e.g. only statements in loops or reduce the window in which we consider queries, for instance only within the same block.

johann-beleites-sonarsource (changes requested):
I think we need to reduce our scope, as I can already see FPs in the ITs - see comments.

java-checks/src/main/java/org/sonar/java/checks/BatchSQLStatementsCheck.java
ADarko22: This logic is a bit problematic because there are scenarios that won't be covered:

- when there is a single invocation not in loop in both, neither of them will be reported (with current logic and separating the two if)
- when there is a single invocation not in loop in one and multiple invocations on the other, then the current logic will correctly report both

java-checks/src/main/java/org/sonar/java/checks/BatchSQLStatementsCheck.java
ADarko22: Very nice observation! We can discuss it. I was considering indeed to reduce the scope to loops only. The idea of same block is very interesting!

java-checks/src/main/java/org/sonar/java/checks/BatchSQLStatementsCheck.java
ADarko22: The linked list is more efficient for append-only operations, like in this case. The reading operations are on the whole list and not in random access, so it should be a better choice here.

java-checks/src/main/java/org/sonar/java/checks/BatchSQLStatementsCheck.java
ADarko22: I changed the rule scope to only loop instructions, so there is no need to raise secondary locations anymore :)

java-checks/src/main/java/org/sonar/java/checks/BatchSQLStatementsCheck.java
ADarko22: Changed the implementation to report only on loops

johann-beleites-sonarsource (approved):
LGTM!

PR #4681, opened 2024-02-21
SONARJAVA-4845 S6914: Use Fused Location to optimize battery power

kaufco:


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/Sonar_way_profile.json
```diff
@@ -501,6 +501,7 @@
     "S6863",
     "S6889",
     "S6905",
-    "S6909"
+    "S6909",
+    "S6914"
```
ADarko22, line 505: Should we keep this rule in the Sonar Way? Let's double-check with @johann-beleites-sonarsource.
I think this is a valid rule, although forces to use Google API which may not be available on all devices.

ADarko22 (changes requested):
Good job! Left a small comment to cover invocations in lambdas

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/Sonar_way_profile.json
kaufco: I agree: should not be in Sonar Way profile.

ADarko22 (approved):
LGTM! 

Remember to update the RSPEC and remove the Sonar Way profile from there, otherwise the rule will be added to the profile when executing the `rule-api update`

PR #4682, opened 2024-02-21
SONARJAVA-4823 S6885: Add clamp methods to Math

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/MathClampMethodsCheck.java
```diff
@@ -0,0 +1,105 @@
...
+      && (tree.elseStatement() != null && tree.elseStatement().is(Tree.Kind.IF_STATEMENT))) {
+      var elseIfStatement = (IfStatementTree) tree.elseStatement();
+      if (isGreaterThanOrLessThan(elseIfStatement.condition())) {
+        reportIssue(tree, "Use \"Math.clamp\" instead of an if-else statement.");
+      }
+    }
+  }
+
+  private static boolean isGreaterThanOrLessThan(Tree tree) {
+    return tree.is(Tree.Kind.GREATER_THAN) || tree.is(Tree.Kind.LESS_THAN);
```
alban-auzeill, line 102: We should consider GREATER_THAN_OR_EQUAL_TO like GREATER_THAN.

And be careful:
`Math.max(nextValue(), Math.min(nextValue(), value))`==
`Math.clamp(value, nextValue(), nextValue())`
But
`Math.min(nextValue(), Math.max( nextValue(), value))` !=
`Math.clamp(value, nextValue(), nextValue())`

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/MathClampMethodsCheck.java
```diff
@@ -0,0 +1,183 @@
...
+
+  private void checkIfStatement(IfStatementTree tree) {
+    if (isGreaterThanOrLessThan(tree.condition())
+      && (tree.elseStatement() != null && tree.elseStatement().is(Tree.Kind.IF_STATEMENT))) {
+      var elseIfStatement = (IfStatementTree) tree.elseStatement();
+      if (isGreaterThanOrLessThan(elseIfStatement.condition())) {
+        reportIssue(tree, IF_ELSE_STATEMENT_MESSAGE);
+      }
+    }
+  }
```
alban-auzeill, line 169: The logic about `if` statement is too basic. It will catch a lot of false-positives. For example:
```
    int a,b,c,d,e,f;
    if (a < b) { // Noncompliant, false positive
      System.out.println("no assignment");
    } else if (c < d) { // Noncompliant, false positive
      System.out.println("no assignment");
    } else if (e < f) {
      System.out.println("no assignment");
    }
```

java-checks-test-sources/default/src/main/java/checks/MathClampMethodsCheckSample.java
```diff
@@ -0,0 +1,98 @@
+package checks;
+
+public class MathClampMethodsCheckSample {
+
+  public void testNoncompliant(int min, int max, int value) {
+    int clampedValueConditional1 = value > max ? max : value < min ? min : value; // Noncompliant [[sc=36;ec=81]] {{Use "Math.clamp" instead of a conditional expression.}}
```
alban-auzeill, line 6: The comparison logic to analyze `value > max  ? max : (value < min ? min : value)` does not support some equivalent inversions like `value > max` <=> `max < value`. So we have those false-negatives:
```
    var testCases = new int[] {
      value > max ? max : (value < min ? min : value), // Noncompliant
      max < value ? max : (value < min ? min : value), // false-negative
      value <= max ? (value < min ? min : value) : max, // false-negative
      max >= value ? (value < min ? min : value) : max, // false-negative
      max >= value ? (min > value ? min : value) : max, // false-negative
      max >= value ? (value >= min ? value : min) : max, // false-negative
      value > max ? max : (min > value ? min : value), // Noncompliant
      value > max ? max : (value >= min ? value : min), // Noncompliant
      value > max ? max : (min <= value ? value : min) // Noncompliant
    };
```

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/MathClampMethodsCheck.java
alban-auzeill: Let's remove the `if` from the scope of this rule.

java-checks/src/main/java/org/sonar/java/checks/MathClampMethodsCheck.java
```diff
@@ -0,0 +1,159 @@
...
+      var falseExpression = skipParentheses(firstConditionalExpression.falseExpression());
+      if (shouldReportOnConditional(condition.rightOperand(), trueExpression, falseExpression, isGreater) ||
+        shouldReportOnConditional(condition.rightOperand(), falseExpression, trueExpression, !isGreater)) {
+        reportIssue(firstConditionalExpression, CONDITIONAL_EXPRESSION_MESSAGE);
+      }
+    }
+  }
+
+  private static boolean shouldReportOnConditional(ExpressionTree condition, ExpressionTree tree1, ExpressionTree tree2, boolean isMax) {
+    if (condition.is(Tree.Kind.IDENTIFIER) && tree1.is(Tree.Kind.IDENTIFIER) && (((IdentifierTree) condition).name().equals(((IdentifierTree) tree1).name()))) {
```
alban-auzeill, line 90: Comparing `String IdentifierTree.name()` should not create false-positives here, but it's not always the case and could confuse the reader. The safe way to test is with `!identifierTree1.symbol().isUnknown() && identifierTree1.symbol().equals(identifierTree2.symbol())`.
I didn't find a matching implementation in the helpers, only the opposite: `SyntacticEquivalence.areDifferentVariables`.
It could be useful to add one helper, like `ExpressionUtils.areSameVariables(Tree tree1, Tree tree2, boolean defaultValue)`. In this case, we could even support `Tree.Kind.MEMBER_SELECT`. It would be possible to check `this.field` is same as `field`.

Same comment for the bellow `checkInnerExpression` method.

alban-auzeill (changes requested):

leonardo-pilastri-sonarsource (approved):

PR #4683, opened 2024-02-22
SONARJAVA-4848 S6916 Use guard instead of a single if else in pattern match

leonardo-pilastri-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/SingleIfInsteadOfPatternMatchGuardCheck.java
```diff
@@ -0,0 +1,111 @@
...
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.CASE_GROUP);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    var caseGroup = (CaseGroupTree) tree;
+    var caseLabel = caseGroup.labels().get(0);
```
irina-batinic-sonarsource, line 57: WDYT if we move this check down?
I think we might optimize a bit..

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S6916.html
```diff
@@ -0,0 +1,66 @@
+<h2>Why is this an issue?</h2>
+<p>In Java 21 switch statements and expressions are extended to work on any type, and case labels now also allow patterns instead of just
```
irina-batinic-sonarsource, line 2: I would rewrite it a bit:

> Java 21 has introduced enhancements to switch statements and expressions, allowing them to operate on any type, not just specific ones, as in previous versions. Furthermore, case labels have been upgraded to support patterns, providing an alternative to the previous restriction of only accepting constants.

irina-batinic-sonarsource (changes requested):
Overall, the implementation looks good. Nice job!
I left two comments, feel free to have a look.

java-checks/src/test/java/org/sonar/java/checks/SingleIfInsteadOfPatternMatchGuardCheckTest.java
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java.checks;
+
+import org.junit.jupiter.api.Test;
+import org.sonar.java.checks.verifier.TestUtils;
+import org.sonar.java.checks.verifier.internal.InternalCheckVerifier;
+
+class SingleIfInsteadOfPatternMatchGuardCheckTest {
```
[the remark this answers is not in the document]
leonardo-pilastri-sonarsource: Before java 21 the code would not compile as you would not have those kind of guarded type patterns

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S6916.html
irina-batinic-sonarsource: PR: https://github.com/SonarSource/rspec/pull/3692

PR #4687, opened 2024-02-23
SONARJAVA-4438 S6204: recommendation not applicable when upcast is required

kaufco:


Wohops (changes requested):
Okay, so here is my review:

* Too many changes that are not related to the fix of the FP itself (formatting, new JDK17 constructs). I would keep such changes outside of the PR so we can focus on the essence, or add another commit AFTER the initial review of the functional fix. Can you revert them?
* I feel the code of the actual fix is hard to read and do not reflect the actual semantic requirements in the code to be able to use `toList()`. I suggested an alternative way of writing it.
* I believe the helper method is not needed (I could not find a case where it would be).
* Usages of `Objects.requireNonNull()` are never needed in the context of this rule, and only add extra cognitive complexity.

I would be happy to discuss this with you more in details

java-checks/src/main/java/org/sonar/java/checks/CollectorsToListCheck.java
```diff
@@ -91,13 +92,30 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
...
+      return true;
+    }
+    Type toListType = mit.symbolType();
+    if (toListType.isRawType()) {
+      return false;
+    }
+    Type toListArgType = toListType.typeArguments().get(0);
+    Type ownerArgType = ownerType.typeArguments().get(0);
+    return toListArgType.is(ownerArgType.fullyQualifiedName());
+  }
```
Wohops, line 117: Looking again at it, and thanks to your comment, I feel we can simplify it even more and make it more explicit!
1. If the `Stream` is NOT a raw type, it's semantically impossible to have the `collect` method returning a raw type. So we can skip the second test, we will simply never match it.
2. I changed the names of variables because I made a mistake in my previous suggestion (we are reacting on `collect`, not on `toList`) and we can make it explicit. The owner is a `Stream` also, so let's try to be as close as possible to the reality of the code.

```suggestion
  private static boolean isInvariantTypeArgument(MethodInvocationTree collectMethodInvocation) {
    Type streamType = collectMethodInvocation.methodSymbol().owner().type();
    if (streamType.isRawType()) {
      return true;
    }
    Type collectArgType = collectMethodInvocation.symbolType().typeArguments().get(0);
    Type streamArgType = streamType.typeArguments().get(0);
    return collectArgType.is(streamArgType.fullyQualifiedName());
  }
```

Wohops (changes requested):
Thanks for the update!

I feel it's 100% easier to review now!
I made a small suggestion on top of it for you to consider.

Once this change is done, I would:
* Squash all your changes in a single commit summarizing the fix
* Add an **optional** commit on top of it to refactor the rule if you believe it's still important and use the latest java 17 constructs  (I'm all in for "less is more" so I do not believe it's required). If you do so, I would however appreciate to look at it! We can also pair-program on it if you want.

Wohops (approved):
LGTM!

PR #4690, opened 2024-02-26
Reduce rule S6912 scope

ADarko22:
Reduce the rule scope to raise issues only on Statements executed in the loop body directly, i.e. not enclosed in other instructions.


java-checks-test-sources/default/src/main/java/checks/BatchSQLStatementsCheckSample.java
```diff
@@ -56,62 +65,29 @@ void nonCompliantRequestsInLoop(Statement statement, List<String> queries) throw
           // do nothing
         }
       });
-  }
-
-  void nonCompliantIterableForEach(PreparedStatement statement, List<String> queries) {
```
ADarko22, line 61: moved below

johann-beleites-sonarsource (approved):

PR #4691, opened 2024-02-26
SONARJAVA-4752 FP in S5665 when \\" and \\' are not intended to be escaped

kaufco:


java-checks-test-sources/default/src/main/java/checks/UnnecessaryEscapeSequencesInTextBlock.java
```diff
@@ -31,5 +31,54 @@ public void str() {
...
+
+    String question19 = """
+              \\\\\\\\\'What's the point, really\\\\\\\\\'
+              ?"""; // Noncompliant@-1 {{Remove this unnecessary escape sequence.}}
+
+    String question20 = """
+              What's
+              the 
+              point,
+              really\\n?"""; // Compliant
```
ADarko22, line 82: can you add the examples mentioned in the ticket and the community post?
```
String notEscapedDoubleQuote = """
                                "\\\\\\\\SERVER\\\\SHARE\\\\ETC\\\\" 
                                """;
String notEscapedSingleQuote = """
                                'C:\\\\foo\\\\bar\\\\' 
                                """;
```

ADarko22 (changes requested):
Let's add the example mentioned in the ticket to make sure the community users won't face anymore the reported FP.

ADarko22 (approved):
LGTM! :)

PR #4694, opened 2024-02-27
SONARJAVA-4827 S6877 SequencedCollection reversed view should be used instead of Collections.reverse

alban-auzeill:
Rspec PR to review and merge: https://github.com/SonarSource/rspec/pull/3690


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S6877.html
```diff
@@ -0,0 +1,43 @@
+<h2>Why is this an issue?</h2>
```
alban-auzeill, line 1: This content will need to be updated from RSPEC once https://github.com/SonarSource/rspec/pull/3690 is approved and merged.

leonardo-pilastri-sonarsource (approved):
The implementation looks good and the samples are exhaustive, if you want you can add some other comments on the code just to explain better what some methods do:
- matchReverseParentSafeScope
- matchAccessor
Having a full explanatory method name would be too long, so maybe a comment could do a better job. Up to you!

PR #4695, opened 2024-02-27
SONARJAVA-4749 Improve accuracy of S1170

johann-beleites-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -87,23 +86,28 @@ public void leaveNode(Tree tree) {
   private static boolean hasConstantInitializer(VariableTree variableTree) {
     ExpressionTree init = variableTree.initializer();
     if (init != null) {
```
kaufco, line 88: I think an early return (`if (init == null) {return;}`) might be better here, as it would reduce the nesting level of the inner block (cognitive complexity), and also, the contents of a nested block should not be too long.

java-checks/src/main/java/org/sonar/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -87,23 +86,28 @@ public void leaveNode(Tree tree) {
...
           return false;
         }
-      }
-      if (init.is(Tree.Kind.NEW_ARRAY)) {
+      } else if (deparenthesized.is(Tree.Kind.NEW_ARRAY)) {
         return false;
+      } else if (deparenthesized.is(Tree.Kind.IDENTIFIER)) {
+        var symbol = ((IdentifierTree) deparenthesized).symbol();
+        return symbol.isStatic() && symbol.isFinal();
       }
```
kaufco, line 101: Since we have now three cases, I think a `switch` statement would be better.

java-checks-test-sources/default/src/main/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -116,3 +129,16 @@ public void printCreation() {
   }
 }
 
+class FieldAssignments {
+  private final int[] finalFoo = new int[42]; // Compliant
+  private final Object finalObject = finalFoo; // Compliant - foo is not static
```
kaufco, line 134:
```suggestion
  private final Object finalObject = finalFoo; // Compliant - finalFoo is not static
```

java-checks-test-sources/default/src/main/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -116,3 +129,16 @@ public void printCreation() {
...
+
+  private static int[] staticFoo = new int[42]; // Compliant
+
+  private final Object staticObject = staticFoo; // Compliant - foo is not final
```
kaufco, line 138:
```suggestion
  private final Object staticObject = staticFoo; // Compliant - staticFoo is not final
```

java-checks-test-sources/default/src/main/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -79,6 +79,19 @@ static void someMethod2(Object o) {
   }
 }
 
+class AvoidFPsWhenUsingFieldInInstance_SONARJAVA_4749 {
+  private static long nextID = 1L;
+  private final long localID = nextID++; // Compliant
```
kaufco, line 84: I think we should check all 4: (prefix, postix) x (increment, decrement), since they are explicitly listed in the implementation

kaufco (changes requested):
I suggested a minor refactoring in the `hasConstantInitializer` function.

java-checks/src/main/java/org/sonar/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -86,32 +85,36 @@ public void leaveNode(Tree tree) {
...
+      return false;
+    }
+
+    var deparenthesized = ExpressionUtils.skipParentheses(init);
+
+    if (deparenthesized.is(Tree.Kind.METHOD_REFERENCE)) {
+      MethodReferenceTree methodRef = (MethodReferenceTree) deparenthesized;
+      if (isInstanceIdentifier(methodRef.expression())) {
         return false;
       }
```
kaufco, line 98: This suggestion might be better readable instead. I think in new code, we generally should prefer the `instanceof` with pattern matching approach if we only check against a single `Kind` and need the instance. We should use the "old" way only if:

- We check against multiple kinds, i.e.:
  - `switch` instead of `if`, or
  - `is` with multiple arguments
- We check only against a single kind, but don't need the cast operation

```suggestion
    if (deparenthesized instanceof MethodReferenceTree methodRef && isInstanceIdentifier(methodRef.expression())) {
      return false;
    }
```

java-checks/src/main/java/org/sonar/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -86,32 +85,36 @@ public void leaveNode(Tree tree) {
...
-
-  private static boolean isThisOrSuper(Tree tree) {
-    if (tree.is(Tree.Kind.IDENTIFIER)) {
-      String name = ((IdentifierTree) tree).name();
-      return "super".equals(name) || "this".equals(name);
-    }
-    return false;
+  private static boolean isNonStaticOrFinal(Tree tree) {
+    return switch (tree.kind()) {
+      case METHOD_INVOCATION, NEW_CLASS, NEW_ARRAY -> true;
```
kaufco, line 106: I just realized that there is one operation that returns a non-constant result even if both source operands are static final: array access. We can't make any assumptions over the contents even for static final arrays, so we need to add it to the non-static-or-final-case:

```suggestion
      case METHOD_INVOCATION, NEW_CLASS, NEW_ARRAY, ARRAY_ACCESS_EXPRESSION -> true;
```

java-checks-test-sources/default/src/main/java/checks/ConstantsShouldBeStaticFinalCheck.java
```diff
@@ -116,3 +136,16 @@ public void printCreation() {
...
+  private final int[] finalFoo = new int[42]; // Compliant
+  private final Object finalObject = finalFoo; // Compliant - finalFoo is not static
+
+  private static int[] staticFoo = new int[42]; // Compliant
+
+  private final Object staticObject = staticFoo; // Compliant - staticFoo is not final
+
+  private static final int[] foo = new int[42];
+  private final Object object = foo; // Noncompliant
+
```
kaufco, line 149: Additional test case for the array index fix (see below)

```suggestion
private final int bar = foo[2]; // Compliant

```

kaufco (changes requested):
I think there are only two more fixes needed:
1. Consider array index expressions (FP; see comments)
2. Pattern matching instead of `is(kind)` + cast

kaufco (approved):
Lgtm!

PR #4702, opened 2024-03-04
SONARJAVA-4422 S6204 FP on lists used outside their instantiation scope

kaufco:


java-checks/src/main/java/org/sonar/java/checks/CollectorsToListCheck.java
```diff
@@ -98,42 +107,82 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
...
-    if (list.parent().is(Tree.Kind.ARGUMENTS)) {
-      Tree potentialMethodInvocation = list.parent().parent();
-      if (potentialMethodInvocation.is(Tree.Kind.METHOD_INVOCATION)) {
-        MethodInvocationTree mit = (MethodInvocationTree) potentialMethodInvocation;
-        return COLLECTIONS_MUTATOR_METHODS.matches(mit);
+  private static boolean isPossibleMutatorExpression(ExpressionTree tree, Set<Symbol> graylist) {
+    var parent = tree.parent();
+    return switch (parent.kind()) {
+      case LAMBDA_EXPRESSION ->
+        // Don't perform deeper data flow analysis; assume receiver of lambda result might perform mutations
```
Wohops, line 139: I would not mention data flow analysis here... we are not looking at any flow here, just at the parent tree...

Wohops (changes requested):
Okay, I think we are going in the wrong direction in trying to fix this with this approach. 

I think we need to rethink the fix in order to make the code easier to understand, or rewrite the rule entirely, starting from another angle.

Right now, I keep losing track of what we actually try to achieve, and I strongly feel this is not OK. The code is too complex for what the rule is supposed to do. I would prefer to totally kill the noise and have the rule raising a very small percentage of issues, rather than trying to cover all possible scenarios at the cost of 300 way too complicated LOC.

Some ideas of amelioration I see to improve readability:
* method parameters names are way too generic: `tree`, `symbol`, etc... we are not dealing with any tree or symbol here, we are speaking of `usageOfVariables`, `variableAsList`, etc... The names should reflect the behavior of the rule and be closer from its business logic. Same remark for the grayList and whiteList (of what?).
* some operations are counterintuitive and it reenforces my feeling that the rule "falls into somehow working" without me understanding why. In particular, the method `isPossibleMutatorExpression()` starts its body by doing `tree.parent()`. While it makes sense to do so when we arrive from the starting `collect(...)` method invocation, it absolutely does not when coming looking at the identifiers of the usage list! usages could be absolutely any tree, anywhere in the AST.

kaufco:
We leave the rule as it is due to the very low FP rate it produces, which is better than eliminating these FPs at the cost of eliminating many TPs as well.

Two main aspects have been discussed:

1. Implementation complexity: Most of the implementation is not about the rule logic itself - which is rather trivial - but about detecting, if and where in the data flow the list reference leaves the scope and might be passed to another receiver. We scan the tree for `yield` statements, `return` statements, assignments, method arguments etc. While the implementation seems correct, it should not be the concern of the rule to inspect the data flow on its own at this complexity level. This functionality should rather be provided by a data flow engine; but SE or DBD are not an option at the moment.

2. FP / TP rate: Currently, only about 0.2% of S6204 reports are flagged as FPs. The implementation would eliminate these, but also would it eliminate most of the TPs, because passing the list reference to a receiver outside the scope does not mean that the receiver also will modify the list. For example, `return` a list reference makes up many of the use cases. Solving 0.2% of FPs by sacrificing most TPs does not make sense. Analyzing whether a receiver in fact _would_ modify the list would result in an even more complex analysis though.

richard-alun-williams:
I have to say, I think that the decision to not fix this fundamentally wrong.

There are any number of reasons why the return Collection from a method that uses a Stream to product a List might need to be modifiable, and the decision about when that is the case and when it isn't, should be entirely at the discretion of the engineering team. If a developer explicitly uses "Collectors.toList()" because they _want_ a modifiable, then the 'rule' should pretty much mind its own business.

A very simple use-case for this is: I get a List<someobject> and now I want to sort that list (maybe using a custom comparator in a Service class for a specific business need) ... but oh, no, if you used _toList()_ you can't even sort it.

I could list more, but the point here is, when the developers explicitly choose to get a modifiable List as the result using the Collectors.toList(), then SonarLint should NOT be forcing them to do otherwise and creating unnecessary code-quality flags.

KoenigsKind:
@kaufco, just out of curiosity:

How are these FP/TP rates calculated? For example we are not allowed to mark issues as "false positive", we can only use `// NOSONAR`. In this case, do we only contribute to the TP rates and never to FP?  
Or do you use some other way to determine the ratio?

PR #4704, opened 2024-03-05
SONARJAVA-4770 S2438 FN on arguments whose concrete type is Thread

kaufco:


java-checks/src/main/java/org/sonar/java/checks/ThreadAsRunnableArgumentCheck.java
```diff
@@ -19,85 +19,131 @@
...
+      case RETURN_STATEMENT -> visitReturnStatement((ReturnStatementTree) tree);
+      case YIELD_STATEMENT -> visitYieldStatement((YieldStatementTree) tree);
+      case NEW_ARRAY -> visitNewArray((NewArrayTree) tree);
+      case ASSIGNMENT -> {
+        var assignment = (AssignmentExpressionTree) tree;
+        checkTypeCoercion(assignment.variable().symbolType(), assignment.expression());
+      }
+      case METHOD_INVOCATION -> {
+        var invacation = (MethodInvocationTree) tree;
+        visitInvocation(invacation.methodSymbol(), invacation.arguments());
```
johann-beleites-sonarsource, line 69:
```suggestion
        var invocation = (MethodInvocationTree) tree;
        visitInvocation(invocation.methodSymbol(), invocation.arguments());
```

java-checks/src/main/java/org/sonar/java/checks/ThreadAsRunnableArgumentCheck.java
```diff
@@ -19,85 +19,131 @@
...
+      }
+      case NEW_CLASS -> {
+        var invacation = (NewClassTree) tree;
+        visitInvocation(invacation.methodSymbol(), invacation.arguments());
+      }
+      default -> throw new IllegalArgumentException();
```
johann-beleites-sonarsource, line 75: IMO, this is dead code that we won't be able to test and we shouldn't add this line. I know we have a rule (S131 I believe) that will complain, but I don't think it really makes sense in contexts like this and would consider it a FP. Not really convinced that it is useful enough as defensive programming here either. I'm going to leave the final decision up to you though.

johann-beleites-sonarsource (commented):
Looks nice, a few cosmetic comments

java-checks/src/main/java/org/sonar/java/checks/ThreadAsRunnableArgumentCheck.java
kaufco: Removed it. Now violating Joshua Bloch 6:36. May the lord forgive me.

java-checks/src/main/java/org/sonar/java/checks/ThreadAsRunnableArgumentCheck.java
```diff
@@ -19,85 +19,131 @@
...
     }
-    return parametersTypes.get(index);
   }
 
-  private static String getMessage(ExpressionTree argument, Type providedType, int index) {
-    String array = providedType.isArray() ? "[]" : "";
-    return MessageFormat.format("Replace \"{0}\" of type Thread{1} with an instance of Runnable{1}.", getArgName(argument, index), array);
+  private void visitYieldStatement(YieldStatementTree tree) {
+    Tree enclosing = ExpressionUtils.getEnclosingTree(tree, Tree.Kind.SWITCH_EXPRESSION, Tree.Kind.SWITCH_STATEMENT);
+    if (enclosing == null || enclosing.is(Tree.Kind.SWITCH_STATEMENT)) {
```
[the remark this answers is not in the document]
kaufco: Added a comment about it.

johann-beleites-sonarsource (approved):
LGTM!

PR #4706, opened 2024-03-06
SONARJAVA-4837 Implement rule S6878 Use record pattern instead of explicit field access

leonardo-pilastri-sonarsource:


johann-beleites-sonarsource (changes requested):
A couple of comments, mostly about alignment between RSPEC and implementation.

java-checks/src/test/java/org/sonar/java/checks/RecordPatternInsteadOfFieldAccessCheckTest.java
```diff
@@ -0,0 +1,37 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.java.checks;
+
+import org.junit.jupiter.api.Test;
+import org.sonar.java.checks.verifier.CheckVerifier;
+import org.sonar.java.checks.verifier.TestUtils;
+
+class RecordPatternInsteadOfFieldAccessCheckTest {
```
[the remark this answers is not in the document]
leonardo-pilastri-sonarsource: We cannot compile the test source with java < 20 because of that record pattern

java-checks/src/main/java/org/sonar/java/checks/RecordPatternInsteadOfFieldAccessCheck.java
```diff
@@ -0,0 +1,123 @@
...
+      checkCaseLabel((CaseLabelTree) tree);
+    }
+  }
+
+  private void checkCaseLabel(CaseLabelTree caseLabel) {
+    var typePattern = getTypePatternFromCaseGroup(caseLabel);
+    typePattern.ifPresent(typePatternTree -> checkTypePatternVariableUsage(typePatternTree.patternVariable()));
+  }
+
+  private static Optional<TypePatternTree> getTypePatternFromCaseGroup(CaseLabelTree caseLabel) {
```
johann-beleites-sonarsource, line 72: Minor, I realized I was confused for a moment about what this method does (the fact that it also filters if it is a record pattern). WDYT of renaming to something like `getRecordTypePatternFromCaseGroup`?

java-checks/src/main/java/org/sonar/java/checks/RecordPatternInsteadOfFieldAccessCheck.java
```diff
@@ -0,0 +1,123 @@
...
+    var comps = recordComponents(type);
+    for (Tree usage : patternVariable.symbol().usages()) {
+      if (usage.parent() instanceof MemberSelectExpressionTree mse && isNotRecordGetter(mse)) {
+        secondaryLocationsTrees.add(mse);
+      } else {
+        return;
+      }
+    }
+    // only if all the records components are used we report an issue
+    if (secondaryLocationsTrees.size() == comps.size()) {
```
johann-beleites-sonarsource, line 93: Unfortunately this doesn't work quite as one might think. E.g. try the sample:
```
    if (obj instanceof Point p) {
      return p.x() + p.x();
    }
```
It will raise an issue because the size is 2 for both. However, only `x()` is called (2x), never `y()`. Our AST trees don't work well with equals & hashCode, they usually simply inherit `Object`'s implementation. Instead, you probably need to put the identifier names in a set for proper hashing & equality.

johann-beleites-sonarsource (changes requested):
There is one check that doesn't work quite as expected, apart from that it looks good!

java-checks/src/main/java/org/sonar/java/checks/RecordPatternInsteadOfFieldAccessCheck.java
leonardo-pilastri-sonarsource: Argh, that's what I wanted to avoid.. good catch tho!

johann-beleites-sonarsource (approved):
Nice job, LGTM!

PR #4707, opened 2024-03-06
SONARJAVA-4857: FP in S3457 on certain java.util.logging strings with single quotes

ADarko22:
…ngle quotes

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/PrintfMisuseCheck.java
+++ b/java-checks-test-sources/default/src/main/java/checks/PrintfMisuseCheck.java
@@ -255,6 +255,7 @@ public class PrintfMisuseCheck {
     logger4.log(level, "message " + param1, new Exception()); // Noncompliant {{Lambda should be used to defer string concatenation.}}
 
     logger4.log(level, "Can't load library \"{0}\"!", "foo"); // Noncompliant {{Single quote "'" must be escaped.}}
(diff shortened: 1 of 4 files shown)


java-checks-test-sources/default/src/main/java/checks/PrintfMisuseCheck.java
```diff
@@ -254,6 +254,10 @@ void foo(Calendar c) throws IOException {
     logger4.log(level, "message " + "...");
     logger4.log(level, "message " + param1, new Exception()); // Noncompliant {{Lambda should be used to defer string concatenation.}}
 
+    logger4.log(level, "Can't load library \"{0}\"!", "foo"); // Noncompliant {{Single quote "'" must be escaped.}}
```
kaufco, line 257: The message says "Single quote "'" must be escaped.". How is it supposed to be escaped? `Can\'t` or `Can''t`? 
- `Can\'t`: it still raises an issue. If this is the supposed escape, then the rule must include an exception for it.
- `Can''t`: I'm not sure it produces the expected output "Can't", we should make this sure by checking the docs or test it ourselves if not documented explicitly.

kaufco (changes requested):
Looks good, except for one question about how to escape singular quotes. See the comment inline.

kaufco (approved):
Lgtm!

ADarko22:
Thank you @kaufco, I've updated the rspec to mention the double single quote for escaping a single quote: https://github.com/SonarSource/rspec/pull/3753. It should be easier for the users now :)

PR #4708, opened 2024-03-07
SONARJAVA-4835: FP on S3242 forcing user to add unnecessary logic

ADarko22:


java-checks/src/main/java/org/sonar/java/checks/LeastSpecificTypeCheck.java
```diff
@@ -219,13 +226,11 @@ private Type leastSpecificType() {
...
     }
 
-    private static boolean isOverriding(Symbol s1, Symbol s2) {
-      return s1.isMethodSymbol()
-        && s2.isMethodSymbol()
-        && s1.name().equals(s2.name())
-        && ConfusingOverloadCheck.isPotentialOverride((Symbol.MethodSymbol) s1, (Symbol.MethodSymbol) s2);
+    private static boolean isOverridingWithSameReturnType(Symbol.MethodSymbol m1, Symbol.MethodSymbol m2) {
+      return m1.name().equals(m2.name())
+        && m1.returnType() == m2.returnType()
```
ADarko22, line 231: The actual fix is to check that the return type is also the same in order to ensure that the least specific type is provides the exactly same methods being used

java-checks/src/main/java/org/sonar/java/checks/LeastSpecificTypeCheck.java
```diff
@@ -84,17 +83,20 @@ private static boolean isStringType(Type type) {
     return type.isUnknown() || type.is("java.lang.String");
   }
 
-  private void handleParameter(Symbol parameter, boolean springInjectionAnnotated) {
-    Type parameterType = parameter.type();
+  private void handleParameter(Symbol parameterSymbol, boolean springInjectionAnnotated) {
+    Type parameterType = parameterSymbol.type();
```
kaufco, line 87: Same here: the method got some refactoring but is otherwise untouched. This distracts from the _actual_ change this PR is about.

PR #4711, opened 2024-03-12
Remove obsolete check from generation script that would stop checks in the default check package to be imported in the check list

johann-beleites-sonarsource:


irina-batinic-sonarsource (approved):
LGTM!

PR #4712, opened 2024-03-12
SONARJAVA-4832 Implement S6881, Fix SONARJAVA-4913

johann-beleites-sonarsource:
VirtualThreads should be used for tasks that include heavy blocking operations.


johann-beleites-sonarsource:
Please rebase, don't squash.

irina-batinic-sonarsource (commented):
The implementation of the rule looks good, as does the fix for S1181. Good job!
I left a comment about the rule description on [this PR](https://github.com/SonarSource/rspec/pull/3780).

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S6881.html
```diff
@@ -0,0 +1,69 @@
...
+
+static void processRequest(String data) {
+    try {
+        // do something with data
+        Thread.sleep(1000);
+    } catch (InterruptedException e) {
+        e.printStackTrace();
+    }
+}
+</pre>
```
irina-batinic-sonarsource, line 36: I wondered why we have this Noncompliant code example without a corresponding Compliant Solution.

java-checks/src/main/java/org/sonar/java/checks/BlockingOperationsInVirtualThreadsCheck.java
```diff
@@ -0,0 +1,131 @@
...
+    .ofSubTypes(JAVA_LANG_THREAD)
+    .names("run")
+    .addWithoutParametersMatcher()
+    .build();
+
+  @Override
+  public void visitNode(Tree tree) {
+    switch (tree.kind()) {
+      case NEW_CLASS -> onConstructorFound((NewClassTree) tree);
+      case METHOD -> onMethodFound((MethodTree) tree);
```
leonardo-pilastri-sonarsource, line 95:
```suggestion
      case METHOD -> onMethodFound((MethodTree) tree);
      case METHOD_INVOCATION -> onMethodInvocationFound((MethodInvocationTree) tree);
```

java-checks/src/main/java/org/sonar/java/checks/BlockingOperationsInVirtualThreadsCheck.java
```diff
@@ -0,0 +1,131 @@
...
+      analyzeForIssues(newClassTree.arguments().get(0), newClassTree.identifier());
+    }
+  }
+
+  private void onMethodFound(MethodTree tree) {
+    if (overwritableThreadMethods.matches(tree)) {
+      analyzeForIssues(tree, tree.simpleName());
+    }
+  }
+
```
leonardo-pilastri-sonarsource, line 110: And with this fairly simple implementation we should be able to cover the extra constructor!
```

  private void onMethodInvocationFound(MethodInvocationTree tree) {
    var builderType = ((MemberSelectExpressionTree) tree.methodSelect()).expression().symbolType();
    if (threadCreationConstructors.matches(tree) && builderType.is("java.lang.Thread$Builder$OfPlatform")) {
      analyzeForIssues(tree.arguments().get(0), tree.methodSelect());
    }
  }
```

leonardo-pilastri-sonarsource (changes requested):
Looks good! I think we can easily cover an extra way to construct a platform thread, then I think the PR is good to go!

leonardo-pilastri-sonarsource (approved):
Nice work!

PR #4713, opened 2024-03-13
SONARJAVA-4907 Fix FP on S1481 with type patterns in switch case clauses

Wohops:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -61,6 +66,7 @@ public List<Tree.Kind> nodesToVisit() {
   public void visitNode(Tree tree) {
     if (tree.is(Tree.Kind.COMPILATION_UNIT)) {
       UNRESOLVED_IDENTIFIERS_VISITOR.check(tree);
+      SWITCH_PATTERN_VARIABLE_VISITOR.check(tree);
```
alban-auzeill, line 69: Running 2 visitors on the full AST is time-consuming. Because `UnresolvedIdentifiersVisitor` and `SwitchPatternVariableVisitor` have the same life cycle, it should be possible to merger the two behaviors in a `VariableVisitor` that extends  `UnresolvedIdentifiersVisitor`.

Wohops: You are right. it's smarter. Fixed.

alban-auzeill (approved):

PR #4715, opened 2024-03-14
SONARJAVA-4843 S6906: Virtual threads should not run tasks that include synchronized or native code

kaufco:


java-checks/src/main/java/org/sonar/java/checks/VirtualThreadNotSynchronizedCheck.java
```diff
@@ -0,0 +1,229 @@
...
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.SynchronizedStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = "S6906")
+public class VirtualThreadNotSynchronizedCheck extends IssuableSubscriptionVisitor implements JavaVersionAwareVisitor {
```
leonardo-pilastri-sonarsource, line 46: So you could change this
```suggestion
public class VirtualThreadNotSynchronizedCheck extends AbstractMethodDetection implements JavaVersionAwareVisitor {
```

java-checks/src/main/java/org/sonar/java/checks/VirtualThreadNotSynchronizedCheck.java
```diff
@@ -0,0 +1,229 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    var runnablesCollector = new RunnablesToCheckCollector();
+    tree.accept(runnablesCollector);
+
+    var synchonizedTreesCollector = new TreesWithSynchronizedCodeCollector();
+    runnablesCollector.invocations.stream()
+      .filter(it -> synchonizedTreesCollector.isInvokingSynchronizedCode(it.arguments().get(0)))
+      .forEach(it -> reportIssue(it, synchonizedTreesCollector.getSecondaryLocation(it.arguments().get(0))));
+  }
```
leonardo-pilastri-sonarsource, line 67: And after moving all the logic from the RunnablesToCheckCollector inside the main check class, change these two with something like
```suggestion
    @Override
  protected MethodMatchers getMethodInvocationMatchers() {
    return VIRTUAL_THREAD_BUILDER_METHODS;
  }

  @Override
  protected void onMethodInvocationFound(MethodInvocationTree mit) {
    var syncCollector = new TreesWithSynchronizedCodeCollector();
    if(isRunnableInVirtualThread(mit) && syncCollector.isInvokingSynchronizedCode(mit.arguments().get(0))) {
      reportIssue(mit, syncCollector.getSecondaryLocation(mit.arguments().get(0)));
    }
  }
```

java-checks/src/main/java/org/sonar/java/checks/VirtualThreadNotSynchronizedCheck.java
```diff
@@ -0,0 +1,229 @@
...
+      return expression instanceof IdentifierTree identifier && identifier.symbol().declaration() instanceof VariableTree variableTree ?
+        Optional.ofNullable(variableTree.initializer()) : Optional.empty();
+    }
+
+    private static boolean isCallToExecutorServiceBuilderWithVirtualTasks(ExpressionTree expression) {
+      return expression instanceof MethodInvocationTree mit && EXECUTOR_BUILDER_SERVICE_WITH_VIRTUAL_TASKS_METHOD.matches(mit);
+    }
+  }
+
+  private static class TreesWithSynchronizedCodeCollector extends BaseTreeVisitor {
```
leonardo-pilastri-sonarsource, line 153: If we don't scan the whole CompilationUnit anymore, I think there might be a way for this class to return directly an `Optional<Tree>` with the secondary location whenever synchronized code is spotted, and it could probably be more stateless if it could do so. 

So that in the end you could change the `onMethodInvocationFound` that I suggested earlier with something like:
```
  private static final TreesWithSynchronizedCodeCollector syncCollector = new TreesWithSynchronizedCodeCollector();
  @Override
    if(isRunnableInVirtualThread(mit)) {
      syncCollector.isInvokingSynchronizedCode(mit.arguments().get(0)).ifPresent(secondaryLocation -> reportIssue(mit, secondaryLocation));
    }
  }
```

leonardo-pilastri-sonarsource (changes requested):
Overall very nice work, you covered a lot of cases! I think that we can avoid visiting the compilation unit though, I understand that you probably implemented this way in order to cache methods already marked as synchronized within the same file, but I see more value in having a stateless implementation on more specific method invocation visits, if possible.
Let me know if you want to talk about it, as I might be missing something here

java-checks/src/main/java/org/sonar/java/checks/VirtualThreadNotSynchronizedCheck.java
```diff
@@ -0,0 +1,229 @@
...
+  void reportIssue(MethodInvocationTree tree, Tree secondaryLocation) {
+    reportIssue(
+      tree.methodSelect(),
+      "Use a platform thread instead of a virtual thread",
+      List.of(new JavaFileScannerContext.Location("synchronized", secondaryLocation)),
+      null
+    );
+  }
+
+  private static class RunnablesToCheckCollector extends BaseTreeVisitor {
```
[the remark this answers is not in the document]
kaufco: Yes: the main check is empty, so it can be stateless. That's a pattern I use when the actual implementation needs a state (like the HashSet and HashMap in my case), since we cannot rely on any cleanup-function called on the check after a file is visited.

`AbstractMethodDetection` I didn't use only because of this. I don't want a state in the actual Check class, but `AbstractMethodDetection` is a `IssuableSubscriptionVisitor`. I looked whether we have a `BaseTreeVisitor` implementation with similar functionality, but found nothing. 

I'm happy to discuss that though.

leonardo-pilastri-sonarsource (approved):
After talking, I still feel like an implementation without caching could be simpler and still effective, but it's not a strong opinion and I leave it up to you, I think your implementation is good.
Nice job!

PR #4716, opened 2024-03-14
SONARJAVA-4835: Fix FNs on S3242 when using generic types

ADarko22:


kaufco (approved):
LGTM! Maybe add a more complex test case with generics (which I expected to break the implementation, but it didn't)?

```
  public static class Phi {
    public List<? extends Number> getOtherThanOwnerType() {
      return null;
    }
  }

  public static class Theta extends Phi {
    @Override
    public List<Integer> getOtherThanOwnerType() {
      return null;
    }
  }

  public List<Integer> getNonRelatedA(Theta theta) { // Compliant
    return theta.getOtherThanOwnerType();
  }
  
  public List<? extends Number> getNonRelatedB(Theta theta) { // FN
    return theta.getOtherThanOwnerType();
  }
```

ADarko22:
I'm adding those test cases. Note that 
```
  public List<? extends Number> getNonRelatedB(Theta theta) { // FN
    return theta.getOtherThanOwnerType();
  }
```
is not a FN. It is reported as TP :)

kaufco (changes requested):
Sadly, it turned out that my example above indeed creates an FP, which was just masked by a bug that appears when the same method name is used.

As discussed: I'd suggest to close this PR and not to report on types with type parameters.

java-checks-test-sources/default/src/main/java/checks/LeastSpecificTypeCheck.java
```diff
@@ -263,73 +263,4 @@ public BigDecimal getUnaryOperator(UnaryOperator<BigDecimal> func) { // Complian
   public BigDecimal getFunction(Function<BigDecimal, BigDecimal> func) { // S4276 issue to promote UnaryOperator<BigDecimal>
     return func.apply(BigDecimal.ONE);
   }
-
```
ADarko22, line 266: extracted and re-organized the test samples into a separate sample file

kaufco (approved):
LGTM! I didn't find a counterexample for the Generics. However, this seems like a lot of test cases. I'll approve anyway, but before merging, maybe you should think about if you can reduce them and which of them really increases the coverage.

PR #4717, opened 2024-03-14
SONARJAVA-4825 S6876: SequencedCollection reversed view should be used for reverse iteration order

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ReversedMethodSequencedCollectionCheck.java
```diff
@@ -0,0 +1,105 @@
...
+      return;
+    }
+    boolean conditionOk = isConditionOk(condition);
+
+    if (initializerOk && conditionOk) {
+      reportIssue(forStatement, ISSUE_MESSAGE);
+    }
+  }
+
+  private static boolean isInitializerOk(Tree initializer) {
```
kaufco, line 90: I think we should choose a more speaking name here, because it's unclear what "OK" means in this context. What about `isInitializerListIteratorFromLast` or sth. like that?

java-checks/src/main/java/org/sonar/java/checks/ReversedMethodSequencedCollectionCheck.java
```diff
@@ -0,0 +1,105 @@
...
+      var variableInitializer = ((VariableTree) initializer).initializer();
+      if (variableInitializer != null && variableInitializer.is(Tree.Kind.METHOD_INVOCATION) && LIST_ITERATOR_MATCHER.matches((MethodInvocationTree) variableInitializer)) {
+        var arg = ((MethodInvocationTree) variableInitializer).arguments().get(0);
+        return arg.is(Tree.Kind.METHOD_INVOCATION) && LIST_SIZE_MATCHER.matches((MethodInvocationTree) arg);
+      }
+    }
+    return false;
+  }
+
+  private static boolean isConditionOk(Tree condition) {
```
kaufco, line 101: Also naming: what about `isConditionHasPrevious`?

kaufco (changes requested):
The implementation looks good overall! I just suggested simplifications due to Java 17 as well as renaming two methods.

One thing to note: strictly speaking, the replacement with the for-each statement is semantically equivalent only if in the body, the `Iterator.previous` function is called _exactly_  once. So these both cases are FPs:

```
for (var it = list.listIterator(list.size()); it.hasPrevious(); ) { // FP
}

for (var it = list.listIterator(list.size()); it.hasPrevious(); ) { // FP
  it.previous();
  it.previous();
}
```

I think that this is OK and they are intended FPs (you can construct arbitrarily complex cases: the call to `previous` could reside in an `if` etc., so no chance without CF analysis), but we should add them to the test cases at least.

kaufco (approved):
LGTM!

PR #4719, opened 2024-03-15
SONARJAVA-4904 FP on S1301 when using switch statement with type patterns

leonardo-pilastri-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/SwitchAtLeastThreeCasesCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/SwitchAtLeastThreeCasesCheck.java
@@ -20,34 +20,34 @@
 package org.sonar.java.checks;
 
 import org.sonar.check.Rule;
-import org.sonar.plugins.java.api.JavaFileScanner;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
-import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.CaseGroupTree;
 import org.sonar.plugins.java.api.tree.SwitchStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+import java.util.List;
 
 @Rule(key = "S1301")
-public class SwitchAtLeastThreeCasesCheck extends BaseTreeVisitor implements JavaFileScanner {
+public class SwitchAtLeastThreeCasesCheck extends IssuableSubscriptionVisitor {
 
   private JavaFileScannerContext context;
 
   @Override
-  public void scanFile(JavaFileScannerContext context) {
-    this.context = context;
-    scan(context.getTree());
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.SWITCH_STATEMENT);
   }
 
   @Override
-  public void visitSwitchStatement(SwitchStatementTree tree) {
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/SwitchAtLeastThreeCasesCheck.java
```diff
@@ -20,34 +20,40 @@
...
   @Override
-  public void visitSwitchStatement(SwitchStatementTree tree) {
+  public void visitNode(Tree tree) {
+    SwitchStatementTree switchStatementTree = (SwitchStatementTree) tree;
     int count = 0;
-    for (CaseGroupTree caseGroup : tree.cases()) {
+    for (CaseGroupTree caseGroup : switchStatementTree.cases()) {
+      if (caseGroup.labels().stream()
+        .flatMap(label -> label.expressions().stream())
+        .anyMatch(expression -> expression.is(Tree.Kind.TYPE_PATTERN, Tree.Kind.RECORD_PATTERN, Tree.Kind.GUARDED_PATTERN))) {
```
Wohops, line 48: I think that to make this method super-simple to read, I would extract this part into a method:

```java
if (containsPatterns(caseGroup)) {
  // usage of pattern improve readability and therefore invalidate use of if/else instead of a switch
  return;
}

// ...

private static boolean containsPatterns(CaseGroupTree caseGroup) {
  return caseGroup.labels().stream()
    .flatMap(label -> label.expressions().stream())
    .anyMatch(expression -> expression.is(Tree.Kind.TYPE_PATTERN, Tree.Kind.RECORD_PATTERN, Tree.Kind.GUARDED_PATTERN));
}
```

java-checks/src/main/java/org/sonar/java/checks/SwitchAtLeastThreeCasesCheck.java
```diff
@@ -20,34 +20,40 @@
...
+        // if any case is a type pattern matching, we don't raise an issue
```
Wohops, line 49: I feel this comment is not useful, as it literally only says that we interrupt the loop, and so paraphrase the code. Try to explain WHY we do such, to document the behavior of the rule itself and the business logic, not the algorithmic.

java-checks/src/main/java/org/sonar/java/checks/SwitchAtLeastThreeCasesCheck.java
```diff
@@ -20,34 +20,40 @@
...
-import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.CaseGroupTree;
 import org.sonar.plugins.java.api.tree.SwitchStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+import java.util.List;
 
 @Rule(key = "S1301")
-public class SwitchAtLeastThreeCasesCheck extends BaseTreeVisitor implements JavaFileScanner {
+public class SwitchAtLeastThreeCasesCheck extends IssuableSubscriptionVisitor {
```
Wohops, line 32: I think this move is a good one, as it will make the rule more performant. However, I would perfer having it in a 2nd distinct commit.
1. fix the FP
2. migrate from BaseTreeVisitor to IssuableSubscriptionVisitor

Ultimately, I would then keep the two commits as distinct when merging!

Wohops (changes requested):
Nice work. However, I feel we can do even better :)

Wohops (approved):

PR #4720, opened 2024-03-15
SONARJAVA-4912 S6204 Change rule definition

kaufco:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CollectorsToListCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CollectorsToListCheck.java
@@ -38,7 +38,7 @@ import org.sonar.plugins.java.api.tree.VariableTree;
 
 @Rule(key = "S6204")
 public class CollectorsToListCheck extends AbstractMethodDetection implements JavaVersionAwareVisitor {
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/CollectorsToListCheck.java
```diff
@@ -38,7 +38,7 @@
 
 @Rule(key = "S6204")
 public class CollectorsToListCheck extends AbstractMethodDetection implements JavaVersionAwareVisitor {
-  private static final String MESSAGE = "Replace this usage of 'Stream.collect(Collectors.%s())' with 'Stream.toList()'";
```
Wohops, line 41: I realize it's not your doing, but this issue message does not follow the convention: https://docs.sonarsource.com/sonarqube/latest/extension-guide/adding-coding-rules/#messages

Issue message should finish by a dot. Can you fix this?

Wohops (commented):
The fix LGTM, however, the commit message is not adequate. We don't change the rule definition at the end, but only the issue message. This change should reflect in the commit message.

Wohops (approved):
LGTM

PR #4728, opened 2024-03-20
SONARJAVA-4865 Fix issue when path is named regex

irina-batinic-sonarsource:


leonardo-pilastri-sonarsource (approved):
Looks good to me!

PR #4730, opened 2024-03-21
SONARJAVA-4826: fix FP and semicolon in quickfix

ValentinAebi-sonar:


leonardo-pilastri-sonarsource (approved):
Nice fix, looks good to me!

PR #4746, opened 2024-03-28
SONARJAVA-4875 S4502: Improve support for various CSRF disabling calls

johann-beleites-sonarsource:
le, please make your commits and pull request start with the ticket number (SONARJAVA-XXXX)


irina-batinic-sonarsource (commented):
Overall, it looks good. Is it possible to increase the coverage?

johann-beleites-sonarsource:
> Overall, it looks good. Is it possible to increase the coverage?

I think it is difficult, as there are currently no relevant methods where the number of type arguments != 1. However, a method could be added at some point, in which case I don't want this rule raising FPs, or worse, potentially crashing with an IndexOutOfBoundsException. Happy to discuss if you have an idea, though!

irina-batinic-sonarsource (approved):

PR #4749, opened 2024-03-29
SONARJAVA-4883 S4423: Support detection of TLS Protocol Downgrades for Spring programmatically

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
```diff
@@ -86,7 +103,18 @@ public void visitNode(Tree tree) {
           .skip(1)
           .map(secondary -> new JavaFileScannerContext.Location("Other weak protocol.", secondary))
           .toList();
-        reportIssue(unsecureVersions.get(0), "Change this code to use a stronger protocol.", secondaries, null);
+        reportIssue(unsecureVersions.get(0), ISSUE_MESSAGE, secondaries, null);
+      }
+    } else if (OPTIONS_ENABLED_PROTOCOLS.matches(mit)) {
+      ExpressionTree argument = arguments.get(0);
+      if (argument instanceof MethodInvocationTree methodInvocation && SET_OF.matches(methodInvocation)) {
```
johann-beleites-sonarsource, line 110: I would argue we can perhaps skip the second match using `SET_OF` and just look in the AST node for string literals without this check. There can be many different ways of creating a new set (including custom methods) and I don't think we risk a lot of FPs by checking for the weak TLS strings in the children of the first argument, no matter how exactly the first argument creates the set. WDYT?

java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
```diff
@@ -86,7 +103,18 @@ public void visitNode(Tree tree) {
...
+        boolean anyMatch = methodInvocation.arguments().stream()
+          .filter(arg -> arg.is(Tree.Kind.STRING_LITERAL))
+          .map(LiteralTree.class::cast)
+          .anyMatch(arg -> LiteralUtils.containsValue(arg, WEAK_FOR_SET_ENABLED_PROTOCOLS));
+        if (anyMatch) {
+          reportIssue(methodInvocation.arguments(), ISSUE_MESSAGE);
+        }
```
johann-beleites-sonarsource, line 117: I think it would be even nicer to highlight the protocols that are actually weak. Otherwise, if I have e.g. `Set.of("TLSv1.1", "TLSv1.2")`, it will highlight the entire set, even though it contains some secure protocols. An option could be to highlight the `setEnabledProtocols` method name and add all insecure protocols as secondary locations.

java-checks/src/main/java/org/sonar/java/checks/WeakSSLContextCheck.java
```diff
@@ -86,7 +103,18 @@ public void visitNode(Tree tree) {
...
           .map(secondary -> new JavaFileScannerContext.Location("Other weak protocol.", secondary))
           .toList();
-        reportIssue(unsecureVersions.get(0), "Change this code to use a stronger protocol.", secondaries, null);
+        reportIssue(unsecureVersions.get(0), ISSUE_MESSAGE, secondaries, null);
+      }
+    } else if (OPTIONS_ENABLED_PROTOCOLS.matches(mit)) {
+      ExpressionTree argument = arguments.get(0);
+      if (argument instanceof MethodInvocationTree methodInvocation && SET_OF.matches(methodInvocation)) {
+        boolean anyMatch = methodInvocation.arguments().stream()
+          .filter(arg -> arg.is(Tree.Kind.STRING_LITERAL))
```
johann-beleites-sonarsource, line 112: We can make the method a little more intelligent by using e.g. `ExpressionUtils#resolveAsConstant`, which would allow us to also identify cases where the protocol isn't passed as a string but e.g. where it is a statically defined string elsewhere. I saw you added a method to `LiteralUtils`, perhaps one can combine something here.

johann-beleites-sonarsource (changes requested):

java-checks-test-sources/spring-3.2/src/main/java/checks/WeakSSLContextCheckSample.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class WeakSSLContextCheckSample {
+
+  public static final String TLSV_1_0 = "TLSv1.0";
+
+  @Autowired
+  public WeakSSLContextCheckSample(SslProperties props, DefaultSslBundleRegistry registry, Set<String> propsSet) {
+    props.getBundle().getJks().get("").getOptions().setEnabledProtocols(Set.of("TLSv1.1", "test", "TLSv1.0")); // Noncompliant [[sc=53;ec=72;secondary=18,18]] {{Change this code to use a stronger protocol.}}
```
johann-beleites-sonarsource, line 18: A workaround for avoiding the problem of not really being able to test secondaries very well is to split the line into multiple lines, one per secondary. This way you can be relatively precise, even though the way we test for secondaries in sonar-java isn't great.

PR #4750, opened 2024-04-01
SONARJAVA-4866 S2077: Support detection of formatted SQL queries in Spring

irina-batinic-sonarsource:


java-checks-test-sources/spring-3.2/src/main/java/checks/SQLInjectionCheckSample.java
```diff
@@ -0,0 +1,60 @@
...
+  @Autowired
+  public SQLInjectionCheckSample(JdbcTemplate jdbcTemplate,
+    JdbcOperations jdbcOperations,
+    NamedParameterJdbcTemplate namedParameterJdbcTemplate,
+    JdbcClient jdbcClient) {
+    this.jdbcTemplate = jdbcTemplate;
+    this.jdbcOperations = jdbcOperations;
+    this.namedParameterJdbcTemplate = namedParameterJdbcTemplate;
+    this.jdbcClient = jdbcClient;
+  }
```
kaufco, line 33: I have no strong opinion about it, but: to me, this just seems to make the test more complicated than necessary (thus, distracting). You want to reproduce the pattern for Spring components here (including dependency injection), which is generally a good idea, but on the other hand, it's not required for the test. The sample methods that need any of these instances could just have them as parameters instead.

java-checks-test-sources/spring-3.2/src/main/java/checks/SQLInjectionCheckSample.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.jdbc.core.simple.JdbcClient;
+import org.springframework.security.core.userdetails.jdbc.JdbcDaoImpl;
+import org.springframework.security.provisioning.JdbcUserDetailsManager;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
```
kaufco, line 16: Same as about the dependency injection: this annotation makes sense in the context of Spring, but does not fulfill a purpose within this test.

java-checks-test-sources/spring-3.2/src/main/java/checks/SQLInjectionCheckSample.java
```diff
@@ -0,0 +1,60 @@
...
+
+    JdbcUserDetailsManager jdbcUserDetailsManager = new JdbcUserDetailsManager();
+    jdbcUserDetailsManager.setFindGroupIdSql("SELECT " + input); // Noncompliant
+
+    jdbcClient.sql("SELECT " + input).query(); // Noncompliant
+
+    return "Done";
+  }
+
+}
```
kaufco, line 60: Even though this will not increase the test coverage in the code, I think we should test all the functions you added to the method matchers. This is because they are explicitly mentioned in the specification, and one of them could easily be forgotten in the listing for the matcher builder, or not match the parameters etc. You can keep it minimalistic though, one line per function, just to make sure it triggers `// Noncompliant`.

kaufco (changes requested):
Looks good to me, just some comments about the test (most importantly, to add all the functions that are explicitly mentioned in the specification to make sure none of them was forgotten in the matchers).

java-checks-test-sources/spring-3.2/src/main/java/checks/SQLInjectionCheckSample.java
```diff
@@ -1,60 +1,69 @@
...
+    jdbcUserDetailsManager.setFindAllGroupsSql("SELECT " + input); // Noncompliant
     jdbcUserDetailsManager.setFindGroupIdSql("SELECT " + input); // Noncompliant
+    jdbcUserDetailsManager.setFindUsersInGroupSql("SELECT " + input); // Noncompliant
+    jdbcUserDetailsManager.setGroupAuthoritiesSql("SELECT " + input); // Noncompliant
+    jdbcUserDetailsManager.setInsertGroupAuthoritySql("SELECT " + input); // Noncompliant
+    jdbcUserDetailsManager.setInsertGroupMemberSql("SELECT " + input); // Noncompliant
+    jdbcUserDetailsManager.setInsertGroupSql("SELECT " + input); // Noncompliant
+    jdbcUserDetailsManager.setRenameGroupSql("SELECT " + input); // Noncompliant
+    jdbcUserDetailsManager.setUpdateUserSql("SELECT " + input); // Noncompliant
+    jdbcUserDetailsManager.setUserExistsSql("SELECT " + input); // Noncompliant
```
kaufco, line 64: Two suggestions for improving the test:

1. `"SELECT " + input` appears ca. 30 times here. You could assign a variable `var sqlStatement = "SELECT " + input`, or even have `String sqlStatement` as a method parameter again, since the concrete contents is irrelevant.
2. Since the method is now quite long and has many parameters, we could split it by owners, which would a) reduce the length of the method, b) reduce the parameter list, c) give the method a more speaking name. Such as in: `public void jdbcClientMethods(String sqlStatements, JdbcClient jdbcClient) { jdbcClient.... }`

kaufco (approved):
Lgtm! Two more suggestions for the test, but I'll leave it up to you (it would also be OK as-is).

java-checks-test-sources/spring-3.2/src/main/java/checks/SQLInjectionCheckSample.java
irina-batinic-sonarsource: 1. Not really. That is what the rule is really about.
2. Will do! Nice suggestion.

PR #4761, opened 2024-04-08
SONARJAVA-4699 FP on S3516 when calling a method using objects from "unknown" packages

ValentinAebi-sonar:


kaufco (changes requested):
Good work, very nice implementation! Let's refine the new test samples a bit though:
- should be minimalistic in terms of: what does really contribute to the test? (there is lots of unneeded code)
- There should be one positive example + `// Compliant` / `// Noncompliant` comments
- _Maybe_ one test case missing

I'm happy to discuss that with you!

kaufco (approved):
Lgtm!

PR #4762, opened 2024-04-09
Temporarily downgrade SQ version number to 10.3

kaufco:


pom.xml
```diff
@@ -86,7 +86,7 @@
 
     <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
 
-    <sonar.version>10.4.1.88267</sonar.version>
+    <sonar.version>10.3.0.82913</sonar.version>
     <sonar.plugin.api.version>10.7.0.2191</sonar.plugin.api.version>
```
alban-auzeill, line 90: We should not change `sonar.version` but `sonar.plugin.api.version` and use `10.2.0.1908`.
```suggestion
    <sonar.version>10.4.1.88267</sonar.version>
    <!-- 
      sonar-plugin-api should be compatible with SQ peach. You can see which version SQ is using
      at SonarSource/sonarqube/gradle.properties
     -->
    <sonar.plugin.api.version>10.2.0.1908</sonar.plugin.api.version>
```

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #4767, opened 2024-04-12
SONARJAVA-4935 S1192 should not report on individual lines of multi line string literal

kaufco:


java-checks/src/main/java/org/sonar/java/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -90,13 +92,58 @@ private static List<JavaFileScannerContext.Location> secondaryLocations(Collecti
...
+    } else {
+      var left = binary.leftOperand();
+      binary = asBinaryExpression(left, kind);
+      return binary != null? binary.rightOperand(): binary;
+    }
+  }
+
+  @Nullable
+  private static BinaryExpressionTree asBinaryExpression(Tree tree, Tree.Kind kind) {
+    return tree.is(kind)? (BinaryExpressionTree) tree: null;
```
leonardo-pilastri-sonarsource, line 144: Since we are only dealing with strings here, we will never need to use other kinds than PLUS, so we can remove the method's argument and adapt the other method calls in the check
```suggestion
  private static BinaryExpressionTree asBinaryExpression(Tree tree) {
    return tree.is(Tree.Kind.PLUS) ? (BinaryExpressionTree) tree: null;
```

leonardo-pilastri-sonarsource (changes requested):
Looks good to me! Just a simplification on a method.
Also I didn't want to mark multiple lines for formatting issues, can you reformat the code with the next commit?

leonardo-pilastri-sonarsource (approved):

PR #4769, opened 2024-04-16
SONARJAVA-4520 Rule S3655: False Positive with JUnit assertions

ValentinAebi-sonar:
This PR adds support for some methods of the Junit framework (`assertTrue`, `assertFalse`, and `fail`, not all overloads) to the SE engine.

### DIFF
--- a/java-checks-test-sources/default/src/main/java/symbolicexecution/checks/OptionalGetBeforeIsPresentCheck_jdk11.java
+++ b/java-checks-test-sources/default/src/main/java/symbolicexecution/checks/OptionalGetBeforeIsPresentCheck_jdk11.java
@@ -1,6 +1,8 @@
 package symbolicexecution.checks;
 
 import java.util.Optional;
+import org.junit.Assert;
+import org.junit.jupiter.api.Assertions;
 
 abstract class OptionalGetBeforeIsPresentCheck_jdk11 {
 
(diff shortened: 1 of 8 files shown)


java-checks-test-sources/default/src/main/java/symbolicexecution/checks/OptionalGetBeforeIsPresentCheck_jdk11.java
```diff
@@ -49,4 +51,105 @@ private void usingIsEmpty5() {
     }
     op.get(); // Noncompliant {{Call "op.isPresent()" or "!op.isEmpty()" before accessing the value.}}
   }
+
```
leonardo-pilastri-sonarsource, line 54: There are some missing samples for the signatures:
`org.junit.jupiter.api.Assertions#assertTrue(Z)V`,
`org.junit.jupiter.api.Assertions#assertFalse(ZLjava/lang/String;)V`
`org.junit.Assert#assertTrue(Z)V`
`org.junit.Assert#assertFalse(Ljava/lang/String;Z)V`

```
  void test_junit5_0(Optional<String> op){
    Assertions.assertTrue(op.isPresent());
    op.get();  // Compliant
  }
  
  void test_junit5_0_1(Optional<String> op){
    Assertions.assertFalse(op.isPresent(), "Hello");
    op.get();  // Noncompliant
  }
  
  void test_junit4_1_2(Optional<String> op){
    Assert.assertTrue(op.isPresent());
    op.get();  // Compliant
  }
  
  void test_junit4_4_2(Optional<String> op){
    Assert.assertFalse("", op.isEmpty());
    op.get(); // Compliant
  }
```

leonardo-pilastri-sonarsource (changes requested):
Looks good to me! You just missed some samples for a bunch of signatures, sadly there's no coverage report for the method behaviors we actually use 🤷 so it can happen

leonardo-pilastri-sonarsource (approved):
Nice job!

PR #4770, opened 2024-04-16
SONARJAVA-4936 Add mechanism to enable checks on main and test code at the same time

johann-beleites-sonarsource:


irina-batinic-sonarsource (approved):
LGTM!

PR #4776, opened 2024-04-17
SONARJAVA-4943 FP on S1144 if private method is referenced by name in annotations

kaufco:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -206,4 +220,41 @@ private static boolean isConstructorWithParameters(MethodTree methodTree) {
...
     return methodTree.is(Tree.Kind.METHOD) && !SerializableContract.SERIALIZABLE_CONTRACT_METHODS.contains(symbol.name());
   }
+
+  private static class CheckAnnotationsForMethodNames extends BaseTreeVisitor {
+
+    public CheckAnnotationsForMethodNames(Set<String> methodNames) {
+      this.methodNames = methodNames;
+    }
+
+    private Set<String> methodNames;
```
leonardo-pilastri-sonarsource, line 230: Using a `Set` here is not sufficient: if in the same file you have two different classes like so
```
  @MethodProvided(value = "foo")
  static class TN {
    private void foo() {} // Compliant, method is referenced in annotation
  }
  
  static class FN {
    private void foo() {} // False negative, this method has same name but no references
  }
```

the `CheckAnnotationsForMethodNames` will remove the name `foo` once and for all and the second encounter will not trigger an issue.

Maybe you could use the whole method signature instead of the simple name? Or a different data structure?

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -206,4 +220,41 @@ private static boolean isConstructorWithParameters(MethodTree methodTree) {
...
+      var isMethodAnnotation = isNameIndicatingMethod(annotationTree.annotationType().symbolType().name());
+      for (var arg : annotationTree.arguments()) {
+        if (arg.is(Tree.Kind.STRING_LITERAL)) {
+          if (isMethodAnnotation) {
+            removeMethodName((LiteralTree) arg);
+          }
+        } else if (arg instanceof AssignmentExpressionTree asgn && asgn.expression().is(Tree.Kind.STRING_LITERAL) && (
+          isMethodAnnotation || isNameIndicatingMethod(((IdentifierTree) asgn.variable()).name())
+        )) {
+          removeMethodName((LiteralTree) asgn.expression());
```
leonardo-pilastri-sonarsource, line 255: Not a super strong opinion here, but maybe it could be more comprehensible like
```suggestion
 else if (isAnnotationArgumentIndicatingMethod(arg, isMethodAnnotation)) {
          removeMethodName((LiteralTree) ((AssignmentExpressionTree)arg).expression());
        }
```

with the logic separated in another boolean method
```
    private static boolean isAnnotationArgumentIndicatingMethod(ExpressionTree arg, boolean isMethodAnnotation) {
      return arg instanceof AssignmentExpressionTree asgn && asgn.expression().is(Tree.Kind.STRING_LITERAL) && (
        isMethodAnnotation || isNameIndicatingMethod(((IdentifierTree) asgn.variable()).name())
      );
    }
```

leonardo-pilastri-sonarsource (changes requested):
Nice job! There is just one kinda of a corner case where we would encounter false negatives, other than that the logic looks nice to me!

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -55,33 +60,33 @@
...
-    unusedPrivateMethods.clear();
-    unresolvedMethodNames.clear();
+  public void visitNode(Tree tree) {
+    reportUnusedPrivateMethods(findUnusedPrivateMethods((ClassTree) tree));
   }
 
-  private void reportUnusedPrivateMethods() {
-    unusedPrivateMethods.stream()
-      .filter(methodTree -> !unresolvedMethodNames.contains(methodTree.simpleName().name()))
+  Stream<MethodTree> findUnusedPrivateMethods(ClassTree tree) {
```
leonardo-pilastri-sonarsource, line 73: Can we make this private static?

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -96,114 +101,150 @@ private void reportUnusedPrivateMethods() {
...
 
-  private static boolean isNotMethodFromSerializable(MethodTree methodTree, Symbol symbol) {
-    return methodTree.is(Tree.Kind.METHOD) && !SerializableContract.SERIALIZABLE_CONTRACT_METHODS.contains(symbol.name());
+  private static class MethodsUsedInAnnotationsFilter extends BaseTreeVisitor {
+
+    public MethodsUsedInAnnotationsFilter(Set<String> methodNames) {
+      this.filteredNames = methodNames;
+    }
+
+    private Set<String> filteredNames;
```
leonardo-pilastri-sonarsource, line 220: This could be final

leonardo-pilastri-sonarsource (approved):
Look good to me! Nice job
Just a couple of comments, but it's up to you

PR #4777, opened 2024-04-18
SONARJAVA-4944 S2699 should not raise on SpringBoot empty sanity test

leonardo-pilastri-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/helpers/UnitTestUtils.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/helpers/UnitTestUtils.java
@@ -161,25 +161,16 @@ public final class UnitTestUtils {
   }
 
   public static boolean isUnitTest(MethodTree methodTree) {
-    boolean isSpringSanityTest = isSpringBootSanityTest(methodTree);
-    if (isOrOverridesJunit4TestMethod(methodTree) && !isSpringSanityTest) {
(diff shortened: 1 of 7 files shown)


java-checks/src/main/java/org/sonar/java/checks/helpers/UnitTestUtils.java
```diff
@@ -161,18 +161,27 @@ public static boolean isUnitTestRelatedToObjectMethods(@Nullable MethodTree meth
   }
 
   public static boolean isUnitTest(MethodTree methodTree) {
-    if (isOrOverridesJunit4TestMethod(methodTree)) {
+    boolean isSpringSanityTest = isSpringBootSanityTest(methodTree);
+    if (isOrOverridesJunit4TestMethod(methodTree) && !isSpringSanityTest) {
```
kaufco, line 165: This also disables all other rules that are applied to unit tests (e.g. S5961) for the `contextLoads` method. I think we might not want this and keep it specific for S2699 for now. Maybe it's better to move the `isSpringBootSanityTest` method into the implementation of S2699 and remove it from the test utils?

kaufco (approved):
We should consider skipping `contextLoads` only for the S2699 check, not for other unit test checks (see comment in-line)

kaufco (changes requested):
Approved accidentally.

kaufco (approved):
Lgtm!

PR #4784, opened 2024-04-26
APPSEC-1695 Rule S6437: Remove overlapping method signatures

hendrik-buchwald-sonarsource:


leonardo-pilastri-sonarsource (approved):
LGTM

PR #4801, opened 2024-06-04
SONARJAVA-5021 Update Rules Metadata

alban-auzeill:
There is one change about one link to one rule `{rule:java:S5810}` and about new links to STIG standard


ADarko22 (approved):
LGTM!

PR #4808, opened 2024-06-05
[NO JIRA] Fix quality issues introduced in recent SE engine decoupling effort

johann-beleites-sonarsource:


leonardo-pilastri-sonarsource (approved):
Looks good!

PR #4818, opened 2024-06-18
SONARJAVA-5049 Add CheckListGenerator

irina-batinic-sonarsource:


sonar-java-plugin/pom.xml
```diff
@@ -164,7 +229,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>20000000</maxsize>
+                  <maxsize>30000000</maxsize>
```
johann-beleites-sonarsource, line 232: :eyes:

irina-batinic-sonarsource: Reverted. 
It was a change made during testing, and I forgot to revert it.

check-list/src/main/java/org/sonar/java/CheckListGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+    } catch (ClassNotFoundException e) {
+      throw new IllegalStateException("Cannot find the class for name " + className, e);
+    }
+  }
+
+  public String getRuleKey(Class<?> check) {
+    return check.getAnnotation(Rule.class).key();
+  }
+
+  public Metadata getMetadata(Reader reader) {
```
ValentinAebi-sonar, line 113: `Metadata` is protected, but here it is returned by a public method (cf warning in Intellij)

ValentinAebi-sonar (commented):
Is it really necessary to change the name of the class to `GeneratedCheckList`? Would it not be better to merely generate a class named `CheckList`?

check-list/src/main/java/org/sonar/java/CheckListGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+  public List<? extends Class<?>> getCheckClasses(Path directory, Path awsDirectory) {
+    var modifiedFiles = getModifiedFiles(directory, awsDirectory);
+    return modifiedFiles.stream()
+      .map(CheckListGenerator::getClassByName)
+      .filter(c -> !Modifier.isAbstract(c.getModifiers()))
+      .filter(c -> c.getAnnotationsByType(Rule.class).length != 0)
+      .toList();
+  }
+
+  private static List<String> getModifiedFiles(Path relativePath, Path awsRelativePath) {
```
leonardo-pilastri-sonarsource, line 89: Can't this be called `getCheckFiles`?

check-list/src/main/java/org/sonar/java/CheckListGenerator.java
```diff
@@ -0,0 +1,271 @@
...
+    try (Stream<Path> stream = Stream.concat(Files.walk(relativePath.resolve("org/sonar/java/checks")), Files.walk(awsRelativePath.resolve("org/sonar/java/checks")))) {
+      return stream
```
leonardo-pilastri-sonarsource, line 91: I think we should sort the classes that we are collecting here

johann-beleites-sonarsource (commented):
Cool stuff, just a couple of minor comments. My main concern is if we now need a multi-step process to build

.cirrus.yml
```diff
@@ -180,7 +180,8 @@ sanity_task:
   sanity_script:
     - source cirrus-env QA
     - source set_maven_build_version $BUILD_NUMBER
-    - mvn clean compile --projects java-checks-test-sources --also-make-dependents
+    - mvn clean install -Dmaven.test.skip=true
+    - mvn compile --projects java-checks-test-sources --also-make-dependents
```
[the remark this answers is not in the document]
irina-batinic-sonarsource: The command that was used before, `mvn clean compile --projects java-checks-test-sources --also-make-dependents`, if executed before compiling the whole project, fails because of compilation errors (GeneratedCheckList class doesn't exist).
Now, I updated the script a bit to fix it without building the whole project (see the latest version of this file).
Also, yes, the documentation needed a minor update, and it has been added.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S1135.json
leonardo-pilastri-sonarsource, line 1: I think this and sonarpedia.json update do not belong to this PR

leonardo-pilastri-sonarsource (changes requested):
Everything looks good to me, let's just remove sonarpedia.json and S1135.json from the PR and then we can merge

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S1135.json
irina-batinic-sonarsource: If we completely remove it, some checks will fail.
S1135 is enabled on tests, but it had a `Main` scope defined.
I will create a separate PR to fix S1135 and remove it from this one.

leonardo-pilastri-sonarsource (approved):

PR #4835, opened 2024-07-02
Add releasability status check

irina-batinic-sonarsource:


ADarko22 (approved):
Thank you for taking care ot if! 

LGTM and it will be nice to add to the scanners and other projects as well

PR #4836, opened 2024-07-02
Add releasability status check is triggered manually

irina-batinic-sonarsource:


ADarko22 (approved):
Very well to catch to move it as manual action! :)

PR #4840, opened 2024-07-08
[DONOTMERGEME] Adding some extra debug lines to scan without parsing

dorian-burihabwa-sonarsource:


dorian-burihabwa-sonarsource:
A more formal ticket will be created to improve the debugging logs during incremental analysis.

PR #4849, opened 2024-08-12
SONARJAVA-5098 Fix regexp that finds '\n'

Ivan-Igorevich:
* 'String#contains()` method finds '\\n' as a false-positive new line symbol. In several cases symbols, including 'n', may follow an escaped backslash to generate '\n' as a printed out string, e.g. LaTeX code generator.

- [ ] Unit tests are passing
- [ ] Provided a unit test for your fix

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/PrintfMisuseCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/PrintfMisuseCheck.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import org.sonar.check.Rule;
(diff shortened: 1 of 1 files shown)


Ivan-Igorevich:
cloned repository failed to test even without any changes. now figuring out what am i doing wrong
image

kaufco (commented):
Hi Ivan-Igorevich,

Thanks for your contribution! The negative look-behind would be an elegant solution since it involves no additional code, just trying to solve the problem within the regex.

Sadly, to look for one additional preceding `\` is not enough, since you can have strings like `"aaa\\\naaa"` (invalid) or `"aaa\\\\naaa"` (valid). So we need to count all the `\` and decide whether the number is even or odd. Look behind does not support repetition, so I see no alternative to writing code. Otherwise we could try something like `(?<!\\\\(\\\\\\\\)*)\\\\n` with the regex.

We have created a PR [here ](https://github.com/SonarSource/sonar-java/pull/4862) that counts the preceding `\`.

PR #4857, opened 2024-09-10
SONARJAVA-4950 Fix S6204 IndexOutOfBoundsException when lombok.val is used

leonardo-pilastri-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CollectorsToListCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CollectorsToListCheck.java
@@ -19,6 +19,7 @@
  */
 package org.sonar.java.checks;
 
+import java.util.List;
 import javax.annotation.CheckForNull;
 import org.sonar.check.Rule;
 import org.sonar.java.checks.methods.AbstractMethodDetection;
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/CollectorsToListCheck.java
```diff
@@ -107,6 +107,9 @@ private static boolean isInvariantTypeArgument(MethodInvocationTree collectMetho
     if (streamType.isRawType()) {
       return true;
     }
+    if(collectMethodInvocation.symbolType().typeArguments().isEmpty()){
+      return false;
+    }
     Type collectArgType = collectMethodInvocation.symbolType().typeArguments().get(0);
     Type streamArgType = streamType.typeArguments().get(0);
     return collectArgType.is(streamArgType.fullyQualifiedName());
```
kaufco, line 115: Also, in Java 21 we have `List.getFirst()` which we should use as a replacement for `List.get(0)`. Maybe we can shorten the above code even a bit more like this?

```suggestion
    List<Type> typeArguments = collectMethodInvocation.symbolType().typeArguments();
    return !typeArguments.isEmpty() && typeArguments.getFirst().is(streamType.typeArguments().getFirst().fullyQualifiedName());
```

kaufco (approved):
I have a suggestion for a minor improvement. Since the PR is still good, I approved and leave it up to you whether to take the change.

PR #4858, opened 2024-09-11
SONARJAVA-5089 S1312 should not report on interfaces

pauloreilly-sonarsource:
Should not report on interfaces since fields cannot be private and static final is redundant.


pauloreilly-sonarsource:
Currently, Ruling Issues Number Differences check is failing due to LOGGER issue on Interfaces no longer being raised.

kaufco (approved):
Lgtm, but I'd suggest to change the comment before merging.

kaufco (changes requested):
We should add a unit test for the new case (-> Logger declared inside an interface). Also, one suggestion about changing a comment. Otherwise, the PR looks good!

kaufco (approved):
Lgtm!

PR #4860, opened 2024-09-11
SONARJAVA-5080 S1659: Quickfix breaks syntax when multiple arrays are declared

leonardo-pilastri-sonarsource:

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/OneDeclarationPerLineCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/OneDeclarationPerLineCheck.java
@@ -26,6 +26,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.sonar.check.Rule;
 import org.sonar.java.checks.helpers.QuickFixHelper;
+import org.sonar.java.model.JavaTree;
 import org.sonar.java.model.LineUtils;
 import org.sonar.java.reporting.JavaQuickFix;
 import org.sonar.java.reporting.JavaTextEdit;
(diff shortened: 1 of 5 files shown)


java-checks/src/main/java/org/sonar/java/checks/OneDeclarationPerLineCheck.java
```diff
@@ -120,20 +120,39 @@ private JavaQuickFix getQuickFixes(List<VariableTree> nodesToReport) {
...
+  }
+
+  private static boolean isTypeFragmented(VariableTree variableTree) {
+    var typeTokens = ((JavaTree) variableTree.type()).allTokens();
+    for (var token : typeTokens) {
+      if (token.range().end().isAfter(variableTree.simpleName().identifierToken().range().start())) {
+        return true;
+      }
+    }
+    return false;
```
kaufco, line 155: We can skip the loop, because: if any token in the list is placed after the variable name, then this is true also for the last token. So let's check the last token only:

```suggestion
    var lastToken = typeTokens.get(typeTokens.size()-1);
    return (lastToken.range().end().isAfter(variableTree.simpleName().identifierToken().range().start()));
```

kaufco, line 155: And because of the previous, we don't need all tokens anymore. We only need the last one and can simplify further:

```suggestion
    var lastToken = variableTree.type().lastToken();
    return (lastToken != null && lastToken.range().end().isAfter(
      variableTree.simpleName().identifierToken().range().start())
    );
```

java-frontend/src/main/java/org/sonar/java/model/JavaTree.java
```diff
@@ -72,6 +72,17 @@ public abstract class JavaTree implements Tree {
...
+    List<SyntaxToken> list = new ArrayList<>();
+    if (this instanceof SyntaxToken st) {
+      list.add(st);
+    } else {
+      for (Tree tree : children) {
+        list.addAll(((JavaTree) tree).allTokens());
+      }
+    }
+    return list;
+  }
```
kaufco, line 85: We don't need that  (see comment above)

```suggestion
```

kaufco (changes requested):
I suggested a simplification, otherwise LGTM.

kaufco (approved):
Lgtm!

PR #4861, opened 2024-09-11
SONARJAVA-5115 FP in S5803: when using (VisibleForTesting.PROTECTED)

pauloreilly-sonarsource:
When using (otherwise = androidx.annotation.VisibleForTesting.PROTECTED).

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/VisibleForTestingProtectedUsageCheck/IssueMyObject.java
+++ b/java-checks-test-sources/default/src/main/java/checks/VisibleForTestingProtectedUsageCheck/IssueMyObject.java
@@ -1,0 +1,14 @@
+package checks.VisibleForTestingProtectedUsageCheck;
+
+public class IssueMyObject {
+
+  IssueMyObject() {
+  }
+
+  // androidx.annotation.VisibleForTesting
+  // @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
+  // where VisibleForTesting.PROTECTED = 4
+  @VisibleForTesting(otherwise = 3, othertestcase=1)
+  String foo;
+
+}
--- a/java-checks-test-sources/default/src/main/java/checks/VisibleForTestingProtectedUsageCheck/IssueMyObject.java
+++ b/java-checks-test-sources/default/src/main/java/checks/VisibleForTestingProtectedUsageCheck/IssueMyObject.java
@@ -5,10 +5,13 @@ public class IssueMyObject {
   IssueMyObject() {
   }
 
+  @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED, othertestcase=1, othertypecase="S")
+  String bar;
+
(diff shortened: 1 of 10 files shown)


java-checks-test-sources/default/src/main/java/checks/VisibleForTestingProtectedUsageCheck/IssueMyObject.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class IssueMyObject {
+
+  IssueMyObject() {
+  }
+
+  // androidx.annotation.VisibleForTesting
+  // @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
+  // where VisibleForTesting.PROTECTED = 4
+  @VisibleForTesting(otherwise = 3, othertestcase=1)
```
irina-batinic-sonarsource, line 11: Do we have a test:
`@VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)`

irina-batinic-sonarsource (commented):
Hey Paul,
Good job overall, I just left two comments.
Please let me know if you'd like to talk about it.

java-checks-test-sources/default/src/main/java/checks/VisibleForTestingProtectedUsageCheck/IssueMyObject.java
irina-batinic-sonarsource: I meant a test case like this in`IssueMyObject `:
`  @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED, othertestcase=1)
  String bar; `

Inside VisibleForTesting, we need to add an `int PROTECTED = 4; `

And in an `IssueService` something like this:
`  public String test2() {
    return new IssueMyObject().bar; // Noncompliant
  } `

java-checks/src/main/java/org/sonar/java/checks/VisibleForTestingUsageCheck.java
```diff
@@ -70,7 +71,21 @@ private static boolean isMisusedVisibleForTesting(Symbol symbol) {
   }
 
   private static boolean isVisibleForTestingAnnotation(AnnotationInstance annotationInstance) {
-    return "VisibleForTesting".equals(annotationInstance.symbol().name());
+    return "VisibleForTesting".equals(annotationInstance.symbol().name())
+      && !isOtherwiseProtected(annotationInstance);
+  }
+
+  // Condition added for FP Check in S5803 => issue should not be raised when
+  // (otherwise = androidx.annotation.VisibleForTesting.VisibleForTesting.PROTECTED) is specified
```
irina-batinic-sonarsource, line 79: The comment can be removed.
The code itself should be descriptive enough.

irina-batinic-sonarsource (commented):
Let's increase code coverage, and it looks good! 🚀

irina-batinic-sonarsource (approved):

PR #4862, opened 2024-09-12
SONARJAVA-5098 FP in S3457 when using strings involving \\n

kaufco:


java-checks/src/main/java/org/sonar/java/checks/PrintfMisuseCheck.java
```diff
@@ -391,9 +391,22 @@ private void checkBoolean(MethodInvocationTree mit, String param, Type argType)
...
+      index = formatString.indexOf("\\n", index+2);
+    }
+  }
+
+  private static boolean isOddNumberOfEscapeChars(String formatString, int lastIndex) {
+    var index = lastIndex-1;
+    while (index >= 0 && formatString.charAt(index) == '\\') {
+      index--;
     }
+    return ((lastIndex - index) & 1) != 0;
```
leonardo-pilastri-sonarsource, line 409:
```suggestion
    return (lastIndex - index) % 2 != 0;
```

leonardo-pilastri-sonarsource (approved):
It looks good to me, just one comment but it is up to you, since the performance difference should be negligible between the bitwise operation and the modulus, I would prefer to have something more immediately readable. But it is up to you!

PR #4864, opened 2024-09-12
SONARJAVA-4963 Line and column positions are wrong after text blocks using '\' line continuations

alban-auzeill:


java-frontend/src/main/java/org/sonar/java/model/JParser.java
mike-swainston-rainford-sonar, line 1: Since the root cause of the bug is the lineEndTable array in the CompilationUnit class I had a quick look at the methods that use it. There is a call in JParser (line 288) to the CompilationUnit's getColumnNumber() method which uses the array

mike-swainston-rainford-sonar (changes requested):
There is a call to CompilationUnit.getColumnNumber() in JParser. This should be changed to a call to the new LineConverter class.
I think the LineConverter class should be documented as a workaround for the Eclipse CompilationUnit.lineEndingTable bug when continuation characters are present.
Does the LineConterverTest need a test of text that has no line continuation characters?

PR #4865, opened 2024-09-13
SONARJAVA-5059 Instanceof check for MemberSelectExpressionTree

pauloreilly-sonarsource:


leonardo-pilastri-sonarsource (approved):
Looks good to me!

PR #4867, opened 2024-09-16
SONARJAVA-5099 Fix FP on S1144 if @MethodSource is used without arguments

irina-batinic-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -82,7 +82,11 @@ private static Stream<MethodTree> findUnusedPrivateMethods(ClassTree tree) {
     methodNames.removeAll(collector.unresolvedMethodNames);
     var filter = new MethodsUsedInAnnotationsFilter(methodNames);
     tree.accept(filter);
-    return unusedPrivateMethods.stream().filter(it -> filter.filteredNames.contains(it.simpleName().name()));
+    var methodTreeStream = unusedPrivateMethods.stream().filter(it -> filter.filteredNames.contains(it.simpleName().name()));
+    var annotatedMethodCollector = new AnnotatedMethodCollector();
+    tree.members().forEach(it -> it.accept(annotatedMethodCollector));
+    var methodSourceAnnotatedMethods = annotatedMethodCollector.methodSourceAnnotatedMethods;
+    return methodTreeStream.filter(it -> !methodSourceAnnotatedMethods.contains(it.simpleName().name()));
```
leonardo-pilastri-sonarsource, line 89: I think we might incur in new false negatives with this logic, if it's the unused private method itself to have the `@MethodSource` annotation (which you might argue does not make sense in most cases):
```
  @MethodSource
  private void unusedMethodWithWrongAnnotation(){} // FN
```

You could encounter unused private methods which mistakenly (or not) have that annotation, and they would not raise an issue

leonardo-pilastri-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
irina-batinic-sonarsource: Hmm. On line 138 there is a check:
`if (isUnusedPrivate(symbol) && hasNoAnnotation(methodTree) `
So, we should be safe.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -101,6 +105,23 @@ private void reportUnusedPrivateMethods(Stream<MethodTree> methods) {
...
+    }
+
+    @Override
+    public void visitMethod(MethodTree methodTree) {
+      if (methodTree.modifiers().annotations().stream().anyMatch(annotation -> annotation.annotationType().symbolType().is("org.junit.jupiter.params.provider.MethodSource"))) {
+        methodSourceAnnotatedMethods.add(methodTree.simpleName().name());
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
irina-batinic-sonarsource: It is sufficient!
I tried different approaches, and while refactoring, I stayed with visitor one.
Thanks for seeing this.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
leonardo-pilastri-sonarsource: I think that helps prevent false positives, but with the filtering we do before, around lines ~80, methods annotated with `@MethodSource` will never reach the reporting part.
So if you try adding the example I pasted above, you should notice that no issue is raised for that method, even if is in fact private and unused.

leonardo-pilastri-sonarsource (approved):
LGTM!

PR #4872, opened 2024-09-18
SONARJAVA-5111 S5838 Improve quickfix directly to isEmpty()

pauloreilly-sonarsource:


its/autoscan/src/test/resources/autoscan/diffs/diff_S5838.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S5838",
   "hasTruePositives": false,
-  "falseNegatives": 274,
+  "falseNegatives": 280,
```
kaufco, line 4: This looks suspicious. The number of issues should not change, but we only want a specialized report and quick fix for some of them.

java-checks/src/main/java/org/sonar/java/checks/tests/AssertJChainSimplificationIndex.java
```diff
@@ -147,6 +147,9 @@ private AssertJChainSimplificationIndex() {
       withSubjectArgumentCondition(LiteralUtils::isTrue, AssertJChainSimplificationIndex::isNotObject, "isTrue()"),
       withSubjectArgumentCondition(LiteralUtils::isFalse, AssertJChainSimplificationIndex::isNotObject, "isFalse()"),
       withSubjectArgumentCondition(LiteralUtils::isEmptyString, AssertJChainSimplificationIndex::isNotObject, "isEmpty()"),
+      withSubjectArgumentCondition(AssertJChainSimplificationIndex::isZeroIntOrLong, AssertJChainSimplificationIndex::isStringLength, msgWithActual(IS_EMPTY)),
+      withSubjectArgumentCondition(AssertJChainSimplificationIndex::isZeroIntOrLong, AssertJChainSimplificationIndex::isCollectionSize, msgWithActual(IS_EMPTY)),
+      withSubjectArgumentCondition(AssertJChainSimplificationIndex::isZeroIntOrLong, AssertJChainSimplificationIndex::isArrayLength, msgWithActual(IS_EMPTY)),
```
kaufco, line 152: Because of the additional FNs: probably here the match condition is not equivalent to `x && argument value == 0`?

kaufco (changes requested):
Looks good, and nice tests! I'm concerned about the autoscan diffs though: there should not be a difference, in SONARJAVA-5111 we just want to specialize the report message and the quickfixes.

pauloreilly-sonarsource:
Testing locally, in `target/actual/java-checks-test-sources-mvn/java-S5838.json` we see new 6 issues reported for the newly added Test class:

`"java-checks-test-sources:default/src/test/java/checks/tests/AssertJChainSimplificationCheckTest_ListQuickFix.java": [
18,
20,
27,
29,
36,
38
],`

The line nos above correlate with the 6 new `// Noncompliant` entries in the source test code.

This output of FN correlates with other pre-existing Test classes like `java-checks-test-sources:default/src/test/java/checks/tests/AssertJChainSimplificationCheckTest.java`.

So this would suggest the difference in `diff_S5838.json` rightly transpires because of the new issues added to the test file: hence the increase of 6 is expected.

PR #4874, opened 2024-09-19
SONARJAVA-5116 S1105 sometimes falsely requests the curly brace to be moved

pauloreilly-sonarsource:


java-frontend/src/main/java/org/sonar/java/model/JParser.java
```diff
@@ -827,7 +828,18 @@ private int findLeftBraceTokenIndex(AbstractTypeDeclaration e) {
...
+        e.bodyDeclarations().stream().sorted((Comparator<ASTNode>) (o1, o2) -> {
+          // sort so that items are ordered by their start position
+          if (o1.getStartPosition() < o2.getStartPosition()) {
+            // we want the ordered by ascending start positions
+            return -1;
+          } else {
+            // start positions should never be the same, so just return 1 otherwise
+            return 1;
+          }
+        }).toList().get(0), TerminalTokens.TokenNameLBRACE);
```
alban-auzeill, line 842: The `Comparator` expects to receive zero when `o1` is equal to `o2`. Here, we return `1`, which means `o1` is greater than `o2` even when they are equal. I don't know why it works 🤔 .

The correct sort could be:
```
e.bodyDeclarations().stream().sorted(Comparator.comparing(ASTNode::getStartPosition)).toList()
```

But, I think it's overkill to sort a list if we need only the min value:
```suggestion
      // for records, bodyDeclarations may not be in the order encountered in file, for classes they are
      List<BodyDeclaration> bodyDeclarations = e.bodyDeclarations();
      BodyDeclaration firstDeclaration = bodyDeclarations.get(0);
      for (int i = 1; i < bodyDeclarations.size(); i++) {
        BodyDeclaration declaration = bodyDeclarations.get(i);
        if (firstDeclaration.getStartPosition() > declaration.getStartPosition()) {
          firstDeclaration = declaration;
        }
      }
      return tokenManager.firstIndexBefore(firstDeclaration, TerminalTokens.TokenNameLBRACE);
 ```

File: java-checks/src/test/java/org/sonar/java/checks/LeftCurlyBraceEndLineCheckTest.java [hunk 2]
```diff
@@ -41,6 +41,14 @@ void java_17() {
       .verifyIssues();
   }
 
+  @Test
+  void test_record() {
+    CheckVerifier.newVerifier()
+      .onFile(TestUtils.mainCodeSourcesPath("checks/LeftCurlyBraceEndLineCheck_record.java"))
+      .withCheck(new LeftCurlyBraceEndLineCheck())
+      .verifyNoIssues();
```
alban-auzeill, line 49: It would be nice to also check noncompliant code with records in the same file.

```suggestion
     .verifyIssues();
```
And add this code at the end of `java-checks-test-sources/default/src/main/java/checks/LeftCurlyBraceEndLineCheck_record.java`:
```java
record TestRecordSyntaxV3(Integer value)
{ // Noncompliant
  public TestRecordSyntaxV3(String v)
  { // Noncompliant
    this(Integer.parseInt(v));
  }
  public static final String CONST_1 = "IDK";
  public static final String CONST_2 = "IDK";
}
```

alban-auzeill (changes requested):

java-frontend/src/main/java/org/sonar/java/model/JParser.java
alban-auzeill: After reading your comment `// start positions should never be the same` now I understand why you were never returning 0. I just noticed that I read the code and not the comments...

alban-auzeill (approved):

PR #4876, opened 2024-09-19
SONARJAVA-4012 S2885 should not raise issues when a private field is only used synchronized

alban-auzeill:


johann-beleites-sonarsource:
Closing as stale

PR #4877, opened 2024-09-20
SONARJAVA-3829 S2629 should not report when log level is enabled

pauloreilly-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -185,6 +187,7 @@ private static String[] testMethodNames(String[] lowerCaseNames) {
 
   private JavaFileScannerContext context;
   private Deque<Tree> treeStack = new ArrayDeque<>();
+  private boolean insideLevelTestReturnStatementFound = false;
```
ADarko22, line 190: What do you think of renaming this variable to `foundReturnStatementInsideLevelTest`?

java-checks-test-sources/default/src/main/java/checks/LazyArgEvaluationCheckSampleFPs.java
```diff
@@ -0,0 +1,61 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.Marker;
+import org.slf4j.MarkerFactory;
+
+// https://community.sonarsource.com/t/false-positive-on-java-s2629/42091
+class AFalsePositveFromTheCommunity {
+
+  void foo() {
+    final Marker myMarker = MarkerFactory.getMarker("MY_MARKER");
```
ADarko22, line 12: Why not put these variables as class fields and have a single class testing the different scenarios?
The reference to the community thread can be put as a comment on top of each group of examples

ADarko22 (approved):
Great job! Very nice solution for covering the cases with the return inside the nested levels :)

Left two small remarks to rename the new flag and clean up the sample file

PR #4880, opened 2024-09-23
SONARJAVA-3970 Fix FP on Try.run with onFailure

irina-batinic-sonarsource:


java-checks-test-sources/default/src/main/java/checks/ServletMethodsExceptionsThrownCheckSample.java
```diff
@@ -1,12 +1,52 @@
...
+  protected final void doGet(HttpServletRequest request, HttpServletResponse response) {
+    final Consumer<Throwable> logError = x -> Logger.error("hey"+x.getMessage(), x);
+    Try.run(() -> {
+      // let's try to raise an exception
+    }).onFailure(logError); // Compliant
+
+    Try.run(() -> { // Noncompliant
+      // let's try to raise an exception
+    }).isEmpty();
+  }
```
ADarko22, line 48: Shouldn't this be in the `S1989noncompliantvavr` class?

java-checks-test-sources/default/src/main/java/checks/ServletMethodsExceptionsThrownCheckSample.java
```diff
@@ -1,12 +1,52 @@
...
+class S1989noncompliantvavr extends HttpServlet {
+
+  @Override
+  protected final void doGet(HttpServletRequest request, HttpServletResponse response) {
+
+    Try.run(() -> { // Noncompliant
+      // let's try to raise an exception
+    });
+
+    Try.of(() -> { // Compliant
```
ADarko22, line 27: Shouldn't this be in the S1989compliantvavr class?

java-checks/src/main/java/org/sonar/java/checks/ServletMethodsExceptionsThrownCheck.java
```diff
@@ -95,6 +97,17 @@ private static List<Type> getCaughtExceptions(List<CatchTree> catches) {
   }
 
   private void checkMethodInvocation(MethodInvocationTree node) {
+    if (node.methodSelect() instanceof MemberSelectExpressionTree memberSelect
+      && memberSelect.expression() instanceof IdentifierTree identifier && "Try".equals(identifier.name())
+      && "run".equals(memberSelect.identifier().name())) {
```
ADarko22, line 102: It looks like the same behavior could also happen for `runRunnable. ' It's better to cover it as well. At that point, I think it's better to simplify this if statement, that has too many conditions, by extracting it in a nice method

ADarko22 (changes requested):
It looks good! 
Just a few comments about an extra case to cover and maybe arranging the samples more consistently

ADarko22 (approved):
Let's merge it!

PR #4881, opened 2024-09-23
SONARJAVA-4061 S2226 should ignore fields assigned in `init` method

pauloreilly-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/ServletInstanceFieldCheck.java
```diff
@@ -107,11 +109,19 @@ private void reportIssuesOnVariable() {
...
-          excludedVariables.add((VariableTree) declaration);
-        }
+        maybeAddExcluded(((IdentifierTree) tree.variable()).symbol().declaration());
+      }
+      // handles e.g. "this.first = 42;" assignments -> member "first" is prefixed with "this."
+      if (tree.variable().is(Kind.MEMBER_SELECT)) {
+        maybeAddExcluded(((MemberSelectExpressionTree) tree.variable()).identifier().symbol().declaration());
+      }
+    }
+    private void maybeAddExcluded(@Nullable Tree declaration) {
```
kaufco, line 121: Maybe something like `addExcludedIfVaraible` or `addVariableToExcluded` would be a more descriptive name here?

java-checks/src/main/java/org/sonar/java/checks/ServletInstanceFieldCheck.java
```diff
@@ -107,11 +109,19 @@ private void reportIssuesOnVariable() {
   private class AssignmentVisitor extends BaseTreeVisitor {
     @Override
     public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+      // handles e.g. "second = this.first * 2;" assignments -> no "this" prefix to member "second"
       if (tree.variable().is(Kind.IDENTIFIER)) {
```
kaufco, line 113: I have no strong opinion on that one, but since we're using `tree.variable()` four times in this method, maybe a local variable `var variable = tree.variable()` could be better? As said, no strong opinion here, because some people prefer _not_ to assign getter results to local variables.

kaufco (changes requested):
We should check in the implementation that the left side of the `MemberSelect` is indeed a `this`. Apart from this, just some minor remarks.

kaufco (approved):

kaufco (approved):

PR #4882, opened 2024-09-24
SONARJAVA-4287 S3012 has a false positive when using either auto boxing or auto unboxing

pauloreilly-sonarsource:


dorian-burihabwa-sonarsource (commented):
LGTM 👍🏿, I tested manually test with all primitive types and we should be good to go here.

PR #4894, opened 2024-09-30
SONARJAVA-5138 [Support Java 22] Update Eclipse Compiler for Java 3.39.0

alban-auzeill:


java-checks-test-sources/default/src/main/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -293,4 +302,76 @@ case MyRecord(int x, int y) -> { } // Compliant
...
+    var ball = queue.remove();
+    switch (ball) {
+      case RedBall _ -> System.out.println("Red"); // Compliant
+      case BlueBall _ -> System.out.println("Blue"); // Compliant
+      default -> throw new IllegalStateException("Unexpected value: " + ball);
+    }
+
+    switch (ballHolder) {
+      case BallHolder(RedBall _) -> System.out.println("One Red"); // Compliant
+// not supported: case BallHolder(BlueBall _), BallHolder(GreenBall _) -> System.out.println("Blue or Green Ball"); // Compliant
```
alban-auzeill, line 347: add FIXME

java-checks/src/main/java/org/sonar/java/checks/naming/KeywordAsIdentifierCheck.java
```diff
@@ -33,7 +33,7 @@ public class KeywordAsIdentifierCheck extends BaseTreeVisitor implements JavaFil
 
   private JavaFileScannerContext context;
 
-  private static final Set<String> FORBIDDEN_IDENTIFIERS = SetUtils.immutableSetOf("enum", "_","assert");
```
ADarko22, line 36: If we decide to change the implementation then the [RSPEC](https://github.com/SonarSource/rspec/blob/master/rules/S1190/java/rule.adoc) for [S1190](https://sonarsource.github.io/rspec/#/rspec/S1190) should be updated accordingly

java-frontend/src/main/java/org/sonar/java/model/JParser.java
```diff
@@ -1228,7 +1276,12 @@ private VariableTreeImpl convertVariable(SingleVariableDeclaration e) {
       );
     }
     t.variableBinding = e.resolveBinding();
-    declaration(t.variableBinding, t);
+    if (t.variableBinding != null) {
+      if (t.type() instanceof InferedTypeTree inferredType) {
+        inferredType.typeBinding = t.variableBinding.getType();
+      }
+      declaration(t.variableBinding, t);
```
ADarko22, line 1283: What do you think if we log a warming when there is no binding? Would it be too noisy?

sonar-java-plugin/src/test/java/org/sonar/plugins/java/SanityTest.java
```diff
@@ -88,6 +88,18 @@ class SanityTest {
     "../java-checks/src/test/files"
   };
 
+  private static final String[] COMPILING_FILE_DIRECTORIES = {
```
ADarko22, line 91: the `FILE_DIRECTORIES` above seems to be unused

ADarko22 (approved):
A few small comments and questions

PR #4896, opened 2024-10-02
SONARJAVA-4546 Support @org.jspecify.annotations.NullMarked in existing nullability checks

pauloreilly-sonarsource:


kaufco (approved):
Lgtm. Nice work!

PR #4898, opened 2024-10-04
SONARJAVA-4544 Support org.jspecify.annotations. @Nullable and @NonNull in existing nullability checks

irina-batinic-sonarsource:


pauloreilly-sonarsource (approved):
Looks good.

PR #4901, opened 2024-10-10
[NO JIRA] Fix Quality Flaws

alban-auzeill:


its/autoscan/src/test/resources/autoscan/diffs/diff_S1130.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S1130",
   "hasTruePositives": true,
-  "falseNegatives": 26,
+  "falseNegatives": 25,
```
alban-auzeill, line 4: This change is because AutoScanTest analyzes not only source examples in `java-checks-test-sources` but also a real project: `java-checks-test-sources/test-classpath-reader`, and this file has been changed: `java-checks-test-sources/test-classpath-reader/src/test/java/org/sonar/java/test/classpath/TestClasspathUtilsTest.java`

leonardo-pilastri-sonarsource (approved):
Looks good to me, thanks for all the fixes!
I suggest we do not squash these commits as it could be easier to understand with the complete commit list

PR #4902, opened 2024-10-11
[NO JIRA] Fix Autoscan result, it was wrong in the CI because of the missing tests compilation

alban-auzeill:


erwan-serandour (approved):
LGTM

leonardo-pilastri-sonarsource (approved):

PR #4903, opened 2024-10-14
 SONARJAVA-5144 Update Custom rules documentation to match MyJavaFileCheckRegistrarTest

erwan-serandour:


docs/CUSTOM_RULES_101.md
```diff
@@ -566,22 +566,35 @@ public class MyJavaFileCheckRegistrar implements CheckRegistrar {
 ```
 
 Now, because we added a new rule, we also need to update our tests to make sure it is taken into account. 
-To do so, navigate to its corresponding test class, named `MyJavaFileCheckRegistrarTest`, and update the expected number of rules from 8 to 9.
+To do so, navigate to its corresponding test class, named `MyJavaFileCheckRegistrarTest`, add your rule key to the end of the mainRuleKeys list
+and your rule class name at the end of the mainCheckClasses list.
```
dorian-burihabwa-sonarsource, line 570: We can probably put the whole sentence on single line here. This should help review changes to the text in the future
```suggestion
To do so, navigate to its corresponding test class, named `MyJavaFileCheckRegistrarTest`, add your rule key to the end of the mainRuleKeys list and your rule class name at the end of the mainCheckClasses list.
```

dorian-burihabwa-sonarsource (approved):
LGTM! Feel free to apply the suggestion on the separate lines that form a single sentence.

PR #4946, opened 2024-12-04
SONARJAVA-5185 exclude Optional.orElseThrow from S2201 as it has side effects

erwan-serandour:
[SONARJAVA-5185](https://sonarsource.atlassian.net/browse/SONARJAVA-5185)

Part of 

[SONARJAVA-5185]: https://sonarsource.atlassian.net/browse/SONARJAVA-5185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/S2201_IgnoredReturnValueCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/S2201_IgnoredReturnValueCheckSample.java
@@ -178,6 +178,6 @@ class S2201_IgnoredReturnValueCheckSample {
   }
 
   void validateValue(Object o){
-    Optional.ofNullable(o).orElseThrow(); //FP here orElseThrow() is used for the side effect
+    Optional.ofNullable(o).orElseThrow(); // Compliant: orElseThrow() is used for the side effect
   }
 }
--- a/java-checks-test-sources/default/src/main/java/checks/S2201_IgnoredReturnValueCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/S2201_IgnoredReturnValueCheckSample.java
@@ -176,4 +176,8 @@ class S2201_IgnoredReturnValueCheckSample {
   List<String> makeList() {
     return new ArrayList<>();
   }
+
+  void validateValue(Object o){
+    Optional.ofNullable(o).orElseThrow(); //FP here orElseThrow() is used for the side effect
+  }
 }
--- a/java-checks/src/main/java/org/sonar/java/checks/IgnoredReturnValueCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/IgnoredReturnValueCheck.java
@@ -40,6 +40,7 @@ public class IgnoredReturnValueCheck extends IssuableSubscriptionVisitor {
   private static final String JAVA_LANG_STRING = "java.lang.String";
   private static final String JAVA_UTIL_FUNCTION_SUPPLIER = "java.util.function.Supplier";
   private static final String JAVA_UTIL_STREAM_STREAM = "java.util.stream.Stream";
+  private static final String JAVA_UTIL_OPTIONAL = "java.util.Optional";
   private static final String COLLECT = "collect";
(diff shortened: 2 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/IgnoredReturnValueCheck.java
```diff
@@ -67,14 +68,15 @@ public class IgnoredReturnValueCheck extends IssuableSubscriptionVisitor {
...
-    "java.util.Optional",
+    JAVA_UTIL_OPTIONAL,
     "com.google.common.base.Optional");
 
   private static final List<String> EXCLUDED_PREFIX = Arrays.asList("parse", "format", "decode", "valueOf");
 
   private static final MethodMatchers EXCLUDED = MethodMatchers.or(
     MethodMatchers.create().ofTypes("java.lang.Character").names("toChars").addParametersMatcher("int", "char[]", "int").build(),
-    MethodMatchers.create().ofTypes(JAVA_LANG_STRING).names("intern").addWithoutParametersMatcher().build());
+    MethodMatchers.create().ofTypes(JAVA_LANG_STRING).names("intern").addWithoutParametersMatcher().build(),
```
dorian-burihabwa-sonarsource, line 78: What is this `intern` method and why do we need to support it?

java-checks-test-sources/default/src/main/java/checks/S2201_IgnoredReturnValueCheckSample.java
```diff
@@ -176,4 +176,8 @@ public Collection<String> get() {
   List<String> makeList() {
     return new ArrayList<>();
   }
+
+  void validateValue(Object o){
+    Optional.ofNullable(o).orElseThrow(); //FP here orElseThrow() is used for the side effect
```
dorian-burihabwa-sonarsource, line 181: Let's make the comment more helpful for the future.
```suggestion
    Optional.ofNullable(o).orElseThrow(); // Compliant: orElseThrow() is used for the side effect
```

dorian-burihabwa-sonarsource (changes requested):
Let's clarify why we need the `intern` method and change the comment in the code sample

java-checks/src/main/java/org/sonar/java/checks/IgnoredReturnValueCheck.java
erwan-serandour: intern, make a copy of the string and store it in the constant pool; this can be used to improve performance when you use it a lot of times the same string.

java-checks-test-sources/default/src/main/java/checks/S2201_IgnoredReturnValueCheckSample.java
erwan-serandour: :+1:

java-checks/src/main/java/org/sonar/java/checks/IgnoredReturnValueCheck.java
erwan-serandour: I don't think intern is used a lot, but when it is, it is for the side effect

dorian-burihabwa-sonarsource: Why do we need to add it to the list? It looks like it is [already ignored in the code sample](https://github.com/SonarSource/sonar-java/blob/3e2cb4478100a5ea082dde30004067579eef58a4/java-checks-test-sources/default/src/main/java/checks/S2201_IgnoredReturnValueCheckSample.java#L77). So this new line does not seem to have any effect on the existing code

erwan-serandour: I did add it, it was here before. The line has changed because I added a new case at the end of the or matcher (we don't see it with this view). The only logical change is at line 79.

dorian-burihabwa-sonarsource (approved):

PR #4953, opened 2024-12-06
SONARJAVA-4424 add methods annotated with @Test to the list of testNg test methods

erwan-serandour:
[SONARJAVA-4424](https://sonarsource.atlassian.net/browse/SONARJAVA-4424)

Part of 

[SONARJAVA-4424]: https://sonarsource.atlassian.net/browse/SONARJAVA-4424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/files/non-compiling/checks/NoTestInTestClassCheckSample.java
+++ b/java-checks-test-sources/default/src/main/files/non-compiling/checks/NoTestInTestClassCheckSample.java
@@ -121,6 +121,16 @@ public class TestNGClassTest { // Noncompliant
   private void test1() { }
   public static void foo() {}
 }
+
+@org.testng.annotations.Test
+public class TestNGClassTestUseAnnotation {
+
+  @org.testng.annotations.Test
+  void myMethod(){
+
+  }
+}
+
 @org.testng.annotations.Test(groups ="integration")
 public abstract class AbstractIntegrationTest2{
 }
--- a/java-checks-test-sources/default/src/main/files/non-compiling/checks/NoTestInTestClassCheckSample.java
+++ b/java-checks-test-sources/default/src/main/files/non-compiling/checks/NoTestInTestClassCheckSample.java
@@ -123,8 +123,8 @@ public class TestNGClassTest { // Noncompliant
 }
 
 @org.testng.annotations.Test
-public class TestNGClassTestUseAnnotation {
-
+class TestNGClassTestWithMethodAnnotated { // compliant, with testng when the class is annotated with @Test all public methods are considered as tests
+                                          // non public methods can also be added to tests with the @Test annotation
   @org.testng.annotations.Test
   void myMethod(){
 
--- a/java-checks-test-sources/default/src/main/files/non-compiling/checks/NoTestInTestClassCheckSample.java
+++ b/java-checks-test-sources/default/src/main/files/non-compiling/checks/NoTestInTestClassCheckSample.java
@@ -131,6 +131,23 @@ class TestNGClassTestWithMethodAnnotated { // compliant, with testng when the cl
   }
 }
 
+@org.testng.annotations.Test
+class TestNGClassWithUnkownAnnotation {
+  @Unkown
+  void myMethod(){
+
+  }
+}
+
+@org.testng.annotations.Test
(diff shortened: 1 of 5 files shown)


java-checks-test-sources/default/src/main/files/non-compiling/checks/NoTestInTestClassCheckSample.java
```diff
@@ -122,6 +122,16 @@ public class TestNGClassTest { // Noncompliant
   private void test1() { }
   public static void foo() {}
 }
+
+@org.testng.annotations.Test
+public class TestNGClassTestUseAnnotation {
```
dorian-burihabwa-sonarsource, line 127: We can change the name of the class to `TestNGDoesNotRequirePublicTestMethods` or add a `compliant` comment explaining why this case is not an issue

java-checks/src/main/java/org/sonar/java/checks/tests/NoTestInTestClassCheck.java
```diff
@@ -113,7 +113,11 @@ private static boolean isArchUnitTestClass(Symbol.TypeSymbol classSymbol) {
   }
 
   private void checkTestNGmembers(IdentifierTree className, Stream<Symbol> members) {
-    if (members.noneMatch(member -> member.isMethodSymbol() && member.isPublic() && !member.isStatic() && !"<init>".equals(member.name()))) {
+    if (members.noneMatch(member -> {
+      boolean annotatedWithTest = isTestFieldOrMethod(member);
+      boolean publicMethod = member.isMethodSymbol() && member.isPublic() && !member.isStatic() && !"<init>".equals(member.name());
+      return annotatedWithTest || publicMethod;
+    })) {
```
dorian-burihabwa-sonarsource, line 120: Can we extract the second test into its own method and simplify the code?
```suggestion
    if (members.noneMatch(member -> isTestFieldOrMethod(member) || isPublicInstanceMethod(member)) {
```

dorian-burihabwa-sonarsource (changes requested):
Nice work. There is a couple of things to revisit here:
1. Simplify the lambda to find test members in a TestNG class
2. Clarify which test case is raising a FP here (and if it cannot be fixed, clearly mark it as an FP when semantic is incomplete)

java-checks/src/main/java/org/sonar/java/checks/tests/NoTestInTestClassCheck.java
```diff
@@ -113,7 +115,17 @@ private static boolean isArchUnitTestClass(Symbol.TypeSymbol classSymbol) {
...
   private void checkTestNGmembers(IdentifierTree className, Stream<Symbol> members) {
-    if (members.noneMatch(member -> member.isMethodSymbol() && member.isPublic() && !member.isStatic() && !"<init>".equals(member.name()))) {
+    Predicate<SymbolMetadata.AnnotationInstance> isTestNgAnnotation = ann -> {
+      Type type = ann.symbol().type();
+      return type.isUnknown() || type.is(TEST_NG_TEST);
+    };
+    Predicate<Symbol> isTestMethod = member -> {
+      boolean annotatedWithTest = member.isMethodSymbol() && member.metadata().annotations().stream().anyMatch(isTestNgAnnotation);
+      boolean publicMethod = member.isMethodSymbol() && member.isPublic() && !member.isStatic() && !"<init>".equals(member.name());
+      return annotatedWithTest || publicMethod;
```
dorian-burihabwa-sonarsource, line 125: It looks like we are only testing on methods because we have `isMethodSymbol` in both branches. Can we return early if the member is not a method?
```suggestion
      if (!member.isMethodSymbol()) {
        return false;
      }
      boolean annotatedWithTest = member.metadata().annotations().stream().anyMatch(isTestNgAnnotation);
      boolean publicMethod = member.isPublic() && !member.isStatic() && !"<init>".equals(member.name());
      return annotatedWithTest || publicMethod;
```

dorian-burihabwa-sonarsource (changes requested):

dorian-burihabwa-sonarsource (approved):

PR #4954, opened 2024-12-06
SONARJAVA-4263 Fix FP on S2325 when method requires type parameter from parent class.

tomasz-tylenda-sonarsource:
[SONARJAVA-4263](https://sonarsource.atlassian.net/browse/SONARJAVA-4263)

Fix FP on S2325 when method requires type parameter from parent class.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/StaticMethodCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/StaticMethodCheck.java
@@ -16,8 +16,10 @@
  */
 package org.sonar.java.checks;
 
+import java.util.ArrayList;
 import java.util.Deque;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/StaticMethodCheck.java
```diff
@@ -95,6 +103,37 @@ public void visitMethod(MethodTree tree) {
...
+      .map(Type::symbol)
+      .anyMatch(parentTypeParam::contains);
+  }
+
+  private static void collectTypeVars(List<Type> accumulator, Type t) {
+    for(Type tt: t.typeArguments()) {
+      if(tt.isTypeVar()) {
+        accumulator.add(tt);
+      } else {
+        collectTypeVars(accumulator, tt);
```
dorian-burihabwa-sonarsource, line 132: Is there a way to document, using a comment or casting it into a new variable, what type `tt` could be apart from a `TypeVar`? No need to be exhaustive but an example would be nice

java-checks/src/main/java/org/sonar/java/checks/StaticMethodCheck.java
```diff
@@ -95,6 +103,37 @@ public void visitMethod(MethodTree tree) {
...
+    ClassTree parent = (ClassTree) symbol.declaration().parent();
+    Set<Symbol> parentTypeParam =
+      parent
+        .typeParameters()
+        .stream()
+        .map(TypeParameterTree::symbol)
+        .collect(Collectors.toUnmodifiableSet());
+    Symbol.TypeSymbol returnType = symbol.returnType();
+    List<Type> returnTypeVars = new ArrayList<>();
+    collectTypeVars(returnTypeVars, returnType.type());
```
dorian-burihabwa-sonarsource, line 120: Open question: Would it be more efficient to collect the type variables in the return type before moving back up the tree and recovering the parent type parameters? I am not sure I can see a point in the flow of the check that would prevent us from entering this block if the method has no type parameter. In such a case, the collection of type parameters from the parent class would be useless here
```suggestion
    List<Type> returnTypeVars = new ArrayList<>();
    collectTypeVars(returnTypeVars, returnType.type());
    if (returnTypeVars.isEmpty()) {
      return false;
    }
    ClassTree parent = (ClassTree) symbol.declaration().parent();
    Set<Symbol> parentTypeParam =
      parent
        .typeParameters()
        .stream()
        .map(TypeParameterTree::symbol)
        .collect(Collectors.toUnmodifiableSet());
    Symbol.TypeSymbol returnType = symbol.returnType();
```

dorian-burihabwa-sonarsource (changes requested):
Nice work, I think there is room for a little improvements but this change already solves the problem quite nicely!

PR #4957, opened 2024-12-09
SONARJAVA-4436 Fix FP on S2095 when using @lombok.Cleanup

tomasz-tylenda-sonarsource:
https://sonarsource.atlassian.net/browse/SONARJAVA-4436

### DIFF
--- a/java-symbolic-execution/java-symbolic-execution-plugin/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
+++ b/java-symbolic-execution/java-symbolic-execution-plugin/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
@@ -487,7 +487,7 @@ public class UnclosedResourcesCheck extends SECheck {
         .metadata()
         .annotations()
         .stream()
-        .anyMatch(annotation -> "lombok.Cleanup".equals(annotation.symbol().type().fullyQualifiedName()));
+        .anyMatch(annotation -> annotation.symbol().type().fullyQualifiedName().endsWith("Cleanup"));
     }
   }
 
--- a/java-symbolic-execution/java-symbolic-execution-plugin/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
+++ b/java-symbolic-execution/java-symbolic-execution-plugin/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
@@ -474,10 +474,20 @@ public class UnclosedResourcesCheck extends SECheck {
     @Override
     public void visitIdentifier(IdentifierTree tree) {
       // close resource as soon as it is encountered in the resource declaration
-      if (isWithinTryHeader(tree)) {
+      // or if it is annotated with @lombok.Cleanup
+      if (isWithinTryHeader(tree) || isAnnotatedLombokCleanup(tree)) {
         Symbol symbol = tree.symbol();
         closeResource(programState.getValue(symbol));
       }
+    }
+
+    private static boolean isAnnotatedLombokCleanup(IdentifierTree tree) {
+      return tree
+        .symbol()
+        .metadata()
+        .annotations()
+        .stream()
+        .anyMatch(annotation -> "lombok.Cleanup".equals(annotation.symbol().type().fullyQualifiedName()));
     }
   }
 
--- a/java-symbolic-execution/java-symbolic-execution-plugin/src/test/files/se/UnclosedResourcesLombokCheck.java
+++ b/java-symbolic-execution/java-symbolic-execution-plugin/src/test/files/se/UnclosedResourcesLombokCheck.java
@@ -4,12 +4,12 @@ import java.io.InputStream;
 import lombok.Cleanup;
 
 class UnclosedResourcesLombokCheck {
-  public void wrongHandling() throws IOException {
-    FileInputStream stream = new FileInputStream("myFile"); // Noncompliant
-    stream.read();
+  public void missingAnnotation(String fileName) throws IOException {
+    InputStream in = new FileInputStream(fileName); // Noncompliant
+    in.read();
   }
 
-  void withLombokCleanup(String fileName) throws IOException {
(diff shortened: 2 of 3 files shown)


java-symbolic-execution/java-symbolic-execution-plugin/src/test/java/org/sonar/java/se/checks/UnclosedResourcesCheckTest.java
```diff
@@ -33,6 +33,15 @@ void test() {
       .verifyIssues();
   }
 
+  @Test
+  void test_lombok() {
```
dorian-burihabwa-sonarsource, line 37: Nitpick: I know it does not occur in the unit tests we have for rules, but I think it would be nice to rename this test with a more meaningful name. WDYT of  `does_not_raise_on_lombok_cleanup_annottated_variable`? It is kind of a mouthful but the contract would be clearer this way.

java-symbolic-execution/java-symbolic-execution-plugin/src/test/files/se/UnclosedResourcesLombokCheck.java
```diff
@@ -0,0 +1,18 @@
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import lombok.Cleanup;
+
+class UnclosedResourcesLombokCheck {
+  public void missingAnnotation(String fileName) throws IOException {
+    InputStream in = new FileInputStream(fileName); // Noncompliant
+    in.read();
+  }
```
dorian-burihabwa-sonarsource, line 10: We can replace this part of the sample with a fully qualified lombok annotation so that this file only contains cases where  we should not raise
```suggestion
  public void fullyQualified(String fileName) throws IOException {
    @lombok.Cleanup
    InputStream in = new FileInputStream(fileName);
    in.read();
  }
```

java-symbolic-execution/java-symbolic-execution-plugin/src/test/java/org/sonar/java/se/checks/UnclosedResourcesCheckTest.java
dorian-burihabwa-sonarsource: Let's add another test without semantics

java-symbolic-execution/java-symbolic-execution-plugin/src/main/java/org/sonar/java/se/checks/UnclosedResourcesCheck.java
```diff
@@ -474,11 +474,21 @@ private void closeResource(@Nullable final SymbolicValue target) {
...
       }
     }
+
+    private static boolean isAnnotatedLombokCleanup(IdentifierTree tree) {
+      return tree
+        .symbol()
+        .metadata()
+        .annotations()
+        .stream()
+        .anyMatch(annotation -> "lombok.Cleanup".equals(annotation.symbol().type().fullyQualifiedName()));
```
dorian-burihabwa-sonarsource, line 490: Let's try to support the case where we do not have semantics but the annotation looks like `Cleanup` or `lombok.Cleanup`

java-symbolic-execution/java-symbolic-execution-plugin/src/test/java/org/sonar/java/se/checks/UnclosedResourcesCheckTest.java
tomasz-tylenda-sonarsource: Renamed with camelCase. Added a test `_noSemantic`.

java-symbolic-execution/java-symbolic-execution-plugin/src/test/java/org/sonar/java/se/checks/UnclosedResourcesCheckTest.java
```diff
@@ -34,12 +34,22 @@ void test() {
...
+      .verifyNoIssues();
+  }
+
+  @Test
+  void doesNotRaiseOnLombokCleanupAnnotatedVariable_noSemantic() {
+    SECheckVerifier.newVerifier()
+      .onFile("src/test/files/se/UnclosedResourcesLombokCheck.java")
+      .withCheck(new UnclosedResourcesCheck())
+      .withoutSemantic()
+      .withClassPath(SETestUtils.CLASS_PATH)
```
dorian-burihabwa-sonarsource, line 51: I think we can remove this line as the `withoutSemantic` indicates that we should not pass any library in the class path (the flag is eventually used [here](https://github.com/SonarSource/sonar-java/blob/889e74199f6ffd01bcae78a794352d6a3866a459/java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/internal/InternalCheckVerifier.java#L299)). 
```suggestion
```

java-symbolic-execution/java-symbolic-execution-plugin/src/test/java/org/sonar/java/se/checks/UnclosedResourcesCheckTest.java
```diff
@@ -34,12 +34,22 @@ void test() {
...
     SECheckVerifier.newVerifier()
       .onFile("src/test/files/se/UnclosedResourcesLombokCheck.java")
       .withCheck(new UnclosedResourcesCheck())
       .withClassPath(SETestUtils.CLASS_PATH)
-      .verifyIssues();
+      .verifyNoIssues();
+  }
+
+  @Test
+  void doesNotRaiseOnLombokCleanupAnnotatedVariable_noSemantic() {
```
dorian-burihabwa-sonarsource, line 46: If we are not consistent within the file, we could still try to be consistent in this signature.
```suggestion
  void doesNotRaiseOnLombokCleanupAnnotatedVariableNoSemantic() {
```

dorian-burihabwa-sonarsource (changes requested):

dorian-burihabwa-sonarsource (approved):
Perfect! Thanks again for diving into this complex issue and crafting a simple but efficient fix!

PR #4960, opened 2024-12-12
SONARJAVA-4973 Fix FP on S1118 with Lombok @UtilityClass in automatic analysis

tomasz-tylenda-sonarsource:
https://sonarsource.atlassian.net/browse/SONARJAVA-4973

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
+++ b/java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
@@ -186,7 +186,7 @@ public class LombokFilter extends BaseTreeVisitorIssueFilter {
   }
 
 
-  private static boolean usesAnnotation(ClassTree classTree, List<String> annotations, boolean checkLocalName) {
+  private static boolean usesAnnotation(ClassTree classTree, List<String> annotations, boolean shouldCheckAnnotationLocalName) {
     SymbolMetadata classMetadata = classTree.symbol().metadata();
 
     for(String fullyQualified: annotations) {
--- a/java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
+++ b/java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
@@ -182,6 +182,12 @@ public class LombokFilter extends BaseTreeVisitorIssueFilter {
   }
 
   private static boolean usesAnnotation(ClassTree classTree, List<String> annotations) {
+    return usesAnnotation(classTree, annotations, !CHECK_LOCAL_NAME);
+  }
+
+  private static final boolean CHECK_LOCAL_NAME = true;
+
+  private static boolean usesAnnotation(ClassTree classTree, List<String> annotations, boolean checkLocalName) {
     SymbolMetadata classMetadata = classTree.symbol().metadata();
 
     for(String fullyQualified: annotations) {
@@ -189,7 +195,7 @@ public class LombokFilter extends BaseTreeVisitorIssueFilter {
         return true;
       }
       // In automatic analysis use only the last part of the annotation.
-      if(classMetadata.isAnnotatedWith(localName(fullyQualified))) {
+      if(checkLocalName && classMetadata.isAnnotatedWith(localName(fullyQualified))) {
         return true;
       }
     }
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -181,12 +182,34 @@ public void visitAssignmentExpression(AssignmentExpressionTree tree) {
   }
 
   private static boolean usesAnnotation(ClassTree classTree, List<String> annotations) {
+    return usesAnnotation(classTree, annotations, !CHECK_LOCAL_NAME);
```
dorian-burihabwa-sonarsource, line 185: When calling `usesAnnotation` with 2 parameters, then we should expect a predictable behavior where the value of 3rd parameter should always be `true` or always be ` false`.
```suggestion
    return usesAnnotation(classTree, annotations, false);
```

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -181,12 +182,34 @@ public void visitAssignmentExpression(AssignmentExpressionTree tree) {
...
+  }
+
+  private static final boolean CHECK_LOCAL_NAME = true;
```
dorian-burihabwa-sonarsource, line 188: If the value is always `true` then we can remove this field that introduces dependencies between static methods and fields.
```suggestion
```

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -181,12 +182,34 @@ public void visitAssignmentExpression(AssignmentExpressionTree tree) {
...
+  }
+
+  @VisibleForTesting
+  static String localName(String fullyQualified) {
+    return fullyQualified.substring(fullyQualified.lastIndexOf('.') + 1);
   }
 
   private static boolean generatesNonPublicConstructor(ClassTree classTree) {
-    if (usesAnnotation(classTree, UTILITY_CLASS)) {
+    if (usesAnnotation(classTree, UTILITY_CLASS, CHECK_LOCAL_NAME)) {
```
dorian-burihabwa-sonarsource, line 212:
```suggestion
    if (usesAnnotation(classTree, UTILITY_CLASS, true)) {
```

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -181,12 +182,34 @@ public void visitAssignmentExpression(AssignmentExpressionTree tree) {
...
+      if(checkLocalName && classMetadata.isAnnotatedWith(localName(fullyQualified))) {
+        return true;
+      }
+    }
+
+    return false;
+  }
+
+  @VisibleForTesting
+  static String localName(String fullyQualified) {
```
dorian-burihabwa-sonarsource, line 207: Nitpick: what about a more [specific name](https://docs.oracle.com/javase/specs/jls/se21/html/jls-6.html#jls-TypeName)?
```suggestion
  static String annotationTypeIdentifier(String fullyQualified) {
```

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
dorian-burihabwa-sonarsource, line 1: We should use some formatting before committing this new code

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
```diff
@@ -181,12 +182,34 @@ public void visitAssignmentExpression(AssignmentExpressionTree tree) {
   }
 
   private static boolean usesAnnotation(ClassTree classTree, List<String> annotations) {
+    return usesAnnotation(classTree, annotations, !CHECK_LOCAL_NAME);
+  }
+
+  private static final boolean CHECK_LOCAL_NAME = true;
+
+  private static boolean usesAnnotation(ClassTree classTree, List<String> annotations, boolean checkLocalName) {
```
dorian-burihabwa-sonarsource, line 190: Nitpick: because we have the notion of checks all over the codebase, I think it would be wiser to rename this parameter to be:
1. More indicative of the context it will be used in
2. More indicative of its use within the method
```suggestion
  private static boolean usesAnnotation(ClassTree classTree, List<String> annotations, boolean shouldCheckAnnotationLocalName) {
```

dorian-burihabwa-sonarsource (changes requested):
We have the substance here, let's try and fix the form

java-checks/src/main/java/org/sonar/java/filters/LombokFilter.java
tomasz-tylenda-sonarsource: I did reformat the code in the IDE. Please let me know if you think my IDE is misconfigured.

PR #4962, opened 2024-12-16
SONARJAVA-5149 report issues with path delimiters in variable declaration

erwan-serandour:
[SONARJAVA-5149](https://sonarsource.atlassian.net/browse/SONARJAVA-5149)

Part of 

[SONARJAVA-5149]: https://sonarsource.atlassian.net/browse/SONARJAVA-5149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/HardcodedURICheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/HardcodedURICheck.java
@@ -264,7 +264,7 @@ public class HardcodedURICheck extends IssuableSubscriptionVisitor {
 
     private void checkPathDelimiter(ExpressionTree expr) {
       ExpressionTree newExpr = ExpressionUtils.skipParentheses(expr);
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/HardcodedURICheck.java
```diff
@@ -179,6 +179,8 @@ private void checkVariable(VariableTree tree) {
 
     String stringLiteral = stringLiteral(initializer);
     if (stringLiteral == null) {
+      // the visitor only report path delimiter on binary expressions, so the initializer must not be a string literal
```
tomasz-tylenda-sonarsource, line 182: It took me a while to understand the meaning of the comment. I suggest changing it, for example, "The check below applies only to expression, so we need it only if the initializer is not a literal."

tomasz-tylenda-sonarsource (approved):

tomasz-tylenda-sonarsource (approved):

PR #4972, opened 2024-12-24
SONARJAVA-5262 Fix FP on S2187 in Cucumber tests with JUnit5.

tomasz-tylenda-sonarsource:
[SONARJAVA-5262](https://sonarsource.atlassian.net/browse/SONARJAVA-5262)

[SONARJAVA-5262]: https://sonarsource.atlassian.net/browse/SONARJAVA-5262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/test/java/checks/tests/NoTestInTestClassCheckCucumberTest.java
+++ b/java-checks-test-sources/default/src/test/java/checks/tests/NoTestInTestClassCheckCucumberTest.java
@@ -13,3 +13,12 @@ class NoTestInTestClassCheckBellPepperTest {} // Noncompliant
 
 @IncludeEngines({"spring", "cucumber"})
 class NoTestInTestClassCheckTwoEnginesTest{}
+
+@NoTestInTestClassIncompatibleAnnotations.IncludeEngines(42)
+class ClassWithFakeIncludeEnginesAnnotationTest {} // Noncompliant
+
+class NoTestInTestClassIncompatibleAnnotations {
+  @interface IncludeEngines {
+    int value();
+  }
+}
--- a/java-checks-test-sources/default/src/test/java/checks/tests/NoTestInTestClassCheckCucumberTest.java
+++ b/java-checks-test-sources/default/src/test/java/checks/tests/NoTestInTestClassCheckCucumberTest.java
@@ -1,0 +1,23 @@
+package checks.tests;
+
+import static io.cucumber.junit.platform.engine.Constants.*;
+
+import org.junit.platform.suite.api.*;
+
+@Suite
+@IncludeEngines("cucumber")
+@SelectClasspathResource("com.project.class.path")
+@ConfigurationParameter(key = GLUE_PROPERTY_NAME, value="com.project.class.path")
+class NoTestInTestClassCheckCucumberStandardTest {}
+
+@Suite
+@org.junit.platform.suite.api.IncludeEngines("cucumber")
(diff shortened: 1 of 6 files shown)


tomasz-tylenda-sonarsource:
This PR is not stale. It was on hold during holidays.

java-checks-test-sources/default/src/test/java/checks/tests/NoTestInTestClassCheckCucumberTest.java
```diff
@@ -0,0 +1,23 @@
+package checks.tests;
+
+import static io.cucumber.junit.platform.engine.Constants.*;
+
+import org.junit.platform.suite.api.*;
+
+@Suite
+@IncludeEngines("cucumber")
+@SelectClasspathResource("com.project.class.path")
+@ConfigurationParameter(key = GLUE_PROPERTY_NAME, value="com.project.class.path")
```
dorian-burihabwa-sonarsource, line 10: Is this part of the code needed to test the scenario in the ticket?

java-checks/src/main/java/org/sonar/java/checks/tests/NoTestInTestClassCheck.java
```diff
@@ -199,6 +201,28 @@ private static boolean checkRunWithType(Symbol.TypeSymbol value, String... runne
...
+  private static boolean annotatedIncludeEnginesCucumber(Symbol.TypeSymbol symbol) {
+    for (var annotation: symbol.metadata().annotations()) {
+      if (annotation.symbol().type().fullyQualifiedName().endsWith("IncludeEngines")) {
+        // values are not available in automatic analysis, so assume "cucumber" is there
+        if (annotation.values().isEmpty()) {
+          return true;
+        }
+        // otherwise check the list
+        boolean containsCucumber = annotation.values().stream().anyMatch(annotationValue ->
+          Objects.deepEquals(annotationValue.value(), new Object[]{"cucumber"}));
```
dorian-burihabwa-sonarsource, line 217: What happens here if we have a case where we have multiple values for the annotation? See this [example](https://github.com/yuzawa-san/googolplex-theater/blob/e3a3b2c7edb68646d6c6e5a4972577ab171ab427/src/test/java/com/jyuzawa/googolplex_theater/RunCucumber.java#L14)

dorian-burihabwa-sonarsource (changes requested):

java-checks-test-sources/default/src/test/java/checks/tests/NoTestInTestClassCheckCucumberTest.java
tomasz-tylenda-sonarsource: No and neither is `@Suite`. I originally included everything that was reported to us to make the test more realistic. Here I deleted the annotation that we don't need. This change let me remove cucumber from `pom.xml`.

dorian-burihabwa-sonarsource (changes requested):
I think we can complete the coverage by testing a class test annotated with type clashes
```java
@IncompatibleAnnotations.IncludeEngines(42)
class ClassWithFakeIncludeEnginesAnnotationTest {// Noncompliant
}

class IncompatibleAnnotations {
  @interface IncludeEngines {
    int value();
  }
}
```

dorian-burihabwa-sonarsource (approved):

PR #4981, opened 2025-01-10
SONARJAVA-5259 Migrate secondary locations in RedundantRegexAlternativesCheckSample to the new syntax

tomasz-tylenda-sonarsource:
[SONARJAVA-5259](https://sonarsource.atlassian.net/browse/SONARJAVA-5259)

[SONARJAVA-5259]: https://sonarsource.atlassian.net/browse/SONARJAVA-5259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/regex/RedundantRegexAlternativesCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/regex/RedundantRegexAlternativesCheckSample.java
@@ -9,60 +9,96 @@ public class RedundantRegexAlternativesCheckSample {
   public void f(Pattern pattern) {
     f(compile("" +
       "." +
-      "|a")); // Noncompliant  {{Remove or rework this redundant alternative.}} [[sc=9;ec=10;secondary=-1]]
+  //   ^>
+      "|a")); // Noncompliant  {{Remove or rework this redundant alternative.}}
+  //    ^
     f(compile("" +
-      "a" + // Noncompliant [[sc=8;ec=9;secondary=+1]]
+      "a" + // Noncompliant
+  //   ^
       "|."));
+  //    ^<
     f(compile("" +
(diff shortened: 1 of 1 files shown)


java-checks-test-sources/default/src/main/java/checks/regex/RedundantRegexAlternativesCheckSample.java
```diff
@@ -9,60 +9,96 @@ public class RedundantRegexAlternativesCheckSample {
...
+      "|a" + // Noncompliant
+//      ^
+//     ^@-3<
+//      ^@+8<
+//      ^@+9<
+      "|b" + // Noncompliant
+//      ^
+//      ^@-7<
+//      ^@+4<
+//      ^@+5<
```
dorian-burihabwa-sonarsource, line 53: Is there way to declare the secondary locations closer to their location? We are gaining in our ability to detect regression but I am not sure this is more readable. WDYT?

dorian-burihabwa-sonarsource (approved):
The PR looks good overall to me but I left a minor comment on the clarity of 1 specific issue. It might be worth spending 5 min looking for a more readable alternative but the change could be merged as-is IMO

java-checks-test-sources/default/src/main/java/checks/regex/RedundantRegexAlternativesCheckSample.java
tomasz-tylenda-sonarsource: I am not aware of any way to simplify this. This is definitely less readable than the previous syntax and hard to modify. I already spent enough time on this, so I'll follow up only if I find similar examples elsewhere.

PR #4984, opened 2025-01-21
SONARJAVA-5232 S1192: Do not consider arguments in Throwables when checking for duplicated strings

tomasz-tylenda-sonarsource:
[SONARJAVA-5232](https://sonarsource.atlassian.net/browse/SONARJAVA-5232)

[SONARJAVA-5232]: https://sonarsource.atlassian.net/browse/SONARJAVA-5232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
@@ -165,7 +165,7 @@ class DuplicatedExceptionArguments {
     return s + suffix;
   }
 
-  private void nestedString2(int r) {
+  private void areNonCompliantWhenPassedToMethod(int r) {
     if(r == 0) {
       throw new IllegalArgumentException(transform("nested string 2", "AAA")); // Noncompliant {{Define a constant instead of duplicating this literal "nested string 2" 3 times.}}
     } else if (r == 1) {
--- a/java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
@@ -114,8 +114,8 @@ class Coverage {
   private final String prevRightNull = 3 + "SELECT";
 }
 
-class ExceptionArguments {
-  private void simple(int r) {
+class DuplicatedExceptionArguments {
+  private void areCompliantByDefault(int r) {
     if(r == 0) {
       throw new IllegalArgumentException("simple IAE");
     } else if (r == 1) {
--- a/java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
@@ -116,7 +116,7 @@ class Coverage {
 
 class DuplicatedExceptionArguments {
   private void areCompliantByDefault(int r) {
-    if(r == 0) {
+    if (r == 0) {
       throw new IllegalArgumentException("simple IAE");
     } else if (r == 1) {
       throw new IllegalArgumentException("simple IAE");
@@ -132,7 +132,7 @@ class DuplicatedExceptionArguments {
   }
 
   private void twoArgs(int r, Throwable throwable) throws MyException {
-    if(r == 0) {
+    if (r == 0) {
       throw new MyException("my exception message", throwable);
     } else if (r == 1) {
       throw new MyException("my exception message", throwable);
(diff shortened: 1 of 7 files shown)


java-checks/src/main/java/org/sonar/java/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -89,13 +90,29 @@ private static List<JavaFileScannerContext.Location> secondaryLocations(Collecti
...
   }
 
+  /**
+   * Verify that <code>tree</code> is an argument in
+   * <code>throw new SomeException(arg1, arg2, ...)</code>.
+   * For simplicity and to avoid surprises there is no recursion on arguments.
+   */
+  private static boolean isThrowableArgument(LiteralTree tree) {
+    return
+      Optional.ofNullable(tree.parent())
```
dorian-burihabwa-sonarsource, line 107:
```suggestion
    return Optional.ofNullable(tree.parent())
```

java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
```diff
@@ -113,3 +113,65 @@ class Coverage {
...
+    } else if (r == 2) {
+      throw new IllegalArgumentException("nested string 1");
+    }
+  }
+
+  private String transform(String s, String suffix) {
+    return s + suffix;
+  }
+
+  private void nestedString2(int r) {
```
dorian-burihabwa-sonarsource, line 168: I think it would be useful here to rename the block by something a little more relevant
```suggestion
  private void areNonCompliantWhenPassedToMethod(int r) {
```

java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
```diff
@@ -113,3 +113,65 @@ class Coverage {
   private final String prevLeftNull = "SELECT" + 3;
   private final String prevRightNull = 3 + "SELECT";
 }
+
+class ExceptionArguments {
+  private void simple(int r) {
```
dorian-burihabwa-sonarsource, line 118: Let's help our future selves by renaming some of this methods or adding some comments.
```suggestion
class DuplicatedExceptionArguments {
  private void areCompliantByDefault(int r) {
```

java-checks/src/main/java/org/sonar/java/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -89,13 +90,29 @@ private static List<JavaFileScannerContext.Location> secondaryLocations(Collecti
...
+        .filter(t -> t.is(Tree.Kind.ARGUMENTS))
+        .map(Tree::parent)
+        .filter(t -> t.is(Tree.Kind.NEW_CLASS))
+        .map(Tree::parent)
+        .filter(t -> t.is(Tree.Kind.THROW_STATEMENT))
+        .isPresent();
```
dorian-burihabwa-sonarsource, line 113: The approach should be robust enough to work with broken semantics.

java-checks/src/main/java/org/sonar/java/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -89,13 +90,29 @@ private static List<JavaFileScannerContext.Location> secondaryLocations(Collecti
...
 
+  /**
+   * Verify that <code>tree</code> is an argument in
+   * <code>throw new SomeException(arg1, arg2, ...)</code>.
+   * For simplicity and to avoid surprises there is no recursion on arguments.
+   */
+  private static boolean isThrowableArgument(LiteralTree tree) {
+    return
+      Optional.ofNullable(tree.parent())
+        .filter(t -> t.is(Tree.Kind.ARGUMENTS))
```
dorian-burihabwa-sonarsource, line 108: Curiosity: Do we know what happens when a literal like "ABC" is part of the concatenation?
```java
  if (a == 1) {
    throw new IllegalArgumentException("ABC" + a);
  } else if (a == 2) {
    throw new IllegalArgumentException("ABC" + a);
  } else {
    throw new IllegalArgumentException("ABC" + a);
  }
```

tomasz-tylenda-sonarsource: We were still reporting these and I thought about this case too, so I decided to relax the rule here as well. There is a new test case and I also had to trim our golden files in ruling tests.

java-checks/src/main/java/org/sonar/java/checks/StringLiteralDuplicatedCheck.java
tomasz-tylenda-sonarsource: I added a test with `.withoutSemantic()`. Please reopen if you meant something else. We don't need any external dependencies for this rule, but the file should complile.

dorian-burihabwa-sonarsource (commented):
LGTM 👍🏿

dorian-burihabwa-sonarsource (approved):

PR #4994, opened 2025-02-04
SONARJAVA-5273 FP for S125 for Apache 2.0 license header

alban-auzeill:
[SONARJAVA-5273](https://sonarsource.atlassian.net/browse/SONARJAVA-5273)

[SONARJAVA-5273]: https://sonarsource.atlassian.net/browse/SONARJAVA-5273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CommentedOutCodeLineCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CommentedOutCodeLineCheck.java
@@ -19,6 +19,7 @@ package org.sonar.java.checks;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 import javax.annotation.Nullable;
 import org.apache.commons.lang3.StringUtils;
 import org.sonar.check.Rule;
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/CommentedOutCodeLineCheck.java
```diff
@@ -53,6 +58,15 @@ public List<Tree.Kind> nodesToVisit() {
...
 
+  @Override
+  public void setContext(JavaFileScannerContext context) {
+    super.setContext(context);
+    compilationUnitFirstTokenPosition = Optional.ofNullable(context.getTree())
+      .map(Tree::firstToken)
+      .map(SyntaxToken::range)
+      .map(Range::start).orElse(FILE_START);
+  }
+
```
dorian-burihabwa-sonarsource, line 69: Would it make sense to reset `compilationUnitFirstTokenPosition` to `FILE_START` when leaving the file?
```suggestion

  @Override
  public void leaveFile(JavaFileScannerContext context) {
    compilationUnitFirstTokenPosition = FILE_START;
  }

```

dorian-burihabwa-sonarsource (commented):
The change looks good and simple enough to revisit in the future. I just have one question about holding on to position information when leaving the file. Let me know what you think

java-checks/src/main/java/org/sonar/java/checks/CommentedOutCodeLineCheck.java
alban-auzeill: I thought about it and decided that there's not enough benefit to justify adding some code. Because the Position class is just a small class with two `int` fields. There's not risk to keep the full AST in memory in case of exception. And there's no risk to use an invalid `compilationUnitFirstTokenPosition` form the previous AST because of the `setContext` execution.

Conclusion: Yes we can but I'm not sure we should.

dorian-burihabwa-sonarsource (approved):
LGTM! Thanks for including the reset at the end!

PR #4998, opened 2025-02-04
SONARJAVA-5317 S7178 Injecting data into static fields is not supported by Spring

erwan-serandour:
[SONARJAVA-5317](https://sonarsource.atlassian.net/browse/SONARJAVA-5317)

[SONARJAVA-5317]: https://sonarsource.atlassian.net/browse/SONARJAVA-5317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
```diff
@@ -0,0 +1,90 @@
...
+  private static final Set<String> INJECTIONS_ANNOTATIONS = Set.of(
+    "javax.inject.Inject",
+    "org.springframework.beans.factory.annotation.Autowired",
+    "jakarta.inject.Inject",
+    "org.springframework.beans.factory.annotation.Value"
+  );
+
+  private static final String STATIC_FIELD_MESSAGE = "Remove the injection annotation targeting the static field.";
+  private static final String STATIC_METHOD_MESSAGE = "Remove the injection annotation targeting the static method.";
+  private static final String STATIC_PARAMETER_MESSAGE = "Remove the injection annotation targeting the parameter.";
```
leonardo-pilastri-sonarsource, line 44: I thinks it reads better like this since we are reporting on the annotation
```suggestion
  private static final String STATIC_FIELD_MESSAGE = "Remove this injection annotation targeting a static field.";
  private static final String STATIC_METHOD_MESSAGE = "Remove this injection annotation targeting a static method.";
  private static final String STATIC_PARAMETER_MESSAGE = "Remove this injection annotation targeting a parameter.";
```

java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
```diff
@@ -0,0 +1,90 @@
...
+        .filter(child -> child instanceof VariableTree v && ModifiersUtils.hasModifier(v.modifiers(), Modifier.STATIC))
+        .map(VariableTree.class::cast);
+
+      staticFields
+        .map(v -> injectionAnnotations(v.modifiers()))
+        .filter(anns -> !anns.isEmpty())
+        .forEach(anns ->
+          anns.forEach(ann -> reportIssue(ann, STATIC_FIELD_MESSAGE))
+        );
+      return;
```
leonardo-pilastri-sonarsource, line 64: I think since we are handling two separate and parallel cases, it could be clearer to have 
```
if (tree instanceof ClassTree clazz) {
...
} else if (tree instanceof MethodTree method) {
...
}
```
instead of the early return

java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
```diff
@@ -0,0 +1,90 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree instanceof ClassTree clazz) {
+      Stream<VariableTree> staticFields = clazz.members().stream()
+        .filter(child -> child instanceof VariableTree v && ModifiersUtils.hasModifier(v.modifiers(), Modifier.STATIC))
+        .map(VariableTree.class::cast);
+
+      staticFields
+        .map(v -> injectionAnnotations(v.modifiers()))
+        .filter(anns -> !anns.isEmpty())
```
leonardo-pilastri-sonarsource, line 60: We can avoid filtering empty annotations, as the next operation will not perform any action on them
```suggestion
```

java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
```diff
@@ -0,0 +1,90 @@
...
+      //report on parameters
+      method.parameters().stream()
+        .map(p -> injectionAnnotations(p.modifiers()))
+        .filter(anns -> !anns.isEmpty())
+        .forEach(anns -> anns.forEach(ann -> reportIssue(ann, STATIC_PARAMETER_MESSAGE)));
+
+    }
+  }
+
+  private static List<AnnotationTree> injectionAnnotations(ModifiersTree m) {
```
leonardo-pilastri-sonarsource, line 84: Not a strong opinion here, but maybe this could be named `getInjectionAnnotations` or `mapToInjectionAnnotations` to communicate the action better

java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
```diff
@@ -0,0 +1,90 @@
...
+    if (ModifiersUtils.hasModifier(method.modifiers(), Modifier.STATIC)) {
+
+      //report on method annotations
+      injectionAnnotations(method.modifiers())
+        .forEach(ann -> reportIssue(ann, STATIC_METHOD_MESSAGE));
+
+      //report on parameters
+      method.parameters().stream()
+        .map(p -> injectionAnnotations(p.modifiers()))
+        .filter(anns -> !anns.isEmpty())
```
leonardo-pilastri-sonarsource, line 78:
```suggestion
```

java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
```diff
@@ -0,0 +1,90 @@
...
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.METHOD, Tree.Kind.CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree instanceof ClassTree clazz) {
+      Stream<VariableTree> staticFields = clazz.members().stream()
+        .filter(child -> child instanceof VariableTree v && ModifiersUtils.hasModifier(v.modifiers(), Modifier.STATIC))
```
leonardo-pilastri-sonarsource, line 55: Not a strong opinion here, this looks better but relies on semantics, but I think it should be ok, as without semantics we don't want to raise issues anyways
```suggestion
        .filter(child -> child instanceof VariableTree v && v.symbol().isStatic())
```

java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
```diff
@@ -0,0 +1,90 @@
...
+        .filter(anns -> !anns.isEmpty())
+        .forEach(anns ->
+          anns.forEach(ann -> reportIssue(ann, STATIC_FIELD_MESSAGE))
+        );
+      return;
+
+    }
+
+    MethodTree method = (MethodTree)tree;
+    if (ModifiersUtils.hasModifier(method.modifiers(), Modifier.STATIC)) {
```
leonardo-pilastri-sonarsource, line 69:
```suggestion
    if (method.symbol().isStatic()) {
```

java-checks/src/test/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheckTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+class StaticFieldInjectionNotSupportedCheckTest {
+
+  @Test
+  void test_compiling(){
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/spring/StaticFieldInjectionNotSupportedCheckSample.java"))
+      .withCheck(new StaticFieldInjectionNotSupportedCheck())
+      .verifyIssues();
+  }
```
leonardo-pilastri-sonarsource, line 32: I would add a unit test to guarantee that we do not raise issues when we don't know where annotations are coming from
```
  @Test
  void test_no_semantics(){
    CheckVerifier.newVerifier()
      .onFile(mainCodeSourcesPath("checks/spring/StaticFieldInjectionNotSupportedCheckSample.java"))
      .withCheck(new StaticFieldInjectionNotSupportedCheck())
      .withoutSemantic()
      .verifyNoIssues();
  }
  ```

File: java-checks-test-sources/default/src/main/java/checks/spring/StaticFieldInjectionNotSupportedCheckSample.java [hunk 9]
```diff
@@ -0,0 +1,91 @@
...
+  @Target({ElementType.FIELD})
+  @interface AnAnnotation {
+
+  }
+
+  @Target({ElementType.FIELD})
+  @Autowired
+  @interface MyAutowired {}
+
+  // we should only raise if the DI framework is spring
```
leonardo-pilastri-sonarsource, line 21: Is this something we should check by looking at the annotation below? We are currently ignoring the context

leonardo-pilastri-sonarsource (changes requested):

leonardo-pilastri-sonarsource (changes requested):
Ah forgot one thing, hitting Ctrl+Alt+L on my intellij reformatted a bunch of lines, can you also make sure to reformat before pushing?

java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
erwan-serandour: :+1:

java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
erwan-serandour: :+1:

java-checks/src/main/java/org/sonar/java/checks/spring/StaticFieldInjectionNotSupportedCheck.java
erwan-serandour: Indeed it is shorter :pray:, I did not know about this api. Thank you :+1:

leonardo-pilastri-sonarsource (changes requested):
LGTM, just need some reformatting, make sure you have the sonarsource code style imported in your Intellij
https://github.com/SonarSource/sonar-developer-toolset/blob/master/intellij/codestyle_sonar_developer_toolset.xml

leonardo-pilastri-sonarsource (approved):
Looks good!

PR #5001, opened 2025-02-05
SONARJAVA-5319 S7184 @Scheduled annotation should only be applied to no-arg methods

tomasz-tylenda-sonarsource:
SONARJAVA-5319


java-checks/src/main/java/org/sonar/java/checks/spring/ScheduledOnlyOnNoArgMethodCheck.java
```diff
@@ -0,0 +1,69 @@
...
+  public void visitNode(Tree tree) {
+    MethodTree methodTree = (MethodTree) tree;
+
+    if (hasParameters(methodTree)) {
+      var scheduledAnnotations = getScheduledAnnotations(methodTree);
+      if (!scheduledAnnotations.isEmpty()) {
+        var secondaryLocations = scheduledAnnotations.stream()
+            .map(annotation ->
+              new JavaFileScannerContext.Location("Triggered by this annotation", annotation.annotationType()))
+            .toList();
```
leonardo-pilastri-sonarsource, line 50: I think there's an extra tab here, can you try to reformat the file?
```suggestion
          .map(annotation ->
            new JavaFileScannerContext.Location("Triggered by this annotation", annotation.annotationType()))
          .toList();
```

java-checks/src/test/java/org/sonar/java/checks/spring/ScheduledOnlyOnNoArgMethodCheckTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.jupiter.api.Test;
+import org.sonar.java.checks.verifier.CheckVerifier;
+
+import static org.sonar.java.checks.verifier.TestUtils.mainCodeSourcesPath;
+
+class ScheduledOnlyOnNoArgMethodCheckTest {
+  @Test
+  void test() {
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/spring/ScheduledOnlyOnNoArgMethodCheckSample.java"))
```
leonardo-pilastri-sonarsource, line 28: Not a strong opinion here, but you might consider having 2 constants in this class, one for the filename and one for the check, to avoid changing the string twice if we move the file and to avoid instantiating the check twice

leonardo-pilastri-sonarsource (changes requested):
A couple of minor things, otherwise it looks good to me!

java-checks/src/test/java/org/sonar/java/checks/spring/ScheduledOnlyOnNoArgMethodCheckTest.java
tomasz-tylenda-sonarsource: Done.

leonardo-pilastri-sonarsource (approved):

PR #5003, opened 2025-02-06
SONARJAVA-5318 S7180: `@Cache*` annotations should only be applied on concrete classes

erwan-serandour:
[SONARJAVA-5318](https://sonarsource.atlassian.net/browse/SONARJAVA-5318)

[SONARJAVA-5318]: https://sonarsource.atlassian.net/browse/SONARJAVA-5318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


java-checks/src/main/java/org/sonar/java/checks/spring/CacheAnnotationsShouldOnlyBeAppliedToConcreteClassesCheck.java
```diff
@@ -0,0 +1,77 @@
...
+import org.sonar.plugins.java.api.tree.AnnotationTree;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.ModifiersTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S7180")
+public class CacheAnnotationsShouldOnlyBeAppliedToConcreteClassesCheck extends IssuableSubscriptionVisitor {
+
+  private static final String ISSUE_MESSAGE = "Move this \"@%s\" annotation from interface to a concrete class.";
```
leonardo-pilastri-sonarsource, line 33: What about:
```suggestion
  private static final String ISSUE_MESSAGE = "\"@%s\" annotation should only be applied to concrete classes.";
```

java-checks-test-sources/default/src/main/java/checks/spring/CacheAnnotationsShouldOnlyBeAppliedToConcreteClassesCheckSample.java
```diff
@@ -0,0 +1,89 @@
...
+  public interface InterfaceCacheable {
+    @Cacheable("aCache") // Noncompliant {{Move this "@Cacheable" annotation from interface to a concrete class.}}
+  //^^^^^^^^^^^^^^^^^^^^
+    String getData(String id);
+  }
+
+
+  @Cacheable("aCache")
+  abstract class AbstractClassCacheable {
+    @Cacheable("aCache")
```
leonardo-pilastri-sonarsource, line 21: After some testing I can confirm that 2 different implementations of an abstract class will have conflicts with the cache if it's configured this way, so we should raise also in this case. Let's discuss the specifics and I can share my experiment

leonardo-pilastri-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/spring/CacheAnnotationsShouldOnlyBeAppliedToConcreteClassesCheck.java
erwan-serandour: :+1:

leonardo-pilastri-sonarsource (approved):

PR #5004, opened 2025-02-06
SONARJAVA-5041 S5786 should raise an issue on JUnit5 annotated lifecycle methods with a public modifier

alban-auzeill:
[SONARJAVA-5041](https://sonarsource.atlassian.net/browse/SONARJAVA-5041)

[SONARJAVA-5041]: https://sonarsource.atlassian.net/browse/SONARJAVA-5041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/tests/AbstractJUnit5NotCompliantModifierChecker.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/tests/AbstractJUnit5NotCompliantModifierChecker.java
@@ -37,8 +37,11 @@ import org.sonar.plugins.java.api.tree.VariableTree;
 public abstract class AbstractJUnit5NotCompliantModifierChecker extends IssuableSubscriptionVisitor {
 
   public enum ModifierScope {
+    // annotation like @Nested that applies to class
     CLASS,
+    // annotation like @BeforeAll that applies to static class method
     CLASS_METHOD,
+    // annotation like @BeforeEach that applies to non-static method
     INSTANCE_METHOD
   }
 
@@ -94,7 +97,8 @@ public abstract class AbstractJUnit5NotCompliantModifierChecker extends Issuable
 
     raiseIssueOnMethods(junit5ClassMethods, ModifierScope.CLASS_METHOD);
     raiseIssueOnMethods(junit5InstanceMethods, ModifierScope.INSTANCE_METHOD);
-    if (!junit5InstanceMethods.isEmpty()) {
+    boolean classHasJunit5InstanceMethods = !junit5InstanceMethods.isEmpty();
+    if (classHasJunit5InstanceMethods) {
       raiseIssueOnClass(nonJunit5Methods, classTree);
     }
   }
@@ -118,9 +122,10 @@ public abstract class AbstractJUnit5NotCompliantModifierChecker extends Issuable
       .anyMatch(AbstractJUnit5NotCompliantModifierChecker::isPublicStatic);
 
     // Can we change the visibility of the class?
-    if (!hasPublicStaticMethods && !hasPublicStaticFields) {
-      raiseIssueOnNotCompliantModifiers(classTree.modifiers(), ModifierScope.CLASS);
+    if (hasPublicStaticMethods || hasPublicStaticFields) {
+      return;
     }
+    raiseIssueOnNotCompliantModifiers(classTree.modifiers(), ModifierScope.CLASS);
   }
 
   private static boolean isPublicStatic(ModifiersTree modifiers) {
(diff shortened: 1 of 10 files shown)


java-checks/src/main/java/org/sonar/java/checks/tests/AbstractJUnit5NotCompliantModifierChecker.java
```diff
@@ -36,21 +36,27 @@
...
       .forRule(this)
       .onTree(modifier)
-      .withMessage(String.format(WRONG_MODIFIER_ISSUE_MESSAGE, modifier.keyword().text()))
+      .withMessage(WRONG_MODIFIER_ISSUE_MESSAGE, modifier.keyword().text())
       .withQuickFix(() ->
-        JavaQuickFix.newQuickFix("Remove modifier")
-          .addTextEdit(JavaTextEdit.removeTree(modifier))
-          .build())
+        JavaQuickFix.newQuickFix("Remove \"%s\" modifier", modifier.keyword().text())
+          .addTextEdit(JavaTextEdit.removeTextSpan(AnalyzerMessage.textSpanBetween(modifier, true, QuickFixHelper.nextToken(modifier), false)))
```
dorian-burihabwa-sonarsource, line 58: Why do we have to change the way that we generate the quickfix?

java-checks/src/main/java/org/sonar/java/checks/tests/AbstractJUnit5NotCompliantModifierChecker.java
```diff
@@ -65,23 +71,43 @@ public void visitNode(Tree tree) {
...
+            junit5ClassMethods.add(method);
+          }
+        } else {
+          nonJunit5Methods.add(method);
+        }
+      });
+
+    raiseIssueOnMethods(junit5ClassMethods, ModifierScope.CLASS_METHOD);
+    raiseIssueOnMethods(junit5InstanceMethods, ModifierScope.INSTANCE_METHOD);
+    if (!junit5InstanceMethods.isEmpty()) {
```
dorian-burihabwa-sonarsource, line 97: Can we extract this for readability?
```suggestion
    boolean classHasJunit5InstanceMethods = !junit5InstanceMethods.isEmpty();
    if (classHasJunit5InstanceMethods) {
```

java-checks/src/main/java/org/sonar/java/checks/tests/AbstractJUnit5NotCompliantModifierChecker.java
```diff
@@ -91,23 +117,19 @@ public void visitNode(Tree tree) {
...
 
-    if (hasPublicStaticMethods || hasPublicStaticFields) {
-      // we can not ask for a change of visibility of the class
-      return;
-    }
-
-    if (!testMethods.isEmpty()) {
-      raiseIssueOnNotCompliantModifiers(classTree.modifiers(), false);
+    // Can we change the visibility of the class?
+    if (!hasPublicStaticMethods && !hasPublicStaticFields) {
```
dorian-burihabwa-sonarsource, line 121: I am not sure I like seeing doubling down on negatives here

java-checks/src/main/java/org/sonar/java/checks/tests/AbstractJUnit5NotCompliantModifierChecker.java
```diff
@@ -36,21 +36,27 @@
 
 public abstract class AbstractJUnit5NotCompliantModifierChecker extends IssuableSubscriptionVisitor {
 
+  public enum ModifierScope {
+    CLASS,
+    CLASS_METHOD,
+    INSTANCE_METHOD
+  }
```
dorian-burihabwa-sonarsource, line 43: It would be very helpful to document how these enums represent and how they can be used

dorian-burihabwa-sonarsource (changes requested):
The change looks good overall but it looks like we don't have examples of issues raised on classes by S5786 with the new logic. It probably would help if we could get some documentation on the modifier scope enums to understand their usage

java-checks/src/main/java/org/sonar/java/checks/tests/AbstractJUnit5NotCompliantModifierChecker.java
alban-auzeill: To have only one `raiseIssueOnNonCompliantModifier` implementation.

dorian-burihabwa-sonarsource (approved):

PR #5008, opened 2025-02-07
SONARJAVA-5256 Fix false positives on S2699 for AssertJ assertions

romainbrenguier:
[SONARJAVA-5256](https://sonarsource.atlassian.net/browse/SONARJAVA-5256)

There are many methods that can be used to write assertions with AssertJ. The previous pattern was only recognizing a couple of assertJ classes.
With this PR we generalize the pattern for recognizing AssertJ assertion methods so that it applies to many of these classes.

[SONARJAVA-5256]: https://sonarsource.atlassian.net/browse/SONARJAVA-5256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/helpers/UnitTestUtils.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/helpers/UnitTestUtils.java
@@ -20,12 +20,16 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import javax.annotation.Nullable;
+
+import org.sonar.java.annotations.VisibleForTesting;
 import org.sonar.java.model.ExpressionUtils;
 import org.sonar.plugins.java.api.semantic.MethodMatchers;
 import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.semantic.SymbolMetadata;
+import org.sonar.plugins.java.api.semantic.Type;
 import org.sonar.plugins.java.api.tree.ClassTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodTree;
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/helpers/UnitTestUtils.java
```diff
@@ -42,6 +44,13 @@ public final class UnitTestUtils {
       "|laxCheckpoint|succeedingThenComplete");
   private static final Pattern TEST_METHODS_PATTERN = Pattern.compile("test.*|.*Test");
 
+  private static final List<String> ASSERTJ_ASSERTION_METHODS_PREFIXES = List.of("has", "doesNot", "is", "contains", "assert");
+
+  @VisibleForTesting
+  static final Pattern ASSERTJ_ASSERTION_METHODS_PATTERN = Pattern.compile(
+    String.join("|", "returns", "contains", "allMatch", "is", "has", "satisfies") + "|" +
+      "((" + String.join("|", ASSERTJ_ASSERTION_METHODS_PREFIXES) + ")" + "[A-Z].*)");
```
alban-auzeill, line 52: I don't see the advantage of this indirection `String.join("|", "a", "b"))` instead of just `"a|b"`. And I think we could treat all those elements as `method name` or `method name prefix`, I don't have in mind a false-negative if we do so. So I suggest:
```suggestion
  @VisibleForTesting
  static final Pattern ASSERTJ_ASSERTION_METHODS_PATTERN = Pattern.compile(
    "(allMatch|assert|contains|doesNot|has|is|returns|satisfies)([A-Z].*)?");
```

alban-auzeill (changes requested):

alban-auzeill (changes requested):

alban-auzeill (approved):
Thank you for the changes!

PR #5017, opened 2025-02-10
SONARJAVA-5146 Fix false positive in Optional.map lambdas in rule S5411

romainbrenguier:
[SONARJAVA-5146](https://sonarsource.atlassian.net/browse/SONARJAVA-5146)

When we have a `Optional.map` called on a lambda, we now the value for the parameters will never be null, so the check for rule S5411 is not useful and can be skipped. 
We this PR we keep track of the variables in the current scope that are known to be safe. 

[SONARJAVA-5146]: https://sonarsource.atlassian.net/browse/SONARJAVA-5146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/BoxedBooleanExpressionsCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/BoxedBooleanExpressionsCheckSample.java
@@ -3,11 +3,46 @@ package checks;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Function;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import javax.validation.constraints.NotNull;
 
 class BoxedBooleanExpressionsCheckSample {
+  public String mapOptionalBoolean() {
+    return optionalBoolean()
+      // by construction we know that b is not null, so using a primitive boolean expression is not necessary.
+      .map(b -> b ? "truthy" : "falsey")
+      .orElse("mystery");
+  }
+
+  public String lambdaWithBooleanParameter() {
+    Function<Boolean, String> function = b -> b ? "truthy" : "falsey"; // Noncompliant
+    return  "foo";
+  }
+
+  public Optional<Boolean> optionalBoolean() {
+    return Optional.of(true);
+  }
+
+  public String mapOptionalBoolean_extraBrackets() {
+    return optionalBoolean()
+      .map((b) -> b ? true : false)
+      .map((Boolean b) -> b ? "truthy" : "falsey")
+      .orElse("mystery");
+  }
+
+  public String mapOptionalBoolean2() {
+    Boolean b1 = getSurprizeBoxedBoolean();
+    return optionalBoolean()
+      .map(b -> b1 ? "truthy" : "falsey") // Noncompliant
+      .orElse("mystery");
+  }
+  public String lambdaUsedInMapLater() {
+    // FP because the lambda is only called in places where the argument cannot be null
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -116,6 +123,26 @@ public void visitIfStatement(IfStatementTree tree) {
...
   }
 
+  @Override
+  public void visitLambdaExpression(LambdaExpressionTree lambdaExpressionTree) {
+    // parameters on lambda applied on optional are always non-null
+    if (lambdaExpressionTree.parent() instanceof Arguments arguments &&
+      arguments.parent() instanceof MethodInvocationTree methodInvocationTree
+      && MethodMatchers.create().ofTypes("java.util.Optional")
+      .names("filter", "flatMap", "ifPresent", "ifPresentOrElse", "map").withAnyParameters()
+      .build().matches(methodInvocationTree)) {
```
romainbrenguier, line 133: Extract method matcher to static field

java-checks/src/main/java/org/sonar/java/checks/BoxedBooleanExpressionsCheck.java
```diff
@@ -67,6 +69,11 @@ public class BoxedBooleanExpressionsCheck extends BaseTreeVisitor implements Jav
   private static final Map<Tree, IfStatementTree> ifStatementCache = new HashMap<>();
   private static final Map<Symbol, ExpressionTree> firstNullCheckCache = new HashMap<>();
 
+  /**
+   * Symbols we know are non-null, so can safely be used in boolean expressions
+   */
+  private final Deque<Symbol> safeSymbols = new ArrayDeque<>();
```
romainbrenguier, line 75: We can use a Set instead

java-checks-test-sources/default/src/main/java/checks/BoxedBooleanExpressionsCheckSample.java
```diff
@@ -3,12 +3,36 @@
...
 class BoxedBooleanExpressionsCheckSample {
 
+  public Optional<Boolean> optionalBoolean() {
+    return Optional.of(true);
+  }
+
+  public String mapOptionalBoolean() {
+    return optionalBoolean()
+      // by construction we know that b is not null, so using a primitive boolean expression is not necessary.
+      .map(b -> b ? "truthy" : "falsey")
```
romainbrenguier, line 20: Test what happens with `(b) -> b ? ...` or `(Boolean b) -> ....`

java-checks-test-sources/default/src/main/java/checks/BoxedBooleanExpressionsCheckSample.java
```diff
@@ -3,12 +3,36 @@
...
+    Boolean b1 = getSurprizeBoxedBoolean();
+    return optionalBoolean()
+      .map(b -> b1 ? "truthy" : "falsey") // Noncompliant
+      .orElse("mystery");
+  }
+
+  public String lambdaWithBooleanParameter() {
+    Function<Boolean, String> function = b -> b ? "truthy" : "falsey"; // Noncompliant
+    return function.apply(null);
+  }
```
romainbrenguier, line 34: Add FN examples with variable declaration for lambda and use later

alban-auzeill (approved):

PR #5022, opened 2025-02-11
SONARJAVA-5336 S1068 Fix FP with @Getter and @Setter annotations in automatic analysis

tomasz-tylenda-sonarsource:
[SONARJAVA-5336](https://sonarsource.atlassian.net/browse/SONARJAVA-5336)

[SONARJAVA-5336]: https://sonarsource.atlassian.net/browse/SONARJAVA-5336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/test/java/org/sonar/java/checks/unused/UnusedPrivateFieldCheckTest.java
+++ b/java-checks/src/test/java/org/sonar/java/checks/unused/UnusedPrivateFieldCheckTest.java
@@ -33,6 +33,15 @@ class UnusedPrivateFieldCheckTest {
   }
 
   @Test
+  void test_without_semantic() {
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/unused/UnusedPrivateFieldCheck.java"))
+      .withCheck(new UnusedPrivateFieldCheck())
+      .withoutSemantic()
+      .verifyIssues();
+  }
+
+  @Test
   void test_non_compiling() {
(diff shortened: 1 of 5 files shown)


java-checks/src/test/java/org/sonar/java/checks/unused/UnusedPrivateFieldCheckTest.java
```diff
@@ -32,6 +32,15 @@ void test() {
       .verifyIssues();
   }
 
+  @Test
+  void test_without_semantic() {
```
leonardo-pilastri-sonarsource, line 36: I would just add a comment around this test as it is not usual to expect issues without semantics, and in the future it might be useful to have a reference here

leonardo-pilastri-sonarsource (changes requested):
LGTM, let's just add a comment around the new test

java-checks/src/test/java/org/sonar/java/checks/unused/UnusedPrivateFieldCheckTest.java
tomasz-tylenda-sonarsource: Done. You can check if this is what you meant.

leonardo-pilastri-sonarsource (approved):

PR #5035, opened 2025-02-17
SONARJAVA-4972 S1118 Fix FPs on Lombok annotations in automatic analysis

tomasz-tylenda-sonarsource:
[SONARJAVA-4972](https://sonarsource.atlassian.net/browse/SONARJAVA-4972)

[SONARJAVA-4972]: https://sonarsource.atlassian.net/browse/SONARJAVA-4972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/UtilityClassWithPublicConstructorCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/UtilityClassWithPublicConstructorCheckSample.java
@@ -48,7 +48,7 @@ class UtilityClassWithPublicConstructorCheckSample {
   }
 
   @AllArgsConstructor(access = PRIVATE)
-  class LombokClassAllArgs { // Compliant, a private constructor will be generated
+  class LombokClassAllArgsPrivate { // Compliant, a private constructor will be generated
     public static void foo() {
     }
   }


java-checks-test-sources/default/src/main/java/checks/UtilityClassWithPublicConstructorCheckSample.java
dorian-burihabwa-sonarsource, line 1: I see that we test on `AccessLevel.PUBLIC` but do we test on `AccessLevel.NONE`

dorian-burihabwa-sonarsource: I see that the code explicitly tests for `PUBLIC` so I would be curious to see what happens with `NONE`

java-checks/src/main/java/org/sonar/java/checks/UtilityClassWithPublicConstructorCheck.java
```diff
@@ -84,11 +95,17 @@ private static boolean hasPublicModifier(MethodTree methodTree) {
...
-    return classTree.modifiers().annotations().stream().anyMatch(it -> isLombokConstructorGenerator(it) && !hasPublicAccess(it));
+    return classTree.modifiers().annotations().stream()
+      .anyMatch(it -> isLombokConstructorGenerator(it.symbolType()) && !hasPublicAccess(it));
   }
 
-  private static boolean isLombokConstructorGenerator(AnnotationTree annotation) {
-    return LOMBOK_CONSTRUCTOR_GENERATORS.contains(annotation.annotationType().symbolType().fullyQualifiedName());
+  private static boolean isLombokConstructorGenerator(Type symbolType) {
+    // This happens in automatic analysis. We match only the last part of the name.
+    if(symbolType.isUnknown()) {
```
dorian-burihabwa-sonarsource, line 104: Formatting

dorian-burihabwa-sonarsource (changes requested):
The change looks pretty good already but I see that the ticket mentions access level `NONE`. Whether we decide to consider it as TP or FP, we should probably add an example in the tests

java-checks/src/main/java/org/sonar/java/checks/UtilityClassWithPublicConstructorCheck.java
tomasz-tylenda-sonarsource: Done, added space after `if`.

java-checks-test-sources/default/src/main/java/checks/UtilityClassWithPublicConstructorCheckSample.java
tomasz-tylenda-sonarsource: Good point. I added a test with `@NoArgsConstructor(access = AccessLevel.NONE)`. I also renamed test classes to make them easier to manage.

dorian-burihabwa-sonarsource (approved):

PR #5036, opened 2025-02-18
SONARJAVA-5232 S1192 Relax duplicate string check for Throwable arguments

tomasz-tylenda-sonarsource:
[SONARJAVA-5232](https://sonarsource.atlassian.net/browse/SONARJAVA-5232)

Previously  #4984 and #5006

[SONARJAVA-5232]: https://sonarsource.atlassian.net/browse/SONARJAVA-5232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
@@ -113,3 +113,110 @@ class Coverage {
   private final String prevLeftNull = "SELECT" + 3;
   private final String prevRightNull = 3 + "SELECT";
 }
+
+class DuplicatedExceptionArguments {
+  private void areCompliantByDefault(int r) {
+    if (r == 0) {
+      throw new IllegalArgumentException("simple IAE");
+    } else if (r == 1) {
+      throw new IllegalArgumentException("simple IAE");
+    } else if (r == 2) {
+      throw new IllegalArgumentException("simple IAE");
+    }
+  }
+
+  static class MyException extends Exception {
+    public MyException(String message, Throwable throwable) {
+      super(message, throwable);
+    }
+  }
+
+  private void twoArgs(int r, Throwable throwable) throws MyException {
+    if (r == 0) {
+      throw new MyException("my exception message", throwable);
+    } else if (r == 1) {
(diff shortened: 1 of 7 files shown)


java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
dorian-burihabwa-sonarsource, line 1: Would it make sense to add, but not cover, the case where the exception is declared in one statement and thrown in another?
```java
  public static void constructAndThenThrow(int value) {
    if (value == 0) {
      var first = new IllegalArgumentException("this is a repeated message"); // Noncompliant
      throw first;
    } else if (value < 0) {
      var second = new IllegalArgumentException("this is a repeated message");
      throw second;
    } else {
      var third = new IllegalArgumentException("this is a repeated message");
      throw third;
    }
  }
```

dorian-burihabwa-sonarsource (changes requested):
This looks good overall but before merging, it would make sense to:
- Document the last suggested case with exceptions constructed and then thrown
- Squash the PR into a single commit
- Rewrite the commit message to be very explicit in the fact that this change is a relaxation of the existing rule

java-checks-test-sources/default/src/main/java/checks/StringLiteralDuplicatedCheckSample.java
tomasz-tylenda-sonarsource: Done.

tomasz-tylenda-sonarsource:
> This looks good overall but before merging, it would make sense to:
> 
> * Document the last suggested case with exceptions constructed and then thrown
> * Squash the PR into a single commit
> * Rewrite the commit message to be very explicit in the fact that this change is a relaxation of the existing rule

Done.

tomasz-tylenda-sonarsource:
> This PR is stale because it has been open 7 days with no activity. If there is no activity in the next 7 days it will be closed automatically

The PR will be merged after we review dogfood.

dorian-burihabwa-sonarsource (approved):

erwan-serandour (approved):

PR #5041, opened 2025-02-26
SONARJAVA-5341 Fix FP reported for too many cases on switches when enum types are unknown

romainbrenguier:
[SONARJAVA-5341](https://sonarsource.atlassian.net/browse/SONARJAVA-5341)

When we don't know if the symbol in the switch is an enum, we shouldn't raise that issue.

[SONARJAVA-5341]: https://sonarsource.atlassian.net/browse/SONARJAVA-5341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/helpers/TernaryValue.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/helpers/TernaryValue.java
@@ -1,0 +1,36 @@
+/*
+ * SonarQube Java
+ * Copyright (C) 2012-2025 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the Sonar Source-Available License Version 1, as published by SonarSource SA.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/helpers/TernaryValue.java
```diff
@@ -0,0 +1,35 @@
...
+public enum TernaryValue {
+  TRUE, FALSE, UNKNOWN;
+
+  public static TernaryValue ofBoolean(@Nullable Boolean b) {
+    if (b == null) {
+      return UNKNOWN;
+    }
+    return b ? TRUE : FALSE;
+  }
+
```
alban-auzeill, line 31: I'm not sure about the usefulness of `ofBoolean` compared to `of`, I don't see where there could be ambiguity.
When we have a boolean like `isEnum()`, I would prefer having a dedicated method instead of converting it to `Boolean` in a method that converts it back to `boolean`.
```suggestion
  public static TernaryValue of(boolean b) {
    return b ? TRUE : FALSE;
  }

  public static TernaryValue of(@Nullable Boolean b) {
    return b == null ? UNKNOWN : of(b.booleanValue());
  }
```

alban-auzeill (changes requested):

alban-auzeill (approved):

PR #5046, opened 2025-02-28
A “NullPointerException” could be thrown; “ofNullable()” can return null

reda-alaoui:
Reproducer + hacky fix for https://community.sonarsource.com/t/a-nullpointerexception-could-be-thrown-ofnullable-can-return-null/134892 .

The hacky fix breaks `org.sonar.java.checks.ChangeMethodContractCheckTest#test`.

IMHO, nullable annotations on method generic return parameters are structurally wrongly implemented. A nullable annotation on the return type or a generic element of the return type is viewed as the same by analyzer.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/model/JSymbol.java
+++ b/java-frontend/src/main/java/org/sonar/java/model/JSymbol.java
@@ -377,7 +377,7 @@ abstract class JSymbol implements Symbol {
         if (returnType == null) {
           return Symbols.EMPTY_METADATA;
         }
-        return new JSymbolMetadata(sema, this, getAnnotations(returnType), binding.getAnnotations());
+        return new JSymbolMetadata(sema, this, returnType.getTypeAnnotations(), binding.getAnnotations());
       default:
         return new JSymbolMetadata(sema, this, binding.getAnnotations());
     }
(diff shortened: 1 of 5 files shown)


romainbrenguier:
Thanks for the PR. There is indeed an issue with `JSymbol#convertMetadata`. We are going to look for a proper fix.

reda-alaoui:
Hello @romainbrenguier,
Was the fix released? If it was, in which Sonarqube version?

romainbrenguier:
@reda-alaoui 
> Hello @romainbrenguier, Was the fix released? If it was, in which Sonarqube version?

Hello, sorry I lost track of this. There has been several changes related to this issue, but I'm not sure whether it was fixed for your case or not. In particular, the NullPointerException rule was migrated to a new engine: https://community.sonarsource.com/t/new-version-of-our-bug-detection-engine-for-java-and-python/143287, I hope that helps.

reda-alaoui:
@romainbrenguier Even with v25.9.0.112764, the issue still exists.

This is a deal breaker. This should be the 101 of a static analysis tool: protect us from null pointer exception. This issue has been there for months.
I bet you are losing customers (and potential ones) because of this. Maybe without you knowing it.

rupert-jung-mw:
This is just laughable.

PR #5052, opened 2025-03-07
SONARJAVA-5383 DependencyVersionAware interface

romainbrenguier:
[SONARJAVA-5383](https://sonarsource.atlassian.net/browse/SONARJAVA-5383)

[SONARJAVA-5383]: https://sonarsource.atlassian.net/browse/SONARJAVA-5383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


leonardo-pilastri-sonarsource (changes requested):
Just a couple of minor things, for the lombok rule, it might be better to improve it if we want to have a round of tests on peach for instance.

I am still unsure if the current implementation of the inference service is overly complex, but that we can discuss it in a broader scope also with people from the squad

PR #5053, opened 2025-03-11
SONARJAVA-5383 Lookup Spring-web dependency version to filter out visitors

romainbrenguier:
[SONARJAVA-5383](https://sonarsource.atlassian.net/browse/SONARJAVA-5383)

Minimal example of what can be achieved by looking at dependency versions.
Improves the SpringComposedRequestMappingCheck

[SONARJAVA-5383]: https://sonarsource.atlassian.net/browse/SONARJAVA-5383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


its/autoscan/src/test/java/org/sonar/java/it/AutoScanTest.java
```diff
@@ -120,7 +120,7 @@ public void javaCheckTestSources() throws Exception {
       .setProjectName(PROJECT_NAME)
       .setProjectVersion("0.1.0-SNAPSHOT")
       .setSourceEncoding("UTF-8")
-      .setSourceDirs("aws/src/main/java/,default/src/main/java/,java-17/src/main/java/,spring-3.2/src/main/java/")
+      .setSourceDirs("aws/src/main/java/,default/src/main/java/,java-17/src/main/java/,spring-3.2/src/main/java/,spring-web-4.0/src/main/java/")
```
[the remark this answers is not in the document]
leonardo-pilastri-sonarsource: I think it's just at line 89 and 121

leonardo-pilastri-sonarsource (approved):
LGTM

java-frontend/src/main/java/org/sonar/java/classpath/Version.java
```diff
@@ -0,0 +1,125 @@
+/*
```
leonardo-pilastri-sonarsource, line 1: We have an issue with this class, as it's not part of the public API but the `DependencyVersionAware` depends on it, it will be missing in the classpath for the analysis at runtime if a custom plugin tries to implement the interface.
I think we should also move the `Version` object to the public API.

leonardo-pilastri-sonarsource (changes requested):
We need to move 1 file, and we should also document the new public APIs in the "API changes" section of the `sonar-java-plugin/src/main/resources/static/documentation.md` file

java-frontend/src/main/java/org/sonar/plugins/java/api/DependencyVersionAware.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.plugins.java.api;
+
+import java.util.Optional;
+import java.util.function.Function;
+
+public interface DependencyVersionAware {
```
leonardo-pilastri-sonarsource, line 22: We need to document this API as well

java-frontend/src/main/java/org/sonar/plugins/java/api/DependencyVersionAware.java
```diff
@@ -0,0 +1,26 @@
...
+
+  boolean isCompatibleWithDependencies(Function<String, Optional<Version>> dependencyFinder);
```
leonardo-pilastri-sonarsource, line 24: Since the usage of this method is not super straight-forward I would document it with a couple of example on how to implement it.

leonardo-pilastri-sonarsource (changes requested):
If you also tested that a custom plugin can implement this new interface, besides a couple more javadoc I think it looks good

romainbrenguier:
To test that a custom plugin could use the new part of the API, here is the process I followed:
 
* In the java-custom-rules-example module I modified the pom_SQ_10_6_LATEST.xml file to use sonarjava version 8.13.0-SNAPSHOT
* I modified the AvoidAnnotationRule to implement the DependencyVersionAware interface, so that it only applies when spring-web is found with version at least 4.0.0
* I install sonar-java-plugin and the custom-rules-example
* I moved the two jar files in my local sonarqube installation, in lib/extensions and extensions/plugins respectively
* I started sonarqube and created a sample project, with a profile that included the modified custom rule
* I ran the analysis on a sample project contaning a module with spring-web
* I checked that the custom rule reported an issue in the analysis
* As an additional check, I ran the maven scanner in debug mode and put a break point in the added code of the modified custom rule to check it was behaving as expected

leonardo-pilastri-sonarsource (approved):
LGTM now! Thanks for the nice work!

PR #5069, opened 2025-03-31
SONARJAVA-5403 Rule S7435: Processing persistent unique identifiers is security-sensitive

antonioaversa:
[SONARJAVA-5403](https://sonarsource.atlassian.net/browse/SONARJAVA-5403)

RSPEC PR [MERGED]: https://github.com/SonarSource/rspec/pull/4776
RSPEC PR to update quickfix [TO REVIEW]: https://github.com/SonarSource/rspec/pull/4864
RSPEC PR to fix missing semicolon in code example [TO REVIEW]: https://github.com/SonarSource/rspec/pull/4878
RSPEC Page: https://sonarsource.github.io/rspec/#/rspec/S7435/java
Specification JIRA: https://sonarsource.atlassian.net/browse/SONARKT-578
Kotlin Implementation JIRA: https://sonarsource.atlassian.net/browse/SONARKT-626

[SONARJAVA-5403]: https://sonarsource.atlassian.net/browse/SONARJAVA-5403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @Override
+  protected MethodMatchers getMethodInvocationMatchers() {
+    return MethodMatchers.or(matchers.keySet().toArray(MethodMatchers[]::new));
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    matchers.keySet().stream()
```
antonioaversa, line 102: [Same remark](https://github.com/SonarSource/sonar-kotlin/pull/600#discussion_r2012490830) here as for the corresponding PR in `sonar-kotlin`.
Using a map is OK here because matchers don't intersect, so there is no risk of matching a different matcher the day Map will change implementation and return a different order of keys.

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
```diff
@@ -0,0 +1,125 @@
...
+      .filter(matcher -> matcher.matches(mit))
+      .findFirst()
+      .filter(matcher ->
+        matcher != staticSettingsSecureGetStringFunMatcher ||
+          mit.arguments().get(1).asConstant().map("android_id"::equals).orElse(false))
+      .ifPresent(matcher -> {
+        var tree = mit.methodSelect().is(Tree.Kind.MEMBER_SELECT)
```
antonioaversa, line 109: I couldn't cover the else branch of this ternary here, but I didn't want to just cast to `MemberSelectExpressionTree ` either. If you are confident that the cast will always succeed for a method invocation, I can remove the ternary. Otherwise I would keep it: it seems something that can easily change in the future, and being a bit defensive doesn't hurt.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S7435.json
```diff
@@ -0,0 +1,34 @@
...
+      "8.3.3"
+    ],
+    "OWASP Mobile Top 10 2024": [
+      "M6"
+    ],
+    "CWE": [
+      213
+    ]
+  },
+  "quickfix": "infeasible",
```
antonioaversa, line 27: I have set the quickfix to infeasible because, while we could just remove the call, the resulting code would very likely be non-compiling and possibly functionally wrong: the user would be using these identifiers to build composite identifiers, intents extra, or the like.
It is not possible to determine what is the right replacement for the telephony/subscription/ads identifier used - it depends on the specific scenario.

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
dorian-burihabwa-sonarsource: I think we can keep the uncovered branch because there is one (very very unlikely) case that can make the code crash without it.

If sonar-java analyzes the code of the file where the type and method we are trying to match is both defined and invoked/referenced, then we cannot rely on the method select being available.

For instance you can make the analyzer crash by adding the following segment to the body of `BluetoothAdapter` in `android/bluetooth/BluetoothAdapter.java` and removing the dead branch in the check
```java
  public void crashZone() {
    getAddress();
  }
```

java-checks/src/test/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheckTest.java
```diff
@@ -0,0 +1,34 @@
...
+class AndroidPersistentUniqueIdentifierCheckTest {
+
+  @Test
+  void test() {
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/security/AndroidPersistentUniqueIdentifierCheckSample.java"))
+      .withCheck(new AndroidPersistentUniqueIdentifierCheck())
+      .verifyIssues();
+  }
+
```
dorian-burihabwa-sonarsource, line 33: We could probably add a test here without semantic to detect regression or changes in the way the check operates
```suggestion
@Test
  void without_semantic() {
    CheckVerifier.newVerifier()
      .onFile(mainCodeSourcesPath("checks/security/AndroidPersistentUniqueIdentifierCheckSample.java"))
      .withCheck(new AndroidPersistentUniqueIdentifierCheck())
      .withoutSemantic()
      .verifyNoIssues();
  }
```

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
```diff
@@ -0,0 +1,125 @@
...
+  protected MethodMatchers getMethodInvocationMatchers() {
+    return MethodMatchers.or(matchers.keySet().toArray(MethodMatchers[]::new));
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    matchers.keySet().stream()
+      .filter(matcher -> matcher.matches(mit))
+      .findFirst()
+      .filter(matcher ->
```
dorian-burihabwa-sonarsource, line 105: Is there a way we combine the two filter methods together by extracting things to a static boolean method?

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
```diff
@@ -0,0 +1,125 @@
...
+        reportIssue(tree, message);
+      });
+  }
+
+  @Override
+  protected void onMethodReferenceFound(MethodReferenceTree methodReferenceTree) {
+    matchers.keySet().stream()
+      .filter(matcher -> matcher.matches(methodReferenceTree))
+      .findFirst()
+      .filter(Predicate.not(staticSettingsSecureGetStringFunMatcher::equals))
```
dorian-burihabwa-sonarsource, line 122: Similarly to the method above, is there a way we can combine the two filters in a single static boolean method?

dorian-burihabwa-sonarsource (changes requested):
The implementation looks good to me and I think we can improve readability a little by making the stream operations easier to read by extracting and grouping some of the filtering logic.

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
```diff
@@ -0,0 +1,125 @@
...
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    matchers.keySet().stream()
+      .filter(matcher -> matcher.matches(mit))
+      .findFirst()
+      .filter(matcher ->
+        matcher != staticSettingsSecureGetStringFunMatcher ||
+          mit.arguments().get(1).asConstant().map("android_id"::equals).orElse(false))
+      .ifPresent(matcher -> {
+        var tree = mit.methodSelect().is(Tree.Kind.MEMBER_SELECT)
+          ? ((MemberSelectExpressionTree) mit.methodSelect()).identifier()
```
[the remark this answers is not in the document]
antonioaversa: Nice!

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    matchers.keySet().stream()
+      .filter(matcher -> matcher.matches(mit))
+      .findFirst()
+      .filter(matcher ->
+        matcher != staticSettingsSecureGetStringFunMatcher ||
+          mit.arguments().get(1).asConstant().map("android_id"::equals).orElse(false))
```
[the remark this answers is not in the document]
antonioaversa: Indeed! I have extracted the entire boolean expression into a `isSettingSecureGetStringWithAndroidId` and inverted the logic.

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
antonioaversa: Oh ok, I see! Ok, keeping the check then.

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
antonioaversa: Yes, I could define an `isInvocationCandidate` method (for lack of a better name):

```java
  private static boolean isInvocationCandidate(MethodMatchers matcher, MethodInvocationTree mit) {
    return matcher.matches(mit) &&
      (matcher != STATIC_SETTINGS_SECURE_GET_STRING_MATCHER || hasAndroidIdArgument(mit));
  }
```

and use it like so:

```java
  @Override
    MATCHERS.keySet().stream()
      .filter(matcher -> isInvocationCandidate(matcher, mit))
      .findFirst()
      .ifPresent(matcher -> {
        var tree = mit.methodSelect() instanceof MemberSelectExpressionTree mset ? mset.identifier() : mit.methodSelect();
        var message = MATCHERS.get(matcher);
        reportIssue(tree, message);
      });
  }
```

I made the change, but I feel this is hiding the filtering logic from the `onMethodInvocationFound`, which was simple enough to not be needing externalized filtering logic.

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
antonioaversa: Yes, as above, I could coalesce the two filters into a single `isReferenceCandidate` method.

antonioaversa:
@dorian-burihabwa-sonarsource I think I have addressed all the comments. Back to you for a second round.

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
```diff
@@ -0,0 +1,131 @@
...
+  );
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.METHOD_INVOCATION, Tree.Kind.METHOD_REFERENCE);
+  }
+
+  @Override
+  protected MethodMatchers getMethodInvocationMatchers() {
+    return MethodMatchers.or(MATCHERS.keySet().toArray(MethodMatchers[]::new));
```
leonardo-pilastri-sonarsource, line 96: The `getMethodInvocationMatchers` is called by the superclass for every node we visit; I would store the matcher we are creating here somewhere else instead of instantiating it every time, to save up on performance

leonardo-pilastri-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/security/AndroidPersistentUniqueIdentifierCheck.java
antonioaversa: Fair point! I have extracted the matcher construction into a static field.

leonardo-pilastri-sonarsource (approved):
Thanks for the change, looks good to me now!

dorian-burihabwa-sonarsource (approved):

PR #5070, opened 2025-04-02
SONARJAVA-5420 Improve S5344: Passwords should not be stored in plaintext or with a fast hashing algorithm

antonioaversa:
[SONARJAVA-5420](https://sonarsource.atlassian.net/browse/SONARJAVA-5420)

RSPEC PR [TO MERGE]: https://github.com/SonarSource/rspec/pull/4822
RSPEC PR to update the quickfix [TO REVIEW]: https://github.com/SonarSource/rspec/pull/4866
RSPEC Page: https://sonarsource.github.io/rspec/#/rspec/S5344/java
Kotlin Implementation Jira: https://sonarsource.atlassian.net/browse/SONARKT-639
Kotlin Implementation PR: https://github.com/SonarSource/sonar-kotlin/pull/601  

[SONARJAVA-5420]: https://sonarsource.atlassian.net/browse/SONARJAVA-5420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


java-checks-test-sources/default/src/main/java/checks/security/PasswordEncoderJavaxCryptoSample.java
```diff
@@ -0,0 +1,244 @@
...
+  // endregion
+
+  // region Compliant cases
+
+  public void compliantIntLiteralAboveThresholdForSHA512(String password, byte[] salt) throws Exception {
+    PBEKeySpec keySpec = new PBEKeySpec(password.toCharArray(), salt, 210000, 256); // Compliant: 210_000 >= 210_000
+    SecretKeyFactory secretKeyFactory = SecretKeyFactory.getInstance("PBKDF2withHmacSHA512");
+    secretKeyFactory.generateSecret(keySpec);
+  }
+
```
erwan-serandour, line 183: let's add a new case

  public void compliantIntLiteralAboveThresholdForSHA512(String password, byte[] salt) throws Exception {
    int it = 10;
    it = 210000;
    PBEKeySpec keySpec = new PBEKeySpec(password.toCharArray(), salt, it, 256);
    SecretKeyFactory secretKeyFactory = SecretKeyFactory.getInstance("PBKDF2withHmacSHA512");
    secretKeyFactory.generateSecret(keySpec);
  }

java-checks/src/main/java/org/sonar/java/checks/security/PasswordEncoderCheck.java
```diff
@@ -87,6 +129,92 @@ public void visitNode(Tree tree) {
...
+    var initializerExpression = getInitializerExpressionOfVariableIdentifier(expression);
+    return initializerExpression
+      .flatMap(e -> e.asConstant(type))
+      .map(v -> new ExpressionsAndValue<>(expression, initializerExpression.get(), v))
+      .or(() -> expression.asConstant(type).map(t -> new ExpressionsAndValue<>(expression, expression, t)));
+  }
+
+  // Given any identifier "x", returns the initializer y from the declaration of x: "var x = y" (whether in the same
+  // statement or not, whether it was reassigned between declaration and reference or not).
+  private static Optional<ExpressionTree> getInitializerExpressionOfVariableIdentifier(ExpressionTree expression) {
```
erwan-serandour, line 207: I think we can update a bit the name to make it clearer. The return type already gives us information about the expression and the if make clearer when we return a value.
```suggestion
  private static Optional<ExpressionTree> getInitializerIfExpressionIsVariableIdentifier(ExpressionTree expression) {
```

java-checks/src/main/java/org/sonar/java/checks/security/PasswordEncoderCheck.java
```diff
@@ -87,6 +129,92 @@ public void visitNode(Tree tree) {
...
+      .filter(e -> e.is(Tree.Kind.METHOD_INVOCATION))
+      .map(MethodInvocationTree.class::cast)
+      .filter(SECRET_KEY_FACTORY_GET_INSTANCE_METHOD::matches)
+      .map(gi -> gi.arguments().get(0))
+      .flatMap(e -> getInitializerExpressionAndValue(e, String.class));
+  }
+
+  // Given any expression e, attempts constant value resolution of the expression itself and of the initializer
+  // of the declaration of e, if any.
+  private static <T> Optional<ExpressionsAndValue<T>> getInitializerExpressionAndValue(ExpressionTree expression, Class<T> type) {
```
erwan-serandour, line 197: we can also make the name a bit clearer
```suggestion
  private static <T> Optional<ExpressionsAndValue<T>> getInitLiteralIfExpressionIsVariable(ExpressionTree expression, Class<T> type) {
```

erwan-serandour: maybe Init can be removed.

erwan-serandour (changes requested):
Small suggestion about the names of the methods and I propose to add a new sample cases. The new test sample seems to be an FP, I am not sure if we want to handle it.

Otherwise, thanks a lot to have added the link to rspec in the PR :).

java-checks-test-sources/default/src/main/java/checks/security/PasswordEncoderJavaxCryptoSample.java
antonioaversa: Good catch! This one is actually an FP, which is good to document. 
For info, we discussed a similar scenario [here](https://github.com/SonarSource/sonar-kotlin/pull/598/files#r2007963988), where I filtered only for `val` (`final` in Java). 
However the use of local `final` in Java is way less common than the use of `val` in Kotlin, and we are thinking of removing the `?.takeIf { !it.isVar }` to have a higher reach, since:
- the chance of hardcoding the algorithm and then reassigning the iteration are very low
- these security rules have generally very low reach, as they signal potentially serious security issues
- this rule in particular is a vulnerability, so it's better to slightly err on the side of a very low probability FP than having a FN for the very common scenario where a non-`final` variable is used
- dead store would trigger if the initially assigned value (10) is never read and, if it is, there is at least one path where 10 is used - so it's worth triggering.

java-checks/src/main/java/org/sonar/java/checks/security/PasswordEncoderCheck.java
antonioaversa: Sure, it definitely reads better!

erwan-serandour (approved):
LGTM

java-checks/src/main/java/org/sonar/java/checks/security/PasswordEncoderCheck.java
antonioaversa: Discussed offline: we went for `getValueIfExpressionIsVariableIdentifier`, since what basically this method does on top of `getInitializerIfExpressionIsVariableIdentifier` is to calculate its value, using `asConstant`.

PR #5083, opened 2025-04-11
SONARCH-708, SONARJAVA-5497: Extend SonarJava CheckRegistrar API for registering custom file scanner hooks

kaufco:
[SONARCH-708](https://sonarsource.atlassian.net/browse/SONARCH-708)

[SONARCH-708]: https://sonarsource.atlassian.net/browse/SONARCH-708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/TestCheckRegistrarContext.java
+++ b/java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/TestCheckRegistrarContext.java
@@ -24,9 +24,11 @@ import java.util.List;
 import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
 import org.sonar.api.rule.RuleKey;
+import org.sonar.api.rule.RuleScope;
 import org.sonar.api.rules.RuleAnnotationUtils;
 import org.sonar.plugins.java.api.CheckRegistrar;
 import org.sonar.plugins.java.api.JavaCheck;
+import org.sonar.plugins.java.api.JavaFileScanner;
 
 public class TestCheckRegistrarContext extends CheckRegistrar.RegistrarContext {
 
@@ -52,16 +54,34 @@ public class TestCheckRegistrarContext extends CheckRegistrar.RegistrarContext {
 
   @Override
   public void registerMainSharedCheck(JavaCheck check, Collection<RuleKey> ruleKeys) {
+    registerMainHook(check);
+    mainRuleKeys.addAll(ruleKeys);
+  }
+
+  private void registerMainHook(JavaCheck check) {
     mainCheckClasses.add(check.getClass());
     mainCheckInstances.add(check);
-    mainRuleKeys.addAll(ruleKeys);
   }
 
   @Override
   public void registerTestSharedCheck(JavaCheck check, Collection<RuleKey> ruleKeys) {
+    registerTestHook(check);
+    testRuleKeys.addAll(ruleKeys);
+  }
+
+  private void registerTestHook(JavaCheck check) {
     testCheckClasses.add(check.getClass());
     testCheckInstances.add(check);
-    testRuleKeys.addAll(ruleKeys);
+  }
+
+  @Override
+  public void registerCustomFileScanner(RuleScope ruleScope, JavaFileScanner scanner) {
+    if (ruleScope != RuleScope.TEST) {
+      registerMainHook(scanner);
+    }
+    if (ruleScope != RuleScope.MAIN) {
+      registerTestHook(scanner);
+    }
   }
 
   @Override
--- a/java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/TestCheckRegistrarContext.java
+++ b/java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/TestCheckRegistrarContext.java
@@ -76,10 +76,10 @@ public class TestCheckRegistrarContext extends CheckRegistrar.RegistrarContext {
 
   @Override
   public void registerCustomFileScanner(RuleScope ruleScope, JavaFileScanner scanner) {
-    if (ruleScope != RuleScope.TEST) {
+    if (ruleScope == RuleScope.MAIN || ruleScope == RuleScope.ALL) {
       registerMainHook(scanner);
     }
-    if (ruleScope != RuleScope.MAIN) {
+    if (ruleScope == RuleScope.TEST || ruleScope == RuleScope.ALL) {
       registerTestHook(scanner);
     }
   }
--- a/java-frontend/src/main/java/org/sonar/java/SonarComponents.java
+++ b/java-frontend/src/main/java/org/sonar/java/SonarComponents.java
@@ -309,10 +309,10 @@ public class SonarComponents extends CheckRegistrar.RegistrarContext {
 
   @Override
   public void registerCustomFileScanner(RuleScope ruleScope, JavaFileScanner scanner) {
-    if (ruleScope != RuleScope.TEST) {
+    if (ruleScope == RuleScope.MAIN || ruleScope == RuleScope.ALL) {
       mainChecks.add(scanner);
     }
-    if (ruleScope != RuleScope.MAIN) {
+    if (ruleScope == RuleScope.TEST || ruleScope == RuleScope.ALL) {
       testChecks.add(scanner);
     }
   }
--- a/java-frontend/src/main/java/org/sonar/java/SonarComponents.java
+++ b/java-frontend/src/main/java/org/sonar/java/SonarComponents.java
@@ -56,6 +56,7 @@ import org.sonar.api.config.Configuration;
 import org.sonar.api.measures.FileLinesContext;
 import org.sonar.api.measures.FileLinesContextFactory;
 import org.sonar.api.rule.RuleKey;
+import org.sonar.api.rule.RuleScope;
 import org.sonar.api.utils.Version;
 import org.sonar.java.annotations.VisibleForTesting;
 import org.sonar.java.caching.ContentHashCache;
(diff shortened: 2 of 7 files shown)


java-frontend/src/test/java/org/sonar/plugins/java/api/CheckRegistrarTest.java
```diff
@@ -96,6 +97,12 @@ class Scanner implements JavaCheck {
       "register 3 instantiated test checks.");
   }
 
+  @Test
+  void empty_default_method_coverage() {
+    var context = new CheckRegistrar.RegistrarContext();
+    assertDoesNotThrow(() -> context.registerCustomFileScanner(null, null));
```
dorian-burihabwa-sonarsource, line 103: What are we testing here? Isn't the default implementation of `registerCustomFileScanner` empty?

java-frontend/src/test/java/org/sonar/plugins/java/api/CheckRegistrarTest.java
```diff
@@ -96,6 +97,12 @@ class Scanner implements JavaCheck {
       "register 3 instantiated test checks.");
   }
 
+  @Test
+  void empty_default_method_coverage() {
```
dorian-burihabwa-sonarsource, line 101: To help understanding the intent of the test, it would help if the test a name in the form of an assertion (that should not be broken)

java-frontend/src/main/java/org/sonar/plugins/java/api/CheckRegistrar.java
```diff
@@ -171,6 +172,13 @@ public void registerMainSharedCheck(JavaCheck check, Collection<RuleKey> ruleKey
       // to be overridden
     }
 
+    /**
+     * Registers a custom file scanner not related to any rule or repository.
+     */
+    public void registerCustomFileScanner(RuleScope ruleScope, JavaFileScanner scanner) {
```
dorian-burihabwa-sonarsource, line 178: Nitpicking: Have we considered returning some value here? To test that something was added, we need to look into the internals of the `Checkregistrar` implementation. It works here but does not promote good testing practices in the custom plugins downstream.

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -305,6 +307,16 @@ public void registerTestSharedCheck(JavaCheck check, Collection<RuleKey> ruleKey
...
   }
 
+  @Override
+  public void registerCustomFileScanner(RuleScope ruleScope, JavaFileScanner scanner) {
+    if (ruleScope != RuleScope.TEST) {
+      mainChecks.add(scanner);
+    }
+    if (ruleScope != RuleScope.MAIN) {
+      testChecks.add(scanner);
+    }
```
dorian-burihabwa-sonarsource, line 317: Can we arrange these tests in a less clever but more explicit way?
```suggestion
    if (ruleScope == RuleScope.MAIN || ruleScope ==  RuleScope.ALL) {
      mainChecks.add(scanner);
    }
    if (ruleScope == RuleScope.TEST || ruleScope ==  RuleScope.ALL) {
      testChecks.add(scanner);
    }
```

java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/TestCheckRegistrarContext.java
```diff
@@ -52,16 +54,34 @@ public void registerTestChecks(String repositoryKey, Collection<?> javaCheckClas
...
+  }
+
+  @Override
+  public void registerCustomFileScanner(RuleScope ruleScope, JavaFileScanner scanner) {
+    if (ruleScope != RuleScope.TEST) {
+      registerMainHook(scanner);
+    }
+    if (ruleScope != RuleScope.MAIN) {
+      registerTestHook(scanner);
+    }
```
dorian-burihabwa-sonarsource, line 84: If we change the implemetation in `SonarComponents`, we might as well change it here

dorian-burihabwa-sonarsource (changes requested):
I think the change looks simple enough to integrated as-is, but there are a couple of questions I have left for you to look at.

In addition,  I thin that he ticket attached to this PR should be a SONARJAVA ticket because this change implies a modification of the public API and should be properly documented as part of the release

java-frontend/src/test/java/org/sonar/plugins/java/api/CheckRegistrarTest.java
kaufco: Yes, but QG failed due to low coverage. Even if the method is empty, it must still be tested. So just making sure the method is there and can be called without issues is the best thing we can do here I think.

java-frontend/src/test/java/org/sonar/plugins/java/api/CheckRegistrarTest.java
kaufco: The purpose of the method is just to get coverage for an empty default method (see the other comment). Naming the method `registerCustomFileScanner_does_not_throw_exception` would satisfy your request, but still I think it doesn't capture well the purpose of the method. I renamed it to `cover_empty_default_implementations`. Wdyt?

java-frontend/src/main/java/org/sonar/plugins/java/api/CheckRegistrar.java
kaufco: I think we should stick to the pattern used in `CheckRegistrar` and not doing it different for a single method: all the `register*` methods have a `void` return value.

If we want to change that, I think it should be done in a separate ticket, because changing it for all methods would clearly go beyond the scope of this PR (and also break backwards compatibility on the bytecode level, because even though source code level compatibility does remain, the method signatures would change in the class files).

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
kaufco: Fixed.

java-frontend/src/test/java/org/sonar/java/SonarComponentsTest.java
```diff
@@ -471,6 +512,21 @@ void no_issue_when_check_not_found() {
     verify(context, never()).newIssue();
   }
 
+  @Test
+  void no_issue_when_reporting_from_custom_file_scanner() {
```
leonardo-pilastri-sonarsource, line 516: Sorry to have requested this test, as it does not really bring much value, I would rather have a test checking that scanners registered with the new API are actually invoked.
We could have a simple 
```java
    class CustomJFS implements JavaFileScanner {
      int counter = 0;

      @Override
        counter++;
      }
    }
```
    
added through 
    
```java
    CheckRegistrar registrar = registrarContext ->
      registrarContext.registerCustomFileScanner(RuleScope.ALL, customScanner);
```
And after running a scan on a file, check that the counter actually increased

leonardo-pilastri-sonarsource (changes requested):
LGTM, let's just change the test in SonarComponents to an end-to-end check that the scanner is actually invoked when registered through the new API

java-frontend/src/test/java/org/sonar/java/SonarComponentsTest.java
kaufco: Would this actually make sense? To my understanding: If I write a unit test for it like for the regular checks, it'll use `TestCheckRegistrarcontext` instead of `SonarComponents`. So we'd end up testing the test infrastructure, not the API implementation in `SonarComponents`. And `SonarComponentsTest` doesn't run any rule checks at all. It just checks if they are registered in the list, and for the new API, I have added a test for that as well.

leonardo-pilastri-sonarsource: What I would like to have is an end-to-end test to check that the JavaFileScanners registered through this new API are indeed invoked during the analysis, and their `scan` method is executed. In the end that is the purpose of the PR so I think it could be a nice test to have

kaufco: Done.

leonardo-pilastri-sonarsource (approved):

java-frontend/src/main/java/org/sonar/plugins/java/api/CheckRegistrar.java
```diff
@@ -171,6 +172,16 @@ public void registerMainSharedCheck(JavaCheck check, Collection<RuleKey> ruleKey
       // to be overridden
     }
 
+    /**
+     * Registers a custom file scanner not related to any rule or repository.
+     * CheckRegistrars call this function to register a custom file scanner for execution during the analysis
+     * on all source files that match the given rule scope (MAIN, TEST or ALL).
+     * Custom file scanners reporting an issue will have no effect, since no rule is associated.
```
dorian-burihabwa-sonarsource, line 179: `will have no effect`: meaning that it is safe to report and not trigger any issue at runtime?

java-frontend/src/test/java/org/sonar/java/SonarComponentsTest.java
```diff
@@ -471,6 +512,21 @@ void no_issue_when_check_not_found() {
...
+    JavaFileScanner customScanner = scannerContext -> {
+      // empty
+    };
+    CheckRegistrar registrar =registrarContext ->
```
dorian-burihabwa-sonarsource, line 520:
```suggestion
    CheckRegistrar registrar = registrarContext ->
```

dorian-burihabwa-sonarsource (approved):
Thanks for coming back to this. I just left a question about the Javadoc and a small typo

PR #5086, opened 2025-04-16
SONARJAVA-5457 S7466: Use var instead of a type with unnamed variable

tomasz-tylenda-sonarsource:
[SONARJAVA-5457](https://sonarsource.atlassian.net/browse/SONARJAVA-5457)

[SONARJAVA-5457]: https://sonarsource.atlassian.net/browse/SONARJAVA-5457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


java-checks/src/main/java/org/sonar/java/checks/UnnamedVariableShouldUseVarCheck.java
```diff
@@ -0,0 +1,84 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree instanceof ForEachStatement forEachTree) {
+      visitForEachStatement(forEachTree);
+    } else if (tree instanceof TryStatementTree tryTree) {
+      visitTryStatement(tryTree);
+    }
+  }
+
+  void visitForEachStatement(ForEachStatement forEachTree) {
```
erwan-serandour, line 51: I think we can inline visitForEachStatement and visitTryStatement. It will remove 1 level of indirection and make the code slightly easier to read. I think both functions are rather short.

tomasz-tylenda-sonarsource: Done.

erwan-serandour (approved):

PR #5090, opened 2025-04-16
SONARJAVA-5485 S1481: Report on unnamed variables in try with resources

tomasz-tylenda-sonarsource:
[SONARJAVA-5485](https://sonarsource.atlassian.net/browse/SONARJAVA-5485)

[SONARJAVA-5485]: https://sonarsource.atlassian.net/browse/SONARJAVA-5485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/autoscan/src/test/resources/autoscan/diffs/diff_S1481.json
+++ b/its/autoscan/src/test/resources/autoscan/diffs/diff_S1481.json
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S1481",
   "hasTruePositives": true,
-  "falseNegatives": 8,
+  "falseNegatives": 10,
   "falsePositives": 0
 }
--- a/its/autoscan/src/test/resources/autoscan/diffs/diff_S1481.json
+++ b/its/autoscan/src/test/resources/autoscan/diffs/diff_S1481.json
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S1481",
   "hasTruePositives": true,
-  "falseNegatives": 7,
+  "falseNegatives": 8,
   "falsePositives": 0
 }
--- a/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
@@ -93,7 +93,7 @@ public class UnusedLocalVariableCheck extends IssuableSubscriptionVisitor {
    * For instance in {@code for (String element : list) {}}, it is only since Java 22 that it can be rewritten {@code for (var _ : list) {}}.
    */
   private boolean canBeReplaced(VariableTree variable) {
-  return context.getJavaVersion().isJava22Compatible()
+    return context.getJavaVersion().isJava22Compatible()
       || (!isForeachVariable(variable) && !isTryResource(variable));
   }
 
--- a/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
@@ -93,7 +93,10 @@ public class UnusedLocalVariableCheck extends IssuableSubscriptionVisitor {
    * Before Java 22 it was not possible to remove the variable in a foreach statement or try with resources even it is unused.
    * For instance in {@code for (String element : list) {}}, it is only since Java 22 that it can be rewritten {@code for (var _ : list) {}}.
    */
-  private boolean isMandatory(VariableTree variable) {
+  private boolean canBeReplaced(VariableTree variable) {
+  return context.getJavaVersion().isJava22Compatible()
+      || (!isForeachVariable(variable) && !isTryResource(variable));
+  }
     return !context.getJavaVersion().isJava22Compatible()
       && (isForeachVariable(variable) || isTryResource(variable));
   }
--- a/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
@@ -76,7 +76,7 @@ public class UnusedLocalVariableCheck extends IssuableSubscriptionVisitor {
       IdentifierTree simpleName = variable.simpleName();
       if (!simpleName.isUnnamedVariable()) {
         boolean unresolved = UNRESOLVED_IDENTIFIERS_AND_SWITCH_CASE_VISITOR.isUnresolved(simpleName.name());
-        if (!unresolved && isProperLocalVariable(variable) && isUnused(variable.symbol()) && !isMandatoryForeachVariable(context.getJavaVersion(), variable)) {
+        if (!unresolved && isProperLocalVariable(variable) && isUnused(variable.symbol()) && !isMandatory(variable)) {
           QuickFixHelper.newIssue(context)
             .forRule(this)
             .onTree(simpleName)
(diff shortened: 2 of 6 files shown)


its/autoscan/src/test/resources/autoscan/diffs/diff_S1481.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S1481",
   "hasTruePositives": true,
-  "falseNegatives": 7,
+  "falseNegatives": 8,
```
romainbrenguier, line 4: Why do we have a false positive?

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -143,11 +147,14 @@ private static List<JavaQuickFix> computeQuickFix(VariableTree variable) {
...
-    return List.of(JavaQuickFix.newQuickFix("Replace unused local variable with _")
-      .addTextEdit(JavaTextEdit.replaceTextSpan(textSpan, "var _"))
-      .build());
+  private static JavaQuickFix makeQuickFixReplacingWithUnnamedVariable(VariableTree variable) {
+    return JavaQuickFix.newQuickFix("Replace unused local variable with _")
+      // This works with both enhanced for loop and try-with-resources.
+      // In the latter case we keep the initializer:
+      //   `for(int elem: elems)`  turns into  `for(var _: elems)`
+      //   `try(Resource res = initializer())`  turns into  `try(var _ = initializer ())`
+      .addTextEdit(JavaTextEdit.replaceBetweenTree(variable.type(), true, variable.simpleName(), true, "var _"))
```
romainbrenguier, line 156: Not a request, just an idea: instead of comments here we could have some specific unit tests to demonstrate that.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -90,11 +90,12 @@ public void leaveNode(Tree tree) {
   }
 
   /**
-   * Before Java 22 it was not possible to remove the variable in a foreach statement even it is unused.
+   * Before Java 22 it was not possible to remove the variable in a foreach statement or try with resources even it is unused.
    * For instance in {@code for (String element : list) {}}, it is only since Java 22 that it can be rewritten {@code for (var _ : list) {}}.
    */
-  private static boolean isMandatoryForeachVariable(JavaVersion javaVersion, VariableTree variable) {
-    return variable.parent() instanceof ForEachStatement && !javaVersion.isJava22Compatible();
+  private boolean isMandatory(VariableTree variable) {
```
erwan-serandour, line 96:
```suggestion
  private boolean canBeReplaced(VariableTree variable) {
  return context.getJavaVersion().isJava22Compatible()
      || (!isForeachVariable(variable) && !isTryResource(variable));
  }
```
I find that the negation on mandatory make it a bit complex to read. We can avoid it by taking the negation of isMandtory with canBeReplaced

erwan-serandour (changes requested):

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
tomasz-tylenda-sonarsource: I started writing such a test, but I think it is both long and redundant. The comment serves the purpose of explaining how exactly one method handles two different constructs.

Below is the test that I wrote. Please let me know if you would like to include it (after cleaning it up and expanding it):

```
  private static final String INPUT = """
    class C {
      void m1() {
        try (Stream stream = Stream.of()) {
        } catch (Exception _) {
        }
      }
    }
    """;

  @Test
  void test_quickfix() {
    CompilationUnitTree compilationUnitTree = JParserTestUtils.parse(INPUT);
    var classTree = (ClassTree) compilationUnitTree.types().get(0);
    var m1 = (MethodTree) classTree.members().get(0);
    var tryStmt = (TryStatementTree) m1.block().body().get(0);
    var streamVar = (VariableTree) tryStmt.resourceList().get(0);

    var quickFix = UnusedLocalVariableCheck.makeQuickFixReplacingWithUnnamedVariable(streamVar);
    assertThat(quickFix.getTextEdits()).hasSize(1);
    var textEdit = quickFix.getTextEdits().get(0);
    assertThat(textEdit.getReplacement()).isEqualTo("var _");
    var textSpan = textEdit.getTextSpan();
    assertThat(textSpan.startLine).isEqualTo(4);
    assertThat(textSpan.startCharacter).isEqualTo(9);
    assertThat(textSpan.endLine).isEqualTo(4);
    assertThat(textSpan.endCharacter).isEqualTo(22);
  }
```

its/autoscan/src/test/resources/autoscan/diffs/diff_S1481.json
tomasz-tylenda-sonarsource: There are no false positives. This check produces false negatives, so we got more of them when we expanded the tests.

romainbrenguier: Sorry I meant false negatives. Do you have an idea why the check produces false negatives?

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
romainbrenguier: As you which. It's true that tests like that are not as clear as the other tests you already have.

its/autoscan/src/test/resources/autoscan/diffs/diff_S1481.json
tomasz-tylenda-sonarsource: I'm investigating. I don't know why it was produced, but it looks like the problem happens here: https://github.com/SonarSource/sonar-java/blob/master/java-checks-test-sources/default/src/test/java/checks/tests/AssertionsCompletenessCheckSample.java#L205

tomasz-tylenda-sonarsource: I found examples when we generate FNs and added them to unit tests. The problem affects enhanced for as well, so I think we could tackle it separately.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
tomasz-tylenda-sonarsource: I prefer not to add such a test.  I could remove the comment if you prefer. I orginally added it to explain why the replacement, because I tought it is non-obvious.

erwan-serandour (approved):

java-checks/src/test/java/org/sonar/java/checks/unused/UnusedLocalVariableCheckTest.java
```diff
@@ -59,4 +59,24 @@ void test_non_compiling() {
...
+  /** Test for false negative when a name is used in a lambda expression. */
+  @Test
+  void test_with_lambda_without_semantics() {
+    CheckVerifier.newVerifier()
+      .onFile(TestUtils.mainCodeSourcesPath("checks/unused/UnusedLocalVariableCheck_withLambda.java"))
+      .withCheck(new UnusedLocalVariableCheck())
+      .withJavaVersion(22)
+      .withoutSemantic()
+      // False negatives.
+      .verifyNoIssues();
```
romainbrenguier, line 80: If I understand correctly, this is checking the current behavior, which has false negatives. This means the test will fail when we fix the behavior. Wouldn't it be better to have a test that checks the behavior we want, and disable it for now. Having a ticket for fixing the issue would also be ideal.

tomasz-tylenda-sonarsource: I created a ticket, https://sonarsource.atlassian.net/browse/SONARJAVA-5504, and added a reference in to it in the code.

Having a test that is enabled is better in my opinion. If somebody in the future works on this rule and fixes the problem for some other reason, they will learn about it. If the test is disabled, then any changes to the behavior may not be noticed.

romainbrenguier (approved):

PR #5091, opened 2025-04-17
SONARJAVA-5430 Check Java version in case of foreach variable

romainbrenguier:
The quickfix will only work starting with Java 22.
We should not raise an issue before that.


leonardo-pilastri-sonarsource (approved):
Looks good, thanks for the fix!

PR #5092, opened 2025-04-17
SONARJAVA-4881 Spring @Transactional and @Async annotations can be protected and package private

SanderKnauff:
[SONARJAVA-4881](https://sonarsource.atlassian.net/browse/SONARJAVA-4881)

Since [Spring 6](https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/annotations.html#transaction-declarative-annotations-method-visibility), `@Transactional` annotated methods are allowed to be protected and package-private. This change updates S2230 rule to only look for private methods. 

[SONARJAVA-4881]: https://sonarsource.atlassian.net/browse/SONARJAVA-4881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/files/non-compiling/checks/spring/TransactionalMethodVisibilityCheckSample.java
+++ b/java-checks-test-sources/default/src/main/files/non-compiling/checks/spring/TransactionalMethodVisibilityCheckSample.java
@@ -5,12 +5,12 @@ import org.springframework.transaction.annotation.Transactional;
 class TransactionalMethodVisibilityCheckSample {
   // Cannot compile because a Transactional method should be overridable
   @org.springframework.transaction.annotation.Transactional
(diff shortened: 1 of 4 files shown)


romainbrenguier:
Thanks @SanderKnauff for your contribution!
I think the rule should still apply in its current form for projects that are using Spring-5 or lower. It's now possible to activate or deactivate rules based on dependency versions in the project. So I'm suggesting an alternative way of fixing the issue in this pull request: https://github.com/SonarSource/sonar-java/pull/5094. 
On top of that would probably need a new rule that would specifically apply to private methods in Spring >= 6.

SanderKnauff:
@romainbrenguier That sounds like a better solution, thanks for picking that up! I will close this PR then in favor of #5094.

PR #5093, opened 2025-04-17
SONARJAVA-5483 implement S7475: RemoveTypeFromUnusedPatternCheck

erwan-serandour:
Part of


java-checks-test-sources/default/src/main/java/checks/RemoveTypeFromUnusedPatternCheckSample.java
```diff
@@ -0,0 +1,106 @@
...
+
+public class RemoveTypeFromUnusedPatternCheckSample {
+  record Guest(String name, String email, String phoneNumber) {
+  }
+
+  record Bar(String a, int b){}
+  record Foo(String a, Bar b){}
+
+  void nonCompliantExamples(Object o) {
+    if (o instanceof Guest(String name, String _, String _)) { // Noncompliant 2
```
tomasz-tylenda-sonarsource, line 13: What is the `2` in in `// Noncompliant 2`?

java-checks/src/test/java/org/sonar/java/checks/RemoveTypeFromUnusedPatternCheckTest.java
```diff
@@ -0,0 +1,45 @@
...
+import static org.sonar.java.checks.verifier.TestUtils.mainCodeSourcesPath;
+
+class RemoveTypeFromUnusedPatternCheckTest {
+
+  @Test
+  void test() {
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/RemoveTypeFromUnusedPatternCheckSample.java"))
+      .withCheck(new RemoveTypeFromUnusedPatternCheck())
+      .withJavaVersion(22)
```
tomasz-tylenda-sonarsource, line 31: Should we check it in Java 21 as well?

You can't use `_`, but we could have a very simple sample file and `verifyNoIssues` on input like `if (o instanceof Guest(String name, String b, String c)) {}`

java-checks-test-sources/default/src/main/java/checks/RemoveTypeFromUnusedPatternCheckSample.java
erwan-serandour: 2 nonCompliant examples on the same line, I will remove one of them it is a bit confusing

java-checks/src/test/java/org/sonar/java/checks/RemoveTypeFromUnusedPatternCheckTest.java
erwan-serandour: :+1:

tomasz-tylenda-sonarsource (approved):

PR #5094, opened 2025-04-22
SONARJAVA-4881 With Spring 6, @Transactional and @Async annotated methods don't need to be public

romainbrenguier:
https://sonarsource.atlassian.net/browse/SONARJAVA-4881

Since [Spring 6](https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/annotations.html#transaction-declarative-annotations-method-visibility), @Transactional annotated methods are allowed to be protected and package-private.
This PR updates the rule S2230 to only apply to projects where Spring appears with a version less than 6.

This is based on the PR https://github.com/SonarSource/sonar-java/pull/5092

Part of

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/spring/TransactionalMethodVisibilityCheckSample_Spring5.java
+++ b/java-checks-test-sources/default/src/main/java/checks/spring/TransactionalMethodVisibilityCheckSample_Spring5.java
@@ -1,0 +1,33 @@
+package checks.spring;
+
+import java.util.concurrent.Future;
+import org.springframework.scheduling.annotation.Async;
+import org.springframework.transaction.annotation.Transactional;
+
+abstract class TransactionalMethodVisibilityCheckSample_Spring5 {
+  
+  public interface C {
+    @Transactional
+    int bar(); // Compliant
+  }
+  
+  @Async
+  protected abstract Future<String> aMethod(); // Noncompliant
+
+  @Async
+  Future<String> defaultVisibilityAsyncMethod(){ // Noncompliant
+    return  null;
+  }
+
+  @Async
+  protected Future<String> protectedVisibilityAsyncMethod(){ // Noncompliant
+    return  null;
+  }
+
+  @org.springframework.transaction.annotation.Transactional
+  protected void protectedTransactionalMethod() {} // Noncompliant {{Make this method "public" or remove the "@Transactional" annotation.}}
+//               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+  @org.springframework.transaction.annotation.Transactional
+  void defaultVisibilityTransactionalMethod() {} // Noncompliant {{Make this method "public" or remove the "@Transactional" annotation.}}
+}
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
@@ -19,14 +19,18 @@ package org.sonar.java.checks.spring;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
 import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.DependencyVersionAware;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.Version;
 import org.sonar.plugins.java.api.tree.AnnotationTree;
 import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
 @Rule(key = "S2230")
-public class TransactionalMethodVisibilityCheck extends IssuableSubscriptionVisitor {
+public class TransactionalMethodVisibilityCheck extends IssuableSubscriptionVisitor implements DependencyVersionAware {
 
   private static final List<String> proxyAnnotations = List.of(
     "org.springframework.transaction.annotation.Transactional",
@@ -62,4 +66,8 @@ public class TransactionalMethodVisibilityCheck extends IssuableSubscriptionVisi
     return false;
   }
 
+  @Override
+  public boolean isCompatibleWithDependencies(Function<String, Optional<Version>> dependencyFinder) {
+    return dependencyFinder.apply("spring-tx").map(version -> version.isLowerThan("6.0")).orElse(false);
+  }
 }
--- a/java-checks/src/test/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheckTest.java
+++ b/java-checks/src/test/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheckTest.java
@@ -16,6 +16,8 @@
  */
 package org.sonar.java.checks.spring;
 
+import java.io.File;
+import java.util.List;
 import org.junit.jupiter.api.Test;
 import org.sonar.java.checks.verifier.CheckVerifier;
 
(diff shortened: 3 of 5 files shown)


dorian-burihabwa-sonarsource (changes requested):
I will have a look when the autoscan issue is addressed

dorian-burihabwa-sonarsource (changes requested):
The change looks clean enough but I am afraid we might start introducing a new rule for each to Spring component scanning. My preference would be for us to apply the same rule with a different reporting behavior depending on the detected Spring version. WDYT?

README.md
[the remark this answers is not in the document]
romainbrenguier: https://github.com/SonarSource/sonar-java/pull/5099

java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
```diff
@@ -62,4 +66,8 @@ private static boolean hasAnnotation(MethodTree method, String annotationSymbol)
     return false;
   }
 
+  @Override
+  public boolean isCompatibleWithDependencies(Function<String, Optional<Version>> dependencyFinder) {
+    return dependencyFinder.apply("spring-tx").map(version -> version.isLowerThan("6.0")).orElse(false);
+  }
```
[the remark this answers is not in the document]
romainbrenguier: Yes that's correct. I created a rule idea proposal for the Spring 6 case: https://sonarsource.atlassian.net/browse/SONARJAVA-5508.
I think we can change the behavior by recording the version during the check. I will make that change to this PR.

java-checks-test-sources/default/src/main/java/checks/spring/TransactionalMethodVisibilityCheckSample_Spring5.java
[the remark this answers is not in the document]
romainbrenguier: I've now changed the samples, to split between Spring5 and Spring6. I'm not sure "compliant" would make it clearer in that case. But let me know your thought after the change.

its/autoscan/src/test/resources/autoscan/diffs/diff_S2230.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S2230",
-  "hasTruePositives": true,
-  "falseNegatives": 15,
+  "hasTruePositives": false,
```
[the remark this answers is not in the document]
romainbrenguier: With the changes in that PR, the rule will always be deactivated in autoscan, yes.

its/autoscan/src/test/resources/autoscan/diffs/diff_S6810.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S6810",
   "hasTruePositives": false,
-  "falseNegatives": 6,
+  "falseNegatives": 5,
```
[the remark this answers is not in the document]
romainbrenguier: There was an Async method returning `int` in the samples https://github.com/SonarSource/sonar-java/pull/5094/files#diff-6073b988466d5c00c736c4987d18d535e9da06e9eb97bc1b2e2e82f7aba4b776L21 which is not allowed (it should return void or Future)

java-checks/src/test/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheckTest.java
```diff
@@ -25,13 +28,35 @@
...
       .verifyIssues();
   }
 
+  @Test
+  void test_spring6() {
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/spring/TransactionalMethodVisibilityCheckSample_Spring6.java"))
+      .withCheck(new TransactionalMethodVisibilityCheck())
+      .withClassPath(List.of(new File("spring-tx-6.0.1.jar"),
+        new File("spring-context-6.0.2.jar")))
```
dorian-burihabwa-sonarsource, line 44: Why do we test with 2 versions of Spring 6?

java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
```diff
@@ -36,6 +40,8 @@ public class TransactionalMethodVisibilityCheck extends IssuableSubscriptionVisi
     "org.springframework.transaction.annotation.Transactional", "@Transactional",
     "org.springframework.scheduling.annotation.Async", "@Async");
 
+  private boolean isSpring6OrLater = false;
```
dorian-burihabwa-sonarsource, line 43: From a few issues with non-primitve attributes in the past, and a bit of superstition ever since, we try to avoid having state in the check. Could we imagine removing the field, replacing it with a local variable in `visitNode` and passing that local variable to a static version of `requiredVisibilityMessage`?

java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
```diff
@@ -19,14 +19,18 @@
...
+import org.sonar.plugins.java.api.Version;
 import org.sonar.plugins.java.api.tree.AnnotationTree;
 import org.sonar.plugins.java.api.tree.MethodTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
 @Rule(key = "S2230")
-public class TransactionalMethodVisibilityCheck extends IssuableSubscriptionVisitor {
+public class TransactionalMethodVisibilityCheck extends IssuableSubscriptionVisitor implements DependencyVersionAware {
 
   private static final List<String> proxyAnnotations = List.of(
```
dorian-burihabwa-sonarsource, line 35: Can we take the opportunity to rename these constants with UPPERCASE_SNAKECASE? I got a little confused in `visitNode` trying to understand where theses values came from?

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/test/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheckTest.java
romainbrenguier: No particular reason. I will use the same for both to avoid confusion.

java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
romainbrenguier: The field is initialized in `isCompatibleWithDependencies` so it can't be made local. To be able to get rid of that state we would need to change how the `DependencyVersionAware` interface works and use the `JavaFileScannerContext` to store that information.

dorian-burihabwa-sonarsource (approved):

java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
```diff
@@ -62,4 +73,18 @@ private static boolean hasAnnotation(MethodTree method, String annotationSymbol)
...
+  @Override
+  public boolean isCompatibleWithDependencies(Function<String, Optional<Version>> dependencyFinder) {
+    Optional<Version> springContextVersion = dependencyFinder.apply("spring-context");
+    Optional<Version> springTxVersion = dependencyFinder.apply("spring-tx");
+    if (springTxVersion.isEmpty() && springContextVersion.isEmpty()) {
+      return false;
+    }
+    isSpring6OrLater = springContextVersion
+      .or(() -> springTxVersion)
+      .map(v -> v.isGreaterThanOrEqualTo("6.0"))
```
tomasz-tylenda-sonarsource, line 86: If this check becomes common, then we should add static method returning a predicate, so one can write `.map(isGreaterThanOrEqualTo("6.0"))`

java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
```diff
@@ -62,4 +73,18 @@ private static boolean hasAnnotation(MethodTree method, String annotationSymbol)
...
 
+  /** Check that Spring transaction artifact is present, and record whether its version is before or after 6.0. */
+  @Override
+  public boolean isCompatibleWithDependencies(Function<String, Optional<Version>> dependencyFinder) {
+    Optional<Version> springContextVersion = dependencyFinder.apply("spring-context");
+    Optional<Version> springTxVersion = dependencyFinder.apply("spring-tx");
+    if (springTxVersion.isEmpty() && springContextVersion.isEmpty()) {
+      return false;
+    }
+    isSpring6OrLater = springContextVersion
```
tomasz-tylenda-sonarsource, line 84: Is there a way to avoid writing into variables from `isCompatibleWithDependencies`? It's quite weird when `isSomething` modifies state. If this keeps happening, then maybe the `DependencyVersionAware` should be refactored in future?

java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
romainbrenguier: We could write `VersionImpl.parse("6.0")::isSmallerThan` but I don't think that's clearer

java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodVisibilityCheck.java
romainbrenguier: Yes that's not great and I agree we need to change the DependencyVersionAware interface for these kind of cases. But the changes to have the proper interface would be more complex, according to the experiment we did with Leonardo.

tomasz-tylenda-sonarsource (approved):

PR #5095, opened 2025-04-22
SONARJAVA-5492 stop to raising issues with S1481 inside record pattern as no fix exists before Java 22

erwan-serandour:
[SONARJAVA-5492](https://sonarsource.atlassian.net/browse/SONARJAVA-5492)

[SONARJAVA-5492]: https://sonarsource.atlassian.net/browse/SONARJAVA-5492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -192,32 +194,50 @@ private static Optional<SyntaxToken> getPrecedingComma(VariableTree variable) {
     return QuickFixHelper.previousVariable(variable).map(VariableTree::lastToken);
   }
 
-  private static class UnresolvedIdentifiersAndSwitchCaseVisitor extends UnresolvedIdentifiersVisitor {
-    private final Set<VariableTree> switchPatternVariables = new HashSet<>();
-    private boolean withinCaseLabel = false;
+  private static class CollectIdentifiersProperties extends UnresolvedIdentifiersVisitor {
```
tomasz-tylenda-sonarsource, line 197: Lets use the same name for the class and the constant. `IDENTIFIER_PROPERTIES` and `IdentifierProperties` is fine.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -192,32 +194,50 @@ private static Optional<SyntaxToken> getPrecedingComma(VariableTree variable) {
...
+    private final Set<VariableTree> patternVariables = new HashSet<>();
+    private final Set<VariableTree> caseVariables = new HashSet<>();
+    private int nestedRecordPattern = 0;
+    private boolean caseLabel = false;
```
tomasz-tylenda-sonarsource, line 201: `withinCaseLabel` sounded fine to me (but maybe there is a good reason to change it?). I typically try to avoid renaming existing variable to keep the diffs small, so it is easier to understand what is going on?

java-checks-test-sources/default/src/main/java/checks/unused/UnusedLocalVariableCheck.java
```diff
@@ -375,5 +379,16 @@ void unnamedVariablesUseCases(Queue<Ball> queue, BallHolder<? extends Ball> ball
     if(coloredPoint instanceof ColoredPoint(Point(_, int y), _)) { // Compliant
       System.out.println("Point (_:" + y + ") with color not important");
     }
+
+    if(coloredPoint instanceof ColoredPoint(Point(int x, int y), String color)) { // Compliant, full pattern is necessary before java 22
```
tomasz-tylenda-sonarsource, line 383: I'm getting lost in the these tests. Do we have a case where we raise an isssue on `x` and `y` in Java 22?

erwan-serandour: No, I made the choice to never raise to avoid to complex-iffy too much the rule. My thinking was maybe a bit quick.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
erwan-serandour: :+1:

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedLocalVariableCheck.java
erwan-serandour: :+1:

PR #5104, opened 2025-04-28
SONARJAVA-5400 Fix false positive for S6241 and S6242 when builder is not AwsClientBuilder

romainbrenguier:
Previously these checks were only taking account calls on some specific kind of builders, but there can be other implementations.


erwan-serandour (approved):
LGTM

PR #5106, opened 2025-04-29
SONARJAVA-4334 Fix FP in S6207, in case where parameter was changed in constructor

romainbrenguier:
[SONARJAVA-4334](https://sonarsource.atlassian.net/browse/SONARJAVA-4334)

With the changes in this PR, we more precisely keep track of which component was assigned to what parameter and whether the parameter was changed by the execution of the constructor.

[SONARJAVA-4334]: https://sonarsource.atlassian.net/browse/SONARJAVA-4334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/RedundantRecordMethodsCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/RedundantRecordMethodsCheckSample.java
@@ -6,6 +6,16 @@ import java.util.Locale;
 import java.util.Random;
 
 public class RedundantRecordMethodsCheckSample {
+  record AssignmentInConstructor(int arg1, int arg2) {
+    public AssignmentInConstructor(int arg1, int arg2) { // Compliant
+      this.arg1 = arg1;
+      if (arg2 == 5) {
+        arg2 = 4;
+      }
+      this.arg2 = arg2;
+    }
+  }
+
   record RedundantConstructorAndGetters(String name, int age) {
 
     static Object variable = null;
--- a/java-checks-test-sources/default/src/main/java/checks/RedundantRecordMethodsCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/RedundantRecordMethodsCheckSample.java
@@ -6,15 +6,6 @@ import java.util.Locale;
 import java.util.Random;
 
 public class RedundantRecordMethodsCheckSample {
-  record AssignmentInConstructor(int arg1, int arg2) {
-    public AssignmentInConstructor(int arg1, int arg2) { // Compliant
-      this.arg1 = arg1;
-      if (arg2 == 5) {
-        arg2 = 4;
-      }
-      this.arg2 = arg2;
-    }
-  }
 
   record RedundantConstructorAndGetters(String name, int age) {
 
@@ -174,6 +165,38 @@ public class RedundantRecordMethodsCheckSample {
     @MyAnnotation
     RecordWithCustomAnnotation(String value) { // Compliant: annotated
       this.value = value;
+    }
+  }
+
+  record AssignmentInConstructor(int arg1, int arg2) {
+    public AssignmentInConstructor(int arg1, int arg2) { // Compliant
+      this.arg1 = arg1;
+      if (arg2 == 5) {
+        arg2 = 4;
+      }
+      this.arg2 = arg2;
+    }
+  }
+  record BranchInConstructor(int arg1, int arg2) {
+    public BranchInConstructor(int arg1, int arg2) { // Noncompliant
+      this.arg1 = arg1;
+      if (arg2 == 5) {
+        this.arg2 = arg2;
+      } else {
+        arg1 = 0;
+        this.arg2 = arg2;
+      }
+    }
+  }
+
+  record BranchInConstructor2(int arg1, int arg2) {
+    public BranchInConstructor2(int arg1, int arg2) { // Compliant
+      this.arg1 = arg1;
+      if (arg2 == 5) {
+        this.arg2 = arg2;
+      } else {
+        this.arg2 = 0;
+      }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


java-checks-test-sources/default/src/main/java/checks/RedundantRecordMethodsCheckSample.java
```diff
@@ -4,6 +4,39 @@
 import java.util.Random;
 
 public class RedundantRecordMethodsCheckSample {
+  record AssignmentInConstructor(int arg1, int arg2) {
```
tomasz-tylenda-sonarsource, line 7: IMO sample files should start with typical case and this ticket is about a special case. Let's move the new examples lower.

java-checks/src/main/java/org/sonar/java/checks/RedundantRecordMethodsCheck.java
```diff
@@ -111,66 +113,100 @@ public static boolean onlyDoesSimpleAssignments(MethodTree constructor, List<Sym
...
+  /**
+   * Class to perform a simple symbolic execution of a record constructor. The state keeps track of which components have been
+   * assigned to the corresponding parameters and which parameters have been changed from their initial values.
+   */
+  private static class ConstructorExecutionState {
+    /**
+     * Immutable list of the components in the record
+     */
+    final List<String> componentNames;
+    /**
```
tomasz-tylenda-sonarsource, line 131: Let's use a blank line between fields if they have JavaDoc.

java-checks/src/main/java/org/sonar/java/checks/RedundantRecordMethodsCheck.java
```diff
@@ -111,66 +113,100 @@ public static boolean onlyDoesSimpleAssignments(MethodTree constructor, List<Sym
...
+     * Set of names of components of the record that have been assigned to the corresponding parameter so far
+     */
+    Set<String> assignedComponents = new HashSet<>();
```
tomasz-tylenda-sonarsource, line 134: Let's move all initializations to the constructor.

java-checks/src/main/java/org/sonar/java/checks/RedundantRecordMethodsCheck.java
```diff
@@ -111,66 +113,100 @@ public static boolean onlyDoesSimpleAssignments(MethodTree constructor, List<Sym
...
       return Optional.empty();
     }
-    Symbol variableSymbol = ((MemberSelectExpressionTree) leftHandSide).identifier().symbol();
-    Optional<Symbol.VariableSymbol> component = components.stream()
-      .filter(variableSymbol::equals)
-      .findFirst();
-    if (!component.isPresent()) {
-      return Optional.empty();
+
+    public void applyAssignment(ExpressionTree lhs, ExpressionTree rhs) {
```
tomasz-tylenda-sonarsource, line 162: This method is not used outside the class, so it should be private.

java-checks/src/main/java/org/sonar/java/checks/RedundantRecordMethodsCheck.java
```diff
@@ -111,66 +113,100 @@ public static boolean onlyDoesSimpleAssignments(MethodTree constructor, List<Sym
...
-      return Optional.empty();
+    private ConstructorExecutionState(List<String> componentNames) {
+      this.componentNames = componentNames;
+      unchangedParameters.addAll(componentNames);
     }
-    ExpressionStatementTree initialStatement = (ExpressionStatementTree) statement;
-    if (!initialStatement.expression().is(Tree.Kind.ASSIGNMENT)) {
-      return Optional.empty();
+
+    public boolean isParameter(Symbol symbol) {
```
tomasz-tylenda-sonarsource, line 145: This method is not used outside the class, so it should be private.

java-checks/src/main/java/org/sonar/java/checks/RedundantRecordMethodsCheck.java
```diff
@@ -111,66 +113,100 @@ public static boolean onlyDoesSimpleAssignments(MethodTree constructor, List<Sym
...
-    AssignmentExpressionTree assignment,
-    List<Symbol.VariableSymbol> components,
-    List<Symbol.VariableSymbol> parameters) {
-    ExpressionTree leftHandSide = assignment.variable();
-    if (!leftHandSide.is(Tree.Kind.MEMBER_SELECT)) {
+    /**
+     * Return the name of the component if the given expression is of the form `this.name`.
+     */
+    public Optional<String> isComponent(ExpressionTree expression) {
+      if (expression instanceof MemberSelectExpressionTree memberSelect) {
```
tomasz-tylenda-sonarsource, line 153: If you like doing everything in streams, then this is equivalent:
```
return Optional.of(expression)
        .filter(MemberSelectExpressionTree.class::isInstance)
        .map(MemberSelectExpressionTree.class::cast)
        .map(memberSelect -> memberSelect.identifier().name())
        .filter(componentNames::contains);
```
but I'm not sure if it make the code more or less readable.

java-checks/src/main/java/org/sonar/java/checks/RedundantRecordMethodsCheck.java
```diff
@@ -111,66 +113,100 @@ public static boolean onlyDoesSimpleAssignments(MethodTree constructor, List<Sym
...
-  private static Optional<Symbol.VariableSymbol> assignsParameterToComponent(
-    AssignmentExpressionTree assignment,
-    List<Symbol.VariableSymbol> components,
-    List<Symbol.VariableSymbol> parameters) {
-    ExpressionTree leftHandSide = assignment.variable();
-    if (!leftHandSide.is(Tree.Kind.MEMBER_SELECT)) {
+    /**
+     * Return the name of the component if the given expression is of the form `this.name`.
+     */
+    public Optional<String> isComponent(ExpressionTree expression) {
```
tomasz-tylenda-sonarsource, line 152: Methods that start with "is..." should return boolean. I suggest changing it to something like "getComponent" or "getOptionalComponent"

PR #5108, opened 2025-04-29
SONARJAVA-4481 S6207: Fix FP when a constructor in a record is annotated

tomasz-tylenda-sonarsource:
[SONARJAVA-4481](https://sonarsource.atlassian.net/browse/SONARJAVA-4481)

[SONARJAVA-4481]: https://sonarsource.atlassian.net/browse/SONARJAVA-4481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/RedundantRecordMethodsCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/RedundantRecordMethodsCheckSample.java
@@ -144,9 +144,16 @@ public class RedundantRecordMethodsCheckSample {
     }
   }
 
-  record RecordWithJsonAnnotation(String value) {
+  record RecordWithConstructorAnnotation(String value) {
     @JsonCreator
-    RecordWithJsonAnnotation(@JsonProperty("myName") String value) { // Compliant: annotated
+    RecordWithConstructorAnnotation(@JsonProperty("myName") String value) { // Compliant: annotated
+      this.value = value;
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


java-checks-test-sources/default/src/main/java/checks/RedundantRecordMethodsCheckSample.java
```diff
@@ -141,4 +143,20 @@ public String name() {
       }
     }
   }
+
+  record RecordWithJsonAnnotation(String value) {
+    @JsonCreator
+    RecordWithJsonAnnotation(@JsonProperty("myName") String value) { // Compliant: annotated
```
romainbrenguier, line 149: Could we also have cases where the annotation is only on the parameters and not the method? What should be the behaviour in that case?

romainbrenguier (approved):

java-checks-test-sources/default/src/main/java/checks/RedundantRecordMethodsCheckSample.java
tomasz-tylenda-sonarsource: In this case the rule should still apply because the annotations can be moved to the canonical constructor. I verified it by applying JsonProperty on the canonnical constructor (we don't need JsonCreator in this case).

The ticket is is about DTO annotations on the constructor. I am leaning towards making minimal changes to the current bahavior.

PR #5117, opened 2025-04-30
SONARJAVA-4901 S6856 does not raise when parent class defines model attribute

leonardo-pilastri-sonarsource:
[SONARJAVA-4901](https://sonarsource.atlassian.net/browse/SONARJAVA-4901)

Part of 

[SONARJAVA-4901]: https://sonarsource.atlassian.net/browse/SONARJAVA-4901?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/spring/MissingPathVariableAnnotationCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/spring/MissingPathVariableAnnotationCheckSample.java
@@ -13,6 +13,26 @@ import org.springframework.web.bind.annotation.RequestMapping;
 
 public class MissingPathVariableAnnotationCheckSample {
 
+  class ParentController {
+    @ModelAttribute("viewCfg")
+    public String getView(@PathVariable("view") final String view){
+      return "";
+    }
+  }
+  class ChildController extends ParentController {
+    @GetMapping("/model/{view}") //Compliant, parent class defines 'view' path var in the model attribute
+    public String list(@ModelAttribute("viewCfg") final String viewConfig){
+      return "";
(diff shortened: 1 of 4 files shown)


java-checks-test-sources/default/src/main/java/checks/spring/MissingPathVariableAnnotationCheckSample.java
```diff
@@ -13,6 +13,26 @@
 
 public class MissingPathVariableAnnotationCheckSample {
 
+  class ParentController {
```
erwan-serandour, line 16: What happens if the parent is in a different file?

erwan-serandour (changes requested):

java-checks-test-sources/default/src/main/java/checks/spring/MissingPathVariableAnnotationCheckSample.java
leonardo-pilastri-sonarsource: At line 30 there is a case marked as FP where the `MissingParentChildController` extends `MissingPathVariableParentInDifferentSample` which is a different file. In those cases we do not have access to the parent class declaration and so we cannot see its methods, hence the FP

erwan-serandour: thank you :+1:, I missed it

erwan-serandour (approved):

PR #5123, opened 2025-05-02
SONARJAVA-4376 Fix S2129 in the case of missing semantic

romainbrenguier:
[SONARJAVA-4376](https://sonarsource.atlassian.net/browse/SONARJAVA-4376)

When semantic is missing for the argument of String constructor for instance, the parser takes a guess at which constructor it is and this may cause some false positives. The fix in this PR, skips the case where we detect an unknown in the argument types. The second sample is comming from SONARJAVA-5388

[SONARJAVA-4376]: https://sonarsource.atlassian.net/browse/SONARJAVA-4376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/StringPrimitiveConstructorCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/StringPrimitiveConstructorCheckSample.java
@@ -1,7 +1,11 @@
 package checks;
 
+import com.google.api.client.http.HttpResponse;
+import com.google.common.primitives.Bytes;
+import java.io.IOException;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.util.List;
 
 class StringPrimitiveConstructorCheckSample {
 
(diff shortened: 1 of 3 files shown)


java-checks/src/test/java/org/sonar/java/checks/StringPrimitiveConstructorCheckTest.java
```diff
@@ -28,6 +28,7 @@ void test() {
     CheckVerifier.newVerifier()
       .onFile(mainCodeSourcesPath("checks/StringPrimitiveConstructorCheckSample.java"))
       .withCheck(new StringPrimitiveConstructorCheck())
+      .withoutSemantic()
```
tomasz-tylenda-sonarsource, line 31: Let's have two tests - one with semantics and one without. We usually call them `test` and `test_without_semantics`. You may need two sample files.

java-checks-test-sources/default/src/main/java/checks/StringPrimitiveConstructorCheckSample.java
```diff
@@ -57,6 +61,14 @@ void foo() {
     BigDecimal doubleBigDecimal = BigDecimal.valueOf(1.1);
     Short myShort = (short) 0;
   }
+
+  String withoutSemantic(HttpResponse response) throws IOException {
```
tomasz-tylenda-sonarsource, line 65: Maybe we can make the test method names more uniform (`fromHttp` and `fromGuava`) and add a comment explaining that these sample require special treatment when semantics is not available?

PR #5130, opened 2025-05-06
SONARJAVA-4473 Check that mutable member storage is reachable by public calls

romainbrenguier:
[SONARJAVA-4473](https://sonarsource.atlassian.net/browse/SONARJAVA-4473)

This is to be more precise and avoid false positives where the storage only happens in private methods, and calls to this private methods by public ones are protected in some way.

[SONARJAVA-4473]: https://sonarsource.atlassian.net/browse/SONARJAVA-4473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/MutableMembersUsageCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/MutableMembersUsageCheckSample.java
@@ -564,16 +564,43 @@ class Java9Methods {
 
 class SetThroughPrivateMethods {
   private byte[] buf;
-  private SetThroughPrivateMethods(final byte[] buf) {
-    this.buf = buf;  // Compliant: the constructor is private
+  private Map<String, String> map;
+
+  private SetThroughPrivateMethods(final byte[] buf, final Map<String, String> map) {
+    this.map = map; // Noncompliant
+    this.buf = buf; // Compliant: callers have to go through `of` which makes a clone of the parameter
   }
 
   public static SetThroughPrivateMethods of(final byte[] buf) {
-    return new SetThroughPrivateMethods(buf.clone());
+    return new SetThroughPrivateMethods(buf.clone(), Collections.emptyMap());
   }
 
-  public void set(final byte[] buf) {
+  public static SetThroughPrivateMethods of(final byte[] buf, final int off, final int len) {
+    return of(buf);
+  }
+
+  private void set(final byte[] buf) {
+    // It is possible to call `set` from outside the class using `setBuf`
     this.buf = buf;  // Noncompliant
   }
 
+  private void setBuffer(final byte[] buf) {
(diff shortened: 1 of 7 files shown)


java-checks/src/main/java/org/sonar/java/checks/MutableMembersUsageCheck.java
```diff
@@ -81,12 +89,183 @@ public class MutableMembersUsageCheck extends BaseTreeVisitor implements JavaFil
...
+
+  private final Deque<String> methodSignatureStack = new ArrayDeque<>();
+
+  /**
+   * Maps index of arguments at the call site to parameters of the calling method.
+   * For instance in `int f(int a, int b) { g(b, 2, 3); h(4, a);}`, we would have two call-sites:
+   * `{g, <0 -> 1>} and {h, <1 -> 0>}`
+   */
+  @VisibleForTesting
+  record CallSite(String methodSignature, Map<Integer, Integer> parameters) {
```
erwan-serandour, line 102: why using a Map: parameter => parameter of the enclosing function instead of a list or even a Map: enclosing function parameter to List<Parameter>.
We could simply the code below if we reverse the map.
```
 callSite.parameters.entrySet().stream()
              .filter(entry -> entry.getValue().equals(current.paramIndex()))
              .map(entry -> new MethodEntryPoint(callSite.methodSignature, entry.getKey()))
              .forEach(toProcess::add);
```

java-checks-test-sources/default/src/main/java/checks/MutableMembersUsageCheckSample.java
```diff
@@ -564,16 +564,43 @@ public List<String> modifiable() {
...
+    this.buf = buf; // Compliant: this is only callable by private methods
+  }
+
+  private void internalSetBuffer(byte[] buf) {
+    internalSet(buf);
+  }
+
+  static SetThroughPrivateMethods ofMap(final Map<String, String> map) {
+    return new SetThroughPrivateMethods(null, map);
+  }
```
erwan-serandour, line 605: let's add an example with a cycle such as:

```
public void cycle(List<Integer> l){
   cycleA(l);
}
private void cycleA(List<Integer> l){
   cycleB(l);
}
private void cycleB(List<Integer> l){
   cycleA(l);
   pirvateField = l; // Noncompliant
}
```

java-checks/src/main/java/org/sonar/java/checks/MutableMembersUsageCheck.java
romainbrenguier: I changed for a List.

PR #5132, opened 2025-05-07
SONARJAVA-4634 Test for duplicates in credential method json file

romainbrenguier:
[SONARJAVA-4634](https://sonarsource.atlassian.net/browse/SONARJAVA-4634)

Adding a test for duplicated entries found a few cases where we could have false positives because of wildcard usage.

[SONARJAVA-4634]: https://sonarsource.atlassian.net/browse/SONARJAVA-4634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-aws/src/test/java/org/sonar/java/checks/security/HardCodedCredentialsShouldNotBeUsedCheckTest.java
+++ b/java-checks-aws/src/test/java/org/sonar/java/checks/security/HardCodedCredentialsShouldNotBeUsedCheckTest.java
@@ -17,12 +17,16 @@
 package org.sonar.java.checks.security;
 
 
+import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.slf4j.event.Level;
 import org.sonar.api.testfixtures.log.LogAndArguments;
 import org.sonar.api.testfixtures.log.LogTesterJUnit5;
+import org.sonar.java.checks.helpers.CredentialMethod;
+import org.sonar.java.checks.helpers.CredentialMethodsLoader;
 import org.sonar.java.checks.verifier.CheckVerifier;
 import org.sonar.java.checks.verifier.TestUtils;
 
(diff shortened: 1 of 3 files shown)


java-checks-aws/src/main/resources/org/sonar/java/checks/security/S6437-methods.json
```diff
@@ -7438,7 +7438,7 @@
   {"cls":"redis.clients.jedis.JedisSentinelPool","name":"JedisSentinelPool","args":["java.lang.String","java.util.Set","*","int","java.lang.String","int","java.lang.String"],"indices":[4]},
   {"cls":"redis.clients.jedis.JedisSentinelPool","name":"JedisSentinelPool","args":["java.lang.String","java.util.Set","*","int","java.lang.String","java.lang.String","int"],"indices":[5]},
   {"cls":"redis.clients.jedis.JedisSentinelPool","name":"JedisSentinelPool","args":["java.lang.String","java.util.Set","*","int","java.lang.String","java.lang.String","int","java.lang.String"],"indices":[5]},
-  {"cls":"redis.clients.jedis.JedisSentinelPool","name":"JedisSentinelPool","args":["java.lang.String","java.util.Set","*","java.lang.String"],"indices":[3]},
+  {"cls":"redis.clients.jedis.JedisSentinelPool","name":"JedisSentinelPool","args":["java.lang.String","java.util.Set","org.apache.commons.pool2.impl.GenericObjectPoolConfig","java.lang.String"],"indices":[3]},
```
leonardo-pilastri-sonarsource, line 7441: I think the entry with `*` that was removed here was covering both line 7441 and line 7443, maybe instead of replacing this entry we could remove line 7443 as it's redundant? 
I'm not sure if the intent of these entries is to prevent or to fix, the `*` approach would cover possible future new signatures, while being more specific with the naming allows to intervene only when needed.
I think we might want an opinion from  Hendrik here, as he's the main contributor to this list

leonardo-pilastri-sonarsource (approved):

PR #5136, opened 2025-05-08
SONARJAVA-5553 Raise only when returned mutable actually reaches outside the class for S2384

romainbrenguier:
https://sonarsource.atlassian.net/browse/SONARJAVA-5553

This is to reduce FPs where a mutable field is returned by a private method but this value is only used within the class and never escapes.

[SONARJAVA-4473]: https://sonarsource.atlassian.net/browse/SONARJAVA-4473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/ruling/src/test/resources/eclipse-jetty-similar-to-main/java-S2384.json
+++ b/its/ruling/src/test/resources/eclipse-jetty-similar-to-main/java-S2384.json
@@ -88,8 +88,7 @@
 196
 ],
 "org.eclipse.jetty:jetty-project:jetty-server/src/main/java/org/eclipse/jetty/server/SecureRequestCustomizer.java": [
-354,
-441
+354
 ],
 "org.eclipse.jetty:jetty-project:jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java": [
 343,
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/MutableMembersUsageCheck.java
```diff
@@ -150,6 +154,16 @@ private static class MutableDataPropagationGraph {
      */
     private final Set<String> nonPrivateMethods = new HashSet<>();
 
+    /**
+     * Maps method that return the result of another method
+     */
+    private final Map<String, String> passingThroughMethod = new HashMap<>();
```
erwan-serandour, line 160: methods can have several returns

erwan-serandour (changes requested):

erwan-serandour (approved):

erwan-serandour (changes requested):

erwan-serandour (approved):

PR #5153, opened 2025-05-14
SONARJAVA-5438 Fix false positive in S125 for comments ending with semicolon

benzonico:
[SONARJAVA-5438](https://sonarsource.atlassian.net/browse/SONARJAVA-5438)

* Improve detection of natural language comments with semicolons

* Add test cases for license headers and natural text

* Reduce probability of semicolon detection in non-code contexts

Part of 

[SONARJAVA-5438]: https://sonarsource.atlassian.net/browse/SONARJAVA-5438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CommentedOutCodeLineCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CommentedOutCodeLineCheck.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 import org.apache.commons.lang3.StringUtils;
 import org.sonar.check.Rule;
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/CommentedOutCodeLineCheck.java
```diff
@@ -114,6 +118,12 @@ public AnalyzerMessage collectIssues(List<AnalyzerMessage> issues, SyntaxTrivia
     return issue;
   }
 
+  private boolean isLicenseOrNaturalLanguage(String line) {
+    String trimmedLine = line.trim();
```
alban-auzeill, line 122: - why doing a `.trim()` before applying regular expressions that start and end with `.*` or`\s*`?
- why complexifying `NATURAL_LANGUAGE_PATTERN` with a `$` at the end, if it is only used with `.matches()` 
- why using regular expression starting and ending with `.*` and trying of match the full line `.matcher(trimmedLine).matches()` instead of removing the useless `.*` and stoping the evaluation as soon as we found what we want `.matcher(line).find()`

java-checks/src/main/java/org/sonar/java/checks/JavaFootprint.java
```diff
@@ -30,7 +30,9 @@ public final class JavaFootprint implements LanguageFootprint {
   private final Set<Detector> detectors = new HashSet<>();
 
   public JavaFootprint() {
-    detectors.add(new EndWithDetector(0.95, '}', ';', '{'));
+    // Reduce probability for semicolon to avoid false positives with natural language
+    detectors.add(new EndWithDetector(0.95, '}', '{'));
+    detectors.add(new EndWithDetector(0.3, ';'));
```
alban-auzeill, line 35: I strongly disagree with the change of probability from `0.95` to `0.3` for a `;` at the end of the line when detecting Java code! Java is not Kotlin, the lines almost always end with a `;`

alban-auzeill (changes requested):

alban-auzeill:
Closed as duplicated, already fixed by #4994 / SONARJAVA-5273

PR #5158, opened 2025-05-16
SONARJAVA-5537 remove FPs and reduce noise

erwan-serandour:
The rule doesn't raise anymore on:
- licence headers at the start of files
- on java code with source version above or equal 23. we don't raise an issue if a markdown comment start with 4 slashes.


java-checks/src/test/java/org/sonar/java/checks/CommentsMustStartWithCorrectNumberOfSlashesCheckTest.java
```diff
@@ -18,27 +18,38 @@
...
   @Test
   void test_after_java23() {
     CheckVerifier.newVerifier()
-      .onFile(mainCodeSourcesPath("checks/CommentsMustStartWithCorrectNumberOfSlashesCheckAfterJava23.java"))
-      .withCheck(new CommentsMustStartWithCorrectNumberOfSlashesCheck())
+      .onFile(mainCodeSourcesPath("checks/CommentsMustStartWithCorrectNumberOfSlashesCheckSample.java"))
+      .withCheck(check)
       .withJavaVersion(23)
-      .verifyIssues();
+      .verifyNoIssues();
```
tomasz-tylenda-sonarsource, line 52: We don't have any examples starting with 4 or 5 `/`? Why not? I can imagine someone writing `////` instead of `///`.

its/autoscan/src/test/resources/autoscan/diffs/diff_S7476.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S7476",
-  "hasTruePositives": true,
+  "hasTruePositives": false,
```
tomasz-tylenda-sonarsource, line 3: Is this rule skipped in autoscan? I thought we would have a few examples from the code introduces in unit test.

java-checks/src/test/java/org/sonar/java/checks/CommentsMustStartWithCorrectNumberOfSlashesCheckTest.java
```diff
@@ -18,27 +18,38 @@
...
-      .onFile(mainCodeSourcesPath("checks/CommentsMustStartWithCorrectNumberOfSlashesCheckBeforeJava23.java"))
-      .withCheck(new CommentsMustStartWithCorrectNumberOfSlashesCheck())
+      .onFile(mainCodeSourcesPath("checks/CommentsMustStartWithCorrectNumberOfSlashesCheckSample.java"))
+      .withCheck(check)
       .withJavaVersion(22)
       .verifyIssues();
   }
 
   @Test
   void test_after_java23() {
```
tomasz-tylenda-sonarsource, line 47: "after 23" would be at least 24 :) Can we call is `test_java23`?

erwan-serandour: :+1:

its/autoscan/src/test/resources/autoscan/diffs/diff_S7476.json
erwan-serandour: it because autoscan set the java sources to 23.

java-checks-test-sources/default/src/main/java/checks/CommentsMustStartWithCorrectNumberOfSlashesCheckJava23.java
```diff
@@ -0,0 +1,58 @@
...
+//
+// this a license header, we don't raise on it as it will not generate javadoc
+//
+///////////////////////////////////////////////////////////////////////////////////////////////
+package checks;
+
+public class CommentsMustStartWithCorrectNumberOfSlashesCheckJava23 {
+  // This is a comment
+  public void twoSlashes() {}
+  /// javadoc using markdown
```
tomasz-tylenda-sonarsource, line 11: Let's use blank lines between declarations when they have a comment.

tomasz-tylenda-sonarsource (approved):

PR #5176, opened 2025-05-28
SONARJAVA-5534 Remove SE engine and unused duplicated check test sources

alban-auzeill:


its/plugin/tests/src/test/java/com/sonar/it/java/suite/PackageInfoTest.java
```diff
@@ -74,7 +74,7 @@ public void should_take_into_account_package_annotations() {
     orchestrator.executeBuild(build);
 
     List<Issue> issues = TestUtils.issuesForComponent(orchestrator, projectKey);
-    assertThat(issues).hasSize(11);
+    assertThat(issues).isEmpty();
```
alban-auzeill, line 77: We should understand why?

alban-auzeill: This test use a profile `profile-package-info-annotations.xml` that activate only symbolic execution ruiles (S2259, S2589). This is why the test raise no issue without the SE rules. We need to remove the test.

tomasz-tylenda-sonarsource (approved):
LGTM with a minor comments about unnecessary changes to new lines at the end of a few files.

We discussed the change offline. I verified that everything removed looks related to symbolic execution.

PR #5181, opened 2025-06-03
Update tomcat-embed-jasper to from 9.0.104 to 9.0.105 to suppress ale…

tomasz-tylenda-sonarsource:


dorian-burihabwa-sonarsource (approved):
Thanks for the change. I see that [JspC's implementation of `getUseNonstandardTagOptimizations`](https://github.com/apache/tomcat/blob/15c524d11651708bc046929139e6b633f2a50fb5/java/org/apache/jasper/JspC.java#L1801) returns some value but since this object is not meant  to be used as anything else than a placeholder, it should be fine.

PR #5185, opened 2025-06-10
SONARJAVA-5598 S3063: StringBuilder data should be used

tomasz-tylenda-sonarsource:
[SONARJAVA-5598](https://sonarsource.atlassian.net/browse/SONARJAVA-5598)

[SONARJAVA-5598]: https://sonarsource.atlassian.net/browse/SONARJAVA-5598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


java-checks-test-sources/default/src/main/java/checks/unused/UnusedStringBuilderCheckSample.java
```diff
@@ -0,0 +1,123 @@
...
+    StringBuilder stringBuilder;
+
+    UnusedField() {
+      this.stringBuilder = new StringBuilder();
+    }
+
+    void appendHello() {
+      stringBuilder.append("Hello");
+    }
+  }
```
erwan-serandour, line 156: I think we can add an example with a chained call
```suggestion
  sb.append("a").append("b").toString()
```

java-checks-test-sources/default/src/main/java/checks/unused/UnusedStringBuilderCheckSample.java
```diff
@@ -108,16 +123,58 @@ private void unrelated() {
...
+    return sb.append("a").append("b").toString();
+  }
+
+  String unusedManyChainedCalls() {
+    StringBuilder sb = new StringBuilder(); // Noncompliant {{Consume or remove this unused StringBuilder}}
+//                ^^
+    sb.append("a").append("b");
+    return "ab";
+  }
+
```
erwan-serandour, line 144:
```suggestion
 String unusedManyChainedCallsAssign() {
    StringBuilder sb = new StringBuilder();
    StringBuilder sb2 = sb.append("a").append("b");
    return "ab";
  }
```
tricky interaction as a string builder return itself, not sure if we should handle the case. You can choose, but let's add the test case.

tomasz-tylenda-sonarsource: We do a recursive call when a method other than `toString` is called, so we will check both appends, but neither consumes `sb`. `sb2` is ignored, because it is not initialized by a constructor.

PR #5190, opened 2025-06-12
SONARJAVA-5619 NPE when semantic can not resolve "java.lang.Object"

alban-auzeill:
[SONARJAVA-5619](https://sonarsource.atlassian.net/browse/SONARJAVA-5619)


alban-auzeill:
Duplicated and not the right PR

PR #5191, opened 2025-06-13
Fix issue in CFG computation with generic record pattern

romainbrenguier:


java-frontend/src/test/java/org/sonar/java/cfg/CFGTest.java
```diff
@@ -3149,6 +3151,37 @@ void throw_statement_within_try_catch() {
 
   }
 
+  @Test
+  void generic_record_pattern() {
+    CFG cfg = buildCFG("""
+        private static boolean testGen3(Object o) throws Throwable {
+          return o instanceof GenRecord1<?, ?>(Integer i, var s) && i.intValue() == 3 && s.length() == 0;
```
dorian-burihabwa-sonarsource: I am trying some corner cases here but we should probably test with a bounded wild card

dorian-burihabwa-sonarsource: `return o instanceof GenRecord1<?, ? extends CharSequence>(Integer i, var s) && i.intValue() == 3 && s.length() == 0;`

PR #5196, opened 2025-06-16
SONARJAVA-5622 extend S7158 to work with all CharSequence not just Srings

erwan-serandour:


java-checks/src/main/java/org/sonar/java/checks/StringIsEmptyCheck.java
```diff
@@ -145,9 +160,11 @@ private static JavaQuickFix getQuickFix(Tree comparisonExpression, MethodInvocat
...
 
   @Nullable
-  private static MethodInvocationTree getLengthCall(ExpressionTree tree) {
-    if (tree instanceof MethodInvocationTree mit && LENGTH_METHOD.matches(mit)) {
-      return mit;
+  private MethodInvocationTree getLengthCall(ExpressionTree tree) {
+    if (tree instanceof MethodInvocationTree mit) {
+      if (STRING_LENGTH_METHOD.matches(mit) || (javaVersionAsInt >= 15 && CHARSEQUENCE_LENGTH_METHOD.matches(mit))) {
+        return mit;
+      }
```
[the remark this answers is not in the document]
erwan-serandour: I think it is better when there is two ifs.

java-checks/src/test/java/org/sonar/java/checks/StringIsEmptyCheckTest.java
```diff
@@ -31,6 +31,15 @@ void test() {
       .verifyIssues();
   }
 
+  @Test
+  void charSequence() {
```
tomasz-tylenda-sonarsource: Let's also test on Java 8, because it's still popular and CharSequence doesn't have isEmpty.

java-checks/src/main/java/org/sonar/java/checks/StringIsEmptyCheck.java
```diff
@@ -112,7 +127,7 @@ public void visitNode(Tree tree) {
         .newIssue(context)
         .forRule(this)
         .onTree(tree)
-        .withMessage("Use isEmpty() to check whether a string is empty or not.")
+        .withMessage("Use \"isEmpty()\" to check whether a \"CharSequence\" is empty or not.")
```
tomasz-tylenda-sonarsource: `CharSequence` is rarely used directly, so it would be nice to be specific here and say `String`, `StringBuilder`, etc. Could you please try to do this? I see that `mit.methodSymbol().owner().type()` contains the name that we need.

java-checks/src/main/java/org/sonar/java/checks/StringIsEmptyCheck.java
```diff
@@ -71,11 +78,19 @@ private static boolean isEmptinessCheck(ComparisonType comparisonType) {
     return comparisonType == ComparisonType.IS_EMPTY || comparisonType == ComparisonType.IS_NOT_EMPTY;
   }
 
-  // `String.isEmpty()` is available since Java 6.
+  private int javaVersionAsInt;
```
tomasz-tylenda-sonarsource: Why introduce a new field? Can't we just do `context.getJavaVersion().isJava15Compatible()` when we need to check it?

java-checks/src/test/java/org/sonar/java/checks/StringIsEmptyCheckTest.java
erwan-serandour: :+1:

java-checks/src/main/java/org/sonar/java/checks/StringIsEmptyCheck.java
erwan-serandour: :+1:

java-checks/src/main/java/org/sonar/java/checks/StringIsEmptyCheck.java
erwan-serandour: :+1:

java-checks-test-sources/default/src/main/java/checks/CharSequenceIsEmptyCheckSample.java
```diff
@@ -0,0 +1,92 @@
+package checks;
+
+import java.nio.CharBuffer;
+
+public class CharSequenceIsEmptyCheckSample {
+
```
tomasz-tylenda-sonarsource: There are multiple occurrences of multiple blank lines in this file? Is this intentional?

java-checks-test-sources/default/src/main/java/checks/CharSequenceIsEmptyCheckSample.java
```diff
@@ -0,0 +1,92 @@
...
+import java.nio.CharBuffer;
+
+public class CharSequenceIsEmptyCheckSample {
+  
+
+
+  public boolean testStringBuilder(StringBuilder sb1, StringBuilder sb2) {
+    boolean b;
+
+    b = sb1.length() == 0; // Noncompliant {{Use "isEmpty()" to check whether a "AbstractStringBuilder" is empty or not.}}
```
tomasz-tylenda-sonarsource: This is surprising. Why is this an `AbstractStringBuilder` and not a `StringBuilder`?

java-checks/src/main/java/org/sonar/java/checks/StringIsEmptyCheck.java
```diff
@@ -127,7 +120,7 @@ public void visitNode(Tree tree) {
         .newIssue(context)
         .forRule(this)
         .onTree(tree)
-        .withMessage("Use \"isEmpty()\" to check whether a \"CharSequence\" is empty or not.")
+        .withMessage("Use \"isEmpty()\" to check whether a \""+lengthCall.methodSymbol().owner().name()+"\" is empty or not.")
```
tomasz-tylenda-sonarsource: Spaces around operators.

java-checks/src/test/java/org/sonar/java/checks/StringIsEmptyCheckTest.java
```diff
@@ -24,13 +24,40 @@
 class StringIsEmptyCheckTest {
 
   @Test
-  void test() {
+  void stringBuilder() {
```
tomasz-tylenda-sonarsource: Shouldn't this be just "string"? Same for `stringBuilder_15` case.

java-checks/src/main/java/org/sonar/java/checks/StringIsEmptyCheck.java
erwan-serandour: :+1:

java-checks/src/test/java/org/sonar/java/checks/StringIsEmptyCheckTest.java
erwan-serandour: yes, indeed. good catch

java-checks-test-sources/default/src/main/java/checks/CharSequenceIsEmptyCheckSample.java
erwan-serandour: StringBuilder does not override the method, so length is implemented inside AbstractStringBuilder. I have an idea how to solve it.

PR #5200, opened 2025-06-18
SONARJAVA-5598 S3063: Fix FPs on usage in expressions and getChars

tomasz-tylenda-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedStringBuilderCheck.java
```diff
@@ -114,7 +109,7 @@ private static boolean isConsumed(Tree tree) {
           .filter(UnusedStringBuilderCheck::isConsumed)
           .isPresent();
       }
-    } else if (parent instanceof ReturnStatementTree || parent instanceof ArgumentListTreeImpl) {
+    } else if (parent instanceof ReturnStatementTree || parent instanceof ArgumentListTreeImpl || parent instanceof ExpressionTree) {
```
erwan-serandour: Is there a specific reason to use ExpressionTree over BinaryExpressionTree

tomasz-tylenda-sonarsource: No. I'll change it to BinaryExpressionTree.

PR #5201, opened 2025-06-19
SONARJAVA-5633 Expose Configuration inside ModuleScannerContext

irina-batinic-sonarsource:
[SONARJAVA-5633](https://sonarsource.atlassian.net/browse/SONARJAVA-5633)

[SONARJAVA-5633]: https://sonarsource.atlassian.net/browse/SONARJAVA-5633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-frontend/src/test/java/org/sonar/java/SonarComponentsTest.java
+++ b/java-frontend/src/test/java/org/sonar/java/SonarComponentsTest.java
@@ -1372,4 +1372,17 @@ class SonarComponentsTest {
 
     assertThat(sonarComponentsWithSonarLintCache02.sonarLintCache()).isSameAs(sonarLintCache);
   }
+
+  @Test
+  void test_getConfiguration() {
+    var settings = new MapSettings();
+    var sonarComponents = new SonarComponents(null, null, null, null, null, null);
+    sonarComponents.setSensorContext(SensorContextTester.create(new File("")).setSettings(settings));
+
+    var configuration = sonarComponents.getConfiguration();
+    assertThat(configuration)
+      .isNotNull()
+      .isInstanceOf(Configuration.class);
+  }
+
 }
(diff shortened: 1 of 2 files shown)


sonar-java-plugin/src/main/java/org/sonar/plugins/java/JavaSensor.java
```diff
@@ -101,6 +101,7 @@ public void execute(SensorContext context) {
 
     sonarComponents.setSensorContext(context);
     sonarComponents.setCheckFilter(createCheckFilter(sonarComponents.isAutoScanCheckFiltering()));
+    sonarComponents.setConfiguration(settings);
```
alban-auzeill: When we call `sonarComponents.setSensorContext(context);` we already provide `context.config()` so we don't need `sonarComponents.setConfiguration(settings);`.
Why do we need `setConfiguration`?

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -683,4 +684,13 @@ public SensorContext context() {
...
     return sonarLintCache;
   }
+
+  public void setConfiguration(Configuration settings) {
+    this.configuration = settings;
+  }
+
+  public Configuration getConfiguration() {
+    return configuration;
+  }
```
alban-auzeill: We can already access `Configuration` using `sonarComponents.context().config()`, no need to add `getConfiguration/setConfiguration`.

java-frontend/src/main/java/org/sonar/java/model/DefaultModuleScannerContext.java
```diff
@@ -102,4 +103,11 @@ public SonarProduct sonarProduct() {
 
     return context.runtime().getProduct();
   }
+
+  @Override
+  public Configuration getConfiguration() {
+    return sonarComponents.getConfiguration();
```
alban-auzeill, line 109:
```suggestion
    return sonarComponents.context().config();
```

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
irina-batinic-sonarsource: That is very true!
Updated the code.

PR #5204, opened 2025-06-20
SONARJAVA-5599 S3024: Report issues only inside loops

tomasz-tylenda-sonarsource:
[SONARJAVA-5599](https://sonarsource.atlassian.net/browse/SONARJAVA-5599)

[SONARJAVA-5599]: https://sonarsource.atlassian.net/browse/SONARJAVA-5599?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


java-checks/src/main/java/org/sonar/java/checks/StringBufferAndBuilderConcatenationCheck.java
```diff
@@ -54,27 +60,43 @@ protected MethodMatchers getMethodInvocationMatchers() {
...
-    );
+        .forRule(this)
+        .onTree(arg)
+        .withMessage("Use multiple calls to \"append\" instead of string concatenation.")
+        .withQuickFix(() -> getQuickFix(mit.arguments()))
+        .report();
+    }
+  }
+
+  private static boolean isInsideLoop(Tree tree) {
```
erwan-serandour: It might be slow to do it for every concatenation. Move the check to a standard IssuableSubscriptionVisitor. And keep track if we are inside a loop when visiting the tree.


```
    if(tree instanceof Loop){
       loopNesting+=1;
    }
  }

    if(tree instanceof Loop){
       loopNesting-=1;
    }
  }
```

erwan-serandour (approved):

java-checks/src/main/java/org/sonar/java/checks/StringBufferAndBuilderConcatenationCheck.java
tomasz-tylenda-sonarsource: Done.

PR #5208, opened 2025-06-23
SONARJAVA-2511 S4030: Do not raise if there is not usage at all

tomasz-tylenda-sonarsource:


java-checks-test-sources/default/src/main/java/checks/unused/UnusedCollectionCheckSample.java
```diff
@@ -17,6 +17,13 @@ int getLengthComp(String a, String b, String c) {
     return a.length() + b.length() + c.length();
   }
 
+  int getLengthNoUsageAtAll(String a, String b, String c) {
+    // Do not report in this case: lack of usage may indicate unreliable data about the symbol
+    // and this case should be another rule anyway.
```
dorian-burihabwa-sonarsource: In this case, we can explicitly comment that this case is covered by [java:S1481](https://sonarsource.github.io/rspec/#/rspec/S1481/java) and [java:S1854](https://sonarsource.github.io/rspec/#/rspec/S1854/java)
```suggestion
    // and this case should be covered by S1481 and S1854.
```

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedCollectionCheck.java
```diff
@@ -52,6 +52,9 @@ public void visitNode(Tree tree) {
       if (symbol.type().isSubtypeOf("java.util.Collection") &&
         symbol.isLocalVariable() &&
         isInitializedByConstructor(variableTree.initializer()) &&
+        // Do not raise without any usage: usages() may be unreliable,
+        // and it should be handled as a regular unused variable (different rule).
```
dorian-burihabwa-sonarsource: Let's mention the rules expclitly here

tomasz-tylenda-sonarsource: Done.

dorian-burihabwa-sonarsource (approved):

PR #5212, opened 2025-06-25
SONARJAVA-5651 org.sonarsource.java:java-extension-plugin should comply with maven central requirements

alban-auzeill:
[SONARJAVA-5651](https://sonarsource.atlassian.net/browse/SONARJAVA-5651)

[SONARJAVA-5651]: https://sonarsource.atlassian.net/browse/SONARJAVA-5651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/plugin/projects/pom.xml
+++ b/its/plugin/projects/pom.xml
@@ -9,6 +9,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.deploy.skip>true</maven.deploy.skip>
   </properties>
 
   <build>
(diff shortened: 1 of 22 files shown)


its/plugin/projects/pom.xml
dorian-burihabwa-sonarsource, line 1: Not sure why we have a parent pom file here that is neither referred to as a submodule or a parent pom. Probably candidate for deletion

dorian-burihabwa-sonarsource (approved):

PR #5225, opened 2025-07-01
SONARJAVA-5494 Add quickfixes for S7477

romainbrenguier:
[SONARJAVA-5661](https://sonarsource.atlassian.net/browse/SONARJAVA-5661)

[SONARJAVA-5661]: https://sonarsource.atlassian.net/browse/SONARJAVA-5661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/ClassNameInClassTransformCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/ClassNameInClassTransformCheckSample.java
@@ -21,8 +21,10 @@ public class ClassNameInClassTransformCheckSample {
     ClassFile classFile = ClassFile.of();
     ClassModel classModel = classFile.parse(path);
     return classFile.transformClass(classModel,
-      classModel.thisClass(), // Noncompliant {{Use `transformClass` overload without the class name.}}
+      classModel.thisClass(), // Noncompliant {{Use `transformClass` overload without the class name.}} [[quickfixes=qf1]]
 //    ^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


java-checks-test-sources/default/src/main/java/checks/ClassNameInClassTransformCheckSample.java
```diff
@@ -21,17 +21,21 @@ public byte[] transformClassFileNonCompliantDesc(Path path) throws IOException {
     ClassFile classFile = ClassFile.of();
     ClassModel classModel = classFile.parse(path);
     return classFile.transformClass(classModel,
-      classModel.thisClass(), // Noncompliant {{Use `transformClass` overload without the class name.}}
+      classModel.thisClass(), // Noncompliant {{Use `transformClass` overload without the class name.}} [[quickfixes=qf1]]
 //    ^^^^^^^^^^^^^^^^^^^^^^
+// fix@qf1 {{Remove second argument.}}
+// edit@qf1 [[sc=7;ec=29]] {{}}
```
erwan-serandour: If I am not mistaken, the comma is not removed by the quickfix (this would lead to ```transformClass(classModel, , classTransform)```). I would quickly check that it is not the case.

romainbrenguier: I just tested that in the IDE and you are right. I'll look for a fix.

PR #5230, opened 2025-07-02
SONARJAVA-5500 Implement rule S7479

romainbrenguier:


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S7479.html
```diff
@@ -0,0 +1,48 @@
+<h2>Why is this an issue?</h2>
```
erwan-serandour, line 1: le'ts merge https://github.com/SonarSource/rspec/pull/4962 to master

java-checks/src/main/java/org/sonar/java/checks/ClassBuilderWithMethodCheck.java
```diff
@@ -0,0 +1,86 @@
...
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.METHOD_INVOCATION);
+  }
+
+  /** Raise if last argument of withMethod is a lambda which only calls `MethodBuilder.withCode` */
+  @Override
+  public void visitNode(Tree tree) {
+    MethodInvocationTree invocation = (MethodInvocationTree) tree;
+    if (withMethod.matches(invocation)) {
+      assert !invocation.arguments().isEmpty() : "withMethod should have at least one argument";
```
erwan-serandour: I don't know what to think about the assert. If `invocation.arguments().isEmpty()` is true, we will get an index out of bound exception, the assert brings only moderate new information and I don't think assert is used inside the codebase.

PR #5236, opened 2025-07-04
Update external linters metadata

romainbrenguier:


dorian-burihabwa-sonarsource:
@romainbrenguier Can we get the version of the checkstyle from which the new mappings come from in the ticket?

erwan-serandour (approved):

PR #5246, opened 2025-07-07
SONARJAVA-5683 S2077 Fix FN on strings built with String.format()/formatted().

tomasz-tylenda-sonarsource:
[SONARJAVA-5683](https://sonarsource.atlassian.net/browse/SONARJAVA-5683)

[SONARJAVA-5683]: https://sonarsource.atlassian.net/browse/SONARJAVA-5683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/SQLInjectionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/SQLInjectionCheck.java
@@ -24,6 +24,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import org.sonar.check.Rule;
+import org.sonar.java.checks.helpers.QuickFixHelper;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
 import org.sonar.plugins.java.api.semantic.MethodMatchers;
@@ -112,6 +113,14 @@ public class SQLInjectionCheck extends IssuableSubscriptionVisitor {
       .withAnyParameters()
       .build());
 
+  private static final String JAVA_LANG_STRING = "java.lang.String";
+
+  private static final MethodMatchers FORMAT_METHODS = MethodMatchers.create()
+    .ofTypes(JAVA_LANG_STRING)
+    .names("format", "formatted")
+    .withAnyParameters()
+    .build();
+
   private static final String MAIN_MESSAGE = "Make sure using a dynamically formatted SQL query is safe here.";
 
   @Override
@@ -123,12 +132,12 @@ public class SQLInjectionCheck extends IssuableSubscriptionVisitor {
   public void visitNode(Tree tree) {
     if (anyMatch(tree)) {
       Optional<ExpressionTree> sqlStringArg = arguments(tree)
-        .filter(arg -> arg.symbolType().is("java.lang.String"))
+        .filter(arg -> arg.symbolType().is(JAVA_LANG_STRING))
         .findFirst();
 
       if (sqlStringArg.isPresent()) {
         ExpressionTree sqlArg = sqlStringArg.get();
-        if (isDynamicConcatenation(sqlArg)) {
+        if (isDynamicString(sqlArg)) {
           reportIssue(sqlArg, MAIN_MESSAGE);
         } else if (sqlArg.is(Tree.Kind.IDENTIFIER)) {
           IdentifierTree identifierTree = (IdentifierTree) sqlArg;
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/SQLInjectionCheck.java
```diff
@@ -197,7 +205,26 @@ private static boolean isDynamicPlusAssignment(ExpressionTree arg) {
...
+  }
+
+  private static boolean hasDynamicStringParameters(MethodInvocationTree mit) {
+    for (ExpressionTree arg: mit.arguments()) {
+      if (arg.symbolType().is(JAVA_LANG_STRING) && arg.asConstant().isEmpty()) {
+        return true;
+      }
+    }
+    return false;
+  }
```
alban-auzeill: Is it intentional not to support `String.format(String format, Object... args)` when `args` is not a `String`?

For example, if the argument is a `int`. Why does the rule not support the following case?
```
  public void formatIntVar(int input) throws SQLException {
    String query = String.format("SELECT %s", input);
    this.stmt.execute(query); // Noncompliant
  }

  public void formatIntConst() throws SQLException {
    String query = String.format("SELECT %s", 1);
    this.stmt.execute(query);
  }
```

Could be done with this kind of logic:

```suggestion
  private static boolean hasDynamicStringParameters(MethodInvocationTree mit) {
    boolean firstArg = true;
    for (ExpressionTree arg: mit.arguments()) {
      Type type = arg.symbolType();
      boolean notTheFirstLocaleArgument = !firstArg || !type.isUnknown() || !type.is("java.util.Locale");
      if (notTheFirstLocaleArgument && arg.asConstant().isEmpty()) {
        return true;
      }
      firstArg = false;
    }
    return false;
  }
```

tomasz-tylenda-sonarsource: I had to change it due to a failing test and then expanded it to exclude primitive arguments as well. Please see code comments for details.

java-checks/src/main/java/org/sonar/java/checks/SQLInjectionCheck.java
```diff
@@ -197,7 +206,35 @@ private static boolean isDynamicPlusAssignment(ExpressionTree arg) {
...
+   * Checks if parameters to format/formatted are dynamic variables susceptible to SQL injection.
+   */
+  private static boolean hasDynamicStringParameters(MethodInvocationTree mit) {
+    boolean firstArg = true;
+    for (ExpressionTree arg: mit.arguments()) {
+      Type type = arg.symbolType();
+      // `format` has a variant with Locale as the first argument - we do not need to check that parameter.
+      boolean isFirstLocaleArgument = firstArg && !type.isUnknown() && type.is("java.util.Locale");
+      // Primitives will not lead to SQL injection, so the code is compliant.
+      if (!isFirstLocaleArgument && !type.isPrimitive() && arg.asConstant().isEmpty()) {
```
alban-auzeill: We could also use isPrimitiveWrapper

java-checks/src/main/java/org/sonar/java/checks/SQLInjectionCheck.java
```diff
@@ -197,7 +206,35 @@ private static boolean isDynamicPlusAssignment(ExpressionTree arg) {
...
+
+  /**
+   * Checks if parameters to format/formatted are dynamic variables susceptible to SQL injection.
+   */
+  private static boolean hasDynamicStringParameters(MethodInvocationTree mit) {
+    boolean firstArg = true;
+    for (ExpressionTree arg: mit.arguments()) {
+      Type type = arg.symbolType();
+      // `format` has a variant with Locale as the first argument - we do not need to check that parameter.
+      boolean isFirstLocaleArgument = firstArg && !type.isUnknown() && type.is("java.util.Locale");
```
alban-auzeill: `&& !type.isUnknown()` seems useless, remove or invert the logic

alban-auzeill (changes requested):

java-checks/src/main/java/org/sonar/java/checks/SQLInjectionCheck.java
tomasz-tylenda-sonarsource: Moved and tested.

java-checks/src/main/java/org/sonar/java/checks/SQLInjectionCheck.java
tomasz-tylenda-sonarsource: Done.

PR #5269, opened 2025-07-30
SONARJAVA-5713 S1176 Support Markdown documentation comments

tomasz-tylenda-sonarsource:
[SONARJAVA-5713](https://sonarsource.atlassian.net/browse/SONARJAVA-5713)

[SONARJAVA-5713]: https://sonarsource.atlassian.net/browse/SONARJAVA-5713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -247,10 +250,11 @@ private static List<String> cleanLines(@Nullable String javadoc) {
       // Empty or malformed javadoc. for instance: '/**/'
       return Collections.emptyList();
     }
-    // remove start and end of Javadoc as well as stars
+    // remove start and end of Javadoc as well as stars (JavaDoc) and tripple slashes (MarkDown)
     String[] lines = trimmedJavadoc
       .substring(3, trimmedJavadoc.length() - 2)
```
alban-auzeill: In case of markdown comment, it does not end with "*/", so we should not remove the last two characters.
```
  /// abc
  /// def
```
The above comment will be cleaned by cleanLines(javadoc) where javadoc is:
```
javadoc = "/// abc\n  /// def";
```
And the current implementation will produce:
```
return [ "abc", "d"];
```

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -247,10 +250,11 @@ private static List<String> cleanLines(@Nullable String javadoc) {
...
       .replaceAll("(?m)^\\s*\\*", "")
+      .replaceAll("(?m)^\\s*///", "")
```
alban-auzeill: Asking the JVM to convert a String to a Pattern at each execution is not ideal, it is better to use constant Pattern fields. Furthermore, I would simplify the code by creating one regex to remove the unwanted prefix and suffix. And one regex to split the comment.
For example:
```
  private static final Pattern JAVADOC_CLEANUP_REGEX = Pattern.compile("^/\\*\\*\\s*+(?:\\*[ \t]*+)|\\s*+\\*/$");
  private static final Pattern JAVADOC_SPLIT = Pattern.compile("[ \t]*+\\R[ \t]*+(?:\\*[ \t]*+)");
  private static final Pattern MARKDOWN_CLEANUP_REGEX = Pattern.compile("^///[ \t]*+");
  private static final Pattern MARKDOWN_SPLIT = Pattern.compile("[ \t]*+\\R[ \t]*+(?:///[ \t]*+)");

  public static List<String> cleanLines(@Nullable String javadoc) {
    if (javadoc == null) {
      return Collections.emptyList();
    }
    boolean isJavaDoc = javadoc.startsWith("/**");
    Pattern cleanupRegex = isJavaDoc ? JAVADOC_CLEANUP_REGEX : MARKDOWN_CLEANUP_REGEX;
    Pattern splitRegex = isJavaDoc ? JAVADOC_SPLIT : MARKDOWN_SPLIT;
    return Arrays.asList(splitRegex.split(cleanupRegex.matcher(javadoc.trim()).replaceAll(""), -1));
  }
```
And we can have a test like:
```
  @Test
  void clear_lines() {
    assertThat(Javadoc.cleanLines("""
      /// Valid descriptions.
          /// @param value Valid descriptions."""))
      .containsExactly(
        "Valid descriptions.",
        "@param value Valid descriptions.");


    assertThat(Javadoc.cleanLines("""
      /**
        * Valid descriptions.
        * @param value Valid descriptions.
        */"""))
      .containsExactly(
        "Valid descriptions.",
        "@param value Valid descriptions.");

  }
```

alban-auzeill (changes requested):

java-checks/src/test/java/org/sonar/java/checks/UndocumentedApiCheckTest.java
```diff
@@ -46,6 +46,17 @@ void java_16() {
       .verifyIssues();
   }
 
+  @Test
+  void java_23() {
+    UndocumentedApiCheck check = new UndocumentedApiCheck();
+    assertThat(check.forClasses).isEqualTo("**.api.**");
+    assertThat(check.exclusion).isEqualTo("**.internal.**");
+    CheckVerifier.newVerifier()
+      .onFile("src/test/files/checks/UndocumentedApiCheck/UndocumentedApiJava23.java")
```
dorian-burihabwa-sonarsource: Consider using `nonCompilingTestSourcesPath`
```suggestion
      .onFile(nonCompilingTestSourcesPath("checks/UndocumentedApiCheck/UndocumentedApiJava23.java"))
```

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
tomasz-tylenda-sonarsource: Fixed.

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
tomasz-tylenda-sonarsource: Done.

I added tests and changed the code, but kept the original regexes and code structure, because it's hard to understand what is happening inside complex regulat expressions (I tried your snippet, but some tests were failing).

java-checks/src/test/java/org/sonar/java/checks/UndocumentedApiCheckTest.java
tomasz-tylenda-sonarsource: Done.

The new file compiles. All exisitng files were moved to `nonCompilingTestSourcesPath`, because they requireq multiple fixes.

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
```diff
@@ -238,22 +241,35 @@ private static String exceptionName(TypeTree typeTree) {
...
+  @VisibleForTesting
+  static List<String> cleanLines(@Nullable String javadoc) {
     if (javadoc == null) {
       return Collections.emptyList();
     }
-    String trimmedJavadoc = javadoc.trim();
-    if (trimmedJavadoc.length() <= 4) {
+
+    String clean = javadoc.trim();
+    if (clean.length() <= 4) {
```
dorian-burihabwa-sonarsource, line 255: This seems to work for empty markdown comments as well, but it might help to modify the comments on the line below to include that case.
```
// Empty or malformed javadoc. for instance: '/**/' or '///'
```

java-checks-test-sources/default/src/main/java/checks/api/undocumentedAPI/UndocumentedApiJava23.java
```diff
@@ -0,0 +1,61 @@
+package com.example.api.mypackage;
+
+/// Undocumented Api for Java 23 (and an example of bad documentation).
+public class UndocumentedApiJava23 {
+
+  public class BadNested {} // Noncompliant {{Document this public class by adding an explicit description.}}
+
+  ///  Documentated.
```
dorian-burihabwa-sonarsource: Should this say "Documented"?
```suggestion
  ///  Documented.
```

java-checks-test-sources/default/src/main/java/checks/api/undocumentedAPI/UndocumentedApiJava23.java
```diff
@@ -0,0 +1,61 @@
...
+  public class GoodNested {} // Compliant
+
+  public String badField; // Noncompliant {{Document this public field by adding an explicit description.}}
+
+  /// Documentated.
```
dorian-burihabwa-sonarsource: Should this read "Documented" as well?
```suggestion
  /// Documented.
```

java-checks-test-sources/default/src/main/java/checks/api/undocumentedAPI/UndocumentedApiJava23.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /// Valid descriptions.
+  /// @param value Valid descriptions.
+  /// @return some text
+  public String missingThrows(int value) throws NumberFormatException { // Noncompliant {{Document this method thrown exception(s): NumberFormatException}}
+    return "";
+  }
+
+  /// Documented, but not the type.
+  public class SomethingGenericBad<T> { // Noncompliant {{Document the parameter(s): <T>}}
```
dorian-burihabwa-sonarsource, line 54: Idea for an onboarding ticket: Should we change the message to say "Document the type parameter(s): <T>"?
```suggestion
  public class SomethingGenericBad<T> { // Noncompliant {{Document the type parameter(s): <T>}}
```

dorian-burihabwa-sonarsource (approved):
The changes look good to me 👍🏿 I left a couple of suggestions that you should feel free to include with this PR and a suggestion for an onboarding ticket

java-checks-test-sources/default/src/main/java/checks/api/undocumentedAPI/UndocumentedApiJava23.java
tomasz-tylenda-sonarsource: I also thought that types should have a separate message. It's out of scope for this PR, but we could fix it in future. Created https://sonarsource.atlassian.net/browse/SONARJAVA-5719

java-checks/src/main/java/org/sonar/java/checks/helpers/Javadoc.java
tomasz-tylenda-sonarsource: Done.

heruan:
Thanks for this! Will this be available on both the IntelliJ plugin and on SonarCloud?

tomasz-tylenda-sonarsource:
> Thanks for this! Will this be available on both the IntelliJ plugin and on SonarCloud?

Yes, both. It will reuquire 2 releases to reach the products: one for the Java analyzer, and another one for the product.

The rule (S1176) is not enabled by default, but since you reported the problem (thank you!), it must be enabled in your environment.

heruan:
> > Thanks for this! Will this be available on both the IntelliJ plugin and on SonarCloud?
> 
> Yes, both. It will reuquire 2 releases to reach the products: one for the Java analyzer, and another one for the product.

Got it. Any way I can test it early on my local IntelliJ?

> The rule (S1176) is not enabled by default, but since you reported the problem (thank you!), it must be enabled in your environment.

It is, indeed, in the company profile on SonarCloud. I guess the IntelliJ plugin uses the same profile/rules as in the connected SonarCloud project! But that's fine, we run it locally first so to avoid failing checks on PRs.

tomasz-tylenda-sonarsource:
> > > Thanks for this! Will this be available on both the IntelliJ plugin and on SonarCloud?
> > 
> > 
> > Yes, both. It will reuquire 2 releases to reach the products: one for the Java analyzer, and another one for the product.
> 
> Got it. Any way I can test it early on my local IntelliJ?
> 
> > The rule (S1176) is not enabled by default, but since you reported the problem (thank you!), it must be enabled in your environment.
> 
> It is, indeed, in the company profile on SonarCloud. I guess the IntelliJ plugin uses the same profile/rules as in the connected SonarCloud project! But that's fine, we run it locally first so to avoid failing checks on PRs.

Testing this change locally requires some work:

1. Build sonar-java (including this PR) using instructions from https://github.com/SonarSource/sonar-java/blob/master/README.md 
2. Install SonarQube locally. Community build works fine. You will need to configure a new project, and since this rule is not enabled by default, you’ll need to set up a custom quality profile as well.
3. Copy Java plugin that you build into the SonarQube (`cp ${SONAR_JAVA}/sonar-java-plugin/target/sonar-java-plugin-${VERSION}.jar ${SONAR_QUBE}/lib/extensions/`). The current version is 8.19.0-SNAPSHOT, SONAR-JAVA is the location of the sonar-java repo, and SONAR_QUBE is the path to the local SonarQube installation.
4. Delete old sonar-java-plugin in the SonarQube. If you forget to do this, SonarQube will not start. Restart SonarQube.

Now you have two ways of testing the change:

- Scan a test project with `mvn verify sonar:sonar …`
- Bind IntelliJ to your local SonarQube and test in IDE - this one is less reliable, because SonarQube for IDE running inside the IDE has to update to what is in your local SonarQube.

heruan:
Thanks, I'll give this a try! Hope it will land soon in products too 😄

alban-auzeill (approved):

heruan:
@tomasz-tylenda-sonarsource I was finally able to test this locally (it appears not to be available in products yet). The reported issue is indeed now fixed, thanks! I'm now experiencing another one related to Markdown JavaDoc:

~~~java
/// Runtime hints registrar for Fabric8 API.
public class Fabric8RuntimeHints implements RuntimeHintsRegistrar {
~~~

This should be valid JavaDoc, but triggers `java:S7476`:

> A single-line comment should start with exactly two slashes, no more.

Not sure where to open the issue, there's no issues here and I have no permission on Jira.

PR #5281, opened 2025-08-14
Stop using org.apache.commons.lang3.ArrayUtils

tomasz-tylenda-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/DiamondOperatorCheck.java
```diff
@@ -61,7 +60,15 @@ public class DiamondOperatorCheck extends SubscriptionVisitor implements JavaVer
...
   };
-  private static final Tree.Kind[] JAVA_8_KINDS = ArrayUtils.add(JAVA_7_KINDS, Tree.Kind.CONDITIONAL_EXPRESSION);
+
+  private static final Tree.Kind[] JAVA_8_KINDS = new Tree.Kind[] {
+    Tree.Kind.VARIABLE,
+    Tree.Kind.TYPE_CAST,
+    Tree.Kind.RETURN_STATEMENT,
+    Tree.Kind.ASSIGNMENT,
+    Tree.Kind.CONDITIONAL_EXPRESSION
+  };
```
dorian-burihabwa-sonarsource: I get the gist and limitations in the API but I feel like this is a little to error-prone (ie: we need to make sure our tests cover all cases otherwise we might update one list and not the other)

dorian-burihabwa-sonarsource: The test sources currently provide enough example to detect regression in one of the lists.
WDYT about moving `JAVA_8_KINDS` 's initialization (or overall declaration) in `isCompatibleWithJavaVersion` and explicitly copy the values from `JAVA_7_KINDS`?

tomasz-tylenda-sonarsource: I'm not sure if I follow. I'd rather focus on removing `ArrayUtils`. Do you also want to refactor `isCompatibleWithJavaVersion`? I find that method ugly, because it pretends to be a predicate, but has a side-effect. I tried refactoring, but everything I come up with is rather long.

Could you please take a look at a solution with Streams?

dorian-burihabwa-sonarsource (approved):

PR #5287, opened 2025-08-18
SONARJAVA-5743 Prepare next development iteration

dorian-burihabwa-sonarsource:


dorian-burihabwa-sonarsource:
Release 8.18.1 is not needed at this time

PR #5288, opened 2025-08-19
SONARJAVA-5745 Stop using org.apache.commons.lang3.StringUtils::countMatches

tomasz-tylenda-sonarsource:
[SONARJAVA-5745](https://sonarsource.atlassian.net/browse/SONARJAVA-5745)

[SONARJAVA-5745]: https://sonarsource.atlassian.net/browse/SONARJAVA-5745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/test/java/org/sonar/java/checks/helpers/StringUtilsTest.java
+++ b/java-checks/src/test/java/org/sonar/java/checks/helpers/StringUtilsTest.java
@@ -39,6 +39,7 @@ class StringUtilsTest {
     assertThat(StringUtils.countMatches("abababab", "cccc")).isZero();
     assertThat(StringUtils.countMatches("abababab", "ab")).isEqualTo(4);
 
+    assertThat(StringUtils.countMatches("abaTaba", "aba")).isEqualTo(2);
     assertThat(StringUtils.countMatches("abababa", "aba")).isEqualTo(2);
   }
 }


java-checks/src/test/java/org/sonar/java/checks/helpers/StringUtilsTest.java
```diff
@@ -0,0 +1,44 @@
...
+  void testCountMatches() {
+    assertThat(StringUtils.countMatches(null, "ab")).isZero();
+    assertThat(StringUtils.countMatches("", "ab")).isZero();
+    assertThat(StringUtils.countMatches("ab", null)).isZero();
+    assertThat(StringUtils.countMatches("ab", "")).isZero();
+
+    assertThat(StringUtils.countMatches("abababab", "cccc")).isZero();
+    assertThat(StringUtils.countMatches("abababab", "ab")).isEqualTo(4);
+
+    assertThat(StringUtils.countMatches("abababa", "aba")).isEqualTo(2);
```
dorian-burihabwa-sonarsource, line 43: So the contract here is that the sequences cannot overlap, correct?
For the sake of completeness, can we add another test where the repeated sequence is separated by a different pattern?
```suggestion
    assertThat(StringUtils.countMatches("abaTaba", "aba")).isEqualTo(2);
    assertThat(StringUtils.countMatches("abababa", "aba")).isEqualTo(2);
```

dorian-burihabwa-sonarsource (approved):
LGTM but I would an extra test case to make the contract more explicit in the tests

PR #5302, opened 2025-09-25
SONARJAVA-5797 False Negative with java:S2698 for org.testng

leonardo-pilastri-sonarsource:
[SONARJAVA-5797](https://sonarsource.atlassian.net/browse/SONARJAVA-5797)
[USER-1011](https://sonarsource.atlassian.net/browse/USER-1011)

Part of 

[USER-1011]: https://sonarsource.atlassian.net/browse/USER-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SONARJAVA-5797]: https://sonarsource.atlassian.net/browse/SONARJAVA-5797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/tests/AssertionsWithoutMessageCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/tests/AssertionsWithoutMessageCheck.java
@@ -105,18 +105,19 @@ public class AssertionsWithoutMessageCheck extends AbstractMethodDetection {
       checkFestLikeAssertion(mit, symbol, reportLocation);
     } else if (type.is("org.junit.jupiter.api.Assertions")) {
       checkJUnit5(mit, reportLocation);
-    } else if (mit.arguments().isEmpty() || expectedMessageArgIsNotString(mit, type) || isAssertingOnStringWithNoMessage(mit)) {
+    } else if (mit.arguments().isEmpty() || !hasMessageArg(mit, type) || isAssertingOnStringWithNoMessage(mit)) {
       reportIssue(reportLocation, MESSAGE);
     }
   }
 
-  private static boolean expectedMessageArgIsNotString(MethodInvocationTree mit, Type type) {
+  /**
+   * True if the call has a message argument. Such an argument is a string
+   * and it is the first of the last argument (depending on the assertion library).
+   */
+  private static boolean hasMessageArg(MethodInvocationTree mit, Type type) {
+    int expectedMessageArgIndex = (type.is("org.testng.Assert") || type.is("org.testng.AssertJUnit")) ? 1 : 0;
     List<ExpressionTree> args = mit.arguments();
-    int expectedMessageArgIndex = 0;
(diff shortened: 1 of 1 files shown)


java-checks-test-sources/default/src/test/java/checks/tests/AssertionsWithoutMessageCheckSample.java
```diff
@@ -14,6 +14,8 @@
 
 class AssertionsWithoutMessageCheckSample {
   void foo() {
+    org.testng.Assert.assertTrue(true); // Noncompliant
```
tomasz-tylenda-sonarsource, line 18: I changed it to `org.testng.Assert.assertTrue(true, "message"); // Noncompliant` and it still passes. I guess that's because the messages in testng are the last parameter.

java-checks/src/main/java/org/sonar/java/checks/tests/AssertionsWithoutMessageCheck.java
```diff
@@ -105,11 +105,20 @@ protected void onMethodInvocationFound(MethodInvocationTree mit) {
       checkFestLikeAssertion(mit, symbol, reportLocation);
     } else if (type.is("org.junit.jupiter.api.Assertions")) {
       checkJUnit5(mit, reportLocation);
-    } else if (mit.arguments().isEmpty() || !isString(mit.arguments().get(0)) || isAssertingOnStringWithNoMessage(mit)) {
+    } else if (mit.arguments().isEmpty() || expectedMessageArgIsNotString(mit, type) || isAssertingOnStringWithNoMessage(mit)) {
       reportIssue(reportLocation, MESSAGE);
     }
   }
 
+  private static boolean expectedMessageArgIsNotString(MethodInvocationTree mit, Type type) {
```
tomasz-tylenda-sonarsource: I tried to clean it up: I'd avoid negatives in method names and do `!` where it is callled, conditional expression, so that variables don't change, also checking `index < size` is more idiomatic IMO. Wdyt?

```
  /**
   * True if the call has a message argument. Such an argument is a string
   * and it is the first of the last argument (depending on the assertion library).
   */
  private static boolean hasMessageArg(MethodInvocationTree mit, Type type) {
    int expectedMessageArgIndex = (type.is("org.testng.Assert") || type.is("org.testng.AssertJUnit")) ? 1 : 0;
    List<ExpressionTree> args = mit.arguments();
    return expectedMessageArgIndex < args.size() && isString(args.get(expectedMessageArgIndex));
  }
```

tomasz-tylenda-sonarsource (approved):

tomasz-tylenda-sonarsource:
Let's use SONARJAVA ticket.

PR #5378, opened 2025-12-17
SONARJAVA-4960 False positive for S1854

rombirli:
-

### DIFF
--- a/java-checks-test-sources/default/src/main/files/non-compiling/checks/UnusedVariablesFPCheck.java
+++ b/java-checks-test-sources/default/src/main/files/non-compiling/checks/UnusedVariablesFPCheck.java
@@ -3,6 +3,8 @@ package checks;
 
 public class UnusedVariablesFPCheck {
   public class DeobfuscatedUpdateManager {
+    // @formatter:off
+// uncommenting the following code makes the issue disappear, as the semantic is fully resolved
 //    private interface DataContainer {
 //      Iterable<ItemElement> getItems();
 //    }
@@ -29,25 +31,8 @@ public class UnusedVariablesFPCheck {
 //    static class A {
 //      interface GenericCallback<T> { }
 //    }
+    // @formatter:on
 
-    /**
-     * Deobfuscated names with AI from :
-     * ```java
-     * package com;
-     * <p>
-     * public class BCid51 {
-     * void HJid232(EJid229 YHid199, Yid24.RIid217<MBid37> QJid241) {
-     * for (XFid148 RJid242 : YHid199.WIid222()) {
-     * int DHid178;
-     * if (Fid5.BGid151() == CGid152.DGid153) {
-     * MBid37 IHid183 = RJid242.OFid139();
-     * IHid183.AGid150();
-     * }
-     * }
-     * }
-     * }
-     * ```
-     */
     void processUpdates(
       DataContainer container
       // REMARK : the issue arises from the A.GenericCallback<ModelObject> callback that is not even used (indirect type resolution problem)
@@ -55,7 +40,7 @@ public class UnusedVariablesFPCheck {
     ) {
       for (ItemElement element : container.getItems()) {
         if (SystemConfig.getMode() == ConfigMode.ENABLED) {
-          ModelObject dataModel = element.getDataModel();
+          ModelObject dataModel = element.getDataModel(); // Compliant - false positive was raised here, dataModel is used in the next line
           dataModel.performAction();
         }
       }
@@ -64,17 +49,25 @@ public class UnusedVariablesFPCheck {
   }
 
   static class StringConcatenation {
+    // @formatter:off
+// uncommenting the following code makes the issue disappear, as the semantic is fully resolved
 //    private class AClass {
 //      private class BClass<T> {
 //        public T b;
 //      }
 //    }
+    // @formatter:on
 
     public String doSomething(AClass.BClass<String> instance) {
-      String c = "Hi"; // Rule S1854
+      String c = "Hi"; // Compliant - false positive was raised here, c is used in the next line
       return instance.b + c;
     }
   }
+
+/*
+  A user reported a FP on enhanced switch statements like the one below.
+  However I was not able to reproduce it in a minimal example.
+  https://community.sonarsource.com/t/false-positive-for-s1854-unused-assignments-should-be-removed/114110/12
 
   static class EnhancedSwitch {
 //    private enum DocumentStatus {
@@ -108,12 +101,5 @@ public class UnusedVariablesFPCheck {
       documentRepository.save(document);
     }
 
-  }
-
-  class Obvious {
-//    void obvious() {
-//      int i = 0; // doesn't raise issue
-//      i = 1; // raises issue
-//    }
-  }
+  }*/
 }
--- a/java-checks-test-sources/default/src/main/files/non-compiling/checks/UnusedVariablesFPCheck.java
+++ b/java-checks-test-sources/default/src/main/files/non-compiling/checks/UnusedVariablesFPCheck.java
@@ -1,4 +1,4 @@
-package com;
+package checks;
 
 
 public class UnusedVariablesFPCheck {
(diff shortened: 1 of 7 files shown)


java-checks/src/test/java/org/sonar/java/checks/DeadStoreCheckTest.java
```diff
@@ -51,6 +51,15 @@ void test_non_compiling() {
       .verifyIssues();
   }
 
+  @Test
+  void test_fp() {
```
tomasz-tylenda-sonarsource: Test name should be more precise.

java-frontend/src/main/java/org/sonar/java/model/Symbols.java
```diff
@@ -467,7 +472,7 @@ public boolean isIntersectionType() {
 
     @Override
     public Type[] getIntersectionTypes() {
-      return new Type[] { this };
```
tomasz-tylenda-sonarsource, line 470: Let's avoid reformatting unrelated code.

java-checks-test-sources/default/src/main/files/non-compiling/checks/UnusedVariablesFPCheck.java
```diff
@@ -0,0 +1,119 @@
...
+//    }
+
+    public String doSomething(AClass.BClass<String> instance) {
+      String c = "Hi"; // Rule S1854
+      return instance.b + c;
+    }
+  }
+
+  static class EnhancedSwitch {
+//    private enum DocumentStatus {
```
tomasz-tylenda-sonarsource, line 73: This is commented out, so that the parser does not generate semantics, right? Could you please add a comment about it?

java-checks-test-sources/default/src/main/files/non-compiling/checks/UnusedVariablesFPCheck.java
```diff
@@ -0,0 +1,119 @@
...
+
+  static class StringConcatenation {
+//    private class AClass {
+//      private class BClass<T> {
+//        public T b;
+//      }
+//    }
+
+    public String doSomething(AClass.BClass<String> instance) {
+      String c = "Hi"; // Rule S1854
```
tomasz-tylenda-sonarsource: This is the place where we had FP, right? We typically use `// Compliant` comment in such case. Could you please add this in all places where FPs where happening before the change.

java-frontend/src/test/java/org/sonar/java/model/SymbolsTest.java
```diff
@@ -171,7 +171,6 @@ private static void assertCommonProperties(Symbol unknownSymbol) {
     assertThat(unknownSymbol.isPackageSymbol()).isFalse();
     assertThat(unknownSymbol.isTypeSymbol()).isFalse();
     assertThat(unknownSymbol.isVariableSymbol()).isFalse();
-    assertThat(unknownSymbol.isMethodSymbol()).isFalse();
```
tomasz-tylenda-sonarsource, line 174: I didn't read this carefully, but I'm wondering why this is removed instead of reverting the condition.

java-frontend/src/main/java/org/sonar/java/model/Symbols.java
rombirli: rolled it back 👍

java-checks-test-sources/default/src/main/files/non-compiling/checks/UnusedVariablesFPCheck.java
rombirli: I cleaned a bit this whole reproducer file (including these lines, and explained all commented out code), could you re-review this file specifically please ?

java-checks-test-sources/default/src/main/files/non-compiling/checks/UnusedVariablesFPCheck.java
rombirli: Applied your suggestion 👍

java-checks/src/test/java/org/sonar/java/checks/DeadStoreCheckTest.java
rombirli: I proposed `test_incomplete_semantic` as a more precise name in my last commit

java-frontend/src/test/java/org/sonar/java/model/SymbolsTest.java
rombirli: I added it back

rombirli: Edit : 
this line 174 is inside of function `private static void assertCommonProperties(Symbol unknownSymbol)` which is used by all kinds of unknown symbols tests (`unknown_type_symbol`, `unknown_method_symbol`, ...), the `isMethodSymbol` function is no longer final in the parent abstact class `UnknownSymbol` , the result can be overriden differently in each child class (`UnkownTypeSymbol`, `UnknownMethodSymbol`).

java-frontend/src/main/java/org/sonar/java/model/Symbols.java
```diff
@@ -340,6 +340,11 @@ public boolean isVarArgsMethod() {
     public boolean isNativeMethod() {
       return false;
     }
+
+    @Override
+    public boolean isMethodSymbol() {
+      return true;
```
tomasz-tylenda-sonarsource, line 346: I flipped it to `false` and re-run unit tests (`SymbolsTest`) and everything passed. Could we unit test it?

java-checks/src/test/java/org/sonar/java/checks/DeadStoreCheckTest.java
```diff
@@ -51,6 +51,15 @@ void test_non_compiling() {
       .verifyIssues();
   }
 
+  @Test
+  void test_incomplete_semantic() {
+    CheckVerifier.newVerifier()
+      .onFile(TestUtils.nonCompilingTestSourcesPath("checks/UnusedVariablesFPCheck.java"))
+      .withJavaVersion(14)
```
tomasz-tylenda-sonarsource: What's the reason for Java 14 here?

rombirli: No reason, removing it 👍

java-frontend/src/main/java/org/sonar/java/model/Symbols.java
rombirli: Adding a test, however flipping it back to false makes `DeadStoreCheckTest.test_incomplete_semantic` fail

rombirli:
Addressed both issues in https://github.com/SonarSource/sonar-java/pull/5378/commits/d78ed6bafb7517e6e8ee924ddf578ab300c635b1

tomasz-tylenda-sonarsource:
Is this ready for re-review?

rombirli:
> Is this ready for re-review?

Yes it is

java-frontend/src/test/java/org/sonar/java/model/SymbolsTest.java
```diff
@@ -171,7 +171,7 @@ private static void assertCommonProperties(Symbol unknownSymbol) {
...
+    assertThat(unknownSymbol.isMethodSymbol()).isEqualTo(unknownSymbol instanceof Symbols.UnknownMethodSymbol);
```
tomasz-tylenda-sonarsource: We should not have logic in tests. Each test that calls `assertCommonProperties` should have an assertion with a constant, either `true` or `false`. After all, if this property changes, it is not a _common_ property.

Reference: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

tomasz-tylenda-sonarsource (commented):
Let's use dogfooding on peachee to test this PR. If you haven't used it before, it's a good thing to learn. We can talk about it offline.

tomasz-tylenda-sonarsource (commented):
LGTM. I see that my last comment was resolved. Since @aurelien-coet-sonarsource wanted to review this change as well, please wait for his approval.

aurelien-coet-sonarsource (approved):

PR #5475, opened 2026-02-19
SONARJAVA-6141 Save ncloc metric on test files

henryju:
Enable saving the NCLOC metric for test files in TestFileMeasurer, skipped in SonarLint context, consistent with main file behavior.

Part of

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/Measurer.java
+++ b/java-frontend/src/main/java/org/sonar/java/Measurer.java
@@ -64,8 +64,11 @@ public class Measurer extends SubscriptionVisitor {
     @Override
     public void scanFile(JavaFileScannerContext context) {
       sonarFile = context.getInputFile();
-      var metricsComputer = ((MetricsScannerContext)context).getMetricsComputer();
+      var metricsComputer = ((MetricsScannerContext) context).getMetricsComputer();
       noSonarFilter.noSonarInFile(sonarFile, metricsComputer.getNoSonarLines(context.getTree()));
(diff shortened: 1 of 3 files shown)


java-frontend/src/main/java/org/sonar/java/Measurer.java
```diff
@@ -64,8 +64,11 @@ public class TestFileMeasurer implements JavaFileScanner {
     @Override
     public void scanFile(JavaFileScannerContext context) {
       sonarFile = context.getInputFile();
-      var metricsComputer = ((MetricsScannerContext)context).getMetricsComputer();
+      var metricsComputer = ((MetricsScannerContext) context).getMetricsComputer();
       noSonarFilter.noSonarInFile(sonarFile, metricsComputer.getNoSonarLines(context.getTree()));
+      if (!isSonarLintContext()) {
+        saveMetricOnFile(CoreMetrics.NCLOC, metricsComputer.getLinesOfCode(context.getTree()));
+      }
```
dorian-burihabwa-sonarsource: Nitpick, can we flip the test to avoid negation and have a similar structure to the one in the `Measurer.scanFile`?
```suggestion
      if (isSonarLintContext()) {
        return;
      }
      saveMetricOnFile(CoreMetrics.NCLOC, metricsComputer.getLinesOfCode(context.getTree()));
```

dorian-burihabwa-sonarsource (approved):
LGTM but I have think we can make a small style improvement here

PR #5553, opened 2026-04-07
SONARJAVA-6234 Autoscan check filter silently removes custom file scanners

kaufco:


leonardo-pilastri-sonarsource (approved):
LGTM, the change only affects the UnaryOperator that we create in the JavaSensor, which is only used a few lines later to provide the JavaFrontend with a set of checks, namely all the registered checks in SonarComponents (also the custom ones registered through our CheckRegistrar component) that also pass the updated filter.

I verified that all implementations of JavaCheck we own that do not have a RuleKey are provided to the frontend components in other ways, and do not go through this updated filter.

PR #5563, opened 2026-04-10
SONARJAVA-6247 Use orchestrator-cache in autoscan tests

tomasz-tylenda-sonarsource:
* We forgot autoscan when adding caching to plugin and ruling qa.
* Tested: https://github.com/SonarSource/sonar-java/actions/runs/24245885636/job/70792206325 downloads `sonarqube-enterprise-lw-2026.2.1.121354.zip` which was previousy caches by other qa jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -342,7 +342,7 @@ jobs:
         run: >
           mvn clean package --batch-mode --errors --show-version
           --activate-profiles it-autoscan
-          "-Dsonar.runtimeVersion=$SQ_VERSION"
+          -Dsonar.runtimeVersion="$SQ_VERSION"
           -Dmaven.test.redirectTestOutputToFile=false
           -Dparallel=methods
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -337,7 +342,7 @@ jobs:
         run: >
           mvn clean package --batch-mode --errors --show-version
           --activate-profiles it-autoscan
-          -Dsonar.runtimeVersion=LATEST_RELEASE
+          "-Dsonar.runtimeVersion=$SQ_VERSION"
```
dorian-burihabwa-sonarsource:
```suggestion
          -Dsonar.runtimeVersion="$SQ_VERSION"
```

dorian-burihabwa-sonarsource (commented):
LGTM but I think it would be nicer if we moved the starting double a bit further

dorian-burihabwa-sonarsource (approved):
Thanks for the change!

PR #5566, opened 2026-04-13
SONARJAVA-6251 Reduce test log noise by restricting debug to selected packages

tomasz-tylenda-sonarsource:
* Logs are used to test telemetry. `-X` implied by `setDebugLogs(true)` is too verbose - it is enough to use debug only on our own package.
* Reduction 50k lines -> 14k lines in Plugin QA LATEST_RELEASE.
* Also tested with DEV sq: https://github.com/SonarSource/sonar-java/actions/runs/24337680858


asya-vorobeva (approved):
👍

PR #5570, opened 2026-04-14
SONARJAVA-6208 S2699: Add approve to assertion method name pattern

NoemieBenard:
Add approve keyword to the assertion methods pattern to support ApproveJ library

### DIFF
--- a/java-checks-test-sources/default/pom.xml
+++ b/java-checks-test-sources/default/pom.xml
@@ -1016,6 +1016,18 @@
       <artifactId>junit-platform-suite</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.approvej</groupId>
+      <artifactId>core</artifactId>
+      <version>1.6.0</version>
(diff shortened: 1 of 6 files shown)


tomasz-tylenda-sonarsource (commented):
Overall this change looks good, but we should also update the documentation in RSpec. Let me know when you have time and I'll show you how to do this.

java-checks-test-sources/default/pom.xml
```diff
@@ -1016,6 +1016,12 @@
       <artifactId>junit-platform-suite</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.approvej</groupId>
+      <artifactId>core</artifactId>
+      <version>1.6.0</version>
+    <scope>test</scope>
+</dependency>
```
tomasz-tylenda-sonarsource: Please fix formatting.

tomasz-tylenda-sonarsource (approved):

PR #5571, opened 2026-04-14
SONARJAVA-6254 Use mise to install mvn in PrepareNextIteration.yml

tomasz-tylenda-sonarsource:
Recent migration from GH runners to sonar runners broke the workflow, because our runners do not have `mvn`.

### DIFF
--- a/.github/workflows/PrepareNextIteration.yml
+++ b/.github/workflows/PrepareNextIteration.yml
@@ -20,6 +20,10 @@ jobs:
       - name: Checkout Sources
         uses: actions/checkout@v4
 
+      - uses: jdx/mise-action@5ac50f778e26fac95da98d50503682459e86d566 # v3.2.0
+        with:
+          version: 2025.7.12
+
       - name: Normalize Version
         id: version
         env:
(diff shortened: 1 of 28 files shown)


.github/workflows/PrepareNextIteration.yml
```diff
@@ -20,6 +20,10 @@ jobs:
       - name: Checkout Sources
         uses: actions/checkout@v4
 
+      - uses: jdx/mise-action@5ac50f778e26fac95da98d50503682459e86d566 # v3.2.0
```
dorian-burihabwa-sonarsource: Is there a reason we do not target a greater version of mise-action like [3.6.3](https://github.com/jdx/mise-action/releases/tag/v3.6.3)?
```suggestion
      - uses: jdx/mise-action@5228313ee0372e111a38da051671ca30fc5a96db # v3.6.3
```

dorian-burihabwa-sonarsource (commented):
Thanks for the fix. Could we update to a greater version of mise-action?

dorian-burihabwa-sonarsource (approved):

PR #5575, opened 2026-04-16
SONARJAVA-5719 S1176 Should have a separate message for undocumented type parameters.

NoemieBenard:
Add separate message for type parameters in rule S1176.

### DIFF
--- a/java-checks-test-sources/default/src/main/files/non-compiling/checks/api/undocumentedAPI/UndocumentedApi.java
+++ b/java-checks-test-sources/default/src/main/files/non-compiling/checks/api/undocumentedAPI/UndocumentedApi.java
@@ -331,6 +331,13 @@ class A {
 
 }
 
+/**
+ * Description.
+ */
+public record MyRecord<U>(U a, int b) { // Noncompliant {{Document the type parameter(s): <U>}}
+  //          ^^^^^^^^
+}
+
 @interface MyAnnotation {}
 
 class UsesVisibleForTesting {
--- a/java-checks-test-sources/default/src/main/files/non-compiling/checks/api/undocumentedAPI/UndocumentedApi.java
+++ b/java-checks-test-sources/default/src/main/files/non-compiling/checks/api/undocumentedAPI/UndocumentedApi.java
@@ -67,7 +67,7 @@ public class AClass { // Noncompliant {{Document this public class by adding an 
 /**
  * This is a Javadoc comment
  */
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/UndocumentedApiCheck.java
```diff
@@ -129,7 +129,8 @@ private void visitNode(Tree tree, Tree reportTree, SymbolMetadata symbolMetadata
       } else {
         Set<String> undocumentedParameters = javadoc.undocumentedParameters();
         if (!undocumentedParameters.isEmpty()) {
-          context.reportIssue(this, reportTree, "Document the parameter(s): " + undocumentedParameters.stream().collect(Collectors.joining(", ")));
+          String label = getParamLabel(tree);
+          context.reportIssue(this, reportTree, "Document the " + label + undocumentedParameters.stream().collect(Collectors.joining(", ")));
```
tomasz-tylenda-sonarsource: We can try to improve existing code: `String.join(", ", undocumentedParameters)` should do the same as the long expression with a stream.

java-checks-test-sources/default/src/main/files/non-compiling/checks/api/undocumentedAPI/UndocumentedApi.java
```diff
@@ -67,7 +67,7 @@ public A() { // Noncompliant {{Document this public constructor by adding an exp
 /**
  * This is a Javadoc comment
  */
-public class MyClass<T> implements Runnable { // Noncompliant {{Document the parameter(s): <T>}}
+public class MyClass<T> implements Runnable { // Noncompliant {{Document the type parameter(s): <T>}}
```
tomasz-tylenda-sonarsource, line 70: How about something like this?
```
  /**
   * Description.
   */
  public record MyRecord<U>(U a, int b) {
  }
```

tomasz-tylenda-sonarsource: We should add such a test I think. In addtion to that, Let's file a ticket that this rule should also flag record fields that do not have a description (`b` in the example will not trigger an issue).

tomasz-tylenda-sonarsource (commented):
Do we have a test for a record with fields and type parameters. If not, then let's add it. See the comment.

tomasz-tylenda-sonarsource (approved):

PR #5576, opened 2026-04-16
SONARJAVA-6198 S1451 should accept an empty headerFormat as the absence of any line

NoemieBenard:
Modify S1451 to recognize an empty headerFormat as the absence of any line.

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/ClassBlankLine.java
+++ b/java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/ClassBlankLine.java
@@ -1,7 +1,7 @@
 
 package checks.FileHeaderCheck;
 
-public class Class4 {
+public class ClassBlankLine {
 }
 // Compliant
 
--- a/java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/ClassNoBlankLine.java
+++ b/java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/ClassNoBlankLine.java
@@ -1,5 +1,5 @@
 package checks.FileHeaderCheck;
 
-public class Class5 {
+public class ClassNoBlankLine {
 }
 // Compliant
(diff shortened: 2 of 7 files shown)


java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/ClassBlankLine.java
tomasz-tylenda-sonarsource, line 1: I see that the existing samples do not use descriptive names, but I think this should not stop us here. How about a name like `ClassBlankLine.java` or something similar?

java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/ClassNoBlankLine.java
tomasz-tylenda-sonarsource, line 1: `ClassNoHeader.java` ?

java-checks/src/main/java/org/sonar/java/checks/FileHeaderCheck.java
```diff
@@ -67,14 +67,18 @@ public void setContext(JavaFileScannerContext context) {
...
+        expectedLines = headerFormat.split("(?:\r)?\n|\r");
+      }
     }
     visitFile();
   }
 
   private String getHeaderFormat() {
     String format = headerFormat;
-    if(format.charAt(0) != '^') {
+    if (format.charAt(0) != '^') {
```
tomasz-tylenda-sonarsource, line 81: Let's avoid reformatting unchanged code.

java-checks/src/main/java/org/sonar/java/checks/FileHeaderCheck.java
```diff
@@ -67,14 +67,18 @@ public void setContext(JavaFileScannerContext context) {
         }
       }
     } else {
-      expectedLines = headerFormat.split("(?:\r)?\n|\r");
+      if (headerFormat.isEmpty()) {
```
tomasz-tylenda-sonarsource, line 70: Conditional expression is IMO more readable:
```
expectedLines = headerFormat.isEmpty())
    ? new String[] {}
    :  headerFormat.split("(?:\r)?\n|\r");
``

NoemieBenard: I ended up handling the empty case before as it is also necessary to handle it when headerFormat is a regular expression (see #5577)

PR #5578, opened 2026-04-17
SONARJAVA-6196 S1451 should provide a default template for headers

NoemieBenard:
Add default template for headerFormat in S1451.


java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/Class4.java
tomasz-tylenda-sonarsource, line 1: Let's use desciprive file names for test samples.

java-checks/src/main/java/org/sonar/java/checks/FileHeaderCheck.java
```diff
@@ -58,23 +65,28 @@ public List<Tree.Kind> nodesToVisit() {
...
-    } else {
-      expectedLines = headerFormat.split("(?:\r)?\n|\r");
     }
     visitFile();
   }
 
   private String getHeaderFormat() {
     String format = headerFormat;
-    if(format.charAt(0) != '^') {
+    if (format.charAt(0) != '^') {
```
tomasz-tylenda-sonarsource, line 89: IMO it's a bad practice to do unintentional chages, because they make it hard to understand what is happening, if someone is reading a PR for some reason in the future.

java-checks/src/test/java/org/sonar/java/checks/FileHeaderCheckTest.java
```diff
@@ -98,6 +98,26 @@ void test() {
       .onFile(mainCodeSourcesPath("checks/FileHeaderCheck/Class3.java"))
       .withCheck(check)
       .verifyNoIssues();
+
```
tomasz-tylenda-sonarsource, line 101: Are these a part of this PR or is it a copied from the other changes?

java-checks/src/main/java/org/sonar/java/checks/FileHeaderCheck.java
tomasz-tylenda-sonarsource, line 57: Let's remove `expectedLines`.

tomasz-tylenda-sonarsource, line 95: Let's remove this method.

tomasz-tylenda-sonarsource, line 99: This should be extracted to `checkExpectedLines`

tomasz-tylenda-sonarsource (approved):
LGTM. I think we should take a look at the experience in local SQ to see how changing this rule will affect the users.

PR #5579, opened 2026-04-20
SONARJAVA-6268 Fix issue in `primitveType()` and `primitiveWrapperType()` when the sema is incomplete

GabrielFleischer:


alban-auzeill:
Good fix. The null check before passing the result of `sema.resolveType(name)` to `sema.type(...)` is the right defensive move. Even though in practice ECJ almost always returns `UnknownType` rather than `null`, the incomplete-classpath edge case is real enough to warrant this guard — a NPE in `primitiveType()` or `primitiveWrapperType()` would be a confusing failure with no obvious cause.

The test approach is pragmatic: since reproducing a half-initialized ECJ semantic model in an integration test is essentially infeasible, using a spy to simulate the `resolveType` returning `null` is the cleanest way to verify the contract. The test comments explaining *why* the scenario is contrived are a nice touch.

The rename of `names()` / `@MethodSource("names")` to `fullyQualifiedNamesToNames()` is also a welcome cleanup — the previous name was a collision waiting to confuse.

alban-auzeill (approved):

GabrielFleischer:
> Good fix. The null check before passing the result of `sema.resolveType(name)` to `sema.type(...)` is the right defensive move. Even though in practice ECJ almost always returns `UnknownType` rather than `null`, the incomplete-classpath edge case is real enough to warrant this guard — a NPE in `primitiveType()` or `primitiveWrapperType()` would be a confusing failure with no obvious cause.
> 
> The test approach is pragmatic: since reproducing a half-initialized ECJ semantic model in an integration test is essentially infeasible, using a spy to simulate the `resolveType` returning `null` is the cleanest way to verify the contract. The test comments explaining _why_ the scenario is contrived are a nice touch.
> 
> The rename of `names()` / `@MethodSource("names")` to `fullyQualifiedNamesToNames()` is also a welcome cleanup — the previous name was a collision waiting to confuse.

Thanks ~~Claude~~Alban

PR #5580, opened 2026-04-20
SONARJAVA-6269 S1451 should correctly handle empty headerFormat

NoemieBenard:
Refactor FileHeaderCheck to handle empty headers correctly.

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/ClassDefaultHeader.java
+++ b/java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/ClassDefaultHeader.java
@@ -6,6 +6,6 @@
  */
 package checks.FileHeaderCheck;
 
-public class Class6 {
+public class ClassDefaultHeader {
 }
 // Compliant (default header)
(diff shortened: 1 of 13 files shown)


java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/ClassDefaultHeader.java
tomasz-tylenda-sonarsource, line 1: `ClassDefaultHeader.java`?

java-checks-test-sources/default/src/main/java/checks/FileHeaderCheck/RegexBlankLine.java
tomasz-tylenda-sonarsource, line 1: Do you really need this file and `RegexNoBlankLine.java`? Why not reuse `ClassBlankLine.java‎`?

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaSensorTest.java
```diff
@@ -429,8 +429,6 @@ void do_not_filter_checks_when_no_autoscan() throws IOException {
         "CustomRepository:CustomMainCheck",
         "CustomRepository:CustomJspCheck",
         "CustomRepository:CustomTestCheck",
-        // not in SonarWay (FileHeaderCheck)
```
tomasz-tylenda-sonarsource, line 432: Does this test still need a fix? I could not reproduce it locally. If it does, could we find a different rule, not in SonarWay, which fails? Maybe S1166.

We can discuss it offline.

tomasz-tylenda-sonarsource (approved):

PR #5581, opened 2026-04-23
SONARJAVA-6273 S3706: "stream" should not be used for Collection "forEach" calls

NoemieBenard:
Implementation of rule S3706.


java-checks/src/main/java/org/sonar/java/checks/StreamForeachCheck.java
```diff
@@ -0,0 +1,65 @@
...
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S3706")
+public class StreamForeachCheck extends IssuableSubscriptionVisitor {
+
+  private static final MethodMatchers STREAM_METHOD = MethodMatchers.create()
+    .ofSubTypes("java.util.Collection")
+    .names("stream")
+    .withAnyParameters()
```
tomasz-tylenda-sonarsource: Since `stream()` doesn't take any parameters, could we make this match stricter?

java-checks/src/main/java/org/sonar/java/checks/StreamForeachCheck.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodInvocationTree mit = (MethodInvocationTree) tree;
+    if (STREAM_FOREACH_METHOD.matches(mit)) {
```
tomasz-tylenda-sonarsource: `tree instanceof MethodInvocationTree mit && STREAM_FOREACH_METHOD.matches(mit)` looks nicer IMO.

java-checks/src/main/java/org/sonar/java/checks/StreamForeachCheck.java
```diff
@@ -0,0 +1,65 @@
...
+      checkUnnecessaryForEach(mit);
+    }
+  }
+
+  private void checkUnnecessaryForEach(MethodInvocationTree mit) {
+    ExpressionTree methodSelect = mit.methodSelect();
```
tomasz-tylenda-sonarsource: You can try something along these lines:
```
    if (mitForEach.methodSelect() instanceof MemberSelectExpressionTree mset
      && mset.expression() instanceof MethodInvocationTree mitStream
      && STREAM_METHOD.matches(mitStream)) {
        reportIssue(mitStream, "Replace unnecessary call to .stream().forEach() by .forEach()");
      }
    }
```
Also, we can call strings `streamMit` and `forEachMit` or something like this.

java-checks/src/test/java/org/sonar/java/checks/StreamForeachCheckTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+class StreamForeachCheckTest {
+  @Test
+  void test() {
+    StreamForeachCheck check = new StreamForeachCheck();
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/StreamForeachCheck.java"))
+      .withCheck(check)
+      .verifyIssues();
+  }
```
tomasz-tylenda-sonarsource, line 42: Let's also add this one:
```
  @Test
  void test_without_semantic() {
    StreamForeachCheck check = new StreamForeachCheck();
    CheckVerifier.newVerifier()
      .onFile(mainCodeSourcesPath("checks/StreamForeachCheck.java"))
      .withoutSemantic()
      .withCheck(check)
      .verifyIssues();
  }
```

java-checks-test-sources/default/src/main/java/checks/StreamForeachCheck.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Stream;
+
+public class StreamForeachCheck {
+
+  void unnecessaryStreamForEach(Collection<Integer> collection) {
+    collection.stream().forEach(System.out::println); // Noncompliant
```
tomasz-tylenda-sonarsource: Let's use `^^^^` to assert the location of the problem exactly and let's also use `// Noncompliant {{Replace unnecessary (...)}}` to assert the message. It doesn't have to be everywhere, just once is enough.

java-checks/src/test/java/org/sonar/java/checks/StreamForeachCheckTest.java
```diff
@@ -27,6 +27,7 @@ void test() {
     StreamForeachCheck check = new StreamForeachCheck();
     CheckVerifier.newVerifier()
       .onFile(mainCodeSourcesPath("checks/StreamForeachCheck.java"))
+      .withoutSemantic()
```
tomasz-tylenda-sonarsource, line 39: I meant, we should have two tests, one with semantic, one without.

java-checks-test-sources/default/src/main/java/checks/StreamForeachCheck.java
```diff
@@ -8,7 +8,8 @@
 public class StreamForeachCheck {
 
   void unnecessaryStreamForEach(Collection<Integer> collection) {
-    collection.stream().forEach(System.out::println); // Noncompliant
+    collection.stream().forEach(System.out::println); // Noncompliant {{Replace unnecessary call to .stream().forEach() by .forEach()}}
+  //^^^^^^^^^^^^^^^^^^^
```
tomasz-tylenda-sonarsource: If we are pointing at `collection.stream()` then maybe the message should be `Remove unnecessary call to .stream()` ? It's possible to tweak the location of the issue quite freely with some adjustments to report issue.

Btw, I told you that quckfixes are difficult. This is generally true, but in this case it's not too difficult, so maybe we could do it in a follow-up PR later.

tomasz-tylenda-sonarsource (approved):

PR #5584, opened 2026-04-27
SONARJAVA-6283 S3706 Implement quickfix

NoemieBenard:
Implement quick fix for S3706.

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/StreamForeachCheck.java
+++ b/java-checks-test-sources/default/src/main/java/checks/StreamForeachCheck.java
@@ -37,4 +37,17 @@ public class StreamForeachCheck {
     s.forEach(System.out::println); // Compliant (the rule ignores this case)
   }
 
+  // fix@qf1 {{Remove .stream() call}}
+  // edit@qf1 [[sc=15;ec=24]] {{}}
+  void quickFix(Collection<Integer> collection) {
+    collection.stream().forEach(System.out::println); // Noncompliant [[quickfixes=qf1]]
+    //         ^^^^^^^^^^^^^^^^
+  }
+
+  // fix@qf2 {{Remove .stream() call}}
+  // edit@qf2 [[sc=8;ec=17]] {{}}
+  void quickFix2(Set<String> set) {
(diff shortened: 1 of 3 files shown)


java-checks-test-sources/default/src/main/java/checks/StreamForeachCheck.java
```diff
@@ -37,4 +37,17 @@ void sequentialStreamForEach(Collection<String> col) {
     s.forEach(System.out::println); // Compliant (the rule ignores this case)
   }
 
+  // fix@qf1 {{Remove .stream() call}}
```
tomasz-tylenda-sonarsource: `Remove the call to stream()` sound better imo. I'm not sure if we quote code in this context (`Remove the call to "stream()"`?)

java-checks/src/main/java/org/sonar/java/checks/StreamForeachCheck.java
```diff
@@ -56,7 +59,19 @@ private void checkUnnecessaryForEach(MethodInvocationTree mitForEach) {
       && msetForEach.expression() instanceof MethodInvocationTree mitStream
       && STREAM_METHOD.matches(mitStream)
       && mitStream.methodSelect() instanceof MemberSelectExpressionTree msetStream) {
-      reportIssue(msetStream.identifier(), msetForEach.identifier(), "Simplify the code by replacing .stream().forEach() with .forEach().");
+      QuickFixHelper.newIssue(context)
+        .forRule(this)
+        .onRange(msetStream.identifier(), msetForEach.identifier())
+        .withMessage("Simplify the code by replacing .stream().forEach() with .forEach().")
+        .withQuickFixes(() -> computeQuickFix(msetStream, mitStream))
```
tomasz-tylenda-sonarsource: There is also `withQuickFix`, which does not need a List. Let's use it if possible.

tomasz-tylenda-sonarsource (commented):
The title of the PR should be something better for a Git commit title. How about this? `SONARJAVA-6283 S3706: implement quickfix`

tomasz-tylenda-sonarsource (approved):

PR #5586, opened 2026-04-28
SONARJAVA-6291 SubmitReview: Use Vault token

pavel-mikula-sonarsource:
With the latest automation changes, we need the Vault-based token now. It's the same token as the one in RequestReview.yml file. Please take care of merging this, I have 200+ repos to update.


rombirli (approved):
LGTM

PR #5591, opened 2026-04-30
Add Java path traversal detection for archive and derived paths

FORIMOC:
# Add a Java path traversal rule for archive-derived, request-derived, and resource-lookup paths

## Summary

This change adds a new Java security rule to improve path traversal detection in three representative situations:

- archive-entry-derived paths, such as `ZipEntry.getName()` or `JarEntry.getName()`
- request-derived paths that are propagated through one path-construction step
- template or resource names that later reach `ClassLoader.getResource(...)`

The new rule is implemented as `ArchiveEntryPathTraversalCheck` and is registered with rule key `S7099`.

## Motivation

Several real vulnerability patterns share the same core shape: attacker-controlled path components are propagated into file-system paths or resource lookups without sufficient validation.

Representative examples include:

- `CVE-2022-4494`
- `CVE-2022-39367`
- `CVE-2022-31194`
- `CVE-2022-29253`

One representative archive-entry case looks like this:

```java
while ((zipEntry = zipInputStream.getNextEntry()) != null) {
    final File destFile = new File(importSandboxDirectory, zipEntry.getName());
    ServiceUtilities.ensureFileCreated(destFile);
    final FileOutputStream destOutputStream = new FileOutputStream(destFile);
}
```

A representative request-derived case looks like this:

```java
String resumableIdentifier = request.getParameter("resumableIdentifier");
tempDir = tempDir + File.separator + resumableIdentifier;
File fileDir = new File(tempDir);
fileDir.mkdir();
```

A representative resource-lookup case looks like this:

```java
String templatePath = suffixPath + templateName;
return classloader.getResource(templatePath);
```

## What this rule detects

The rule reports when untrusted path components reach sensitive path construction or resource lookup operations.

It currently models:

- `ZipEntry.getName()` and `JarEntry.getName()` as archive-derived path sources
- `getParameter(...)` as a request-derived path source
- conservative path-like method parameters such as `path`, `file`, `filename`, `dir`, `template`, or `resource`
- propagation through variable assignment and string concatenation
- the following sinks:
  - `new File(...)`
  - `mkdir()`, `mkdirs()`, `createNewFile()`
  - `ClassLoader.getResource(...)`

## Validation

The rule has been integrated into the real `sonar-java` source tree and validated with targeted tests.

The following representative cases are detected by the new rule:

- `CVE-2022-4494`
- `CVE-2022-39367`
- `CVE-2022-31194`
- `CVE-2022-29253`

The rule is also included in the generated rule list after running the normal `GeneratedCheckList` build step.

## Scope and design choice

This implementation is intentionally conservative. It is not a full general-purpose taint engine. Instead, it focuses on a compact set of source, propagation, and sink shapes that repeatedly appear in real Java path traversal vulnerabilities.

The goal of this change is to improve coverage for important and recurring path traversal patterns while keeping the implementation understandable and maintainable.

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/ArchiveEntryPathTraversalCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/ArchiveEntryPathTraversalCheckSample.java
@@ -1,0 +1,33 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+import java.io.File;
+import java.io.FileOutputStream;
+import java.util.zip.ZipEntry;
+
+class ArchiveEntryPathTraversalCheckSample {
+  void bad(ZipEntry ze, File destDir) throws Exception {
+    String fileName = ze.getName();
+    File newFile = new File(destDir, fileName); // Noncompliant
+    FileOutputStream fos = new FileOutputStream(newFile);
+  }
+
+  void good(File destDir) throws Exception {
+    String fileName = "safe.txt";
+    File newFile = new File(destDir, fileName);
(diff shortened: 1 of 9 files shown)


rombirli:
Hi @FORIMOC ,

Thank you for your proposal.

Before we can review this PR, we require all new rule proposals to be discussed first on our [Community Forum](https://community.sonarsource.com/). This step is necessary for our team to evaluate the proposal and determine if it can be accepted.

Could you please:
* **Open a thread** on the forum to present your proposal.
* **Review our [Contributing Guidelines](https://github.com/SonarSource/sonar-java/blob/master/CONTRIBUTING.md)** for further details on the process.

Once the discussion has taken place on the community channel, we will be able to follow up here regarding the next steps for this PR.

Thank you in advance for taking these steps!

FORIMOC:
Thanks for the guidance. I opened a discussion on the Community Forum here:
https://community.sonarsource.com/t/java-proposal-path-traversal-rule-for-archive-derived-and-constructed-paths/181940

I’ll wait for feedback there before pushing this PR further.

PR #5606, opened 2026-05-06
SONARJAVA-6328 - Implement S8688: Time-based .now() methods should specify a ZoneId or a Clock

NoemieBenard:


java-checks/src/main/java/org/sonar/java/checks/NowWithoutParametersCheck.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    if (tree instanceof MethodInvocationTree mit && NOW.matches(mit) && mit.methodSelect() instanceof MemberSelectExpressionTree mset) {
+      reportIssue(mset.identifier(), mit, "Explicitly specify the time zone by passing a ZoneId or a Clock to the .now() method.");
```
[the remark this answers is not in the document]
NoemieBenard: It's not a secondary location but a range, which allows to raise on `now()` and not just `now`

java-checks/src/main/java/org/sonar/java/checks/NowWithoutParametersCheck.java
```diff
@@ -0,0 +1,49 @@
...
+    .addWithoutParametersMatcher()
+    .build();
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.METHOD_INVOCATION);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
```
aurelien-coet-sonarsource: If you extend the `AbstractMethodDetection` class, you can override `getMethodInvocationMatchers` method which will trigger the rule on invocations of specific method matchers. Then, instead of calling `visitNode` and explicitly having to check that the node is a `MethodInvocationTree`, you can override the `onMethodInvocationFound` method and directly raise the issue without the need for the condition.

java-checks/src/main/java/org/sonar/java/checks/NowWithoutParametersCheck.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S8688")
+public class NowWithoutParametersCheck extends IssuableSubscriptionVisitor {
```
aurelien-coet-sonarsource: The class should probably extend `JavaVersionAwareVisitor` and check that the Java version is greater than 8.

java-checks/src/main/java/org/sonar/java/checks/NowWithoutParametersCheck.java
```diff
@@ -0,0 +1,53 @@
...
+  @Override
+  protected MethodMatchers getMethodInvocationMatchers() {
+    return NOW_MATCHER;
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    if (NOW_MATCHER.matches(mit) && mit.methodSelect() instanceof MemberSelectExpressionTree mset) {
+      reportIssue(mset.identifier(), mit, "Explicitly specify the time zone by passing a ZoneId or a Clock to the .now() method.");
+    }
```
aurelien-coet-sonarsource: The call to `NOW_MATCHER.matches(mit)` is unnecessary here: the method will only be called on method invocations that match the method matchers returned by `getMethodInvocationMatchers()`. You also know that the method will only ever be invoked on method calls where `methodSelect` will be a `MemberSelectExpressionTree`, so you could do something like:
```suggestion
reportIssue(((MemberSelectExpressionTree) mit.methodSelect()).identifier(), mit, "Explicitly specify the time zone by passing a ZoneId or a Clock to the .now() method.");
```

aurelien-coet-sonarsource (approved):
LGTM

PR #5607, opened 2026-05-06
Add quick fix for `Map.get(key)` `isEqualTo`/`isNotEqualTo(value)`

asm0dey:
Wires a quick fix for the AssertJ chain simplifications targeting
`Map.get(key)`: rewrites `assertThat(map.get(k)).isEqualTo(v)` to
`assertThat(map).containsEntry(k, v)`, and the `isNotEqualTo` variant to
doesNotContainEntry. Adds the new `ActualKeyExpectedQuickFix` which
combines the inner-call argument with the predicate argument, plus
positive (`Map`, `HashMap`) and negative (non-Map `.get`) tests.


asm0dey:
@NoemieBenard hey, I'm sorry, I'm not sure what happened here — somehow there are 0 changes in my PR, should I create a new one with the fixes requested?

PR #5611, opened 2026-05-08
SONARJAVA-6330: Implement S8692 - The system clock should not be used in tests

aurelien-coet-sonarsource:


java-checks/src/main/java/org/sonar/java/checks/tests/SystemClockCheck.java
```diff
@@ -0,0 +1,64 @@
...
+  );
+
+  @Override
+  protected MethodMatchers getMethodInvocationMatchers() {
+    return SYSTEM_CLOCK_MATCHERS;
+  }
+
+  @Override
+  protected void onMethodInvocationFound(MethodInvocationTree mit) {
+    reportIssue(mit, "Replace this use of the system clock with a fixed clock.");
```
tomasz-tylenda-sonarsource: Let's keep the message generic, so it also covers reading the clock to measure elapsed time..

java-checks-test-sources/default/src/test/java/checks/tests/SystemClockCheckSample.java
```diff
@@ -0,0 +1,75 @@
...
+      this.clock = clock;
+    }
+
+    public boolean isTokenValid(Instant issuedAt) {
+      return issuedAt.isAfter(Instant.now(clock).minus(1, ChronoUnit.HOURS));
+    }
+  }
+
+  @Test
+  void testSystemClockInstants() {
```
tomasz-tylenda-sonarsource, line 38: `Instant.ofEpochMilli(System.currentTimeMillis());` is compliant. Should this rule catch it too? (There is also `System.nanoTime()`)

aurelien-coet-sonarsource: Yes, I think that in the first example, we should probably raise an issue on `System.currentTimeMillis()` (rather than `Instant.ofEpochMilli()`, since it doesn't have anything to do with the system clock). `System.nanoTime()` doesn't read the system clock and is very commonly used to measure elapsed time, so I don't think we should raise on it either.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8692.json
```diff
@@ -0,0 +1,26 @@
...
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "1h"
+  },
+  "tags": [
+    "java8",
+    "datetime",
+    "tests"
+  ],
+  "defaultSeverity": "Major",
```
NoemieBenard: We should probably update the RSPEC because the default severity doesn't match the highest impact severity

java-checks-test-sources/default/src/test/java/checks/tests/SystemClockCheckSample.java
NoemieBenard: Even if System.nanoTime doesn't read the wall-clock system time, it still uses the hardware to measure time (which still makes the tests non deterministic), so I think it could be targeted by the rule too

aurelien-coet-sonarsource: Yes, however my issue with including `System.nanoTime` is that it is often one of the recommended ways to measure elapsed time between two points in tests. It is not a problem to read `System.nanoTime` in a test because it does not depend on the system clock (which could change between runs or even during a run). As specified [in the doc](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--): "This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time".

java-checks-test-sources/default/src/test/java/checks/tests/SystemClockCheckSample.java
```diff
@@ -0,0 +1,84 @@
...
+    assertTrue(securityService.isTokenValid(Instant.now(Clock.system(ZoneId.systemDefault())).minusSeconds(60))); // Noncompliant {{Do not use the system clock in tests.}}
+//                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  }
+
+  @Test
+  void testSystemMethods() {
+    long currentTimeMillis = System.currentTimeMillis(); // Noncompliant {{Do not use the system clock in tests.}}
+//                           ^^^^^^^^^^^^^^^^^^^^^^^^^^
+    long currentTimeNanos = System.nanoTime(); // Compliant: nanoTime is typically used to measure elapsed time in tests
+    assertTrue(currentTimeMillis < currentTimeNanos);
```
tomasz-tylenda-sonarsource, line 68: Here is an idea for another rule: if you use a suffix such as `millis` or `nanos`, don't mix them up :) It may not be straightfoward to implement though.

java-checks-test-sources/default/src/test/java/checks/tests/SystemClockCheckSample.java
NoemieBenard: Ok, you're right, it seems to be out of scope for this rule

java-checks/src/main/java/org/sonar/java/checks/tests/SystemClockCheck.java
```diff
@@ -0,0 +1,65 @@
...
+      .addParametersMatcher("java.time.ZoneId")
+      .build(),
+    MethodMatchers.create()
+      .ofTypes("java.time.Clock")
+      .names("systemUTC", "systemDefaultZone", "system")
+      .withAnyParameters()
+      .build(),
+    MethodMatchers.create()
+      .ofTypes("java.lang.System")
+      .names("currentTimeMillis")
```
tomasz-tylenda-sonarsource, line 53: It just occured to me that `new java.util.Date()` is another way to [read the system clock](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Date.html#%3Cinit%3E()) and I think it is used frequently. I know that is class should not be used, and we probably have a rule for that, but this is an orthogonal concern - someone may allow using `java.util.Date` in their code (compatibility with 3rd party code is a good reason to do so) and still want to have an issue about reading system clock in tests.

its/ruling/src/test/resources/eclipse-jetty-similar-to-main/java-S8692.json
tomasz-tylenda-sonarsource, line 1: I assume you reviewed all these sources.

java-checks/src/main/java/org/sonar/java/checks/tests/SystemClockCheck.java
aurelien-coet-sonarsource: You're right, I'll add calls to the `Date` constructors to the nodes we should raise the issue on.

its/ruling/src/test/resources/eclipse-jetty-similar-to-main/java-S8692.json
aurelien-coet-sonarsource: Yes, these are all occurrences of `System.currentTimeMillis` in tests !

java-checks/src/main/java/org/sonar/java/checks/tests/SystemClockCheck.java
NoemieBenard: I think it is the same for `Calendar.getInstance()`

NoemieBenard (commented):
I think we should also add the `java.util.Calendar` `getInstance` method as you did for `new Date()`

NoemieBenard (approved):
LGTM

PR #5620, opened 2026-05-19
SONARJAVA-6321- Implement S8694: DayOfWeek and Month Enums should be used instead of numeric values

NoemieBenard:


java-checks-test-sources/default/src/main/java/checks/DateEnumsCheckSample.java
```diff
@@ -0,0 +1,70 @@
...
+
+  void compliantDateCreation(int monthNumber) {
+    LocalDate date = LocalDate.of(2024, Month.JANUARY, 15);
+    LocalDateTime dateTime = LocalDateTime.of(2024, Month.DECEMBER, 25, 10, 30);
+    YearMonth yearMonth = YearMonth.of(2024, Month.JUNE);
+    LocalDateTime dateTimeWithVariable = LocalDateTime.of(2024, monthNumber, 25, 10, 30);
+  }
+
+  boolean noncompliantDateManipulation(LocalDate date, DayOfWeek day, Month month) {
+    if (date.getMonthValue() == 9) { // Noncompliant {{Use a "java.time.Month" enum constant instead of this int literal.}}
```
aurelien-coet-sonarsource, line 45: Nitpick: Could we report the issue only on the `9` here instead of the entire line ?

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -0,0 +1,192 @@
...
+        reportIssue(binaryExpressionTree, MONTH_ISSUE_MESSAGE);
+        return;
+      }
+      if (DAY_OF_WEEK_GET_VALUE_MATCHER.matches(methodInvocationSide)) {
+        reportIssue(binaryExpressionTree, DAY_ISSUE_MESSAGE);
+      }
+    }
+  }
+
+  private static boolean isIntLiteral(ExpressionTree arg) {
```
aurelien-coet-sonarsource, line 175: I think we should also handle parenthesized expressions in this method. For example, if we encounter `DayOfWeek.of((2))` or `DayOfWeek.of(-(1))`, we should report an issue.
There is a helper method for this in `ExpressionUtils` called `skipParentheses`.

aurelien-coet-sonarsource (commented):
Overall LGTM, just two comments.

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -0,0 +1,192 @@
...
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S8694")
+public class DateEnumsCheck extends AbstractMethodDetection implements JavaVersionAwareVisitor {
+
+  private static final MethodMatchers LOCAL_DATE_OF_MATCHER = MethodMatchers.create()
+    .ofTypes("java.time.LocalDate")
+    .names("of")
+    .addParametersMatcher("int", "int", "int")
+    .build();
```
alban-auzeill: We can group method with a name like `METHOD_WITH_MONTH_AS_SECOND_ARGUMENT`, and use `MethodMatchers.or(`

alban-auzeill (changes requested):

java-checks-test-sources/default/src/main/java/checks/DateEnumsCheckSample.java
NoemieBenard: Since the fix requires to also change the `getMonthValue` method, I thought it was better to highlight everything. Maybe I should update the issue message to make that clearer instead of highlighting only the `9` ? What do you think ?

aurelien-coet-sonarsource: As discussed offline, let's keep it this way here.

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -0,0 +1,186 @@
...
+      }
+    }
+  }
+
+  private static boolean isIntLiteral(ExpressionTree arg) {
+    ExpressionTree argWithoutParentheses = ExpressionUtils.skipParentheses(arg);
+    if (argWithoutParentheses.is(Tree.Kind.INT_LITERAL)) {
+      return true;
+    }
+    if (argWithoutParentheses.is(Tree.Kind.UNARY_MINUS, Tree.Kind.UNARY_PLUS)) {
```
aurelien-coet-sonarsource, line 180: What happens if we have a (silly) expression like `-(-(3))` ? I think we should use something like:
```suggestion
    ExpressionTree expr = ExpressionUtils.skipParentheses(arg);
    while (expr.is(Tree.Kind.UNARY_MINUS, Tree.Kind.UNARY_PLUS)) {
        expr = ExpressionUtils.skipParentheses(((UnaryExpressionTree) expr).expression());
     }
     return expr.is(Tree.Kind.INT_LITERAL);
```

alban-auzeill (approved):

PR #5624, opened 2026-05-20
SONARJAVA-6303: S8696 - Value-based types should be compared using their value

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **New rule S8696 - Value-based type comparison:**
  - Added `ValueBasedObjectIdentityCheck` to detect identity-sensitive operations (`==`, `!=`, `System.identityHashCode()`) on value-based types
  - Created `TypeUtils` utility with `isValueBasedType()` method supporting primitives, `java.util.Optional*`, and `java.time.*` types
- **Refactoring:**
  - Extracted value-based type detection logic from `ValueBasedObjectUsedForLockCheck` into `TypeUtils`
- **Documentation & profiles:**
  - Added rule documentation (S8696.html, S8696.json) with examples and rationale
  - Registered S8696 in Sonar agentic AI and Sonar way profiles
- **Tests:**
  - Added `ValueBasedObjectIdentityCheckTest` and `ValueBasedObjectIdentityCheckSample` with positive and negative cases
  - Added `TypeUtilsTest` covering wrapper types, Optional, java.time classes, and edge cases


java-frontend/src/main/java/org/sonar/java/model/TypeUtils.java
```diff
@@ -0,0 +1,60 @@
...
+    "java.util.OptionalDouble",
+    "java.util.OptionalInt",
+    "java.util.OptionalLong",
+    "java.time.chrono.HijrahDate",
+    "java.time.chrono.JapaneseDate",
+    "java.time.chrono.MinguoDate",
+    "java.time.chrono.ThaiBuddhistDate"
+  );
+
+  private static final Pattern JAVA_TIME_PACKAGE_PATTERN = Pattern.compile("java\\.time\\.\\w+");
```
dorian-burihabwa-sonarsource: Should the regex be more restrictive on the start of the pattern?
```suggestion
  private static final Pattern JAVA_TIME_PACKAGE_PATTERN = Pattern.compile("^java\\.time\\.\\w+");
```

java-frontend/src/main/java/org/sonar/java/model/TypeUtils.java
```diff
@@ -0,0 +1,60 @@
...
+
+  private static final List<String> VALUE_BASED_TYPES = Arrays.asList(
+    "java.lang.Boolean",
+    "java.lang.Byte",
+    "java.lang.Character",
+    "java.lang.Double",
+    "java.lang.Float",
+    "java.lang.Integer",
+    "java.lang.Long",
+    "java.lang.Short",
```
dorian-burihabwa-sonarsource: Is this the complete list of primitive wrappers? If so, we can maybe remove them from the and use `Type::isPrimitiveWrapper()` in `isValueBasedType` ?

java-frontend/src/main/java/org/sonar/java/model/TypeUtils.java
```diff
@@ -0,0 +1,60 @@
...
+    "java.util.Optional",
+    "java.util.OptionalDouble",
+    "java.util.OptionalInt",
+    "java.util.OptionalLong",
+    "java.time.chrono.HijrahDate",
+    "java.time.chrono.JapaneseDate",
+    "java.time.chrono.MinguoDate",
+    "java.time.chrono.ThaiBuddhistDate"
```
dorian-burihabwa-sonarsource, line 34: How did we identify these types? Did we review them? Are they marked with a specific interface? Will we need to update this list with newer versions of Java?

java-frontend/src/test/java/org/sonar/java/model/TypeUtilsTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Test
+  void private_constructor() throws Exception {
+    assertThat(isFinal(TypeUtils.class.getModifiers())).isTrue();
+    Constructor<TypeUtils> constructor = TypeUtils.class.getDeclaredConstructor();
+    assertThat(isPrivate(constructor.getModifiers())).isTrue();
+    assertThat(Modifier.isPrivate(constructor.getModifiers())).isTrue();
+    constructor.setAccessible(true);
+    constructor.newInstance();
+  }
```
dorian-burihabwa-sonarsource, line 37: Does the empty private constructor count towards coverage? If so, do we really want to have a test that looks this thoroughly at the utility class implementation details?

java-frontend/src/main/java/org/sonar/java/model/TypeUtils.java
dorian-burihabwa-sonarsource, line 1: Why don't we move this utility class from the frontend to the checks module? It does not seem to be used elsewhere?

java-frontend/src/main/java/org/sonar/java/model/TypeUtils.java
dorian-burihabwa-sonarsource: I see that these types come from the previous implementation of the synchronization check. Feel free to dismiss it.

java-checks/src/main/java/org/sonar/java/checks/ValueBasedObjectIdentityCheck.java
```diff
@@ -0,0 +1,72 @@
...
+    } else if (tree instanceof MethodInvocationTree methodInvocation &&
+      IDENTITY_HASHCODE_MATCHER.matches(methodInvocation) &&
+      isValueBasedType(methodInvocation.arguments().get(0).symbolType())) {
+      reportIssue(methodInvocation, HASHCODE_MESSAGE);
+    }
+  }
+
+  private static boolean isValueBasedTypeExcludingPrimitiveWrappers(Type type) {
+    // We need to exclude wrappers for primitive classes to avoid duplicating issues with S4973.
+    return isValueBasedType(type) && !type.isPrimitiveWrapper();
```
dorian-burihabwa-sonarsource: Wouldn't it be cheaper/faster to reverse this conditions?
```suggestion
    return !type.isPrimitiveWrapper() && isValueBasedType(type);
```

dorian-burihabwa-sonarsource: Maybe we should consider deprecating S4973 in favor of this new rule?

dorian-burihabwa-sonarsource (commented):
The change looks good to me but I think there are couple of questions we want to answer to make sure we

java-frontend/src/main/java/org/sonar/java/model/TypeUtils.java
aurelien-coet-sonarsource: Yes, these types were already used in the previous implementation for the synchronization check and upon checking the docs, these indeed seem to be the only other value-based classes not at the root of the `java.time` package.

java-frontend/src/main/java/org/sonar/java/model/TypeUtils.java
aurelien-coet-sonarsource: Yes, this is the full list and you're right, they are all covered by `Type::isPrimitiveWrapper()`.

java-frontend/src/test/java/org/sonar/java/model/TypeUtilsTest.java
aurelien-coet-sonarsource: I think it does count towards coverage, yes. This test is essentially a mirror of the one in `ExpressionUtilsTest.java`, but I agree, this might be a bit too thorough.

java-frontend/src/main/java/org/sonar/java/model/TypeUtils.java
aurelien-coet-sonarsource: The idea was to mirror the `ExpressionUtils` class located in the frontend but for types. `TypeUtils` is indeed only used in the checks module, but I think it would be more consistent to keep the two utility classes in the same location (and `TypeUtils` might end up being used elsewhere). WDYT ?

java-checks/src/main/java/org/sonar/java/checks/ValueBasedObjectIdentityCheck.java
aurelien-coet-sonarsource: Yes, I think it would make sense. We'd have to determine if `String` should be treated as a value-based type or a special case to be included in the rule.

NoemieBenard (approved):
LGTM 👍

PR #5625, opened 2026-05-20
SONARJAVA-6389 Implement a quickfix for rule S8694

NoemieBenard:
----
## Summary by Gitar

- **Quickfix implementation for S8694:**
  - Added quickfix to replace int literals with `Month` and `DayOfWeek` enum constants in date/time methods
  - Quickfix automatically adds necessary imports (`java.time.Month`, `java.time.DayOfWeek`) when replacing literals
- **Code improvements:**
  - Extracted type names to constants `JAVA_TIME_MONTH` and `JAVA_TIME_DAY_OF_WEEK`
  - Added `ImportSupplier` for managing import additions across quickfix invocations
  - Added lifecycle methods (`setContext`, `leaveFile`) to reset import supplier state
- **Tests:**
  - Added test case `DateEnumsCheckImportSample.java` verifying quickfix includes proper import statements

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/DateEnumsCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/DateEnumsCheckSample.java
@@ -26,12 +26,12 @@ public class DateEnumsCheckSample {
     //                                ^
     MonthDay md = MonthDay.of(2, 16); // Noncompliant
     //                        ^
-    MonthDay monthDay = MonthDay.of(+2, 16); // Noncompliant
-    LocalDate ld = LocalDate.of(2024, -1, 15); // Noncompliant
+    MonthDay monthDay = MonthDay.of(+2, 16); // Compliant; ignored by the rule to keep it simple
+    LocalDate ld = LocalDate.of(2024, -1, 15); // Compliant; ignored by the rule to keep it simple
     DayOfWeek day = DayOfWeek.of(2); // Noncompliant {{Use a "java.time.DayOfWeek" enum constant instead of this int literal.}}
     Month month = Month.of(10); // Noncompliant {{Use a "java.time.Month" enum constant instead of this int literal.}}
-    DayOfWeek parenthesized = DayOfWeek.of((-3)); // Noncompliant {{Use a "java.time.DayOfWeek" enum constant instead of this int literal.}}
-    DayOfWeek parenthesizedOutside = DayOfWeek.of(-(3)); // Noncompliant {{Use a "java.time.DayOfWeek" enum constant instead of this int literal.}}
+    DayOfWeek parenthesized = DayOfWeek.of((-3)); // Compliant; ignored by the rule to keep it simple
+    DayOfWeek parenthesizedOutside = DayOfWeek.of(-(3)); // Compliant; ignored by the rule to keep it simple
   }
 
   void compliantDateCreation(int monthNumber) {
@@ -67,5 +67,42 @@ public class DateEnumsCheckSample {
     return day == DayOfWeek.THURSDAY;
   }
 
+
+  void quickfixDateCreation() {
+    LocalDate date = LocalDate.of(2024, 1, 15); // Noncompliant [[quickfixes=qf1]]
+    // fix@qf1 {{Replace with Month.JANUARY.}}
+    // edit@qf1 [[sc=41;ec=42]]{{Month.JANUARY}}
+
+    LocalDateTime dateTime = LocalDateTime.of(2024, 10, 15, 1, 20); // Noncompliant [[quickfixes=qf2]]
+    // fix@qf2 {{Replace with Month.OCTOBER.}}
+    // edit@qf2 [[sc=53;ec=55]]{{Month.OCTOBER}}
+
+    MonthDay monthDay = MonthDay.of(2, 16); // Noncompliant [[quickfixes=qf3]]
+    // fix@qf3 {{Replace with Month.FEBRUARY.}}
+    // edit@qf3 [[sc=37;ec=38]]{{Month.FEBRUARY}}
+
+    DayOfWeek day = DayOfWeek.of(2); // Noncompliant [[quickfixes=qf4]]
+    // fix@qf4 {{Replace with DayOfWeek.TUESDAY.}}
+    // edit@qf4 [[sc=34;ec=35]]{{DayOfWeek.TUESDAY}}
+
+    Month month = Month.of(12); // Noncompliant [[quickfixes=qf5]]
+    // fix@qf5 {{Replace with Month.DECEMBER.}}
+    // edit@qf5 [[sc=28;ec=30]]{{Month.DECEMBER}}
+  }
+
+  void quickfixDateManipulation(LocalDate date, DayOfWeek day, Month month) {
+    if (date.getMonthValue() == 9) { // Noncompliant [[quickfixes=qf6]]
+      // fix@qf6 {{Replace with date.getMonth() == Month.SEPTEMBER.}}
+      // edit@qf6 [[sc=9;ec=34]]{{date.getMonth() == Month.SEPTEMBER}}
+    }
+    if (day.getValue() != 3) { // Noncompliant [[quickfixes=qf7]]
+      // fix@qf7 {{Replace with day != DayOfWeek.WEDNESDAY.}}
+      // edit@qf7 [[sc=9;ec=28]]{{day != DayOfWeek.WEDNESDAY}}
+    }
+    if (3 == month.getValue()) { // Noncompliant [[quickfixes=qf8]]
+      // fix@qf8 {{Replace with Month.MARCH == month.}}
+      // edit@qf8 [[sc=9;ec=30]]{{Month.MARCH == month}}
+    }
+  }
 
 }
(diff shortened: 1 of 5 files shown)


java-checks-test-sources/default/src/main/java/checks/DateEnumsCheckSample.java
```diff
@@ -26,12 +26,12 @@ void noncompliantDateCreation() {
     //                                ^
     MonthDay md = MonthDay.of(2, 16); // Noncompliant
     //                        ^
-    MonthDay monthDay = MonthDay.of(+2, 16); // Noncompliant
-    LocalDate ld = LocalDate.of(2024, -1, 15); // Noncompliant
+    MonthDay monthDay = MonthDay.of(+2, 16); // Compliant; ignored by the rule to keep it simple
```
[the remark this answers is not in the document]
NoemieBenard: Since that is an edge case and can be problematic for the quickfix, we decided to restrain the rule to integer literals only

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -124,23 +127,53 @@ public List<Tree.Kind> nodesToVisit() {
...
     }
   }
 
+  private void reportAndCreateQuickfix(ExpressionTree arg, String replacement, String issueMessage) {
+    QuickFixHelper.newIssue(context)
+      .forRule(this)
+      .onTree(arg)
+      .withMessage(issueMessage)
+      .withQuickFix(() -> computeQuickfix(arg,
+        replacement))
```
aurelien-coet-sonarsource: Nitpick: Why the newline here ?

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -150,37 +183,71 @@ public void visitNode(Tree tree) {
...
       }
     }
   }
 
-  private void checkComparison(BinaryExpressionTree binaryExpressionTree, MethodInvocationTree methodInvocationSide, ExpressionTree literalSide) {
-    if (isIntLiteral(literalSide)) {
-      if (METHOD_TO_GET_MONTH_VALUE_AS_INT.matches(methodInvocationSide)) {
-        reportIssue(binaryExpressionTree, MONTH_ISSUE_MESSAGE);
+  private void checkComparison(BinaryExpressionTree binaryExpressionTree,
+    MethodInvocationTree methodInvocationSide, ExpressionTree literalSide, boolean isReversed) {
```
aurelien-coet-sonarsource, line 212: Formatting nitpick: not sure why we have a newline here either.

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -124,23 +127,53 @@ public List<Tree.Kind> nodesToVisit() {
   protected void onMethodInvocationFound(MethodInvocationTree mit) {
     if (METHOD_WITH_MONTH_AS_SECOND_ARGUMENT.matches(mit)) {
       ExpressionTree secondArgument = mit.arguments().get(1);
-      if (isIntLiteral(secondArgument)) {
-        reportIssue(secondArgument, MONTH_ISSUE_MESSAGE);
+      int secondArgumentLiteral = getIntLiteral(secondArgument);
+      if (isValidMonth(secondArgumentLiteral)) {
+        reportAndCreateQuickfix(secondArgument, getMonthEnumName(secondArgumentLiteral), MONTH_ISSUE_MESSAGE);
```
aurelien-coet-sonarsource: This quickfix works for `LocalDateTime` because there exists an overload of the static `of` method with a `Month` as the second argument. However, for other classes such as `ZonedDateTime` or `OffsetDateTime`, the quickfix will produce a solution that does not compile, because there is no overload of e.g. `ZonedDateTime.of(int, int, int, int, int, int, int, java.time.ZoneOffset)` with a `Month` instead of an `int` as second argument. In those cases, the quickfix should suggest replacing the literal with `Month.X.getValue()` instead of `Month.X`.

java-checks-test-sources/default/src/main/java/checks/DateEnumsCheckSample.java
```diff
@@ -68,4 +68,44 @@ boolean compliantDateManipulation(LocalDate date, DayOfWeek day, Month month) {
...
 
+  void quickfixDateCreation() {
+    LocalDate date = LocalDate.of(2024, 1, 15); // Noncompliant [[quickfixes=qf1]]
+    // fix@qf1 {{Replace with Month.JANUARY.}}
+    // edit@qf1 [[sc=41;ec=42]]{{Month.JANUARY}}
+
+    LocalDateTime dateTime = LocalDateTime.of(2024, 10, 15, 1, 20); // Noncompliant [[quickfixes=qf2]]
+    // fix@qf2 {{Replace with Month.OCTOBER.}}
+    // edit@qf2 [[sc=53;ec=55]]{{Month.OCTOBER}}
+
```
aurelien-coet-sonarsource, line 79: We should add examples for `ZonedDateTime` or `OffsetDateTime` where the quickfix suffixes the `Month` enum value with `.getValue()`.

java-checks-test-sources/default/src/main/java/checks/DateEnumsCheckSample.java
```diff
@@ -68,4 +68,44 @@ boolean compliantDateManipulation(LocalDate date, DayOfWeek day, Month month) {
...
+    MonthDay monthDay = MonthDay.of(2, 16); // Noncompliant [[quickfixes=qf3]]
+    // fix@qf3 {{Replace with Month.FEBRUARY.}}
+    // edit@qf3 [[sc=37;ec=38]]{{Month.FEBRUARY}}
+
+    DayOfWeek day = DayOfWeek.of(2); // Noncompliant [[quickfixes=qf4]]
+    // fix@qf4 {{Replace with DayOfWeek.TUESDAY.}}
+    // edit@qf4 [[sc=34;ec=35]]{{DayOfWeek.TUESDAY}}
```
aurelien-coet-sonarsource: I think this quickfix is wrong: the entire `DayOfWeek.of(2)` expression should be replaced with `DayOfWeek.TUESDAY`. Here, the quickfix will produce `DayOfWeek.of(DayOfWeek.TUESDAY)`, which does not compile.

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -124,23 +127,53 @@ public List<Tree.Kind> nodesToVisit() {
...
         return;
       }
     }
     ExpressionTree firstArgument = mit.arguments().get(0);
+    int firstArgumentLiteral = getIntLiteral(firstArgument);
     if (DAY_OF_WEEK_OF_MATCHER.matches(mit)
-      && isIntLiteral(firstArgument)) {
-      reportIssue(firstArgument, DAY_ISSUE_MESSAGE);
+      && isValidDay(firstArgumentLiteral)) {
+      reportAndCreateQuickfix(firstArgument, getDayOfWeekEnumName(firstArgumentLiteral), DAY_ISSUE_MESSAGE);
```
aurelien-coet-sonarsource: The entire `mit` expression should be replaced here, not just the `firstArgument`.

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -124,23 +127,53 @@ public List<Tree.Kind> nodesToVisit() {
...
       return;
     }
     if (METHOD_WITH_MONTH_AS_FIRST_ARGUMENT.matches(mit)
-      && isIntLiteral(firstArgument)) {
-      reportIssue(firstArgument, MONTH_ISSUE_MESSAGE);
+      && isValidMonth(firstArgumentLiteral)) {
+      reportAndCreateQuickfix(firstArgument, getMonthEnumName(firstArgumentLiteral), MONTH_ISSUE_MESSAGE);
```
aurelien-coet-sonarsource: Again, here this will work for `MonthDay.of(int, int)` because there is a `MonthDay.of(Month, int)` overload, but not for `Month.of(int)`. In the last case, the entire expression should be replaced with the computed `Month` enum value.

aurelien-coet-sonarsource (commented):
There are a few issues with the quickfix for some methods that need to be addressed before we can merge, but otherwise good work!

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
NoemieBenard: Ah yes you're right I made a mistake here, I believe in this case the rule should just not fire, because I am not sure the replacement with `.getValue()` is an improvement. WDYT ?

aurelien-coet-sonarsource: I think it is a bit more readable than an integer literal, but the value is debatable in those cases, yes. We could just remove them from the rule.

aurelien-coet-sonarsource (approved):
LGTM 👍

PR #5626, opened 2026-05-22
SONARJAVA-6305 Fix the expected exceptions filter to work without semantic information

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **Refactored checks:**
  - Renamed `InstantConversionsCheck` to `DateTimeConversionsCheck` to reflect its broadened scope.
- **Implemented new logic:**
  - Added detection for invalid date-time conversions involving `LocalDate`, `LocalTime`, and `ZonedDateTime` types.
  - Enhanced `ExpectedExceptionFilter` to allow these conversions to work correctly without full semantic information.
- **Expanded test coverage:**
  - Added `DateTimeConversionsCheckSample` and `DateTimeConversionsCheckTest` to verify the new detection logic.
  - Added additional test cases in `ExpectedExceptionFilter.java` to validate filter behavior with various exception types and signatures.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/filters/ExpectedExceptionFilter.java
+++ b/java-checks/src/main/java/org/sonar/java/filters/ExpectedExceptionFilter.java
@@ -38,7 +38,6 @@ import org.sonar.plugins.java.api.tree.TypeTree;
 import org.sonar.plugins.java.api.tree.UnionTypeTree;
 
 import static org.sonar.java.checks.helpers.MethodTreeUtils.consecutiveMethodInvocation;
-import static org.sonar.java.checks.helpers.UnitTestUtils.isTryCatchFail;
 
 /**
  * A filter to deactivate rules that catch expressions raising specific exceptions in contexts where these exceptions are expected.
@@ -205,11 +204,8 @@ public class ExpectedExceptionFilter extends BaseTreeVisitorIssueFilter {
    * Extract the type name of a class literal expression.
    */
   private static Optional<Type> classLiteralType(ExpressionTree expression) {
-    if (expression.is(Tree.Kind.MEMBER_SELECT)) {
-      MemberSelectExpressionTree memberSelect = (MemberSelectExpressionTree) expression;
-      if ("class".equals(memberSelect.identifier().name())) {
-        return Optional.of(memberSelect.expression().symbolType());
-      }
+    if (expression instanceof MemberSelectExpressionTree memberSelect && "class".equals(memberSelect.identifier().name())) {
+      return Optional.of(memberSelect.expression().symbolType());
     }
     return Optional.empty();
   }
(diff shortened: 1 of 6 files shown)


java-checks/src/test/files/filters/ExpectedExceptionFilter.java
```diff
@@ -151,6 +152,50 @@ void tryCatchFail() {
...
     }
+
+    try {
+      LocalDate.from(instant); // NoIssue
+    } catch (DateTimeException e) {
+      // expected
+    }
+  }
+
+  // When semantic information is missing, the filter should conservatively activate.
```
NoemieBenard, line 163: Is it really about semantic here ?

java-checks/src/test/files/filters/ExpectedExceptionFilter.java
```diff
@@ -151,6 +152,50 @@ void tryCatchFail() {
...
+
+  @Test(expect = DateTimeException.class)
+  void wrongAnnotationAttribute() {
+    Instant.from(date); // WithIssue
+  }
+
+  @Test
+  void wrongAssertMethodSignature() {
+    assertThrows(Instant.from(date)); // WithIssue
+    catchThrowableOfType(Instant.from(date)); // WithIssue
```
NoemieBenard, line 198: I wonder if we are not being too strict with these tests, we could prevent the filter from activating on patterns that we don't recognize even though they are correct. (but maybe we can keep it as it is for now, WDYT?)

java-checks/src/main/java/org/sonar/java/filters/ExpectedExceptionFilter.java
NoemieBenard: This looks nicer IMO: 
```suggestion
if (expression instanceof MemberSelectExpressionTree memberSelect
      && "class".equals(memberSelect.identifier().name())) {
      return Optional.of(memberSelect.expression().symbolType());
    }
```

java-checks/src/main/java/org/sonar/java/filters/ExpectedExceptionFilter.java
```diff
@@ -39,152 +37,173 @@
 import org.sonar.plugins.java.api.tree.TypeTree;
 import org.sonar.plugins.java.api.tree.UnionTypeTree;
 
+import static org.sonar.java.checks.helpers.MethodTreeUtils.consecutiveMethodInvocation;
 import static org.sonar.java.checks.helpers.UnitTestUtils.isTryCatchFail;
```
NoemieBenard: The unused import should be removed

NoemieBenard (commented):
Overall LGTM and should work for the FPs that we have, just some comments

java-checks/src/test/files/filters/ExpectedExceptionFilter.java
aurelien-coet-sonarsource: Yes, here this is an example that wouldn't even compile, because `@Test` is used without a fully qualified path and there is no import for the annotation, so when the filter processes the annotation it doesn't know its actual type. However, since the annotation looks like a test annotation that matches the criteria for the filter, we still want to trigger it.

java-checks/src/test/files/filters/ExpectedExceptionFilter.java
aurelien-coet-sonarsource: I think we don't want to trigger the filter in cases where we do not have enough information to know what specific exceptions are expected. Here, we cannot know if the intent of the user was to assert that the call throws a `DateTimeException` or one of its super/sub-classes. WDYT ?

NoemieBenard: Yes, but we could say that we don't raise if there is any indication that an exception is expected, risking false negatives but avoiding false positives on specific behavior intended by the developer (but I am not sure this can happen). Anyway I think for now the way you did it is fine and we'll have feedback later if there are any issues

NoemieBenard (approved):
LGTM

PR #5633, opened 2026-05-27
SONARJAVA-4426 S5778 allow calling factory methods to create empty collections

tomasz-tylenda-sonarsource:
Expand the list of methods that we can safely ignore.

----
## Summary by Gitar

- **Rule update:**
  - Extended `OneExpectedRuntimeExceptionCheck` to ignore `Collections.singleton`, `Collections.singletonList`, and `Collections.singletonMap` method calls.
- **Test update:**
  - Added test cases in `OneExpectedRuntimeExceptionCheckSample` to verify the exclusion of `Collections.singleton*` methods.
  - Updated `its/autoscan/src/test/resources/autoscan/diffs/diff_S5961.json` to reflect new false negative count for rule `S5961`.

### DIFF
--- a/its/autoscan/src/test/resources/autoscan/diffs/diff_S5778.json
+++ b/its/autoscan/src/test/resources/autoscan/diffs/diff_S5778.json
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S5778",
   "hasTruePositives": false,
-  "falseNegatives": 41,
+  "falseNegatives": 44,
   "falsePositives": 0
 }
--- a/its/autoscan/src/test/resources/autoscan/diffs/diff_S5778.json
+++ b/its/autoscan/src/test/resources/autoscan/diffs/diff_S5778.json
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S5778",
   "hasTruePositives": false,
-  "falseNegatives": 32,
+  "falseNegatives": 41,
   "falsePositives": 0
 }
--- a/java-checks/src/main/java/org/sonar/java/checks/tests/OneExpectedRuntimeExceptionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/tests/OneExpectedRuntimeExceptionCheck.java
@@ -73,6 +73,8 @@ public class OneExpectedRuntimeExceptionCheck extends AbstractOneExpectedExcepti
   /**
    * Frequently used methods known not to throw runtime exceptions.
    */
+  // If adding to this list, make sure that the methods do not call NullPointerExceptions
+  // (for example, `Arrays.asList(null)` throws).
   private static final MethodMatchers AUTHORIZED_METHODS = MethodMatchers.or(
     FAIL_METHOD_MATCHER,
     MOCKITO_MOCK_METHOD_MATCHERS,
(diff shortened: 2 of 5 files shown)


its/autoscan/src/test/resources/autoscan/diffs/diff_S5778.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S5778",
   "hasTruePositives": false,
-  "falseNegatives": 32,
+  "falseNegatives": 41,
```
rombirli: I'm not sure to understand correctly what this value is, but is it expected that we have more FN than before? Even if TPs are missclassified as FNs we shouldn't have more issues raised than before by just expanding the whitelist

rombirli (approved):
It seems quite difficult to maintain an exhaustive list of all commonly used functions that never throw an exception. It might make more sense to rethink the underlying implementation of this rule in the future. That said, I'm approving this PR anyway, as it should still eliminate a solid number of false positives.

java-checks/src/main/java/org/sonar/java/checks/tests/OneExpectedRuntimeExceptionCheck.java
```diff
@@ -33,15 +33,48 @@ public class OneExpectedRuntimeExceptionCheck extends AbstractOneExpectedExcepti
...
+   */
   private static final MethodMatchers AUTHORIZED_METHODS = MethodMatchers.or(
     FAIL_METHOD_MATCHER,
     MOCKITO_MOCK_METHOD_MATCHERS,
-    ENUM_FINAL_METHODS
+    ENUM_FINAL_METHODS,
+    COLLECTIONS_EMPTY,
+    COLLECTION_OF,
+    COLLECTION_CTOR,
+    EMPTY
```
rombirli, line 86: Looks good, I've seen your comment under the jira ticket, but there are many otheres non-raising functions, here are a few commonly used that could make sense to add. 
- Arrays.asList(...)
- `Collections.singleton*` : Collections.singleton(o) / Collections.singletonList(o) / Collections.singletonMap(k, v)...
- `*.of`: `Stream.of(t)` / `Optional.ofNullable(v)`...
- **boxing functions** :`Byte.valueOf(b)`, `Boolean.valueOf(b)`, `Integer.valueOf(i)` 
- ...
I'm not requesting any change, just proposing very commonly used functions you can add if you think it makes sense

its/autoscan/src/test/resources/autoscan/diffs/diff_S5778.json
tomasz-tylenda-sonarsource: For some reason, this check does not work without semantics, so nothing is reported in autoscan. The new false negatives are the additional test cases that I added.

java-checks/src/main/java/org/sonar/java/checks/tests/OneExpectedRuntimeExceptionCheck.java
tomasz-tylenda-sonarsource: I added `Collections.singleton*`. `Arrays.asList` can throw NPE. While the other examples can lead to FPs, I don't think we should be adding a large collection of cases unless we have some evidence that they happen in practice.

I think we can validate this PR on Peachee and follow up if we see more cases worth excluding.

its/autoscan/src/test/resources/autoscan/diffs/diff_S5778.json
rombirli: Ah I understand, thanks for this explanation.

java-checks/src/main/java/org/sonar/java/checks/tests/OneExpectedRuntimeExceptionCheck.java
rombirli: Sounds good!

PR #5634, opened 2026-05-27
SONARJAVA-5360 Implement S8700: Time zone-aware types should be used to compute durations

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **New Rule Implementation:**
  - Added `DateTimeDurationCheck` (S8700) to detect calculations involving `LocalDateTime` that ignore time zone offsets.
- **Detection Logic:**
  - Flags usage of `Duration.between` and `ChronoUnit.between` when arguments are `LocalDateTime` types.
  - Updated detection to handle cases where only one of the two arguments is `LocalDateTime`.
- **Test Coverage:**
  - Created `DateTimeDurationCheckSample.java` and `DateTimeDurationCheckTest.java` to verify compliant and noncompliant code patterns.
- **Rule Metadata:**
  - Included rule definition in `S8700.json`, documentation in `S8700.html`, and enabled it in the `Sonar_way_profile.json`.


java-checks-test-sources/default/src/main/java/checks/DateTimeDurationCheckSample.java
```diff
@@ -0,0 +1,50 @@
...
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneId;
+import java.time.temporal.ChronoUnit;
+
+class DateTimeDurationCheckSample {
+
+  private static final ZoneId UTC = ZoneId.of("UTC");
+
```
NoemieBenard, line 12: The jira ticket also mentions flagging `Instant.now().plus(60, ChronoUnit.DAYS)`, should we add this type of operation too ?

java-checks-test-sources/default/src/main/java/checks/DateTimeDurationCheckSample.java
```diff
@@ -0,0 +1,50 @@
...
+    long halfDays = ChronoUnit.HALF_DAYS.between(localDateTime1, localDateTime2); // Noncompliant {{Convert these arguments to time zone-aware types before computing a duration between them.}}
+//                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    long days = ChronoUnit.DAYS.between(localDateTime1, localDateTime2); // Noncompliant {{Convert these arguments to time zone-aware types before computing a duration between them.}}
+//                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    long days2 = ChronoUnit.DAYS.between(localDateTime1, localDateTime2.atZone(UTC)); // Noncompliant {{Convert this argument to a time zone-aware type before computing a duration on it.}}
+//                                       ^^^^^^^^^^^^^^
+  }
+
+  void compliantExamples(LocalDateTime localDateTime1, LocalDateTime localDateTime2) {
+    Duration duration = Duration.between(LocalTime.now(), LocalTime.now().plusHours(1)); // Compliant: computing a duration between two local time objects without a date means we don't care about time zones
```
NoemieBenard, line 41: Why do we not care about time zones ? For example if we measure elapsed time during a DST transition, the duration will not be correct

NoemieBenard (commented):
The implementation looks good to me, I just have some questions about the scope of the rule

java-checks-test-sources/default/src/main/java/checks/DateTimeDurationCheckSample.java
aurelien-coet-sonarsource: I eventually decided to only focus on the "second" part mentioned in the ticket because the first could be very controversial: the API deliberately offers the possibility to users to add days to an `Instant`, and the semantics of `DAYS` is exactly 24 hours here. Since an instant is just a timestamp on the UTC timeline, this just means adding this amount of hours to it.

java-checks-test-sources/default/src/main/java/checks/DateTimeDurationCheckSample.java
aurelien-coet-sonarsource: The problem here is that there is no way to convert a `LocalTime` to a `ZonedDateTime` without associating it with a specific date (there is no `ZonedTime` class in the API). IMO, we can assume that, if a user computes a duration between `LocalTime`s, they know what they are doing and just want the elapsed wall-clock time.

NoemieBenard: Ok, you're right

NoemieBenard (approved):
LGTM

PR #5636, opened 2026-05-27
SONARJAVA-6405 Create S8745: Test classes should not contain multiple lifecycle methods of the same type

tomasz-tylenda-sonarsource:
----
## Summary by Gitar

- **New inspection rule:**
  - Added `OneTestLifecycleAnnotationCheck` (rule `S8745`) to identify test classes with multiple methods using the same JUnit lifecycle annotation.
- **Rule metadata:**
  - Added `S8745.json` and `S8745.html` with full rule specifications and remediation guidance.
  - Registered the rule in `Sonar_way_profile.json` and `Sonar_agentic_AI_profile.json`.
- **Test coverage:**
  - Created `OneTestLifecycleAnnotationCheckTest` and `OneTestLifecycleAnnotationCheckSample` to verify the rule functionality.
- **Integration testing:**
  - Added `diff_S8745.json` for autoscan validation and updated `java-S8745.json` for ruling tests.
  - Updated `JavaAgenticWayProfileTest` to include the new rule in the profile size count.


java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
```diff
@@ -0,0 +1,91 @@
...
+    reportIssues(lifecycleMethods);
+  }
+
+  /**
+   * Populates the lifecycleMethods map passed as the second parameter.
+   */
+  private static void analyzeMethods(ClassTree classTree, Map<String, Set<IdentifierTree>> lifecycleMethods) {
+    for (Tree member : classTree.members()) {
+      if (member instanceof MethodTree methodTree) {
+        for(SymbolMetadata.AnnotationInstance annotation: methodTree.symbol().metadata().annotations()) {
```
dorian-burihabwa-sonarsource: Formatting
```suggestion
        for (SymbolMetadata.AnnotationInstance annotation: methodTree.symbol().metadata().annotations()) {
```

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
```diff
@@ -0,0 +1,91 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  /**
+   * Reports an issue for each lifecycle annotation that is used more than once.
+   * Reporting happens on the method name.
+   */
```
dorian-burihabwa-sonarsource, line 80: The rule name is `**Test classes** should not contain multiple lifecycle methods of the same type``. We will report on each annotated method. What do you think about reporting once per annotation type on the class declaration (or the first annotated method) and then, using the (other) annotated methods as secondary locations? My feeling is that it would be easier to navigate to all concerned locations at in-product, make the issue remediation a transactional thing where everything is done at once and potentially make the generation of a quickfix easier. WDYT?

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
```diff
@@ -0,0 +1,91 @@
...
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    Map<String, Set<IdentifierTree>> lifecycleMethods = new HashMap<>();
+    analyzeMethods((ClassTree) tree, lifecycleMethods);
```
dorian-burihabwa-sonarsource: Nitpicking: Any reason why analyze methods cannot return a new data structure and it has modify the passed argument to add information to it?

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Populates the lifecycleMethods map passed as the second parameter.
+   */
+  private static void analyzeMethods(ClassTree classTree, Map<String, Set<IdentifierTree>> lifecycleMethods) {
+    for (Tree member : classTree.members()) {
+      if (member instanceof MethodTree methodTree) {
+        for(SymbolMetadata.AnnotationInstance annotation: methodTree.symbol().metadata().annotations()) {
+          String fqn = annotation.symbol().type().fullyQualifiedName();
+          if (LIFECYCLE_ANNOTATIONS.contains(fqn)) {
```
dorian-burihabwa-sonarsource, line 67: What happens if  a method is annotated with more than just the just the lifecycle annotations listed above? For example if we have one of `@Override`, or `@BeforeTransaction` from [Spring](https://docs.spring.io/spring-framework/reference/testing/annotations/integration-spring/annotation-beforetransaction.html)?

dorian-burihabwa-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+      }
+    }
+    return lifecycleMethods;
+  }
+
+  /**
+   * Report issues on methods if lifecycle annotations that are used more than once.
+   */
+  private void reportIssues(Map<String, Set<IdentifierTree>> lifecycleMethods) {
```
dorian-burihabwa-sonarsource: Now that we rely on the order of identifier trees to raise issues, would it make sense to change the value type to `List<IdentiferTree>`. 
This would allow us to extract the first element with `methods.getFirst()` or `methods.get(0)` and then to stream the rest with `methods.stream().skip(1).//..`. WDYT?

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
```diff
@@ -0,0 +1,103 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    reportIssues(analyzeMethods((ClassTree) tree));
+  }
+
+  /**
+   * Returns a map from lifecycle method annotations to a set of method names that are annotated with it.
+   */
+  private static Map<String, Set<IdentifierTree>> analyzeMethods(ClassTree classTree) {
+    Map<String, Set<IdentifierTree>> lifecycleMethods = new LinkedHashMap<>();
```
dorian-burihabwa-sonarsource: It seems unlikely that iterating though class members, we might encounter the same identifier twice for the same annotation. Should we convert the type of the value to `List<IdentifierTree>`?

dorian-burihabwa-sonarsource (approved):
Looking much better. I have left some suggestions for simpler data structure changes

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
tomasz-tylenda-sonarsource: Done. We report on the first method and use the rest as secondary locations.

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
tomasz-tylenda-sonarsource: Done. It was initiali a private field.

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
tomasz-tylenda-sonarsource: It should not affect this rule, but I'll pay attention to this issue during validation.

I marked one of the methods in the unit test `@Deprecated` to consider this scenario.

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
tomasz-tylenda-sonarsource: That's a good idea. Done.

java-checks/src/main/java/org/sonar/java/checks/tests/OneTestLifecycleAnnotationCheck.java
tomasz-tylenda-sonarsource: Done.

PR #5638, opened 2026-05-28
SONARJAVA-5160 S1144 Handle arrays in @MethodSource

tomasz-tylenda-sonarsource:
----
## Summary by Gitar

- **Logic enhancements:**
  - Updated `UnusedPrivateMethodCheck` to correctly identify methods referenced as arrays in `@MethodSource` annotations.
  - Added logic to traverse `NEW_ARRAY` trees within annotation arguments to extract and filter method names.
- **Test coverage:**
  - Added `UnusedPrivateMethodSampleTest.java` to verify false-positive prevention with array-based `@MethodSource` configurations.
  - Included a new test case in `UnusedPrivateMethodCheckTest` to validate the check against the sample file.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
@@ -37,12 +37,14 @@ import org.sonar.plugins.java.api.tree.Arguments;
 import org.sonar.plugins.java.api.tree.AssignmentExpressionTree;
 import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.IdentifierTree;
 import org.sonar.plugins.java.api.tree.LiteralTree;
 import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodInvocationTree;
 import org.sonar.plugins.java.api.tree.MethodReferenceTree;
 import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.NewArrayTree;
 import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.ParameterizedTypeTree;
 import org.sonar.plugins.java.api.tree.Tree;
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -273,26 +283,44 @@ private static boolean isNameIndicatingMethod(String name) {
...
       return withQuotes.substring(1, withQuotes.length() - 1);
     }
 
+    private void removeMethodName(LiteralTree literal) {
+      removeMethodName(literal.value());
+    }
+
+    private void removeMethodNames(NewArrayTree newArrayTree) {
+      for (ExpressionTree initializer : newArrayTree.initializers()) {
+        if (initializer.is(Tree.Kind.STRING_LITERAL)) {
```
rombirli, line 300: The condition should always be true (or the code doesn't compile)

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
```diff
@@ -273,26 +283,44 @@ private static boolean isNameIndicatingMethod(String name) {
...
     public void visitAnnotation(AnnotationTree annotationTree) {
       var isMethodAnnotation = isNameIndicatingMethod(annotationTree.annotationType().symbolType().name());
       for (var arg : annotationTree.arguments()) {
-        if (arg.is(Tree.Kind.STRING_LITERAL)) {
-          if (isMethodAnnotation) {
-            removeMethodName((LiteralTree) arg);
+        if (arg instanceof AssignmentExpressionTree asgn) {
+          if (isMethodAnnotation || isNameIndicatingMethod(((IdentifierTree) asgn.variable()).name())) {
+            removeMethodName(asgn.expression());
           }
```
rombirli, line 322:
```suggestion
        if (!(arg instanceof AssignmentExpressionTree asgn)) { 
          continue;
        }
        if (isMethodAnnotation || isNameIndicatingMethod(((IdentifierTree) asgn.variable()).name())) {
          removeMethodName(asgn.expression());
        }
```

rombirli (approved):
LGTM, minor suggestions

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
tomasz-tylenda-sonarsource: The initializer must be known at compile time, but does not have to be a literal. For example, the following is fine:
```
private static final String PROVIDE_DATA_ARRAY = "provideDataArray";

  @ParameterizedTest
  @MethodSource(value = {PROVIDE_DATA_ARRAY, PROVIDE_DATA_ARRAY + "2"})
  void testAddArray(int num) {}
```
This is, however, a corner case, and I don't think supporting it is a good use of time.

java-checks/src/main/java/org/sonar/java/checks/unused/UnusedPrivateMethodCheck.java
tomasz-tylenda-sonarsource: Let men know if I missed something. 

I don't think this is correct. I tried the suggestion locally in the IDE and the unit tests failed. If the arg is not assignment, then w have something like `@MethodSource("methodName")` and we still need to handle it. 

I added comments in the code to clarify this.

PR #5644, opened 2026-05-29
SONARJAVA-6425 Support records in S1068

tomasz-tylenda-sonarsource:
----
## Summary by Gitar

- **Check logic:**
  - Updated `UnusedPrivateFieldCheck` to include `Tree.Kind.RECORD` in `nodesToVisit` and switch case.
- **Test updates:**
  - Added test case for record private field usage in `UnusedPrivateFieldCheck.java` test sources.

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/unused/UnusedPrivateFieldCheck.java
+++ b/java-checks-test-sources/default/src/main/java/checks/unused/UnusedPrivateFieldCheck.java
@@ -198,5 +198,12 @@ record MyRecord(int a) {
   private static final int XXX = 3; // Noncompliant
 //                         ^^^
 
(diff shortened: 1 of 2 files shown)


asya-vorobeva (changes requested):
I think gitar-bot's comment makes a perfect sense. Let's add this test case: int sum() { return a + XXX; }

tomasz-tylenda-sonarsource:
> I think gitar-bot's comment makes a perfect sense. Let's add this test case: int sum() { return a + XXX; }

Done.

The bot also talks about "explicitly declared private instance fields", but I don't think it is possible.

asya-vorobeva (approved):
💯

PR #5645, opened 2026-06-01
SONARJAVA-6423 Implement S8715 No JUnit 4 assertions in JUnit 5 tests

tomasz-tylenda-sonarsource:
----
## Summary by Gitar

  - Added `JUnit4AssertionsCheck` to flag usage of JUnit 4 assertions (`org.junit.Assert`) in JUnit Jupiter tests.
- **Rule metadata:**
  - Added documentation (`S8715.html`), configuration (`S8715.json`), and integrated the rule into `Sonar agentic AI` and `Sonar way` profiles.
- **Testing:**
  - Created `JUnit4AssertionsCheckTest` and sample code to verify detection, and updated `JavaAgenticWayProfileTest` to account for the new rule.
- **Infrastructure:**
  - Added `diff_S8715.json` to the autoscan test resources and updated `sonarpedia.json` with the latest update timestamp.


java-checks-test-sources/default/src/test/java/checks/tests/NoJUnit4AssertionsInJUnit5TestsCheck_JUnit5WithNoncompliantAssertionsSampleTest.java
rombirli, line 1: Small improvement : make the file name ends with `Sample.java` (also valid for the 2 other files in this folder)

java-checks/src/test/java/org/sonar/java/checks/tests/NoJUnit4AssertionsInJUnit5TestsCheckTest.java
```diff
@@ -0,0 +1,48 @@
...
+      .verifyNoIssues();
+  }
+
+  @Test
+  void test_junit5mixed() {
+    CheckVerifier.newVerifier()
+      .onFile(testCodeSourcesPath("checks/tests/JUnit4AssertionsCheckSampleJUnit5MixedTest.java"))
+      .withCheck(new JUnit4AssertionsCheck())
+      .verifyIssues();
+  }
```
rombirli, line 47: We could add the same test without semantic to ensure no issue is raised

rombirli (approved):
LGTM, 
gitar comment about potential assertions in methods such as BeforeEach could be considered (even if it's not very common)

java-checks-test-sources/default/src/test/java/checks/tests/JUnit4AssertionsCheckSampleJUnit5MixedTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+class JUnit4AssertionsCheckSampleJUnit5MixedTest {
+  @Test
+  void good() {
+    System.out.println("nothing to see here");
```
dorian-burihabwa-sonarsource: We can probably add a compliant assertion and remove the change to S2699 autoscan expected results

java-checks-test-sources/default/src/test/java/checks/tests/JUnit4AssertionsCheckSampleJUnit4Test.java
```diff
@@ -0,0 +1,23 @@
...
+import static org.junit.Assert.fail;
+
+public class JUnit4AssertionsCheckSampleJUnit4Test {
+
+  // We do not raise in old JUnit 4 tests.
+
+  @Test
+  public void testOne() {
+    assertEquals(2, 1 + 1); // Compliant
+    System.out.println();
```
dorian-burihabwa-sonarsource: Do we need to call `System.out.println` in these tests? We would benefit from keeping the test samples as lean as possible to reduce the side-effects of unrelated rules

java-checks/src/main/java/org/sonar/java/checks/tests/JUnit4AssertionsCheck.java
```diff
@@ -0,0 +1,74 @@
...
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+import static org.sonar.java.checks.helpers.UnitTestUtils.JUNIT5_TEST_ANNOTATIONS;
+
+/**
+ * Check that JUnit Jupiter (JUnit 5) tests do not use JUnit 4 assertions.
+ */
+@Rule(key = "S8715")
+public class JUnit4AssertionsCheck extends IssuableSubscriptionVisitor {
```
dorian-burihabwa-sonarsource: Should we change the name of the check to be more explicit? To something like `NoJUnit4AssertionsInJUnit5TestsCheck`?

java-checks-test-sources/default/src/test/java/checks/tests/NoJUnit4AssertionsInJUnit5TestsCheck_JUnit5WithNoncompliantAssertionsSampleTest.java
dorian-burihabwa-sonarsource, line 1: I am not sure I understand the filename for this sample. What is the mixed part of this code?

java-checks-test-sources/default/src/test/java/checks/tests/NoJUnit4AssertionsInJUnit5TestsCheck_JUnit4SampleTest.java
dorian-burihabwa-sonarsource, line 1: To be more explicit in what assertions we can raise on, we could add the FN case where an assertion method is defined in a class that extends `org.junit.Assert` and called in JUnit5 annotated class

sonarpedia.json
dorian-burihabwa-sonarsource, line 1: I think we can exclude this change until full metadata update. WDYT?

dorian-burihabwa-sonarsource (approved):
Looking good overall, I left a couple of comments for potential cleanups and improvements

sonarpedia.json
tomasz-tylenda-sonarsource: Done.

java-checks-test-sources/default/src/test/java/checks/tests/JUnit4AssertionsCheckSampleJUnit4Test.java
tomasz-tylenda-sonarsource: Done. Removed all references to `System.out`. I added some calls which are not assertions to better simulate real world examples.

java-checks-test-sources/default/src/test/java/checks/tests/JUnit4AssertionsCheckSampleJUnit5MixedTest.java
tomasz-tylenda-sonarsource: Done.

java-checks-test-sources/default/src/test/java/checks/tests/NoJUnit4AssertionsInJUnit5TestsCheck_JUnit5WithNoncompliantAssertionsSampleTest.java
tomasz-tylenda-sonarsource: Since these are sample test, they should end in `Test.java` to be realistic. I'll make them `SampleTest.java`.

java-checks/src/test/java/org/sonar/java/checks/tests/NoJUnit4AssertionsInJUnit5TestsCheckTest.java
tomasz-tylenda-sonarsource: Done.

java-checks/src/main/java/org/sonar/java/checks/tests/JUnit4AssertionsCheck.java
tomasz-tylenda-sonarsource: Done.

java-checks-test-sources/default/src/test/java/checks/tests/NoJUnit4AssertionsInJUnit5TestsCheck_JUnit5WithNoncompliantAssertionsSampleTest.java
tomasz-tylenda-sonarsource: Done.

java-checks-test-sources/default/src/test/java/checks/tests/NoJUnit4AssertionsInJUnit5TestsCheck_JUnit4SampleTest.java
tomasz-tylenda-sonarsource: Done.

tomasz-tylenda-sonarsource:
> LGTM, gitar comment about potential assertions in methods such as BeforeEach could be considered (even if it's not very common)

I think this method has enough scope as as, but I'm adding a test to make the behavior explicit (I don't think code should assert in `@Before`).

PR #5646, opened 2026-06-01
SONARJAVA-6410 Create SpringContextModel and related classes

asya-vorobeva:
Create SpringContextModel and related classes. See [SONARJAVA-6410](https://sonarsource.atlassian.net/browse/SONARJAVA-6410).

[SONARJAVA-6410]: https://sonarsource.atlassian.net/browse/SONARJAVA-6410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionHolder.java
+++ b/java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionHolder.java
@@ -144,7 +144,7 @@ public class BeanDefinitionHolder {
 
     public BeanDefinitionHolder build() {
       BeanDefinitionHolder holder = new BeanDefinitionHolder(type, module, beanPackage, location);
-      holder.setDependingBeans(Collections.unmodifiableList(dependingBeans));
+      holder.setDependingBeans(List.copyOf(dependingBeans));
       holder.setProfiles(profiles);
       if (isPrimary) {
         holder.setPrimary();
--- a/java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionHolder.java
+++ b/java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionHolder.java
@@ -1,0 +1,155 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.model.springcontext;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Immutable representation of a Spring bean definition discovered during project scanning.
+ *
+ * <p>Captures the bean's fully-qualified type, the module and package it belongs to,
+ * its source {@link BeanLocation location}, and optional metadata such as active profiles,
+ * dependency names, and whether the bean is marked as {@code @Primary}.
+ *
+ * <p>Use {@link Builder} to construct instances:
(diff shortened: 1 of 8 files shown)


java-frontend/src/main/java/org/sonar/java/model/springcontext/EntityClassToPropertiesIndex.java
```diff
@@ -0,0 +1,56 @@
...
+  private final Map<String, Set<Map.Entry<String, String>>> propertiesByEntityClass = new HashMap<>();
+
+  /**
+   * Registers a property for the given {@code @Entity} class.
+   *
+   * @param entityClass   fully-qualified name of the {@code @Entity} class
+   * @param propertyKey   the property name
+   * @param propertyValue the property value
+   */
+  public void addProperty(String entityClass, String propertyKey, String propertyValue) {
```
leonardo-pilastri-sonarsource, line 43: In this context, what do we mean by `propertyKey` and `propertyValue`?
My assumption is that for each entity we want to know its list of fields, mapping their name to their type FQN. If that's the case I would make it explicit by changing these 2 parameter names to `propertyName` and `propertyType` or something alone those lines.

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionHolder.java
```diff
@@ -0,0 +1,155 @@
...
+ * BeanDefinitionHolder bean = new BeanDefinitionHolder.Builder(type, module, pkg, location)
+ *     .profiles("prod")
+ *     .primary()
+ *     .build();
+ * }</pre>
+ *
+ * @see BeanDefinitionRegistry
+ * @see BeanLocation
+ */
+public class BeanDefinitionHolder {
```
leonardo-pilastri-sonarsource, line 41: Could this be a record?

java-frontend/src/main/java/org/sonar/java/model/springcontext/TypeToBeanNamesIndex.java
```diff
@@ -0,0 +1,56 @@
...
+ * Index mapping fully-qualified bean type names to the names of all beans of that type
+ * discovered during Spring context scanning.
+ *
+ * <p>A single type may have multiple bean names registered (e.g. when the same class is
+ * declared as several distinct beans). Lookup returns an empty set for types with no
+ * registered beans.
+ */
+public class TypeToBeanNamesIndex {
+  /** Bean names indexed by fully-qualified type name. */
+  private final Map<String, Set<String>> beanNamesByType = new HashMap<>();
```
leonardo-pilastri-sonarsource, line 35: I feel like if we want to keep the issue reporting logic in a separate layer from the one that will be collection the information, having a Set here will cause us to miss duplicate bean names being defined.
It is true that the visitor in change to populate this index could check if the name it's trying to add is present in getNamesForType, and report at that moment, but it is going to mix the responsibilities of indexing and reporting. WDYT?

leonardo-pilastri-sonarsource (changes requested):

java-frontend/src/main/java/org/sonar/java/model/springcontext/TypeToBeanNamesIndex.java
asya-vorobeva: I initially planned to use this map together with BeanDefinitionRegistry's one. Combining them together we can easily understand:

- if we have several different bean names for some type (normal situation when we have different profiles)
- if we have several bean definitions with one name for some type (should be reported).

Adding of new entities will be done without reporting or checking for existing names.

And I supposed to have reporting implemented not in visitor, but in some method which simply investigates this structure after it's fully created, in final ProjectSensor.

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionHolder.java
asya-vorobeva: It can, but we have some difficulties because we have List<String> as one of the field, and it's hard to provide immutability for such field in record approach. Besides, I intended to use Builder instead of constructor (which is presented in this class), to make creation more flexible.

java-frontend/src/main/java/org/sonar/java/model/springcontext/EntityClassToPropertiesIndex.java
asya-vorobeva: Currently I'm not sure what we'll need to have here. E.g., we can want to add Hibernate annotations used for the field. Or setter-getter presence.

leonardo-pilastri-sonarsource (approved):

PR #5654, opened 2026-06-04
SONARJAVA-6442 Introduce SpringContextModelGatherer and SpringContextModelSensor

leonardo-pilastri-sonarsource:
SpringContextModel is now a ScannerSide component living across modules and plugged into the DefaultModuleScannerContext (for gathering data) and the SpringContextModelSensor (for issue reporting, after all modules)

Added SpringContextModelGatherer abstract class that allows subclasses to visit the AST and to gather spring-related data at the end of a module analysis, inside the SpringContextModel.

----
## Summary by Gitar

- **New Testing Rules:**
  - Added `S8714`: Identifies `try-catch` blocks with `fail()` calls that should be replaced by `assertThrows` or `assertDoesNotThrow`.
  - Added `S8715`: Prohibits the use of JUnit 4 assertions within JUnit 5 test classes to ensure consistency and correct failure message handling.
  - Added `S8745`: Enforces that classes contain at most one lifecycle method per annotation type (`@BeforeEach`, `@AfterEach`, etc.) to guarantee predictable execution.
- **Refactoring & Infrastructure:**
  - Improved `UnitTestUtils` to better identify JUnit 5 annotations and fail-method invocations, and updated `SuppressWarningFilter` to handle lowercase rule names.
  - Updated test suites across the project to replace manual `try-catch-fail` blocks with AssertJ's `assertThatThrownBy` and `assertThatCode` for more robust testing.


java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextModel.java
```diff
@@ -41,6 +46,13 @@ public class SpringContextModel {
   /** Index of properties associated with Spring Data / Hibernate {@code @Entity} classes. */
   private EntityClassToPropertiesIndex entityClassToPropertiesIndex;
 
+  public SpringContextModel() {
```
asya-vorobeva: I'd suggest to remove setters as we'll modify the fields through getters manipulation. (like in the sample's sensor: springContextModel.getProjectPackageScan().addPackage(data, data)).

Also we can remove constructor and initialize fields directly:

```
  private final BeanDefinitionRegistry beanDefinitionRegistry = new BeanDefinitionRegistry();

  /** Packages registered for Spring component scanning, grouped by module. */
  private final ProjectPackageScan projectPackageScan = new ProjectPackageScan();

  /** Index for resolving bean names by their fully-qualified type. */
  private final TypeToBeanNamesIndex typeToBeanNamesIndex = new TypeToBeanNamesIndex();

  /** Index of properties associated with Spring Data / Hibernate {@code @Entity} classes. */
  private final EntityClassToPropertiesIndex entityClassToPropertiesIndex = new EntityClassToPropertiesIndex();

```

java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextModelGatherer.java
```diff
@@ -0,0 +1,42 @@
...
+  @Override
+  public final void endOfAnalysis(ModuleScannerContext context) {
+    var defaultModuleContext = (DefaultModuleScannerContext) context;
+    gatherSpringContextData(defaultModuleContext.getSpringContextModel());
+  }
+
+  /**
+   * Method called at the end of the analysis of a module, allowing to store gathered data in the SpringContextModel.
+   */
+  public abstract void gatherSpringContextData(SpringContextModel springContextModel);
```
asya-vorobeva: Add parameter: ModuleScannerContext context

asya-vorobeva (changes requested):

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
```diff
@@ -136,21 +137,22 @@ public class SonarComponents extends CheckRegistrar.RegistrarContext {
   private SensorContext context;
   private UnaryOperator<List<JavaCheck>> checkFilter = UnaryOperator.identity();
   private final Set<RuleKey> additionalAutoScanCompatibleRuleKeys;
+  private SpringContextModel springContextModel;
```
asya-vorobeva, line 140: The same here. We can remove setter and initialize it here:
`private final SpringContextModel springContextModel = new SpringContextModel();`

asya-vorobeva (changes requested):

java-frontend/src/test/java/org/sonar/java/model/springcontext/SpringContextModelGathererTest.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  private SpringContextModel getSpringModelAfterVisitingFile(String filePath, JavaCheck check) {
+    File file = new File(filePath);
+    InputFile inputFile = TestUtils.inputFile(file);
+    var compilationUnit = JParserTestUtils.parse(file);
+    SensorContextTester sensorContextTester = SensorContextTester.create(new File(""));
+    var sonarComponents = new SonarComponents(null, null, null, null, null, null);
+    sonarComponents.setSensorContext(sensorContextTester);
+    sonarComponents.setSpringContextModel(new SpringContextModel());
```
asya-vorobeva: Let's simplify:
```
class SpringContextModelGathererTest {
private final SpringContextModel model = new SpringContextModel();

@Test
  void testGatherSpringContextData() {
    visitingFile("src/test/files/model/SimpleClass.java", new SampleGatherer());
    assertThat(model.getTypeToBeanNamesIndex().getNamesForType("com.example.MyService")).containsExactly("myServiceBean");
  }

private void visitingFile(String filePath, JavaCheck check) {
   ...
sonarComponents.setSpringContextModel(model);
...
 }
}
```

asya-vorobeva (changes requested):

java-frontend/src/main/java/org/sonar/java/SonarComponents.java
leonardo-pilastri-sonarsource: The instance created this way in SonarComponents would not be the same living in the SensorContext and later provided to the ProjectSensor, so we still need to rely on the SensorContext prividing the model instance

asya-vorobeva: Yep, you're right, that one should be injected via JavaPlugin.

asya-vorobeva (approved):
Generally LGTM, please fix QG failures before merge. And rebase to master to fix broken CI (was fixed in https://github.com/SonarSource/sonar-java/pull/5658/changes).

PR #5657, opened 2026-06-05
SONARJAVA-6446 Refactor merging list of arguments to MethodMatchers

tomasz-tylenda-sonarsource:
----
## Summary by Gitar

- **New utility method:**
  - Added `StringUtils.flatten` to concatenate `String`, `String[]`, and `Collection<String>` arguments into a single `String[]` array.
- **Refactoring:**
  - Simplified `MethodMatchers` definitions in `UnitTestUtils` by replacing complex `Stream.concat` logic with the new `flatten` helper.
- **Testing:**
  - Added `StringUtilsTest.testFlatten` to verify argument merging and exception handling for unsupported types.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/helpers/StringUtils.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/helpers/StringUtils.java
@@ -17,6 +17,10 @@
 package org.sonar.java.checks.helpers;
 
 import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 
 public class StringUtils {
   private StringUtils() {}
@@ -40,5 +44,31 @@ public class StringUtils {
     }
 
     return count;
+  }
+
+  /**
+   * Build String[] by concatenating Strings, arrays of Strings, and Collections of Strings.
+   * Nested collections and arrays are not supported, and will throw an IllegalArgumentException if encountered.
+   */
+  public static String[] stringArgs(Object ... args) {
+    List<String> result = new ArrayList<>();
+    for (Object arg : args) {
+      if (arg instanceof String s) {
+        result.add(s);
+      } else if (arg instanceof String[] arr) {
+        Collections.addAll(result, arr);
+      } else if (arg instanceof Collection<?> col) {
+        for (Object o : col) {
+          if (o instanceof String s) {
+            result.add(s);
+          } else {
+            throw new IllegalArgumentException("Unsupported collection element type: " + o.getClass());
+          }
+        }
+      } else {
+        throw new IllegalArgumentException("Unsupported argument type: " + arg.getClass());
+      }
+    }
+    return result.toArray(new String[0]);
   }
 }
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/helpers/StringUtils.java
```diff
@@ -41,4 +45,30 @@ public static int countMatches(@Nullable String string, @Nullable String pattern
...
+        result.add(s);
+      } else if (arg instanceof String[] arr) {
+        Collections.addAll(result, arr);
+      } else if (arg instanceof Collection<?> col) {
+        for (Object o : col) {
+          if (o instanceof String s) {
+            result.add(s);
+          } else {
+            throw new IllegalArgumentException("Unsupported collection element type: " + o.getClass());
+          }
```
dorian-burihabwa-sonarsource: Can simplify the code here since we throw exceptions if any of the collection elements is not of the expected type in the `else block below`?

```suggestion
      } else if (arg instanceof Collection<String> col) {
          result.addAll(col);
```

java-checks/src/main/java/org/sonar/java/checks/helpers/StringUtils.java
```diff
@@ -41,4 +45,30 @@ public static int countMatches(@Nullable String string, @Nullable String pattern
 
     return count;
   }
+
+  /**
+   * Build String[] by concatenating Strings, arrays of Strings, and Collections of Strings.
+   * Nested collections and arrays are not supported, and will throw an IllegalArgumentException if encountered.
+   */
+  public static String[] stringArgs(Object ... args) {
```
dorian-burihabwa-sonarsource: Can we name this method with something more active? `concat`? `join`?

java-checks/src/main/java/org/sonar/java/checks/helpers/StringUtils.java
```diff
@@ -41,4 +45,30 @@ public static int countMatches(@Nullable String string, @Nullable String pattern
 
     return count;
   }
+
+  /**
+   * Build String[] by concatenating Strings, arrays of Strings, and Collections of Strings.
+   * Nested collections and arrays are not supported, and will throw an IllegalArgumentException if encountered.
```
dorian-burihabwa-sonarsource:
```suggestion
   * Build String[] by concatenating arguments of types:
   * <ol>
   *   <li>java.lang.String</li>
   *   <li>java.lang.String[]</li>
   *   <li>java.util.Collection<java.lang.strings></li>
   * </ol> 
   * Nested collections and arrays are not supported, and will throw an IllegalArgumentException if encountered.
   * @throws  IllegalArgumentException If one of the argument is not of the supported types
```

dorian-burihabwa-sonarsource (approved):
The change looks OK to me but I would rather we keep the code simpler and make it evolve if we feel a need for a dedicated type of exception message when we encounter the issue in a collection or a regular argument

java-checks/src/main/java/org/sonar/java/checks/helpers/StringUtils.java
tomasz-tylenda-sonarsource: I chose `flatten`, because `concat` and `join` suggest that the results is a String.

java-checks/src/main/java/org/sonar/java/checks/helpers/StringUtils.java
tomasz-tylenda-sonarsource: Due to type erasure, we can only cast to `Collection<?>`. I added an unsafe cast later, which results in ArrayStoreException later.

PR #5662, opened 2026-06-07
SONARJAVA-6411 Implement Visitor for collecting packages scanned by Spring

asya-vorobeva:
See [SONARJAVA-6411](https://sonarsource.atlassian.net/browse/SONARJAVA-6411).

[SONARJAVA-6411]: https://sonarsource.atlassian.net/browse/SONARJAVA-6411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

----
## Summary by Gitar

- **New feature:**
  - Added `ComponentScanPackageGatherer` to collect packages configured via `@ComponentScan` and `@SpringBootApplication`.
  - Integrated the gatherer into `SpringContextModel` to track Spring configuration across project modules.
- **Refactoring:**
  - Moved `SpringUtils` and `UnitTestUtils` to `java-frontend` to centralize model-related utilities.
- **Testing:**
  - Implemented `ComponentScanPackageGathererTest` and extended `CheckVerifier` to support asserting on `SpringContextModel` data.
- **Caching:**
  - Added per-file caching logic for collected package data to optimize incremental analysis performance.
- **Infrastructure:**
  - Enabled skipping of `SpringContextModelGatherer` visitors when `spring-context` is missing from the classpath using `DependencyVersionAware`.


java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextModelGatherer.java
```diff
@@ -16,17 +16,33 @@
...
+ * {@code IssuableSubscriptionVisitorsRunner} after each file. This is required for per-file cache
+ * writes: without it, cache entries are never stored and {@link #scanWithoutParsing} always misses,
+ * preventing unchanged files from being skipped in incremental analyses.
  */
-public abstract class SpringContextModelGatherer extends SubscriptionVisitor implements EndOfAnalysis {
+public abstract class SpringContextModelGatherer extends IssuableSubscriptionVisitor implements EndOfAnalysis, DependencyVersionAware {
+
+  @Override
+  public boolean isCompatibleWithDependencies(Function<String, Optional<Version>> dependencyFinder) {
+    return dependencyFinder.apply("spring-context").isPresent();
```
leonardo-pilastri-sonarsource: I fear this is not enough, as a module could still provide bean definitions without depending on spring-context. We probably need to extend this condition to a list of known dependencies that would allow a module to define beans, repositories, and so on
Something like
```java
return dependencyFinder.apply("spring-context")
      .or(() -> dependencyFinder.apply("spring-beans"))
      .or(() -> dependencyFinder.apply("spring-data-jpa"))
      ....... more here ......
      .isPresent();
```

java-frontend/src/main/java/org/sonar/java/model/UnitTestUtils.java
```diff
@@ -14,7 +14,7 @@
  * You should have received a copy of the Sonar Source-Available License
  * along with this program; if not, see https://sonarsource.com/license/ssal/
  */
-package org.sonar.java.checks.helpers;
+package org.sonar.java.model;
```
leonardo-pilastri-sonarsource: I feel like this is not the best package to contain this class, maybe we can put it under `org.sonar.java.utils`

java-frontend/src/main/java/org/sonar/java/model/SpringUtils.java
```diff
@@ -14,10 +14,10 @@
  * You should have received a copy of the Sonar Source-Available License
  * along with this program; if not, see https://sonarsource.com/license/ssal/
  */
-package org.sonar.java.checks.helpers;
+package org.sonar.java.model;
```
leonardo-pilastri-sonarsource: Same here, this would better fit under the `org.sonar.java.utils` pacakge

leonardo-pilastri-sonarsource (changes requested):
Great work, especially with the tests and the cache implementation 🚀 
I left a couple of things we should look at before merging

java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextModelGatherer.java
asya-vorobeva: I'd say that spring-data-jpa is additional thing and if we have it we'll definitely have spring-context or spring-beans.
Another thing that I've forgot is spring-boot-starter / spring-boot-starter-web (which pulls in spring-context, spring-beans, spring-core and so on).

leonardo-pilastri-sonarsource (approved):
Looks good, but I think the its/sources submodule slipped through, if you can remove that, then it's good to go

PR #5665, opened 2026-06-10
SONARJAVA-6413 Implement BeanDefinitionGatherer to collect Spring bean definitions

asya-vorobeva:
- Add BeanDefinitionGatherer that discovers beans from stereotype annotations (@Component, @Service, @Repository, @Controller, @RestController, @Configuration) and @Bean methods inside configuration/component classes
- Capture @Primary flag and dependencies (@Autowired fields, constructors, setters; @Bean method parameters)
- Register BeanDefinitionGatherer in SpringContextModelGatherers

What's not done here:
- Didn't implement caching layer as it was done in ComponentScanPackageGatherer, needs to be added in following PRs
- Ignored @Qualifier annotations (most important) and @DependsOn / @Lookup annotations (less important)
- Didn't fill TypeToBeanNamesIndex, should be done when working on SONARJAVA-6412
- To properly fill dependantBeans we need to do the second pass when beans from all modules will be collected. It should be done in aggregation step in SpringContextModelSensor, before to create and raise issues..


java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionGatherer.java
```diff
@@ -0,0 +1,213 @@
...
+        AnalyzerMessage.textSpanFor(classTree.simpleName()),
+        meta.isAnnotatedWith(PRIMARY_ANNOTATION),
+        deps));
+
+      // @Bean methods — only if class is a configuration/component class
+      for (MethodTree method : methodsOf(classTree)) {
+        if (method.symbol().metadata().isAnnotatedWith(SpringUtils.BEAN_ANNOTATION)) {
+          collectBeanMethod(method, pkg);
+        }
+      }
```
aurelien-coet-sonarsource: Suggestion: the call to `methodsOf` needs to iterate over the class members to filter them and only keep methods, then instantiates a list, then this list is iterated over again. It would probably be more efficient to directly operate on the stream:

```suggestion
      classTree.members().stream()
        .filter(m -> m.is(Tree.Kind.METHOD))
        .map(MethodTree.class::cast)
        .filter(m -> m.symbol().metadata().isAnnotatedWith(SpringUtils.BEAN_ANNOTATION))
        .forEach(method -> collectBeanMethod(method, pkg));
```

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionGatherer.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  @Override
+  public boolean scanWithoutParsing(InputFileScannerContext ctx) {
+    // Bean data is not cached yet; force parsing so beans are
+    // always collected even for unchanged files in incremental runs.
+    return false;
+  }
+
+  private static boolean isSpringBeanDefinitionsClass(SymbolMetadata meta) {
```
aurelien-coet-sonarsource: Would it make sense to move this method in `SpringUtils` ?

java-frontend/src/test/java/org/sonar/java/model/springcontext/SpringContextModelGathererTest.java
```diff
@@ -62,16 +62,6 @@ void isCompatibleWithDependencies_false_when_no_spring_dependency_is_present() {
     assertThat(new SampleGatherer().isCompatibleWithDependencies(finderFor())).isFalse();
   }
 
-  // ---- ComponentScanPackageGatherer -----------------------------------------
-
-  @Test
-  void componentScanPackageGatherer_collects_package_from_springBootApplication() {
```
aurelien-coet-sonarsource, line 68: Why was this test deleted ?

aurelien-coet-sonarsource (commented):
Overall LGTM, just a few small comments. I think that the first of the two remaining suggestions by the Gitar bot is also correct ?

java-frontend/src/test/java/org/sonar/java/model/springcontext/SpringContextModelGathererTest.java
asya-vorobeva: Because we provide the tests for this gatherer in ComponentScanPackageGathererTest.

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionGatherer.java
asya-vorobeva: Better solution is to remove this method :) I do it in following commit

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionGatherer.java
asya-vorobeva: I extract the method already used in another visitor to SpringUtils.

asya-vorobeva:
> Implicit single-constructor injection not captured as dependencies

This one as added as TODO in Jira ticket. Feel free to do it in follow-up

> packageName removes guard, casts expr to IdentifierTree unconditionally

This is really super-low risk, I don't think that such situations will happen in real-world apps.

java-frontend/src/test/java/org/sonar/java/model/springcontext/SpringContextModelGathererTest.java
aurelien-coet-sonarsource: Ah ok, got it!

PR #5667, opened 2026-06-10
SONARJAVA-6440 S2245: Add security-context heuristic to reduce false positives

Luqmansonar:
## Summary

Ports the DART-276 security-context heuristic to Java for S2245 (`PseudoRandomCheck`). Pseudorandom number generator usages are now reported only when the surrounding code is plausibly security-relevant.

The heuristic has two parts:

- **Part 1 — file-level crypto import check.** If the compilation unit imports any class under one of these prefixes, all PRNG calls in the file are reported (matches ticket spec):
  - `java.security.`, `javax.crypto.`, `org.springframework.security.`, `org.bouncycastle.`, `io.jsonwebtoken.`, `com.auth0.jwt.`, `at.favre.lib.crypto.`, `de.mkammerer.argon2.`
- **Part 2 — scope-level keyword scan.** If no crypto import is present, walk the parent chain from the PRNG call to find the closest enclosing `MethodTree` (method/constructor); fall back to the enclosing `ClassTree` for field/static initializers. Collect every identifier in that scope, tokenize it (split on `_`, then on camelCase / PascalCase boundaries; all-uppercase parts stay whole), lowercase each token, and intersect with the 29-keyword set from the ticket. Any hit ⇒ flag.

## Decisions

- **Heuristic applies to all 6 currently-flagged PRNG sources** (`Random`, `JVMRandom`, `Math.random()`, `ThreadLocalRandom`, `RandomUtils`, `RandomStringUtils`) — not only the two literally listed in the ticket. Confirmed with the requester. The ticket's motivation (false positives in non-security code) applies equally to all six.
- **Scope = enclosing method/constructor; falls back to enclosing class for field-level usage.** Mirrors Dart's `declarationScope` semantics (method for local code, class for field code). Balances FP reduction against recall on the common `private static final Random R = new Random();` pattern.
- **Tokenization matches the Dart reference exactly** (`_splitIntoWords` from `sonar-dart`). One faithful-port quirk worth knowing: the keyword `randombytes` is all-lowercase, so `randomBytes` (camelCase) which tokenizes to `[random, bytes]` does NOT match — only literal `randombytes` or `RANDOMBYTES` does. Same behavior as Dart; documented in the test sample.
- **Rule HTML description (`S2245.html`) was NOT modified** — the ticket does not request a doc change. Happy to update separately if desired.

## Test plan

- [x] Existing `PseudoRandomCheckSample.java` assertions still pass (file imports `java.security.SecureRandom` → Part 1 triggers, all current `// Noncompliant` markers preserved).
- [x] New `PseudoRandomCheckNoContextSample.java`: neutral file, all calls Compliant (heuristic suppresses).
- [x] New `PseudoRandomCheckSecurityKeywordsSample.java`: camelCase / snake_case / UPPER / method-name / parameter keyword cases trigger; negative cases (`unrelatedScope`, `splitRandomBytes`) do not.
- [x] New `PseudoRandomCheckCryptoImportSample.java`: `org.bouncycastle` import → all calls flagged via Part 1.
- [x] New `PseudoRandomCheckFieldScopeSample.java`: field / static-initializer calls fall back to class scope; `TokenGenerator`, `Holder` (with `password` field), `CipherBundle` flagged; neutral class is not.
- [x] `mvn -pl java-checks test -Dtest=PseudoRandomCheckTest` → 5 tests, 0 failures.


java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -94,4 +133,109 @@ private static boolean isStaticCallToInsecureRandomMethod(MethodInvocationTree m
...
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private static boolean matchesCryptoPrefix(String importName) {
+    for (String prefix : CRYPTO_IMPORT_PREFIXES) {
+      if (importName.startsWith(prefix)) {
+        return true;
```
nils-werner-sonarsource, line 144: **Bug: wildcard import not detected**

A wildcard import like `import java.security.*;` concatenates to `"java.security.*"`. The check `importName.startsWith("java.security.")` won't match because the string ends with `".*"`.

Suggested fix — normalise the import name before matching:
```java
private static boolean matchesCryptoPrefix(String importName) {
  // Normalise wildcard imports: "java.security.*" -> "java.security."
  String name = importName.endsWith(".*")
    ? importName.substring(0, importName.length() - 1)
    : importName + ".";
  for (String prefix : CRYPTO_IMPORT_PREFIXES) {
    if (name.startsWith(prefix)) return true;
  }
  return false;
}
```
Please also add a test case with `import java.security.*;` in `PseudoRandomCheckCryptoImportSample.java`.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -94,4 +133,109 @@ private static boolean isStaticCallToInsecureRandomMethod(MethodInvocationTree m
...
+      current = current.parent();
+    }
+    return null;
+  }
+
+  // Mirrors Dart's `_splitIntoWords`: split on underscores first; for each part either keep it as
+  // a single lowercase word when all-uppercase, or split further on capital-letter boundaries.
+  static List<String> tokenizeIdentifier(String identifier) {
+    List<String> words = new java.util.ArrayList<>();
+    for (String part : identifier.split("_")) {
```
nils-werner-sonarsource, line 194: **Nit: inline `java.util.ArrayList` should be a proper import**

```java
List<String> words = new java.util.ArrayList<>();
```

This is the only class in this file referenced by fully-qualified name. Please add `import java.util.ArrayList;` at the top and use `new ArrayList<>()` here.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -66,22 +78,49 @@ public class PseudoRandomCheck extends IssuableSubscriptionVisitor {
...
+  private static final Set<String> SECURITY_KEYWORDS = SetUtils.immutableSetOf(
+    "aes", "asymmetric", "auth", "certificate", "chacha20", "cipher", "crypto", "cryptography",
+    "decrypt", "ecc", "encrypt", "encryption", "hmac", "iv", "nonce", "password", "pbkdf2",
+    "poly1305", "randombytes", "rsa", "salt", "scrypt", "secret", "secure", "security",
+    "sessionid", "signature", "token", "verify"
+  );
+
+  private boolean cryptoImportPresent = false;
+  private final Map<Tree, Boolean> scopeSecurityContextCache = new IdentityHashMap<>();
+
```
nils-werner-sonarsource, line 101: **Concern: cache may not be worth the complexity**

The `scopeSecurityContextCache` avoids re-walking a scope when multiple PRNG calls appear in the same method. In practice this is rare, and the `IdentifierCollector` walk over a method body is cheap. Meanwhile:
- The cache pins AST node references for the file's lifetime.
- The `clear()` in `visitNode(COMPILATION_UNIT)` creates an implicit lifecycle contract that is easy to break.

Consider removing the cache and calling `scopeHasSecurityKeyword` directly:
```java
private boolean isInSecurityContext(Tree tree) {
  if (cryptoImportPresent) return true;
  Tree scope = findDeclarationScope(tree);
  return scope != null && scopeHasSecurityKeyword(scope);
}
```
If profiling ever shows this matters, the cache can be reintroduced — but the `clear()` should then be in `leaveNode(COMPILATION_UNIT)` to make the lifecycle explicit.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -94,4 +133,109 @@ private static boolean isStaticCallToInsecureRandomMethod(MethodInvocationTree m
...
+    return false;
+  }
+
+  private boolean isInSecurityContext(Tree tree) {
+    if (cryptoImportPresent) {
+      return true;
+    }
+    Tree scope = findDeclarationScope(tree);
+    if (scope == null) {
+      return false;
```
nils-werner-sonarsource: **Minor: silent `false` on null scope suppresses issues without explanation**

`findDeclarationScope` returning `null` means the PRNG call has no enclosing `MethodTree` or `ClassTree`, which shouldn't occur in valid Java. Returning `false` silently drops the issue in this edge case.

Prefer returning `true` (flag by default when context is unknown) or add a comment explaining why suppressing is the right tradeoff:
```java
if (scope == null) {
  // Defensive: shouldn't happen in valid Java; flag to avoid silent FN.
  return true;
}
```

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -66,22 +78,49 @@ public class PseudoRandomCheck extends IssuableSubscriptionVisitor {
     "org.apache.commons.lang.math.JVMRandom"
   );
 
+  // SONARJAVA-6440: ported from DART-276 security-context heuristic.
+  // When any of these prefixes appear in the file's imports, all PRNG usages in the file are reported.
```
nils-werner-sonarsource: **Nit: comment explains origin, not the invariant**

Ticket references rot as context. The comment should explain *why* this is the right design, not where the idea came from.

Suggestion:
```java
// A crypto library import signals the whole file is security-relevant;
// report all PRNG calls without needing per-scope keyword analysis.
private static final List<String> CRYPTO_IMPORT_PREFIXES = List.of(
```

java-checks-test-sources/default/src/main/java/checks/PseudoRandomCheckSecurityKeywordsSample.java
```diff
@@ -0,0 +1,61 @@
+package checks;
+
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import org.apache.commons.lang3.RandomUtils;
+import org.apache.commons.lang3.RandomStringUtils;
+
+// SONARJAVA-6440: no crypto import. Security keywords reached via per-scope identifier scan.
+class PseudoRandomCheckSecurityKeywordsSample {
```
nils-werner-sonarsource, line 9: **Suggestion: a few edge cases worth adding to the test suite**

1. **Digit-containing uppercase identifiers** — `AES256` has no lowercase letter so `isAllUppercaseWithLetter` returns `true` and it lowercases to `"aes256"`, which does **not** match the keyword `"aes"`. A variable `AES256_KEY` would therefore not trigger the rule. Please add a test to document this (Compliant or Noncompliant — either is fine, but it should be intentional).

2. **Wildcard crypto import** — `import java.security.*;` should trigger Part 1 (see comment on `matchesCryptoPrefix`). Add a test case in `PseudoRandomCheckCryptoImportSample.java`.

3. **Nested inner class isolation** — `findDeclarationScope` returns the inner `ClassTree` first, so a PRNG call in an inner class should *not* pick up keywords from the outer class. Worth an explicit Compliant test to confirm this isolation is intentional.

4. **Static import of a crypto class** — `import static javax.crypto.Cipher.ENCRYPT_MODE;` is still `Tree.Kind.IMPORT`, so it should be caught by `hasCryptoImport`. A targeted test would confirm this.

its/autoscan/src/test/resources/autoscan/diffs/diff_S2440.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S2440",
   "hasTruePositives": true,
-  "falseNegatives": 2,
+  "falseNegatives": 4,
   "falsePositives": 0
```
nils-werner-sonarsource, line 5: **Unrelated regression: S2440 false-negatives increased**

S2440 (`StringConcatenationInLoop`) is a completely different rule from S2245. The `falseNegatives` going from 2 → 4 here doesn't appear related to this PR's changes. Can you clarify whether this is expected? If not, please revert.

nils-werner-sonarsource (commented):
Overall this is a well-structured and well-documented implementation — the PR description is thorough and the per-scenario test files are a good pattern. A few bugs and concerns below.

nils-werner-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
Luqmansonar: Verified locally: `"java.security.*".startsWith("java.security.")` returns `true`. `ExpressionsHelper.concatenate` produces `"java.security.*"` and the prefix `"java.security."` contains no `*`, so `startsWith` already matches. Added `PseudoRandomCheckWildcardImportSample.java` (and the static-import sample) in c1708df to lock both cases in.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
Luqmansonar: Fixed in c1708df.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
Luqmansonar: Kept the cache; moved `clear()` + the `cryptoImportPresent` reset to a new `leaveNode(COMPILATION_UNIT)` per your sub-suggestion. Lifecycle is now explicit. c1708df.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
Luqmansonar: Changed to `return true` with the suggested comment. c1708df.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
Luqmansonar: Rewritten as suggested. c1708df.

java-checks-test-sources/default/src/main/java/checks/PseudoRandomCheckSecurityKeywordsSample.java
Luqmansonar: All four added in c1708df: `AES256_KEY` (Compliant, with comment documenting Dart-faithful behaviour), wildcard import sample, static import sample, inner-class isolation (`TokenAware` outer + `NeutralInner` inner). Tokenization gap (digit-suffix + `ChaCha20`-style) documented on the SONARJAVA-6440 description; cross-language follow-up tracked under APPSEC-3004.

its/autoscan/src/test/resources/autoscan/diffs/diff_S2440.json
Luqmansonar: Root cause: my new `PseudoRandomCheckNoContextSample.java` instantiated `new RandomUtils()` and `new RandomStringUtils()` — utility classes with only static methods, which **S2440 = `ClassWithOnlyStaticMethodsInstantiationCheck`** (not StringConcatenationInLoop) legitimately flags but autoscan misses without semantics. Dropped those instantiations and reverted `diff_S2440.json` to FN=2. c1708df.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
Luqmansonar: Update on this: I had to remove the explanatory comment in 29e9bef. SQS flagged the two-line block as S125 (Sections of code should not be commented out) — a false positive on the rule's part, since the lines are plain English, but the cleanest unblock was just to drop the comment. The constant name `CRYPTO_IMPORT_PREFIXES` plus the prefix list are reasonably self-evident on their own. Happy to reinstate a tighter wording if you'd like.

Luqmansonar:
Quick context on the autoscan baseline updates in commit 8123b0db. The failing rules were S1874 and S2119, not S2245. The new test samples added to exercise the heuristic include a few deprecated Apache Commons calls like RandomUtils.nextFloat and RandomStringUtils.random, plus one extra new Random constructor. Full analysis catches all of those, autoscan does not because it lacks semantics to resolve the Deprecated annotation and the Random type. That accounts for plus two on S1874, plus one on S2119, total three, matching the expected 4137 actual 4140 message. The S2245 baseline was already correct at 17 on the branch, so it stays as is.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -66,32 +79,171 @@ public class PseudoRandomCheck extends IssuableSubscriptionVisitor {
...
+      if (RANDOM_CONSTRUCTOR_TYPES.contains(newClass.symbolType().fullyQualifiedName())
+        && isInSecurityContext(newClass)) {
         reportIssue(newClass.identifier(), MESSAGE);
       }
     }
   }
 
+  @Override
+  public void leaveNode(Tree tree) {
+    if (tree.is(Tree.Kind.COMPILATION_UNIT)) {
```
nils-werner-sonarsource: **Convention mismatch: use `leaveFile()` instead of `leaveNode(COMPILATION_UNIT)` for per-file state reset**

Every other check in this repo that resets per-file state uses `leaveFile(JavaFileScannerContext)`, not `leaveNode`. Examples: `ArrayHashCodeAndToStringCheck`, `AbstractInSynchronizeChecker`, `AssertionsInProductionCodeCheck`, `BatchSQLStatementsCheck`, `CallOuterPrivateMethodCheck`, etc.

The current approach works (since `COMPILATION_UNIT` is in `nodesToVisit()`, `leaveNode` is called), but it's non-idiomatic and fragile: if `COMPILATION_UNIT` were ever removed from `nodesToVisit()`, the reset would silently stop firing and state would leak across files.

Suggested fix — drop the `leaveNode` override entirely and use `leaveFile` instead:
```java
@Override
  cryptoImportPresent = false;
  scopeSecurityContextCache.clear();
}
```
You can then also remove `COMPILATION_UNIT` from `nodesToVisit()` and the `if (tree.is(Tree.Kind.COMPILATION_UNIT))` guard in `visitNode`, simplifying the visit dispatch.

java-checks-test-sources/default/src/main/java/checks/PseudoRandomCheckWildcardImportSample.java
```diff
@@ -0,0 +1,22 @@
+package checks;
+
+import java.security.*;
+import java.util.Random;
+
+// SONARJAVA-6440 Part 1: wildcard import `java.security.*` -> all PRNG calls flagged.
+// ExpressionsHelper.concatenate returns "java.security.*", which still starts with the
```
nils-werner-sonarsource, line 7: ✓ Confirmed your reply is correct: `"java.security.*".startsWith("java.security.")` returns `true` in Java, so wildcard imports are already handled. My earlier comment was wrong — apologies for the noise. The test here is a nice explicit confirmation of this.

java-checks-test-sources/default/src/main/java/checks/PseudoRandomCheckSecurityKeywordsSample.java
```diff
@@ -0,0 +1,89 @@
...
+  // from trailing digits would help here AND in cases like `ChaCha20` (where the keyword
+  // `chacha20` is also unreachable today) but is out of scope for this faithful port.
+  // Tracked separately under APPSEC-3004.
+  void digitSuffixedAcronym() {
+    final int AES256_KEY = 32;
+    Random r = new Random(); // Compliant
+    r.nextInt(AES256_KEY);
+  }
+}
+
```
nils-werner-sonarsource, line 74: The `AES256_KEY` test and cross-reference to APPSEC-3004 is exactly the right way to document a known limitation — good call on adding the ticket reference.

nils-werner-sonarsource (commented):
Thanks for the thorough follow-up — most of the first-round feedback has been addressed well. A few remaining points below.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
Luqmansonar: Done.

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -66,32 +80,169 @@ public class PseudoRandomCheck extends IssuableSubscriptionVisitor {
...
+        if (SECURITY_KEYWORDS.contains(token)) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
+
+  // Mirrors Dart's `declarationScope`: the closest enclosing MethodTree (method/constructor) for
+  // local code; falls back to the enclosing ClassTree for field/initializer-block code.
```
nils-werner-sonarsource: Please remove the Dart reference. The comment should describe the Java behaviour, not its ancestry.

```java
// Scope for the keyword scan: use the enclosing method/constructor for local code;
// fall back to the enclosing class for field and static-initializer code.
private static Tree findDeclarationScope(Tree tree) {
```

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -66,32 +80,169 @@ public class PseudoRandomCheck extends IssuableSubscriptionVisitor {
...
+      if (current instanceof ClassTree classTree) {
+        return classTree;
+      }
+      current = current.parent();
+    }
+    return null;
+  }
+
+  // Mirrors Dart's `_splitIntoWords`: split on underscores first; for each part either keep it as
+  // a single lowercase word when all-uppercase, or split further on capital-letter boundaries.
```
nils-werner-sonarsource: Same here — drop `Mirrors Dart's \`_splitIntoWords\`:`.

```java
// Split on underscores first; for each part either keep it as a single lowercase word
// when all-uppercase, or split further on capital-letter boundaries.
static List<String> tokenizeIdentifier(String identifier) {
```

java-checks-test-sources/default/src/main/java/checks/PseudoRandomCheckSecurityKeywordsSample.java
```diff
@@ -0,0 +1,89 @@
...
+  // `randomBytes` tokenizes to [random, bytes]; neither is in the keyword set
+  // (the keyword `randombytes` only matches an all-lowercase or all-uppercase literal).
+  void splitRandomBytes() {
+    byte[] randomBytes = new byte[16];
+    Random r = new Random(); // Compliant
+    r.nextBytes(randomBytes);
+  }
+
+  // --- Digit-suffixed all-uppercase crypto acronym (Dart-faithful behaviour). ---
+  // `AES256_KEY` splits on `_` to ["AES256", "KEY"]. The "AES256" part has no lowercase
```
nils-werner-sonarsource: The block comment referencing `DART-276` and `Dart-faithful behaviour` should be rewritten without those references. The key fact for a reader is *what the behaviour is* and *why it is intentional*, not where the tokenizer was ported from.

Suggested replacement:
```java
// --- Digit-suffixed all-uppercase crypto acronym: documents a known tokenizer gap. ---
// `AES256_KEY` splits on `_` to ["AES256", "KEY"]. `AES256` has no lowercase letter so
// isAllUppercaseWithLetter returns true and it lowercases to `aes256` as a single token
// (NOT `aes` + `256`). The keyword set holds `aes`, not `aes256`, so no match fires.
// `KEY` -> `[key]`, which is not in the Java security-keyword set.
// Intentional: splitting letters from trailing digits is out of scope here; tracked under APPSEC-3004.
void digitSuffixedAcronym() {
```
(Remove the SONARKT-770 / `_splitIntoWords` references entirely — those belong in the Jira ticket.)

java-checks/src/main/java/org/sonar/java/checks/PseudoRandomCheck.java
```diff
@@ -66,32 +80,169 @@ public class PseudoRandomCheck extends IssuableSubscriptionVisitor {
...
+  );
+
+  private boolean cryptoImportPresent = false;
+  private final Map<Tree, Boolean> scopeSecurityContextCache = new IdentityHashMap<>();
+
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.NEW_CLASS, Tree.Kind.METHOD_INVOCATION);
+    return Arrays.asList(Tree.Kind.COMPILATION_UNIT, Tree.Kind.NEW_CLASS, Tree.Kind.METHOD_INVOCATION);
   }
```
nils-werner-sonarsource, line 105: Minor follow-up from the previous round: now that state reset is in `leaveFile`, `COMPILATION_UNIT` only needs to be in `nodesToVisit()` to set `cryptoImportPresent`. If you move that into `setContext` or `leaveFile`'s counterpart `setContext`, you could drop `COMPILATION_UNIT` from the list and remove the guard in `visitNode`. Not a blocker — happy to approve as-is — but it would simplify the visit dispatch.

nils-werner-sonarsource (commented):
Almost there — `leaveFile` is in, all other feedback addressed. One remaining ask: strip out the Dart/DART references from comments. The origin story belongs in the Jira ticket, not the code.

nils-werner-sonarsource (approved):

PR #5670, opened 2026-06-12
SONARJAVA-6465 : Rule S8714 - fix quickfix indexoutofbound / NPE exception

rombirli:
Despite [AssertJ doc pretends fail always has a string argument message](https://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/Fail.html), it is not the case in practice : causing a index out of bound exception

----
## Summary by Gitar

- **Fixed bugs:**
  - Resolved `IndexOutOfBoundsException` in `AssertThrowsInsteadOfTryCatchFailCheck` by handling `fail()` calls without arguments.
  - Added safeguard against `null` method owners during `fail()` invocation analysis.
  - Corrected `Throwable` to `Throwable.class` in expected type resolution.
- **Refactored logic:**
  - Rebuilt `JavaTextEdit` generation to handle diverse `TryStatementTree` configurations, including `try-with-resources` without `catch` or `finally`.
- **Dependencies:**
  - Updated `assertj-core` to `3.27.7` and updated test samples accordingly.

### DIFF
--- a/java-checks-test-sources/default/src/test/java/checks/tests/AssertionsCompletenessCheckSample.java
+++ b/java-checks-test-sources/default/src/test/java/checks/tests/AssertionsCompletenessCheckSample.java
@@ -9,6 +9,7 @@ import java.util.Optional;
 import java.util.OptionalInt;
 import java.util.stream.Stream;
 import org.assertj.core.api.junit.jupiter.SoftAssertionsExtension;
+import org.assertj.core.description.Description;
 import org.fest.assertions.BooleanAssert;
 import org.junit.Rule;
 import org.junit.Test;
@@ -123,8 +124,8 @@ public class AssertionsCompletenessCheckSample {
     // BDD
     org.assertj.core.api.BDDAssertions.thenRuntimeException(); // Noncompliant
     org.assertj.core.api.BDDAssertions.thenRuntimeException().isThrownBy(() -> System.out.println("b"));
-    org.assertj.core.api.BDDAssertions.thenRuntimeException().describedAs(null); // Noncompliant
-    org.assertj.core.api.BDDAssertions.thenRuntimeException().describedAs(null).isThrownBy(() -> System.out.println("b"));
+    org.assertj.core.api.BDDAssertions.thenRuntimeException().describedAs((Description) null); // Noncompliant
+    org.assertj.core.api.BDDAssertions.thenRuntimeException().describedAs((Description) null).isThrownBy(() -> System.out.println("b"));
 
 
     Comparator customComparator = null;
--- a/java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
@@ -189,7 +189,7 @@ public class AssertThrowsInsteadOfTryCatchFailCheck extends IssuableSubscription
 
 
   private static String typeClass(@Nullable Type caughtType) {
-    if (caughtType == null) return "Throwable";
+    if (caughtType == null) return "Throwable.class";
     return caughtType.name() + ".class";
   }
 
--- a/java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
@@ -24,16 +24,11 @@ import org.sonar.java.reporting.JavaQuickFix;
 import org.sonar.java.reporting.JavaTextEdit;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.semantic.Type;
-import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
-import org.sonar.plugins.java.api.tree.MethodInvocationTree;
-import org.sonar.plugins.java.api.tree.BlockTree;
-import org.sonar.plugins.java.api.tree.MethodTree;
-import org.sonar.plugins.java.api.tree.Tree;
-import org.sonar.plugins.java.api.tree.TryStatementTree;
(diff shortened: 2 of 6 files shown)


rombirli:
I wasn't able to add a unit test for this fix : 
- assertj fail without argument doesn't exist 
- adding a nonCompilingTest sample doesn't work as it doesnt have semantic and the junit test annotation is not resolved (so no issue raised in the test)

tomasz-tylenda-sonarsource (changes requested):
I'll wait for the fix for NPE.

java-checks-test-sources/default/src/test/java/checks/tests/AssertionsCompletenessCheckSample.java
```diff
@@ -123,8 +124,8 @@ public void assertj_assertions() {
     // BDD
     org.assertj.core.api.BDDAssertions.thenRuntimeException(); // Noncompliant
     org.assertj.core.api.BDDAssertions.thenRuntimeException().isThrownBy(() -> System.out.println("b"));
-    org.assertj.core.api.BDDAssertions.thenRuntimeException().describedAs(null); // Noncompliant
-    org.assertj.core.api.BDDAssertions.thenRuntimeException().describedAs(null).isThrownBy(() -> System.out.println("b"));
+    org.assertj.core.api.BDDAssertions.thenRuntimeException().describedAs((Description) null); // Noncompliant
```
tomasz-tylenda-sonarsource, line 127: This is because of AssertJ update, right?

java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
```diff
@@ -114,22 +114,36 @@ private JavaQuickFix provideQuickFix(
...
-      var lastCatchFinallyToken = tryStatement.finallyBlock() != null ?
-        tryStatement.finallyBlock().lastToken() :
-        tryStatement.catches().get(tryStatement.catches().size() - 1).block().lastToken();
-
-      return JavaQuickFix.newQuickFix(issueMessage).addTextEdit(
+      JavaTextEdit lastEdit;
+      if (!tryStatement.catches().isEmpty()) {
+        var start = tryStatement.catches().get(0).firstToken();
+        var end = tryStatement.finallyBlock() != null ?
+          tryStatement.finallyBlock().lastToken() :
```
tomasz-tylenda-sonarsource: My IDE still shows that `lastToken()` can produce NPE. Could you extract `BlockTree finallyBlock = tryStatement.finallyBlock();` and use it, so that the IDE does not show a warning?

java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
```diff
@@ -163,16 +177,15 @@ private Replacements assertJReplacement(
       TryStatementTree tryStatement,
       boolean isTryBlock
     ) {
-      // in assertJ the failure message is mandatory
-      var failureMessage = contentFor(failArguments.get(0));
+      var failureMessagePart = failArguments.stream().findFirst().map(this::contentFor).map(".withFailMessage(%s)"::formatted).orElse("");
```
tomasz-tylenda-sonarsource: I find this easier to read:
`failArguments.isEmpty() ? "" :  ".withFailMessage(%s)".formatted(contentFor(failArguments.get(0)));`

java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
tomasz-tylenda-sonarsource: Optional: I was thinking of changing this to `failMethodInvocation.methodSymbol().owner().type().fullyQualifiedName().startsWith("org.assertj")`, which seems like a more direct way to check what we need here.

java-checks-test-sources/default/src/test/java/checks/tests/AssertionsCompletenessCheckSample.java
rombirli: Yes it is

java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
rombirli: Good idea. It eliminated a few `... might be null` warnings.

java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
```diff
@@ -76,7 +76,12 @@ private void checkBlock(
     ) {
       UnitTestUtils.findFail(block).ifPresent(failMethodInvocation -> {
 
-          var isAssertJ = failMethodInvocation.methodSymbol().signature().contains("org.assertj");
+          var isAssertJ = failMethodInvocation
+            .methodSymbol()
+            .owner()
+            .type()
```
tomasz-tylenda-sonarsource, line 84: I just realized that NPE is possible here. Let's guard against if just in case.
```
          var failOwner = failMethodInvocation.methodSymbol().owner();
          if (failOwner == null) {
            return;
          }
          var isAssertJ = failOwner.(...)
```

tomasz-tylenda-sonarsource (commented):
This is still attached to the old ticket, right>

java-checks/src/main/java/org/sonar/java/checks/AssertThrowsInsteadOfTryCatchFailCheck.java
rombirli: It should not be possible as this point is not reached if owner is null (will not match any fail method in UnitTestUtils), but lets add this guard anyway for safety

tomasz-tylenda-sonarsource (approved):
LGTM

PR #5683, opened 2026-06-19
RIS-496 Implement S8795: Code after unconditional control flow statements should be removed

romainbrenguier:
Implements detection of unreachable code after unconditional control flow statements (return, throw, break, continue).

## Changes
- Added  - Rule implementation
- Added  - Unit tests  
- Added  - Test cases
- Added S8795 metadata (HTML/JSON)

## Related
- Jira: RIS-493
- RSpec: S8795

Generated using vibe-bot new-rule workflow


romainbrenguier:
This was only an experiment

PR #5684, opened 2026-06-22
[SONARJAVA-6484] Implement new rule S8899

romainbrenguier:
This rule detects a critical security vulnerability where HTTP request paths are used in security checks without normalizing multiple consecutive slashes. Attackers can bypass authentication and authorization by inserting extra slashes in URLs (e.g., //admin instead of /admin). The rule flags path-based security checks that lack proper normalization via replaceAll("/+", "/") and applies to both javax.servlet and jakarta.servlet frameworks, as well as JAX-RS UriInfo.

Part of 

----
## Summary by Gitar

- **New Sonar Rule:**
  - Implemented `S8899` to detect unnormalized HTTP request paths in security checks.
  - Added full rule specification, including metadata, HTML documentation, and test samples covering various frameworks.
- **Project Configuration:**
  - Added `.cursor/rules/sonar_code_context.mdc` and `.sonar-code-context/settings.json` to manage development quality standards.
  - Updated `.gitignore` to track critical Sonar configuration while excluding transient files.
  - Added `java-S8899.json` tracking files to multiple test projects in `its/ruling/`.
- **Refactoring:**
  - Extracted logic into `updatePathStateFromMethodInvocation` helper within `RequestPathNormalizationCheck.java`.
- **Environment Updates:**
  - Updated `mise.toml` to lock Java to version `21`.
  - Integrated rule into `Sonar_agentic_AI_profile.json` and `Sonar_way_profile.json` and adjusted `AutoScanTest` expectations.


.cursor/rules/sonar_code_context.mdc
romainbrenguier, line 1: This shouldn't be commited

.sonar-code-context/settings.json
romainbrenguier, line 1: This shouldn't be commited

.gitignore
```diff
@@ -39,3 +39,8 @@ Desktop.ini
 
 .java-version
 .mvn/.develocity/
+
+# Sonar Code Context configuration (track settings.json but ignore other files)
+!.sonar-code-context/
+.sonar-code-context/*
+!.sonar-code-context/settings.json
```
romainbrenguier:
```suggestion
```

.agents/RIS_PREPROCESSING.md
romainbrenguier, line 1: This should be a separate PR

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8899.html
romainbrenguier, line 1: The rule should be added to a quality profile

java-checks/src/main/java/org/sonar/java/checks/security/RequestPathNormalizationCheck.java
```diff
@@ -0,0 +1,234 @@
...
+      .names("normalize")
+      .addWithoutParametersMatcher()
+      .build()
+  );
+
+  private static final MethodMatchers SECURITY_CHECK_METHODS = MethodMatchers.create()
+    .ofTypes("java.lang.String")
+    .names("startsWith", "equals", "matches")
+    .withAnyParameters()
+    .build();
```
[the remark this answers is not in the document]
romainbrenguier: Good catch! I've expanded the rule to detect these alternative security check patterns:

**Added detection for:**
- Reversed String comparisons: `"/admin".equals(path)`, `"/admin".equalsIgnoreCase(path)`, `"/admin".regionMatches(..., path, ...)`
- `Objects.equals(path, "/admin")` in both argument positions
- `Pattern.matches(pattern, path)` where path appears as the second argument
- `String.equalsIgnoreCase` (both receiver and argument positions)

The implementation now checks not only the receiver of security check methods but also inspects arguments for unnormalized path expressions in RequestPathNormalizationCheck.java:145-189. Added comprehensive test coverage for all these patterns in the sample file, including both compliant (normalized) and non-compliant cases.

alex-meseldzija-sonarsource (changes requested):
This rule is useful, but it is a security rule.
I think this means we need to get sign off/reviews from the security domain.
WDYT?

romainbrenguier:
> This rule is useful, but it is a security rule. I think this means we need to get sign off/reviews from the security domain. WDYT?

This makes sense. I'll #ask-security-domain

romainbrenguier:
I'm withdrawing this rule for now, since it falls under the security domain.

PR #5685, opened 2026-06-22
SONARJAVA-6482 Change number of issues raised for S5786 to one per class

asya-vorobeva:
Instead of reporting a separate issue on each noncompliant modifier, the rule
now reports a single issue on the class name listing all offending modifiers as
secondaries. The quick fix removes all of them in one click.

See [SONARJAVA-6482](https://sonarsource.atlassian.net/browse/SONARJAVA-6482).

[SONARJAVA-6482]: https://sonarsource.atlassian.net/browse/SONARJAVA-6482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.claude/commands/quickfix-tests.md
rombirli, line 1: IMO this file shouldn't be part of this PR. 
Quickfixes testing should be improved in sonar-java (like in sonar-python, where they can compare the code after applying them). 
All the content of this skill is specific to our current way of testing quickfixes and comparing the list of edits

its/vibebot/pom.xml
```diff
@@ -11,6 +11,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.compiler.release>21</maven.compiler.release>
+    <slf4j.version>1.7.30</slf4j.version>
```
rombirli, line 14: Why ?

java-checks/src/main/java/org/sonar/java/checks/tests/JUnit5DefaultPackageClassAndMethodCheck.java
```diff
@@ -16,22 +16,113 @@
...
+  private static final String MESSAGE = "Remove redundant visibility modifiers from this test class and its methods.";
+  private static final String MODIFIER_MESSAGE = "Remove this '%s' modifier.";
+  private static final String QUICK_FIX_ALL = "Remove all redundant visibility modifiers";
 
   @Override
-  protected boolean isNonCompliantModifier(Modifier modifier, ModifierScope modifierScope) {
-    // All visibility modifiers except 'private' handled by S5810
-    return modifier == Modifier.PUBLIC || modifier == Modifier.PROTECTED;
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.CLASS);
```
rombirli, line 47: I personally prefer List.of

java-checks/src/main/java/org/sonar/java/checks/tests/JUnit5DefaultPackageClassAndMethodCheck.java
```diff
@@ -16,22 +16,113 @@
...
   }
 
   @Override
-  protected void raiseIssueOnNonCompliantReturnType(MethodTree methodTree) {
-    // Handled by S5810
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+    if (classTree.symbol().isAbstract() || (classTree.simpleName() == null)) {
```
rombirli, line 53: redundant parentheses 
```suggestion
    if (classTree.symbol().isAbstract() || classTree.simpleName() == null) {
```

java-checks/src/main/java/org/sonar/java/checks/tests/JUnit5DefaultPackageClassAndMethodCheck.java
```diff
@@ -16,22 +16,113 @@
...
+        .map(m -> new JavaFileScannerContext.Location(String.format(MODIFIER_MESSAGE, m.keyword().text()), m))
+        .toList();
+      List<ModifierKeywordTree> modifiersForFix = new ArrayList<>(noncompliantModifiers);
+      QuickFixHelper.newIssue(context)
+        .forRule(this)
+        .onTree(classTree.simpleName())
+        .withMessage(MESSAGE)
+        .withSecondaries(secondaries)
+        .withQuickFixes(() -> List.of(buildQuickFix(modifiersForFix)))
+        .report();
```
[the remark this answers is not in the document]
rombirli: I agree with it. It would also be worth trying to manually apply quickfixes to see the resulting code after all edits.

java-checks/src/main/java/org/sonar/java/checks/tests/JUnit5DefaultPackageClassAndMethodCheck.java
```diff
@@ -16,22 +16,113 @@
...
+    modifiers.modifiers().stream()
+      .filter(m -> isNonCompliantModifier(m.modifier()))
+      .findFirst()
+      .ifPresent(result::add);
+  }
+
+  private static JavaQuickFix buildQuickFix(List<ModifierKeywordTree> modifiers) {
+    List<JavaTextEdit> edits = modifiers.stream()
+      .map(m -> JavaTextEdit.removeTextSpan(
+        AnalyzerMessage.textSpanBetween(m, true, QuickFixHelper.nextToken(m), false)))
```
rombirli, line 114: This map is a bit hard to read, all closing parentheses are on the same line, i propose: 
```suggestion
      .map(m ->
        AnalyzerMessage.textSpanBetween(m, true, QuickFixHelper.nextToken(m), false)
      )
      .map(JavaTextEdit::removeTextSpan)
```

java-checks/src/main/java/org/sonar/java/checks/tests/JUnit5SilentlyIgnoreClassAndMethodCheck.java
```diff
@@ -19,33 +19,107 @@
...
+    INSTANCE_METHOD
+  }
+
+  private static final String WRONG_MODIFIER_ISSUE_MESSAGE = "Remove this '%s' modifier.";
 
   @Override
-  protected boolean isNonCompliantModifier(Modifier modifier, ModifierScope modifierScope) {
-    return modifier == Modifier.PRIVATE || (modifierScope == ModifierScope.INSTANCE_METHOD && modifier == Modifier.STATIC);
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.CLASS);
```
rombirli, line 57: I prefer List.of

java-checks/src/main/java/org/sonar/java/checks/tests/AbstractJUnit5NotCompliantModifierChecker.java
rombirli, line 1: A few things are now duplicated : 
`MODIFIER_MESSAGE`, `nodesToVisit`, `UnitTestUtils.groupJUnit5Methods` and `ClassPatternsUtils.shouldBePublicClass` are only used in `JUnit5DefaultPackageClassAndMethodCheck` and `JUnit5SilentlyIgnoreClassAndMethodCheck` 
IMO it makes sense to keep this superclass even if it only contains a few things.

rombirli: I think both rules should follow the same direction here.

  S5786 was moved to a bulk quick-fix model, while S5810 still keeps a separate per-issue/per-modifier behavior. In my view, that creates unnecessary divergence between two checks that still share the same core traversal and JUnit 5 classification logic.

  I would prefer:

  - keeping AbstractJUnit5NotCompliantModifierChecker with the shared visit / grouping / class-handling logic;
  - aligning S5810 with S5786 on the reporting model, with a bulk quick fix for the noncompliant modifiers in the same test class.

The shared abstraction still looks valid to me, and keeping it would make future changes safer.

rombirli (changes requested):
I’m requesting changes because I don’t understand why these two rules diverged this much, or why the common abstraction was removed.

They still seem to share the same core traversal and JUnit 5 classification logic, so I would expect them to follow the same direction and keep the shared parent class.

Could you either keep the common abstraction, or explain why this level of divergence is necessary?

The other comments are personal preferences and optional.

rombirli (approved):
Approved after discussion on why these two rules need to diverge for business reasons.

.claude/commands/quickfix-tests.md
asya-vorobeva: Yep, but it's a good case when we can add it. I don't want to create separate PR for such little change (and I think that adding new skills can be iterative process, on-demand).

java-checks/src/main/java/org/sonar/java/checks/tests/JUnit5DefaultPackageClassAndMethodCheck.java
asya-vorobeva: Because this project is separated one and don't inherit root project's properties.

java-checks/src/main/java/org/sonar/java/checks/tests/JUnit5SilentlyIgnoreClassAndMethodCheck.java
asya-vorobeva: Me too, but it's our style among the whole project.

java-checks/src/main/java/org/sonar/java/checks/tests/AbstractJUnit5NotCompliantModifierChecker.java
asya-vorobeva: Was clarified in Slack discussion.

francois-mora-sonarsource (approved):

PR #5687, opened 2026-06-22
SONARJAVA-6488 Implement new rule S8908: Methods annotated with @CacheResult should not return void

romainbrenguier:
## Summary
- Implement new rule S8908 to detect Quarkus `@CacheResult` annotation used on void methods
- Update ruling baseline for S8899 in eclipse-jetty project
- Address review feedback on RequestPathNormalizationCheck (S8899)

### New Rule: S8908
Detects methods annotated with `@CacheResult` that return void. Since void methods have no return value to cache, applying `@CacheResult` to them is meaningless and indicates a misunderstanding of how caching works.

**Rule details:**
- Type: CODE_SMELL
- Severity: Minor
- Impact: MAINTAINABILITY (LOW)
- Fix time: 5min

**Implementation:**
- Added `QuarkusCacheResultOnVoidMethodCheck` to detect the pattern
- Includes comprehensive test cases covering various scenarios
- Proper documentation with examples and Quarkus documentation references

### S8899 Updates
- Updated ruling baseline for eclipse-jetty (1 new issue detected)
- Addressed security-context detection review feedback to reduce false positives
- Added additional test cases for non-security string operations

## Test plan
- [x] Unit tests for S8908 pass
- [x] Ruling baselines updated
- [ ] CI checks pass

----
## Summary by Gitar

- **New Rule: S8908:**
  - Implemented `QuarkusCacheResultOnVoidMethodCheck` and added corresponding metadata, documentation, and test cases.
- **CI/Test Infrastructure:**
  - Updated `Sonar_way_profile.json` to include `S8908` in the default rule set.
  - Registered `diff_S8908.json` and updated `autoscan-diff-by-rules.json` to reflect changes in integration testing baselines.


.sonar-code-context/SONAR.md
romainbrenguier, line 1: Delete this file from the PR.

java-checks-test-sources/default/src/main/files/non-compiling/checks/QuarkusCacheResultOnVoidMethodCheckSample.java
```diff
@@ -0,0 +1,21 @@
+package checks;
+
+import io.quarkus.cache.CacheResult;
+
+public class QuarkusCacheResultOnVoidMethodCheckSample {
+
+  @CacheResult(cacheName = "my-cache")
+  public void processData(UnknownType key) { // Noncompliant
```
rombirli: It could be slightly more clear to report on the whole `CacheResule` annotation instead of the void keyword but up to you to decide that

java-checks-test-sources/default/src/main/java/io/quarkus/cache/CacheInvalidate.java
rombirli, line 1: What is the purpose of this file? (and the 2 other files in this folder)? Could we move the annotations declarations to the sample (simpy as nested classes)

its/autoscan/src/test/resources/autoscan/diffs/diff_S8908.json
```diff
@@ -0,0 +1,6 @@
+{
+  "ruleKey": "S8908",
+  "hasTruePositives": true,
+  "falseNegatives": 0,
```
rombirli, line 4: There should be FN because when semantic annotation are not available the `CacheResult` annotation is not resolved correctly. Reading the rule implementation I couldn't figure out how this works anyway. There is an early return in case the symbol is unknow. Could you add a test with 
`.withoutSemantic().verifyNoIssue()` to figure out if it works as expected please?

rombirli (approved):
The implementation is clear, looks solid, but there is a point about FN that i dont understand. Usually rules that dont work without semantic informations have FNs. Could we add a unit test to ensure that without semantic information this rule doesnt works, as we expected it doesnt work in this case

java-checks-test-sources/default/src/main/java/io/quarkus/cache/CacheInvalidate.java
romainbrenguier: It is a mockup of the CacheInvalidate annotation, that is needed in the samples to not have to import quarkus. It can't be in the same file as the sample, because it should be in the io.quarkus.cache package for name matching.

its/autoscan/src/test/resources/autoscan/diffs/diff_S8908.json
romainbrenguier: I'm adding a test without semantic and it doesn't raise issues in that case. My understanding is that with autoscan we can still get some semantic in some cases.

PR #5688, opened 2026-06-22
SONARJAVA-6504 Remove S1192 from Sonar way and Sonar agentic AI

asya-vorobeva:
See [SONARJAVA-6504](https://sonarsource.atlassian.net/browse/SONARJAVA-6504).

[SONARJAVA-6504]: https://sonarsource.atlassian.net/browse/SONARJAVA-6504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

----
## Summary by Gitar

- **Rule configuration:**
  - Removed rule `java:S1192` from `Sonar_way_profile.json` and `Sonar_agentic_AI_profile.json`.
- **Test updates:**
  - Deleted `diff_S1192.json` test resource from `its/autoscan` as the rule is no longer active.


erwan-leforestier-sonarsource:
I'm a bit skeptical about removing this rule from the default profile. Duplicating string literals several times in the code is really bad for maintainability and this is, I would say, one of the top 10 rules that helps keeping code clean. Now considering the "why" developers are not fixing this rule, I wonder if it's not due to context. As a developer, it happened to me to duplicate string literals inside tests because it makes the test easier to read. This is a personal preference, some others would argue that "test" is code so it should remain clean as well.
  
I had a look to the rule definition and the scope is "MAIN" so it shouldn't run on "TEST" files, now I don't know if this is taken into consideration in sonar-java, I think it's worth having a look.

Running on test files is a hypothesis, it could be something else but I would think twice before disabling this rule.

asya-vorobeva:
Agree with the points, decided to narrow this rule to the main sources' scope.

PR #5689, opened 2026-06-23
SONARJAVA-5708 Implement rule S8924: Mockito core methods should be imported statically

NoemieBenard:
----
## Summary by Gitar

  - Implemented `MockitoStaticImportCheck` (rule `S8924`) to enforce static imports for common Mockito methods.
- **Check coverage:**
  - Detects non-compliant usage of methods like `mock`, `when`, `verify`, `spy`, `any`, and `eq` when called via `Mockito.` qualifier.
- **Testing:**
  - Added `MockitoStaticImportCheckTest` and corresponding sample file `MockitoStaticImportCheckSample.java` to verify rule reporting.


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8924.json
```diff
@@ -0,0 +1,25 @@
...
+  "tags": [
+    "mockito",
+    "tests",
+    "convention"
+  ],
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-8924",
+  "sqKey": "S8924",
+  "scope": "Tests",
+  "quickfix": "targeted",
```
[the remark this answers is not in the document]
NoemieBenard: "Targeted" means that the quickfix is feasible and can be implemented in the future. It is different from "covered" which means that the quickfix is implemented.

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoStaticImportCheck.java
```diff
@@ -0,0 +1,56 @@
...
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S8924")
+public class MockitoStaticImportCheck extends IssuableSubscriptionVisitor {
+
+  private static final String MOCKITO_CLASS = "org.mockito.Mockito";
+
+  private static final Set<String> MOCKITO_METHODS = Set.of(
+    "doReturn", "doThrow", "mock", "never", "spy", "times", "verify", "when"
+  );
```
rombirli: Optional : it would better to create method matchers like [here](https://github.com/SonarSource/sonar-java/blob/643e39b208e5cc0f34fb6e5730843d2e0509c99f/java-checks/src/main/java/org/sonar/java/checks/ConfusingVarargCheck.java#L38)

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoStaticImportCheck.java
```diff
@@ -0,0 +1,56 @@
...
+  @Override
+  public void visitNode(Tree tree) {
+    MethodInvocationTree mit = (MethodInvocationTree) tree;
+    ExpressionTree methodSelect = mit.methodSelect();
+    if (!methodSelect.is(Tree.Kind.MEMBER_SELECT)) {
+      return;
+    }
+    MemberSelectExpressionTree mset = (MemberSelectExpressionTree) methodSelect;
+    String methodName = mset.identifier().name();
+    if (MOCKITO_METHODS.contains(methodName) && mset.expression().symbolType().is(MOCKITO_CLASS)) {
```
rombirli: this would become 

```java
MOCKITO_METHODS.matches(mit.methodSymbol())
```

rombirli (approved):
LGTM, small improvement suggestion

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoStaticImportCheck.java
```diff
@@ -0,0 +1,56 @@
...
+      reportIssue(methodSelect, "Use a static import for \"" + methodName + "\".");
```
rombirli:
```suggestion
      reportIssue(methodSelect, "Use a static import for %s\".".formatted(methodName));
```

PR #5692, opened 2026-06-23
SONARJAVA-6510 Build with Java 26

alex-meseldzija-sonarsource:
Please compare with the [Java 25 upgrade PR](https://github.com/SonarSource/sonar-java/pull/5390/changes)


mise.toml
```diff
@@ -1,3 +1,3 @@
 [tools]
-java = ["25", "21", "17"]
+java = ["26", "25", "21", "17"]
```
romainbrenguier: Do we still need Java 25 for something?

romainbrenguier (approved):
🎉

mise.toml
alex-meseldzija-sonarsource: Good question. JDK 25 is an LTS so i though to keep it because thats what we do in dotnet.
But i asked pavel and he says its unnecessary :D.
So i will delete this before merge.

PR #5694, opened 2026-06-23
SONARJAVA-6491: Implement S8911 - Methods annotated with @Startup should be non-static, non-producer, and parameter-free

romainbrenguier:
## Summary

This PR implements rule S8911 to detect methods annotated with `@Startup` (from `io.quarkus.runtime.Startup`) that violate CDI initialization requirements by being static, having parameters, or being annotated with `@Produces`.

## Rule Behavior

The rule detects methods annotated with `@Startup` that:
1. Are declared as `static` (static methods cannot be managed by CDI lifecycle)
2. Have one or more parameters (framework cannot provide arguments during automatic invocation)
3. Are annotated with `@Produces` (producer methods serve a different purpose than initialization)

## Implementation Approach

The implementation follows the pattern established by `ScheduledOnlyOnNoArgMethodCheck` (S7184):

1. Extends `IssuableSubscriptionVisitor` to subscribe to `Tree.Kind.METHOD` nodes
2. For each method, checks if it has the `@Startup` annotation (`io.quarkus.runtime.Startup`)
3. If `@Startup` is present, checks for three violations:
   - Method is static (using `methodTree.symbol().isStatic()`)
   - Method has parameters (using `!methodTree.parameters().isEmpty()`)
   - Method has `@Produces` annotation (checking for `jakarta.enterprise.inject.Produces`)
4. Reports issues with descriptive messages indicating which requirement is violated
5. Includes secondary locations pointing to the `@Startup` annotation for context

## Test Coverage

The test file covers:

### Noncompliant patterns:
- Static method with `@Startup`
- Method with `@Startup` and parameters
- Method with both `@Startup` and `@Produces`
- Multiple violations on same method (static + parameters, static + producer, etc.)

### Compliant patterns:
- Non-static, parameter-free method with `@Startup` (valid usage)
- Methods using dependency injection (`@Inject` fields) instead of parameters
- Methods without `@Startup` annotation (static, producer, or parameterized methods are OK)
- Private methods with `@Startup` (access level doesn't matter for this rule)

## Additional Changes

- Added mock `io.quarkus.runtime.Startup` annotation in `java-checks-test-sources/default/src/main/java/io/quarkus/runtime/Startup.java` following the pattern used for other external dependencies (similar to `io.realm.RealmConfiguration`)
- Test sample uses existing Jakarta dependencies (`jakarta.enterprise.cdi-api` and `javax.inject`) already available in the test sources pom

## Resources

- [Quarkus Application Lifecycle Guide](https://quarkus.io/guides/lifecycle)
- [Jakarta CDI Specification](https://jakarta.ee/specifications/cdi/)


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/Sonar_agentic_AI_profile.json
rombirli, line 1: [Rspec PR](https://github.com/SonarSource/rspec/pull/7152/changes#diff-de722320e2e762504eb80a52f56b8891ffe8c3076d3524802ac0208edd9e33adR19) says it's only in Sonar-Way which seems correct. Could you update rule metadata to exclude it from agentic ai profile

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaAgenticWayProfileTest.java
```diff
@@ -53,7 +53,7 @@ void profile_is_registered_as_expected() {
     BuiltInQualityProfilesDefinition.BuiltInQualityProfile actualProfile = profilesPerLanguages.get("java").get("Sonar agentic AI");
     assertThat(actualProfile.isDefault()).isFalse();
     assertThat(actualProfile.rules())
-      .hasSize(465)
+      .hasSize(466)
```
rombirli: Revert

java-checks/src/main/java/org/sonar/java/checks/StartupAnnotationCheck.java
```diff
@@ -0,0 +1,91 @@
...
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.tree.AnnotationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S8911")
+public class StartupAnnotationCheck extends IssuableSubscriptionVisitor {
+  private static final String STARTUP_FQN = "io.quarkus.runtime.Startup";
+  private static final String PRODUCES_FQN = "jakarta.enterprise.inject.Produces";
```
rombirli, line 30: Usually we prefer matchers

java-checks/src/main/java/org/sonar/java/checks/StartupAnnotationCheck.java
```diff
@@ -0,0 +1,91 @@
...
+    MethodTree methodTree = (MethodTree) tree;
+    var startupAnnotations = getStartupAnnotations(methodTree);
+
+    if (startupAnnotations.isEmpty()) {
+      return;
+    }
+
+    var secondaryLocations = startupAnnotations.stream()
+      .map(annotation ->
+        new JavaFileScannerContext.Location("Triggered by this annotation", annotation.annotationType()))
```
rombirli: code style suggestion

```suggestion
      .map(annotation ->
        new JavaFileScannerContext.Location("Triggered by this annotation", annotation.annotationType())
      )
```

java-checks/src/main/java/org/sonar/java/checks/StartupAnnotationCheck.java
```diff
@@ -0,0 +1,91 @@
...
+      return;
+    }
+
+    var producesAnnotations = getProducesAnnotations(methodTree);
+    if (!producesAnnotations.isEmpty()) {
+      reportIssue(
+        methodTree.simpleName(),
+        "\"@Startup\" annotation should not be applied to producer methods",
+        secondaryLocations,
+        null);
```
rombirli: code style suggestion

```suggestion
      reportIssue(
        methodTree.simpleName(),
        "\"@Startup\" annotation should not be applied to producer methods",
        secondaryLocations,
        null
      );
```

java-checks/src/main/java/org/sonar/java/checks/StartupAnnotationCheck.java
```diff
@@ -0,0 +1,91 @@
...
+      return;
+    }
+
+    if (!methodTree.parameters().isEmpty()) {
+      reportIssue(
+        methodTree.simpleName(),
+        "\"@Startup\" annotation should only be applied to no-arg methods",
+        secondaryLocations,
+        null);
```
rombirli:
```suggestion
      reportIssue(
        methodTree.simpleName(),
        "\"@Startup\" annotation should only be applied to no-arg methods",
        secondaryLocations,
        null
      );
```

java-checks/src/main/java/org/sonar/java/checks/StartupAnnotationCheck.java
```diff
@@ -0,0 +1,91 @@
...
+        new JavaFileScannerContext.Location("Triggered by this annotation", annotation.annotationType()))
+      .toList();
+
+    // Only report one issue per method, prioritizing: static > producer > parameters
+    if (methodTree.symbol().isStatic()) {
+      reportIssue(
+        methodTree.simpleName(),
+        "\"@Startup\" annotation should not be applied to static methods",
+        secondaryLocations,
+        null);
```
rombirli, line 58:
```suggestion
      reportIssue(
        methodTree.simpleName(),
        "\"@Startup\" annotation should not be applied to static methods",
        secondaryLocations,
        null
      );
```

java-checks/src/main/java/org/sonar/java/checks/StartupAnnotationCheck.java
rombirli: I don't know if it's possible, I never did it for annotations. Optional : let's give it a try with an agent

its/autoscan/src/test/resources/autoscan/diffs/diff_S6813.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S6813",
   "hasTruePositives": true,
-  "falseNegatives": 65,
+  "falseNegatives": 66,
```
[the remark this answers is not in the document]
rombirli: Raised because of : https://github.com/SonarSource/sonar-java/pull/5694/changes#diff-7c98e778e7fe5793d8ef3c1f013690dde172fe5c4333c4ff8d26f7d41c0549ceR65
seems correct

rombirli (approved):
LGTM. Please be careful about https://github.com/SonarSource/rspec/pull/7152#discussion_r3472686968 during validation

java-checks/src/main/java/org/sonar/java/checks/StartupAnnotationCheck.java
romainbrenguier: After exploring the codebase, the current pattern of using `symbolType().is(FQN)` for annotation checking is actually the standard approach used throughout the java-checks module.

`MethodMatchers` is specifically designed for matching method invocations by type, name, and parameters - it doesn't support annotation matching. For annotation checks, the codebase consistently uses either:
1. `annotation.annotationType().symbolType().is(FQN)` when working with `AnnotationTree` (syntax tree level)
2. `symbol.metadata().isAnnotatedWith(FQN)` for simple presence checks

The current implementation follows pattern #1, which is appropriate here since we need to create secondary locations pointing to the specific `AnnotationTree` objects. Examples using the same pattern include:
- `TransactionalMethodVisibilityCheck.java:69-74`
- `NullableInjectedFieldsHaveDefaultValueCheck.java:161-165`

Given that your follow-up comment indicates this is optional and you're uncertain if matchers work for annotations, I believe the current implementation is correct as-is.

romainbrenguier: After searching the codebase, there doesn't appear to be a built-in annotation matcher utility (like `MethodMatchers` for method calls). Other similar annotation-based checks (e.g., `ScheduledOnlyOnNoArgMethodCheck.java:66`) use the same pattern: `.filter(annotation -> annotation.annotationType().symbolType().is(FQN))`.

Creating an annotation matcher would be a larger architectural addition. Since this is marked optional, I'm keeping the current approach which follows the established pattern in the codebase.

PR #5695, opened 2026-06-23
SONARJAVA-6506 Avoid suggesting records for DTO contracts

francois-mora-sonarsource:
## Summary
- prevent S6206 from suggesting records for classes that participate in Java serialization contracts
- skip the rule when class, constructor, fields, or getter methods carry known framework annotations or unresolved annotations
- narrow the Spring-related prefix handling and add regression coverage to avoid over-skipping unrelated annotations

## Root cause
S6206 only checked structural record eligibility. That missed cases where converting a class to a record would change runtime behavior because serialization hooks or framework metadata define part of the class contract.

## Validation
- `mvn -pl java-checks -am -Dtest=org.sonar.java.checks.RecordInsteadOfClassCheckTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test`
- `git diff --check`

## RSPEC
https://github.com/SonarSource/rspec/pull/7177


java-checks/src/test/java/org/sonar/java/checks/RecordInsteadOfClassCheckTest.java
```diff
@@ -33,6 +33,16 @@ void test() {
       .verifyIssues();
   }
 
+  @Test
+  void test_framework_annotation_prefix_scope() {
+    CheckVerifier.newVerifier()
```
asya-vorobeva, line 38: We store all new test samples in java-checks-test-sources. The ones still located in java-checks/src/test/files are remainders from legacy style which should be moved some day. See .claude/commands/migrate-test-samples.md skill which describes this procedure.

asya-vorobeva (changes requested):
Makes sense also to update RSpec desc for this rule and shortly list exceptions.

java-checks/src/test/java/org/sonar/java/checks/RecordInsteadOfClassCheckTest.java
francois-mora-sonarsource: Thanks, I have moved the whole file to `java-checks-test-sources`.

java-checks/src/test/java/org/sonar/java/checks/RecordInsteadOfClassCheckTest.java
```diff
@@ -33,6 +34,16 @@ void test() {
...
+      .onFile(mainCodeSourcesPath("checks/RecordInsteadOfClassCheckPackagePrefixSample.java"))
+      .withCheck(new RecordInsteadOfClassCheck())
+      .withClassPath(DEFAULT_MODULE.getClassPath())
```
asya-vorobeva: DEFAULT_MODULE is automatically used when not specified. So you can remove this line

asya-vorobeva (approved):
Generally LGTM, one little comment.

java-checks/src/test/java/org/sonar/java/checks/RecordInsteadOfClassCheckTest.java
francois-mora-sonarsource: Removed, thanks.

PR #5696, opened 2026-06-23
SONARJAVA-6512 Enable analysis of Java 26

alex-meseldzija-sonarsource:
Part of 

----
## Summary by Gitar

- **Java 26 support:**
  - Updated `JavaVersionImpl` and `JavaVersion` to support Java 26 as the maximum supported version.
  - Added `isJava26Compatible` predicate to check for Java 26 compatibility.
  - Updated `JavaVersionImplTest` to include validation for Java 26 and updated default effective version to 26.

### DIFF
--- a/java-frontend/src/main/java/org/sonar/plugins/java/api/JavaVersion.java
+++ b/java-frontend/src/main/java/org/sonar/plugins/java/api/JavaVersion.java
@@ -177,7 +177,7 @@ public interface JavaVersion {
    * Test if java version of the project is greater than or equal to 26.
    * Remark - Contrary to other isJava*Compatible methods, this one will NOT return true if version is not set
    * @return true if java version used is >= 26
-   * @since SonarJava 8.32: Support of Java 26
+   * @since SonarJava 8.34: Support of Java 26
    */
   boolean isJava26Compatible();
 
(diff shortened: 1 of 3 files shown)


java-frontend/src/main/java/org/sonar/plugins/java/api/JavaVersion.java
```diff
@@ -173,6 +173,14 @@ public interface JavaVersion {
    */
   boolean isJava25Compatible();
 
+  /**
+   * Test if java version of the project is greater than or equal to 26.
+   * Remark - Contrary to other isJava*Compatible methods, this one will NOT return true if version is not set
+   * @return true if java version used is >= 26
+   * @since SonarJava 8.32: Support of Java 26
```
rombirli: Current release is [8.33.0.44481](https://github.com/SonarSource/sonar-java/releases/tag/8.33.0.44481). 

```suggestion
   * @since SonarJava 8.34: Support of Java 26
```

rombirli (approved):

java-frontend/src/main/java/org/sonar/plugins/java/api/JavaVersion.java
rombirli: https://github.com/SonarSource/sonar-java/blob/master/pom.xml#L13

PR #5698, opened 2026-06-23
SONARJAVA-6511 Decrease number of FPs for the rule S125

asya-vorobeva:
Semicolons in the end of comment's lines can be punctuation marks and not code indicators. We have a lot of user feedback pointing on it.

This PR removes FPs for comments where **the only** sign of code is that line ends with ';'. Comments that additionally have some indicators like CamelCase or keywords are still recognized as code.
This change unavoidably adds some amount of FNs. To reduce them the pattern is tuned to react on typical Java commented messages like System.out(err).print.

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/CommentedCode.java
+++ b/java-checks-test-sources/default/src/main/java/checks/CommentedCode.java
@@ -110,6 +110,24 @@ public class CommentedCode {
      * }
      */
     int a;
+
+    // This is a legal comment that ends with semi-colon;
+
+// Noncompliant@+2
+    // uses System/.out/.println
+    // System.out.println("testit");
+
+// Noncompliant@+2
+    // uses System/.err/.println
+    // System.err.println("err");
+
+// Noncompliant@+2
+    // ends by ';' and uses CamelCase
+    // assertEquals(a, "smth");
+
+// Noncompliant@+2
+    // ends by ';' and uses int keyword
+    // int i = 1;
   }
 
   // TODo
--- a/java-checks/src/main/java/org/sonar/java/checks/JavaFootprint.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/JavaFootprint.java
@@ -30,12 +30,14 @@ public final class JavaFootprint implements LanguageFootprint {
   private final Set<Detector> detectors = new HashSet<>();
 
   public JavaFootprint() {
-    detectors.add(new EndWithDetector(0.95, '}', ';', '{'));
+    detectors.add(new EndWithDetector(0.95, '}', '{'));
+    // add detection of ';' at the end of a line, it is less than a threshold used in the detector to avoid false positives
+    detectors.add(new EndWithDetector(0.87, ';'));
     detectors.add(new KeywordsDetector(0.7, "++", "||", "&&"));
     detectors.add(new KeywordsDetector(0.3, "public", "abstract", "class", "implements", "extends", "return", "throw",
         "private", "protected", "enum", "continue", "assert", "package", "synchronized", "boolean", "this", "double", "instanceof",
         "final", "interface", "static", "void", "long", "int", "float", "super", "true", "case:"));
-    detectors.add(new ContainsDetector(0.95, "for(", "if(", "while(", "catch(", "switch(", "try{", "else{"));
+    detectors.add(new ContainsDetector(0.95, "for(", "if(", "while(", "catch(", "switch(", "try{", "else{", "System.out.print", "System.err.print"));
     detectors.add(new CamelCaseDetector(0.5));
   }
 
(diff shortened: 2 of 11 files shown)


java-checks/src/main/java/org/sonar/java/checks/JavaFootprint.java
```diff
@@ -30,12 +30,13 @@ public final class JavaFootprint implements LanguageFootprint {
   private final Set<Detector> detectors = new HashSet<>();
 
   public JavaFootprint() {
-    detectors.add(new EndWithDetector(0.95, '}', ';', '{'));
+    detectors.add(new EndWithDetector(0.95, '}', '{'));
+    detectors.add(new EndWithDetector(0.9, ';'));
```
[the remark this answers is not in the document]
asya-vorobeva: Can we replace 0.9 for this EndWIthDetector with 0.85?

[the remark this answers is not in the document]
asya-vorobeva: Let's tune it to 0.87. It will fix tests.

java-checks/src/main/java/org/sonar/java/checks/JavaFootprint.java
```diff
@@ -30,12 +30,14 @@ public final class JavaFootprint implements LanguageFootprint {
   private final Set<Detector> detectors = new HashSet<>();
 
   public JavaFootprint() {
-    detectors.add(new EndWithDetector(0.95, '}', ';', '{'));
+    detectors.add(new EndWithDetector(0.95, '}', '{'));
+    // add detection of ';' at the end of a line, it is less than a threshold used in the detector to avoid false positives
+    detectors.add(new EndWithDetector(0.87, ';'));
```
erwan-leforestier-sonarsource, line 35: With `;` at 0.87, comments like `// foo = bar;` or `// list.add(x);` no longer get flagged. No keyword, no CamelCase, nothing else fires. Did you check the FN impact on a real corpus?

java-checks/src/main/java/org/sonar/java/checks/JavaFootprint.java
```diff
@@ -30,12 +30,14 @@ public final class JavaFootprint implements LanguageFootprint {
...
     detectors.add(new KeywordsDetector(0.7, "++", "||", "&&"));
     detectors.add(new KeywordsDetector(0.3, "public", "abstract", "class", "implements", "extends", "return", "throw",
         "private", "protected", "enum", "continue", "assert", "package", "synchronized", "boolean", "this", "double", "instanceof",
         "final", "interface", "static", "void", "long", "int", "float", "super", "true", "case:"));
-    detectors.add(new ContainsDetector(0.95, "for(", "if(", "while(", "catch(", "switch(", "try{", "else{"));
+    detectors.add(new ContainsDetector(0.95, "for(", "if(", "while(", "catch(", "switch(", "try{", "else{", "System.out.print", "System.err.print"));
```
erwan-leforestier-sonarsource, line 40: `ContainsDetector` will match prose too. `// Use System.out.println to debug this` will now be flagged as commented-out code. Maybe require a `(` after, or use `EndWithDetector`?

java-checks-test-sources/default/src/main/java/checks/CommentedCode.java
```diff
@@ -110,6 +110,24 @@ public CommentedCode(int field) {
...
+    // uses System/.out/.println
+    // System.out.println("testit");
+
+// Noncompliant@+2
+    // uses System/.out/.err
+    // System.err.println("err");
+
+// Noncompliant@+2
+    // ends by ';' and uses CamelCase
+    // assertEquals(a, "smth);
```
erwan-leforestier-sonarsource: Missing closing quote: `"smth)` should be `"smth")`.

java-checks/src/main/java/org/sonar/java/checks/JavaFootprint.java
asya-vorobeva: Yep, I've checked. As I noted in PR desc, some FNs are unavoidable. We can't get rid of them. I think that in this case it's ok, we prefer to increase FNs but decrease FPs generally.

java-checks/src/main/java/org/sonar/java/checks/JavaFootprint.java
asya-vorobeva: The problem is that you can always find an exception. Say // Use System.out.println() to debug will still be raised. Initially I wanted to use RegexDetector to check that this thing is in the beginning of the string, but it's also not super-reliable.

asya-vorobeva: After observing the code in our ruling projects I see that huge amount of real commented code looks like:
// System.out.println("smth");

So this tuning is kind of compensation for considering ; in the end as not reliable indicator.

erwan-leforestier-sonarsource (approved):

PR #5701, opened 2026-06-24
SONARJAVA-6473 Fix FP in S3626: return inside try-finally in a loop

francois-mora-sonarsource:
## Summary

- S3626 (RedundantJumpCheck) was incorrectly flagging `return;` statements inside `try` blocks as redundant when the `try` was wrapped in a `finally` clause inside a loop. Both the return path and the fall-through path share the same CFG successor (the `finally` block), causing the equality check to misidentify the jump as redundant.
- The fix adds `isFinallyBlockWithDistinctContinuation`: before reporting, check whether the common successor is a finally block that leads somewhere genuinely different for the return path vs. the fall-through path (i.e., it has a non-exit successor that is not a dead-loop condition block).
- Dead-loop conditions (`do-while(false)`, `while(false)`, and constant-false expressions like `Boolean.FALSE` or `!true`) are correctly excluded — returns inside them *are* redundant.

## Related

- RSPEC update: https://github.com/SonarSource/rspec/pull/7186

## Test plan

- [ ] Existing test cases continue to pass (no regression on `// Noncompliant` and compliant cases)
- [ ] New compliant cases: `while`/`for`/`do-while` loops with early return in try-finally — no issue raised
- [ ] New `// Noncompliant` cases: `return` at end of method in try-finally, `do-while(false)`, `while(false)`, `do-while(Boolean.FALSE)`, `do-while(!true)` — issue correctly raised


java-checks/src/main/java/org/sonar/java/checks/RedundantJumpCheck.java
```diff
@@ -77,12 +90,109 @@ private static boolean isSwitchCaseChild(Tree tree) {
...
+      current = current.parent();
+    }
+    return false;
+  }
+
+  private static boolean isLoop(Tree tree) {
+    return tree.is(Tree.Kind.WHILE_STATEMENT, Tree.Kind.DO_STATEMENT, Tree.Kind.FOR_STATEMENT, Tree.Kind.FOR_EACH_STATEMENT);
+  }
+
   private static Block nonEmptySuccessor(Block initialBlock) {
```
asya-vorobeva: Let's refactor it to improve readability. I suggest to move all Block-related methods to CFG class (or even better, as it's already >1300LOC, create separate CFGUtils class) and provide Javadoc for all public methods moved.

asya-vorobeva (changes requested):

java-checks/src/main/java/org/sonar/java/checks/RedundantJumpCheck.java
francois-mora-sonarsource: Refactor implemented

asya-vorobeva (approved):
Great job!
💯

PR #5702, opened 2026-06-24
SONARJAVA-6526 Implement rule S2654: JEE applications should not use threads or synchronization

NoemieBenard:
----
## Summary by Gitar

- **New Check:**
  - Implemented `S2654` to detect forbidden thread and synchronization usage in JEE `Servlet` and `EJB` components.
- **Rule Metadata:**
  - Added rule definition `S2654.json`, documentation `S2654.html`, and enabled it in `Sonar_way` and `Sonar_agentic_AI` profiles.
- **Testing & Infrastructure:**
  - Added comprehensive test samples covering both `javax` and `jakarta` namespaces, and updated profile test expectations.
  - Updated `its/autoscan` configuration to include `S2654` validation.


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2654.html
```diff
@@ -0,0 +1,22 @@
+<h2>Why is this an issue?</h2>
+<p>Proper synchronization and thread management can be tricky under the best of circumstances, but it’s particularly difficult in JEE application, and
+is even forbidden under some circumstances by the JEE standard.</p>
+<p>This rule raises an issue for each <code>Runnable</code>, and use of the <code>synchronized</code> keyword.</p>
+<h3>Noncompliant code example</h3>
```
romainbrenguier, line 5: There is no explicit "how to fix it" section and compliant example. I think we could improve a bit on the rspec side.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2654.html
```diff
@@ -0,0 +1,22 @@
+<h2>Why is this an issue?</h2>
+<p>Proper synchronization and thread management can be tricky under the best of circumstances, but it’s particularly difficult in JEE application, and
+is even forbidden under some circumstances by the JEE standard.</p>
+<p>This rule raises an issue for each <code>Runnable</code>, and use of the <code>synchronized</code> keyword.</p>
```
romainbrenguier, line 4: It's not clear that this applies only to jakarta applications.

romainbrenguier (approved):
Implementation looks good, but I think we can improve the Rspec

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaAgenticWayProfileTest.java
```diff
@@ -53,7 +53,7 @@ void profile_is_registered_as_expected() {
     BuiltInQualityProfilesDefinition.BuiltInQualityProfile actualProfile = profilesPerLanguages.get("java").get("Sonar agentic AI");
     assertThat(actualProfile.isDefault()).isFalse();
     assertThat(actualProfile.rules())
-      .hasSize(466)
+      .hasSize(467)
```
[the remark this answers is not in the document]
NoemieBenard: These rules appear after rebase from master, which was necessary so the assertion `.hasSize(...)` stays true on master when merging.

PR #5703, opened 2026-06-25
SONARJAVA-6519 Tune S2143 to reduce noise

asya-vorobeva:
- Stop raising issues for Joda-Time imports: Joda-Time is a recognized legacy library and its presence does not necessarily indicate a problem.
- Stop raising issues for java.sql.Timestamp: it is excluded because older JDBC drivers (pre-4.2) do not support java.time types, so java.sql.Timestamp may still be required for compatibility.

----
## Summary by Gitar

- **Rule logic enhancements:**
  - Added `java.sql.Date` and `java.sql.Time` to the list of checked types in `DateAndTimesCheck`.
- **Test infrastructure:**
  - Migrated legacy test samples to `java-checks-test-sources` and added `SqlTimestamp.java` to verify the new exception.


java-checks/src/main/java/org/sonar/java/checks/DateAndTimesCheck.java
```diff
@@ -42,15 +42,14 @@ public class DateAndTimesCheck extends AbstractMethodDetection implements JavaVe
     .build();
 
   private static final MethodMatchers DATE_CONSTRUCTOR = MethodMatchers.create()
-    .ofSubTypes("java.util.Date")
+    .ofTypes("java.util.Date", "java.sql.Date", "java.sql.Time")
     .constructor()
     .withAnyParameters()
     .build();
```
erwan-leforestier-sonarsource, line 48: Switching from `ofSubTypes("java.util.Date")` to three exact types silently stops reporting on project-specific `Date` subclasses. Was that intentional? If not, the constructor matcher should keep `ofSubTypes`.

asya-vorobeva: Yep, it's intentional sure. java.sql.Timestamp is a subtype of java.util.Date, and one of the goals of this PR was to stop raising issues for this case. As for project-specific Date inheritors, I'd say that it was mistakenly excluded, I don't think that we should raise issues for custom Date types.

erwan-leforestier-sonarsource: Ok thanks for the explanation :)

erwan-leforestier-sonarsource (approved):

PR #5704, opened 2026-06-25
SONARJAVA-6521 Exclude lifecycle annotations from S2325

romainbrenguier:
## Summary
Fixes false positives in rule S2325 for methods annotated with lifecycle annotations.

## Changes
Methods annotated with lifecycle annotations are now excluded from S2325, as they must remain instance methods to be invoked by frameworks on bean instances.

### Excluded annotations:
- `@PostConstruct` and `@PreDestroy` (javax.annotation / jakarta.annotation)
- `@PostActivate` and `@PrePassivate` (javax.ejb / jakarta.ejb)
- `@Startup` (io.quarkus.runtime)

## Rationale
Lifecycle methods are part of the framework contract and must be instance methods even if they don't access instance data. This is similar to the existing exclusions for serialization methods (`readObject`, `writeObject`, etc.).

## Test plan
- [x] Added comprehensive test cases in `StaticMethodCheckSample.java`
- [x] All existing tests pass
- [x] New tests verify all lifecycle annotations are properly excluded

### DIFF
--- a/java-checks-test-sources/default/src/main/files/non-compiling/checks/StaticMethodCheckSample.java
+++ b/java-checks-test-sources/default/src/main/files/non-compiling/checks/StaticMethodCheckSample.java
@@ -35,3 +35,36 @@ class UtilitiesExtension extends Utilities {
 class SerializableExclusions implements Serializable {
   private Object writeReplace() throws ObjectStreamException { }
 }
+
+class LifecycleAnnotations {
+  private static int counter = 0;
+
+  @javax.annotation.PostConstruct
+  private void initPostConstruct() { // Compliant - lifecycle method
+    System.out.println("PostConstruct");
+  }
+
+  @jakarta.annotation.PreDestroy
+  private void cleanupPreDestroy() { // Compliant - lifecycle method
+    System.out.println("PreDestroy");
+  }
+
+  @javax.ejb.PostActivate
+  private void activatePostActivate() { // Compliant - lifecycle method
+    System.out.println("PostActivate");
+  }
+
+  @jakarta.ejb.PrePassivate
+  private void passivatePrePassivate() { // Compliant - lifecycle method
+    System.out.println("PrePassivate");
+  }
+
+  @io.quarkus.runtime.Startup
+  private void startupMethod() { // Compliant - Quarkus startup method
+    System.out.println("Startup");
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/StaticMethodCheck.java
```diff
@@ -168,7 +180,21 @@ private static boolean shouldBePlacedAfterStatic(Modifier modifier) {
...
+
+  private static boolean hasLifecycleAnnotation(MethodTree tree) {
+    Symbol.MethodSymbol symbol = tree.symbol();
+    for (String annotation : LIFECYCLE_ANNOTATIONS) {
+      if (symbol.metadata().isAnnotatedWith(annotation)) {
+        return true;
+      }
+    }
+    return false;
   }
```
rombirli, line 195:
```suggestion
  private static boolean hasLifecycleAnnotation(MethodTree tree) {
    var metadata = tree.symbol().metadata();
    return Arrays.stream(LIFECYCLE_ANNOTATIONS).anyMatch(metadata::isAnnotatedWith);
  }
```

java-checks-test-sources/default/src/main/files/non-compiling/checks/StaticMethodCheckSample.java
rombirli, line 1: Can we put it in `java-checks-test-sources/default/src/main/java/checks/StaticMethodCheckSample.java` instead? It should be the default places to put samples unless it's impossible for some reasons

rombirli (approved):

java-checks-test-sources/default/src/main/files/non-compiling/checks/StaticMethodCheckSample.java
romainbrenguier: I think we won't be able to compile that file because it contains both javax and jakarta annotations and these are from incompatible dependencies.

java-checks/src/main/java/org/sonar/java/checks/StaticMethodCheck.java
romainbrenguier, line 20:
```suggestion

import java.util.ArrayList;
import java.util.Arrays;
```

PR #5706, opened 2026-06-26
SONARJAVA-6527 Implement rule S2206: Either fields or getters should be annotated for persistence but not both

NoemieBenard:
----
## Summary by Gitar

  - Implemented `S2206` to detect mixed persistence annotations (JPA/Jakarta) on fields and getters within the same class.
  - Excluded members annotated with `@Access` from the check as they explicitly define access mode.
- **Testing and configuration:**
  - Added `PersistenceAnnotationsMixedCheck` with unit tests and a sample source file covering various scenarios.
  - Included rule metadata, description in HTML, and enabled the rule in both `Sonar way` and `Sonar agentic AI` profiles.


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2206.json
```diff
@@ -0,0 +1,15 @@
+{
+  "title": "Either fields or getters should be annotated for persistence but not both",
+  "type": "BUG",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "20min"
+  },
+  "tags": [],
```
romainbrenguier: Maybe we could add a jakarta tag

romainbrenguier (approved):

PR #5707, opened 2026-06-26
SONARJAVA-6492: Implement S8912 - CredentialsProvider implementations should be annotated with @Unremovable

romainbrenguier:
## Summary

- Implements rule S8912: Custom `CredentialsProvider` implementations in Quarkus should be annotated with `@Unremovable` when they have a CDI scope annotation, to prevent them from being removed by Quarkus's bean pruning mechanism.
- Rule raises an issue on CDI-scoped classes implementing `io.quarkus.credentials.CredentialsProvider` that are missing the `@io.quarkus.arc.Unremovable` annotation.
- Test sample uses mock stub types for `io.quarkus.arc.Unremovable` and `io.quarkus.credentials.CredentialsProvider` (no external library dependency added to pom.xml), following the same pattern as S8911.

Rspec PR: https://github.com/SonarSource/rspec/pull/7153

## Test plan

- [ ] Rule implementation correctly detects missing `@Unremovable` on CDI-scoped `CredentialsProvider` implementations
- [ ] Compliant cases (with `@Unremovable`, or without CDI scope) are not flagged
- [ ] `withoutSemantic()` test verifies no issues raised without semantic analysis
- [ ] `JavaAgenticWayProfileTest` passes
- [ ] CI ruling and autoscan tests pass (to be updated after CI run)


java-checks/src/main/java/org/sonar/java/checks/CredentialsProviderUnremovableCheck.java
```diff
@@ -0,0 +1,76 @@
...
+public class CredentialsProviderUnremovableCheck extends IssuableSubscriptionVisitor {
+  private static final String CREDENTIALS_PROVIDER_FQN = "io.quarkus.credentials.CredentialsProvider";
+  private static final String UNREMOVABLE_FQN = "io.quarkus.arc.Unremovable";
+
+  private static final List<String> CDI_SCOPE_ANNOTATIONS = List.of(
+    "jakarta.enterprise.context.ApplicationScoped",
+    "jakarta.enterprise.context.RequestScoped",
+    "jakarta.enterprise.context.SessionScoped",
+    "jakarta.enterprise.context.Dependent",
+    "jakarta.inject.Singleton"
```
romainbrenguier: Maybe we should also add the javax equivalent here?

NoemieBenard: Since version 3.0, Quarkus is based on Jakarta EE 10 and I think it is mandatory  to use the jakarta.* packages, so I don't know if it is necessary to add the javax variant

java-checks/src/main/java/org/sonar/java/checks/CredentialsProviderUnremovableCheck.java
```diff
@@ -0,0 +1,76 @@
...
+    if (!hasCdiScopeAnnotation(metadata)) {
+      return;
+    }
+
+    if (metadata.isAnnotatedWith(UNREMOVABLE_FQN)) {
+      return;
+    }
+
+    reportIssue(
+      classTree.simpleName(),
```
NoemieBenard, line 72: Not sure it matters here, but this can be null

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8912.json
```diff
@@ -0,0 +1,26 @@
...
+    "cdi"
+  ],
+  "defaultSeverity": "Critical",
+  "ruleSpecification": "RSPEC-8912",
+  "sqKey": "S8912",
+  "scope": "All",
+  "quickfix": "unknown",
+  "code": {
+    "impacts": {
+      "SECURITY": "MEDIUM",
```
NoemieBenard, line 21: I thought we were not supposed to do security rules, did you ask code security about this rule? We could just keep the "reliability" impact, as it is probably the most important for this rule.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8912.json
```diff
@@ -0,0 +1,26 @@
...
+  "tags": [
+    "quarkus",
+    "credentials",
+    "cdi"
+  ],
+  "defaultSeverity": "Critical",
+  "ruleSpecification": "RSPEC-8912",
+  "sqKey": "S8912",
+  "scope": "All",
+  "quickfix": "unknown",
```
NoemieBenard: The rule is probably a good candidate for a quickfix, we can put it as "targeted" so that it will be easier to find it if we want to implement quickfixes later

java-checks-test-sources/default/src/main/java/checks/CredentialsProviderUnremovableCheckSampleMinimal.java
```diff
@@ -0,0 +1,16 @@
+package checks;
+
+import io.quarkus.arc.Unremovable;
```
NoemieBenard: This import is unused
```suggestion
```

NoemieBenard (commented):
Just a few comments

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8912.json
romainbrenguier: It's clearly a rule more about reliability than security, I'll remove the "SECURITY" entry from the rspec.

java-checks/src/main/java/org/sonar/java/checks/CredentialsProviderUnremovableCheck.java
romainbrenguier: You're correct that Quarkus 3.0+ is based on Jakarta EE 10 and requires jakarta.* packages. However, I included the javax variants following the pattern used in other rules in this codebase (e.g., JEEThreadCheck at java-checks/src/main/java/org/sonar/java/checks/JEEThreadCheck.java:48-52, which supports both javax and jakarta variants for EJB annotations).

The rationale is that SonarJava analyzes existing codebases, some of which may still be using Quarkus 2.x with javax packages. Supporting both variants ensures the rule works across different Quarkus versions, which is helpful for users in the process of migrating from Quarkus 2 to 3.

That said, if the team's policy is to only support current framework versions (Quarkus 3.x+), I can remove the javax variants. What's your preference?

romainbrenguier: Good point. While Quarkus 3.0+ requires Jakarta EE 10, following the codebase convention seen in other rules (e.g., `AtLeastOneConstructorCheck`, `ServletInstanceFieldCheck`, Spring checks), I've added the javax equivalents to support potential legacy code or older Quarkus versions:

```java
"javax.enterprise.context.ApplicationScoped",
"javax.enterprise.context.RequestScoped",
"javax.enterprise.context.SessionScoped",
"javax.enterprise.context.Dependent",
"javax.inject.Singleton"
```

romainbrenguier: Sorry about the verbose replies, that comes from the vibe-bot :sweat_smile:

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8912.json
romainbrenguier: I did the update on the rspec side. It will be taken into account by the next "update metadata" run

NoemieBenard (approved):
LGTM

PR #5710, opened 2026-06-29
SONARJAVA-6525 Decrease FPs for the rule S117

asya-vorobeva:
Local variables inside methods may now use leading or trailing underscores (e.g. `_good`, `good_`, `__myVar__`), as a signal that the variable is temporary, unimportant, or intentionally unnamed. 
The core name between underscores is still validated against the configured format. The leniency does not apply to method parameters or fields.


francois-mora-sonarsource (changes requested):
There is a gap between the documented exception and the implementation. The RSPEC PR says the underscore leniency applies only to local variables declared inside a method body, and that method parameters are not affected. The current implementation instead strips leading/trailing underscores from any visited variable that is not a method or constructor parameter, which includes named lambda parameters such as _name -> _name. This makes lambda parameters compliant even though they are not covered by the documented exception. I do not think we need to assume here whether lambda parameters should or should not be exempt; either direction can be valid, but the implementation and documentation should make the same choice explicitly.

francois-mora-sonarsource (approved):

PR #845, opened 2016-05-30
SONARJAVA-1689 Migrate to LTS

benzonico:

PR #5712, opened 2026-06-29
SONARJAVA-6518 Prevent JAVA analysis failure when missing java binaries

alban-auzeill:
Note `sonar-java classpath properties` is used below to reference `sonar.java.binaries`, `sonar.java.libraries`, `sonar.java.test.binaries`, `sonar.java.test.libraries`.

Technical decision part of this PR:
* All the logic about properties "sonar.binaries" and "sonar.libraries" has been removed from the Java analyzer code.
  * 2014.10.17 Deprecation of the properties with a warning: "sonar.binaries and sonar.libraries are deprecated since version 2.5 of sonar-java-plugin"
  * 2016.05.10 In the PR #845, if the properties are present, the analysis fails with an exception: sonar.binaries and sonar.libraries are not supported since version 4.0 of sonar-java-plugin
  * Design decision of today: the code has been cleaned from those ancestral properties logic, the analysis does not fail if the properties are present, the user still has warnings related to the missing 'sonar.java.binaries',  'sonar.java.libraries', 'sonar.java.test.binaries', 'sonar.java.test.libraries'

* An invalid element in one of the `sonar-java classpath properties`:
  * Will not throw any more IllegalStateException, e.g.: `No files nor directories matching 'lib/*.jar'`
  * But only a console log and SQ warning. e.g.: `Invalid value for 'sonar.java.libraries', no files nor directories matching 'lib/*.jar'.`

* Empty or missing sonar.java.libraries sonar.java.test.libraries:
  * Will not log any more this warning message, e.g.: `Dependencies/libraries were not provided for analysis of SOURCE files. The 'sonar.java.libraries' property is empty. Verify your configuration, as you might end up with less precise results.`
  * But a console log warning and SQ warning when the property is missing e.g.: `Missing 'sonar.java.libraries' property. You might end up with less precise analysis results.`

* Empty or missing sonar.java.binaries:
  * Will not throw any more IllegalStateException, e.g.: `Your project contains .java files, please provide compiled classes with sonar.java.binaries property, or exclude them from the analysis with sonar.exclusions property.`
  * But a console log warning and SQ warning when the property is missing e.g.: `Missing 'sonar.java.binaries' property. You might end up with less precise analysis results.`

* The logic in `logUndefinedTypes()` has not changed. Warning related to `sonar-java classpath properties` are only logged/sent if `problemsToFilePaths` is not empty. So if there were no unresolved imports/types during the analysis, then there are no warnings.

* There is no longer a need to have a dedicated logic for SonarLint; the class ClasspathForMainForSonarLint has been removed.

----
## Summary by Gitar

- **Refactored classpath initialization:**
  - Consolidated classpath logic into `AbstractClasspath`, introducing a unified `init()` method and `logClasspathWarnings()` to handle configuration validation.
  - Replaced hard-failing `IllegalStateException` with non-blocking analysis warnings when `binaries` or `libraries` properties are missing or invalid.
- **Improved user experience:**
  - Analysis no longer fails when required classpath properties are missing or invalid, allowing for partial analysis results instead.
  - Updated integration tests to assert successful analysis completion with associated warning logs.
- **Cleaned up codebase:**
  - Removed `ClasspathForMainForSonarLint` and streamlined `ClasspathForMain`/`ClasspathForTest` configurations.
  - Standardized property resolution and warning reporting across all classpath components.


java-frontend/src/main/java/org/sonar/java/classpath/AbstractClasspath.java
```diff
@@ -135,11 +180,25 @@ protected Set<File> getFilesFromProperty(String property) {
...
+    // No need to iterate over the entire collection, checking that there are two elements is enough
+    Iterator<InputFile> iterator = fs.inputFiles(sourcePredicate()).iterator();
+    if (iterator.hasNext()) {
+      iterator.next();
+      return iterator.hasNext();
+    }
+    return false;
+  }
+
+  protected boolean hasJavaFiles() {
```
antoine-vinot-sonarsource, line 192: This method is exactly the same as the `hasJavaSources` method line 182

java-frontend/src/main/java/org/sonar/java/classpath/AbstractClasspath.java
```diff
@@ -58,26 +61,67 @@ public abstract class AbstractClasspath {
...
+      libraries.addAll(extraLibraries);
+      logResolvedFiles(binariesProperty, binaries);
+      logResolvedFiles(librariesProperty, libraries);
+      Set<File> all = new LinkedHashSet<>(binaries);
+      all.addAll(libraries);
+      elements = List.copyOf(all);
+    }
+  }
+
+  protected void validatePropertiesPresence(String binariesProperty, String librariesProperty) {
```
antoine-vinot-sonarsource: The two parameters are already class attributes, we should be able to drop them, right?

antoine-vinot-sonarsource (approved):
Just a couple of minor comments, otherwise LGTM, I was able to test it locally and it works well.
Also I did not check the Gitar comments, not sure if they are relevant

java-frontend/src/main/java/org/sonar/java/classpath/AbstractClasspath.java
```diff
@@ -87,26 +131,29 @@ protected List<File> getJdkJars() {
...
   }
 
-  protected abstract void init();
-
-  public abstract void logSuspiciousEmptyLibraries();
+  public void logClasspathWarnings() {
+    for (String warning : classpathWarnings) {
+      LOG.warn(warning);
+      analysisWarnings.addUnique(warning);
+    }
```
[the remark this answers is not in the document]
alban-auzeill: It is intentional

java-frontend/src/main/java/org/sonar/java/classpath/AbstractClasspath.java
```diff
@@ -87,26 +131,29 @@ protected List<File> getJdkJars() {
...
+    classpathWarnings.clear();
+  }
```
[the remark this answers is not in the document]
alban-auzeill: It is intentional

java-frontend/src/main/java/org/sonar/java/classpath/AbstractClasspath.java
alban-auzeill: I removed the duplicate method.

java-frontend/src/main/java/org/sonar/java/classpath/AbstractClasspath.java
alban-auzeill: I removed the parameters.

PR #5724, opened 2026-07-01
SONARJAVA-6540 Implement S8954: Bean Validation constraints should not be placed on static fields

NoemieBenard:
----
## Summary by Gitar

- **New Rule Implementation:**
  - Added `BeanValidationConstraintOnStaticFieldCheck` to flag constraints on `static` fields.
  - Supports both `javax` and `jakarta` validation packages.
- **Documentation and Metadata:**
  - Created `S8954.json` and `S8954.html` with rule details and remediation guidance.
  - Enabled the rule in the `Sonar_way_profile.json` configuration.
- **Test Coverage:**
  - Implemented unit tests in `BeanValidationConstraintOnStaticFieldCheckTest` using provided sample files for `javax` and `jakarta` validation.
- **Autoscan Configuration:**
  - Updated `diff_S1710.json` and added `diff_S8954.json` in the `autoscan` test resources.


java-checks-test-sources/default/src/main/java/checks/BeanValidationConstraintOnStaticFieldCheckSampleJavax.java
```diff
@@ -0,0 +1,70 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+class BeanValidationConstraintOnStaticFieldCheckSampleJavax {
+
+  @NotNull
+  private static String staticField; // Noncompliant {{Remove the "static" modifier from this field.}}
+  //      ^^^^^^
```
romainbrenguier: I think the messaging is not very clear, because the user may not understand why it's flagged here. I think we should highlight the annotation instead and say something like "Move the constraint to an instance field."

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8954.json
```diff
@@ -0,0 +1,26 @@
...
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "5 min"
+  },
+  "tags": [
+    "validation",
+    "pitfall",
+    "bug"
+  ],
```
romainbrenguier, line 14: we could add "jakarta"

romainbrenguier (changes requested):
The messaging doesn't look great to me. Otherwise implementation looks good.

romainbrenguier (approved):

PR #5755, opened 2026-07-07
S6856: Support Spring 7 path-segment API versioning

patpatpat123:
## Problem

  Spring Framework 7 supports resolving an API version from a request path segment:

  ```java
  @Override
  public void configureApiVersioning(ApiVersionConfigurer configurer) {
    Predicate<RequestPath> versionedPaths = this::isVersionedPath;
    configurer.usePathSegment(1, versionedPaths);
  }

  @GetMapping(value = "/api/v{apiVersion}/resources/{id}", version = "2")
  Resource getResource(@PathVariable UUID id) {
    // ...
  }

  PathApiVersionResolver reads the raw path segment (v2 in this example) independently of controller argument binding. The {apiVersion} placeholder is structural: it makes the route match the version
  segment, but it does not need to be exposed as a controller method parameter.

  S6856 currently reports {apiVersion} as unbound. Adding an unused @PathVariable merely to satisfy the rule would make the controller signature misleading.

  ## Proposed solution

  Teach S6856 to recognize source-visible Spring MVC path-version configuration and defer its final decision until all source files in the module have been inspected.

  The exception is deliberately conservative. It applies only when:

  - Spring Web 7 or later is present;
  - usePathSegment(...) is called directly from an actual configureApiVersioning callback;
  - the segment index is a non-negative compile-time constant;
  - only one distinct path-version index is configured;
  - the controller mapping declares a nonblank version;
  - no fallback resolver, default version, optional-version configuration, dynamic index, or escaped configurer makes the configuration ambiguous; and
  - the placeholder occurs only in the configured segment and is not a catch-all variable.

  Other missing variables are still reported. For example, /api/v{apiVersion}/resources/{id} continues to report an unbound {id}.

  ## Scope

  This proposal handles direct, source-visible Spring MVC Java configuration. It intentionally does not infer WebFlux configuration, configuration supplied by dependencies or properties, delegated
  configurer mutations, or programmatic path prefixes.

  ## Tests

  This change adds a Spring Web 7.0.8 fixture module and regression coverage for:

  - both usePathSegment overloads;
  - method-reference predicates;
  - configuration declared after controller sources;
  - class-level paths;
  - ordinary missing variables;
  - versionless mappings and catch-all variables;
  - fallback resolvers;
  - inactive configuration methods;
  - multiple and dynamic indices; and
  - escaped configurers.

  Verified locally:

  - Spring Web 7 fixture module: successful
  - S6856 focused suite: 29 tests, 0 failures

  Spring API-versioning documentation:
  https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-config/api-version.html

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
           maven-args: >
             --define maven.test.skip=true
             --define sonar.skip=true
-            --projects !java-checks-test-sources/aws,!java-checks-test-sources/default,!java-checks-test-sources/java-17,!java-checks-test-sources/spring-3.2,!java-checks-test-sources/spring-web-4.0
+            --projects !java-checks-test-sources/aws,!java-checks-test-sources/default,!java-checks-test-sources/java-17,!java-checks-test-sources/spring-3.2,!java-checks-test-sources/spring-web-4.0,!java-checks-test-sources/spring-web-7.0
 
   ruling-qa:
     strategy:
--- a/its/autoscan/src/test/java/org/sonar/java/it/AutoScanTest.java
+++ b/its/autoscan/src/test/java/org/sonar/java/it/AutoScanTest.java
@@ -123,7 +123,7 @@ public class AutoScanTest {
       .setProjectName(PROJECT_NAME)
       .setProjectVersion("0.1.0-SNAPSHOT")
       .setSourceEncoding("UTF-8")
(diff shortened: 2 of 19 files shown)


patpatpat123:
Hello team, may i get some feedback on this?

patpatpat123:
Well, can someone take a look instead?

patpatpat123:
Can you guys actually look at this please?

patpatpat123:
Can you guys actually look at this please?

patpatpat123:
What is going on?

patpatpat123:
What is going on?

PR #5770, opened 2026-07-10
SONARJAVA-4926 Implement new rule S8989

romainbrenguier:
This rule detects methods annotated with @Transactional from Spring Framework that declare checked exceptions but do not explicitly configure rollback behavior using rollbackFor or noRollbackFor attributes. By default, Spring only rolls back transactions for unchecked exceptions, which can lead to silent data corruption when checked exceptions are thrown but the transaction commits anyway. The rule helps prevent this by requiring explicit rollback configuration for all transactional methods that throw checked exceptions.

[RSPEC PR](https://github.com/SonarSource/rspec/pull/7353)

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/spring/TransactionalMethodCheckedExceptionCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/spring/TransactionalMethodCheckedExceptionCheckSample.java
@@ -12,22 +12,27 @@ public class TransactionalMethodCheckedExceptionCheckSample {
 
   @Transactional
   public void processOrder(Order order) throws IOException, SQLException { // Noncompliant {{Specify rollback behavior for checked exceptions using "rollbackFor" or "noRollbackFor" attributes.}}
+//            ^^^^^^^^^^^^
   }
 
   @Transactional
   public void importData() throws Exception { // Noncompliant
+//            ^^^^^^^^^^
   }
 
   @Transactional(timeout = 30)
   public void withOtherAttributes() throws SQLException { // Noncompliant
+//            ^^^^^^^^^^^^^^^^^^^
   }
 
   @Transactional
   public void customException() throws CustomCheckedException { // Noncompliant
+//            ^^^^^^^^^^^^^^^
   }
 
   @Transactional
   public void mixedExceptions() throws IOException, RuntimeException { // Noncompliant
+//            ^^^^^^^^^^^^^^^
   }
 
   @Transactional(rollbackFor = IOException.class)
@@ -78,6 +83,7 @@ public class TransactionalMethodCheckedExceptionCheckSample {
   @Transactional
   static class ClassLevelNoConfig {
     public void noConfig() throws IOException { // Noncompliant
+//              ^^^^^^^^
     }
 
     @Transactional(rollbackFor = IOException.class)
@@ -91,6 +97,7 @@ public class TransactionalMethodCheckedExceptionCheckSample {
 
   @org.springframework.transaction.annotation.Transactional
   public void fullyQualified() throws IOException { // Noncompliant
+//            ^^^^^^^^^^^^^^
   }
 
   @Transactional(rollbackFor = IOException.class)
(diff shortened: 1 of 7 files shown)


java-checks-test-sources/default/src/main/java/checks/spring/TransactionalMethodCheckedExceptionCheckSample.java
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.transaction.annotation.Transactional;
+
+class Order {}
+class NotificationException extends Exception {}
+class CustomCheckedException extends Exception {}
+
+public class TransactionalMethodCheckedExceptionCheckSample {
+
+  @Transactional
+  public void processOrder(Order order) throws IOException, SQLException { // Noncompliant {{Specify rollback behavior for checked exceptions using "rollbackFor" or "noRollbackFor" attributes.}}
```
romainbrenguier: We should check precisely where the issue is raised

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8989.json
```diff
@@ -0,0 +1,26 @@
...
+  "tags": [
+    "spring",
+    "transaction",
+    "pitfall"
+  ],
+  "defaultSeverity": "Critical",
+  "ruleSpecification": "RSPEC-8989",
+  "sqKey": "S8989",
+  "scope": "Main",
+  "quickfix": "unknown",
```
NoemieBenard, line 18: quickfix shoud be "covered" (or "partial" since it doesn't suggest using "noRollbackFor") based on the implementation

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8989.html
```diff
@@ -0,0 +1,116 @@
...
+    notificationService.sendEmail(order);
+}
+</pre>
+<h2>Resources</h2>
+<h3>Documentation</h3>
+<ul>
+  <li>Spring Framework Documentation - Transaction Management - <a
+  href="https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/annotations.html">Official Spring documentation
+  explaining the @Transactional annotation and its rollback behavior</a></li>
+  <li>Baeldung - Spring @Transactional Rollback - <a href="https://www.baeldung.com/spring-transactional-rollback">Practical guide to configuring
```
NoemieBenard, line 113: I get a 404 on this link

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8989.html
```diff
@@ -0,0 +1,116 @@
...
+    // We want order saved even if notification fails
+    notificationService.sendEmail(order);
+}
+</pre>
+<h2>Resources</h2>
+<h3>Documentation</h3>
+<ul>
+  <li>Spring Framework Documentation - Transaction Management - <a
+  href="https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/annotations.html">Official Spring documentation
+  explaining the @Transactional annotation and its rollback behavior</a></li>
```
NoemieBenard, line 112: Following the guidelines, it probably should be: 
```suggestion
  <li>Spring Framework Documentation - <a
  href="https://docs.spring.io/spring-framework/reference/data-access/transaction/declarative/annotations.html">Using @Transactional </a></li>
```

oracle-10g-adds-more-to-forall.html
NoemieBenard, line 1: I am not sure I understand the purpose of this file

ruling-diff-comment-config.md
NoemieBenard, line 1: I think this should not be part of the PR (and it is probably the same for TESTING-RULING-DIFF-ACTION.md, or at least we need to move it to an agent directory if we want to keep it as a skill)

NoemieBenard (approved):
The implementation looks good, some questions/suggestions on the documentation and agent-generated files

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S8989.html
romainbrenguier: I'm replacing it in the rspec PR.

oracle-10g-adds-more-to-forall.html
romainbrenguier: Sorry it looks like it was comited by mistake by nigel

PR #5780, opened 2026-07-13
SONARJAVA-3259 Implement new rule S5673

romainbrenguier:
This rule detects classes and interfaces annotated with `@Component` whose names suggest they should use specialized Spring stereotype annotations (`@Service`, `@Repository`, `@Controller`, or `@RestController`) instead.

The rule also covers interfaces, which is relevant for Spring Data repository interfaces — where Spring generates the implementation at runtime, but `@Repository` is still the appropriate stereotype.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
@@ -31,7 +31,7 @@ public class SpringComponentSpecializationCheck extends IssuableSubscriptionVisi
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return List.of(Tree.Kind.CLASS);
+    return List.of(Tree.Kind.CLASS, Tree.Kind.INTERFACE);
   }
 
   @Override
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
@@ -50,7 +50,7 @@ public class SpringComponentSpecializationCheck extends IssuableSubscriptionVisi
     String suggestedAnnotation = getSuggestedAnnotation(className);
 
     if (suggestedAnnotation != null) {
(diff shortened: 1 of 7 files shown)


java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
```diff
@@ -0,0 +1,90 @@
...
+
+    if (componentAnnotation.isEmpty()) {
+      return;
+    }
+
+    String className = classTree.simpleName().name();
+    String suggestedAnnotation = getSuggestedAnnotation(className);
+
+    if (suggestedAnnotation != null) {
+      reportIssue(componentAnnotation.get(), String.format("Use @%s instead of @Component", suggestedAnnotation));
```
rombirli: Optional: The issue message could also suggest renaming the class as an alternative way to fix

java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
```diff
@@ -0,0 +1,90 @@
...
+import org.sonar.plugins.java.api.tree.AnnotationTree;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S5673")
+public class SpringComponentSpecializationCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.CLASS);
```
rombirli: What about records, enums, interfaces, annotation_types? Just to double-check if it is intentional to not include them.

rombirli (approved):
Very clear implementation

java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
romainbrenguier: I don't think this would make sense for records enum and annotations. But for interfaces I think it does, in particular repositories can be just interfaces. Let me check.

PR #5785, opened 2026-07-14
SONARJAVA-6522 Don't raise S2092 for deletion cookies

nils-werner-sonarsource:
## Problem

`S2092` (`SecureCookieCheck`) flags the common "delete this cookie" idiom — `new Cookie(name, null)`/`new Cookie(name, "")` combined with `setMaxAge(0)`, or the Spring equivalent `ResponseCookie.from(name, "").maxAge(0)...` — as if it were a real, value-carrying cookie missing its secure flag. A deletion cookie carries no sensitive value, so whether it's marked secure is moot, but the rule doesn't know that.

[SONARJAVA-6522](https://sonarsource.atlassian.net/browse/SONARJAVA-6522)

## Approach

Guiding principle (shared with the sibling ticket SONARJAVA-6523, #5781): favor fewer false positives, and fail safe on ambiguity — if either condition can't be resolved to a literal, don't suppress, keep raising as before.

The ticket spans two independent code paths in `SecureCookieCheck.java`, since they behave differently today:

| Question | Decision |
|---|---|
| `javax.servlet.http.Cookie`/`jakarta.servlet.http.Cookie`: how to detect the deletion pattern? | `new Cookie(name, value)` (2-arg constructor, no secure param) is *always* flagged today unless a later `setSecure(true)` cancels it — this is where the real FP lives. Added a second, independent `Map<Symbol.VariableSymbol, NewClassTree>` (`deletionCandidateCookies`), populated when the constructor's value argument is a literal `null`/`""`, and consumed by a new `checkMaxAgeCall` when a literal `setMaxAge(0)` is later seen on the same variable — mirroring the existing `unsecuredCookies`/`checkSecureCall` cancellation idiom already in this file. |
| Separate map or generalize `unsecuredCookies`? | Separate map — "value is null/empty" and "secure is false" are orthogonal facts about the same constructor call; keeps each map single-purpose. |
| `ResponseCookie.from(name, value).maxAge(0)...secure(false)`: how to detect the deletion pattern? | This builder is never flagged by itself today (no default-insecure assumption for builders) — only an explicit `.secure(false)` in the chain triggers it, reported immediately. Since it's a single self-contained expression tree, no new tracking map is needed: added a backward structural walk (`isDeletionCookieChain`) from the `secure(false)` call via `methodSelect().expression()`, looking for a `.maxAge(0)` and a root `.from(name, null/"")` call anywhere in the chain. Order-independent by construction (it's a tree walk, not sequential-visitation-based). |
| Precise `MethodMatchers` for `.maxAge(...)`, or name+literal-arg matching? | Precise `MethodMatchers` for `maxAge(long)`, consistent with this file's existing style for constructor-overload matching, and naturally excludes the `Duration` overload (out of scope per the ticket's literal-only requirement). |

## Implementation

`SecureCookieCheck.java`: new `deletionCandidateCookies` map + `addToDeletionCandidates`/`checkMaxAgeCall` for the servlet/JAX-RS `Cookie` path; new `RESPONSE_COOKIE_FROM`/`RESPONSE_COOKIE_MAX_AGE_ZERO` matchers + `isDeletionCookieChain` backward chain-walk for the Spring `ResponseCookie` builder path; one shared `isNullOrEmptyLiteral` helper used by both. No new abstractions beyond what's needed.

## Tests

`SecureCookieCheckSample.java` (default sources) — servlet `Cookie` deletion pattern:
- Compliant: empty/null value + `setMaxAge(0)`.
- Noncompliant: real value (Condition A fails), missing `setMaxAge` call (Condition B fails), non-zero `setMaxAge` (Condition B fails), non-literal `setMaxAge` argument (ambiguous, fails safe).

`SecureCookieCheckSample.java` (spring-3.2 module) — `ResponseCookie` builder chain:
- Compliant: empty value + `.maxAge(0)` + `.secure(false)` in the same chain.
- Noncompliant: missing `.maxAge(0)`, and real value with `.maxAge(0)` present (Condition A fails).

## Test plan

- [x] `mvn -pl java-checks test -Dtest="SecureCookieCheckTest#test,SecureCookieCheckTest#test_jakarta,SecureCookieCheckTest#test_non_compiling"` — all pass
- [x] Confirmed `SecureCookieCheckTest#test_with_spring_3_2` fails identically (same "expect some issues, but there's none" pre-existing environment failure) on unmodified `master` and on this branch — pre-existing, unrelated to this change (same finding as #5781)
- [x] Ran the full `org.sonar.java.checks.security.*Test` suite — identical set of 15 pre-existing unrelated failures (Android/WebView/ClearTextProtocol/CookieHttpOnly/DebugFeatureEnabled/EmptyDatabasePassword/ReceivingIntents + the known `SecureCookieCheckTest` spring-3.2 failure) on this branch; no new regressions introduced

[SONARJAVA-6522]: https://sonarsource.atlassian.net/browse/SONARJAVA-6522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
@@ -54,6 +54,7 @@ public class SecureCookieCheck extends IssuableSubscriptionVisitor {
   private static final String JAX_RS_NEW_COOKIE_JAKARTA = "jakarta.ws.rs.core.NewCookie";
   private static final String SPRING_SAVED_COOKIE = "org.springframework.security.web.savedrequest.SavedCookie";
   private static final String PLAY_COOKIE = "play.mvc.Http$Cookie";
+  private static final String SPRING_HTTP_COOKIE_BUILDER = "org.springframework.http.ResponseCookie$ResponseCookieBuilder";
   private static final List<String> COOKIES = Arrays.asList(
     "javax.servlet.http.Cookie",
     "jakarta.servlet.http.Cookie",
@@ -67,7 +68,7 @@ public class SecureCookieCheck extends IssuableSubscriptionVisitor {
     PLAY_COOKIE,
     "play.mvc.Http$CookieBuilder",
     "org.springframework.boot.web.server.Cookie",
-    "org.springframework.http.ResponseCookie$ResponseCookieBuilder");
+    SPRING_HTTP_COOKIE_BUILDER);
 
   private static final List<String> SETTER_NAMES = Arrays.asList("setSecure", "withSecure", "secure");
 
@@ -114,7 +115,20 @@ public class SecureCookieCheck extends IssuableSubscriptionVisitor {
     .addParametersMatcher(JAVA_LANG_STRING, JAVA_LANG_STRING, "java.lang.Integer", JAVA_LANG_STRING, JAVA_LANG_STRING, BOOLEAN, BOOLEAN, "play.mvc.Http$Cookie$SameSite")
     .build();
 
+  private static final MethodMatchers RESPONSE_COOKIE_FROM = MethodMatchers.create()
+    .ofTypes("org.springframework.http.ResponseCookie")
+    .names("from")
+    .addParametersMatcher(JAVA_LANG_STRING, JAVA_LANG_STRING)
+    .build();
+
+  private static final MethodMatchers RESPONSE_COOKIE_MAX_AGE_ZERO = MethodMatchers.create()
+    .ofTypes(SPRING_HTTP_COOKIE_BUILDER)
+    .names("maxAge")
+    .addParametersMatcher("long")
+    .build();
+
   private final Map<Symbol.VariableSymbol, NewClassTree> unsecuredCookies = new HashMap<>();
+  private final Map<Symbol.VariableSymbol, NewClassTree> deletionCandidateCookies = new HashMap<>();
   private final Set<NewClassTree> cookieConstructors = new HashSet<>();
   private final Deque<Symbol.TypeSymbol> enclosingClass = new ArrayDeque<>();
 
@@ -131,6 +145,7 @@ public class SecureCookieCheck extends IssuableSubscriptionVisitor {
   @Override
   public void setContext(JavaFileScannerContext context) {
     unsecuredCookies.clear();
+    deletionCandidateCookies.clear();
     cookieConstructors.clear();
     enclosingClass.clear();
     super.setContext(context);
@@ -145,10 +160,13 @@ public class SecureCookieCheck extends IssuableSubscriptionVisitor {
   public void visitNode(Tree tree) {
     if (tree.is(Tree.Kind.VARIABLE)) {
       addToUnsecuredCookies((VariableTree) tree);
+      addToDeletionCandidates((VariableTree) tree);
     } else if (tree.is(Tree.Kind.ASSIGNMENT)) {
       addToUnsecuredCookies((AssignmentExpressionTree) tree);
+      addToDeletionCandidates((AssignmentExpressionTree) tree);
     } else if (tree.is(Tree.Kind.METHOD_INVOCATION)) {
       checkSecureCall((MethodInvocationTree) tree);
+      checkMaxAgeCall((MethodInvocationTree) tree);
     } else if (tree.is(Tree.Kind.CLASS)) {
       enclosingClass.push(((ClassTree) tree).symbol());
     } else {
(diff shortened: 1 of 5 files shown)


nils-werner-sonarsource:
Fixed — good catch. \`isDeletionCookieChain\` only walked backward from the \`secure(false)\` call via its receiver chain, so \`maxAge(0)\` placed *after* \`secure(false)\` in the same fluent chain was structurally an ancestor, not reachable by that walk. Added a \`chainRoot\` helper that climbs to the outermost call of the fluent chain via \`Tree.parent()\` before walking, so the whole chain is inspected regardless of call order. Added a regression test (\`deleteCookieMaxAgeAfterSecure\`) covering \`.secure(false).maxAge(0)\` ordering.

java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -192,7 +210,7 @@ private void checkSecureCall(MethodInvocationTree mit) {
       ExpressionsHelper.ValueResolution<Boolean> valueResolution = ExpressionsHelper.getConstantValueAsBoolean(mit.arguments().get(0));
       Boolean secureArgument = valueResolution.value();
       boolean isFalse = secureArgument != null && !secureArgument;
-      if (isFalse) {
+      if (isFalse && !isDeletionCookieChain(mit)) {
```
Luqmansonar, line 227: This suppresses the wrong path for the servlet/JAX-RS Cookie case. `isDeletionCookieChain` (per its own javadoc) only recognizes the Spring `ResponseCookie` fluent-chain pattern — it walks receiver `MethodInvocationTree`s. For `cookie.setSecure(false)` where `cookie` is a plain identifier, the receiver is an `IdentifierTree`, so the loop in `isDeletionCookieChain` never executes and it always returns `false`. `!isDeletionCookieChain(mit)` is therefore always `true` here, and this line unconditionally reports whenever `secure(false)` is a literal — regardless of whether the cookie is a deletion cookie.

The `deletionCandidateCookies`/`checkMaxAgeCall` mechanism only removes the entry from `cookieConstructors`, which is consulted only in `leaveFile()` (for cookies that were never explicitly secured). It's never consulted here.

Repro — this is the ticket's own second compliant example:
```java
Cookie cookie = new Cookie("refreshToken", null);
cookie.setHttpOnly(true);
cookie.setSecure(false);
cookie.setPath("/");
cookie.setMaxAge(0);
response.addCookie(cookie);
```
Running `SecureCookieCheck` against this still raises S2092 on the `cookie.setSecure(false)` line. Verified with both call orders (`setMaxAge(0)` before and after `setSecure(false)`) — same result both times, so it's not an ordering issue.

None of the new sample cases added in this PR call `.setSecure(false)` explicitly (they only rely on the never-explicitly-secured constructor path), so this gap has no test coverage.

Suggested fix: this method needs to also check `deletionCandidateCookies` for the receiver's symbol (like `checkMaxAgeCall` does) before reporting, not just rely on the Spring-chain check.

nils-werner-sonarsource: Good catch — confirmed and fixed. `isDeletionCookieChain` was indeed Spring-chain-only; the servlet/JAX-RS `Cookie` explicit `setSecure(false)` case never consulted `deletionCandidateCookies` at all.

Fixed by deferring the report decision to `leaveFile()` for deletion-candidate variables: `setMaxAge(0)` sightings are now recorded in a plain, never-removed fact set, so the check no longer cares whether `setMaxAge(0)` comes before or after `setSecure(false)` in the method. Added test coverage for both statement orders (including your exact repro) plus the still-raises case with no `setMaxAge` call at all.

Luqmansonar (approved):
LGTM

java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -192,13 +221,19 @@ private void checkSecureCall(MethodInvocationTree mit) {
...
       Boolean secureArgument = valueResolution.value();
       boolean isFalse = secureArgument != null && !secureArgument;
-      if (isFalse) {
-        reportIssue(mit.arguments(), MESSAGE, valueResolution.valuePath(), null);
-      }
       ExpressionTree methodObject = ((MemberSelectExpressionTree) mit.methodSelect()).expression();
-      if (methodObject.is(Tree.Kind.IDENTIFIER)) {
-        IdentifierTree identifierTree = (IdentifierTree) methodObject;
-        NewClassTree newClassTree = unsecuredCookies.remove(identifierTree.symbol());
+      Symbol.VariableSymbol receiverSymbol = methodObject.is(Tree.Kind.IDENTIFIER) ? (Symbol.VariableSymbol) ((IdentifierTree) methodObject).symbol() : null;
```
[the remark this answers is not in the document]
nils-werner-sonarsource: Good catch — fixed. Extracted a shared `variableSymbolOf` helper that guards with `isVariableSymbol()` before casting, used by both `checkSecureCall` and `checkMaxAgeCall`, consistent with the sibling `addToUnsecuredCookies`/`addToDeletionCandidates` methods.

PR #5789, opened 2026-07-14
SONARJAVA-6509 Implement S9024: `@InjectMocks` should be preferred over manual initialization

NoemieBenard:
----
## Summary by Gitar

  - Added `MockitoInjectMocksShouldBeUsedCheck` (S9024) to flag manual instantiation of test objects when `@InjectMocks` could be used.
  - Implemented rule metadata and localized documentation for S9024.
- **Test coverage:**
  - Created `MockitoInjectMocksShouldBeUsedSample` covering various JUnit 4/5 scenarios and Mockito configurations.
  - Added `MockitoInjectMocksShouldBeUsedCheckTest` to verify the detection logic.

### DIFF
--- a/java-checks-test-sources/default/src/test/java/checks/tests/MockitoInjectMocksShouldBeUsedSample.java
+++ b/java-checks-test-sources/default/src/test/java/checks/tests/MockitoInjectMocksShouldBeUsedSample.java
@@ -1,0 +1,277 @@
+package checks.tests;
+
+import org.junit.Before;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+import org.mockito.Spy;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+public class MockitoInjectMocksShouldBeUsedSample {
+
+  interface PaymentGateway {
+    void process();
+  }
+
+  interface InventoryService {
+    int getStock(String item);
+  }
+
+  static class OrderProcessor {
+    OrderProcessor(PaymentGateway gateway, InventoryService inventory) {}
+    OrderProcessor(PaymentGateway gateway) {}
+    OrderProcessor() {}
+    OrderProcessor(PaymentGateway gateway, String name) {}
+  }
+
+  // ===== JUnit 4 - @RunWith(MockitoJUnitRunner.class) =====
+
+  @RunWith(MockitoJUnitRunner.class)
+  public class JUnit4NoncompliantBothMocks {
+    @Mock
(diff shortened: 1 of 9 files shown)


java-checks-test-sources/default/src/test/java/checks/tests/MockitoInjectMocksShouldBeUsedSample.java
rombirli, line 1: This sample is a bit narrow because every case revolves around OrderProcessor, which is an easy `@InjectMocks` target. It would be useful to add another constructed class with less straightforward constructors, for example one mixing mocks with non-mock/primitive parameters, to cover edge cases beyond the happy path.

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoInjectMocksShouldBeUsedCheck.java
```diff
@@ -0,0 +1,191 @@
...
+      }
+      if (expression instanceof MemberSelectExpressionTree memberSelect
+        && memberSelect.expression() instanceof IdentifierTree identifierTree
+        && "this".equals(identifierTree.name())) {
+        return memberSelect.identifier().symbol();
+      }
+      return null;
+    }
+
+    private boolean allArgsMockFields(NewClassTree newClass) {
```
rombirli, line 179: This is purely argument-based and does not inspect constructor injectability. A constructor taking two parameters of the same mock type still matches here, even though `@InjectMocks` resolution is ambiguous/unreliable in that case.

for example : 
```java
  static class SameTypeDependenciesOrderProcessor {
    SameTypeDependenciesOrderProcessor(PaymentGateway primaryGateway, PaymentGateway backupGateway) {}
  }
```

java-checks/src/main/java/org/sonar/java/checks/tests/MockitoInjectMocksShouldBeUsedCheck.java
```diff
@@ -0,0 +1,191 @@
...
+      this.classSymbol = classSymbol;
+    }
+
+    @Override
+    public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+      ExpressionTree expression = tree.expression();
+      if (expression instanceof NewClassTree newClass
+        && isFieldAssignment(tree.variable())
+        && allArgsMockFields(newClass)) {
+        issues.add(newClass);
```
rombirli, line 157: This only checks that we are assigning a new instance to a field; it never verifies that the constructed type is actually a valid `@InjectMocks` target. For example, automatic `@InjectMocks` instantiation fails on non-static inner classes because Mockito cannot implicitly resolve the required enclosing outer instance. We could filter these out.

rombirli (approved):
Looks good to me overall: the implementation is sound and the rule itself makes sense. One thing we could still improve in a follow-up is covering more edge cases in the implementation.

PR #5815, opened 2026-07-17
CLP-211 Update plugin API and Renovate configuration

vdiez:
Part of CLP-211

## Summary
- compile and test against sonar-plugin-api 13.8
- remove repository-specific dependency grouping rules
- remove the disabled sonar-plugin-api Renovate rule so the shared preset can manage updates
- preserve existing ignore paths and non-grouping compatibility/safety rules
- update internal test doubles for API methods introduced since 13.0

## Validation
- `mvn -pl sonar-java-plugin,java-surefire -am -DskipTests package`
- parsed `.github/renovate.json` with `jq`
- verified against master that only grouping rules and the sonar-plugin-api disable rule were removed


java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/internal/InternalSensorContext.java
```diff
@@ -207,4 +213,9 @@ public void addTelemetryProperty(String s, String s1) {
   public void addAnalysisData(String s, String s1, InputStream inputStream) {
     throw notSupportedException("addAnalysisData(String,String,InputStream)");
   }
+
+  @Override
+  public boolean isFeatureAvailable(String featureKey) {
+    return false;
```
aurelien-coet-sonarsource, line 219: Since this is a method in the new API that shouldn't be used, should this throw a `notSupportedException` like `newIssueResolution` rather than return a boolean ?

aurelien-coet-sonarsource (approved):
Just a nitpick, feel free to ignore.

java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/internal/InternalSensorContext.java
vdiez: I think returning `false` is preferable here. `isFeatureAvailable(...)` is a capability query, so `false` accurately reports that the light test context does not support the feature and lets callers take their normal fallback path.

This also matches the defaults in `scanner-engine-light` and `SensorContextTester`. `newIssueResolution()` is different: it is an operation that must return a usable builder, so throwing when unsupported is appropriate there.

PR #5816, opened 2026-07-17
SONARJAVA-6637 - Fix FP on S8924 when mocking parameterized types

NoemieBenard:

### DIFF
--- a/its/autoscan/src/test/resources/autoscan/diffs/diff_S8924.json
+++ b/its/autoscan/src/test/resources/autoscan/diffs/diff_S8924.json
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S8924",
   "hasTruePositives": false,
-  "falseNegatives": 36,
+  "falseNegatives": 37,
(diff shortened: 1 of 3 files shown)


its/autoscan/src/test/resources/autoscan/diffs/diff_S8924.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S8924",
   "hasTruePositives": false,
-  "falseNegatives": 36,
+  "falseNegatives": 37,
```
rombirli, line 4: Do you know why there is 1 more issue now on autoscan?

rombirli (approved):
LGTM

its/autoscan/src/test/resources/autoscan/diffs/diff_S8924.json
NoemieBenard: I guess it is because I added one noncompliant example in the samples

PR #5818, opened 2026-07-17
CLP-212: Migrate scanner test fixtures

vdiez:
## Summary

- replace the SonarQube implementation artifact with scanner-engine 13.4.1 test fixtures across affected modules
- migrate moved scanner implementation imports and use `TestSonarRuntime`
- remove unused implementation and Plugin API test-fixture dependencies from modules that need neither
- declare JUnit 4 directly where the removed artifact previously supplied it transitively

## Dependency changes

Before: analyzer tests relied on `org.sonarsource.sonarqube:sonar-plugin-api-impl`, including for scanner implementation types and transitive scanner helpers.

After: affected modules directly declare `plugin-api-scanner-impl` and `sensor-test-fixtures`; `java-frontend` also directly declares `sonar-scanner-extension-framework` for `PropertyDefinitions`. All scanner artifacts use `13.4.1.4007`. The four modules whose forbidden-API checks inspect `RunWith` now declare JUnit 4 directly. The resolved Maven graph contains no `sonar-plugin-api-impl`.

`sonar-plugin-api-test-fixtures` is retained only in modules that still use its logging helpers.

## Validation

- affected reactor test compilation on Java 26
- Maven dependency-tree check for `org.sonarsource.sonarqube:sonar-plugin-api-impl` (no matches)
- `git diff --check`

Running the complete local test suite additionally requires the repository's prebuilt `java-checks-test-sources/*/target/test-classpath.txt` inputs; affected test compilation is green without them.

### DIFF
--- a/external-reports/pom.xml
+++ b/external-reports/pom.xml
@@ -30,8 +30,13 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.sonarsource.sonarqube</groupId>
-      <artifactId>sonar-plugin-api-impl</artifactId>
+      <groupId>org.sonarsource.scanner.engine</groupId>
+      <artifactId>plugin-api-scanner-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.sonarsource.scanner.engine</groupId>
+      <artifactId>sensor-test-fixtures</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
@@ -53,6 +58,11 @@
       </exclusions>
     </dependency>
     <!-- unit tests -->
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter</artifactId>
--- a/external-reports/pom.xml
+++ b/external-reports/pom.xml
@@ -58,6 +58,7 @@
       </exclusions>
     </dependency>
     <!-- unit tests -->
+    <!-- Required only for the forbid-junit4 testCheck to resolve forbidden classes from forbid_junit4.txt. -->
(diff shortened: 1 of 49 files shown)


external-reports/pom.xml
```diff
@@ -53,6 +58,11 @@
       </exclusions>
     </dependency>
     <!-- unit tests -->
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
```
aurelien-coet-sonarsource, line 66: Why is this now an explicit dependency ?

aurelien-coet-sonarsource (approved):
LGTM, just a small question.

external-reports/pom.xml
vdiez: JUnit 4 was already on this module's test classpath transitively through `sonar-plugin-api-impl`. Removing that artifact made the `forbiddenapis:testCheck` execution (`forbid-junit4`) fail because its signature file references `org.junit.runner.RunWith` and that class could no longer be resolved. This declaration makes that existing build-time requirement explicit; the module's tests remain on JUnit 5.

vdiez: More precisely, no JUnit test was failing. The failure was the [`Test Analyze` CI job](https://github.com/SonarSource/sonar-java/actions/runs/29583196761/job/87894041096), in the `external-reports` Maven execution:

`Failed to execute goal de.thetaphi:forbiddenapis:3.10:testCheck (forbid-junit4) on project external-reports: Parsing signatures failed: Class 'org.junit.runner.RunWith' not found on classpath while parsing signature: org.junit.runner.RunWith`

The referenced signature is [`forbid_junit4.txt:9`](https://github.com/SonarSource/sonar-java/blob/master/forbid_junit4.txt#L9). The explicit dependency is only there so this check can resolve the class it forbids.

vdiez: added a comment in the pom directly so we don't lose track of why this is needed

PR #5832, opened 2026-07-22
SONARJAVA-6654 Fix S2166 false positives for classes with unresolvable supertypes

asya-vorobeva:
Guard against ECJ batch mode behavior where a compilation error causes superClass() to return null instead of UnknownType, which was triggering false positives on classes with an extends clause.


erwan-leforestier-sonarsource:
I don't see any unit test to make sure it's fine. Is it possible to add one ?

asya-vorobeva:
> I don't see any unit test to make sure it's fine. Is it possible to add one ?

Unfortunately no. This situation happens in _very specific case_ when the project fails to be parsed with ECJ parser (which we use in default batch mode). The problem is that our unit tests always use JParser, so this path is inherently untestable at the unit level. And our ruling projects also don't satisfy this condition. I've already added defensive comment to clarify this.

erwan-leforestier-sonarsource (approved):

PR #5834, opened 2026-07-23
SONARJAVA-6666 Fix FP on S8924 when a non-Mockito method with the same name is in scope

NoemieBenard:
## Summary

- Rule S8924 was raising a false positive when `Mockito.<method>(...)` was called with an explicit qualifier but a method with the same name was already in scope (via a static import from another class, or a locally defined method), making the `Mockito.` prefix necessary for disambiguation.
- Fix: before reporting, the check now collects non-Mockito static imports and enclosing class method names, and suppresses the issue if the method name conflicts with either.

## Test plan

- [ ] Existing `MockitoStaticImportCheckSample` tests still pass (no regression on noncompliant cases)
- [ ] New `MockitoStaticImportCheckNameConflictSample` test covers the two FP scenarios: static import conflict and local method conflict

### DIFF
--- a/its/autoscan/src/test/resources/autoscan/diffs/diff_S1128.json
+++ b/its/autoscan/src/test/resources/autoscan/diffs/diff_S1128.json
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S1128",
   "hasTruePositives": true,
-  "falseNegatives": 30,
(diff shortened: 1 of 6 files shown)


nathsou (approved):
Image

PR #5840, opened 2026-07-27
SONARJAVA-6655 Extend S8694 to use EndOfAnalysis with 80% threshold

asya-vorobeva:
Issues are now only raised when at least 80% of targeted DateTime.of() calls across the module use int literals instead of enum constants. Compliant overloads (Month/DayOfWeek parameters) count toward the total but never raise issues.


java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -132,39 +175,40 @@ public List<Tree.Kind> nodesToVisit() {
...
         return;
       }
     }
     ExpressionTree firstArgument = mit.arguments().get(0);
     int firstArgumentLiteral = getIntLiteral(firstArgument);
-    if (DAY_OF_WEEK_OF_MATCHER.matches(mit)
-      && isValidDay(firstArgumentLiteral)) {
-      reportAndCreateQuickfix(mit, getDayOfWeekEnumName(firstArgumentLiteral), DAY_ISSUE_MESSAGE, JAVA_TIME_DAY_OF_WEEK);
+    if (DAY_OF_WEEK_OF_MATCHER.matches(mit) && isValidDay(firstArgumentLiteral)) {
+      collectIssue(mit, getDayOfWeekEnumName(firstArgumentLiteral), DAY_ISSUE_MESSAGE, JAVA_TIME_DAY_OF_WEEK);
```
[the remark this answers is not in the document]
asya-vorobeva: As number of non-literal or out-of-range arguments shouldn't be large, we can deal with them.

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -210,25 +240,50 @@ public void visitNode(Tree tree) {
...
+        MONTH_ISSUE_MESSAGE, JAVA_TIME_MONTH);
+    } else if (DAY_OF_WEEK_GET_VALUE_MATCHER.matches(methodInvocationSide) && isValidDay(intLiteral)) {
+      collectIssue(binaryExpressionTree, getValueReplacement(methodInvocationSide, binaryExpressionTree, getDayOfWeekEnumName(intLiteral), isReversed),
+        DAY_ISSUE_MESSAGE, JAVA_TIME_DAY_OF_WEEK);
+    }
+  }
+
+  @Override
+  public void endOfAnalysis(ModuleScannerContext context) {
+    if (projectTotalCount > 0 && projectRaisedCount * 100 >= RAISED_PERCENTAGE_THRESHOLD * projectTotalCount) {
```
erwan-leforestier-sonarsource: `projectRaisedCount` is confusing because it suggests that this is number of issues that were already raised but nothing has been raised yet, I suggest to rename it to something else to make it clear that this is about number of issues that will be raised for this rule only.  
Same for projectTotalCount, it's not clear that it's counting the number of `DateTime.of` calls found in the project.

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -210,25 +240,50 @@ public void visitNode(Tree tree) {
...
+      collectIssue(binaryExpressionTree, getValueReplacement(methodInvocationSide, binaryExpressionTree, getMonthEnumName(intLiteral), isReversed),
+        MONTH_ISSUE_MESSAGE, JAVA_TIME_MONTH);
+    } else if (DAY_OF_WEEK_GET_VALUE_MATCHER.matches(methodInvocationSide) && isValidDay(intLiteral)) {
+      collectIssue(binaryExpressionTree, getValueReplacement(methodInvocationSide, binaryExpressionTree, getDayOfWeekEnumName(intLiteral), isReversed),
+        DAY_ISSUE_MESSAGE, JAVA_TIME_DAY_OF_WEEK);
+    }
+  }
+
+  @Override
+  public void endOfAnalysis(ModuleScannerContext context) {
```
erwan-leforestier-sonarsource, line 322: [Found by Claude]
When you open a pull request, SonarQube does not re-read every file in your project. That would be slow. It only reads the files you changed. For the files you did not touch, it pulls the old result from a cache instead of reading them again. This is called caching, and it is on by default for pull request and branch analysis.

## The bug

This rule counts by reading files. But it never saved its counts to the cache, and it never reads them back. So for every unchanged file, the rule contributes zero to both counts. The only files that get counted are the ones in your pull request.

The division at the end is done over just those few files, not the whole project. The 80% "across the project" promise is quietly broken.

## Clear example

Say your project has 100 date usages. 90 of them use numbers. That is 90%, well above 80%, so the rule should warn.

Now you open a pull request that touches one file. That file has 2 date usages, and both happen to use the clean `Month.OCTOBER` style.

- Full analysis (no caching): counts 90 out of 100 = 90% -> warns. Correct.
- Pull request analysis (caching on): the other 99 files are not read, so they count as zero. Only your one file counts: 0 numbers out of 2 = 0% -> stays silent.

Same project, opposite answer, just because caching skipped the other files. And it flips the other way too: a project that is mostly clean can suddenly start warning if the one file you edited happens to be full of numbers.

## The fix the ticket asked for

Each time the rule reads a file, it should save that file's two counts into the cache. Next time, for a file that did not change, it reads those saved counts back instead of skipping it. That way the final division always covers the whole project, whether or not caching skipped the reading. There is already another rule in the same codebase (S5693) that does exactly this, so it is a copy-the-existing-pattern job, not new invention.

asya-vorobeva: This is the approach I've tried as the first one. The problem with it is that it produces stale results in many cases. See 'Bug: Cached files' issues not re-evaluated when threshold flips' comment above. To get proper results we can cache all the issues produced, but it creates additional memory usage. So to avoid having cache is some trade-off.

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
asya-vorobeva: Renamed

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
asya-vorobeva: But I realize that I need to test it more strictly, maybe add some project in ruling tests. Let me think about it

asya-vorobeva: Done, and approach with cache is applied.

java-frontend/src/main/java/org/sonar/java/reporting/InternalJavaIssueBuilder.java
```diff
@@ -103,7 +103,12 @@ public final InternalJavaIssueBuilder onRange(Tree from, Tree to) {
     return onRange(AnalyzerMessage.textSpanBetween(from, to));
   }
 
-  private InternalJavaIssueBuilder onRange(AnalyzerMessage.TextSpan range) {
+  @Override
+  public InternalJavaIssueBuilder onRange(int startLine, int startCharacter, int endLine, int endCharacter) {
```
erwan-leforestier-sonarsource, line 107: Was it necessary to make this method public ? I don't see this method being called anywhere else but I might be wrong

erwan-leforestier-sonarsource (approved):

java-frontend/src/main/java/org/sonar/java/reporting/InternalJavaIssueBuilder.java
asya-vorobeva: This one is used in new version of DateEnumCheck. Thanks for review!

PR #5842, opened 2026-07-28
SONARJAVA-6685 Implement new rule S1944

romainbrenguier:
Detect inappropriate type casts that will always fail at runtime with a ClassCastException. The rule flags casts between unrelated concrete classes and casts between a final class and an unrelated interface.

Part of


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/Sonar_agentic_AI_profile.json
romainbrenguier, line 1: It looks like this was done using the wrong version of rule-api

romainbrenguier:
It looks like to raise anything useful, we would need some dataflow analysis, so I'll not implement this rule right now.

PR #5845, opened 2026-07-28
SONARJAVA-6686 Implement new rule S2198: Unnecessary mathematical comparisons

romainbrenguier:
Detect comparisons between a bounded numeric primitive (byte, short, char, int) and a compile-time constant outside the type's value range, which always evaluate to true or false.

Part of

### DIFF
--- a/its/ruling/src/test/resources/diff_S2198.json
+++ b/its/ruling/src/test/resources/diff_S2198.json
@@ -1,0 +1,6 @@
+{
+  "ruleKey": "S2198",
+  "hasTruePositives": true,
+  "falseNegatives": 0,
+  "falsePositives": 0
+}
--- a/its/ruling/src/test/resources/diff_S2198.json
+++ b/its/ruling/src/test/resources/diff_S2198.json
@@ -1,6 +1,0 @@
-{
-  "ruleKey": "S2198",
-  "hasTruePositives": true,
-  "falseNegatives": 0,
-  "falsePositives": 0
-}
--- a/java-checks-test-sources/default/src/main/java/checks/UselessMathematicalComparisonCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/UselessMathematicalComparisonCheckSample.java
@@ -6,6 +6,7 @@ class UselessMathematicalComparisonCheckSample {
 
   void byteComparisons(byte b) {
     if (b > 200) {} // Noncompliant {{Remove this comparison; it will always return false.}}
+//      ^^^^^^^
     if (b > 127) {} // Noncompliant {{Remove this comparison; it will always return false.}}
     if (b >= 128) {} // Noncompliant {{Remove this comparison; it will always return false.}}
     if (b < -200) {} // Noncompliant {{Remove this comparison; it will always return false.}}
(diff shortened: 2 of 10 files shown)


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2198.json
romainbrenguier, line 1: It looks like it's was not addded to the profiles

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2198.html
```diff
@@ -0,0 +1,70 @@
...
+  <li><strong>Wrong type</strong>: The variable should have been declared with a larger type (e.g., 32-bit instead of 8-bit)</li>
+  <li><strong>Wrong constant</strong>: The constant value is incorrect for the intended comparison</li>
+  <li><strong>Dead code</strong>: The condition guards code that can never execute (if always <code>false</code>) or always executes (if always
+  <code>true</code>)</li>
+</ul>
+<p>Such comparisons make the code confusing and harder to maintain. They may also hide genuine bugs where the developer’s intent does not match the
+actual behavior.</p>
+<p>In Java, these types are named: <code>byte</code> (8-bit signed), <code>short</code> (16-bit signed), <code>char</code> (16-bit unsigned),
+<code>int</code> (32-bit signed), <code>long</code> (64-bit signed), <code>float</code> (single-precision), and <code>double</code>
+(double-precision).</p>
```
NoemieBenard, line 29: This mentions long, float and double but the rule actually excludes these cases

java-checks-test-sources/default/src/main/java/checks/UselessMathematicalComparisonCheckSample.java
```diff
@@ -0,0 +1,116 @@
...
+    if (b > (200)) {} // Noncompliant {{Remove this comparison; it will always return false.}}
+  }
+
+  void constantExpressions(byte b) {
+    if (b > CONSTANT_200) {} // Noncompliant {{Remove this comparison; it will always return false.}}
+  }
+
+  void twoVariables(byte a, byte b) {
+    if (a > b) {} // Compliant - comparing two variables
+  }
```
NoemieBenard, line 175: we could add a sample for when the comparison is made against an expression (like b > 127 + 1)

java-checks-test-sources/default/src/main/java/checks/UselessMathematicalComparisonCheckSample.java
```diff
@@ -0,0 +1,116 @@
+package checks;
+
+class UselessMathematicalComparisonCheckSample {
+
+  static final int CONSTANT_200 = 200;
+
+  void byteComparisons(byte b) {
+    if (b > 200) {} // Noncompliant {{Remove this comparison; it will always return false.}}
```
NoemieBenard, line 8: We should add a precise location test

its/ruling/src/test/resources/diff_S2198.json
```diff
@@ -0,0 +1,6 @@
+{
+  "ruleKey": "S2198",
+  "hasTruePositives": true,
+  "falseNegatives": 0,
+  "falsePositives": 0
+}
```
NoemieBenard: I think this was added by mistake (autoscan diff already present in the correct folder)

NoemieBenard (approved):
Implementation looks good, just one file probably added by mistake and some comments on the samples/rspec

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2198.html
romainbrenguier: I'll make changes on the rspec side

PR #5849, opened 2026-07-30
SONARJAVA-6671 Add parametrized exclusion list and some exceptions for java:S1192

asya-vorobeva:
Enhancements:

- Suppress when duplicate strings are used in log’s messages
- Suppress when used in test environment
- Add excludePatterns rule property to ignore strings satisfying such patterns
- Add minimalLength rule property

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
@@ -23,6 +23,7 @@ import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 import org.sonar.check.Rule;
+import org.sonar.java.checks.helpers.LoggingMatchers;
 import org.sonar.plugins.java.api.JavaFileScanner;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
 import org.sonar.plugins.java.api.semantic.MethodMatchers;
@@ -41,13 +42,10 @@ import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.ReturnStatementTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
-import static org.sonar.plugins.java.api.semantic.MethodMatchers.ANY;
-
 @Rule(key = "S2629")
 public class LazyArgEvaluationCheck extends BaseTreeVisitor implements JavaFileScanner {
 
   private static final String STRING = "java.lang.String";
-  private static final String OBJECT_ARR = "java.lang.Object[]";
 
   private static class SLF4J {
 
@@ -62,19 +60,6 @@ public class LazyArgEvaluationCheck extends BaseTreeVisitor implements JavaFileS
     private static final String LOGGER = "org.slf4j.Logger";
     private static final String MARKER = "org.slf4j.Marker";
 
-    private static final MethodMatchers LOG = MethodMatchers.create()
-      .ofSubTypes(LOGGER)
-      .names(METHOD_NAMES)
-      .addParametersMatcher(STRING)
-      .addParametersMatcher(STRING, ANY)
-      .addParametersMatcher(STRING, ANY, ANY)
-      .addParametersMatcher(STRING, OBJECT_ARR)
-      .addParametersMatcher(MARKER, STRING)
-      .addParametersMatcher(MARKER, STRING, ANY)
-      .addParametersMatcher(MARKER, STRING, ANY, ANY)
-      .addParametersMatcher(MARKER, STRING, OBJECT_ARR)
-      .build();
-
     private static final MethodMatchers TEST = MethodMatchers.create()
       .ofSubTypes(LOGGER)
       .names(testMethodNames(METHOD_NAMES))
@@ -85,29 +70,7 @@ public class LazyArgEvaluationCheck extends BaseTreeVisitor implements JavaFileS
 
   private static class JUL {
 
-    private static final String[] METHOD_NAMES = {
-      "severe",
-      "warning",
-      "info",
-      "config",
-      "fine",
-      "finer",
-      "finest"
-    };
-
     private static final String LOGGER = "java.util.logging.Logger";
-
-    private static final MethodMatchers LOG = MethodMatchers.or(
-      MethodMatchers.create()
-        .ofTypes(LOGGER)
-        .names(METHOD_NAMES)
-        .addParametersMatcher(STRING)
-        .build(),
-      MethodMatchers.create()
-        .ofTypes(LOGGER)
-        .names("log")
-        .addParametersMatcher("java.util.logging.Level", STRING)
-        .build());
 
     private static final MethodMatchers TEST = MethodMatchers.create()
       .ofTypes(LOGGER)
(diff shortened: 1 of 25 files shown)


java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
```diff
@@ -85,30 +70,8 @@ private static class SLF4J {
...
-      MethodMatchers.create()
-        .ofTypes(LOGGER)
-        .names(METHOD_NAMES)
-        .addParametersMatcher(STRING)
-        .build(),
-      MethodMatchers.create()
-        .ofTypes(LOGGER)
-        .names("log")
-        .addParametersMatcher("java.util.logging.Level", STRING)
-        .build());
```
victor-peixoto-sonarsource, line 110: I agree with the change to apply `LoggingMatchers` here.
However, the original rule implementation only matched this exact overload, not all overloads. This is correct, but ideally: 1 - this should be covered in a separate PR; 2 - we need dedicated test cases to cover this behavioral change, not just the IT fixes.

java-frontend/src/main/java/org/sonar/java/utils/JavaFileTypeClassifier.java
```diff
@@ -0,0 +1,180 @@
...
+ * Enriches test scope determination beyond the platform's {@link InputFile.Type}.
+ *
+ * <p>The platform scanner (Maven, Gradle, CLI) assigns {@link InputFile.Type#TEST} or
+ * {@link InputFile.Type#MAIN} based on build-tool conventions, but this classification
+ * can be incomplete (e.g. test helpers in {@code src/main/java}, misconfigured projects).
+ *
+ * <p>This classifier combines three signals:
+ * <ol>
+ *   <li>Platform truth: {@link InputFile#type()} from the Sonar scanner</li>
+ *   <li>Naming conventions: file name patterns like {@code FooTest}, {@code ITFoo}, {@code FooSpec}</li>
```
victor-peixoto-sonarsource: The pattern match now is suffix only, please fix this documentation.

java-frontend/src/main/java/org/sonar/java/utils/JavaFileTypeClassifier.java
```diff
@@ -0,0 +1,180 @@
...
+   */
+  private static final List<String> TEST_PATH_SUBPATHS = List.of("src/it/java", "src/its/java");
+
+  /**
+   * Matches file names (without {@code .java} extension) that follow standard test naming conventions
+   * by suffix: {@code Test}, {@code Tests}, {@code TestCase}, {@code IT}, {@code ITCase}, {@code Spec}, {@code Specs}
+   * (e.g. {@code FooTest}, {@code FooSpec}, {@code FooIT}).
+   */
+  private static final Pattern TEST_NAME_PATTERN = Pattern.compile(
+    "^[A-Z]\\w*(Test|Tests|TestCase|IT|ITCase|Spec|Specs)$"
```
victor-peixoto-sonarsource, line 95: Fix documentation above.

victor-peixoto-sonarsource, line 95: Spec/Specs may not always be used for tests. So there is a risk of FPs. Maybe worth checking on GitHub to see how projects are using *Spec.java files.
Example that I just found: https://github.com/DataDog/datadog-api-client-java/blob/bf13740e1bfcb034072ffcf58c3193cbe9ad480a/src/main/java/com/datadog/api/client/v2/model/EntityV3APISpec.java

java-checks/src/main/java/org/sonar/java/checks/StringLiteralDuplicatedCheck.java
```diff
@@ -43,44 +48,69 @@
...
-        List<LiteralTree> duplications = literalTrees.stream().filter(literal -> literal.parent() != constant).toList();
+        List<LiteralTree> duplications = literalTrees.stream()
+          .filter(literal -> literal.parent() != constant && !isLoggingArgument(literal))
+          .toList();
+        if (duplications.isEmpty()) {
+          return;
+        }
         context.reportIssue(this, duplications.iterator().next(),
           "Use already-defined constant '" + constant.simpleName() + "' instead of duplicating its value here.",
           secondaryLocations(duplications.subList(1, duplications.size())), literalOccurrence);
```
victor-peixoto-sonarsource, line 99: `literalOccurrence` is set before the new logging filter is introduced. I believe it should change here, like below in the `else if ...` path.

java-frontend/src/main/java/org/sonar/java/utils/JavaFileTypeClassifier.java
```diff
@@ -0,0 +1,179 @@
...
+import org.sonar.plugins.java.api.semantic.SymbolMetadata;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+/**
+ * Enriches test scope determination beyond the platform's {@link InputFile.Type}.
+ *
+ * <p>The platform scanner (Maven, Gradle, CLI) assigns {@link InputFile.Type#TEST} or
+ * {@link InputFile.Type#MAIN} based on build-tool conventions, but this classification
+ * can be incomplete (e.g. test helpers in {@code src/main/java}, misconfigured projects).
```
victor-peixoto-sonarsource, line 34: Just a comment. I understand enriching this feature, but maybe the scanner would be the right place to change. I am ok with doing it here though.

victor-peixoto-sonarsource (changes requested):
Looks good overall. I added some comments. Can you please take a look and address them if they make sense?

java-frontend/src/main/java/org/sonar/java/utils/JavaFileTypeClassifier.java
victor-peixoto-sonarsource: Resolving this because it is just an observation. I don't have enough knowledge to infer this.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
asya-vorobeva: I understand it but I definitely don't want to fix it here. I suggest to create a separate PR for this. The change of behavior for this rule is super-minor.

java-frontend/src/main/java/org/sonar/java/utils/JavaFileTypeClassifier.java
asya-vorobeva: Recheck, it's already there

java-frontend/src/main/java/org/sonar/java/utils/JavaFileTypeClassifier.java
asya-vorobeva: The same here

asya-vorobeva: This example that you're related is autogenerated stuff, where this rule is ok not be ignored

java-frontend/src/main/java/org/sonar/java/utils/JavaFileTypeClassifier.java
victor-peixoto-sonarsource: Yes, there is correct. I think that the ITFoo is still outdated. Not a bloocker, but nit.

```suggestion
 *   <li>Naming conventions: file name patterns like {@code FooTest}, {@code FooSpec}</li>
```

java-frontend/src/main/java/org/sonar/java/utils/JavaFileTypeClassifier.java
asya-vorobeva: Generally speaking: Spec is typically used from Cucumber / Gerkin frameworks. Java borrowed this pattern. But maybe for some cases it's not the case. I'll remove this case

victor-peixoto-sonarsource (approved):
Approved with suggestions

java-frontend/src/main/java/org/sonar/java/utils/JavaFileTypeClassifier.java
victor-peixoto-sonarsource: I am ok with both. Just raising the concern.

java-checks/src/main/java/org/sonar/java/checks/LazyArgEvaluationCheck.java
victor-peixoto-sonarsource: Makes sense.

PR #5853, opened 2026-07-31
SONARJAVA-6455 S7467: FP on broken semantics

asya-vorobeva:
Without semantics, catch parameter usages() is always empty, causing false positives. Guard with context.getSemanticModel() != null.

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/ReplaceUnusedExceptionParameterWithUnnamedPatternCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/ReplaceUnusedExceptionParameterWithUnnamedPatternCheckSample.java
@@ -1,7 +1,11 @@
 package checks;
 
 import io.restassured.exception.PathException;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
 import java.util.List;
+import java.util.concurrent.*;
 import java.util.function.Supplier;
 
 public class ReplaceUnusedExceptionParameterWithUnnamedPatternCheckSample {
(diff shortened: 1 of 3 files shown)


francois-mora-sonarsource (approved):
Approving — the guard is correct for the case it targets (semantic model entirely absent), and the flipped `verifyNoIssues()` in `test_without_semantics` covers it.

One non-blocking note on test coverage. The added `executor_run_compliant()` sample doesn't exercise the change: it's analyzed with the full default classpath, so `e` resolves and has a usage, and it passes on the parent commit too. It also repeats the lambda-usage case already at line 117.

While checking that, I found a *separate* FP on partial semantics that this PR doesn't cover — `usages()` is also empty when the parameter binds fine but its usage sits in a lambda body ECJ couldn't bind. Reproduced against this branch:

```java
void usedInsideLambdaWithUnresolvedFunctionalInterface() {
  try {
    Integer.parseInt("x");
  } catch (NumberFormatException e) { // FP: "Replace e with an unnamed pattern"
    Unknown.run(() -> System.out.println(e.getMessage()));
  }
}
```

`NumberFormatException` resolves, the semantic model is present, so the new guard passes — but the usage inside the lambda was never registered. Note this is *not* fixable by checking `v.symbol().isUnknown()` instead (I tried; the symbol is known here). It needs something like detecting unresolved constructs in the parameter's scope, so it's a design question rather than a condition tweak.

Worth a follow-up ticket rather than blocking this one. Happy to file it with the reproducer.

asya-vorobeva:
> Approving — the guard is correct for the case it targets (semantic model entirely absent), and the flipped `verifyNoIssues()` in `test_without_semantics` covers it.
> 
> One non-blocking note on test coverage. The added `executor_run_compliant()` sample doesn't exercise the change: it's analyzed with the full default classpath, so `e` resolves and has a usage, and it passes on the parent commit too. It also repeats the lambda-usage case already at line 117.
> 

Absolutely agree, and I've added this test case only to illustrate customer's issue.

> While checking that, I found a _separate_ FP on partial semantics that this PR doesn't cover — `usages()` is also empty when the parameter binds fine but its usage sits in a lambda body ECJ couldn't bind. Reproduced against this branch:
> 
> ```java
> void usedInsideLambdaWithUnresolvedFunctionalInterface() {
>   try {
>     Integer.parseInt("x");
>   } catch (NumberFormatException e) { // FP: "Replace e with an unnamed pattern"
>     Unknown.run(() -> System.out.println(e.getMessage()));
>   }
> }
> ```
> 
> `NumberFormatException` resolves, the semantic model is present, so the new guard passes — but the usage inside the lambda was never registered. Note this is _not_ fixable by checking `v.symbol().isUnknown()` instead (I tried; the symbol is known here). It needs something like detecting unresolved constructs in the parameter's scope, so it's a design question rather than a condition tweak.
> 
> Worth a follow-up ticket rather than blocking this one. Happy to file it with the reproducer.

Yep, it's well-known problem for many rules. Good point, I'll create JIRA ticket to investigate and fix this point.

PR #5855, opened 2026-08-03
SONARJAVA-6695 Implement new rule S9130

romainbrenguier:
Detect when the return value of InputStream.read() or Reader.read() is cast to byte or char before being checked for -1, which can cause false end-of-stream detection or missed end-of-stream conditions.

Part of


rombirli (approved):
Goood rule

PR #5859, opened 2026-08-03
SONARJAVA-6704 Implement new rule S9132

romainbrenguier:
Detect calls to setMaximumPoolSize() on ScheduledThreadPoolExecutor instances. This method has no effect because ScheduledThreadPoolExecutor uses a fixed-size pool determined solely by corePoolSize.

Part of

### DIFF
--- a/sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9132.html
+++ b/sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9132.html
@@ -1,0 +1,27 @@
+<h2>Why is this an issue?</h2>
+<p>Scheduled thread pool executors are designed with a fixed-size thread pool architecture. Unlike standard thread pool implementations, they use an
+unbounded work queue and maintain a fixed number of threads equal to the core pool size configuration.</p>
+<p>When you call <code>setMaximumPoolSize()</code> on a <code>ScheduledThreadPoolExecutor</code>, the call completes without error, but it has no effect
+on the executor's behavior. The pool will never create threads beyond the core pool size, regardless of how high you set the maximum pool size.</p>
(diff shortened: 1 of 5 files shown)


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9132.html
```diff
@@ -0,0 +1,27 @@
...
+<p>Remove the call to <code>setMaximumPoolSize()</code> and use <code>setCorePoolSize()</code> instead to control the number of threads in the pool.</p>
+<h3>Noncompliant code example</h3>
+<pre>
+ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(5);
+executor.setMaximumPoolSize(10); // Noncompliant
+</pre>
+<h3>Compliant solution</h3>
+<pre>
+ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(5);
+executor.setCorePoolSize(10); // Pool now maintains exactly 10 threads
```
NoemieBenard, line 20: I am not sure but from what I understand of the rule the better fix is to directly create a pool with a bigger core size:
```suggestion
ScheduledThreadPoolExecutor executor = new ScheduledThreadPoolExecutor(10); // Pool now maintains exactly 10 threads
```

NoemieBenard (approved):
LGTM, just a comment

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9132.html
romainbrenguier: Yes, but the goal is to illustrate the use of `setCorePoolSize`. I will modify the example on the Rspec side to make more sense.

PR #5862, opened 2026-08-03
SONARJAVA-6703 Implement new rule S9133

romainbrenguier:
Detect hard-coded floating-point literals that approximate well-known mathematical constants (pi, e, sqrt(2), ln(2)) and suggest using the corresponding Math class constants or expressions instead.

Part of

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/HardcodedMathConstantCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/HardcodedMathConstantCheck.java
@@ -112,16 +112,14 @@ public class HardcodedMathConstantCheck extends IssuableSubscriptionVisitor {
     int count = 0;
     for (int i = 0; i < normalized.length(); i++) {
       char c = normalized.charAt(i);
-      if (c == '.' || c == '-' || c == '+') {
+      if (c == '.') {
         continue;
       }
-      if (c >= '0' && c <= '9') {
-        if (c != '0') {
-          foundNonZero = true;
-        }
-        if (foundNonZero) {
-          count++;
-        }
+      if (c != '0') {
+        foundNonZero = true;
+      }
+      if (foundNonZero) {
+        count++;
       }
     }
     return count;
--- a/java-checks/src/main/java/org/sonar/java/checks/HardcodedMathConstantCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/HardcodedMathConstantCheck.java
@@ -25,7 +25,7 @@ import org.sonar.plugins.java.api.tree.Tree;
 @Rule(key = "S9133")
 public class HardcodedMathConstantCheck extends IssuableSubscriptionVisitor {
 
-  private static final int MIN_SIGNIFICANT_DIGITS = 4;
+  private static final int MIN_SIGNIFICANT_DIGITS = 3;
 
   private enum MathConstant {
     PI(Math.PI, "Math.PI", "pi"),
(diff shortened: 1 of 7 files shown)


java-checks/src/main/java/org/sonar/java/checks/HardcodedMathConstantCheck.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.List;
+import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S9133")
+public class HardcodedMathConstantCheck extends IssuableSubscriptionVisitor {
+
+  private static final int MIN_SIGNIFICANT_DIGITS = 4;
```
rombirli: Most commonly used approximation of pi is 3.14, which has only 3 significant digits, i would set it to 3

java-checks/src/main/java/org/sonar/java/checks/HardcodedMathConstantCheck.java
```diff
@@ -0,0 +1,127 @@
...
+
+  private static String normalize(String rawValue) {
+    String value = rawValue.replace("_", "");
+    // Strip type suffix
+    char last = value.charAt(value.length() - 1);
+    if (last == 'f' || last == 'F' || last == 'd' || last == 'D') {
+      value = value.substring(0, value.length() - 1);
+    }
+    // Skip hex float literals
+    if (value.startsWith("0x") || value.startsWith("0X")) {
```
rombirli, line 100: We should extend to binary values (`0b`) and octal values (leading `0`)

java-checks/src/main/java/org/sonar/java/checks/HardcodedMathConstantCheck.java
```diff
@@ -0,0 +1,127 @@
...
+      if (c != '0') {
+        foundNonZero = true;
+      }
+      if (foundNonZero) {
+        count++;
+      }
+    }
+    return count;
+  }
+}
```
rombirli, line 114: Looks to be incorrect on negative numbers at first sight, the reproducer should contain a negative nubmer case to verify

rombirli: it could also be rewritten in a easier-to-understand way : 

```java
private static int countSignificantDigits(String normalized) {
    String sig = normalized.replace("-", "").replace(".", "").replaceFirst("^0+", "");
    return sig.isEmpty() ? 1 : sig.length();
}
```

rombirli (changes requested):
The rule looks good, but a few things should be improved : 
- support octal/binary bases
- what happens with negative numbers?
- some functions could be rewritten in an easier way, less verbose

java-checks/src/main/java/org/sonar/java/checks/HardcodedMathConstantCheck.java
```diff
@@ -0,0 +1,127 @@
...
+        return;
+      }
+    }
+  }
+
+  private static String normalize(String rawValue) {
+    String value = rawValue.replace("_", "");
+    // Strip type suffix
+    char last = value.charAt(value.length() - 1);
+    if (last == 'f' || last == 'F' || last == 'd' || last == 'D') {
```
rombirli: what about 
```java
if ("fd".contains(last.toLowerCase())) { ... 
```

java-checks/src/main/java/org/sonar/java/checks/HardcodedMathConstantCheck.java
romainbrenguier: Java does not support native binary or octal syntax for floating-point literals.

rombirli: you are right. I figured out `0x1.1` is not a valid hex double value. `0x11e-1` is but as this rule doesnt support exponent notation i think we could drop support for hex doubles/floats

rombirli (approved):
thanks for the changes and extending the reproducer

PR #5865, opened 2026-08-03
SONARJAVA-6710: Fix FN on empty condition loop

joke1196:
Part of 

----
## Summary by Gitar

- **Bug fix:**
    - Fixed false negative in `LoopExecutingAtMostOnceCheck` for empty condition `for(;;)` loops by requiring conditional control flow in the body
- **Tests:**
    - Added test cases covering various conditional structures and scope boundaries like lambdas and anonymous classes


rombirli:
Is this one ready for rereview?

PR #5867, opened 2026-08-03
SONARJAVA-6706 Implement new rule S2330

romainbrenguier:
Detect array covariance where an array of a derived type is assigned to a variable declared as an array of its base type, which can lead to ArrayStoreException at runtime.

### DIFF
--- a/its/ruling/src/test/resources/commons-beanutils/java-S2330.json
+++ b/its/ruling/src/test/resources/commons-beanutils/java-S2330.json
@@ -1,0 +1,13 @@
+{
+"commons-beanutils:commons-beanutils:src/main/java/org/apache/commons/beanutils2/LazyDynaList.java": [
+418,
+494
+],
+"commons-beanutils:commons-beanutils:src/main/java/org/apache/commons/beanutils2/MethodUtils.java": [
+1352,
+1352
+],
+"commons-beanutils:commons-beanutils:src/test/java/org/apache/commons/beanutils2/LazyDynaListTestCase.java": [
+209
+]
+}
--- a/its/ruling/src/test/resources/commons-beanutils/java-S2330.json
+++ b/its/ruling/src/test/resources/commons-beanutils/java-S2330.json
@@ -3,10 +3,6 @@
 418,
 494
 ],
-"commons-beanutils:commons-beanutils:src/main/java/org/apache/commons/beanutils2/MethodUtils.java": [
-1352,
-1352
-],
 "commons-beanutils:commons-beanutils:src/test/java/org/apache/commons/beanutils2/LazyDynaListTestCase.java": [
 209
 ]
--- a/sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2330.json
+++ b/sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2330.json
@@ -1,0 +1,23 @@
+{
+  "title": "Array covariance should not be used",
+  "type": "CODE_SMELL",
+  "code": {
+    "impacts": {
+      "RELIABILITY": "MEDIUM"
+    },
+    "attribute": "LOGICAL"
+  },
+  "status": "ready",
+  "remediation": {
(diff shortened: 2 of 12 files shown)


sonar-java-plugin/src/main/resources/profiles/Sonar_way/S2330
NoemieBenard, line 1: The fact that the rule is in sonar way seems contradictory with the RSPEC and the choice of disabling it by default in C# , is it intentional?

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2330.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "10min"
+  },
+  "tags": [
+    "pitfall"
+  ],
+  "defaultSeverity": "Critical",
```
NoemieBenard, line 18: Default severity should be "Major" to match "RELIABILITY": "MEDIUM".

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S2330.json
```diff
@@ -0,0 +1,23 @@
+{
+  "title": "Array covariance should not be used",
+  "type": "CODE_SMELL",
+  "code": {
+    "impacts": {
+      "RELIABILITY": "MEDIUM"
```
NoemieBenard, line 6: Is it intentional to modify the metadata from the C# RSPEC? To me it makes sense to keep the "MAINTAINABILITY" impact together with the RELIABILITY impact (maybe with lower maintainability than reliability)

its/ruling/src/test/resources/commons-beanutils/java-S2330.json
```diff
@@ -0,0 +1,13 @@
+{
+"commons-beanutils:commons-beanutils:src/main/java/org/apache/commons/beanutils2/LazyDynaList.java": [
+418,
+494
+],
+"commons-beanutils:commons-beanutils:src/main/java/org/apache/commons/beanutils2/MethodUtils.java": [
+1352,
```
NoemieBenard: Here the rule raises false positives on both parameters of `Array.equals` because the method takes parameters of type `Object[]` and these are arrays with a different sub-type. We should probably exclude standard library methods from the check (though there will probably be a high FP rate for the rule anyway). Based on the ruling diffs there is the same issue for `Array.fill`

NoemieBenard (changes requested):
Implementation looks good and the examples are very thorough, but I think the rule will be quite controversial. I already see some false positives worth addressing in the ruling diffs. Disabling the rule by default like in C# is also an option.

romainbrenguier:
It looks like the C# version is not performing well. According to the FPS report there are lots of cases where we report an issue, but it is not really relevant, which explains the poor performance.


- **Total Issues Analyzed**: 362
- **Total False Positives**: 136
- **False Positive Rate**: 37.6%
- **Rule Scope**: All

## Types of false positives

1. [Same-Type Array Operations](#cluster-1) (59 issues)
2. [Implicit Type Inference Matching Context](#cluster-2) (33 issues)
3. [Read-Only Array Usage](#cluster-3) (15 issues)
4. [Non-Array Operations Incorrectly Flagged](#cluster-4) (10 issues)
5. [Method Operations Without Assignment](#cluster-5) (8 issues)
6. [Explicit Type Casts](#cluster-6) (4 issues)
7. [Explicit Type Matching](#cluster-7) (3 issues)
8. [Polymorphic Element Assignment](#cluster-8) (2 issues)
9. [Safe Array Copy Operations](#cluster-9) (2 issues)



I think for the Java version of the rule there will also be a high potential for FPs. So I wouldn't spend time on this rule right now, but maybe wait until we have the FPS tool integrated in the PR process.

PR #5869, opened 2026-08-03
SONARJAVA-6430 Centralize class-like tree kinds

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **API additions:**
    - Added centralized `Tree.CLASS_KINDS` list containing all class-like node kinds in `Tree.java`
    - Added unit test in `TreeTest.java` to verify `CLASS_KINDS` matches all kinds backed by `ClassTree`

### DIFF
--- a/its/autoscan/src/test/resources/autoscan/diffs/diff_S1172.json
+++ b/its/autoscan/src/test/resources/autoscan/diffs/diff_S1172.json
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S1172",
   "hasTruePositives": true,
-  "falseNegatives": 32,
+  "falseNegatives": 33,
   "falsePositives": 0
 }
--- a/java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
@@ -43,7 +43,7 @@ public class CallOuterPrivateMethodCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Tree.Kind.CLASS_KINDS;
+    return List.of(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.ENUM, Tree.Kind.RECORD);
   }
 
   @Override
--- a/java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
@@ -16,7 +16,6 @@
  */
 package org.sonar.java.checks;
 
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -44,7 +43,7 @@ public class CallOuterPrivateMethodCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.INTERFACE);
+    return Tree.CLASS_KINDS;
   }
 
   @Override
--- a/java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
@@ -43,7 +43,7 @@ public class CallOuterPrivateMethodCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Tree.CLASS_KINDS;
+    return Tree.Kind.CLASS_KINDS;
   }
 
   @Override
--- a/java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
@@ -33,7 +33,7 @@ public class CallSuperMethodFromInnerClassCheck extends IssuableSubscriptionVisi
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Tree.CLASS_KINDS;
+    return Tree.Kind.CLASS_KINDS;
   }
 
   @Override
--- a/java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
@@ -16,7 +16,6 @@
  */
 package org.sonar.java.checks;
 
-import java.util.Arrays;
 import java.util.List;
 import org.sonar.check.Rule;
 import org.sonar.java.model.ExpressionUtils;
@@ -34,7 +33,7 @@ public class CallSuperMethodFromInnerClassCheck extends IssuableSubscriptionVisi
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.INTERFACE);
+    return Tree.CLASS_KINDS;
   }
 
   @Override
--- a/java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
@@ -33,7 +33,7 @@ public class CallSuperMethodFromInnerClassCheck extends IssuableSubscriptionVisi
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Tree.Kind.CLASS_KINDS;
+    return List.of(Tree.Kind.CLASS, Tree.Kind.INTERFACE);
   }
 
   @Override
--- a/java-checks/src/main/java/org/sonar/java/checks/ClassFieldCountCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ClassFieldCountCheck.java
@@ -41,7 +41,7 @@ public class ClassFieldCountCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Tree.CLASS_KINDS;
+    return Tree.Kind.CLASS_KINDS;
   }
 
   @Override
--- a/java-checks/src/main/java/org/sonar/java/checks/ClassFieldCountCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ClassFieldCountCheck.java
@@ -41,7 +41,7 @@ public class ClassFieldCountCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Tree.Kind.CLASS_KINDS;
+    return List.of(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.ENUM);
   }
 
   @Override
--- a/java-checks/src/main/java/org/sonar/java/checks/ClassFieldCountCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ClassFieldCountCheck.java
@@ -16,6 +16,7 @@
  */
 package org.sonar.java.checks;
 
+import java.util.List;
 import org.sonar.check.Rule;
 import org.sonar.check.RuleProperty;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
@@ -23,9 +24,6 @@ import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.tree.ClassTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.VariableTree;
-
-import java.util.Arrays;
-import java.util.List;
 
 import static org.sonar.java.checks.helpers.ExpressionsHelper.reportOnClassTree;
 
@@ -43,7 +41,7 @@ public class ClassFieldCountCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.ENUM);
+    return Tree.CLASS_KINDS;
   }
 
   @Override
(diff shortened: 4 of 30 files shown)


java-frontend/src/main/java/org/sonar/java/ast/visitors/ComplexityVisitor.java
```diff
@@ -57,14 +57,14 @@ public void visitMethod(MethodTree tree) {
 
   @Override
   public void visitClass(ClassTree tree) {
-    if(root.is(Tree.Kind.CLASS, Tree.Kind.ENUM, Tree.Kind.INTERFACE, Tree.Kind.ANNOTATION_TYPE, Tree.Kind.COMPILATION_UNIT)) {
+    if (Tree.CLASS_KINDS.contains(root.kind()) || root.is(Tree.Kind.COMPILATION_UNIT)) {
```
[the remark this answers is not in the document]
aurelien-coet-sonarsource: This is an improvement and the intended behaviour.

its/autoscan/src/test/resources/autoscan/diffs/diff_S1172.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruleKey": "S1172",
   "hasTruePositives": true,
-  "falseNegatives": 32,
+  "falseNegatives": 33,
```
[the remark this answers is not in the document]
aurelien-coet-sonarsource: This is expected and is not a regression: it is coming from the test samples.

java-checks/src/main/java/org/sonar/java/checks/RedundantNullabilityAnnotationsCheck.java
```diff
@@ -79,7 +78,7 @@ private void checkMembers(SymbolMetadata.NullabilityData classNullabilityData,
       } else if (member.is(Tree.Kind.METHOD)) {
         // check method
         checkMethod(classNullabilityData, (MethodTree) member, scope);
-      } else if (member.is(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.RECORD)) {
+      } else if (member.is(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.RECORD, Tree.Kind.ENUM, Tree.Kind.ANNOTATION_TYPE)) {
```
mostafa-mohammed-sonarsource: An `IMPLICIT_CLASS` cannot occur in a class member list, so the additional kind is unreachable in this context. Should we use `Tree.Kind.CLASS_KINDS.contains(member.kind())` here as well, to avoid keeping a second class-kind list that can drift? WDYT?

java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
```diff
@@ -44,7 +43,7 @@ public class CallOuterPrivateMethodCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.INTERFACE);
+    return Tree.Kind.CLASS_KINDS;
```
mostafa-mohammed-sonarsource: `ANNOTATION_TYPE` is not valid input for this rule.

```java
class Outer {
  private static int[] data() {
    return new int[] {1};
  }

  @interface Nested {
    int[] VALUES = data();
  }
}
```

S3398 reports:

```text
Move this method into "Nested".
```

That transformation is not legal Java. Moving `data` into `Nested` would require the annotation type to declare a static method, which annotation declarations cannot do.

Should we keep `ANNOTATION_TYPE` out of this rule's subscription, or handle it without proposing an illegal move? WDYT?

java-checks/src/main/java/org/sonar/java/checks/ClassFieldCountCheck.java
```diff
@@ -43,7 +41,7 @@ public class ClassFieldCountCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.ENUM);
+    return Tree.Kind.CLASS_KINDS;
```
mostafa-mohammed-sonarsource: `CLASS_KINDS` is too broad here without kind-specific counting and reporting.

For an implicit class:

```java
int field1;
int field2;
int field3;
int field4;
int field5;
int field6;
int field7;
int field8;
int field9;
int field10;
int field11;
int field12;
int field13;
int field14;
int field15;
int field16;
int field17;
int field18;
int field19;
int field20;
int field21;

void main() {}
```

With Java 25, S1820 counts the 21 fields and enters the reporting path. Analysis then fails with an `AnalysisException` caused by:

```text
ClassCastException: CompilationUnitTreeImpl cannot be cast to NewClassTree
```

`reportOnClassTree` assumes that a class without a name is an anonymous `NewClassTree`. That assumption does not hold for an implicit class.

For a record:

```java
record ManyFields(
  int field1,
  int field2,
  int field3,
  int field4,
  int field5,
  int field6,
  int field7,
  int field8,
  int field9,
  int field10,
  int field11,
  int field12,
  int field13,
  int field14,
  int field15,
  int field16,
  int field17,
  int field18,
  int field19,
  int field20,
  int field21
) {}
```

S1820 reports nothing. Record components are exposed through `recordComponents()`, not `members()`, even though each component declares a corresponding private final field.

Should we handle the report location for `IMPLICIT_CLASS` and include record components in the count before subscribing to all `CLASS_KINDS`? WDYT?

java-checks/src/main/java/org/sonar/java/checks/InterfaceOrSuperclassShadowingCheck.java
```diff
@@ -16,24 +16,22 @@
...
-import java.util.List;
-import java.util.Locale;
 
 @Rule(key = "S2176")
 public class InterfaceOrSuperclassShadowingCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.INTERFACE, Tree.Kind.RECORD);
+    return Tree.Kind.CLASS_KINDS;
```
mostafa-mohammed-sonarsource: The newly accepted kinds violate two assumptions in this check.

First, an implicit class has no `simpleName()`:

```java
// Object.java
void main() {}
```

With Java 25, this declares an implicit `Object` whose superclass is `java.lang.Object`. The names match, so the rule tries to report on `simpleName()` and analysis fails with an `AnalysisException` caused by `NullPointerException`.

Second, `Tree.Kind.name()` is not suitable for a diagnostic:

```java
@interface Annotation {}
```

This reports:

```text
Rename this annotation_type.
```

`annotation_type` is an internal enum name, not Java terminology. Should we add explicit reporting for `IMPLICIT_CLASS` and a user-facing name for `ANNOTATION_TYPE` before using `CLASS_KINDS` here? WDYT?

java-checks/src/main/java/org/sonar/java/checks/PublicStaticMutableMembersCheck.java
```diff
@@ -115,7 +116,7 @@ public class PublicStaticMutableMembersCheck extends IssuableSubscriptionVisitor
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.INTERFACE, Tree.Kind.CLASS, Tree.Kind.ENUM, Tree.Kind.ASSIGNMENT);
+    return ListUtils.concat(Tree.Kind.CLASS_KINDS, List.of(Tree.Kind.ASSIGNMENT));
```
mostafa-mohammed-sonarsource: `ANNOTATION_TYPE` has the same field-modifier constraints as an interface.

```java
@interface MutableAnnotation {
  int[] DATA = {1};
}
```

S2386 reports:

```text
Make this member "protected".
```

Annotation fields are implicitly `public static final`; `protected` is not permitted. Should we treat `ANNOTATION_TYPE` like `INTERFACE` and emit the existing move-to-a-class remediation? WDYT?

java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
```diff
@@ -34,7 +33,7 @@ public class CallSuperMethodFromInnerClassCheck extends IssuableSubscriptionVisi
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.INTERFACE);
+    return Tree.Kind.CLASS_KINDS;
```
mostafa-mohammed-sonarsource: A member enum is not an inner class for the purpose of this rule. It is implicitly static.

```java
class Outer {
  void name() {}

  enum Nested {
    VALUE;

    void useName() {
      name();
    }
  }
}
```

The call binds to the inherited `Enum.name()`. `Outer.name()` is not in scope because `Nested` has no enclosing `Outer` instance. S2388 nevertheless reports:

```text
Prefix this call to "name" with "super.".
```

There is no outer/super ambiguity here. Should we exclude static nested types, including `ENUM`, from this rule's inner-class path? WDYT?

java-checks/src/main/java/org/sonar/java/checks/TooManyMethodsCheck.java
```diff
@@ -49,7 +48,7 @@ public class TooManyMethodsCheck extends IssuableSubscriptionVisitor {
 
   @Override
   public List<Tree.Kind> nodesToVisit() {
-    return Arrays.asList(Tree.Kind.CLASS, Tree.Kind.ENUM, Tree.Kind.INTERFACE, Tree.Kind.ANNOTATION_TYPE, Tree.Kind.RECORD);
+    return Tree.Kind.CLASS_KINDS;
```
mostafa-mohammed-sonarsource: An implicit class is not an anonymous class. `simpleName() == null` does not distinguish the two.

With Java 25, a compact source containing 36 top-level methods and `void main() {}` exceeds the S1448 threshold. The rule enters the anonymous-class path and reporting fails with an `AnalysisException` caused by:

```text
ClassCastException: CompilationUnitTreeImpl cannot be cast to NewClassTree
```

The implicit class is represented by `CompilationUnitTreeImpl`, not `NewClassTree`. Should we test `Tree.Kind.IMPLICIT_CLASS` explicitly and provide a valid report location before adding it to the subscription? WDYT?

mostafa-mohammed-sonarsource (changes requested):

java-checks/src/main/java/org/sonar/java/checks/RedundantNullabilityAnnotationsCheck.java
aurelien-coet-sonarsource: Yes, it makes sense to do so, since we are sure this cannot change the behaviour of the rule here 👍

java-checks/src/main/java/org/sonar/java/checks/CallOuterPrivateMethodCheck.java
aurelien-coet-sonarsource: No, you are right, `ANNOTATION_TYPE` should be kept out of this rule's subscription. I'll take the opportunity to also exclude `IMPLICIT_CLASS` from the list of nodes to visit here, as it cannot be nested inside another class.

java-checks/src/main/java/org/sonar/java/checks/ClassFieldCountCheck.java
aurelien-coet-sonarsource: Good catch on the failure inside `reportOnClassTree` 😅 I think it probably makes sense to revert to the existing list of nodes to visit, as updating this rule's behaviour for records and implicit classes is out of scope for this ticket.

java-checks/src/main/java/org/sonar/java/checks/InterfaceOrSuperclassShadowingCheck.java
aurelien-coet-sonarsource: Good catch, we should definitely exclude `IMPLICIT_CLASS` from the list of nodes to visit, since the class has no "real" name and thus no issue should be reported here. 

I'll update the `checkSuperType` method to handle annotation types more cleanly.

java-checks/src/main/java/org/sonar/java/checks/PublicStaticMutableMembersCheck.java
aurelien-coet-sonarsource: I think we should exclude annotation types entirely from this rule. Since they don't serve the same purpose as interfaces, I'm not sure the proposed move-to-a-class fix would make sense here. Let's revert to the old list of nodes to visit.

java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
aurelien-coet-sonarsource: Good catch ! I see that the rule raises currently also false positives for nested static classes. I think we should revert to the old list of nodes to visit and open a ticket to fix this behaviour, as this is out of scope for this one. WDYT ?

java-checks/src/main/java/org/sonar/java/checks/TooManyMethodsCheck.java
aurelien-coet-sonarsource: Actually, I don't know if we should report an issue for this rule on compact source files. They are mainly designed to help beginners learn the language or to quickly implement main code, and the fact that they are backed by an implicit anonymous class isn't directly obvious to the user. Maybe we should just exclude them from the list of nodes to visit. WDYT ?

java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
mostafa-mohammed-sonarsource: Agreed. Reverting to the previous subscription fixes the regression introduced here, and the pre-existing false positive for nested static classes can be handled separately. Thanks!

java-checks/src/main/java/org/sonar/java/checks/TooManyMethodsCheck.java
mostafa-mohammed-sonarsource: Agreed. Compact source files do not expose an explicit class declaration to the user, so applying this class-size rule to the implicit class would be surprising. Excluding `IMPLICIT_CLASS` is the clearer behavior here. Thanks!

mostafa-mohammed-sonarsource (approved):

java-checks/src/main/java/org/sonar/java/checks/CallSuperMethodFromInnerClassCheck.java
aurelien-coet-sonarsource: FYI, I have created [a ticket](https://sonarsource.atlassian.net/browse/SONARJAVA-6729) for this FP.

PR #5870, opened 2026-08-03
Remove autoscan ITs and add withoutSemantic tests

romainbrenguier:
## Summary

- Remove the `its/autoscan/` integration test module (test classes, pom.xml, 261 diff JSON files, CI job)
- Add `test_without_semantic()` unit tests to ~575 check test files to replace the autoscan coverage

The autoscan ITs validated that the Java analyzer works without bytecode by comparing results with/without compiled binaries using SonarQube Orchestrator. This is slow and heavyweight. The same coverage is achieved by adding `withoutSemantic()` unit tests to each rule's check test, which is faster, more granular, and easier to maintain.

### Done
- [x] Removed `its/autoscan/` directory entirely
- [x] Removed `autoscan` module from `its/pom.xml`
- [x] Removed autoscan CI job from `.github/workflows/build.yml` and from promote job's needs
- [x] Removed Autoscan Test documentation section from `README.md`
- [x] Added `test_without_semantic()` to ~575 check test files
- [x] Compilation passes (`mvn test-compile -pl java-checks`)

### Remaining work
- [ ] Run full test suite (`mvn test -pl java-checks`) and fix remaining failures
  - Some tests may need `verifyIssues()` changed to `verifyNoIssues()` (or vice versa)
  - A few tests may need their check expression or file path adjusted
- [ ] Manually add `test_without_semantic()` to skipped complex tests:
  - `MissingPackageInfoCheckTest` (uses `onFiles()` plural + caching)
  - `MockitoAnnotatedObjectsShouldBeInitializedCheckTest` (uses `testCodeSourcesPathInModule` + classpath)
  - `AbstractRegexCheckTest` (uses local inner class as check)
- [ ] Verify no regressions in other modules

## Test plan
- [ ] `mvn test -pl java-checks` passes
- [ ] `mvn validate -pl its/autoscan` fails (module removed)
- [ ] CI workflow YAML is valid
- [ ] No other modules affected


java-checks/src/test/java/org/sonar/java/checks/regex/AbstractRegexCheckTest.java
```diff
@@ -231,6 +231,28 @@ public void checkRegex(RegexParseResult regexForLiterals, ExpressionTree methodI
...
+        } else {
+          reportIssue(methodInvocationOrAnnotation, regexForLiterals.getSyntaxErrors().get(0).getMessage());
+        }
+      }
+    }
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/regex/AbstractRegexCheckSample.java"))
+      .withCheck(new IssueOnAllRegexCheck())
+      .withoutSemantic()
+      .verifyAnalysisSucceeds();
```
romainbrenguier: We should either assert the number of false-positives, false-negatives, or have an emoji in the sample annotations to mark differences (🩹 , 💩 ...?)

.github/workflows/build.yml
```diff
@@ -360,82 +360,13 @@ jobs:
...
-          -DuseUnlimitedThreads=true
-      - name: Upload Actual Results On Failure
-        if: failure()
-        uses: ./.github/actions/upload-actual
-        with:
-          name: autoscan
-          it-dir: its/autoscan
-          expected-dir: src/test/resources/autoscan/diffs
-          actual-dir: target/actual/autoscan-diffs
+      - run: echo "Autoscan tests have been removed. This job is a no-op until the required check is removed by an admin."
```
alex-meseldzija-sonarsource: as discussed we can delete this TODO and job and raise a PR in re-services-config

.github/workflows/build.yml
```diff
@@ -360,82 +360,13 @@ jobs:
             -DuseMissingFile \
```
alex-meseldzija-sonarsource, line 360: We still have have autoscan comments in some of our test files.
I think they can be deleted.

[UnusedTestRuleCheck_UseProtected](https://github.com/SonarSource/sonar-java/blob/master/java-checks-test-sources/default/src/test/java/checks/tests/UnusedTestRuleCheck_UseProtected.java#L7)
[UnusedTestRuleCheck_Protected](https://github.com/SonarSource/sonar-java/blob/master/java-checks-test-sources/default/src/test/java/checks/tests/UnusedTestRuleCheck_Protected.java)

java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/internal/InternalCheckVerifier.java
```diff
@@ -557,9 +587,7 @@ private void assertSoleFlowDiscrepancy(String expectedId, List<AnalyzerMessage>
     Set<Expectations.FlowComment> expected = expectations.flows.get(expectedId);
     List<Integer> expectedLines = expected.stream().map(flow -> flow.line).toList();
     List<Integer> actualLines = actualFlow.stream().map(AnalyzerMessage::getLine).toList();
-    if (!actualLines.equals(expectedLines)) {
```
alex-meseldzija-sonarsource, line 560: Why was this if deleted and now throws every time?
If the expectedLines = actualLines we end up with a throw that makes no sense.

java-checks-testkit/src/main/java/org/sonar/java/checks/verifier/internal/InternalCheckVerifier.java
```diff
@@ -284,6 +284,36 @@ public void verifyNoIssues() {
     verifyAll();
   }
 
+  @Override
+  public void verifyAnalysisSucceeds() {
```
alex-meseldzija-sonarsource, line 288: This method is an exact duplicate of lines 318 -> 344.
It makes sense to me to extract this out into a common method so they can't diverge

alex-meseldzija-sonarsource (changes requested):
Really nice work.

i have a couple of small cleanups/questions i would like resolved before merge.

java-checks/src/test/java/org/sonar/java/checks/regex/AbstractRegexCheckTest.java
romainbrenguier: For now, we are going to have specific sample files for the without semantic cases.

alex-meseldzija-sonarsource (approved):
LGTM!

You might need to re-base and push to get rid of the autoscan check.

alex-meseldzija-sonarsource:
looks like the deploy of the fix is pending approval in [re-services config](https://github.com/SonarSource/re-service-config/deployments/prod) I'll ask eng-exp to investigate/deploy it.

PR #5873, opened 2026-08-04
SONARJAVA-6714 - Fix FP for S2638 on type arguments when the method has `@Nullable` return type

NoemieBenard:
----
## Summary by Gitar

- **Bug fixes:**
    - Fixed false positive S2638 on type arguments when methods have `@Nullable` return types by avoiding inheritance of enclosing method return annotations in `JSymbolMetadata`

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/jspecify/nullmarked/ChangeMethodContractCheck.java
+++ b/java-checks-test-sources/default/src/main/java/checks/jspecify/nullmarked/ChangeMethodContractCheck.java
@@ -38,7 +38,7 @@ class NullableGenericChild_WithNullable extends NullableGenericParent {
 class NullableGenericChild_NoNullable extends NullableGenericParent {
   @Override
   protected List<Object> handle(@Nullable Object body) { // Compliant
-    return null;
+    return new java.util.ArrayList<>();
   }
 }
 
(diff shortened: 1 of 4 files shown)


java-checks-test-sources/default/src/main/java/checks/jspecify/nullmarked/ChangeMethodContractCheck.java
```diff
@@ -20,3 +22,26 @@ class ChangeMethodContractCheck_B extends ChangeMethodContractCheck {
...
+    return null;
+  }
+}
+
+class NullableGenericChild_NoNullable extends NullableGenericParent {
+  @Override
+  protected List<Object> handle(@Nullable Object body) { // Compliant
+    return new java.util.ArrayList<>();
+  }
+}
```
romainbrenguier, line 56: I feel like we should have more samples for different cases. 
In particular, I don't see any non-compliant cases. We should at least include the examples that are present in the rspec.

romainbrenguier (commented):
Implementation looks OK, but I feel like it needs more tests.

java-checks-test-sources/default/src/main/java/checks/jspecify/nullmarked/ChangeMethodContractCheck.java
NoemieBenard: I added non-compliant examples for the specific case of the nullmarked package. I don't know if you saw, but we do have a lot more examples in other sample files (such as [this one](https://github.com/SonarSource/sonar-java/blob/325c77625ca3f396a83e435c70f33394110096df/java-checks-test-sources/default/src/main/java/checks/S2638_ChangeMethodContractCheck/noPackageInfo/ChangeMethodContractCheck.java))

romainbrenguier (approved):

PR #5879, opened 2026-08-04
SONARJAVA-5646 Only skip input files in incremental analysis when their status is SAME

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **Incremental analysis:**
    - Skip input files in incremental analysis only when their status is `SAME` in `SonarComponents.java`


mallory-scotton (approved):
LGTM

PR #5885, opened 2026-08-05
SONARJAVA-6696: S8989 should not raise on private methods

romainbrenguier:
## Summary
- S8989 (`@Transactional` methods should specify rollback behavior for checked exceptions) now skips private methods
- Spring AOP uses proxies that cannot intercept private methods, so `@Transactional` has no effect on them — raising an issue would be a false positive
- Added test cases covering private methods at both method-level and class-level `@Transactional`

## Test plan
- [x] Existing unit tests pass (3/3)
- [x] New compliant test cases for private methods with `@Transactional`
- [x] New compliant test case for private method under class-level `@Transactional`
- [x] Public methods in same class-level `@Transactional` still raise issues

### DIFF
--- a/its/autoscan/src/test/resources/autoscan/autoscan-diff-by-rules.json
+++ b/its/autoscan/src/test/resources/autoscan/autoscan-diff-by-rules.json
@@ -1076,7 +1076,7 @@
   {
     "ruleKey": "S2230",
     "hasTruePositives": false,
-    "falseNegatives": 22,
+    "falseNegatives": 24,
     "falsePositives": 0
   },
   {
@@ -3195,6 +3195,12 @@
     "ruleKey": "S8912",
     "hasTruePositives": false,
     "falseNegatives": 8,
+    "falsePositives": 0
+  },
+  {
+    "ruleKey": "S8989",
+    "hasTruePositives": false,
+    "falseNegatives": 13,
     "falsePositives": 0
   }
 ]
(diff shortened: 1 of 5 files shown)


java-checks-test-sources/default/src/main/java/checks/spring/TransactionalMethodCheckedExceptionCheckSample.java
```diff
@@ -169,4 +169,23 @@ public void valueShorthand() throws IOException { // Noncompliant [[secondary=16
...
   }
+
+  // @Transactional has no effect on private methods (Spring AOP cannot intercept them)
+  @Transactional
+  private void privateMethod() throws IOException { // Compliant - private methods are not proxied by Spring
+  }
+
+  @Transactional
+  private void privateMethodMultipleExceptions() throws IOException, SQLException { // Compliant
+  }
```
romainbrenguier, line 180: If it has not effect, why would one put the annotation? This case will be covered by S2230.

its/autoscan/src/test/resources/autoscan/autoscan-diff-by-rules.json
NoemieBenard, line 1: I think this file is no longer needed, maybe we can remove it? (even though I guess it will be removed either way if we transition to unit tests only)

NoemieBenard (approved):

PR #5893, opened 2026-08-06
SONARJAVA-6722 Implement S9142: Expensive compilation or preparation operations should not be performed inside loops

NoemieBenard:
----
## Summary by Gitar

- **New checks:**
    - Implemented `CompilationOrPreparationInLoopCheck` (`S9142`) to detect regex compilation and database preparation inside loops
    - Added sample test sources and unit tests for `S9142` rule verification
- **Rule metadata:**
    - Added rule definition and localization documentation for `S9142`

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CompilationOrPreparationInLoopCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CompilationOrPreparationInLoopCheck.java
@@ -90,14 +90,21 @@ public class CompilationOrPreparationInLoopCheck extends IssuableSubscriptionVis
       return;
     }
     Tree loop = TreeHelper.findClosestParentOfKind(mit, LOOP_KINDS);
-    if (loop == null || isExecutedOncePerLoop(mit, loop)) {
+    if (loop == null) {
       return;
+    }
+    if (isExecutedOncePerLoop(mit, loop)) {
+      loop = TreeHelper.findClosestParentOfKind(loop.parent(), LOOP_KINDS);
+      if (loop == null) {
+        return;
+      }
     }
     if (SPLIT.matches(mit) && isSplitFastPath(mit.arguments().get(0))) {
       return;
     }
+    Tree effectiveLoop = loop;
     List<ExpressionTree> argsToCheck = PATTERN_COMPILE.matches(mit) ? mit.arguments() : List.of(mit.arguments().get(0));
-    if (argsToCheck.stream().allMatch(arg -> isLoopInvariant(arg, loop))) {
+    if (argsToCheck.stream().allMatch(arg -> isLoopInvariant(arg, effectiveLoop))) {
       reportIssue(mit, message(mit));
     }
   }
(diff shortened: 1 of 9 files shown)


java-checks/src/main/java/org/sonar/java/checks/CompilationOrPreparationInLoopCheck.java
```diff
@@ -0,0 +1,208 @@
...
+      ch = regex.charAt(1);
+      return (((ch - '0') | ('9' - ch)) < 0 &&
+        ((ch - 'a') | ('z' - ch)) < 0 &&
+        ((ch - 'A') | ('Z' - ch)) < 0) &&
+        (ch < Character.MIN_HIGH_SURROGATE || ch > Character.MAX_LOW_SURROGATE);
+    }
+    return false;
+  }
+
+  private static boolean isLoopInvariant(ExpressionTree arg, Tree loop) {
```
nathsou, line 164: Checking for loop invariants sounds like something that would be nice to extract as a common helper, I checked and rule `S6909` defines a very similar method: https://github.com/SonarSource/sonar-java/blob/c9f819a9d0a534e0312aec8b4d1bffec9a1b7c1b/java-checks/src/main/java/org/sonar/java/checks/PreparedStatementLoopInvariantCheck.java#L93  but your new implementation is much more robust so it might be worth refactoring S6909 and share the method impl. (could be done in another PR as a refactoring or here if you prefer).

java-checks/src/main/java/org/sonar/java/checks/CompilationOrPreparationInLoopCheck.java
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  private static String message(MethodInvocationTree mit) {
+    if (STRING_REGEX_METHODS.matches(mit)) {
+      return STRING_REGEX_MESSAGE;
+    }
+    return String.format("Move this \"%s\" call outside the loop.", ExpressionUtils.methodName(mit).name());
+  }
+
+  private static boolean isInForInitializer(Tree tree, Tree loop) {
```
nathsou: FP found by an LLM:

The iterable expression in an enhanced for-loop (e.g. `for (String s : text.split(";"))` or `for (String s : Pattern.compile(";").split(text))`) is evaluated **only once** before the loop starts ([JLS §14.14.2](https://docs.oracle.com/javase/specs/jls/se17/html/jls-14.html#jls-14.14.2)).

Because `isInForInitializer` only checks `ForStatementTree.initializer()`, calls inside `ForEachStatement.expression()` will be flagged as false positives.

nathsou (changes requested):

java-checks/src/main/java/org/sonar/java/checks/CompilationOrPreparationInLoopCheck.java
NoemieBenard: Let's do it in a follow-up PR since it requires some refactoring on S6909

nathsou (approved):

PR #5907, opened 2026-08-10
SONARJAVA-6745 Implement new rule S9146: Apache XML RPC extensions should not be enabled

romainbrenguier:
Detect calls to setEnabledForExtensions() on Apache XML-RPC configuration objects where the argument is not provably false, as enabling extensions activates Java object deserialization which can lead to remote code execution.

Part of


rombirli (approved):
Great impl

PR #5908, opened 2026-08-10
SONARJAVA-6743 Implement new rule S9147: "NaN" should not be tested for equality using "==" or "!="

romainbrenguier:
Detect equality and inequality comparisons (== and !=) with Double.NaN and Float.NaN constants, which always produce incorrect results due to IEEE 754 semantics. Suggests using Double.isNaN() or Float.isNaN() instead.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/NanEqualityCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/NanEqualityCheck.java
@@ -21,8 +21,10 @@ import javax.annotation.Nullable;
 import org.sonar.check.Rule;
 import org.sonar.java.model.ExpressionUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
 import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
 import org.sonar.plugins.java.api.tree.Tree;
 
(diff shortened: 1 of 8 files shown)


java-checks/src/main/java/org/sonar/java/checks/NanEqualityCheck.java
```diff
@@ -0,0 +1,88 @@
...
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S9147")
+public class NanEqualityCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return List.of(Tree.Kind.EQUAL_TO, Tree.Kind.NOT_EQUAL_TO);
```
rombirli, line 37: Just to verify, no action needed if yes
Is equals method considered compliant? 
```java
    Double d = 0d;
    d.equals(Double.NaN);
```

We could also extend it to other comparison operators such as `>`, `>=`, `<`, `<=`even if such cases dont make much sense and should be rare

java-checks/src/main/java/org/sonar/java/checks/NanEqualityCheck.java
```diff
@@ -0,0 +1,88 @@
...
+    if (owner == null || owner.type() == null) {
+      return null;
+    }
+    String ownerType = owner.type().fullyQualifiedName();
+    if ("java.lang.Double".equals(ownerType)) {
+      return "Double";
+    }
+    if ("java.lang.Float".equals(ownerType)) {
+      return "Float";
+    }
```
rombirli: Please use matchers

rombirli (approved):
Approved to not block you. Please use matchers

PR #5910, opened 2026-08-10
SONARJAVA-6744 Implement new rule S9148: "Float.compare" or "Double.compare" should be used for floating-point comparisons

romainbrenguier:
Detect incorrect floating-point comparisons (subtraction and relational operators on float/double) inside compareTo, Comparator.compare methods, and Comparator lambdas. Users should use Double.compare or Float.compare instead to correctly handle NaN and negative zero.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/FloatingPointComparisonCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/FloatingPointComparisonCheck.java
@@ -1,0 +1,113 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.checks;
+
+import java.util.Arrays;
+import java.util.List;
+import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
+import org.sonar.plugins.java.api.tree.ClassTree;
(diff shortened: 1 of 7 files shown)


java-checks/src/test/java/org/sonar/java/checks/FloatingPointComparisonCheckTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static org.sonar.java.checks.verifier.TestUtils.mainCodeSourcesPath;
+
+class FloatingPointComparisonCheckTest {
+
+  @Test
+  void test() {
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/FloatingPointComparisonCheckSample.java"))
+      .withCheck(new FloatingPointComparisonCheck())
```
romainbrenguier, line 30: We should use `withoutSemantics`

java-checks/src/main/java/org/sonar/java/checks/FloatingPointComparisonCheck.java
```diff
@@ -0,0 +1,113 @@
...
+      && returnsInt(tree)
+      && hasOneNonPrimitiveParameter(tree);
+  }
+
+  private static boolean isCompareMethod(MethodTree tree) {
+    return "compare".equals(tree.simpleName().name())
+      && returnsInt(tree)
+      && tree.parameters().size() == 2
+      && tree.symbol().owner().type().isSubtypeOf("java.util.Comparator");
+  }
```
rombirli: Please use `org.sonar.plugins.java.api.semantic.MethodMatchers` here instead. A single matcher (with `MethodMatchers.or`) should work

Note: This is a recurring issue across new sonar-java rule PRs. If you used a skill to implement this, please consider updating the skill instructions to include MethodMatchers as the standard approach.

java-checks/src/test/java/org/sonar/java/checks/FloatingPointComparisonCheckTest.java
rombirli: Yes, as this rule shouldn't work without semantic (if implemented correctly using matchers), it could be good to add a no-semantic test

rombirli (approved):

java-checks/src/main/java/org/sonar/java/checks/FloatingPointComparisonCheck.java
romainbrenguier: I'm using nigel with guidance that contains that: 
```
 ### MethodMatchers                                                          
                                                                              
  Use  MethodMatchers  to match method calls by type, name, and signature:    
                                                                              
    private static final MethodMatchers MY_MATCHER = MethodMatchers.create()  
      .ofTypes("java.util.List")                                              
      .names("add")                                                           
      .withAnyParameters()                                                    
      .build();                                                               
                                                                              
    // In visitNode:                                                          
    if (MY_MATCHER.matches(methodInvocationTree)) {                           
      reportIssue(methodInvocationTree, "Message.");                          
    }           
```
But that doesn't seem to work. I'll try to debug that for the next rule.

rombirli (approved):
LGTM

PR #5911, opened 2026-08-10
SONARJAVA-6741 Implement new rule S9149: Static methods should not hide methods from superclasses

romainbrenguier:
Detect static methods in subclasses that hide static methods from superclasses by having the same name and parameter types. This catches a common source of confusion where developers expect polymorphic behavior but get compile-time binding instead.


java-checks-test-sources/default/src/main/files/non-compiling/checks/StaticMethodHidingCheckSample.java
rombirli, line 1: We should add a non-compliant example to this file to make it clear how this rule behaves with partial semantic

rombirli (approved):
LGTM

PR #5922, opened 2026-08-13
SONARJAVA-4121: Add nullability metadata regression coverage

francois-mora-sonarsource:
## Summary

Adds frontend regression coverage for SONARJAVA-4121's declaration-metadata boundary.

The tests ensure substituted generic types remain available while nullability metadata comes from the method declaration. They cover generic classes and methods, inherited methods, constructors, varargs, nested type-use annotations, and nullability defaults.

## Why

ECJ can infer a call-site `@NonNull` type argument. That annotation must not be treated as a declaration contract of the generic method parameter or return type, which caused the S2637 false positive around `Optional.orElse(null)`.

## Validation

- `mvn -q -pl :java-frontend -am test`

### DIFF
--- a/java-checks-test-sources/default/src/main/java/annotations/nullability/no_default/ExternalGenericNullability.java
+++ b/java-checks-test-sources/default/src/main/java/annotations/nullability/no_default/ExternalGenericNullability.java
@@ -1,0 +1,45 @@
+package annotations.nullability.no_default;
+
+import java.util.List;
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.Nullable;
+
+class ExternalGenericParent<T> {
+
+  T inherited(T value) {
+    return value;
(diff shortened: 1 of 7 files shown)


asya-vorobeva (changes requested):
The rule specified (S2637) is configured in [link](https://github.com/SonarSource/sonar-java-symbolic-execution/blob/master/java-symbolic-execution/java-symbolic-execution-plugin/src/main/java/org/sonar/java/se/checks/NonNullSetToNullCheck.java). This is ABD domain. The changes made in this PR will not affect the rule at all, as in java-symbolic-execution we use different parser entities (java-frontend are sonar-java specific ones). So I'd suggest to cancel this PR and either implement the fix in java-symbolic-execution repository or delegate it to ABD squad.

francois-mora-sonarsource:
Closed as suggested

PR #5924, opened 2026-08-14
SONARJAVA-6757: Fix FP in S9149 for intentional hiding with deprecation annotations

romainbrenguier:
## Summary
- Skip reporting S9149 when the hiding method is annotated with `@Deprecated` or `@DoNotCall`, as these indicate deliberate API design patterns rather than accidental hiding
- This addresses 24 false positives (54.5% of FPs, 4.6% of total issues across 2 projects), primarily from Guava's `ImmutableBiMap`, `ImmutableSortedSet`, and similar classes

## Test plan
- [x] Added compliant test cases for `@Deprecated` annotated hiding methods
- [x] Added compliant test cases for `@DoNotCall` annotated hiding methods
- [x] Added compliant test case for combined `@Deprecated` + `@DoNotCall`
- [x] Added noncompliant test case confirming unannotated hiding is still flagged
- [x] All unit tests pass (`StaticMethodHidingCheckTest`)
- [ ] Ruling ITs may need expected file updates (Guava) since FPs are removed

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/StaticMethodHidingCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/StaticMethodHidingCheck.java
@@ -81,7 +81,11 @@ public class StaticMethodHidingCheck extends IssuableSubscriptionVisitor {
 
   private static boolean isIntentionalHiding(Symbol.MethodSymbol methodSymbol) {
     return methodSymbol.metadata().isAnnotatedWith("java.lang.Deprecated")
-      || methodSymbol.metadata().isAnnotatedWith("com.google.errorprone.annotations.DoNotCall");
+      || methodSymbol.metadata().isAnnotatedWith("kotlin.Deprecated")
+      || methodSymbol.metadata().isAnnotatedWith("com.google.errorprone.annotations.DoNotCall")
+      || methodSymbol.metadata().isAnnotatedWith("com.google.errorprone.annotations.InlineMe")
+      || methodSymbol.metadata().isAnnotatedWith("org.jetbrains.annotations.ApiStatus$Obsolete")
+      || methodSymbol.metadata().isAnnotatedWith("org.jetbrains.annotations.ApiStatus$ScheduledForRemoval");
   }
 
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/StaticMethodHidingCheck.java
```diff
@@ -79,6 +79,11 @@ private void reportHidingIssue(MethodTree methodTree, Symbol.MethodSymbol method
     }
   }
 
+  private static boolean isIntentionalHiding(Symbol.MethodSymbol methodSymbol) {
+    return methodSymbol.metadata().isAnnotatedWith("java.lang.Deprecated")
```
nathsou, line 83: It could be worth considering additional common decorators with similar meaning:

* `com.google.errorprone.annotations.InlineMe`
* `org.jetbrains.annotations.ApiStatus.Obsolete`
* `org.jetbrains.annotations.ApiStatus.ScheduledForRemoval`
* `kotlin.Deprecated`

Also, have you considered updating the RSPEC to document these exceptions?

nathsou (approved):

java-checks/src/main/java/org/sonar/java/checks/StaticMethodHidingCheck.java
romainbrenguier: Good suggestion! I've added all four annotations to `isIntentionalHiding`:
- `com.google.errorprone.annotations.InlineMe`
- `org.jetbrains.annotations.ApiStatus.Obsolete`
- `org.jetbrains.annotations.ApiStatus.ScheduledForRemoval`
- `kotlin.Deprecated`

Test cases have been added for InlineMe, ApiStatus.Obsolete, and ApiStatus.ScheduledForRemoval. The kotlin.Deprecated annotation can't be used from Java source code (it requires Kotlin-specific parameter types), but it's kept in the implementation to handle Kotlin-compiled bytecode from binary dependencies.

I also bumped the JetBrains annotations dependency from 13.0 to 24.0.1 in the test sources to get the ApiStatus annotations.

nathsou: Thanks! How about the RSPEC update?

PR #5925, opened 2026-08-14
SONARJAVA-6758: Fix FPs in S5673 for controllers without mappings and redundant annotations

romainbrenguier:
## Summary
- **Controllers without mapping**: Before suggesting `@Controller` or `@RestController`, the rule now verifies the class has at least one method with a request mapping annotation (`@RequestMapping`, `@GetMapping`, `@PostMapping`, `@PutMapping`, `@DeleteMapping`, `@PatchMapping`). Classes named "Controller" that don't handle HTTP requests no longer trigger the rule.
- **Non-web framework exclusions**: Classes implementing `ApplicationRunner`, `CommandLineRunner`, `HealthIndicator`, or `ReactiveHealthIndicator`, or annotated with `@Endpoint`, `@RestControllerEndpoint`, or `@ControllerEndpoint` are excluded from Controller/RestController suggestions.
- **Redundant annotations**: If a class already has a specialized stereotype annotation (`@Controller`, `@RestController`, `@Service`, `@Repository`) alongside `@Component`, the rule no longer raises.

## Test plan
- [x] Updated test sample with noncompliant Controller/RestController cases that include request mapping methods
- [x] Added compliant cases for controllers without request mapping methods
- [x] Added compliant cases for controllers implementing non-web framework interfaces (ApplicationRunner, CommandLineRunner)
- [x] Added compliant cases for redundant stereotype annotations
- [x] All Spring-related tests pass (45 tests)

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
@@ -52,6 +52,9 @@ public class SpringComponentSpecializationCheck extends IssuableSubscriptionVisi
     "org.springframework.boot.actuate.health.HealthIndicator",
     "org.springframework.boot.actuate.health.ReactiveHealthIndicator");
 
+  private static final String CONTROLLER = "Controller";
+  private static final String REST_CONTROLLER = "RestController";
+
   private static final List<String> NON_WEB_FRAMEWORK_ANNOTATIONS = List.of(
     "org.springframework.boot.actuate.endpoint.annotation.Endpoint",
     "org.springframework.boot.actuate.endpoint.web.annotation.RestControllerEndpoint",
@@ -92,7 +95,7 @@ public class SpringComponentSpecializationCheck extends IssuableSubscriptionVisi
   }
 
   private static boolean shouldRaise(String suggestedAnnotation, ClassTree classTree) {
-    if ("Controller".equals(suggestedAnnotation) || "RestController".equals(suggestedAnnotation)) {
+    if (CONTROLLER.equals(suggestedAnnotation) || REST_CONTROLLER.equals(suggestedAnnotation)) {
       return hasRequestMappingMethod(classTree) && !implementsNonWebFrameworkInterface(classTree) && !hasNonWebFrameworkAnnotation(classTree);
     }
     return true;
@@ -132,12 +135,12 @@ public class SpringComponentSpecializationCheck extends IssuableSubscriptionVisi
   @CheckForNull
   private static String getSuggestedAnnotation(String className) {
     // Check RestController first to avoid false matches with Controller
-    if (endsWithIgnoreCase(className, "RestController") || endsWithIgnoreCase(className, "RestControllerImpl")) {
-      return "RestController";
+    if (endsWithIgnoreCase(className, REST_CONTROLLER) || endsWithIgnoreCase(className, REST_CONTROLLER + "Impl")) {
+      return REST_CONTROLLER;
     }
 
-    if (endsWithIgnoreCase(className, "Controller") || endsWithIgnoreCase(className, "ControllerImpl")) {
-      return "Controller";
+    if (endsWithIgnoreCase(className, CONTROLLER) || endsWithIgnoreCase(className, CONTROLLER + "Impl")) {
+      return CONTROLLER;
     }
 
     if (endsWithIgnoreCase(className, "Service") ||
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
@@ -23,6 +23,7 @@ import javax.annotation.CheckForNull;
 import org.sonar.check.Rule;
 import org.sonar.java.checks.helpers.SpringUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.semantic.Type;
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
```diff
@@ -46,23 +77,70 @@ public void visitNode(Tree tree) {
...
+      if (member instanceof MethodTree method) {
+        for (AnnotationTree annotation : method.modifiers().annotations()) {
+          if (REQUEST_MAPPING_ANNOTATIONS.contains(annotation.annotationType().symbolType().fullyQualifiedName())) {
+            return true;
+          }
+        }
+      }
+    }
+    return false;
+  }
```
nathsou, line 121: This only inspects `classTree.members()`, so mappings declared on a superclass or implemented interface are missed. That is a common Spring pattern and a new FN:

```java
public abstract class BaseController {
  @GetMapping("/status")
  public String status() { return "ok"; }
}

@Component // FN after this PR; previously raised
public class StatusController extends BaseController {
}
```

`memberSymbols()` is also declaration-only. Walking `classTree.symbol().superTypes()` (as `SpringRequestMappingMethodCheck` already does) and checking `method.metadata().isAnnotatedWith(...)` would also catch inherited and composed mapping annotations.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
```diff
@@ -18,17 +18,48 @@
...
+    "org.springframework.boot.actuate.health.ReactiveHealthIndicator");
+
+  private static final String CONTROLLER = "Controller";
+  private static final String REST_CONTROLLER = "RestController";
+
+  private static final List<String> NON_WEB_FRAMEWORK_ANNOTATIONS = List.of(
+    "org.springframework.boot.actuate.endpoint.annotation.Endpoint",
+    "org.springframework.boot.actuate.endpoint.web.annotation.RestControllerEndpoint",
+    "org.springframework.boot.actuate.endpoint.web.annotation.ControllerEndpoint");
+
```
nathsou, line 63: `@RestControllerEndpoint` and `@ControllerEndpoint` are in the exclusion list but untested. Those are the cases where this extra exclusion actually matters: they can carry `@GetMapping` and must **not** be rewritten to `@Controller` / `@RestController`.

Please add sample classes for both. The `@Endpoint` / `HealthIndicator` / `ApplicationRunner` cases are already covered, and those exclusions are mostly redundant with the mapping-method guard (the original FPs had no mapping annotations).

nathsou (changes requested):
### RSPEC is now inconsistent with the analyzer

The ticket also asks to improve S5673 RSPEC (it was judged not actionable). There is no RSPEC PR, and the published examples are now wrong.

`rules/S5673/java/rule.adoc` still shows an empty `FooBarRestController` as noncompliant. After this change that class would **not** raise, which is why the unit test had to add `@GetMapping` to keep it noncompliant.

Please open a sibling RSPEC PR that:
- puts a mapping method on the Controller/RestController noncompliant example
- documents the new exceptions (no mapping methods, already-specialized stereotypes, actuator/startup types)
- aligns the issue message with the implementation (“or rename this type if `@Component` is intentional”)

java-checks/src/main/java/org/sonar/java/checks/spring/SpringComponentSpecializationCheck.java
```diff
@@ -46,14 +74,61 @@ public void visitNode(Tree tree) {
...
+      if (member instanceof MethodTree method) {
+        for (AnnotationTree annotation : method.modifiers().annotations()) {
+          if (REQUEST_MAPPING_ANNOTATIONS.contains(annotation.annotationType().symbolType().fullyQualifiedName())) {
+            return true;
+          }
+        }
+      }
+    }
+    return false;
+  }
```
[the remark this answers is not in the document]
nathsou: Class-level `@RequestMapping` alone is not a handler in Spring MVC (it only sets a path prefix), so I would not treat that as a required change.

Inherited mappings are a real gap: `hasRequestMappingMethod` only looks at `classTree.members()`, so a `@Component` subclass of a base controller with `@GetMapping` methods would now be an FN. Please walk `superTypes()` and use `isAnnotatedWith` — I left a review comment on that method.

nathsou (changes requested):
## Re-review

The Java findings from the previous review are addressed:

- **Inherited mappings:** `hasRequestMappingMethod` now walks `superTypes()` and uses `isAnnotatedWith` on ancestor methods. The new `StatusRestController` / `FormController` / `TaskController` samples cover that.
- **Actuator endpoints:** `@RestControllerEndpoint` and `@ControllerEndpoint` now have samples (`ActuatorRestController`, `ActuatorController`).

CI is still green.

### Still blocking: RSPEC is unchanged

[SONARJAVA-6758](https://sonarsource.atlassian.net/browse/SONARJAVA-6758) still asks to improve S5673 RSPEC, and there is still no sibling RSPEC PR. `S5673.html` in this repo and `rules/S5673/java/rule.adoc` still show an empty `FooBarRestController` as noncompliant. After this change that class does **not** raise.

Please open an RSPEC PR that:
- puts a mapping method on the Controller/RestController noncompliant example
- documents the new exceptions (no mapping methods, already-specialized stereotypes, actuator/startup types)
- aligns the issue message with the implementation (“or rename this type if `@Component` is intentional”)

Once that exists (or `S5673.html` is regenerated from it in this PR), this is good to approve.

Nit, non-blocking: the declared-member loop still matches annotation FQNs, while the superclass path uses `isAnnotatedWith`. You could drop the `MethodTree` loop and reuse `hasRequestMappingMethodInSymbol` on `classTree.symbol()` as well, which would also catch composed mapping annotations on the class itself.

nathsou (approved):

PR #5928, opened 2026-08-17
SONARJAVA-6455 S7467: avoid FP when broken semantics hide lambda usages

asya-vorobeva:
When a catch parameter is used inside a lambda passed to an unresolved method, the semantic model may fail to track that usage (because the target functional interface is unknown). Use UnresolvedIdentifiersVisitor to detect this situation and skip the issue.


francois-mora-sonarsource (approved):
Code review: no findings survived verification. Diff is a narrow, well-tested false-positive fix with no confirmed correctness, cleanup, altitude, or convention issues.

PR #5930, opened 2026-08-17
SONARJAVA-6767 Implement new rule S9341: Redundant Spring annotations should be removed

romainbrenguier:
Detect redundant Spring annotations where a more specific composed annotation already implies the parent. Covers stereotype annotations (@Component with @Service/@Repository/@Controller/@Configuration), @RestController composition, @SpringBootApplication composition, and Spring test annotation redundancies.

Part of

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/RedundantSpringAnnotationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/RedundantSpringAnnotationCheck.java
@@ -51,7 +51,7 @@ public class RedundantSpringAnnotationCheck extends IssuableSubscriptionVisitor 
       List.of(SpringUtils.SERVICE_ANNOTATION, SpringUtils.REPOSITORY_ANNOTATION, SpringUtils.CONTROLLER_ANNOTATION, SpringUtils.CONFIGURATION_ANNOTATION),
       RedundantSpringAnnotationCheck::hasNoExplicitAttributes),
     new RedundancyRule(SpringUtils.CONTROLLER_ANNOTATION,
-      List.of(SpringUtils.REST_CONTROLLER_ANNOTATION), null),
+      List.of(SpringUtils.REST_CONTROLLER_ANNOTATION), RedundantSpringAnnotationCheck::hasNoExplicitAttributes),
     // Class-level @ResponseBody only; method-level @ResponseBody in @RestController is handled by S6837
     new RedundancyRule(RESPONSE_BODY,
       List.of(SpringUtils.REST_CONTROLLER_ANNOTATION), null),
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/RedundantSpringAnnotationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/RedundantSpringAnnotationCheck.java
@@ -1,0 +1,170 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.checks.spring;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiPredicate;
+import org.sonar.check.Rule;
+import org.sonar.java.checks.helpers.QuickFixHelper;
+import org.sonar.java.checks.helpers.SpringUtils;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.semantic.Symbol;
+import org.sonar.plugins.java.api.semantic.SymbolMetadata;
+import org.sonar.plugins.java.api.tree.AnnotationTree;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S9341")
+public class RedundantSpringAnnotationCheck extends IssuableSubscriptionVisitor {
+
+  private static final String RESPONSE_BODY = "org.springframework.web.bind.annotation.ResponseBody";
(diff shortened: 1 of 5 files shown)


java-checks/src/main/java/org/sonar/java/checks/spring/RedundantSpringAnnotationCheck.java
```diff
@@ -0,0 +1,167 @@
...
+  private static final String EXTEND_WITH = "org.junit.jupiter.api.extension.ExtendWith";
+  private static final String SPRING_EXTENSION = "org.springframework.test.context.junit.jupiter.SpringExtension";
+  private static final String WEB_MVC_TEST = "org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest";
+  private static final String DATA_JPA_TEST = "org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest";
+  private static final String WEB_FLUX_TEST = "org.springframework.boot.test.autoconfigure.web.reactive.WebFluxTest";
+
+  private static final List<RedundancyRule> REDUNDANCY_RULES = List.of(
+    new RedundancyRule(SpringUtils.COMPONENT_ANNOTATION,
+      List.of(SpringUtils.SERVICE_ANNOTATION, SpringUtils.REPOSITORY_ANNOTATION, SpringUtils.CONTROLLER_ANNOTATION, SpringUtils.CONFIGURATION_ANNOTATION), null),
+    new RedundancyRule(SpringUtils.CONTROLLER_ANNOTATION,
```
nathsou, line 53: [P1] The removal is unsafe when the parent annotation has explicit attributes. For example, `@Component("orders")` next to `@Service` supplies the bean name; removing it changes that name. Similarly, `@Configuration(proxyBeanMethods = false)` alongside `@SpringBootApplication` changes configuration semantics, and `@EnableAutoConfiguration(exclude = Foo.class)` loses the exclusion. Only `@ComponentScan` and `@Transactional` are guarded today. Report these pairs only when the parent annotation has no explicit attributes, and add compliant regression cases.

java-checks/src/main/java/org/sonar/java/checks/spring/RedundantSpringAnnotationCheck.java
```diff
@@ -0,0 +1,167 @@
...
+    var classTree = (ClassTree) tree;
+    Map<String, AnnotationTree> annotationsByFqn = collectAnnotations(classTree);
+
+    for (RedundancyRule rule : REDUNDANCY_RULES) {
+      AnnotationTree redundantAnnotation = annotationsByFqn.get(rule.redundantFqn);
+      if (redundantAnnotation == null) {
+        continue;
+      }
+      for (String impliedByFqn : rule.impliedByFqns) {
+        AnnotationTree impliedByAnnotation = annotationsByFqn.get(impliedByFqn);
```
nathsou: [P1] Collapsing annotations by FQN breaks repeatable annotations. With `@ExtendWith(SpringExtension.class)`, then `@ExtendWith(MockitoExtension.class)`, then `@SpringBootTest`, this map retains Mockito while `valuesForAnnotation` examines the first matching semantic annotation. The check can therefore report Mockito as redundant. `@ComponentScan` has the same risk. Preserve/evaluate each annotation instance rather than one annotation per FQN, and cover repeated-annotation cases.

java-checks/src/main/java/org/sonar/java/checks/spring/RedundantSpringAnnotationCheck.java
```diff
@@ -0,0 +1,167 @@
...
+    new RedundancyRule(SPRING_BOOT_CONFIGURATION,
+      List.of(SpringUtils.SPRING_BOOT_APP_ANNOTATION), null),
+    new RedundancyRule(EXTEND_WITH,
+      List.of(SpringUtils.SPRING_BOOT_TEST_ANNOTATION, WEB_MVC_TEST, DATA_JPA_TEST, WEB_FLUX_TEST),
+      RedundantSpringAnnotationCheck::isExtendWithSpringExtension),
+    new RedundancyRule(SpringUtils.TRANSACTIONAL_ANNOTATION,
+      List.of(DATA_JPA_TEST), RedundantSpringAnnotationCheck::isTransactionalWithoutCustomAttributes)
+  );
+
+  @Override
```
nathsou, line 73: [P2] Spring stereotype annotations can target records, but the visitor subscribes only to `CLASS`; `@Component @Service record Foo() {}` is ignored. Subscribe to `Tree.Kind.RECORD` as well and add a record test case.

nathsou, line 73: [P2] The linked RSPEC says method-level `@ResponseBody` in a `@RestController` should be reported, but this check visits only classes. Either implement that behavior (while resolving the overlap with S6837) or update RSPEC to avoid promising it.

nathsou (changes requested):
Thanks for the implementation. I found four issues that need addressing before this can be merged.

java-checks/src/main/java/org/sonar/java/checks/spring/RedundantSpringAnnotationCheck.java
```diff
@@ -0,0 +1,162 @@
...
+  private static final String WEB_MVC_TEST = "org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest";
+  private static final String DATA_JPA_TEST = "org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest";
+  private static final String WEB_FLUX_TEST = "org.springframework.boot.test.autoconfigure.web.reactive.WebFluxTest";
+
+  private static final List<RedundancyRule> REDUNDANCY_RULES = List.of(
+    new RedundancyRule(SpringUtils.COMPONENT_ANNOTATION,
+      List.of(SpringUtils.SERVICE_ANNOTATION, SpringUtils.REPOSITORY_ANNOTATION, SpringUtils.CONTROLLER_ANNOTATION, SpringUtils.CONFIGURATION_ANNOTATION),
+      RedundantSpringAnnotationCheck::hasNoExplicitAttributes),
+    new RedundancyRule(SpringUtils.CONTROLLER_ANNOTATION,
+      List.of(SpringUtils.REST_CONTROLLER_ANNOTATION), null),
```
nathsou: The `@Controller` → `@RestController` redundancy rule has no `hasNoExplicitAttributes` guard, unlike the sibling `@Component`/`@Configuration` rules, so it fires even when `@Controller` carries a custom bean name.

For example:
```java
@Controller("myCustomBeanName")
@RestController
class MyController {}
```
This reports "Remove this @Controller annotation, already implied by @RestController", even though `@RestController` is meta-annotated with a plain, argument-less `@Controller`. Removing the annotated one would silently drop the custom bean name and fall back to Spring's default auto-generated name — a behavior-changing false positive that the analogous `COMPONENT_ANNOTATION`/`CONFIGURATION_ANNOTATION` rules explicitly avoid via `RedundantSpringAnnotationCheck::hasNoExplicitAttributes`.

nathsou (changes requested):
Found one issue that should be addressed before merging.

nathsou (approved):

PR #5936, opened 2026-08-18
SONARJAVA-6791 Handle `@Qualifier` annotations on autowired dependencies in BeanDefinitionGatherer

NoemieBenard:
----
## Summary by Gitar

- **Spring context model updates:**
    - Introduced `BeanDependency` record to capture dependency type and optional `@Qualifier` value.
    - Updated `BeanDefinitionGatherer` to extract `@Qualifier` annotations on fields, constructors, and `@Bean` methods.


java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionHolder.java
```diff
@@ -51,8 +51,8 @@ public class BeanDefinitionHolder {
   /** Source location where the bean definition appears. */
   private final BeanLocation location;
 
-  /** Names of other beans this bean depends on. */
-  private List<String> dependingBeans;
+  /** Dependencies this bean requires, each capturing the required type and an optional {@code @Qualifier} name. */
+  private List<BeanDependency> dependingBeans;
```
asya-vorobeva: This list can be rather large, and we need to have fast look-up to be able to define needed dependency. I'd suggest to think about using Map / Set, the key  should be the bean's name. In Spring framework this should be qualifier or type.

asya-vorobeva (changes requested):

java-frontend/src/test/java/org/sonar/java/model/springcontext/BeanDefinitionGathererTest.java
```diff
@@ -206,6 +205,39 @@ static Stream<Arguments> dependencyCollectionArguments() {
     );
   }
 
+  // ---- @Qualifier handling --------------------------------------------------
+
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("qualifiedDependencyArguments")
+  void qualifier_is_captured_on_qualified_dependency(String filePath, String expectedBeanName) {
```
asya-vorobeva, line 212: To illustrate properly how it will work we should create two beans of the same type (and different names) and specify one of them via Qualifier. An example:
```
public interface PaymentProcessor {
    void process();
}

@Component("creditCard") // Bean Name: "creditCard"
public class CreditCardProcessor implements PaymentProcessor {
    public void process() { System.out.println("Processing Credit Card..."); }
}

@Component("paypal") // Bean Name: "paypal"
public class PayPalProcessor implements PaymentProcessor {
    public void process() { System.out.println("Processing PayPal..."); }
}
```

```
@Service
public class OrderService {

    private final PaymentProcessor paymentProcessor;

    // @Qualifier explicitly picks the "paypal" bean instead of "creditCard"
    @Autowired
    public OrderService(@Qualifier("paypal") PaymentProcessor paymentProcessor) {
        this.paymentProcessor = paymentProcessor;
    }

    public void checkout() {
        paymentProcessor.process(); // Outputs: Processing PayPal...
    }
}
```

asya-vorobeva (changes requested):

java-frontend/src/test/java/org/sonar/java/model/springcontext/BeanDefinitionGathererTest.java
NoemieBenard: Maybe I understand this incorrectly, but from what I have implemented, `BeanDefinitionGatherer` only looks at the injection point and resolves the name of the dependent bean, either normally if there is no `@Qualifier` annotation (here it would map the fqn of the type `PaymentProcessor` to the name `paymentProcessor`) or with the qualifier value if there is one (in this case it would map the fqn of the type `PaymentProcessor` to the name `paypal`). Having multiple beans of this type or not doesn't change anything here since we are only looking at the injection point. Then I would expect that it is the `TypeToBeanNamesGatherer` responsability to collect beans names by type so that we can check if the expected bean in bean dependencies resolves to a single candidate or multiple ones. So I don't think this check adds value here, it will be useful when we implement the rule. WDYT?

asya-vorobeva: Generally speaking you're right. I prefer my examples because they're more realistic. In reality there's no sense to use Qualifier if we don't have at least two existing beans of the same type, and your examples look confusing.

asya-vorobeva (approved):
💯

PR #5937, opened 2026-08-18
SONARJAVA-6759: Avoid overlap between S1244 and S9147 for NaN comparisons

romainbrenguier:
## Summary
- S1244 (`FloatEqualityCheck`) no longer raises issues when a float/double is compared to `Double.NaN` or `Float.NaN`, since this case is already covered by S9147 (`NanEqualityCheck`)
- Added S1244 as a related rule in the S9147 Rspec HTML documentation
  See rspec PR: https://github.com/SonarSource/rspec/pull/7904
- Added test cases in `FloatEquality.java` to verify NaN comparisons are compliant for S1244

## Test plan
- [x] `FloatEqualityCheckTest` passes (both with and without semantic analysis)
- [x] `NanEqualityCheckTest` passes (unchanged)
- [ ] CI integration tests pass

### DIFF
--- a/java-frontend/src/test/java/org/sonar/java/model/ExpressionUtilsTest.java
+++ b/java-frontend/src/test/java/org/sonar/java/model/ExpressionUtilsTest.java
@@ -41,11 +41,17 @@ import org.sonar.plugins.java.api.tree.StaticInitializerTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.VariableTree;
 
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+
 import static java.lang.reflect.Modifier.isFinal;
 import static java.lang.reflect.Modifier.isPrivate;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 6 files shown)


java-frontend/src/test/java/org/sonar/java/model/ExpressionUtilsTest.java
```diff
@@ -470,6 +474,130 @@ void areVariablesSame_unknown_symbol() {
...
+
+    IdentifierTree identifier = mock(IdentifierTree.class);
+    when(identifier.name()).thenReturn("NaN");
+    when(identifier.symbol()).thenReturn(nanSymbol);
+    when(identifier.is(Tree.Kind.MEMBER_SELECT)).thenReturn(false);
+    when(identifier.is(Tree.Kind.IDENTIFIER)).thenReturn(true);
+
+    assertThat(ExpressionUtils.getNanOwnerTypeName(identifier)).isNull();
+  }
+
```
rombirli, line 600: I don't really see the point of these 2 tests

rombirli (approved):

java-frontend/src/test/java/org/sonar/java/model/ExpressionUtilsTest.java
romainbrenguier: They are here so that we can get 90% coverage on the PR

PR #5938, opened 2026-08-18
SONARJAVA-6769 Implement new rule S9342: Archive entries should not be empty

romainbrenguier:
Detect empty archive entries where closeEntry() is called on a ZipOutputStream or JarOutputStream after putNextEntry() without any intervening write() call, which creates useless empty entries in the archive.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
@@ -174,11 +174,8 @@ public class EmptyArchiveEntryCheck extends IssuableSubscriptionVisitor {
     ExpressionTree methodSelect = mit.methodSelect();
     if (methodSelect.is(Tree.Kind.MEMBER_SELECT)) {
       ExpressionTree expression = ((MemberSelectExpressionTree) methodSelect).expression();
-      if (expression.is(Tree.Kind.IDENTIFIER)) {
-        Symbol symbol = ((IdentifierTree) expression).symbol();
-        if (!symbol.isUnknown()) {
-          return symbol;
-        }
+      if (expression.is(Tree.Kind.IDENTIFIER) && !((IdentifierTree) expression).symbol().isUnknown()) {
+        return ((IdentifierTree) expression).symbol();
       }
     }
     return null;
--- a/java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
@@ -1,0 +1,169 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.checks;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.CheckForNull;
+import org.sonar.check.Rule;
+import org.sonar.java.model.ExpressionUtils;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
+import org.sonar.plugins.java.api.semantic.Symbol;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.BlockTree;
+import org.sonar.plugins.java.api.tree.ExpressionStatementTree;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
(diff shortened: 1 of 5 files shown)


java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
```diff
@@ -0,0 +1,209 @@
...
+        handleMethodInvocation(mit, pendingEntries);
+      } else {
+        scanForTrackedSymbolUsage(statement, pendingEntries);
+      }
+    }
+  }
+
+  private void handleMethodInvocation(MethodInvocationTree mit, Map<Symbol, MethodInvocationTree> pendingEntries) {
+    Symbol receiver = getReceiverSymbol(mit);
+    if (receiver == null) {
```
nathsou, line 88: Argument escape handling only runs when the invocation has no receiver. Consequently, `this.writeContent(stream)` (or `helper.writeContent(stream)`) leaves the entry pending and `closeEntry()` is reported even though the helper writes content. Apply the conservative argument clearing for other non-rule invocations too, and add a qualified-helper regression case.

java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
```diff
@@ -0,0 +1,209 @@
...
+    if (statement.is(Tree.Kind.EXPRESSION_STATEMENT)) {
+      ExpressionTree expression = ((ExpressionStatementTree) statement).expression();
+      if (expression.is(Tree.Kind.METHOD_INVOCATION)) {
+        return (MethodInvocationTree) expression;
+      }
+    }
+    return null;
+  }
+
+  private static boolean isDirectoryEntry(MethodInvocationTree putNextEntry) {
```
nathsou, line 131: The directory exemption only recognizes an inline `new ZipEntry("dir/")`. A common equivalent—constructing `ZipEntry directory = new ZipEntry("dir/")` first and passing `directory` to `putNextEntry()`—is reported as an empty file. Either recognize the entry value through the local symbol or conservatively avoid reporting when directory status cannot be established; add this form to the sample.

java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
```diff
@@ -0,0 +1,209 @@
...
+      if (!isDirectoryEntry(mit)) {
+        pendingEntries.put(receiver, mit);
+      }
+    } else if (CLOSE_ENTRY.matches(mit)) {
+      MethodInvocationTree putNextEntry = pendingEntries.remove(receiver);
+      if (putNextEntry != null) {
+        reportIssue(ExpressionUtils.methodName(mit), "Write content to this archive entry; it is empty.",
+          Collections.singletonList(new JavaFileScannerContext.Location("Entry opened here", ExpressionUtils.methodName(putNextEntry))), null);
+      }
+    } else if (WRITE.matches(mit)) {
```
nathsou, line 102: Pending entries are keyed only by the `ZipOutputStream` symbol. If an `OutputStream writer = new BufferedOutputStream(stream)` is created before `putNextEntry()`, `writer.write(...)` removes `writer` rather than `stream`, so a subsequent `stream.closeEntry()` is falsely reported. Track aliases/wrappers or conservatively clear the wrapped stream’s state, including for wrappers created before an entry opens.

nathsou (changes requested):
I found three verified false-positive paths in S9342. The focused submitted test passes (`mvn -pl java-checks -Dtest=EmptyArchiveEntryCheckTest test`), but each case below produces an unexpected issue when analyzed with the new check.

java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+    if (PUT_NEXT_ENTRY.matches(mit)) {
+      if (!isDirectoryEntry(mit)) {
+        pendingEntries.put(receiver, mit);
+      }
+    } else if (CLOSE_ENTRY.matches(mit)) {
+      MethodInvocationTree putNextEntry = pendingEntries.remove(receiver);
+      if (putNextEntry != null) {
+        reportIssue(ExpressionUtils.methodName(mit), "Write content to this archive entry; it is empty.",
+          Collections.singletonList(new JavaFileScannerContext.Location("Entry opened here", ExpressionUtils.methodName(putNextEntry))), null);
```
nathsou, line 100: putNextEntry/closeEntry correlation is scoped to a single BLOCK's immediate statement list, so the common try/finally idiom (`putNextEntry()` in the outer block, `closeEntry()` inside a nested `finally`/`if` block) is never analyzed, silently missing genuinely empty archive entries.

Example: `zos.putNextEntry(new ZipEntry("empty.txt")); try { } finally { zos.closeEntry(); }` creates a truly empty entry, but the rule never reports it — the nested block gets its own fresh, empty `pendingEntries` map in `visitNode`, and `TrackedSymbolVisitor` never treats `CLOSE_ENTRY` as clearing/using a tracked symbol.

java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
```diff
@@ -0,0 +1,232 @@
...
+          if (pendingEntries.containsKey(argSymbol)) {
+            usedSymbols.add(argSymbol);
+          }
+        }
+      }
+      super.visitMethodInvocation(mit);
+    }
+
+    @Override
+    public void visitNewClass(NewClassTree tree) {
```
nathsou, line 211: TrackedSymbolVisitor.visitNewClass clears a pending entry as soon as the tracked stream is passed to any constructor, even if the resulting wrapper's write method is never actually called.

Example: `zos.putNextEntry(new ZipEntry("file.txt")); OutputStream w = new BufferedOutputStream(zos); zos.closeEntry();` (wrapper constructed but `w.write(...)` never invoked) produces a genuinely empty entry, yet this removes `zos` from `pendingEntries` merely because it appears as a constructor argument — a false negative for the exact bug the rule targets.

nathsou (approved):
Approving — found two false-negative gaps worth tracking, not blocking:

PR #5939, opened 2026-08-18
SONARJAVA-6768 Implement new rule S7438: Incompatible bit masks should not be used in comparisons

romainbrenguier:
Detect incompatible bit masks in equality comparisons where bitwise AND or OR operations are compared against values that are impossible given the mask, making the comparison always true or always false.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/IncompatibleBitMaskCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/IncompatibleBitMaskCheck.java
@@ -22,6 +22,7 @@ import org.sonar.check.Rule;
 import org.sonar.java.model.ExpressionUtils;
 import org.sonar.java.model.LiteralUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.Type;
 import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.Tree;
@@ -55,6 +56,10 @@ public class IncompatibleBitMaskCheck extends IssuableSubscriptionVisitor {
     if (mask == null || value == null) {
       return;
     }
+    if (isIntOperation(bitwiseOp, possibleConstant)) {
+      mask = (long) mask.intValue();
+      value = (long) value.intValue();
+    }
     if (isIncompatible(unwrapped.kind(), mask, value)) {
       String message = comparison.is(Kind.EQUAL_TO)
         ? "This comparison is always false."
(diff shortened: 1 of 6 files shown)


java-checks/src/main/java/org/sonar/java/checks/IncompatibleBitMaskCheck.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+    BinaryExpressionTree bitwiseOp = (BinaryExpressionTree) unwrapped;
+    Long mask = extractMask(bitwiseOp);
+    Long value = signExtendedLongValue(possibleConstant);
+    if (mask == null || value == null) {
+      return;
+    }
+    if (isIntOperation(bitwiseOp, possibleConstant)) {
+      mask = (long) mask.intValue();
+      value = (long) value.intValue();
```
nathsou, line 68: **[P1] Do not truncate a `long` comparison operand**

An `int` bitwise result is promoted to `long` for comparison; Java does not truncate the RHS. For `if ((x & 0xFFFFFFFF) == 0xFFFFFFFFL)`, the condition is always false, but both values become `-1` here and no issue is raised.

`isIntOperation()` returns `true` from the known-type branch whenever `bitwiseOp.symbolType()` is `int`, without checking whether the comparison operand is itself a `long` (unlike the heuristic fallback branch, which does call `hasLongLiteral(comparisonValue)`). Please preserve the comparison operand's width / account for the sign-extended `int` result in the known-type branches too.

java-checks/src/main/java/org/sonar/java/checks/IncompatibleBitMaskCheck.java
```diff
@@ -0,0 +1,123 @@
...
+      String message = comparison.is(Kind.EQUAL_TO)
+        ? "This comparison is always false."
+        : "This comparison is always true.";
+      reportIssue(comparison.operatorToken(), message);
+    }
+  }
+
+  @Nullable
+  private static Long signExtendedLongValue(ExpressionTree operand) {
+    Long value = LiteralUtils.longLiteralValue(operand);
```
nathsou: **[P2] Unwrap parentheses before parsing literals**

`LiteralUtils.longLiteralValue` only unwraps `UNARY_MINUS`/`UNARY_PLUS` before requiring `INT_LITERAL`/`LONG_LITERAL`; it never calls `ExpressionUtils.skipParentheses`. So `if ((x & (1)) == 2)` and `if ((x & 1) == (2))` are silently skipped despite being always false. `isIntLiteral`/`hasLongLiteral` elsewhere in this file already call `skipParentheses`, so this looks like an oversight rather than intentional.

nathsou: **[P2] Support unsigned high-bit `long` masks**

`Long.decode` throws `NumberFormatException` for hex literals above `Long.MAX_VALUE`, such as `0xFFFF_FFFF_FFFF_FFFEL` (this is even called out in a comment in `LiteralUtils.java`). Consequently `if ((flags & 0xFFFF_FFFF_FFFF_FFFEL) == -1L)` is missed, since `longLiteralValue` returns `null` and the check bails out via the `mask == null || value == null` guard.

nathsou (changes requested):
Nice rule idea! Found 3 issues in `IncompatibleBitMaskCheck`

nathsou (approved):

PR #5941, opened 2026-08-18
SONARJAVA-6779 Implement new rule S9343: Methods should not increase accessibility when overriding or hiding

romainbrenguier:
Add rule S9343 which detects methods that increase accessibility when overriding instance methods or hiding static methods in parent classes.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/MethodOverrideAccessibilityCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/MethodOverrideAccessibilityCheck.java
@@ -19,6 +19,8 @@ package org.sonar.java.checks;
 import java.util.Collections;
 import java.util.List;
 import org.sonar.check.Rule;
+import org.sonar.java.checks.helpers.MethodTreeUtils;
+import org.sonar.java.model.JUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
 import org.sonar.plugins.java.api.semantic.Symbol;
@@ -64,7 +66,7 @@ public class MethodOverrideAccessibilityCheck extends IssuableSubscriptionVisito
         if (symbol.isMethodSymbol() && symbol.isStatic() && !symbol.isPrivate()
           && !(symbol.isPackageVisibility() && !samePackage(methodSymbol, symbol))) {
           Symbol.MethodSymbol candidate = (Symbol.MethodSymbol) symbol;
-          if (hasSameParameterTypes(methodSymbol, candidate)) {
+          if (MethodTreeUtils.hasSameParameterTypes(methodSymbol, candidate)) {
             reportIfAccessIncreased(methodTree, methodSymbol, candidate, "hiding");
             return;
           }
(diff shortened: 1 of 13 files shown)


java-checks/src/main/java/org/sonar/java/checks/MethodOverrideAccessibilityCheck.java
```diff
@@ -0,0 +1,106 @@
...
+    Symbol.MethodSymbol methodSymbol = methodTree.symbol();
+    if (methodSymbol.isStatic()) {
+      checkStaticMethodHiding(methodTree, methodSymbol);
+    } else {
+      checkInstanceMethodOverride(methodTree, methodSymbol);
+    }
+  }
+
+  private void checkInstanceMethodOverride(MethodTree methodTree, Symbol.MethodSymbol methodSymbol) {
+    List<Symbol.MethodSymbol> overriddenSymbols = methodSymbol.overriddenSymbols();
```
nathsou, line 53: [P1] Exclude `java.lang.Object#clone()` from this rule. `Object.clone()` is protected, while a public override is the conventional usable clone API and is the compliant style documented by S2157. This lookup selects the non-interface `Object#clone()` and reports a protected-to-public increase; the new ruling baseline already contains that false positive at `BeanMap.java:209`.

java-checks/src/main/java/org/sonar/java/checks/MethodOverrideAccessibilityCheck.java
```diff
@@ -0,0 +1,106 @@
...
+    overriddenSymbols.stream()
+      .filter(s -> !s.owner().isInterface())
```
nathsou, line 55: [P1] Do not discard a public superinterface before deciding whether the access increase is actionable. In the new `AbstractDoer` / `ConcreteDoer` sample, `ConcreteDoer#doIt` must be public to implement `Doer#doIt`; keeping the protected superclass visibility does not compile. Filtering interfaces here makes the rule report an impossible-to-fix issue. Compare against the most accessible overridden declaration (or skip when an interface requires the child visibility).

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9343.json
```diff
@@ -0,0 +1,24 @@
+{
+  "title": "Methods should not increase accessibility when overriding or hiding",
+  "type": "CODE_SMELL",
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
+  },
+  "tags": [
```
nathsou, line 9: [P1] Regenerate the plugin rule assets from the S9343 RSPEC branch instead of committing hand-written copies. This directory's README declares these files generated; draft RSPEC #7894 has tags `encapsulation`, `design`, and `pitfall`, plus `MODULAR` (not `CLEAR`). The committed HTML also omits RSPEC's potential-impact section and other explanatory text. Generate from `rule/add-RSPEC-S9343` and commit the resulting JSON/HTML/profile assets.

nathsou (changes requested):
Requesting changes: these three blockers create unavoidable false positives or publish rule assets that diverge from the S9343 RSPEC. Please address them before merge.

romainbrenguier:
Given there are legitimate cases were this rule can be violated as pointed at by the review, I think it's better to not implement this rule for now.

PR #5946, opened 2026-08-18
SONARJAVA-6790: Implement S9350: equals() implementations should not compare mismatched members

nathsou:
## Summary

- Implement S9350 in SonarJava as a native check that flags `equals()` implementations comparing unmatched instance fields or getters of the enclosing type.
- Cover `==`/`!=`, `Objects.equals`, Guava `Objects.equal`, `Arrays.equals`, instance `equals`, and 0-arg getters/record accessors, including the swapped-pair exception for order-independent equality.

## Links

- Jira: https://sonarsource.atlassian.net/browse/SONARJAVA-6790
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7901

## AI disclosure

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/EqualsMismatchedMembersCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/EqualsMismatchedMembersCheck.java
@@ -154,7 +154,9 @@ public class EqualsMismatchedMembersCheck extends IssuableSubscriptionVisitor {
       }
       MemberRef leftMember = left.get();
       MemberRef rightMember = right.get();
-      if (leftMember.kind != rightMember.kind || leftMember.symbol.equals(rightMember.symbol) || leftMember.onThis == rightMember.onThis) {
+      if (leftMember.kind != rightMember.kind
+        || leftMember.displayName.equals(rightMember.displayName)
+        || leftMember.onThis == rightMember.onThis) {
         return;
       }
       MemberRef thisMember = leftMember.onThis ? leftMember : rightMember;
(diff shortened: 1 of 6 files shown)


java-checks/src/test/java/org/sonar/java/checks/EqualsMismatchedMembersCheckTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.sonar.java.checks.verifier.TestUtils.nonCompilingTestSourcesPath;
+
+class EqualsMismatchedMembersCheckTest {
+
+  @Test
+  void test() {
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/EqualsMismatchedMembersCheckSample.java"))
+      .withCheck(new EqualsMismatchedMembersCheck())
+      .verifyIssues();
```
romainbrenguier, line 32: We should have a similar test using `.withoutSemantic`

romainbrenguier (changes requested):
It needs a test without semantic, otherwise it looks ok

java-checks/src/test/java/org/sonar/java/checks/EqualsMismatchedMembersCheckTest.java
nathsou: Added `test_without_semantic()` on the compiling sample. The check still reports the same issues from intra-file symbols when no bytecode is provided, so this uses `.withoutSemantic().verifyIssues()` like `SillyEqualsCheckTest`.

java-checks/src/main/java/org/sonar/java/checks/EqualsMismatchedMembersCheck.java
```diff
@@ -0,0 +1,236 @@
...
+      if (tree.arguments().size() != 2) {
+        return false;
+      }
+      if (OBJECTS_EQUALS.matches(tree) || GUAVA_OBJECTS_EQUAL.matches(tree)) {
+        return true;
+      }
+      Symbol.MethodSymbol method = tree.methodSymbol();
+      return (method.isUnknown() || method.isStatic())
+        && "equal".equals(ExpressionUtils.methodName(tree).name());
+    }
```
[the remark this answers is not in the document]
nathsou: Tightened the name-based `equal` fallback: unresolved `equal` still matches (Guava off the classpath), but a resolved static helper is treated as an equality comparison only when it returns `boolean`. Added `NonBooleanEqualHelper` so a void `equal` is not flagged. `EqualsMismatchedMembersCheckTest` is green.

romainbrenguier (approved):

PR #5950, opened 2026-08-18
SONARJAVA-6798 Collect dependencies from implicit single-constructor injection

NoemieBenard:
----
## Summary by Gitar

- **Spring context model:**
    - Collect dependencies from implicit single-constructor injection in `BeanDefinitionGatherer`
    - Add test cases for single-constructor and multiple-constructor scenarios


asya-vorobeva (changes requested):
Theoretically speaking, we can mix single-constructor injection and @Autowired on field injection within the same Spring component, and Spring will successfully inject dependencies for both. E.g. the following configuration will successfully autowire 2 beans:

```
@Component
public class UserService {

    // Injected via Constructor
    private final UserRepository userRepository;

    // Injected via Field Injection after instantiation
    @Autowired
    private NotificationService notificationService;

    // Single constructor: Spring automatically autowires 'userRepository'
    public UserService(UserRepository userRepository) {
        this.userRepository = userRepository;
    }
}
```

asya-vorobeva (approved):
💯

PR #5952, opened 2026-08-19
SONARJAVA-6781 Implement new rule S9344: Bitwise AND operations with zero should be corrected

romainbrenguier:
Detect bitwise AND operations with literal zero (`& 0` and `&= 0`), which always produce zero regardless of the other operand, indicating a likely programming error such as a wrong constant or wrong operator.

Part of

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/BitwiseAndWithZeroCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/BitwiseAndWithZeroCheck.java
@@ -1,0 +1,60 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.checks;
+
+import java.util.Arrays;
+import java.util.List;
+import org.sonar.check.Rule;
+import org.sonar.java.model.LiteralUtils;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.tree.AssignmentExpressionTree;
+import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.Tree.Kind;
+
+@Rule(key = "S9344")
+public class BitwiseAndWithZeroCheck extends IssuableSubscriptionVisitor {
+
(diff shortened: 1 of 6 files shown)


java-checks/src/main/java/org/sonar/java/checks/BitwiseAndWithZeroCheck.java
```diff
@@ -0,0 +1,60 @@
...
+      }
+    } else {
+      AssignmentExpressionTree assignment = (AssignmentExpressionTree) tree;
+      if (isZero(assignment.expression())) {
+        reportIssue(assignment.operatorToken(), MESSAGE);
+      }
+    }
+  }
+
+  private static boolean isZero(ExpressionTree expression) {
```
nathsou, line 56: `isZero()` fails to detect a parenthesized zero literal, so `flags & (0)` is not flagged (false negative). `LiteralUtils.longLiteralValue()` only unwraps `UNARY_MINUS`/`UNARY_PLUS` before checking for `INT_LITERAL`/`LONG_LITERAL` — it never unwraps `PARENTHESIZED_EXPRESSION`. The codebase already has `ExpressionUtils.skipParentheses` used by other checks (e.g. `StringConcatToTextBlockCheck`, `MathOnFloatCheck`, `UselessMathematicalComparisonCheck`) for exactly this purpose. Also not covered by the test sample.

java-checks/src/main/java/org/sonar/java/checks/BitwiseAndWithZeroCheck.java
```diff
@@ -0,0 +1,60 @@
...
+
+  private static final String MESSAGE = "Remove this bitwise AND with zero; the result is always zero.";
+
+  @Override
+  public List<Kind> nodesToVisit() {
+    return Arrays.asList(Kind.AND, Kind.AND_ASSIGNMENT);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
```
nathsou, line 42: Issue location is inconsistent between the two branches: the `&` case reports the whole binary expression (`reportIssue(tree, MESSAGE)`), while the `&=` case reports only the operator token (`reportIssue(assignment.operatorToken(), MESSAGE)`, line 47). The sibling rule S2437 (`UnnecessaryBitOperationCheck`) uses `operatorToken()` uniformly for both binary and assignment forms — worth aligning here too.

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaAgenticWayProfileTest.java
```diff
@@ -60,7 +60,7 @@ void profile_is_registered_as_expected() {
     BuiltInQualityProfilesDefinition.BuiltInQualityProfile actualProfile = profilesPerLanguages.get("java").get("Sonar agentic AI");
     assertThat(actualProfile.isDefault()).isFalse();
     assertThat(actualProfile.rules())
-      .hasSize(465)
+      .hasSize(466)
```
nathsou: S9344 is added to the `Sonar agentic AI` profile (rule count bumped to 466), but the three immediately preceding new-rule PRs (S9147, S9350, S9351) were not added to this profile. That profile is normally populated through a documented CSV-based curation/classification step — can you confirm this addition went through that same review, or was it added ad hoc?

nathsou (changes requested):
Automated code review found a few issues worth addressing before merge.

sonar-java-plugin/src/test/java/org/sonar/plugins/java/JavaAgenticWayProfileTest.java
romainbrenguier, line 63: It's added by the rule-api tool. We will be doing a re-run of rule-api for all rules before the release anyway.

nathsou (approved):

PR #5957, opened 2026-08-19
SONARJAVA-6786 Implement new rule S9346: Integer values should not be cast to long for use as timestamps

romainbrenguier:
Detect 32-bit or smaller integer values (int, short, byte, char) passed as arguments to timestamp-consuming APIs (Date, Timestamp, Instant, Calendar), where the narrow type causes overflow or data corruption.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/IntegerToLongTimestampCastCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/IntegerToLongTimestampCastCheck.java
@@ -82,6 +82,9 @@ public class IntegerToLongTimestampCastCheck extends AbstractMethodDetection {
     if (arg.is(Tree.Kind.TYPE_CAST)) {
       arg = ((TypeCastTree) arg).expression();
     }
+    if (arg.is(Tree.Kind.INT_LITERAL)) {
+      return;
+    }
     Type type = arg.symbolType();
     if (type.isUnknown()) {
       return;
--- a/java-checks/src/main/java/org/sonar/java/checks/IntegerToLongTimestampCastCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/IntegerToLongTimestampCastCheck.java
@@ -26,6 +26,7 @@ import org.sonar.plugins.java.api.tree.MethodInvocationTree;
 import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.TypeCastTree;
+import org.sonar.plugins.java.api.tree.UnaryExpressionTree;
 
 @Rule(key = "S9346")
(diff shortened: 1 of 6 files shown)


nathsou (changes requested):
Request changes requested: preserve narrowing casts, regenerate the rule resources from the corrected RSPEC, and complete the literal exemption. Inline comments follow.

java-checks/src/main/java/org/sonar/java/checks/IntegerToLongTimestampCastCheck.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  protected void onConstructorFound(NewClassTree nct) {
+    checkArgument(nct.arguments().get(0));
+  }
+
+  private void checkArgument(ExpressionTree argument) {
+    ExpressionTree arg = ExpressionUtils.skipParentheses(argument);
+    if (arg.is(Tree.Kind.TYPE_CAST)) {
```
nathsou, line 83: [P2] Preserve narrowing casts. This unwraps every TYPE_CAST, so new Date((int) millis) inspects millis as long and reports nothing, even though the argument has been truncated to int and then implicitly widened back to long. Only unwrap casts to long; otherwise inspect the cast expression type. Please add this case to the sample.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9346.json
```diff
@@ -0,0 +1,21 @@
...
+      "RELIABILITY": "HIGH"
+    },
+    "attribute": "COMPLETE"
+  },
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "5min"
+  },
+  "tags": [],
```
nathsou: [P2] Regenerate these generated resources from a corrected RSPEC. RSPEC PR 7897 specifies the pitfall and datetime tags plus the LOGICAL attribute, while this file contains no tags and COMPLETE. The resource directory README identifies RSPEC as the source of truth. Align the RSPEC example with the accepted direct-call scope, then regenerate HTML and JSON so the rule ships with the correct metadata.

java-checks/src/main/java/org/sonar/java/checks/IntegerToLongTimestampCastCheck.java
```diff
@@ -0,0 +1,103 @@
...
+      arg = ((TypeCastTree) arg).expression();
+    }
+    if (arg.is(Tree.Kind.INT_LITERAL)) {
```
nathsou: [P3] Complete the literal exemption. new Date(0) is skipped, but equivalent literals such as new Date(-1) and new Date((long) (0)) still report because they are unary or parenthesized AST nodes. Skip parentheses after a long cast and recognize signed integer literals.

nathsou (approved):

PR #5959, opened 2026-08-19
SONARJAVA-6780: Implement S9354: Comparable.compareTo() and Comparator.compare() should not use subtraction on numerical fields

nathsou:
## Summary

- Implement S9354 as a native SonarJava check that flags `int`/`long` subtraction inside `Comparable.compareTo()`, `Comparator.compare()`, and Comparator lambdas.
- Report a parameterized message on the `-` operator, for example `Subtracting numeric values in compareTo can overflow; use Long.compare instead.`
- Leave floating-point subtraction to S9148. Byte/short/char subtraction is ignored because the promoted `int` difference cannot overflow.

## Links

- Jira: https://sonarsource.atlassian.net/browse/SONARJAVA-6780
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7915

## AI disclosure


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9354.json
```diff
@@ -0,0 +1,23 @@
...
+    "pitfall"
+  ],
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-9354",
+  "sqKey": "S9354",
+  "scope": "All",
+  "quickfix": "targeted",
+  "code": {
+    "impacts": {
+      "RELIABILITY": "HIGH"
```
romainbrenguier: This doesn't seem coherent with the "defaultSeverity". I think it should be "MEDIUM" here. (To be fixed on the rspec PR).

sonar-java-plugin/src/test/java/org/sonar/plugins/java/GeneratedCheckListTest.java
```diff
@@ -50,7 +50,8 @@ class GeneratedCheckListTest {
     "AbstractXPathBasedCheck.java",
     "AbstractWebXmlXPathBasedCheck.java",
     "AbstractRedosCheck.java",
-    "AbstractRegexCheck.java");
+    "AbstractRegexCheck.java",
+    "AbstractComparisonMethodCheck.java");
```
romainbrenguier: Why is it blacklisted? I don't understand what this test is.

java-checks/src/main/java/org/sonar/java/checks/AbstractComparisonMethodCheck.java
```diff
@@ -0,0 +1,84 @@
...
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.LambdaExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+/**
+ * Shared entry point for checks that inspect {@code Comparable.compareTo} and {@code Comparator.compare}.
+ */
+abstract class AbstractComparisonMethodCheck extends IssuableSubscriptionVisitor {
```
romainbrenguier: I would stay away from creating abstract classes for checks. That creates coupling between rules, which we always want to avoid. If we need to avoid duplication, utility methods may be a solution.

romainbrenguier (changes requested):
I don't think adding an abstract check class is a good idea.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9354.json
nathsou: Fixed. `defaultSeverity` Major maps to a MEDIUM impact (same as S9350); HIGH is the Critical mapping used by S9148.

RSPEC: https://github.com/SonarSource/rspec/pull/7915 (`37ad141ec9`)
Analyzer metadata updated to match in `66c7c48b2d`.

sonar-java-plugin/src/test/java/org/sonar/plugins/java/GeneratedCheckListTest.java
nathsou: `count()` asserts that every on-disk `*Check.java` is a registered rule in `GeneratedCheckList`. Abstract bases without `@Rule` have to be excluded (same list as `AbstractRegexCheck`).

That blacklist entry only existed because of `AbstractComparisonMethodCheck`. The abstract class is gone in `66c7c48b2d`, so the entry is gone too.

java-checks/src/main/java/org/sonar/java/checks/AbstractComparisonMethodCheck.java
nathsou: Agreed, sharing an abstract check coupled S9148 and S9354.

`66c7c48b2d` deletes `AbstractComparisonMethodCheck`. Each rule is a standalone `IssuableSubscriptionVisitor` again, and they only share `ComparisonMethodUtils` for the `compareTo` / `Comparator.compare` matchers.

romainbrenguier (approved):

PR #5960, opened 2026-08-19
SONARJAVA-6782: Implement S9355: use Javadoc syntax for almost-Javadoc comments

nathsou:
## Summary

- Implement S9355 in SonarJava as a native `AlmostJavadocCheck`.
- Flag `/*` comments and `// ... */` comments that contain Javadoc or HTML tags immediately before a documentable declaration.
- Add a quick fix that converts the comment to `/**`.
- Generate rule metadata from the S9355 RSPEC branch.

## Links

- Jira: https://sonarsource.atlassian.net/browse/SONARJAVA-6782
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7916

## AI disclosure

### DIFF
--- a/sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9355.html
+++ b/sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9355.html
@@ -84,7 +84,6 @@ interface Repository {
 <h3>Related rules</h3>
 <ul>
   <li>{rule:java:S1176} - Public types, methods and fields (API) should be documented with Javadoc</li>
-  <li>{rule:java:S2920} - Javadoc tags should not be used in non-Javadoc comments</li>
   <li>{rule:java:S8491} - Dangling Javadoc comments should be removed</li>
 </ul>
 
--- a/sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9355.html
+++ b/sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9355.html
@@ -1,0 +1,90 @@
+<p>A documentation comment must start with <code>/**</code>. A regular comment that contains Javadoc tags or HTML documentation tags, and that sits
+immediately above a declaration, is ignored by the Javadoc tool, so the following declaration stays undocumented.</p>
+<h2>Why is this an issue?</h2>
+<p>The Javadoc tool recognizes documentation comments that start with <code>/**</code> and that are placed immediately before a class, interface,
+constructor, method, field, or enum constant. A comment that starts with <code>/*</code> or <code>//</code> in that position is ordinary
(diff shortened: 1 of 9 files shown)


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9355.html
```diff
@@ -0,0 +1,90 @@
...
+<h2>Resources</h2>
+<h3>Documentation</h3>
+<ul>
+  <li>Java Documentation - <a href="https://docs.oracle.com/en/java/javase/21/docs/specs/javadoc/doc-comment-spec.html">Documentation Comment
+  Specification for the Standard Doclet</a></li>
+</ul>
+<h3>Related rules</h3>
+<ul>
+  <li>{rule:java:S1176} - Public types, methods and fields (API) should be documented with Javadoc</li>
+  <li>{rule:java:S2920} - Javadoc tags should not be used in non-Javadoc comments</li>
```
romainbrenguier: The distinction with S2920 is not very clear to me. But it's not in any profile anyway. Maybe we could clarify were the two apply.

romainbrenguier (approved):

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9355.html
nathsou: Good catch, and I went further than clarifying: I removed the S2920 reference entirely.

S2920 has no implementation in the Java analyzer. It is not among the rules shipped in `sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/`, so the generated `{rule:java:S2920}` reference does not resolve and would have pointed users at a rule they cannot enable anywhere. The remaining two related rules are both implemented, S1176 and S8491 (`DanglingJavadocCheck`), so those stay.

For the record, the conceptual boundary was:

- **S2920** covers Javadoc tags in *any* non-Javadoc comment, wherever it sits, including inside method bodies. Its own example is `// FIXME should be moved to {@link Foo#bar(String)}`, which documents nothing, and its remediation is "remove the tag or convert to Javadoc".
- **S9355** only fires when a comment sits immediately above a documentable declaration and looks like it was meant to be Javadoc, so the single correct fix is always to start it with `/**` (shipped as a quick fix). It also covers the HTML-only case such as `/* Returns the display name as <code>String</code>. */`, which has no `@tag` and is therefore out of S2920's scope.

The implementation keeps them apart: S9355 skips comments inside methods and constructors, and does not flag a plain `// ... {@link X}` line comment unless it ends with `*/`. But since S2920 is unimplemented, `Minor`, and in no profile, documenting that distinction to users was not worth it.

Updated in rspec (SonarSource/rspec#7916) and regenerated here.

PR #5981, opened 2026-08-21
SONARJAVA-6783: Implement rule S9345 Classes with throwing constructors should be protected against Finalizer attacks

romainbrenguier:
## Summary
- Implement new rule S9345 that detects classes with throwing constructors vulnerable to Finalizer attacks
- Non-compliant: non-final class with public/protected/package-private constructor that has a `throws` clause or contains `throw` statements
- Compliant: `final` classes, classes with only `private` throwing constructors (factory pattern), enums, records, sealed classes (where all permitted subclasses are final/sealed), classes with `final finalize()` method, and local classes

## Review feedback addressed (commit 09126f5)
All 6 findings from @nathsou have been addressed:
- **[P1]** Handle instance/field initializers and classes with no explicit constructor
- **[P1]** Abstract classes are no longer exempt (attacker can subclass)
- **[P1]** Sealed classes with non-sealed subclasses are correctly flagged
- **[P2]** Classes with `protected final void finalize() {}` recognized as safe
- **[P2]** Removed inaccurate CWE-586 mapping, kept CERT OBJ11-J reference
- **[P3]** Local classes are excluded (cannot be subclassed from outside)

## Test plan
- [x] Unit test with `CheckVerifier` covering noncompliant and compliant patterns
- [x] Test without semantic analysis
- [x] Ruling results updated
- [ ] CI passes

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
@@ -1,0 +1,92 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.checks;
+
+import java.util.Collections;
+import java.util.List;
+import org.sonar.check.Rule;
+import org.sonar.java.model.ModifiersUtils;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.BlockTree;
+import org.sonar.plugins.java.api.tree.ClassTree;
+import org.sonar.plugins.java.api.tree.LambdaExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Modifier;
+import org.sonar.plugins.java.api.tree.ThrowStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.Tree.Kind;
+
+@Rule(key = "S9345")
+public class FinalizerAttackCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Kind> nodesToVisit() {
+    return Collections.singletonList(Kind.CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+    if (ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.FINAL) ||
+      ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.ABSTRACT)) {
+      return;
+    }
+    for (Tree member : classTree.members()) {
+      if (member.is(Kind.CONSTRUCTOR) && isVulnerableConstructor((MethodTree) member)) {
+        reportIssue(classTree.simpleName(), "Make this class \"final\" or make the throwing constructors \"private\".");
+        return;
+      }
+    }
+  }
+
+  private static boolean isVulnerableConstructor(MethodTree constructor) {
+    if (ModifiersUtils.hasModifier(constructor.modifiers(), Modifier.PRIVATE)) {
+      return false;
+    }
+    return !constructor.throwsClauses().isEmpty() || containsThrowStatement(constructor);
+  }
+
+  private static boolean containsThrowStatement(MethodTree constructor) {
+    BlockTree block = constructor.block();
+    if (block == null) {
+      return false;
+    }
+    ThrowStatementVisitor visitor = new ThrowStatementVisitor();
+    block.accept(visitor);
+    return visitor.hasThrow;
+  }
+
+  private static class ThrowStatementVisitor extends BaseTreeVisitor {
+    boolean hasThrow;
+
+    @Override
+    public void visitThrowStatement(ThrowStatementTree tree) {
+      hasThrow = true;
+    }
+
+    @Override
+    public void visitClass(ClassTree tree) {
+      // skip nested classes
+    }
+
+    @Override
+    public void visitLambdaExpression(LambdaExpressionTree tree) {
+      // skip lambdas
+    }
+  }
+}
--- a/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
@@ -23,7 +23,6 @@ import org.sonar.check.Rule;
 import org.sonar.java.model.ModifiersUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
-import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.semantic.Type;
 import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.BlockTree;
--- a/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
@@ -70,10 +70,7 @@ public class FinalizerAttackCheck extends IssuableSubscriptionVisitor {
     for (Tree member : classTree.members()) {
       if (member.is(Kind.CONSTRUCTOR)) {
         hasExplicitConstructor = true;
-      } else if (member.is(Kind.INITIALIZER) && containsThrowStatementInBlock((BlockTree) member)) {
-        throwingInitializers.add(member);
-        hasThrowingInitializers = true;
-      } else if (member.is(Kind.VARIABLE) && hasThrowingFieldInitializer((VariableTree) member)) {
+      } else if (isThrowingInitializer(member)) {
         throwingInitializers.add(member);
         hasThrowingInitializers = true;
       }
@@ -108,6 +105,11 @@ public class FinalizerAttackCheck extends IssuableSubscriptionVisitor {
     reportIssue(classTree.simpleName(),
       "Make this class \"final\" or add a private constructor, because initializers can throw.",
       locations, null);
+  }
+
+  private static boolean isThrowingInitializer(Tree member) {
+    return (member.is(Kind.INITIALIZER) && containsThrowStatementInBlock((BlockTree) member))
+      || (member.is(Kind.VARIABLE) && hasThrowingFieldInitializer((VariableTree) member));
   }
 
   private static boolean isLocalClass(ClassTree classTree) {
--- a/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
@@ -176,18 +176,18 @@ public class FinalizerAttackCheck extends IssuableSubscriptionVisitor {
       if (classTree.simpleName() != null && name.equals(classTree.simpleName().name())) {
         return classTree;
       }
-      for (Tree member : classTree.members()) {
-        ClassTree found = findClassInTree(member, name);
-        if (found != null) {
-          return found;
-        }
-      }
+      return findClassInChildren(classTree.members(), name);
     } else if (tree.is(Kind.COMPILATION_UNIT)) {
-      for (Tree child : ((CompilationUnitTree) tree).types()) {
-        ClassTree found = findClassInTree(child, name);
-        if (found != null) {
-          return found;
-        }
+      return findClassInChildren(((CompilationUnitTree) tree).types(), name);
+    }
+    return null;
+  }
+
+  private static ClassTree findClassInChildren(List<? extends Tree> children, String name) {
+    for (Tree child : children) {
+      ClassTree found = findClassInTree(child, name);
+      if (found != null) {
+        return found;
       }
     }
     return null;
--- a/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
@@ -16,21 +16,28 @@
  */
 package org.sonar.java.checks;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import org.sonar.check.Rule;
 import org.sonar.java.model.ModifiersUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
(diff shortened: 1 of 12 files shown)


java-checks/src/test/java/org/sonar/java/checks/FinalizerAttackCheckTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static org.sonar.java.checks.verifier.TestUtils.mainCodeSourcesPath;
+
+class FinalizerAttackCheckTest {
+
+  @Test
+  void test() {
+    CheckVerifier.newVerifier()
+      .onFile(mainCodeSourcesPath("checks/FinalizerAttackCheckSample.java"))
+      .withCheck(new FinalizerAttackCheck())
```
romainbrenguier, line 31: We should add a test withoutSemantic

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9345.json
```diff
@@ -0,0 +1,32 @@
...
+    "cert",
+    "cwe",
+    "serialization"
+  ],
+  "defaultSeverity": "Critical",
+  "ruleSpecification": "RSPEC-9345",
+  "sqKey": "S9345",
+  "scope": "Main",
+  "defaultQualityProfiles": [
+    "Sonar way"
```
romainbrenguier, line 17: The rule should have been added to the quality profiles. rule-api needs to be re-run.

nathsou (changes requested):
Changes requested. See the inline comments for the six findings; the initializer-path issue is also tracked in the existing unresolved thread.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
```diff
@@ -0,0 +1,99 @@
...
+    if (classTree.simpleName() == null ||
+      ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.FINAL) ||
+      ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.ABSTRACT) ||
+      ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.SEALED)) {
+      return;
+    }
+    List<JavaFileScannerContext.Location> secondaryLocations = Collections.singletonList(
+      new JavaFileScannerContext.Location("Non-final class", classTree.simpleName()));
+    for (Tree member : classTree.members()) {
+      if (member.is(Kind.CONSTRUCTOR) && isVulnerableConstructor((MethodTree) member)) {
```
nathsou: [P1] This only examines explicit constructor members. A throwing instance initializer or field initializer runs during object construction too; a class with no explicit constructor is never checked. The existing unresolved thread identifies this gap, but it remains a blocking false negative.

nathsou (commented):
Inline findings for the requested changes.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
```diff
@@ -0,0 +1,99 @@
...
+  public List<Kind> nodesToVisit() {
+    return Collections.singletonList(Kind.CLASS);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    ClassTree classTree = (ClassTree) tree;
+    if (classTree.simpleName() == null ||
+      ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.FINAL) ||
+      ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.ABSTRACT) ||
```
nathsou: [P1] Do not exempt abstract classes. An attacker can instantiate a concrete subclass that overrides `finalize()`; if this accessible parent constructor throws, the partially constructed subclass instance can still be finalized and resurrected. This condition misses that attack.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
```diff
@@ -0,0 +1,99 @@
...
+      ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.SEALED)) {
```
nathsou: [P1] A sealed class can permit a `non-sealed` subclass. An attacker can extend that reopened branch, and constructing the attacker type still executes this parent constructor. Do not suppress every sealed class solely from its modifier; account for the permitted hierarchy.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
```diff
@@ -0,0 +1,99 @@
...
+      return;
+    }
+    List<JavaFileScannerContext.Location> secondaryLocations = Collections.singletonList(
+      new JavaFileScannerContext.Location("Non-final class", classTree.simpleName()));
+    for (Tree member : classTree.members()) {
+      if (member.is(Kind.CONSTRUCTOR) && isVulnerableConstructor((MethodTree) member)) {
+        MethodTree constructor = (MethodTree) member;
+        reportIssue(constructor.simpleName(),
+          "Make this class \"final\" or make this throwing constructor \"private\".",
```
nathsou: [P2] S9345's RSPEC draft names an empty cleanup implementation that cannot be overridden as a protection mechanism. A throwing constructor in a class with `protected final void finalize() {}` is safe from an attacker-supplied finalizer but is reported here. Recognize that mitigation and add a compliant test.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
```diff
@@ -0,0 +1,99 @@
...
+import org.sonar.plugins.java.api.tree.ThrowStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.Tree.Kind;
+
+@Rule(key = "S9345")
+public class FinalizerAttackCheck extends IssuableSubscriptionVisitor {
+
+  @Override
+  public List<Kind> nodesToVisit() {
+    return Collections.singletonList(Kind.CLASS);
```
nathsou, line 52: [P3] This subscribes to local `CLASS` declarations as well. A local class cannot normally be named and subclassed from an independent Java source file, so flagging its throwing constructor is a likely false positive. Please either exclude local classes or add coverage that documents the intended scope.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9345.json
```diff
@@ -0,0 +1,32 @@
...
+  ],
+  "quickfix": "unknown",
+  "code": {
+    "impacts": {
+      "SECURITY": "HIGH"
+    },
+    "attribute": "COMPLETE"
+  },
+  "securityStandards": {
+    "CWE": [586],
```
nathsou: [P2] CWE-586 specifically covers an explicit call to `finalize()` from outside the finalizer. S9345 detects a different finalizer-resurrection attack, so this mapping is inaccurate. Please align the rule's CWE mapping with the final RSPEC; CWE-459 is not an appropriate replacement because it covers incomplete cleanup of temporary/supporting resources.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: We now exclude local classes (see line 54)

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9345.json
romainbrenguier: I will remove on the rpsec PR side

nathsou (commented):
Single-pass re-review of the follow-up commits.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
```diff
@@ -0,0 +1,260 @@
...
+    block.accept(visitor);
+    return visitor.hasThrow;
+  }
+
+  private static boolean hasThrowingFieldInitializer(VariableTree variable) {
+    if (variable.initializer() == null) {
+      return false;
+    }
+    ThrowStatementVisitor visitor = new ThrowStatementVisitor();
+    variable.initializer().accept(visitor);
```
nathsou, line 275: [P1] The field-initializer part of the original finding is still not detected. A Java field initializer is an expression, so it cannot contain a bare `throw` statement; the only `throw` nodes this visitor can encounter are inside lambdas or anonymous/local classes, which the visitor deliberately skips. Consequently cases such as `private final X value = compute();` remain unreported even when `compute()` throws during construction (the new samples explicitly mark them compliant). Please analyze potentially throwing initializer invocations/constructions, or narrow the rule's documented scope instead of treating this path as fixed.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
```diff
@@ -0,0 +1,260 @@
...
+      parent = parent.parent();
+    }
+    return false;
+  }
+
+  private static boolean isSafelySealedClass(ClassTree classTree) {
+    if (!ModifiersUtils.hasModifier(classTree.modifiers(), Modifier.SEALED)) {
+      return false;
+    }
+    for (TypeTree permitted : classTree.permittedTypes()) {
```
nathsou, line 146: [P1] This only checks direct permitted subclasses. For `sealed Base permits Mid`, `sealed Mid permits Open`, and `non-sealed Open extends Mid`, `Base` is considered safe even though an attacker can extend `Open` and still execute `Base`'s throwing constructor. Walk sealed permitted descendants recursively before suppressing the parent.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
```diff
@@ -0,0 +1,260 @@
...
+      if (isNonSealedPermittedType(classTree, permitted)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private static boolean isNonSealedPermittedType(ClassTree context, TypeTree permitted) {
+    ClassTree permittedDecl = resolvePermittedDeclaration(context, permitted);
+    return permittedDecl != null && ModifiersUtils.hasModifier(permittedDecl.modifiers(), Modifier.NON_SEALED);
```
nathsou: [P1] An unresolved permitted declaration currently returns `false` here and therefore contributes to classifying the sealed parent as safe. Permitted subclasses may live in another source file or be unavailable in incomplete analysis; without their modifiers, the implementation cannot prove that the hierarchy stays closed. Treat an unresolved permitted type conservatively as unsafe rather than suppressing the finding.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
```diff
@@ -0,0 +1,260 @@
...
+    return null;
+  }
+
+  private static boolean hasFinalFinalizer(ClassTree classTree) {
+    for (Tree member : classTree.members()) {
+      if (member.is(Kind.METHOD)) {
+        MethodTree method = (MethodTree) member;
+        if ("finalize".equals(method.simpleName().name()) &&
+          method.parameters().isEmpty() &&
+          ModifiersUtils.hasModifier(method.modifiers(), Modifier.FINAL)) {
```
nathsou: [P2] The documented mitigation is a **final, empty** `finalize()` implementation, but this accepts any final zero-argument method named `finalize`. A body such as `leaked = this;` still resurrects the partially constructed object, yet the whole class is skipped. Verify that the overriding finalizer is empty before considering the class protected.

nathsou (approved):

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: Addressed: `checkMembers()` now detects throwing instance initializers and field initializers via `isThrowingInitializer`. Classes with no explicit constructor but throwing initializers are flagged. Test cases: `InitializerThrower`, `MultipleThrowingInitializers`, `AbstractWithThrowingInitializer`.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: Addressed: Abstract classes are no longer exempt. `visitNode` has no abstract-class check. Test cases: `AbstractService`, `AbstractWithThrowingInitializer`, `AbstractThrowsClause` are all flagged.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: Addressed: `isSafelySealedClass` walks the permitted hierarchy recursively. Non-sealed subclasses make the parent unsafe. Test cases: `SealedServiceWithNonSealed`/`OpenSubclass` and `DeepSealedNonSealedBottom`/`DeepSealedMid`/`DeepSealedOpen`.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: Addressed: `hasFinalFinalizer` checks for empty body via `hasEmptyBody()`. Test cases: `ProtectedByFinalizer` (compliant, empty body) and `FinalFinalizerNonEmpty` (noncompliant, non-empty body).

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: Addressed: Local classes are excluded via `isLocalClass()` in `visitNode()`. Test case: `LocalClass` inside `someMethod()` is compliant.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9345.json
romainbrenguier: Acknowledged: CWE mapping will be fixed on the RSPEC PR side, not in this PR.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: Addressed: `ThrowingExpressionVisitor` now handles `visitMethodInvocation` and `visitNewClass`, checking `canThrowCheckedException` for methods/constructors with declared thrown types. `hasThrowingFieldInitializer` uses this visitor. Semantic test cases: `FieldInitializerCheckedThrow`, `FieldInitializerNewThrows`, `FieldInitWithCheckedAndCtor`.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: Addressed: `isSafePermittedType` calls `isSafelySealedClass` recursively for sealed subclasses. Deep hierarchy test cases: `DeepSealedParent`/`DeepSealedChild`/`DeepSealedGrandchild` (compliant) and `DeepSealedNonSealedBottom`/`DeepSealedMid`/`DeepSealedOpen` (noncompliant).

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: Addressed: `isSafePermittedType` returns `false` (unsafe) when `permittedDecl == null`, conservatively treating unresolved types as unsafe. Non-compiling test cases: `SealedWithUnknown` and `SealedWithOnlyUnknown` are flagged.

java-checks/src/main/java/org/sonar/java/checks/FinalizerAttackCheck.java
romainbrenguier: Addressed: `hasFinalFinalizer` now requires `hasEmptyBody(method)`. A finalize() with any statements is not considered protective. Test case: `FinalFinalizerNonEmpty` is flagged as noncompliant.

PR #5984, opened 2026-08-21
SONARJAVA-6820: Fix S9355: ignore Eclipse (non-Javadoc) override comments

nathsou:
## Summary

- Stop raising S9355 when a comment contains the Eclipse `(non-Javadoc)` marker.
- The marker alone is enough; `@see` is not required.
- Keep flagging a regular `@see` comment that is not an Eclipse stub.
- Refresh the generated S9355 description from the RSPEC branch.

## Links

- Jira: https://sonarsource.atlassian.net/browse/SONARJAVA-6820
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7936

## AI disclosure

### DIFF
--- a/java-checks-test-sources/default/src/main/java/checks/AlmostJavadocCheckSample.java
+++ b/java-checks-test-sources/default/src/main/java/checks/AlmostJavadocCheckSample.java
@@ -121,6 +121,35 @@ class AlmostJavadocCheckSample {
     return name;
   }
 
+  /*
+   * (non-Javadoc)
+   * @see java.lang.Object#hashCode()
+   */
+  @Override
+  public int hashCode() {
+    return 0;
+  }
+
+  /* (non-Javadoc)
+   * @see java.lang.Object#equals(java.lang.Object)
+   */
+  @Override
+  public boolean equals(Object other) {
+    return this == other;
+  }
+
+  /* (non-Javadoc)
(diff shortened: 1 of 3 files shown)


java-checks-test-sources/default/src/main/java/checks/AlmostJavadocCheckSample.java
```diff
@@ -121,6 +121,49 @@ public String toString() {
...
+  /* (non-javadoc) @see java.lang.Object#wait() */
+  public void lowercaseMarkerLookalike() {
+  }
+
+  // (non-Javadoc) {@link Object} */
+  public void eclipseMarkerLineComment() {
+  }
+
+  /* (non-Javadoc) returns </code> */
+  public String eclipseMarkerWithHtml() {
```
romainbrenguier: This is a bit of a strange case. I'm not sure we should include it as a test.

romainbrenguier (approved):

java-checks-test-sources/default/src/main/java/checks/AlmostJavadocCheckSample.java
nathsou: Agreed — that sample mixed the Eclipse marker with an HTML tag in a way that does not show up in real code. I dropped it.

PR #5989, opened 2026-08-21
SONARJAVA-6823: Implemented rule S909 "continue" should not be used

romainbrenguier:
## Summary
- Implemented rule S909 that flags every use of the `continue` statement inside loops (for, for-each, while, do-while), whether labeled or unlabeled
- The rule is a CODE_SMELL with MEDIUM impact on MAINTAINABILITY — loops should use inverted conditional logic instead of `continue`
- Simple `IssuableSubscriptionVisitor` subscribing to `CONTINUE_STATEMENT`

## Test plan
- [x] Unit test `ContinueInLoopCheckTest` passes
- [ ] CI passes (ruling tests may need baseline updates)
- [ ] Verify no false positives on ruling projects


java-checks/src/test/java/org/sonar/java/checks/ContinueInLoopCheckTest.java
```diff
@@ -21,23 +21,14 @@
...
 
   @Test
   void test() {
     CheckVerifier.newVerifier()
-      .onFile(mainCodeSourcesPath("checks/IntegerSubtractionInComparisonCheckSample.java"))
-      .withCheck(new IntegerSubtractionInComparisonCheck())
+      .onFile(mainCodeSourcesPath("checks/ContinueInLoopCheckSample.java"))
+      .withCheck(new ContinueInLoopCheck())
       .verifyIssues();
   }
```
romainbrenguier, line 32: We should also have a test `withoutSemantic`

nathsou (changes requested):
This is a duplicate of `S909` which is already implemented in `C`, `JS` and `PLSQL`.
Please reuse this rule id instead of `S9356`.

Apart from that, LGTM!

nathsou (approved):

PR #5993, opened 2026-08-21
SONARJAVA-6828 Add AGENTS.md to provide guidance to agents making changes to rules

romainbrenguier:
This is to help being more efficient when creating new rules.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,6 +4,8 @@ SonarJava is a static code analyzer for Java.
 
 The rules are implemented in **Java** using the SonarJava API. Test resources are **Java** files annotated with special comment markers.
 
+For step-by-step guidance on implementing a new rule (including metadata generation), see `.claude/skills/new-rule/SKILL.md`.
+
 ---
 
 
@@ -94,7 +96,7 @@ public class MyCheck extends BaseTreeVisitor implements JavaFileScanner {
 - **`reportIssue(tree, message)`** — Report an issue at a specific AST node
 - **`reportIssue(tree, message, secondaries, cost)`** — Report with secondary locations
 - **`Tree.Kind.*`** — AST node type enumeration (e.g., `Tree.Kind.METHOD`, `Tree.Kind.IF_STATEMENT`, `Tree.Kind.METHOD_INVOCATION`)
-- **`TreeUtils.firstAncestorOfKind(tree, Tree.Kind.METHOD)`** — Navigate up the AST
+- **`ExpressionUtils.getParentOfType(tree, Tree.Kind.METHOD)`** — Navigate up the AST
 - **`ExpressionUtils`**, **`MethodMatchers`** — Utility classes for common analysis patterns
 
 ### MethodMatchers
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -0,0 +1,255 @@
...
+  }
+}
+```
+
+### Key API Elements
+
+- **`reportIssue(tree, message)`** — Report an issue at a specific AST node
+- **`reportIssue(tree, message, secondaries, cost)`** — Report with secondary locations
+- **`Tree.Kind.*`** — AST node type enumeration (e.g., `Tree.Kind.METHOD`, `Tree.Kind.IF_STATEMENT`, `Tree.Kind.METHOD_INVOCATION`)
+- **`TreeUtils.firstAncestorOfKind(tree, Tree.Kind.METHOD)`** — Navigate up the AST
```
nathsou: `TreeUtils.firstAncestorOfKind` does not exist in this repository or its compiled API, so following this guidance produces a non-compiling rule. Please replace this with an actual parent-walking pattern, such as `ExpressionUtils.getParentOfType(tree, Tree.Kind.METHOD)`, or remove the example.

AGENTS.md
```diff
@@ -0,0 +1,255 @@
...
+reportIssue(tree, "Message").withQuickFix(() -> quickFix);
+```
+
+### Dependency-Aware Rules
+
+When a rule should only apply when a specific library is present, or when its behaviour depends on a dependency version, implement `DependencyVersionAware` alongside the visitor base class:
+
+```java
+@Rule(key = "S2230")
+public class TransactionalMethodVisibilityCheck extends IssuableSubscriptionVisitor implements DependencyVersionAware {
```
nathsou, line 134: This `DependencyVersionAware` example does not compile: `SubscriptionVisitor.nodesToVisit()` is abstract and `IssuableSubscriptionVisitor` does not provide a default implementation. Please add the required override, mirroring `TransactionalMethodVisibilityCheck`:
```java
@Override
  return Collections.singletonList(Tree.Kind.METHOD);
}
```

AGENTS.md
```diff
@@ -0,0 +1,255 @@
...
+- Some handle Java 11+ or Java 17+ features (var, records, sealed classes)
+
+---
+
+# Test File Conventions
+
+The file general structure for a rule implementation and its tests is as follows:
+- Rule class: `java-checks/src/main/java/org/sonar/java/checks/{RuleId}Check.java`
+- Test class: `java-checks/src/test/java/org/sonar/java/checks/{RuleId}CheckTest.java`
+- Test samples: `java-checks-test-sources/default/src/main/files/checks/{RuleId}CheckSample.java`
```
nathsou: The normal-sample path is invalid. `java-checks-test-sources/default/src/main/files/` contains only `non-compiling`; normal samples live under `src/main/java/checks/` and should be loaded with `TestUtils.mainCodeSourcesPath("checks/{RuleId}CheckSample.java")`. This also contradicts the legacy `src/test/files` example below. Please update both the documented convention and the example. The same incorrect normal-sample path is duplicated in `.claude/skills/new-rule/SKILL.md:75`.

AGENTS.md
```diff
@@ -0,0 +1,255 @@
+# Introduction
```
nathsou, line 1: Non-blocking: since `CLAUDE.md` symlinks to this auto-loaded guide, please add a pointer to `.claude/skills/new-rule/SKILL.md`. That skill contains required new-rule guidance not present here, including rule-api metadata generation, output paths, and the prohibition on adding external test dependencies. Without a link, an agent following `AGENTS.md` alone can implement a rule without its required generated metadata.

nathsou (changes requested):
I found three correctness issues in the agent guidance that can directly lead to non-compiling rules or failing tests. I also left a non-blocking suggestion to point the auto-loaded guide at the existing new-rule skill, which contains required metadata-generation and test-dependency guidance absent here.

nathsou (approved):

PR #5995, opened 2026-08-21
SONARJAVA-6827: Implement S9357 Anonymous classes on functional interfaces should be lambdas

romainbrenguier:
## Summary

Implement S9357: Anonymous classes on functional interfaces should be lambdas.

- S9357 scope is set to **Tests** to complement S1604 (scope: Main) without double-reporting on main code
- Shared detection logic extracted into `AbstractAnonymousClassToLambdaCheck` base class, used by both S9357 and S1604, eliminating code duplication
- Test samples moved to test source path to match the Tests scope

## Test plan

- [x] S9357 unit tests pass (with and without semantic)
- [x] S1604 unit tests pass (pre-existing failure on `Function<Object, Date>` test case unrelated to this change)
- [x] CheckList generation correctly places S9357 in `JAVA_TEST_CHECKS` and S1604 in `JAVA_MAIN_CHECKS`
- [ ] CI ruling tests (will need a new ruling fix PR after scope change)


java-checks/src/main/java/org/sonar/java/checks/AbstractAnonymousClassToLambdaCheck.java
```diff
@@ -0,0 +1,197 @@
...
+import org.sonar.plugins.java.api.tree.EnumConstantTree;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.NewClassTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.TypeTree;
+
+public abstract class AbstractAnonymousClassToLambdaCheck extends BaseTreeVisitor implements JavaFileScanner, JavaVersionAwareVisitor {
```
romainbrenguier: Don't declare any abstract class. This introduce coupling between rules. To avoid duplicating code, you may use utility classes.

java-checks/src/main/java/org/sonar/java/checks/AnonymousClassOnFunctionalInterfaceCheck.java
```diff
@@ -0,0 +1,39 @@
...
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.JavaVersion;
+import org.sonar.plugins.java.api.JavaVersionAwareVisitor;
+
+@Rule(key = "S9357")
+public class AnonymousClassOnFunctionalInterfaceCheck implements JavaFileScanner, JavaVersionAwareVisitor {
+
+  @Override
+  public boolean isCompatibleWithJavaVersion(JavaVersion version) {
+    return version.isJava8Compatible();
```
nathsou, line 31: [P2] Exclude S9357 from AutoScan\n\nAutoScan generally lacks `sonar.java.source`, so an unspecified version is treated as Java 8-compatible. S9357 may therefore recommend lambdas in Java 7 test code, where lambdas are unsupported. S1604 is already excluded from AutoScan for this exact reason. Add `AnonymousClassOnFunctionalInterfaceCheck.class` alongside it in `JAVA_CHECKS_NOT_WORKING_FOR_AUTOSCAN`.

nathsou (approved):

PR #6001, opened 2026-08-24
SONARJAVA-6824: Implemented rule S9358 - Conditional expressions should not duplicate operations in both branches

romainbrenguier:
This PR implements rule S9358 which detects when a ternary operator applies the same operation (method invocation, object creation, or array access) to different arguments in both branches.

Such patterns can be refactored by moving the condition inside the operation for better readability:

The rule handles:
- Method invocations with same method name and different arguments
- Object creation ( vs )
- Array access ( vs )

Test cases cover both compliant and non-compliant scenarios including edge cases like nested ternaries, chained method calls, and generic type arguments.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/TernaryOperatorSameOperationCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/TernaryOperatorSameOperationCheck.java
@@ -21,15 +21,13 @@ import org.sonar.check.Rule;
 import org.sonar.java.model.ExpressionUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.tree.ArrayAccessExpressionTree;
-import org.sonar.plugins.java.api.tree.ArrayDimensionTree;
+import org.sonar.plugins.java.api.tree.ConditionalExpressionTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.IdentifierTree;
 import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodInvocationTree;
 import org.sonar.plugins.java.api.tree.NewClassTree;
 import org.sonar.plugins.java.api.tree.Tree;
-import org.sonar.plugins.java.api.tree.TypeArguments;
-import org.sonar.plugins.java.api.tree.TypeTree;
 
 @Rule(key = "S9358")
 public class TernaryOperatorSameOperationCheck extends IssuableSubscriptionVisitor {
@@ -41,7 +39,7 @@ public class TernaryOperatorSameOperationCheck extends IssuableSubscriptionVisit
 
   @Override
   public void visitNode(Tree tree) {
-    var conditional = (org.sonar.plugins.java.api.tree.ConditionalExpressionTree) tree;
+    var conditional = (ConditionalExpressionTree) tree;
     var trueExpr = ExpressionUtils.skipParentheses(conditional.trueExpression());
     var falseExpr = ExpressionUtils.skipParentheses(conditional.falseExpression());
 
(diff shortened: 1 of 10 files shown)


java-checks/src/main/java/org/sonar/java/checks/TernaryOperatorSameOperationCheck.java
```diff
@@ -0,0 +1,120 @@
...
+      return sameNewClass((NewClassTree) left, (NewClassTree) right);
+    }
+    if (left.is(Tree.Kind.ARRAY_ACCESS_EXPRESSION) && right.is(Tree.Kind.ARRAY_ACCESS_EXPRESSION)) {
+      return sameArrayAccess((ArrayAccessExpressionTree) left, (ArrayAccessExpressionTree) right);
+    }
+    return false;
+  }
+
+  private static boolean sameMethodInvocation(MethodInvocationTree left, MethodInvocationTree right) {
+    return sameMethodSelect(left.methodSelect(), right.methodSelect())
```
nathsou: Method names do not identify the invoked operation. For example, `condition ? overloaded(1) : overloaded("x")` invokes distinct overloads but is reported, while `condition ? foo(a) : this.foo(b)` invokes the same method but is missed. The suggested refactoring for the overload case may not compile. Please compare resolved method symbols when semantic information is available, and avoid reporting ambiguous syntax-only matches.

java-checks/src/main/java/org/sonar/java/checks/TernaryOperatorSameOperationCheck.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+    return sameTree(left, right);
+  }
+
+  private static boolean sameIdentifier(IdentifierTree left, IdentifierTree right) {
+    return left.name().equals(right.name());
+  }
+
+  private static boolean sameNewClass(NewClassTree left, NewClassTree right) {
+    if (!sameTree(left.identifier(), right.identifier())) {
```
nathsou, line 104: This comparison ignores constructor identity and the rest of the creation context. `condition ? new Overloaded(1) : new Overloaded("x")` is reported even when those arguments select different constructors; enclosing instances and anonymous class bodies are ignored as well. Please compare resolved constructor symbols and relevant `NewClassTree` context before treating these branches as the same operation.

nathsou (approved):
Approved with 2 findings I think are worth addressing:

java-checks/src/main/java/org/sonar/java/checks/TernaryOperatorSameOperationCheck.java
romainbrenguier: Addressed in commit 0e2ef4b: the check now compares resolved method symbols via `sameMethodSymbol`. When semantic information is available, overloaded methods with different signatures (e.g. `overloaded(int)` vs `overloaded(String)`) are correctly identified as different operations and not reported. When symbols are unknown (no semantics), the check falls back to treating them as the same method to avoid false negatives. Additionally, `sameTree` now uses `SyntacticEquivalence.areEquivalent` for proper structural comparison, and test cases have been added for these scenarios.

java-checks/src/main/java/org/sonar/java/checks/TernaryOperatorSameOperationCheck.java
romainbrenguier: Addressed in commit 0e2ef4b: `sameNewClass` now compares resolved constructor symbols via `sameMethodSymbol`, checks for anonymous class bodies (returns false if either has a class body), and compares enclosing expressions via `sameNullableTree`. Test cases have been added for overloaded constructors, anonymous class bodies, and qualified instantiations (`obj.new Inner(a)` vs `obj.new Inner(b)`).

romainbrenguier: Addressed in commit 0e2ef4b: `sameNewClass` now compares resolved constructor symbols via `sameMethodSymbol`, checks for anonymous class bodies (returns false if either has a class body), and compares enclosing expressions via `sameNullableTree`. Test cases have been added for overloaded constructors, anonymous class bodies, and qualified instantiations (`obj.new Inner(a)` vs `obj.new Inner(b)`).

PR #6002, opened 2026-08-24
SONARJAVA-6838: Fix FP on S9354: ignore subtraction of operands with a provably bounded range

nathsou:
## Summary

A Peach false-positive run sampled 366 S9354 issues and found 60 false positives (16.4%). About 45 of them subtract two operands whose value range is provably bounded, so the subtraction cannot overflow: array/String lengths, `Collection`/`Map.size()`, `Enum.ordinal()`, `Integer`/`Long.bitCount()` and related bit-scan intrinsics, values masked with a non-negative bitwise `AND` constant, and compile-time constants. The check previously decided purely from operand type (`int`/`long`), so every one of these was reported.

- Add `BoundedIntegerRange`, computing a best-effort `[lo, hi]` range for an expression and testing whether two ranges make an int subtraction safe (`hi - otherLo <= Integer.MAX_VALUE && lo - otherHi >= Integer.MIN_VALUE`). This subsumes the existing `byte`/`short`/`char` exemption and generalizes to any bounded `int` value, including single-write locals resolved via the existing `ExpressionsHelper.getSingleWriteUsage`.
- Consult it in `IntegerSubtractionInComparisonCheck.replacementFor` only for the `Integer.compare` branch (both operands `int`), leaving `Long.compare` cases unaffected — none of the sampled false positives involve `long` operands, and exempting bounded `long`s (e.g. `java.time` nanosecond fields) would require API invariants outside the type system.
- Add compliant samples for each recognized pattern (array length, `String.length()`, `Collection.size()`, `Map.size()`, `Enum.ordinal()`, `Integer.bitCount()`, masked bytes direct and via locals, single-write local length, literal constants) and noncompliant guard cases for a mixed bounded/unbounded subtraction and an unknown-range `hashCode()` subtraction, to catch a regression that would weaken the "both operands must be bounded" requirement to "either operand".

Intentionally out of scope, and still reported: `indexOf()` results (`[-1, size-1]`, so `Integer.MAX_VALUE - (-1)` overflows in principle), `java.time.Duration`/`Instant` internal nanosecond fields, `java.lang.Enum.ordinal` raw field access (only reachable in JDK source, unlike the `.ordinal()` method call which is now recognized), values from unrelated helper methods, and values bounded only by runtime assertions.

## Links

- Jira: https://sonarsource.atlassian.net/browse/SONARJAVA-6838
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7950

## AI disclosure
- LLM model used for implementation: Claude Sonnet 5

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/helpers/BoundedIntegerRange.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/helpers/BoundedIntegerRange.java
@@ -18,7 +18,6 @@ package org.sonar.java.checks.helpers;
 
 import javax.annotation.CheckForNull;
 import org.sonar.java.model.ExpressionUtils;
-import org.sonar.java.model.LiteralUtils;
 import org.sonar.plugins.java.api.semantic.MethodMatchers;
 import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.semantic.Type;
@@ -104,11 +103,17 @@ public final class BoundedIntegerRange {
 
   @CheckForNull
   private static Range constantRange(ExpressionTree tree) {
-    Long value = LiteralUtils.longLiteralValue(tree);
-    if (value == null) {
-      return null;
+    // Resolve through the compiler's own constant folding rather than re-parsing literal text, so the
+    // value respects the expression's static type: an int-typed 0x80000000 is -2147483648, not +2147483648.
+    Integer intValue = tree.asConstant(Integer.class).orElse(null);
+    if (intValue != null) {
+      return new Range(intValue, intValue);
     }
-    return new Range(value, value);
+    Long longValue = tree.asConstant(Long.class).orElse(null);
+    if (longValue != null) {
+      return new Range(longValue, longValue);
+    }
+    return null;
   }
 
   @CheckForNull
@@ -160,7 +165,9 @@ public final class BoundedIntegerRange {
   @CheckForNull
   private static Range identifierRange(IdentifierTree identifier, int depth) {
     Symbol symbol = identifier.symbol();
-    if (symbol.isUnknown()) {
+    if (symbol.isUnknown() || !ExpressionsHelper.isNotReassigned(symbol)) {
+      // getSingleWriteUsage only looks at AssignmentExpressionTree reassignments, so a variable
+      // mutated by ++/-- would otherwise look like a single-write local pinned to its initializer.
       return null;
     }
     ExpressionTree singleWriteUsage = ExpressionsHelper.getSingleWriteUsage(symbol);
--- a/java-checks/src/main/java/org/sonar/java/checks/helpers/BoundedIntegerRange.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/helpers/BoundedIntegerRange.java
@@ -1,0 +1,176 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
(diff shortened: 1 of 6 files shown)


java-checks/src/test/java/org/sonar/java/checks/helpers/BoundedIntegerRangeTest.java
```diff
@@ -0,0 +1,122 @@
...
+      "static class Node { Node parent; }")).isFalse();
+  }
+
+  private static boolean subtractionCannotOverflow(String... classMembers) {
+    MethodTree method = JParserTestUtils.methodTree(JParserTestUtils.newCode(classMembers));
+    List<StatementTree> statements = method.block().body();
+    ReturnStatementTree returnStatement = (ReturnStatementTree) statements.get(statements.size() - 1);
+    BinaryExpressionTree binary = (BinaryExpressionTree) returnStatement.expression();
+    return BoundedIntegerRange.subtractionCannotOverflow(binary.leftOperand(), binary.rightOperand());
+  }
```
romainbrenguier: We could add the examples to the sample file and test the helper class through that, instead of having these tests.

java-checks/src/main/java/org/sonar/java/checks/helpers/BoundedIntegerRange.java
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * Computes a best-effort [lo, hi] bound for an integer-typed expression, used to prove that
+ * subtracting two such expressions cannot produce a 32-bit overflow.
+ * <p>
+ * This is intentionally conservative: it only recognizes a small set of JDK APIs and operations
+ * with a documented or structurally guaranteed range. Any expression it cannot classify is
+ * treated as unbounded, which keeps every unrecognized case reported as before.
+ */
+public final class BoundedIntegerRange {
```
romainbrenguier, line 39: It looks like we would need some kind of abstract interpretation to make it work well, which is probably too complex. But if this works, it's probably fine....

romainbrenguier (approved):

java-checks/src/test/java/org/sonar/java/checks/helpers/BoundedIntegerRangeTest.java
nathsou: Good point — most of these scenarios are actually valid `compare()`/`compareTo()` bodies, so I moved them into `IntegerSubtractionInComparisonCheckSample.java` as Compliant/Noncompliant cases (some overlapped with samples already added on this branch, e.g. `StringLengthComparator`, `DirectMaskedByteComparator`, `MixedUnboundedAndBoundedComparator`). I kept a single direct unit test for the `long`-constant case, since `subtractionCannotOverflow` is only ever called by the check on `int`-typed operands, so there's no real check call site that reaches that branch. See 01fc1bcf.

java-checks/src/main/java/org/sonar/java/checks/helpers/BoundedIntegerRange.java
nathsou: Agreed, this is meant to be a pragmatic, best-effort heuristic rather than sound abstract interpretation — the class Javadoc calls that out explicitly ("intentionally conservative ... any expression it cannot classify is treated as unbounded"). Since every unrecognized shape falls back to the pre-existing behavior (still reported), the worst case is a missed suppression, not a wrong one. No code change here.

PR #6004, opened 2026-08-24
SONARJAVA-6839: Implement S9362: hashCode() and equals() should use consistent fields

nathsou:
## Summary

- Implement `S9362` (`HashCodeMismatchedFieldsCheck`): flag a locally declared `hashCode()` that reads an instance field which the class's `equals(Object)` override never reads, breaking the `Object.hashCode()` contract.
- The check aggregates every mismatched field for a class into a single issue on the `hashCode()` method name, with one secondary location per mismatched field.
- Supports single-level getter delegation, excludes memoized/cached hash fields (name contains `"hash"`), and bails out (no issue) on any unresolved symbol, unresolvable helper call, or an `equals()` that compares no state.
- Includes `HashCodeMismatchedFieldsCheckSample.java` covering both RSPEC examples plus the memoized-hash, unresolvable-helper, inherited-field, static-field, and asymmetric-direction exceptions, and a `.withoutSemantic()` `CheckVerifier` test (the sample has no external dependency, so local symbol resolution still succeeds without semantic info; documented in the test).
- Generated `S9362.html` / `S9362.json` / `profiles/Sonar_way/S9362` from the RSPEC branch with `rule-api generate`.

## Links

- Jira: https://sonarsource.atlassian.net/browse/SONARJAVA-6839
- RSPEC PR: https://github.com/SonarSource/rspec/pull/7948

## AI disclosure
- LLM model used for implementation: Claude Sonnet 5 (authored directly in the driving session; the workflow's `cursor-grok-4.6-high` routing tier was not delegated to a separate subagent for this step)

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
@@ -57,48 +57,73 @@ public class HashCodeMismatchedFieldsCheck extends IssuableSubscriptionVisitor {
   public void visitNode(Tree tree) {
     ClassTree classTree = (ClassTree) tree;
     Symbol owner = classTree.symbol();
-    if (owner.isUnknown() || owner.type().isUnknown()) {
+
+    EqualsAndHashCode methods = EqualsAndHashCode.find(classTree);
+    if (methods == null) {
       return;
     }
 
-    MethodTree equalsMethod = null;
-    MethodTree hashCodeMethod = null;
-    List<MethodTree> otherMethods = new ArrayList<>();
-    for (Tree member : classTree.members()) {
-      if (!(member instanceof MethodTree methodTree) || methodTree.block() == null) {
-        continue;
-      }
-      if (MethodTreeUtils.isEqualsMethod(methodTree)) {
-        equalsMethod = methodTree;
-      } else if (MethodTreeUtils.isHashCodeMethod(methodTree)) {
-        hashCodeMethod = methodTree;
-      } else {
-        otherMethods.add(methodTree);
-      }
-    }
-    if (equalsMethod == null || hashCodeMethod == null) {
-      return;
-    }
+    Map<Symbol.MethodSymbol, Map<Symbol, Tree>> fieldsByHelper = collectHelperFields(owner, methods.otherMethods);
 
-    Map<Symbol.MethodSymbol, Map<Symbol, Tree>> fieldsByHelper = collectHelperFields(owner, otherMethods);
-
-    FieldReadCollector equalsFields = scan(equalsMethod, owner, fieldsByHelper, Role.EQUALS);
-    FieldReadCollector hashCodeFields = scan(hashCodeMethod, owner, fieldsByHelper, Role.HASH_CODE);
+    FieldReadCollector equalsFields = scan(methods.equalsMethod, owner, fieldsByHelper, Role.EQUALS);
+    FieldReadCollector hashCodeFields = scan(methods.hashCodeMethod, owner, fieldsByHelper, Role.HASH_CODE);
     if (equalsFields.failed || hashCodeFields.failed || equalsFields.fields.isEmpty()) {
       // Bail out on unresolved members, or when equals() compares no state (likely reference equality).
       return;
     }
 
+    Map<Symbol, Tree> extraFields = computeExtraFields(equalsFields, hashCodeFields);
+    if (!extraFields.isEmpty()) {
+      reportMismatch(methods.hashCodeMethod, extraFields);
+    }
+  }
+
+  private static Map<Symbol, Tree> computeExtraFields(FieldReadCollector equalsFields, FieldReadCollector hashCodeFields) {
     Map<Symbol, Tree> extraFields = new LinkedHashMap<>(hashCodeFields.fields);
     extraFields.keySet().removeAll(equalsFields.fields.keySet());
     // A field caching a previously computed hash value does not add new identity state: recognize it either by
     // name, or because hashCode() itself assigns to it (the memoization pattern), regardless of its name.
     extraFields.keySet().removeIf(field -> field.name().toLowerCase(Locale.ROOT).contains("hash") || hashCodeFields.assignedFields.contains(field));
-    if (extraFields.isEmpty()) {
-      return;
+    return extraFields;
+  }
+
+  /**
+   * The locally declared {@code equals(Object)} and {@code hashCode()} methods of a class, plus every other
+   * concrete method declared in that class as helper candidates.
+   */
+  private static final class EqualsAndHashCode {
+
+    private final MethodTree equalsMethod;
+    private final MethodTree hashCodeMethod;
+    private final List<MethodTree> otherMethods;
+
+    private EqualsAndHashCode(MethodTree equalsMethod, MethodTree hashCodeMethod, List<MethodTree> otherMethods) {
+      this.equalsMethod = equalsMethod;
+      this.hashCodeMethod = hashCodeMethod;
+      this.otherMethods = otherMethods;
     }
 
-    reportMismatch(hashCodeMethod, extraFields);
+    private static EqualsAndHashCode find(ClassTree classTree) {
+      MethodTree equalsMethod = null;
+      MethodTree hashCodeMethod = null;
+      List<MethodTree> otherMethods = new ArrayList<>();
+      for (Tree member : classTree.members()) {
+        if (!(member instanceof MethodTree methodTree) || methodTree.block() == null) {
+          continue;
+        }
+        if (MethodTreeUtils.isEqualsMethod(methodTree)) {
+          equalsMethod = methodTree;
+        } else if (MethodTreeUtils.isHashCodeMethod(methodTree)) {
+          hashCodeMethod = methodTree;
+        } else {
+          otherMethods.add(methodTree);
+        }
+      }
+      if (equalsMethod == null || hashCodeMethod == null) {
+        return null;
+      }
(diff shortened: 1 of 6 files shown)


java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
```diff
@@ -0,0 +1,216 @@
...
+    FieldReadCollector hashCodeFields = scan(hashCodeMethod, owner, fieldsByHelper, Role.HASH_CODE);
+    if (equalsFields.failed || hashCodeFields.failed || equalsFields.fields.isEmpty()) {
+      // Bail out on unresolved members, or when equals() compares no state (likely reference equality).
+      return;
+    }
+
+    Map<Symbol, Tree> extraFields = new LinkedHashMap<>(hashCodeFields.fields);
+    extraFields.keySet().removeAll(equalsFields.fields.keySet());
+    // A field caching a previously computed hash value does not add new identity state.
+    extraFields.keySet().removeIf(field -> field.name().toLowerCase(Locale.ROOT).contains("hash"));
```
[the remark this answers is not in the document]
nathsou: Addressed in c715ba7d27. The broad `contains("hash")` suppression is now limited to an explicit set of conventional cache-field names, so identity fields such as `contentHash` are reported. I added regression coverage for both an eager `cachedHash` cache and a mismatched `contentHash` identity field.

Validation: `mvn -pl java-checks -am test -Dtest=HashCodeMismatchedFieldsCheckTest -Dsurefire.failIfNoSpecifiedTests=false` passes.

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
```diff
@@ -0,0 +1,216 @@
...
+
+  private void reportMismatch(MethodTree hashCodeMethod, Map<Symbol, Tree> extraFields) {
+    List<String> names = extraFields.keySet().stream()
+      .map(Symbol::name)
+      .sorted()
+      .toList();
+    List<JavaFileScannerContext.Location> secondaryLocations = extraFields.values().stream()
+      .map(location -> new JavaFileScannerContext.Location(SECONDARY_MESSAGE, location))
+      .toList();
+    reportIssue(hashCodeMethod.simpleName(), String.format(ISSUE_MESSAGE, String.join("\", \"", names)), secondaryLocations, null);
```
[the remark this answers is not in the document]
nathsou: Addressed in c715ba7d27. Fields discovered through a helper are now associated with the helper invocation tree in the calling method, so the secondary location points to `getLastName()` in `hashCode()` rather than the field read inside the getter. The precise secondary-location assertion was updated accordingly.

Validation: `mvn -pl java-checks -am test -Dtest=HashCodeMismatchedFieldsCheckTest -Dsurefire.failIfNoSpecifiedTests=false` passes.

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
```diff
@@ -0,0 +1,239 @@
...
+    @Override
+    public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+      if (tree.variable() instanceof IdentifierTree identifier) {
+        Symbol symbol = identifier.symbol();
+        if (!symbol.isUnknown() && symbol.isVariableSymbol() && !symbol.isStatic() && ownedByEnclosing(symbol)) {
+          assignedFields.add(symbol);
+        }
+      }
+      super.visitAssignmentExpression(tree);
+    }
```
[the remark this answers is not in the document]
nathsou: Addressed in c715ba7d27. Assignment tracking now uses `ExpressionUtils.extractIdentifierSymbol`, which recognizes bare assignments and `this.`/`super.`-qualified member selections. The memoization regression sample now uses a non-conventional field name (`hc`) and `this.hc = ...`, proving the exclusion is assignment-based rather than name-based.

Validation: `mvn -pl java-checks -am test -Dtest=HashCodeMismatchedFieldsCheckTest -Dsurefire.failIfNoSpecifiedTests=false` passes.

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
```diff
@@ -0,0 +1,271 @@
...
+    Symbol owner = classTree.symbol();
+
+    EqualsAndHashCode methods = EqualsAndHashCode.find(classTree);
+    if (methods == null) {
+      return;
+    }
+
+    Map<Symbol.MethodSymbol, Map<Symbol, Tree>> fieldsByHelper = collectHelperFields(owner, methods.otherMethods);
+
+    FieldReadCollector equalsFields = scan(methods.equalsMethod, owner, fieldsByHelper, Role.EQUALS);
```
romainbrenguier: `scan` is a bit too generic of a name. I think `collectReadFields` would be more descriptive

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
```diff
@@ -0,0 +1,271 @@
...
+    FieldReadCollector hashCodeFields = scan(methods.hashCodeMethod, owner, fieldsByHelper, Role.HASH_CODE);
```
romainbrenguier: I think we don't need to return a `Collector` object here, since the collection part is done after the scan. To make things clearer, we could have a separate class for the result: `ReadAndAssignedFields` or something like that.

romainbrenguier (commented):
It looks ok but I think we can clarify a bit the Collector class part.

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
nathsou: Addressed in f4b248806a. I renamed `scan` to `collectReadFields`, making the method's purpose explicit at each call site.

Validation: `mvn -pl java-checks -am test -Dtest=HashCodeMismatchedFieldsCheckTest -Dsurefire.failIfNoSpecifiedTests=false` passes.

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
nathsou: Addressed in f4b248806a. `collectReadFields` now returns a dedicated `ReadAndAssignedFields` result, while `FieldReadCollector` remains only the mutable AST visitor used during traversal. Callers no longer consume the visitor itself as the result object.

Validation: `mvn -pl java-checks -am test -Dtest=HashCodeMismatchedFieldsCheckTest -Dsurefire.failIfNoSpecifiedTests=false` passes.

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  private enum Role {
+    /** Pre-scanning a candidate getter/helper method: no instance calls are trusted, and helpers are not chained. */
+    HELPER,
+    EQUALS,
+    HASH_CODE
+  }
+
+  private static final class ReadAndAssignedFields {
```
romainbrenguier: I looks like this could have been a record type.

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
```diff
@@ -0,0 +1,288 @@
...
+   * failed: callers must then skip reporting entirely instead of guessing from a partial field set.
+   */
+  private static final class FieldReadCollector extends BaseTreeVisitor {
+
+    private final Symbol enclosingClass;
+    private final Map<Symbol.MethodSymbol, Map<Symbol, Tree>> fieldsByHelper;
+    private final Role role;
+    private final Map<Symbol, Tree> readFields = new LinkedHashMap<>();
+    private final Set<Symbol> assignedFields = new HashSet<>();
+    private boolean failed;
```
romainbrenguier: These could have been replaced with a ReadAndAssignedFields field

romainbrenguier (approved):
Some non-blocking comments.

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
nathsou: Addressed in 8b7428faa9. `ReadAndAssignedFields` is now a private record, removing the data-carrier boilerplate while retaining the insertion-ordered map and mutable set used during traversal.

Validation: `mvn -pl java-checks -am test -Dtest=HashCodeMismatchedFieldsCheckTest -Dsurefire.failIfNoSpecifiedTests=false` passes.

java-checks/src/main/java/org/sonar/java/checks/HashCodeMismatchedFieldsCheck.java
nathsou: Addressed in 8b7428faa9. `FieldReadCollector` now owns one initialized `ReadAndAssignedFields` field instead of separately declaring its read fields, assigned fields, and failure state. When a scan fails, the collector replaces that record with an `asFailed()` copy while retaining the collected maps.

Validation: `mvn -pl java-checks -am test -Dtest=HashCodeMismatchedFieldsCheckTest -Dsurefire.failIfNoSpecifiedTests=false` passes.

PR #6022, opened 2026-08-25
SONARJAVA-6825: Implemented rule S9360 - Comparisons should not use Yoda conditions

romainbrenguier:
This PR implements rule S9360 'Constant expressions and comparisons should be simplified'.

The rule detects **Yoda conditions** where a constant literal appears on the left side of a comparison operator.

## What the rule detects

- `0 == count` → should be `count == 0`
- `null == obj` → should be `obj == null`
- `true == flag` → should be `flag == true`
- `5 != x` → should be `x != 5`
- `0 < count` → should be `count > 0`
- `5 > x` → should be `x < 5`

## Supported operators

- `==` (equality)
- `!=` (inequality)
- `<` (less than)
- `>` (greater than)

## Supported literal types

- Integer literals (`0`, `5`, `42`)
- Long literals (`0L`, `5L`)
- Floating-point literals (`0.0`, `3.14`)
- Boolean literals (`true`, `false`)
- Character literals (`'a'`)
- String literals (`"hello"`, `""`)
- Null literal (`null`)

## Edge cases handled

- Nested parentheses: `((0)) == count` is detected
- Both operands are literals: `0 == 0` is compliant (nothing to swap)
- Both operands are variables: `count == otherCount` is compliant
- Non-comparison contexts: assignments, arithmetic, method calls are ignored

## Implementation details

- Uses `IssuableSubscriptionVisitor` pattern
- Subscribes to `EQUAL_TO`, `NOT_EQUAL_TO`, `LESS_THAN`, `GREATER_THAN` tree kinds
- Uses `ExpressionUtils.skipParentheses()` to handle nested parentheses
- Reports issue on the literal (left operand) with message "Put the variable on the left side of this comparison."

## Test coverage

- All literal types with `==` and `!=` operators
- Less than and greater than operators
- Nested parentheses handling
- Non-comparison contexts (assignments, arithmetic, method calls)
- Edge cases: both literals, both variables, ternary operators

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
@@ -27,13 +27,18 @@ import org.sonar.plugins.java.api.tree.Tree;
 @Rule(key = "S9360")
 public class YodaConditionCheck extends IssuableSubscriptionVisitor {
 
+  private static final String MESSAGE = "Put the variable on the left side of this comparison.";
+  private static final String MESSAGE_WITH_REVERSE = "Put the variable on the left side of this comparison and reverse the operator.";
+
   @Override
   public List<Tree.Kind> nodesToVisit() {
     return List.of(
       Tree.Kind.EQUAL_TO,
       Tree.Kind.NOT_EQUAL_TO,
       Tree.Kind.LESS_THAN,
-      Tree.Kind.GREATER_THAN
+      Tree.Kind.GREATER_THAN,
+      Tree.Kind.LESS_THAN_OR_EQUAL_TO,
+      Tree.Kind.GREATER_THAN_OR_EQUAL_TO
     );
   }
 
@@ -44,8 +49,17 @@ public class YodaConditionCheck extends IssuableSubscriptionVisitor {
     ExpressionTree right = ExpressionUtils.skipParentheses(binaryExpression.rightOperand());
 
     if (isLiteral(left) && !isLiteral(right)) {
-      reportIssue(left, "Put the variable on the left side of this comparison.");
+      reportIssue(left, isRelationalOperator(tree) ? MESSAGE_WITH_REVERSE : MESSAGE);
     }
+  }
+
+  private static boolean isRelationalOperator(Tree tree) {
+    return tree.is(
+      Tree.Kind.LESS_THAN,
+      Tree.Kind.GREATER_THAN,
+      Tree.Kind.LESS_THAN_OR_EQUAL_TO,
+      Tree.Kind.GREATER_THAN_OR_EQUAL_TO
+    );
   }
 
   private static boolean isLiteral(ExpressionTree tree) {
--- a/java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
@@ -1,0 +1,63 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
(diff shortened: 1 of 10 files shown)


romainbrenguier:
## Changes addressing review comments

1. **Fixed compilation error** - `null == count` where `count` is `int` (primitive) now uses an `Object` variable instead
2. **Narrowed rule description** - Removed "Constant Function Calls" section (`Math.max`, `Math.sqrt`, etc.) since the implementation only detects Yoda conditions
3. **Fixed compliant example** - Changed `flag == true` to `flag` (avoids conflict with S1125) and `4.0 + 5` to `9.0`
4. **Improved relational operator message** - For `<`, `>`, `<=`, `>=`, the message now says "Put the variable on the left side of this comparison **and reverse the operator**"
5. **Added `<=` and `>=` support** - `LESS_THAN_OR_EQUAL_TO` and `GREATER_THAN_OR_EQUAL_TO` are now detected
6. **Merged ruling fix PR #6023** - Ruling expectations for the original findings are included

> Note: The addition of `<=` and `>=` operators may produce new ruling differences that will require a follow-up ruling update.

romainbrenguier:
gitar unblock

java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
```diff
@@ -0,0 +1,77 @@
...
+
+  private static boolean isLiteral(ExpressionTree tree) {
+    return tree.is(
+      Tree.Kind.INT_LITERAL,
+      Tree.Kind.LONG_LITERAL,
+      Tree.Kind.FLOAT_LITERAL,
+      Tree.Kind.DOUBLE_LITERAL,
+      Tree.Kind.BOOLEAN_LITERAL,
+      Tree.Kind.CHAR_LITERAL,
+      Tree.Kind.STRING_LITERAL,
```
nathsou, line 73: These literal kinds cause verified duplicate findings for equality operators. I ran focused `CheckVerifier` tests on this head with exactly each relevant rule pair enabled: `true == flag` produced S9360 + S1125, `"hello" == text` produced S9360 + S4973, and `0.0 == value` produced S9360 + S1244. S1125 and S4973 are both in Sonar Way, so the first two duplicates occur in the default profile. Please make equality handling operator/type-aware: defer boolean equality to S1125, String equality to S4973, and floating-point equality to S1244, while retaining non-overlapping relational comparisons.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9360.json
```diff
@@ -0,0 +1,25 @@
...
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "5 min"
+  },
+  "tags": [
+    "confusing",
+    "convention",
+    "clarity"
+  ],
+  "defaultSeverity": "Major",
```
nathsou: This points to draft RSPEC PR 7941, but that source RSPEC still requires precomputing constant method calls and illustrates `Math.max`, `Math.sqrt`, and `Math.abs`. A runtime test with S9360 and S2185 did *not* double-report those calls because S9360 never visits method invocations; that instead confirms the analyzer does not implement its authoritative specification. Constant math is already owned at least in part by S2185. Please narrow RSPEC-9360 to Yoda conditions before merging and keep the imported documentation aligned with that source.

java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
```diff
@@ -0,0 +1,77 @@
...
+    return tree.is(
+      Tree.Kind.LESS_THAN,
+      Tree.Kind.GREATER_THAN,
+      Tree.Kind.LESS_THAN_OR_EQUAL_TO,
+      Tree.Kind.GREATER_THAN_OR_EQUAL_TO
+    );
+  }
+
+  private static boolean isLiteral(ExpressionTree tree) {
+    return tree.is(
```
nathsou, line 66: This recognizes only bare literal AST nodes, not Java constant expressions. For example, `-1 == result` is a unary-minus expression and is silently missed; named constants such as `Integer.MAX_VALUE == result` and expressions such as `1 + 1 == result` are missed as well. That conflicts with the rule's user-facing `constant`/Yoda-condition scope. Please either recognize constant expressions (at minimum unary `+`/`-` numeric literals) or explicitly narrow the specification and title to bare literals, with tests documenting the boundary.

nathsou (changes requested):
Requesting changes for three blocking scope issues: verified duplicate findings with existing Java rules, divergence from the authoritative draft RSPEC, and incomplete handling of constants. The duplicate-report cases were reproduced with focused CheckVerifier tests on this exact head.

romainbrenguier:
superseeded by https://github.com/SonarSource/sonar-java/pull/6036

PR #6025, opened 2026-08-25
SONARJAVA-4647 Improve S120 to report one project-level issue per bad package name

asya-vorobeva:
Previously, the rule raised an issue on every file within a non-compliant package, producing n issues for n classes in the same package. The rule now collects unique bad package names across all files and reports a single project-level issue per package via endOfAnalysis().

### DIFF
--- a/java-checks/src/test/java/org/sonar/java/checks/naming/BadPackageNameCheckTest.java
+++ b/java-checks/src/test/java/org/sonar/java/checks/naming/BadPackageNameCheckTest.java
@@ -21,12 +21,14 @@ import org.sonar.java.checks.verifier.CheckVerifier;
 
 class BadPackageNameCheckTest {
 
+  private static final String DEFAULT_FORMAT = "^[a-z_]+(\\.[a-z_][a-z0-9_]*)*$";
+
   @Test
   void test() {
     CheckVerifier.newVerifier()
       .onFile("src/test/files/checks/PACKAGE/BadPackageNameNoncompliant.java")
       .withCheck(new BadPackageNameCheck())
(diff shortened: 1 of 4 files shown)


java-checks/src/test/java/org/sonar/java/checks/naming/BadPackageNameCheckTest.java
```diff
@@ -16,17 +16,41 @@
...
+
   @Test
   void test() {
     CheckVerifier.newVerifier()
-      .onFile("src/test/files/checks/PACKAGE/BadPackageNameNoncompliant.java")
+      .onFile(NONCOMPLIANT_FILE)
       .withCheck(new BadPackageNameCheck())
-      .verifyIssues();
+      .verifyIssueOnProject("Rename package \"PACKAGE\" to match the regular expression '" + DEFAULT_FORMAT + "'.");
   }
```
erwan-leforestier-sonarsource, line 54: The PR's core behavioral claim, that N files sharing the same bad package produce exactly one project-level issue, is never tested. Every unit test uses a single file, so a regression (e.g. swapping the dedup HashSet for a List) would go undetected. Could you add a test with two or more files in the same bad package that calls verifyIssueOnProject(expectedMessage) once and asserts no duplicate?

asya-vorobeva: Project issues are designed with the way that they don't raise duplicates. Removing issues in files already is a signal that we'll have one issue (per project).

erwan-leforestier-sonarsource (approved):

PR #6035, opened 2026-08-25
SONARJAVA-6853 S3329: fix FP when cipher.init uses a non-final variable initialized to DECRYPT_MODE

asya-vorobeva:
The DECRYPT_MODE suppression in SecureInitializationFinder relied on asConstant() to resolve the operation mode, which only works for compile-time constants. A non-final variable initialized to Cipher.DECRYPT_MODE was not resolved, causing the suppression to silently fail and a FP to be raised.

Added resolveIntConstant() which falls back to inspecting the variable's declaration initializer when asConstant() returns empty.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/security/CipherBlockChainingCheck.java
```diff
@@ -241,6 +242,21 @@ private static boolean isPartOfArguments(MethodInvocationTree methodInvocation,
...
+    private static Optional<Integer> resolveIntConstant(ExpressionTree expression) {
+      Optional<Integer> constant = expression.asConstant(Integer.class);
+      if (constant.isPresent()) {
+        return constant;
+      }
+      if (expression instanceof IdentifierTree identifierTree
+        && identifierTree.symbol() instanceof Symbol.VariableSymbol variableSymbol) {
+        var declaration = variableSymbol.declaration();
+        if (declaration != null && declaration.initializer() != null) {
+          return declaration.initializer().asConstant(Integer.class);
```
erwan-leforestier-sonarsource, line 254: `resolveIntConstant` reads only the declaration initializer, so a reassignment before `cipher.init` is ignored: `int mode = Cipher.DECRYPT_MODE; mode = Cipher.ENCRYPT_MODE; cipher.init(mode, key, new IvParameterSpec(new byte[16]));` raises no S3329, silently suppressing a real encryption-with-fixed-IV case (a false negative on a security rule). Could you resolve the value at the call site with `ReassignmentFinder.getClosestReassignmentOrDeclarationExpression` instead of the declaration initializer?

asya-vorobeva: In the comment for this rule we specialized that we don't plan to be super-precise for this rule implementation. Proper fix requires Taint Analysis / DBD-based technics. That's why we don't plan to improve it significantly. This fix was added only for fixing recent Community issue.

erwan-leforestier-sonarsource: Ok fair enough

erwan-leforestier-sonarsource (approved):

PR #6036, opened 2026-08-25
SONARJAVA-6825: Implement rule S9360 - Constant expressions and comparisons should be simplified

romainbrenguier:
## Summary
- Implements rule S9360: Constant expressions and comparisons should be simplified
- **Yoda conditions**: Detects comparisons where a literal is on the left side (`0 == count`, `null != obj`) and suggests putting the variable on the left
- **Constant Math calls**: Detects calls to deterministic `Math` methods where all arguments are compile-time literals (`Math.max(5, 10)`, `Math.sqrt(16.0)`, `Math.abs(-5)`) and suggests replacing with precomputed values
- Covers all comparison operators: `==`, `!=`, `<`, `>`, `<=`, `>=`
- Handles unary minus/plus in Math call arguments (e.g., `Math.abs(-10)`)
- Uses `MethodMatchers` for type-safe Math method detection with semantic analysis

## Test plan
- [x] Unit tests pass (2 tests: with semantic, non-compiling)
- [x] Sonar way profile test passes
- [ ] Ruling QA CI checks pass (ruling expectation files will be auto-generated)

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
@@ -20,12 +20,65 @@ import java.util.List;
 import org.sonar.check.Rule;
 import org.sonar.java.model.ExpressionUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
 import org.sonar.plugins.java.api.tree.BinaryExpressionTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
 import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.UnaryExpressionTree;
 
 @Rule(key = "S9360")
 public class YodaConditionCheck extends IssuableSubscriptionVisitor {
+
+  private static final String JAVA_LANG_MATH = "java.lang.Math";
+
+  private static final MethodMatchers CONSTANT_MATH_METHODS = MethodMatchers.or(
+    MethodMatchers.create().ofTypes(JAVA_LANG_MATH)
+      .names("max", "min")
+      .addParametersMatcher("int", "int")
+      .addParametersMatcher("long", "long")
+      .addParametersMatcher("float", "float")
+      .addParametersMatcher("double", "double")
+      .build(),
+    MethodMatchers.create().ofTypes(JAVA_LANG_MATH)
+      .names("abs", "absExact", "negateExact", "incrementExact", "decrementExact")
+      .addParametersMatcher("int")
+      .addParametersMatcher("long")
+      .build(),
+    MethodMatchers.create().ofTypes(JAVA_LANG_MATH)
+      .names("abs")
+      .addParametersMatcher("float")
+      .addParametersMatcher("double")
+      .build(),
+    MethodMatchers.create().ofTypes(JAVA_LANG_MATH)
+      .names("sqrt", "cbrt", "ceil", "floor", "rint", "log", "log10", "exp",
+        "sin", "cos", "tan", "asin", "acos", "atan", "sinh", "cosh", "tanh",
+        "toDegrees", "toRadians", "signum", "expm1", "log1p")
+      .addParametersMatcher("double")
+      .build(),
+    MethodMatchers.create().ofTypes(JAVA_LANG_MATH)
+      .names("round")
+      .addParametersMatcher("float")
+      .addParametersMatcher("double")
+      .build(),
+    MethodMatchers.create().ofTypes(JAVA_LANG_MATH)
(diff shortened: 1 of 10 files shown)


nathsou (changes requested):
PR #6036 and PR #6022 are competing implementations of the same Jira deliverable, not complementary changes. SONARJAVA-6825 has one stated objective—create S9360—and no description or comments defining separate deliverables. Both PRs share the initial S9360 commits through `36d0e824`, then diverge: #6022 implements the Yoda-condition subset, while this PR implements Yoda conditions plus constant `Math` calls. Both add the same rule key, check class, metadata, documentation, Sonar Way activation, and ruling files, so they cannot reasonably both merge. Please establish one canonical implementation before proceeding. If #6036 supersedes #6022, state that explicitly and close #6022; otherwise explain the intended split and re-scope the work so the two PRs no longer compete for S9360.

romainbrenguier:
@nathsou I closed the other PR

java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
```diff
@@ -0,0 +1,158 @@
...
+
+  private static final MethodMatchers CONSTANT_MATH_METHODS = MethodMatchers.or(
+    MethodMatchers.create().ofTypes(JAVA_LANG_MATH)
+      .names("max", "min")
+      .addParametersMatcher("int", "int")
+      .addParametersMatcher("long", "long")
+      .addParametersMatcher(FLOAT, FLOAT)
+      .addParametersMatcher(DOUBLE, DOUBLE)
+      .build(),
+    MethodMatchers.create().ofTypes(JAVA_LANG_MATH)
```
nathsou, line 45: **[P1] S9360 still duplicates S2185 in Sonar Way.** The earlier overlap thread was resolved after narrowing the matcher set, but the overlap remains for the methods still listed here: a focused analyzer run produced two issues for both `Math.abs(5)` and `Math.sqrt(1.0)`, one from S9360 and one from S2185. Both rules are enabled in Sonar Way, so users receive duplicate default-profile findings on the same invocation. Please remove constant-math coverage from S9360 or partition it from S2185 explicitly before enabling both by default.

java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
```diff
@@ -0,0 +1,158 @@
...
+    if (expr.is(Tree.Kind.UNARY_MINUS, Tree.Kind.UNARY_PLUS)) {
+      expr = ExpressionUtils.skipParentheses(((UnaryExpressionTree) expr).expression());
+    }
+    return expr.is(
+      Tree.Kind.INT_LITERAL,
+      Tree.Kind.LONG_LITERAL,
+      Tree.Kind.FLOAT_LITERAL,
+      Tree.Kind.DOUBLE_LITERAL,
+      Tree.Kind.CHAR_LITERAL,
+      Tree.Kind.STRING_LITERAL,
```
nathsou, line 154: **[P1] Exclude equality cases already governed by dedicated rules.** A focused analyzer run on `"hello" == text` produced both S9360 and S4973, which are both enabled in Sonar Way. Following S9360's suggestion merely changes this to `text == "hello"`, leaving the S4973 defect in place. Likewise, `0.0 == value` double-reports with S1244 when that rule is enabled. Please make equality handling type-aware: exclude string equality and floating-point equality while retaining relational comparisons where appropriate.

java-checks/src/main/java/org/sonar/java/checks/YodaConditionCheck.java
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  private void checkConstantMathCall(MethodInvocationTree methodInvocation) {
+    if (CONSTANT_MATH_METHODS.matches(methodInvocation) && allArgumentsAreLiterals(methodInvocation)) {
+      reportIssue(methodInvocation.methodSelect(),
+        String.format("Replace this call to \"%s\" with the precomputed constant value.", methodInvocation.methodSymbol().name()));
+    }
+  }
+
+  private static boolean allArgumentsAreLiterals(MethodInvocationTree methodInvocation) {
```
nathsou, line 126: **[P2] The implementation does not recognize the documented compile-time constants.** This predicate only accepts numeric literals (optionally under unary `+`/`-`). Focused analyzer tests confirmed that neither `static final int FIVE = 5; Math.max(FIVE, 10)` nor `Math.max(2 + 3, 10)` raises S9360, although both arguments are compile-time constants under the RSPEC/HTML wording. Please resolve constant expressions rather than only testing syntax; the frontend already exposes `ExpressionUtils.resolveAsConstant` and symbol constant values for this purpose.

nathsou (changes requested):
The focused analyzer tests confirm three remaining scope and overlap problems that should be addressed before merging.

romainbrenguier:
I will close for now because the scope of the RSpec needs to be reworked, as there is overlap with another rule.

PR #6044, opened 2026-08-26
SONARJAVA-6421 Implement S9352: Bean autowiring ambiguity should be resolved using "@Qualifier" or "@Primary"

NoemieBenard:
----
## Summary by Gitar

- **New checks:**
    - Implemented Spring rule `S9352` to detect ambiguous bean autowiring without `@Qualifier` or `@Primary`
- **Model & Index updates:**
    - Added `TypeToDependenciesIndex` and integrated it into `BeanDefinitionGatherer` to track injection points


java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -0,0 +1,99 @@
...
+        Set<String> effectiveCandidates = excludeFallbackCandidates(candidates, registry);
+        if (effectiveCandidates.size() > 1) {
+          ambiguousDependencies.add(new AmbiguousDependency(bean.getLocation(), message(requiredType, effectiveCandidates)));
+        }
+      }
+    }
+    return ambiguousDependencies;
+  }
+
+  private static boolean isResolved(Set<String> candidates, Set<String> injectionPointNames, BeanDefinitionRegistry registry) {
```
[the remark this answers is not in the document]
NoemieBenard: > Spring drops the fallback bean before by-name matching and throws NoUniqueBeanDefinitionException

That's not correct, Spring will recognize the `@Fallback` bean correctly, so no issue needs to be raised.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -0,0 +1,122 @@
...
+
+  private static final String MESSAGE = "Multiple beans of type \"%s\" match this dependency (%s);"
+    + " disambiguate it with \"@Qualifier\" or mark one bean as \"@Primary\".";
+
+  public List<AmbiguousDependency> findAmbiguousDependencies(SpringContextModel model) {
+    BeanDefinitionRegistry registry = model.getBeanDefinitionRegistry();
+    TypeToBeanNamesIndex typeToBeanNamesIndex = model.getTypeToBeanNamesIndex();
+
+    List<AmbiguousDependency> ambiguousDependencies = new ArrayList<>();
+    for (BeanDefinitionHolder bean : registry.getAll()) {
```
asya-vorobeva: Here makes sense to iterate in outer loop through TypeToBeanNamesIndex map to inspect types, then for each type iterate through specific beans using BeanDefinitionRegistry.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -0,0 +1,82 @@
...
+      for (Map.Entry<String, Set<String>> dependency : bean.getDependingBeans().entrySet()) {
+        String requiredType = dependency.getKey();
+        Set<String> candidates = typeToBeanNamesIndex.getNamesForType(requiredType);
+        if (isAmbiguous(candidates, dependency.getValue(), registry)) {
+          ambiguousDependencies.add(new AmbiguousDependency(bean.getLocation(), message(requiredType, candidates)));
+        }
+      }
+    }
+    return ambiguousDependencies;
```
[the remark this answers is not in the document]
asya-vorobeva: We can postpone handling situation with several bean name aliases, as it's moderately low. Let's fix this as follow-up in a separate JIRA ticket.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  private static boolean isResolved(Set<String> candidates, Set<String> injectionPointNames, BeanDefinitionRegistry registry) {
+    // injectionPointNames merges every injection point of this type declared on the bean: it is only resolved
+    // if EVERY one of them names a candidate (by bean name or by a @Qualifier declared on that candidate bean
+    // itself), otherwise at least one injection point remains ambiguous.
+    return candidates.size() <= 1
+      || injectionPointNames.stream().allMatch(name -> matchesCandidate(name, candidates, registry))
+      || hasExactlyOnePrimaryCandidate(candidates, registry);
```
asya-vorobeva, line 77: Here it's very important to check for having Profile set for the candidates. Candidates having configured profile should be excluded from consideration as possibly mutually exclusive.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -0,0 +1,122 @@
...
+
+    List<AmbiguousDependency> ambiguousDependencies = new ArrayList<>();
+    for (BeanDefinitionHolder bean : registry.getAll()) {
+      for (Map.Entry<String, Set<String>> dependency : bean.getDependingBeans().entrySet()) {
+        String requiredType = dependency.getKey();
+        Set<String> candidates = typeToBeanNamesIndex.getNamesForType(requiredType);
+        if (!isResolved(candidates, dependency.getValue(), registry)) {
+          // A @Fallback candidate is only a real contender when it is the sole remaining one; otherwise it is
+          // ignored by Spring, so the effective candidates are whichever bean(s) are not marked @Fallback.
+          Set<String> effectiveCandidates = excludeFallbackCandidates(candidates, registry);
```
asya-vorobeva: Fallback annotation is modern (introduced in Spring Framework 6.2) and thus currently low used. We can add it later in follow-up ticket.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -0,0 +1,122 @@
...
+          if (effectiveCandidates.size() > 1) {
+            ambiguousDependencies.add(new AmbiguousDependency(bean.getLocation(), message(requiredType, effectiveCandidates)));
```
asya-vorobeva: As we've discussed, another alternative is to raise issues on project level instead. Drawback is that we won't see it in SonarLint. So the current approach looks appropriate.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/SpringContextModelSensor.java
```diff
@@ -49,6 +55,21 @@ public void describe(SensorDescriptor descriptor) {
 
   @Override
   public void execute(SensorContext context) {
-    // Nothing to do for now
+    reportAmbiguousDependencies(context);
```
asya-vorobeva: Here we have in mind that we'll implement multiple checks which will be called here. So I'd recommend to use Strategy design pattern.
For it we need to create some common interface with some execute method which all the needed checks will implement, and common record with information needed to create an issue (generalize AmbiguousDependency).
Then we can inject all of them into this sensor (injection mechanism is on you) and run execute method for all of them in a loop.

asya-vorobeva (commented):
To properly test such checks, please use scanner-integration-framework capabilities provided in [this ticket](https://sonarsource.atlassian.net/browse/SONARJAVA-6365).

its/scanner-integration-tests/src/test/resources/ambiguous-dependencies-should-be-resolved/module-a/src/main/java/com/example/modulea/CacheConfig.java
```diff
@@ -0,0 +1,24 @@
+package com.example.modulea;
+
+import com.example.common.CacheProvider;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class CacheConfig {
+
+    @Bean
```
asya-vorobeva, line 10: Let's configure these beans in different modules / files to be sure that it recognizes such cases.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -0,0 +1,96 @@
...
+ * bean of a given type across the whole analyzed scope, not a single file.
+ */
+@Rule(key = "S9352")
+public class AmbiguousDependencyCheck implements JavaCheck, SpringContextCheck {
+
+  private static final String MESSAGE = "Multiple beans match this dependency (%s);"
+    + " disambiguate it with \"@Qualifier\" or mark one bean as \"@Primary\".";
+
+  @Override
+  public List<SpringContextIssue> execute(SpringContextModel model) {
```
asya-vorobeva, line 54: This method is super-hard to read. Let's add Javadoc comment describing how it acts. If you will fill that it's not enough, add additional internal comments in the method.

asya-vorobeva (changes requested):
This PR became too huge. Would be great to separate it into several ones: one with SpringModel refactoring, another with rule implementation (including tests). Or at least to re-group-squash commits to have clear history and when merging not do squash.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -0,0 +1,96 @@
...
+  private static Set<String> excludeCandidatesWithProfile(Set<String> candidates, BeanDefinitionRegistry registry) {
+    return candidates.stream().filter(candidate -> !hasProfile(registry, candidate)).collect(Collectors.toUnmodifiableSet());
+  }
+
+  private static boolean isPrimary(BeanDefinitionRegistry registry, String beanName) {
+    return registry.getByName(beanName).stream().anyMatch(BeanDefinitionHolder::isPrimary);
+  }
+
+  private static boolean hasProfile(BeanDefinitionRegistry registry, String beanName) {
+    return registry.getByName(beanName).stream().anyMatch(bean -> bean.getProfiles() != null);
```
asya-vorobeva, line 97: I couldn't find a place where we add profiles information to the bean holder. Probably we missed it? Should be done in BeanDefinitionGatherer.

NoemieBenard:
Closing this PR after splitting it into smaller ones: https://github.com/SonarSource/sonar-java/pull/6068, https://github.com/SonarSource/sonar-java/pull/6071, https://github.com/SonarSource/sonar-java/pull/6073

PR #6049, opened 2026-08-27
SONARJAVA-6660 Fix S2229 FP for self-invocations inside TransactionTemplate callbacks

asya-vorobeva:
Self-invocations inside TransactionTemplate.execute/executeWithoutResult callbacks are suppressed because the template guarantees an active transaction at the call site. Resolving the template's configured propagation statically would require data-flow or symbolic-execution techniques, so all self-invocations within such callbacks are treated as compliant.

### DIFF
--- a/java-checks-test-sources/default/pom.xml
+++ b/java-checks-test-sources/default/pom.xml
@@ -371,7 +371,7 @@
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-tx</artifactId>
-      <version>5.0.6.RELEASE</version>
+      <version>5.3.21</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
@@ -31,6 +31,7 @@ import org.sonar.java.checks.helpers.SpringUtils;
 import org.sonar.java.model.ExpressionUtils;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
 import org.sonar.plugins.java.api.semantic.Symbol;
 import org.sonar.plugins.java.api.semantic.SymbolMetadata;
 import org.sonar.plugins.java.api.semantic.SymbolMetadata.AnnotationValue;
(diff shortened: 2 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -59,6 +60,12 @@ public class SpringIncompatibleTransactionalCheck extends IssuableSubscriptionVi
   // Made name to represent no annotation
   private static final String NOT_TRANSACTIONAL = "SONAR_NOT_TRANSACTIONAL";
 
+  private static final MethodMatchers TRANSACTION_TEMPLATE_EXECUTE = MethodMatchers.create()
+    .ofTypes("org.springframework.transaction.support.TransactionTemplate")
+    .names("execute", "executeWithoutResult")
+    .withAnyParameters()
+    .build();
```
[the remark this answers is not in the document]
asya-vorobeva: TransactionOperations has only one standard subclass which executes in transactional mode: TransactionTemplate. Using custom subclasses is possible, but rare.

java-checks/src/main/java/org/sonar/java/checks/spring/SpringIncompatibleTransactionalCheck.java
```diff
@@ -96,14 +103,21 @@ private void checkMethodInvocations(MethodTree method, @Nullable String callerPr
...
+        boolean previousState = insideTransactionTemplateCallback;
+        if (TRANSACTION_TEMPLATE_EXECUTE.matches(methodInvocation)) {
+          insideTransactionTemplateCallback = true;
+        }
         super.visitMethodInvocation(methodInvocation);
+        insideTransactionTemplateCallback = previousState;
         Symbol calleeMethodSymbol = methodInvocation.methodSymbol();
         if (calleeMethodSymbol.isUnknown()) {
           return;
         }
```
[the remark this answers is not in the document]
asya-vorobeva: Fixing this bug looks like significant over-engineering. About examples:
- execute(buildCallback(retryNew(id))): calling a transactional method  eagerly as an argument to a callback factory is very unusual. If retryNew returns void, it can't even appear there syntactically.
- templateFor(retry(id)).execute(status -> ...) — retry would have to return a TransactionTemplate. Essentially impossible in practice.

So, both examples are very unlikely to appear, benefit of catching them is very low.

java-checks-test-sources/default/pom.xml
```diff
@@ -371,7 +371,7 @@
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-tx</artifactId>
-      <version>5.0.6.RELEASE</version>
+      <version>5.3.21</version>
```
lijun-chen-sonarsource, line 374: What is the purpose of this version upgrade?

lijun-chen-sonarsource (approved):
LGTM!

I've seen your counterarguments to Gitar's comments. I think they make sense.
Do you think it'd make sense to add test cases that make explicit that these two cases are FN and FP that we intentionally accept?

java-checks-test-sources/default/pom.xml
asya-vorobeva: Because previous version's TransactionTemplate didn't have executeWithoutResult method.

lijun-chen-sonarsource: I'd suggest adding this information in the final commit message.

asya-vorobeva:
> LGTM!
> 
> I've seen your counterarguments to Gitar's comments. I think they make sense. Do you think it'd make sense to add test cases that make explicit that these two cases are FN and FP that we intentionally accept?

I think no. These cases are super-low, there's no need to document them.

PR #6052, opened 2026-08-27
SONARJAVA-6687 S2129 Java: add BigDecimal(String) to exceptions

asya-vorobeva:
BigDecimal.valueOf has no String overload, so new BigDecimal(String) has no equivalent alternative and should not be flagged.


lijun-chen-sonarsource (approved):
LGTM!

PR #6056, opened 2026-08-28
SONARJAVA-6870: Fix FP in S9342 when zip stream is wrapped in custom OutputStream

romainbrenguier:
## Summary
- Add test example for S9342 where `ObjectMapper.writeValue` writes to a custom `OutputStream` wrapping a `ZipOutputStream`
- Fix `clearTrackedSymbolsUsedAsArguments` to deeply scan non-identifier arguments (e.g., `new CustomOutputStream(zos)`) for tracked symbol usage, preventing false positives when the zip stream is nested inside a constructor call within a method argument

## Test plan
- [x] Existing `EmptyArchiveEntryCheckTest` passes (both with and without semantic)
- [ ] CI passes

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
@@ -163,18 +163,31 @@ public class EmptyArchiveEntryCheck extends IssuableSubscriptionVisitor {
     return null;
   }
 
+  @CheckForNull
+  private static Symbol extractIdentifierSymbol(ExpressionTree expression) {
+    ExpressionTree unwrapped = ExpressionUtils.skipParentheses(expression);
+    while (unwrapped.is(Tree.Kind.TYPE_CAST)) {
+      unwrapped = ExpressionUtils.skipParentheses(((TypeCastTree) unwrapped).expression());
+    }
+    if (unwrapped.is(Tree.Kind.IDENTIFIER)) {
+      return ((IdentifierTree) unwrapped).symbol();
+    }
+    return null;
+  }
+
   private static void clearTrackedSymbolsUsedAsArguments(MethodInvocationTree mit, Map<Symbol, MethodInvocationTree> pendingEntries) {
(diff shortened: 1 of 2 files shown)


java-checks/src/main/java/org/sonar/java/checks/EmptyArchiveEntryCheck.java
```diff
@@ -161,10 +164,18 @@ private static NewClassTree resolveNewClass(ExpressionTree expression) {
...
-        Symbol argSymbol = ((IdentifierTree) arg).symbol();
-        pendingEntries.remove(argSymbol);
+      ExpressionTree unwrapped = ExpressionUtils.skipParentheses(arg);
+      while (unwrapped.is(Tree.Kind.TYPE_CAST)) {
+        unwrapped = ExpressionUtils.skipParentheses(((TypeCastTree) unwrapped).expression());
+      }
+      if (unwrapped.is(Tree.Kind.IDENTIFIER)) {
+        pendingEntries.remove(((IdentifierTree) unwrapped).symbol());
+      } else {
+        removeUsedSymbols(unwrapped, pendingEntries);
```
nathsou: While unwrapping casts and parentheses here covers direct arguments of the method invocation, `TrackedSymbolVisitor` (`visitNewClass` and `visitMethodInvocation`) still checks `arg.is(Tree.Kind.IDENTIFIER)` directly without unwrapping casts or parentheses.

Since `TrackedSymbolVisitor.visitIdentifier` is an intentional no-op, tracked symbols nested inside a constructor/method call argument behind a cast or parentheses (e.g., `new CustomOutputStream((OutputStream) zos)`) won't be detected during the AST traversal.

Extracting a shared helper (e.g. `extractIdentifier(ExpressionTree)`) to unwrap casts and parentheses and reusing it in both `clearTrackedSymbolsUsedAsArguments` and `TrackedSymbolVisitor` would ensure nested cases are handled consistently.

nathsou (approved):
LGTM, approved with a minor observation regarding cast/parenthesis unwrapping in `TrackedSymbolVisitor`.

PR #6059, opened 2026-08-28
SONARJAVA-5397 Support the import of custom checkstyle issues

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **External Reports:**
    - Added support for importing custom checkstyle issues without the standard prefix in `CheckstyleXmlReportReader.java`


vdiez (approved):
The implementation matches SONARJAVA-5397 and the linked customer report: it preserves the shortened IDs for built-in Checkstyle rules while allowing custom rule IDs through. I found no functional blocker.

Non-blocking suggestion: add regression coverage for an `<error>` with a missing or empty `source`. The guard added in `dd442c5` is currently uncovered, even though the earlier review requested both the guard and a fixture. Placing such an element before a valid custom issue and asserting that import continues would protect this error-recovery behavior.

Validated locally with `mvn -pl external-reports test -Dsonar.skip=true` (30 tests passed); all PR checks are green.

PR #6061, opened 2026-08-28
SONARJAVA-6876 S2479 Reduce noise by suppressing on strings that are most likely generated

lijun-chen-sonarsource:
Part of AT-82

----
## Summary by Gitar

- **New helper:**
    - Added `GeneratedStringLiteralRecognizer` to suppress string literals likely generated in Kotlin annotations
- **Check updates:**
    - Updated `ControlCharacterInLiteralCheck` to ignore recognized generated string literals


java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private static boolean isAnnotationType(AnnotationTree annotation, String fullyQualifiedName) {
+    TypeTree annotationTypeTree = annotation.annotationType();
+    Type annotationType = annotationTypeTree.symbolType();
+    if (!annotationType.isUnknown()) {
+      return annotationType.is(fullyQualifiedName);
+    }
+
+    String annotationName = ExpressionsHelper.concatenate((ExpressionTree) annotationTypeTree);
```
asya-vorobeva: We can simplify, there's no need to be so precise. In case when semantic is absent we can simply check for 
`annotationTree.symbolType().name().equals(unqualifiedName)`

java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
```diff
@@ -0,0 +1,102 @@
...
+    if (fullyQualifiedName.equals(annotationName)) {
+      return true;
+    }
+
+    String unqualifiedName = fullyQualifiedName.substring(fullyQualifiedName.lastIndexOf('.') + 1);
+    return annotationName.equals(unqualifiedName)
+      && hasExplicitImport(annotation, fullyQualifiedName);
+  }
+
+  private static boolean hasExplicitImport(Tree tree, String fullyQualifiedName) {
```
asya-vorobeva: Correspondently we can remove this method

java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
```diff
@@ -0,0 +1,102 @@
...
+import org.sonar.plugins.java.api.tree.TypeTree;
+
+/**
+ * Recognizes strings that are most likely not hand-written and maintained by developers.
+ *
+ * Currently we ignore:
+ *
+ * <ul>
+ *   <li>{@code d1} strings in {@code kotlin.Metadata} annotations:
+ *     <a href="https://github.com/Blaze209/phantom-sonar/blob/4297720bd3afef6bf43accd472f34cd732819bfb/src/main/java/app/rive/Fit.java#L11">example</a></li>
```
asya-vorobeva: I understand that these repos are public, but not sure that it's good to add them here. What if they expire / become private? It's ok to add any mock example, e.g. 
```
@kotlin.Metadata(
   d1 = {"\u0000\u0014\n\u0002\u0018\u0002\n\u0002\u0010\u0000\n\u0002\b\u0002\n\u0002\u0010\u0002\n\u0000\u0008\u00c6\u0002\u0018\u00002\u00020\u0001B\u0007\b\u0002\u00a2\u0006\u0002\u0010\u0002J\u0006\u0010\u0003\u001a\u00020\u0004\u00a8\u0006\u0005"},
   d2 = {"Lcom/example/MyService;", "", "()V", "execute", "", "app_mainRelease"}
)
public final class MyService {
   // ...
}
```

java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
```diff
@@ -0,0 +1,102 @@
...
+public final class GeneratedStringLiteralRecognizer {
+
+  private static final String KOTLIN_METADATA = "kotlin.Metadata";
+  private static final String KOTLIN_SOURCE_DEBUG_EXTENSION = "kotlin.jvm.internal.SourceDebugExtension";
+  private static final String KOTLIN_DEBUG_METADATA = "kotlin.coroutines.jvm.internal.DebugMetadata";
+
+  private GeneratedStringLiteralRecognizer() {
+  }
+
+  public static boolean isGenerated(LiteralTree literal) {
```
asya-vorobeva, line 53: Add Javadoc for this method.

java-checks/src/test/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizerTest.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.sonar.java.checks.verifier.CheckVerifier;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+import static org.sonar.java.checks.verifier.TestUtils.mainCodeSourcesPath;
+
+class GeneratedStringLiteralRecognizerTest {
```
asya-vorobeva, line 27: Typically when we implement utils class test, we use JParserTestUtils.parse method. See e.g. how it's done in TryCatchUtilsTest.

asya-vorobeva (changes requested):

java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
lijun-chen-sonarsource: The purpose of the links was to show that there is indeed java code that has this kind of code.
The mock examples are not that useful IMO, since the test themselves are already mock examples.
If you prefer I can remove the links.

asya-vorobeva: It's understood, but as I said we can't rely on the side repos. So I'd prefer to remove, yes.

java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
lijun-chen-sonarsource: I prefer to be precise. I don't want to recognize any arbitrary usage of `@Metadata` and I think the cost of being precise here is neither high from a computational PoV nor from a maintenance PoV.

But no strong opinion. Addressed in b70e237daa

java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
lijun-chen-sonarsource: Same as above. Addressed in b70e237daa

java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
lijun-chen-sonarsource: Addressed in 684f8ddf26

java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
lijun-chen-sonarsource: Addressed in aa0e9c3559

java-checks/src/test/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizerTest.java
lijun-chen-sonarsource: Addressed in 9dcc751e36

asya-vorobeva (approved):
💯

java-checks/src/main/java/org/sonar/java/checks/helpers/GeneratedStringLiteralRecognizer.java
```diff
@@ -0,0 +1,82 @@
...
+        && "d1".equals(ExpressionUtils.annotationAttributeName(argument))
+        && isAnnotationType(annotation, KOTLIN_METADATA));
+  }
+
+  private static boolean isAnnotationType(AnnotationTree annotation, String fullyQualifiedName) {
+    Type annotationType = annotation.symbolType();
+    String unqualifiedName = fullyQualifiedName.substring(fullyQualifiedName.lastIndexOf('.') + 1);
+    return annotationType.is(fullyQualifiedName)
+      || (annotationType.isUnknown() && annotationType.name().equals(unqualifiedName));
+  }
```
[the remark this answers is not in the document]
lijun-chen-sonarsource: `withoutSemantic()` does not disable ECJ bindings; binding resolution and recovery remain enabled. `ControlCharacterInLiteralCheckTest.test_without_semantic()` already covers this case and passes. I could only produce `Type.UNKNOWN` by manually removing the binding, which is not a realistic analysis path.

PR #6069, opened 2026-09-02
SONARJAVA-6421 Implement S9352: Bean autowiring ambiguity should be resolved using "@Qualifier" or "@Primary"

NoemieBenard:
## Summary
- Implements the S9352 rule (`AmbiguousDependencyCheck`) using the `SpringContextModel`/`TypeToDependenciesIndex` built by the model-extension branch.
- Wires check reporting through `SpringContextModelSensor` and `JavaFrontend`.
- Registers the rule (metadata, Sonar way profile) and adds unit + cross-module integration tests.

Not covered:
- `@Fallback` annotations
- `@Qualifier` annotation on the bean definition
- Beans with profile are excluded without checking if the profile is active or not


NoemieBenard:
Closing to recreate with the correct base after inserting nb/sonarjava-6421-add-profile-support into the stack.

PR #6068, opened 2026-09-02
SONARJAVA-6889 Create new structure for storing all collected dependencies

NoemieBenard:
## Summary
- Introduces `TypeToDependenciesIndex`, extending `SpringContextModel` to track injected dependencies (name + `BeanLocation`) indexed by fully-qualified type.
- Wires `BeanDefinitionGatherer` to populate the new index alongside the existing `TypeToBeanNamesIndex`.

PR #6071, opened 2026-09-02
SONARJAVA-6421 Extract @Profile expressions when gathering bean definitions

NoemieBenard:
## Summary
- `BeanDefinitionGatherer` now extracts `@Profile` expressions (single or array-valued) from stereotype-annotated classes and `@Bean` methods, storing them on `BeanDefinitionHolder` via the existing `profiles(...)` builder step.
- For `@Bean` methods, the method's own `@Profile` takes precedence over the one declared on the enclosing `@Configuration`/`@Component` class; if the method has none, it inherits the class's.
- Updates the bean-definition cache serialization format to persist the new field (`profiles` inserted right after `isPrimary`).


java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionGatherer.java
```diff
@@ -268,9 +283,12 @@ private static BeanData deserializeBean(String line, InputFile inputFile) {
       Integer.parseInt(spanParts[2]),
       Integer.parseInt(spanParts[3]));
     boolean isPrimary = Boolean.parseBoolean(fields[4]);
+    String profiles = !fields[5].isEmpty()
```
asya-vorobeva: This serialization-deserialization becomes more and more complex. I'd say we should think about using some format, JSON or maybe something binary like Protobuf. Let's create JIRA ticket as follow-up for this in epic https://sonarsource.atlassian.net/browse/SONARJAVA-6237.

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionGatherer.java
```diff
@@ -422,6 +442,42 @@ private static String extractQualifier(SymbolMetadata metadata) {
       .orElse(null);
   }
 
+  @Nullable
```
asya-vorobeva: The same as for PR https://github.com/SonarSource/sonar-java/pull/6068. After extracting this method please add comment.

asya-vorobeva (changes requested):
@Profile annotation luckily does not evaluate SpEL expressions. But it supports syntax like this:
`@Profile("dev & !test")`

More precisely, it supports `!, &, | `operators.
Would be great to add support for such evaluation. But I'd suggest to do it in separate PR.

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionGatherer.java
NoemieBenard: I created this ticket: https://sonarsource.atlassian.net/browse/SONARJAVA-6896

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDefinitionGatherer.java
```diff
@@ -471,6 +491,46 @@ private static String extractQualifier(SymbolMetadata metadata) {
       .orElse(null);
   }
 
+
+  /** Reads the {@code @Profile} annotation's "value" attribute, joining every profile name it lists. */
+  @Nullable
+  private static String extractProfiles(SymbolMetadata metadata) {
```
asya-vorobeva: The same as for PR https://github.com/SonarSource/sonar-java/pull/6068. Instead of making this class  more and more unreadable, let's extract these methods and related machinery to SpringUtils. And also let's make documentation standard. Moreover, another benefit of extraction is that you can test methods independently in SpringUtilsTest class (which also will provide additional documentation).

asya-vorobeva (changes requested):

java-frontend/src/main/java/org/sonar/java/utils/SpringUtils.java
```diff
@@ -171,4 +173,24 @@ public static String extractQualifier(SymbolMetadata metadata) {
...
   }
 
+  /**
+   * Reads the {@code @Profile} annotation's "value" attribute, joining every profile name it lists with ",".
+   *
+   * @param metadata The symbol metadata of the class or {@code @Bean} method to check for a {@code @Profile}
+   * @return The joined profile expression, or {@code null} if none is declared
+   */
+  @Nullable
+  public static String extractProfiles(SymbolMetadata metadata) {
```
asya-vorobeva, line 257: Add tests for this method in SpringUtilsTest (covering all the cases from files). Then we can probably reduce the number of added test case files below.

java-frontend/src/test/java/org/sonar/java/model/springcontext/BeanDefinitionGathererTest.java
```diff
@@ -145,6 +145,76 @@ void non_primary_bean_has_isPrimary_false() {
...
+      Arguments.of("src/test/files/springcontext/ProfiledConfigurationWithBeanMethods.java", "inheritedProfileBean", "prod"),
+      // @Bean method's own @Profile is combined (AND-ed) with the enclosing class's one, not overriding it
+      Arguments.of("src/test/files/springcontext/ProfiledConfigurationWithBeanMethods.java", "ownProfileBean", "prod;test"),
+      // @Bean method's own @Profile is kept as-is when the enclosing class has none
+      Arguments.of("src/test/files/springcontext/ConfigurationWithBeanMethods.java", "methodOnlyProfileBean", "test")
+    );
+  }
+
+  @Test
+  void leaveFile_writes_profile_to_cache() {
```
asya-vorobeva: I think we can reduce number of tests for leaveFile to one providing one test file covering all the things.

The same relates to scanWithoutParsing test.

asya-vorobeva (changes requested):

asya-vorobeva (approved):
💯

PR #6076, opened 2026-09-02
SONARJAVA-6892: S8989 Fix FP for class-level @Transactional when no transaction occurs

romainbrenguier:
## Summary
- When `@Transactional` is at the class level, only raise S8989 when the method likely performs a transactional operation
- A method is considered transactional if its name or the name of any method it calls starts with: `save`, `delete`, `update`, `persist`, `merge`, or `flush`
- Methods with method-level `@Transactional` are unaffected and always raise as before

## Test plan
- [x] Existing tests for method-level `@Transactional` still pass
- [x] Class-level `@Transactional` with transactional method names (e.g., `saveOrder`, `deleteRecord`) raises issues
- [x] Class-level `@Transactional` with non-transactional method names (e.g., `noConfig`, `readData`) no longer raises
- [x] Class-level `@Transactional` with transactional method calls (e.g., calling `saveOrder()` inside body) raises issues
- [x] `test_without_semantic` and `test_noDependencies` tests pass

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodCheckedExceptionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodCheckedExceptionCheck.java
@@ -42,10 +42,12 @@ import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
 import org.sonar.plugins.java.api.tree.ClassTree;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.LambdaExpressionTree;
 import org.sonar.plugins.java.api.tree.LiteralTree;
 import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
 import org.sonar.plugins.java.api.tree.MethodInvocationTree;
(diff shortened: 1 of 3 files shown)


java-checks/src/main/java/org/sonar/java/checks/spring/TransactionalMethodCheckedExceptionCheck.java
```diff
@@ -274,6 +286,41 @@ private static boolean hasRollbackConfiguration(AnnotationTree annotation) {
       });
   }
 
+  private static boolean looksTransactional(MethodTree method) {
+    if (hasTransactionalPrefix(method.simpleName().name())) {
+      return true;
+    }
+    var visitor = new MethodInvocationVisitor();
+    method.accept(visitor);
```
nathsou, line 295: `method.accept(visitor)` also traverses lambda bodies, anonymous/local-class methods, and other nested executable scopes. Consequently, a method that only creates a callback containing `repository.save()` can be classified as transactional even when that callback is never executed within the transaction. Please consider limiting this visitor to invocations executed directly by the analyzed method, skipping nested classes and deferred lambda bodies. This is non-blocking for this PR.

nathsou (approved):
Approved. I left one non-blocking false-positive concern about traversing deferred and nested executable bodies.

PR #6080, opened 2026-09-04
SONARJAVA-6898 S5786/S5810: stop flagging class modifiers to avoid FPs

asya-vorobeva:
Both rules previously attempted to flag the visibility modifier of the test class itself, but this produced false positives for classes that are legitimately public — for example, when referenced in a @Suite or extended from another package.

The fix removes class-modifier flagging from both rules entirely:
- S5786 now only reports redundant public/protected on test methods, never on the class declaration.
- S5810 now raises issues only for test methods. Exception is private @Nested class.

### DIFF
--- a/java-checks-test-sources/default/src/test/java/checks/tests/JUnit5SilentlyIgnoreClassAndMethodCheckSample.java
+++ b/java-checks-test-sources/default/src/test/java/checks/tests/JUnit5SilentlyIgnoreClassAndMethodCheckSample.java
@@ -44,10 +44,16 @@ class JUnit5SilentlyIgnoreClassAndMethodCheckSample {
   }
 
   @Nested
-  private class PrivateWithOneTest { // Noncompliant {{Remove this 'private' modifier.}}
-//^^^^^^^
+  private class PrivateWithOneTest { // Compliant - test method is already package-private, class may legitimately need to be private
     @Test
     void test() {}
+  }
+
+  @Nested
+  private class PrivateWithPrivateMethod { // Compliant - class modifier is not flagged
+    @Test
+    private void test() {} // Noncompliant {{Remove this 'private' modifier.}}
+//  ^^^^^^^
   }
 
   @Nested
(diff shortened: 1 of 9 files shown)


java-checks-test-sources/default/src/test/java/checks/tests/JUnit5SilentlyIgnoreClassAndMethodCheckSample.java
```diff
@@ -50,6 +50,13 @@ private class PrivateWithOneTest { // Noncompliant {{Remove this 'private' modif
     void test() {}
   }
 
+  @Nested
+  private class PrivateWithPrivateMethod { // Compliant - class modifier is not flagged, only the method is
```
lijun-chen-sonarsource: IMO this is not desirable.
Users will first fix the issue on the method and then forget to remove the modifier also on the class and then the next analysis will flag the issue on the class.

lijun-chen-sonarsource (changes requested):
LGTM for S5786, but I'm not sure about changes to S5810.

---

> Both rules previously attempted to flag the visibility modifier of the test class itself, but this produced false positives for classes that are legitimately public
> ...
> S5810 now skips the class-level private-modifier check, for the same reason. Exception is private @nested class.

IIUC you're saying that in some cases we shouldn't raise on `public` modifiers for classes. But how does this relate to "skips the class-level private-modifier check"?

java-checks-test-sources/default/src/test/java/checks/tests/JUnit5SilentlyIgnoreClassAndMethodCheckSample.java
asya-vorobeva: Fixed

asya-vorobeva:
> LGTM for S5786, but I'm not sure about changes to S5810.
> 
> > Both rules previously attempted to flag the visibility modifier of the test class itself, but this produced false positives for classes that are legitimately public
> > ...
> > S5810 now skips the class-level private-modifier check, for the same reason. Exception is private @nested class.
> 
> IIUC you're saying that in some cases we shouldn't raise on `public` modifiers for classes. But how does this relate to "skips the class-level private-modifier check"?

Updated PR description.

lijun-chen-sonarsource (approved):
LGTM!

PR #6086, opened 2026-09-07
SONARJAVA-6906 - Modify S8924 to raise only on inconsistent Mockito import usage

NoemieBenard:
## Summary
- S8924 now only raises an issue when a file mixes prefixed `Mockito.xxx()` calls with at least one static import of a Mockito method, instead of flagging every prefixed call unconditionally.
- Files that consistently use the `Mockito.` prefix throughout, without any static import of a Mockito method, are no longer flagged.
- Added `MockitoStaticImportCheckNoStaticImportSample.java` and a corresponding test case covering the fully-prefixed, no-static-import scenario.

----
## Summary by Gitar

- **Rule renaming and refactoring:**
    - Renamed `CompilationOrPreparationInLoopCheck` to `PreparedStatementInsideLoopCheck` (rule S9142) and removed regex checks.
- **Documentation updates:**
    - Updated rule S9142 metadata, title, HTML description, and resources.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/tests/MockitoStaticImportCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/tests/MockitoStaticImportCheck.java
@@ -50,6 +50,7 @@ public class MockitoStaticImportCheck extends IssuableSubscriptionVisitor {
     .build();
 
   private Set<String> conflictingImportedNames = new HashSet<>();
+  private boolean hasStaticMockitoImport;
   private final Deque<Set<String>> classMethodsStack = new ArrayDeque<>();
 
   @Override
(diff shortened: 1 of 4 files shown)


java-checks/src/main/java/org/sonar/java/checks/tests/MockitoStaticImportCheck.java
```diff
@@ -76,9 +77,14 @@ public void leaveNode(Tree tree) {
   }
 
   private void collectConflictingImports(CompilationUnitTree compilationUnitTree) {
-    conflictingImportedNames = compilationUnitTree.imports().stream()
+    List<String> staticImportFqns = compilationUnitTree.imports().stream()
       .filter(clause -> clause instanceof ImportTree importTree && importTree.isStatic())
       .map(clause -> ExpressionsHelper.concatenate((ExpressionTree) ((ImportTree) clause).qualifiedIdentifier()))
+      .toList();
+
+    hasStaticMockitoImport = staticImportFqns.stream().anyMatch(fqn -> fqn.startsWith(MOCKITO_IMPORT_PREFIX));
```
nathsou, line 85: [P2] Only count static Mockito method imports when enabling the mixed-usage check. This predicate treats any `org.mockito.Mockito.*` import as evidence of static method usage, including field imports such as `import static org.mockito.Mockito.RETURNS_DEEP_STUBS;`. A file that consistently calls `Mockito.mock(...)` but imports only that constant is then reported, although it does not mix prefixed and statically imported Mockito methods. Consider excluding explicit field imports (or matching the supported method names) and adding a regression test.

nathsou (approved):
Approved with the inline finding noted below.

PR #6096, opened 2026-09-08
SONARJAVA-6833: S2699 - Fix false positives for disabled tests, Java assert, and AssertJ methods

asya-vorobeva:
- Add missing AssertJ assertion method prefixes (accepts, matches, startsWith) to ASSERTJ_ASSERTION_METHODS_PREDICATE so methods like accepts(), startsWith(), and matches() are recognized when called on AssertJ assertion types
- Skip assertion checking for test methods annotated with @Disabled (JUnit 5) or @Ignore (JUnit 4) since those tests are never executed
- Recognize Java built-in assert statement as a valid assertion in AbstractAssertionVisitor, fixing FPs for both S2699 and S6103


java-checks/src/main/java/org/sonar/java/checks/tests/AssertionsInTestsCheck.java
```diff
@@ -96,6 +97,21 @@ public void visitMethod(MethodTree methodTree) {
...
 
+  @Override
+  public void visitClass(ClassTree classTree) {
+    if (isAnnotatedWithSkippedTestAnnotation(classTree.symbol().metadata())) {
+      // Static nested classes are separate test containers;
+      // JUnit does not propagate the disabled state to them, so we still visit them.
+      classTree.members().stream()
+        .filter(member -> member.is(Tree.Kind.CLASS))
+        .map(ClassTree.class::cast)
+        .filter(nestedClass -> ModifiersUtils.hasModifier(nestedClass.modifiers(), Modifier.STATIC))
```
victor-peixoto-sonarsource, line 108: nit. reuses existing method that does the same thing, plus catches implicit static records (that probably never happen in real projects, but I am leaving this as non-blocking suggestion).

```suggestion
       .filter(member -> member instanceof ClassTree)
       .map(ClassTree.class::cast)
       .filter(nestedClass -> nestedClass.symbol().isStatic())
```

victor-peixoto-sonarsource (approved):
LGTM

java-checks/src/main/java/org/sonar/java/checks/tests/AssertionsInTestsCheck.java
asya-vorobeva: The difference is that ModifiersUtils.hasModifier is purely syntactic — it doesn't depend on symbol resolution — which makes it slightly more robust. Say it will be solved when we use scanner CLI to run analysis (so don't have needed semantic information).

PR #6114, opened 2026-09-11
SONARJAVA-6896 Improve the serialization format for Spring context caching

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **Caching enhancements:**
    - Migrated Spring context cache serialization format from custom delimited strings to structured `Gson` JSON objects with versioning
    - Added comprehensive validation and robust error handling for cached metadata entries and component scan packages


java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextCacheHelper.java
```diff
@@ -118,137 +118,242 @@ static <T> Optional<T> readFromCache(InputFileScannerContext context, Logger log
...
-        Set<InjectionPoint> points = Arrays.stream(entry.substring(idx + 1).split(DEP_NAMES_SEPARATOR))
-          .map(token -> decodeInjectionPoint(token, inputFile))
-          .collect(Collectors.toCollection(LinkedHashSet::new));
-        injectionPoints.put(typeFqn, points);
+  static Optional<List<String>> readComponentScanPackagesFromCache(InputFileScannerContext context, Logger log) {
+    var cacheKey = cacheKey(COMPONENT_SCAN_CACHE_KEY_PREFIX, context);
+    return readFromCache(context, log, cacheKey, SpringContextCacheHelper::deserializePackages);
+  }
+
+  private static JsonObject serializeBeans(List<BeanData> beans) {
```
asya-vorobeva: I'd highly recommend to use Gson's TypeAdapter mechanism for serialization/deserialization stuff. It allows us using steaming API (JsonReader / JsonWriter). You'll need to create several adapters for the classes to serialize (actually only two I guess: BeanData and InjectionPoint) and use them in serialize / deserialize methods. The main benefit is that each adapter is self-contained (and you can use its field names are local string literals, not top-level constants). E.g.:
```
class BeanDataAdapter extends TypeAdapter<BeanData> {

    @Override
    public void write(JsonWriter out, BeanData bean) throws IOException {
      out.beginObject();
      out.name("name").value(bean.beanName());
      out.name("type").value(bean.type());
      out.name("package").value(bean.beanPackage());
      // ... etc.
      out.endObject();
    }

    @Override
    public BeanData read(JsonReader in) throws IOException {
      String name = null, type = null, pkg = null;
      // ... locals for all fields
      in.beginObject();
      while (in.hasNext()) {
        switch (in.nextName()) {
          case "name" -> name = in.nextString();
          case "type" -> type = in.nextString();
          case "package" -> pkg = in.nextString();
          // ...
          default -> in.skipValue();
        }
      }
      in.endObject();
      return new BeanData(name, type, pkg, /* ... */);
    }
  }
```

asya-vorobeva (changes requested):
After switching to using TypeAdapter, it makes sense to add tests for SpringContextCacheHelper. Thus we can probably reduce / simplify caching tests in our GathererTest's.

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDataTypeAdapter.java
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * JSON representation of a single cached bean definition.
+ *
+ * <p>The input file of the bean and its {@link BeanData#dependingBeans()} are not serialized: the former is known
+ * from the cache entry being read, and the latter is derived from the dependency injection points.
+ */
+final class BeanDataTypeAdapter extends TypeAdapter<BeanData> {
+
+  private static final String NAME = "name";
```
asya-vorobeva: Let's remove these top-level constants as I've noted in initial comment. It's absolutely acceptable to use string literals as they're local to the adapter. And it's much more simpler to read:
```
 @Override
    public void write(JsonWriter out, BeanData bean) throws IOException {
      out.beginObject();
      out.name("name").value(bean.beanName());
      // ... etc.
      out.endObject();
    }

    @Override
    public BeanData read(JsonReader in) throws IOException {
      String name = null, type = null, pkg = null;
      // ... locals for all fields
      in.beginObject();
      while (in.hasNext()) {
        switch (in.nextName()) {
          case "name" -> name = in.nextString();
          // ...
          default -> in.skipValue();
        }
      }
      in.endObject();
      return new BeanData(name, type, pkg, /* ... */);
    }
```

java-frontend/src/main/java/org/sonar/java/model/springcontext/InjectionPointTypeAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+
+import static org.sonar.java.utils.JsonUtils.readString;
+import static org.sonar.java.utils.JsonUtils.required;
+
+/**
+ * JSON representation of an {@link InjectionPoint} within a cached bean definition.
+ */
+final class InjectionPointTypeAdapter extends TypeAdapter<InjectionPoint> {
+
+  private static final String NAME = "name";
```
asya-vorobeva: The same as for another adapter

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDataTypeAdapter.java
aurelien-coet-sonarsource: Yes, the string literals are local to the adapters, but they are reused at least 3 times in each class, between the `read` and `write` methods, so I would argue that it makes sense to centralize them in static fields, so we only need to modify a single location if we want to change them later (I also suspect Gitar will flag these as code quality issues). WDYT ?

java-frontend/src/main/java/org/sonar/java/serialization/BeanDataTypeAdapter.java
```diff
@@ -0,0 +1,179 @@
...
+  private static final String TYPE = "type";
+  private static final String PACKAGE = "package";
+  private static final String SPAN = "span";
+  private static final String PRIMARY = "primary";
+  private static final String PROFILES = "profiles";
+  private static final String DEPENDENCIES = "dependencies";
+  private static final String INJECTION_POINTS = "injectionPoints";
+  private static final String TYPE_HIERARCHY = "typeHierarchy";
+
+  private final InputFile inputFile;
```
asya-vorobeva, line 57: We use this field only in read method. I think that we can remove it and use as a parameter of read method. It will allow us to create one static adapter in CacheHelper class and thus reduce number of created objects.

java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextCacheHelper.java
```diff
@@ -118,137 +105,52 @@ static <T> Optional<T> readFromCache(InputFileScannerContext context, Logger log
...
+   * @param beans   The beans collected from this file.
    */
   static void writeBeanDefinitionsToCache(JavaFileScannerContext context, Logger log, List<BeanData> beans) {
-    var cacheKey = cacheKey(BEAN_CACHE_KEY_PREFIX, context);
-    var data = beans.stream()
-      .map(SpringContextCacheHelper::serializeBean)
-      .collect(Collectors.joining(BEAN_SEPARATOR));
-    writeToCache(context, log, cacheKey, data);
+    var adapter = new BeanDataTypeAdapter(context.getInputFile());
+    var document = newDocument(CACHE_FORMAT_VERSION);
```
asya-vorobeva: Here it makes sense to also use writers as we did in adapter classes. Something like this:

```
var out = new StringWriter();
try (var writer = new JsonWriter(out)) {
  writer.beginObject();
  writer.name("version").value(CACHE_FORMAT_VERSION);
  writer.name("beans").beginArray();
  for (BeanData bean : beans) {
    BEAN_ADAPTER.write(writer, bean);
  }
  writer.endArray();
  writer.endObject();
} catch (IOException e) {
  throw new IllegalStateException("Failed to serialize beans", e);
}
writeToCache(context, log, cacheKey(BEAN_CACHE_KEY_PREFIX, context), out.toString());
```

java-frontend/src/main/java/org/sonar/java/serialization/TextSpanTypeAdapter.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.gson.stream.JsonWriter;
+import java.io.IOException;
+
+import static org.sonar.java.utils.JsonUtils.readInt;
+import static org.sonar.java.utils.JsonUtils.required;
+
+/**
+ * JSON representation of the source location of a cached element.
+ */
+public final class TextSpanTypeAdapter extends TypeAdapter<AnalyzerMessage.TextSpan> {
```
asya-vorobeva, line 35: It looks more natural to have all adapters stored in one package.

java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextCacheHelper.java
```diff
@@ -118,137 +105,52 @@ static <T> Optional<T> readFromCache(InputFileScannerContext context, Logger log
...
-      encodedName,
-      bean.type(),
-      bean.beanPackage(),
-      span.startLine + ":" + span.startCharacter + ":" + span.endLine + ":" + span.endCharacter,
-      Boolean.toString(bean.isPrimary()),
-      encodedProfiles,
-      deps,
-      typeHierarchy);
+  static void writeComponentScanPackagesToCache(InputFileScannerContext context, Logger log, Collection<String> packages) {
+    var document = newDocument(CACHE_FORMAT_VERSION);
```
asya-vorobeva: The same as before, using writers.

java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextCacheHelper.java
```diff
@@ -118,137 +105,52 @@ static <T> Optional<T> readFromCache(InputFileScannerContext context, Logger log
...
+   * @param log     Logger of the calling gatherer, used to trace ignored duplicate writes.
+   * @param beans   The beans collected from this file.
    */
   static void writeBeanDefinitionsToCache(JavaFileScannerContext context, Logger log, List<BeanData> beans) {
-    var cacheKey = cacheKey(BEAN_CACHE_KEY_PREFIX, context);
-    var data = beans.stream()
-      .map(SpringContextCacheHelper::serializeBean)
-      .collect(Collectors.joining(BEAN_SEPARATOR));
-    writeToCache(context, log, cacheKey, data);
+    var adapter = new BeanDataTypeAdapter(context.getInputFile());
```
asya-vorobeva, line 114: As it was said before, we can create one static adapter BEAN_ADAPTER and use it in all the places.

asya-vorobeva (changes requested):
Also please add JavaDoc to all new public methods.

java-frontend/src/main/java/org/sonar/java/model/springcontext/BeanDataTypeAdapter.java
asya-vorobeva: Actually we don't want to modify them. That's the point. :)

Using static fields is actual when you use it across multiple files, but it's not our case.

java-frontend/src/main/java/org/sonar/java/serialization/BeanDataTypeAdapter.java
aurelien-coet-sonarsource: It would have made sense, but we cannot add the `InputFile` as a parameter of the `read` method, because it is an override to satisfy the `TypeAdapter<T>` interface.

asya-vorobeva: Yep, I see. But we can use overloaded version of read (and probably fromJsonTree). WDYT?

asya-vorobeva (approved):
Great work! 🚀

PR #6121, opened 2026-09-11
SONARJAVA-6943: Implement rule S9386 Float literals should be used instead of casting double literals to float

romainbrenguier:
## Summary
- Implement new rule S9386 that detects explicit type casts of double literals to `float` (e.g. `(float) 3.14`) and suggests using a float literal suffix instead (`3.14f`)
- Includes a quick fix to automatically replace the cast expression with the equivalent float literal
- Handles scientific notation, hex floating-point literals, `d`/`D` suffixed literals, and parenthesized expressions

## Test plan
- [x] Unit tests pass with semantic analysis
- [x] Unit tests pass without semantic analysis
- [x] Check-list generator correctly discovers the new rule
- [x] Plugin metadata validation tests pass
- [ ] Ruling tests (will be validated by CI)

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/CastDoubleToFloatCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/CastDoubleToFloatCheck.java
@@ -26,8 +26,10 @@ import org.sonar.java.reporting.JavaTextEdit;
 import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
 import org.sonar.plugins.java.api.tree.ExpressionTree;
 import org.sonar.plugins.java.api.tree.LiteralTree;
+import org.sonar.plugins.java.api.tree.PrimitiveTypeTree;
 import org.sonar.plugins.java.api.tree.Tree;
 import org.sonar.plugins.java.api.tree.TypeCastTree;
+import org.sonar.plugins.java.api.tree.UnaryExpressionTree;
 
 @Rule(key = "S9386")
 public class CastDoubleToFloatCheck extends IssuableSubscriptionVisitor {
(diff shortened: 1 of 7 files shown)


java-checks/src/main/java/org/sonar/java/checks/CastDoubleToFloatCheck.java
```diff
@@ -0,0 +1,109 @@
...
+  public void visitNode(Tree tree) {
+    TypeCastTree typeCastTree = (TypeCastTree) tree;
+    if (!isFloatCast(typeCastTree)) {
+      return;
+    }
+    ExpressionTree expression = ExpressionUtils.skipParentheses(typeCastTree.expression());
+    boolean negated = false;
+    if (expression.is(Tree.Kind.UNARY_MINUS, Tree.Kind.UNARY_PLUS)) {
+      negated = expression.is(Tree.Kind.UNARY_MINUS);
+      expression = ((UnaryExpressionTree) expression).expression();
```
nathsou: If the operand of the unary minus or plus is parenthesized (e.g. `(float) -(0.5)` or `(float) +(3.14)`), `((UnaryExpressionTree) expression).expression()` returns a `ParenthesizedExpressionTree`. As a result, `expression.is(Tree.Kind.DOUBLE_LITERAL)` returns `false` and the issue is missed.

Wrapping the inner operand with `ExpressionUtils.skipParentheses` ensures parenthesized negated literals are caught:
```suggestion
      expression = ExpressionUtils.skipParentheses(((UnaryExpressionTree) expression).expression());
```

java-checks/src/main/java/org/sonar/java/checks/CastDoubleToFloatCheck.java
```diff
@@ -0,0 +1,109 @@
...
+      if (!Float.isFinite(asFloat)) {
+        return false;
+      }
+      if (asFloat == 0.0f && asDouble != 0.0) {
+        return false;
+      }
+      if (negated) {
+        asDouble = -asDouble;
+        asFloat = -asFloat;
+      }
```
nathsou: Direct equality comparison with `==` on floating-point values can trigger Sonar rule S1244 ("Floating-point numbers should not be tested for equality") and treats `+0.0f == -0.0f` as `true` (though here both `asFloat` and `asDouble` are given the same sign).

Consider using `Float.compare(asFloat, (float) asDouble) == 0` or `Float.floatToIntBits(asFloat) == Float.floatToIntBits((float) asDouble)` for a more explicit and precise check.

nathsou (approved):
LGTM! Approved with two minor suggestions:
1. Unwrap parentheses on the inner unary expression (`ExpressionUtils.skipParentheses`) so parenthesized negated literals like `(float) -(0.5)` are detected.
2. Consider using `Float.compare(asFloat, (float) asDouble) == 0` instead of `==` in `isEquivalentFloatLiteral`.

PR #6123, opened 2026-09-11
SONARJAVA-6951: Implemented rule S9390 - Manual null checks should be replaced with Objects.equals()

romainbrenguier:
Implements rule S9390 which detects ternary expressions performing redundant null-safe equality checks.

The pattern  is semantically equivalent to  but more verbose and error-prone.

Changes:
- Added  - the rule implementation using 
- Added  - test class with semantic and non-semantic tests
- Added  - comprehensive test cases with semantic analysis
- Added  - test cases for non-semantic analysis

The rule detects:
-  → suggests 
-  → suggests 
- Qualified identifiers: 

Accepted false negatives:
- Method calls: 
- Complex expressions with  or  in the false branch


nathsou (approved):
Looks good!

PR #6124, opened 2026-09-11
SONARJAVA-6944: Implement rule S9387 TestNG Javadoc tags should be converted to annotations

romainbrenguier:
## Summary
- Implement new rule S9387 that detects TestNG-specific Javadoc tags (`@test`, `@beforeMethod`, `@afterMethod`, `@beforeClass`, `@afterClass`, `@dataProvider`, etc.) used as configuration markers instead of proper TestNG annotations
- Case-insensitive matching for all 15 TestNG tags with secondary locations when multiple tags appear on a single method
- Includes test sample with compliant and noncompliant examples covering edge cases (case variants, mixed standard/TestNG tags, non-Javadoc comments)

## Test plan
- [x] Unit tests pass (with and without semantic analysis)
- [ ] CI build passes
- [ ] Ruling tests updated if needed (auto-PR)

## Agent workflow

Addressed review comments in pr_report_6124.md using `uv run address_reviews.py pr_report_6124.md`

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/tests/TestNGJavadocTagsCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/tests/TestNGJavadocTagsCheck.java
@@ -1,0 +1,101 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.checks.tests;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.JavaFileScannerContext;
+import org.sonar.plugins.java.api.tree.SyntaxTrivia;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonarsource.analyzer.commons.collections.MapBuilder;
+
+@Rule(key = "S9387")
+public class TestNGJavadocTagsCheck extends IssuableSubscriptionVisitor {
+
+  private static final Pattern BLOCK_TAG_PATTERN = Pattern.compile("^\\s*\\*?\\s*@(\\w+)", Pattern.MULTILINE);
+
+  private static final Map<String, String> TESTNG_TAGS_TO_ANNOTATIONS = MapBuilder.<String, String>newMap()
+    .put("test", "@Test")
+    .put("beforemethod", "@BeforeMethod")
+    .put("aftermethod", "@AfterMethod")
+    .put("beforeclass", "@BeforeClass")
+    .put("afterclass", "@AfterClass")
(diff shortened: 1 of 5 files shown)


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9387.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "tags": [
+    "deprecated",
+    "convention"
+  ],
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-9387",
+  "sqKey": "S9387",
+  "scope": "All",
+  "quickfix": "unknown",
```
nathsou, line 17: This is a TestNG test-configuration rule, but `All` also runs it on production methods where tags such as `@factory` may be ordinary documentation. Sibling checks under `checks/tests` use the `Tests` scope. Could we switch this to `Tests` and regenerate the derived metadata/profile files?

java-checks/src/main/java/org/sonar/java/checks/tests/TestNGJavadocTagsCheck.java
```diff
@@ -0,0 +1,108 @@
...
+
+    Matcher matcher = BLOCK_TAG_PATTERN.matcher(textWithoutCodeBlocks);
+
+    String firstTagOriginal = null;
+    String firstAnnotation = null;
+    List<JavaFileScannerContext.Location> secondaryLocations = new ArrayList<>();
+
+    while (matcher.find()) {
+      String tagName = matcher.group(1);
+      String annotation = TESTNG_TAGS_TO_ANNOTATIONS.get(tagName.toLowerCase(Locale.ROOT));
```
nathsou, line 81: Lowercasing makes annotation names indistinguishable from the legacy case-sensitive tags. For example, documentation containing `@Test is the annotation used by TestNG` is treated as obsolete `@test`. Could we match the documented legacy spellings exactly (`@test`, `@beforeMethod`, etc.) and keep uppercase annotation names compliant?

java-checks/src/main/java/org/sonar/java/checks/tests/TestNGJavadocTagsCheck.java
```diff
@@ -0,0 +1,108 @@
...
+import org.sonar.plugins.java.api.tree.SyntaxTrivia;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonarsource.analyzer.commons.collections.MapBuilder;
+
+@Rule(key = "S9387")
+public class TestNGJavadocTagsCheck extends IssuableSubscriptionVisitor {
+
+  private static final Pattern BLOCK_TAG_PATTERN = Pattern.compile("(?:^|/\\*\\*)[\\t ]*+\\*?[\\t ]*+@(\\w+)", Pattern.MULTILINE);
+  private static final Pattern PRE_BLOCK_PATTERN = Pattern.compile("<pre>.*?</pre>", Pattern.DOTALL);
+  private static final Pattern CODE_TAG_PATTERN = Pattern.compile("\\{@code[\\t ][^}]*+\\}");
```
nathsou: This pattern requires a space or tab after `{@code`, so a valid multiline block such as `{@code
 * @Test
 * void example() {}
 * }` is not masked; the inner `@Test` can then be reported as a block tag. Could we support line breaks here (or use structured Javadoc parsing) and add a multiline regression test?

java-checks/src/main/java/org/sonar/java/checks/tests/TestNGJavadocTagsCheck.java
```diff
@@ -0,0 +1,108 @@
...
+    return Arrays.asList(Tree.Kind.METHOD, Tree.Kind.CONSTRUCTOR);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    tree.firstToken().trivias().stream()
+      .filter(trivia -> trivia.isComment(SyntaxTrivia.CommentKind.JAVADOC))
+      .forEach(trivia -> checkJavadoc((MethodTree) tree, trivia));
+  }
+
```
nathsou, line 68: Java 23 Markdown documentation comments are represented as `SyntaxTrivia.CommentKind.MARKDOWN`, so `/// @test` is currently skipped even though it is still a documentation-based declaration. If the rule is intended to cover all Java documentation comments, could we inspect `MARKDOWN` trivia too, as the existing documentation checks do?

nathsou (approved):
The rule is well structured. I left four non-blocking coverage and false-positive concerns.

PR #6125, opened 2026-09-11
SONARJAVA-6942: Implemented rule S9389 - Data provider names should be unique within a class

romainbrenguier:
This PR implements rule S9389 which detects duplicate @DataProvider names within a single TestNG test class.

## Summary

When multiple methods are annotated with  using the same name, TestNG cannot reliably determine which provider to use, leading to unpredictable test behavior. This rule identifies such duplicates and reports them as issues.

## Implementation Details

- **Rule Class**:  extends 
- **Scope**: Checks methods within each class independently (nested classes have separate scopes)
- **Detection**: Uses semantic model to identify  annotations and extracts the  attribute value
- **Reporting**: Reports on the duplicate (second occurrence) with a secondary location pointing to the first occurrence

## Test Coverage

- Duplicate names within the same class (non-compliant)
- Three or more methods with the same name (reports each duplicate)
- Unique names (compliant)
- Methods without explicit name (compliant - uses method name as default)
- Same name in different classes (compliant - different scopes)
- Nested classes with same names (compliant - different scopes)
- No false positives when semantic is unavailable

## Files Changed

-  - Rule implementation
-  - Test class
-  - Test sample
-  - Rule description
-  - Rule metadata

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/tests/DataProviderNameUniquenessCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/tests/DataProviderNameUniquenessCheck.java
@@ -1,0 +1,121 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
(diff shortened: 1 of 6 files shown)


java-checks/src/main/java/org/sonar/java/checks/tests/DataProviderNameUniquenessCheck.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @Nullable
+  private static String extractDataProviderName(MethodTree method) {
+    for (AnnotationTree annotation : method.modifiers().annotations()) {
+      if (isDataProviderAnnotation(annotation)) {
+        ExpressionTree nameExpression = findNameAttribute(annotation);
+        if (nameExpression == null) {
+          return method.simpleName().name();
+        }
+        return nameExpression.asConstant(String.class).orElse(null);
```
nathsou: `@DataProvider(name = "")` needs the same fallback as an omitted name. TestNG replaces an empty name with the method name, whereas this currently returns `""`. This creates both a false positive for two differently named methods that each specify `name = ""`, and a false negative when `foo()` uses `name = ""` while another provider explicitly uses `name = "foo"`. Could we normalize an empty resolved constant to `method.simpleName().name()` and cover both cases?

nathsou (approved):
The implementation looks good overall. I left one non-blocking edge case to address.

PR #6126, opened 2026-09-11
SONARJAVA-6941: Implement rule S9388 TestNG data provider methods should return valid types

romainbrenguier:
## Summary
- Implement rule S9388 that detects `@DataProvider`-annotated methods in TestNG test code returning invalid types
- The rule flags methods whose return type is not `Object[][]`, `Iterator<Object[]>`, or `Object[]`
- Scoped to test sources; requires semantic analysis (no false positives without bytecode)

## Test plan
- [x] Verified with `CheckVerifier` on test sample covering all noncompliant and compliant patterns
- [x] Verified no issues raised without semantic analysis
- [ ] CI passes

## Agent workflow
Addressed review comments in pr_report_6126.md using `uv run address_reviews.py pr_report_6126.md`

Iterated on the PR with `uv run ci_loop.py` for 2 iterations. 

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/S9388Check.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/S9388Check.java
@@ -1,0 +1,87 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.checks;
+
+import java.util.Collections;
+import java.util.List;
+import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S9388")
+public class S9388Check extends IssuableSubscriptionVisitor {
+
+  private static final String DATA_PROVIDER_ANNOTATION = "org.testng.annotations.DataProvider";
+  private static final String MESSAGE = "Change this return type to \"Object[][]\", \"Iterator<Object[]>\", or \"Object[]\".";
+
+  @Override
+  public List<Tree.Kind> nodesToVisit() {
+    return Collections.singletonList(Tree.Kind.METHOD);
+  }
+
+  @Override
+  public void visitNode(Tree tree) {
+    MethodTree methodTree = (MethodTree) tree;
+    if (!methodTree.symbol().metadata().isAnnotatedWith(DATA_PROVIDER_ANNOTATION)) {
+      return;
+    }
+    Type returnType = methodTree.returnType().symbolType();
+    if (returnType.isUnknown()) {
+      return;
+    }
+    if (!isValidDataProviderReturnType(returnType)) {
+      reportIssue(methodTree.returnType(), MESSAGE);
+    }
+  }
+
+  private static boolean isValidDataProviderReturnType(Type returnType) {
+    return isObjectArray2D(returnType)
+      || isObjectArray1D(returnType)
+      || isIteratorOfObjectArray(returnType);
+  }
+
+  private static boolean isObjectArray2D(Type type) {
+    if (!type.isArray()) {
+      return false;
+    }
+    Type elementType = ((Type.ArrayType) type).elementType();
+    return elementType.isArray() && ((Type.ArrayType) elementType).elementType().is("java.lang.Object");
+  }
+
+  private static boolean isObjectArray1D(Type type) {
+    if (!type.isArray()) {
+      return false;
+    }
+    Type elementType = ((Type.ArrayType) type).elementType();
+    return !elementType.isArray() && elementType.is("java.lang.Object");
+  }
+
+  private static boolean isIteratorOfObjectArray(Type type) {
+    if (!type.is("java.util.Iterator") || !type.isParameterized()) {
+      return false;
+    }
+    List<Type> typeArgs = type.typeArguments();
+    if (typeArgs.size() != 1) {
+      return false;
+    }
+    Type typeArg = typeArgs.get(0);
+    return typeArg.isArray() && ((Type.ArrayType) typeArg).elementType().is("java.lang.Object");
+  }
+
+}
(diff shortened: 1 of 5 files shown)


java-checks/src/main/java/org/sonar/java/checks/S9388Check.java
```diff
@@ -0,0 +1,92 @@
...
+      return false;
+    }
+    Type elementType = ((Type.ArrayType) type).elementType();
+    return !elementType.isArray() && elementType.is(JAVA_LANG_OBJECT);
+  }
+
+  private static boolean isValidIterator(Type type) {
+    if (!type.isSubtypeOf("java.util.Iterator")) {
+      return false;
+    }
```
nathsou, line 79: This does not mirror TestNG’s dispatch based on the method’s declared generic return type. Direct `Iterator<String>` is valid in current TestNG (each value is wrapped as a one-argument row), but this rejects it; conversely, a method declared as a concrete `StringIterator implements Iterator<String>` reaches this non-parameterized branch and is accepted, while TestNG treats that declared class like a raw iterator and expects its elements to already be `Object[]` rows. Could we align this logic with TestNG’s reflection behavior and cover both declarations?

java-checks/src/main/java/org/sonar/java/checks/S9388Check.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  private static boolean isValidDataProviderReturnType(Type returnType) {
+    return isObjectArray2D(returnType)
+      || isObjectArray1D(returnType)
+      || isValidIterator(returnType);
+  }
+
+  private static boolean isObjectArray2D(Type type) {
+    if (!type.isArray()) {
```
nathsou, line 61: There appears to be an RSPEC/runtime discrepancy worth clarifying: current TestNG accepts covariant reference arrays such as `String[]` because they are `instanceof Object[]` and wraps each element as one argument, while the RSPEC explicitly presents `String[]` as noncompliant and this exact-component check follows that narrower contract. Could we confirm whether the rule intentionally enforces a style restriction; otherwise the RSPEC and implementation should accept reference arrays assignable to `Object[]` (while still rejecting `int[]`)?

java-checks/src/main/java/org/sonar/java/checks/S9388Check.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.List;
+import org.sonar.check.Rule;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S9388")
+public class S9388Check extends IssuableSubscriptionVisitor {
+
```
nathsou, line 29: Could we give this check a descriptive name, such as `TestNGDataProviderReturnTypeCheck`, and rename the test/sample accordingly? SonarJava checks conventionally use behavior-oriented names rather than `SXXXXCheck`, which makes them easier to discover and maintain.

java-checks/src/main/java/org/sonar/java/checks/S9388Check.java
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.checks;
```
nathsou, line 17: This is test-specific and the companion TestNG rules S9387/S9389 live under `org.sonar.java.checks.tests`. Could we move this implementation and its test to that package as well for consistency?

nathsou (approved):
The rule is a useful addition. I left two non-blocking behavioral concerns and two consistency suggestions.

java-checks/src/main/java/org/sonar/java/checks/tests/TestNGDataProviderReturnTypeCheck.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  private static boolean isValidDataProviderReturnType(Type returnType) {
+    return isObjectArray2D(returnType)
+      || isObjectArray1D(returnType)
+      || isValidIterator(returnType);
+  }
+
+  private static boolean isObjectArray2D(Type type) {
+    if (!type.isArray()) {
```
romainbrenguier: Updated the implementation and test sample to accept covariant reference arrays such as `String[]` and `String[][]`, while keeping primitive arrays such as `int[]`/`int[][]` noncompliant. The RSPEC wording and examples will be updated separately on the rspec repository side to reflect this runtime-compatible contract.

java-checks/src/main/java/org/sonar/java/checks/tests/TestNGDataProviderReturnTypeCheck.java
```diff
@@ -0,0 +1,92 @@
...
+      return false;
+    }
+    Type elementType = ((Type.ArrayType) type).elementType();
+    return !elementType.isArray() && elementType.is(JAVA_LANG_OBJECT);
+  }
+
+  private static boolean isValidIterator(Type type) {
+    if (!type.isSubtypeOf("java.util.Iterator")) {
+      return false;
+    }
```
romainbrenguier: The `isValidIterator` method has been simplified to just `type.isSubtypeOf("java.util.Iterator")`, which now uniformly accepts all Iterator subtypes including `Iterator<String>` and concrete implementations like `StringIterator`. This addresses the inconsistency you raised.

java-checks/src/main/java/org/sonar/java/checks/tests/TestNGDataProviderReturnTypeCheck.java
```diff
@@ -0,0 +1,92 @@
...
+      return true;
+    }
+    List<Type> typeArgs = type.typeArguments();
+    if (typeArgs.size() != 1) {
+      return false;
+    }
+    Type typeArg = typeArgs.get(0);
+    return typeArg.isSubtypeOf(JAVA_LANG_OBJECT)
+      || (typeArg.isArray() && ((Type.ArrayType) typeArg).elementType().isSubtypeOf(JAVA_LANG_OBJECT));
+  }
```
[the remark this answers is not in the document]
romainbrenguier: Addressed: `isValidIterator` has been simplified to `type.isSubtypeOf("java.util.Iterator")`, removing the unreachable type-argument checks entirely.

java-checks-test-sources/default/src/test/java/checks/tests/TestNGDataProviderReturnTypeCheckSample.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  @DataProvider
+  public Stream<String> returnsStream() { // Noncompliant {{Change this return type to "Object[][]", "Iterator<Object[]>", "Iterator<Object>", or "Object[]".}}
+//       ^^^^^^^^^^^^^^
+    return Stream.of("a", "b");
+  }
+
+  @DataProvider
+  public String[] returnsStringArray() { // Compliant - String[] is assignable to Object[] at runtime
```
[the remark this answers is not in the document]
romainbrenguier: Addressed: The HTML documentation has been updated to reflect that any reference-type array and any Iterator subtype are accepted. The third noncompliant example has been changed from `String[]` (which is now compliant) to `int[][]` (a primitive array that is actually noncompliant).

PR #6127, opened 2026-09-11
SONARJAVA-6952 Implement rule S9391: Detect for-loops that can be replaced with streams

romainbrenguier:
# Summary

This PR implements rule S9391 which detects for-each loops performing filter, map, or collect operations that could be simplified using the Java Stream API.

## Changes

- Added  that identifies loops that can be replaced with streams
- Handles both regular for-each loops and loops with if statements as the body
- Supports semantic analysis when available, falls back to name-based heuristics for syntax-only analysis
- Added test cases for both semantic and non-semantic analysis scenarios
- Added rule metadata (S9391.html, S9391.json) and profile configuration

## Patterns Detected

1. **Filter/Map**: Loop with if condition followed by collection.add()
2. **Collect**: Loop with direct collection.add()
3. **FindFirst**: Loop with break statement (not implemented yet)

## Agent workflow

Iterated on the PR with `uv run ci_loop.py` for 3 iterations. 
:heavy_check_mark: : The PR is now ready for review.

### DIFF
--- a/java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
+++ b/java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
@@ -1,0 +1,405 @@
+/*
+ * SonarQube Java
+ * Copyright (C) SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * You can redistribute and/or modify this program under the terms of
+ * the Sonar Source-Available License Version 1, as published by SonarSource Sàrl.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the Sonar Source-Available License for more details.
+ *
+ * You should have received a copy of the Sonar Source-Available License
+ * along with this program; if not, see https://sonarsource.com/license/ssal/
+ */
+package org.sonar.java.checks;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Pattern;
+import org.sonar.check.Rule;
+import org.sonar.java.model.ExpressionUtils;
+import org.sonar.plugins.java.api.IssuableSubscriptionVisitor;
+import org.sonar.plugins.java.api.semantic.Symbol;
+import org.sonar.plugins.java.api.semantic.Type;
+import org.sonar.plugins.java.api.tree.AssignmentExpressionTree;
+import org.sonar.plugins.java.api.tree.BaseTreeVisitor;
+import org.sonar.plugins.java.api.tree.BlockTree;
+import org.sonar.plugins.java.api.tree.BreakStatementTree;
+import org.sonar.plugins.java.api.tree.ContinueStatementTree;
+import org.sonar.plugins.java.api.tree.ExpressionStatementTree;
+import org.sonar.plugins.java.api.tree.ExpressionTree;
+import org.sonar.plugins.java.api.tree.ForStatementTree;
+import org.sonar.plugins.java.api.tree.ForEachStatement;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.IfStatementTree;
+import org.sonar.plugins.java.api.tree.ListTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodInvocationTree;
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.StatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = "S9391")
+public class ForLoopStreamSuggestionCheck extends IssuableSubscriptionVisitor {
+
+  private static final String MESSAGE = "Use a stream instead of this loop.";
+  private static final Set<String> ADD_METHODS = Set.of("add", "addFirst", "addLast", "offer", "offerFirst", "offerLast");
+  private static final Pattern COLLECTION_NAME_PATTERN = Pattern.compile(
+      "^(result|list|set|collection|queue|deque|items|elements|entries|values|keys|children|nodes|names|ids|users|strings)$",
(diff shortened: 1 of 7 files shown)


java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+    return null;
+  }
+
+  private static boolean isCollectionAdd(Set<Symbol> collectionSymbols, ExpressionTree expr) {
+    if (!expr.is(Tree.Kind.METHOD_INVOCATION)) {
+      return false;
+    }
+    MethodInvocationTree mit = (MethodInvocationTree) expr;
+    if (mit.arguments().size() != 1 || !isAddMethod(mit)) {
```
nathsou: Checking only the name and arity also accepts unrelated overloads. A `Collection` subtype can declare `void add(int metric)`, and a loop calling that method would be reported even though it is not `Collection.add(E)` and has no collector equivalent. Could we match the resolved supported collection/queue/deque methods and signatures with `MethodMatchers`?

java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+    return symbols;
+  }
+
+  private static void addIfCollection(Set<Symbol> symbols, VariableTree varTree) {
+    ExpressionTree initializer = varTree.initializer();
+    if (initializer == null) {
+      return;
+    }
+    Type type = initializer.symbolType();
```
nathsou: A collection-typed initializer does not establish that this is a fresh empty result. For `List<String> result = existing;`, replacing the loop with an assigned collected stream discards existing elements and breaks the alias; `new ArrayList<>(existing)` similarly starts pre-populated. Could we restrict candidates to recognized fresh empty collection constructions/factories?

java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
```diff
@@ -0,0 +1,178 @@
...
+      singleStmt = loopBody.body().get(0);
+    } else {
+      singleStmt = body;
+    }
+    Set<Symbol> collectionSymbols = collectCollectionSymbols(forEach);
+    if (collectionSymbols.isEmpty()) {
+      return;
+    }
+
+    if (singleStmt.is(Tree.Kind.EXPRESSION_STATEMENT)) {
```
nathsou, line 87: Could we also exclude the accumulation target when it is the enhanced-for source? `for (String item : items) { items.add(transform(item)); }` is not a collect-into-result pattern, and a stream rewrite would read and mutate the same collection (often causing `ConcurrentModificationException`). Comparing the source and target symbols should avoid this case.

java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
```diff
@@ -0,0 +1,178 @@
...
+      ExpressionTree expr = ((ExpressionStatementTree) singleStmt).expression();
+      if (isCollectionAdd(collectionSymbols, expr)) {
+        reportIssue(forEach.forKeyword(), MESSAGE);
+      }
+    } else if (singleStmt.is(Tree.Kind.IF_STATEMENT)) {
+      IfStatementTree ifStmt = (IfStatementTree) singleStmt;
+      if (isSimpleFilterCollect(collectionSymbols, ifStmt)) {
```
nathsou, line 94: A direct stream rewrite may not compile when the mapped expression or filter condition throws a checked exception. For example, a loop can call `parse(path)` from a method declaring `throws IOException`, but `paths.stream().map(path -> parse(path))` cannot use `Function` without handling that exception. Could we skip patterns containing checked-throwing invocations?

nathsou (approved):
The main loop patterns look good. I left four non-blocking semantic-equivalence edge cases.

java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
romainbrenguier: Good point. Detecting checked exceptions thrown from within lambda-incompatible expressions would add significant complexity. Since this is a non-blocking suggestion rule (the user still decides whether to apply the refactoring), I think we can defer this to a follow-up if we observe FPs in practice.

java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
romainbrenguier: Addressed: switched from name+arity matching to `MethodMatchers.ofSubTypes("java.util.Collection").names("add", "addLast", "offer", "offerLast").addParametersMatcher(ANY)`, combined with `isInheritedCollectionMethod` which verifies the resolved method symbol belongs to a Collection API type (Collection, List, Set, Queue, Deque) or overrides one. This filters out unrelated overloads like `Metrics.add(int metric)` — see test case `nonCollectionAddOverload()`. Also removed redundant BlockingQueue/BlockingDeque from COLLECTION_API_TYPES since they extend Queue/Deque.

java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
romainbrenguier: Addressed: `addIfCollection` now requires the initializer to pass `isFreshEmptyCollection`, which only accepts `new XxxCollection()` expressions where the type is a subtype of `java.util.Collection` and no constructor argument is itself a Collection (to reject copy constructors like `new ArrayList<>(existing)`). Aliased initializers like `List<String> result = existing` are also rejected since they're not `NEW_CLASS` nodes. See test cases `collectFromExistingCollection()` and `collectFromCopiedCollection()`.

java-checks/src/main/java/org/sonar/java/checks/ForLoopStreamSuggestionCheck.java
romainbrenguier: Addressed: added `excludeSourceSymbol` which removes the enhanced-for source collection from candidate symbols before checking the loop body. When `for (String item : items) { items.add(...); }`, the `items` symbol is excluded from `collectionSymbols`, so no issue is raised. See test case `selfModifyingLoop()`.

PR #6135, opened 2026-09-14
SONARJAVA-6950: Implemented rule S9392 - Redundant range checks should be removed

romainbrenguier:
This rule detects redundant range checks on the same variable within a single boolean AND expression. When multiple range comparisons (>, <, >=, <=) reference the same variable, one comparison may be logically implied by another. For example, `x >= 0 && x >= 5` contains a redundant check because `x >= 5` already implies `x >= 0`.

## Agent workflow
Addressed review comments in pr_report_6135.md using `uv run address_reviews.py pr_report_6135.md`

Addressed review comments in pr_report_6135.md using `uv run address_reviews.py pr_report_6135.md`

Addressed review comments in pr_report_6135.md using `uv run address_reviews.py pr_report_6135.md`


java-checks/src/main/java/org/sonar/java/checks/RedundantRangeCheckCheck.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  @Nullable
+  private static Comparison extractComparison(BinaryExpressionTree comparison) {
+    ExpressionTree left = ExpressionUtils.skipParentheses(comparison.leftOperand());
+    ExpressionTree right = ExpressionUtils.skipParentheses(comparison.rightOperand());
+
+    String operator = comparison.operatorToken().text();
+
+    if (left.is(Kind.IDENTIFIER) && right.is(Kind.INT_LITERAL, Kind.LONG_LITERAL)) {
```
nathsou: Signed constants never reach `LiteralUtils`: Java represents `-10` and `+10` as unary expressions, so this kind guard rejects them even though the literal helpers can resolve them. For example, `x >= -10 && x >= -5` should report the first comparison but currently does not. Could we accept unary plus/minus constant operands and add regression tests?

java-checks/src/main/java/org/sonar/java/checks/RedundantRangeCheckCheck.java
```diff
@@ -0,0 +1,264 @@
...
+  private static boolean hasPotentialSideEffects(ExpressionTree expr) {
+    AtomicBoolean found = new AtomicBoolean(false);
+    expr.accept(new BaseTreeVisitor() {
+      @Override
+      public void visitMethodInvocation(MethodInvocationTree tree) {
+        found.set(true);
+      }
+
+      @Override
+      public void visitMethodReference(MethodReferenceTree tree) {
```
nathsou, line 136: This safety check misses expressions that can throw without having conventional side effects. For example, in `x >= 0 && array[0] > 0 && x >= 5`, reporting/removing `x >= 0` makes the array access execute for negative `x`, potentially introducing an NPE or AIOOBE. Division, casts, and unboxing have the same concern. Could we require intervening operands to be safely evaluable, or conservatively split/skip such chains?

java-checks/src/main/java/org/sonar/java/checks/RedundantRangeCheckCheck.java
```diff
@@ -0,0 +1,264 @@
...
+    return collectOperand(left, comparisonsByVariable) && collectOperand(right, comparisonsByVariable);
+  }
+
+  private static boolean collectOperand(ExpressionTree operand, Map<Symbol, List<Comparison>> comparisonsByVariable) {
+    if (operand.is(Kind.CONDITIONAL_AND)) {
+      return collectOperands((BinaryExpressionTree) operand, comparisonsByVariable);
+    }
+    if (hasPotentialSideEffects(operand)) {
+      comparisonsByVariable.clear();
+      return false;
```
nathsou, line 121: Clearing and aborting the whole chain also loses safe findings that precede a side-effect boundary. In `x >= 0 && x >= 5 && refresh()`, the first comparison is redundant and removing it does not change whether `refresh()` runs, but this reports nothing. Could we segment analysis at unsafe operands instead of discarding comparisons already collected before them?

nathsou (approved):
The core implication logic looks sound. I left three non-blocking edge cases for follow-up.

java-checks/src/main/java/org/sonar/java/checks/RedundantRangeCheckCheck.java
romainbrenguier: Addressed: the analysis now stops collecting at a side-effect boundary without clearing previously collected comparisons. So `x >= 0 && x >= 5 && refresh()` correctly reports `x >= 0` as redundant, while `x >= 0 && refresh() && x >= 5` correctly reports nothing. Added a regression test.

java-checks/src/main/java/org/sonar/java/checks/RedundantRangeCheckCheck.java
romainbrenguier: The current side-effect detection covers method calls, assignments, and increments/decrements. Extending it to array accesses, divisions, casts, and unboxing would require type resolution on every sub-expression in the operand to distinguish safe from potentially-throwing expressions. This level of analysis goes beyond the initial scope and risks over-approximation (flagging too many expressions as unsafe). I'd prefer to track this as a follow-up improvement if real false positives surface from the ruling tests.

PR #6142, opened 2026-09-14
SONARJAVA-6972 fix: clear safeSymbols to prevent memory leak

romainbrenguier:
## Problem

The `safeSymbols` field in `BoxedBooleanExpressionsCheck` accumulates symbols across file scans but is never cleared. This causes a memory leak when the analyzer processes multiple files, as the set continues to grow without bound.

The two static caches (`ifStatementCache` and `firstNullCheckCache`) are properly cleared at the start of each file scan, but `safeSymbols` was overlooked.

## Solution

Add `safeSymbols.clear()` in `scanFile()` alongside the existing cache clear calls. This ensures the set is reset for each file while maintaining correct behavior within a single file.

## Verification

- Existing tests pass
- The fix follows the same pattern used for the other caches in the same method


leveretka:
The code looks fine. Did you manage to reproduce the leak and test locally that it doesn't occur anymore?

You can run the analysis locally and connect to the process with VisualVM and get a heap dump before and after the fix. This way you'll be able to not just say that you fixed it but You could also estimate the impact on a real project. (You can try analyzing SonarJava itself at least)

And another question, don't you need an approval from someone from the CQ team? do you think mine would be enough here?

leveretka (approved):

romainbrenguier:
@leveretka 
On sonar-java at least, it doesn't seem to have an impact on memory used. Do you have an example project where you know the memory leak is visible?

Before the fix:
Screenshot from 2026-09-22 10-06-17



After the fix:
Screenshot from 2026-09-22 09-50-25

leveretka:
@romainbrenguier I'm not sure I can answer the question about the leak based on the charts solely. Of course the retained memory highly depends on the actual usage of `Optional.map` in the repo.

What I'd check is if 'safeSymbols' is growing till the end of analysis by generating heapdumps during the analysis. And see if it contains symbols from the old files and references old files ASTs. Of course at the end of the analysis everything is cleaned. 

Also, I believe it will more affect single huge modules.

Another possible issur is that the source files are still there due to the other leak. But that requires looking at the heapdump between the files.

romainbrenguier:
> @romainbrenguier I'm not sure I can answer the question about the leak based on the charts solely. Of course the retained memory highly depends on the actual usage of `Optional.map` in the repo.
> 
> What I'd check is if 'safeSymbols' is growing till the end of analysis by generating heapdumps during the analysis. And see if it contains symbols from the old files and references old files ASTs. Of course at the end of the analysis everything is cleaned.
> 
> Also, I believe it will more affect single huge modules.
> 
> Another possible issur is that the source files are still there due to the other leak. But that requires looking at the heapdump between the files.

Looking at the heap dump during the analizis of sonar-java, the `safeSymbols` map always have 0 elements, so it's not a good example to test the memory leak.

romainbrenguier:
I don't have a good example to reproduce the leak for now, but I'm going to merge anyway since it can only improve things.

PR #6144, opened 2026-09-15
SONARJAVA-6976 Clear check state after file analysis

romainbrenguier:
## Summary
- Clear collection-backed check state at file or module-analysis boundaries.
- Use finally blocks around per-file reporting where appropriate.
- Extend the ArchUnit lifecycle test to cover module-end cleanup.

## Validation
- CheckStateCleanupTest passes with zero violations.
- 102 focused java-checks tests pass.
- Compilation and whitespace checks pass.

## Agent workflow
Addressed review comments in pr_report_6144.md using `uv run address_reviews.py pr_report_6144.md`


java-checks/src/test/java/org/sonar/java/checks/MissingPackageInfoCheckTest.java
```diff
@@ -68,21 +67,15 @@ void initVerifier() {
...
         mainCodeSourcesPath("checks/packageInfo/package-info.java"),
         mainCodeSourcesPath("checks/packageInfo/nopackageinfo/HelloWorld.java"),
         mainCodeSourcesPath("checks/packageInfo/nopackageinfo/nopackageinfo.java"))
-      .withCheck(check)
+      .withCheck(new MissingPackageInfoCheck())
       .verifyIssueOnProject(EXPECTED_MESSAGE);
-
-    Set<String> set = check.missingPackageWithoutPackageFile;
-    assertThat(set).hasSize(1);
-    assertThat(set.iterator().next()).isEqualTo(EXPECTED_PACKAGE);
```
aurelien-coet-sonarsource, line 85: Why were these assertions removed here ?

java-frontend/src/main/java/org/sonar/plugins/java/api/IssuableSubscriptionVisitor.java
```diff
@@ -69,8 +79,7 @@ public final void scanFile(JavaFileScannerContext context) {
    */
   @Override
   public void leaveFile(JavaFileScannerContext context) {
-    // Explicitly declares the method to make it appears in the IssuableSubscriptionVisitor's members.
-    // Default behaviour is to do nothing
+    clearState();
   }
```
aurelien-coet-sonarsource, line 83: If a class overrides `leaveFile` and doesn't call `super.leaveFile(context)` or `clearState()`, then the state will not be cleared, even if `clearState` is overriden to do so. 
Furthermore, `leaveFile` is called inside a `try .. catch` block in `VisitorsBridge.scanFile`, after calling other methods that could throw, so if they do, then `leaveFile` is never called and the state is never cleared.
It may make more sense to directly call `clearState` on every visitor in the `finally` block inside `VisitorsBridge.scanFile`. WDYT ?

java-checks/src/main/java/org/sonar/java/checks/StaticFieldInitializationCheck.java
```diff
@@ -56,6 +56,9 @@ public List<Tree.Kind> nodesToVisit() {
 
   @Override
   public void setContext(JavaFileScannerContext context) {
+    classWithSynchronizedMethod.clear();
+    withinStaticInitializer.clear();
+    methodUsesLocks.clear();
```
aurelien-coet-sonarsource, line 61: This class is transitively an instance of `IssuableSubscriptionVisitor` , so it would probably make more sense to override `clearState` rather than to repeat the cleanup code in `setContext` and `leaveFile`.

java-checks/src/main/java/org/sonar/java/checks/security/CookieHttpOnlyCheck.java
```diff
@@ -156,14 +156,21 @@ public void setContext(JavaFileScannerContext context) {
...
+        reportIssue(mit.arguments(), MESSAGE);
+      }
+      for (TypeTree typeTree : newClassToReport) {
+        reportIssue(typeTree, MESSAGE);
+      }
+    } finally {
+      ignoredVariables.clear();
+      symbolConstructorMapToReport.clear();
+      settersToReport.clear();
+      newClassToReport.clear();
```
aurelien-coet-sonarsource: Same here, this is an `IssuableSubscriptionVisitor`, so it would make more sense to implement `clearState`.

java-checks/src/main/java/org/sonar/java/checks/security/SecureCookieCheck.java
```diff
@@ -160,12 +160,21 @@ public void setContext(JavaFileScannerContext context) {
...
+          reportIssue(mit.arguments(), MESSAGE);
+        }
+      });
+    } finally {
+      unsecuredCookies.clear();
+      deletionCandidateCookies.clear();
+      cookieConstructors.clear();
+      maxAgeZeroSeen.clear();
+      deletionCandidateSecureCalls.clear();
+      enclosingClass.clear();
```
aurelien-coet-sonarsource: Same as for `CookieHttpOnlyCheck`.

java-checks/src/main/java/org/sonar/java/checks/DateEnumsCheck.java
```diff
@@ -337,6 +337,7 @@ public void endOfAnalysis(ModuleScannerContext context) {
     }
     projectTotalUsageCount = 0;
     projectTotalNoEnumUsageCount = 0;
+    currentFileIssues.clear();
     issuesByFile.clear();
```
aurelien-coet-sonarsource, line 337: Here too, the class is transitively an `IssuableSubscriptionVisitor`.

java-checks/src/main/java/org/sonar/java/checks/HardcodedURICheck.java
```diff
@@ -129,6 +130,11 @@ public void leaveFile(JavaFileScannerContext context) {
         continue;
       }
       reportHardcodedURI(v.initializer());
+      }
+    } finally {
+      annotationsStack.clear();
+      identifiersUsedInAnnotations.clear();
+      hardCodedUri.clear();
```
aurelien-coet-sonarsource: This should also be in `clearState`.

java-checks/src/main/java/org/sonar/java/checks/WaitInWhileLoopCheck.java
```diff
@@ -37,6 +37,7 @@ public class WaitInWhileLoopCheck extends AbstractMethodDetection {
 
   @Override
   public void setContext(JavaFileScannerContext context) {
+    inWhileLoop.clear();
```
aurelien-coet-sonarsource, line 40: Also an `IssuableSubscriptionVisitor`.

aurelien-coet-sonarsource (changes requested):
Maybe we could also document somewhere (e.g. in `.claude/commands/project-level-rules.md`) that checks which use collections to store some state should override `clearState` to clear it between files ?

romainbrenguier:
> ## Ruling Diff Summary
> Detected changes in 5 rule files: 0 issues removed, 75 issues added.
> 
> S1166 (`java`) on commons-beanutils - 0 issues removed, 23 issues added...

This comment is outdated. The PR should not update ruling results.

aurelien-coet-sonarsource (changes requested):
Two comments are still unaddressed.

PR #6146, opened 2026-09-15
SONARJAVA-6971: Implemented rule S9395 Numeric values should be explicitly cast when assigned to narrower types

romainbrenguier:
## Summary
- Implements rule S9395 that detects implicit widening conversions from `int`/`long` to `float`/`double` where precision loss is possible
- Covers variable declarations, assignments, method/constructor arguments, return statements, and compound assignments
- Filters out explicit casts, safe small literals, and non-lossy conversions (e.g. `int` → `double`)

## Test plan
- [x] Unit tests pass with semantic analysis (all noncompliant/compliant patterns verified)
- [x] Unit tests pass without semantic analysis (no false positives when types are unresolved)
- [ ] Ruling tests (will be updated via auto-generated PR if needed)

## Agent workflow
PR created using `uv run new_rule_implementation.py S9395 -j SONARJAVA-6971 -a claude -g AGENTS.md AGENTS.md .claude/skills/new-rule/SKILL.md`

Addressed review comments in pr_report_6146.md using `uv run address_reviews.py pr_report_6146.md`


java-checks/src/main/java/org/sonar/java/checks/S9395Check.java
```diff
@@ -0,0 +1,178 @@
...
+  private static boolean isSafeConstant(ExpressionTree expr, Type targetType) {
+    Object constant = ExpressionUtils.resolveAsConstant(expr);
+    if (constant instanceof Integer intVal) {
+      long absValue = Math.abs((long) intVal);
+      return !targetType.isPrimitive(Type.Primitives.FLOAT) || absValue <= FLOAT_MAX_EXACT_INT;
+    }
+    if (constant instanceof Long longVal) {
+      long absValue = Math.abs(longVal);
+      if (targetType.isPrimitive(Type.Primitives.FLOAT)) {
+        return absValue <= FLOAT_MAX_EXACT_INT;
```
nathsou, line 165: This magnitude threshold produces false positives for exactly representable values above 2^24. For example,  is exactly representable as a , but this helper returns false and the rule reports it. Please determine actual representability (or otherwise avoid assuming every value beyond the threshold is lossy) before reporting.

java-checks/src/main/java/org/sonar/java/checks/S9395Check.java
```diff
@@ -0,0 +1,178 @@
...
+      }
+    }
+  }
+
+  private void checkExpression(Type targetType, @Nullable ExpressionTree expr) {
+    if (expr == null) {
+      return;
+    }
+    ExpressionTree unwrapped = ExpressionUtils.skipParentheses(expr);
+    if (unwrapped.is(Tree.Kind.TYPE_CAST)) {
```
nathsou, line 133: This suppresses every conversion whose operand is cast, including  assigned to . The cast changes the source type to , but the subsequent implicit -to- conversion can still lose precision. Please only suppress casts that eliminate the lossy conversion being checked.

java-checks/src/main/java/org/sonar/java/checks/S9395Check.java
```diff
@@ -0,0 +1,178 @@
...
+      context.reportIssue(this, unwrapped,
+        "Explicitly cast this \"" + sourceType.name() + "\" to \"" + targetType.name() + "\" to document potential precision loss.");
+    }
+  }
+
+  private static boolean isLossyWideningConversion(Type sourceType, Type targetType) {
+    if (sourceType.isUnknown() || targetType.isUnknown()) {
+      return false;
+    }
+    if (sourceType.isPrimitive(Type.Primitives.INT) && targetType.isPrimitive(Type.Primitives.FLOAT)) {
```
nathsou, line 147: Wrapper sources are not recognized here, so Java's implicit unboxing is missed. For example,  loses precision without an issue. Please account for / unboxing before evaluating the widening conversion.

nathsou (approved):
These are non-blocking findings for a follow-up: exact representability, cast handling, and boxed-value unboxing are worth tightening, but they do not block approval.

java-checks/src/main/java/org/sonar/java/checks/S9395Check.java
nathsou, line 165: This magnitude threshold produces false positives for exactly representable values above 2^24. For example, `20_000_000L` is exactly representable as a `float`, but this helper returns false and the rule reports it. Please determine actual representability (or otherwise avoid assuming every value beyond the threshold is lossy) before reporting.

java-checks/src/main/java/org/sonar/java/checks/S9395Check.java
nathsou, line 133: This suppresses every conversion whose operand is cast, including `(int) longValue` assigned to `float`. The cast changes the source type to `int`, but the subsequent implicit `int`-to-`float` conversion can still lose precision. Please only suppress casts that eliminate the lossy conversion being checked.

java-checks/src/main/java/org/sonar/java/checks/S9395Check.java
nathsou, line 147: Wrapper sources are not recognized here, so Java's implicit unboxing is missed. For example, `Integer id = 16_777_217; float value = id;` loses precision without an issue. Please account for `Integer`/`Long` unboxing before evaluating the widening conversion.

nathsou (commented):
Non-blocking follow-up findings: exact representability, cast handling, and boxed-value unboxing are worth tightening.

java-checks/src/main/java/org/sonar/java/checks/S9395Check.java
romainbrenguier: Agreed, the threshold-based approach is an approximation. Values like `20_000_000L` are exactly representable as float despite being > 2^24 (because their trailing zeros allow them to fit in 24 bits of mantissa). I'll track this as a follow-up improvement to check actual IEEE 754 representability instead of using a magnitude threshold.

java-checks/src/main/java/org/sonar/java/checks/S9395Check.java
romainbrenguier: Good catch. The current guard blanket-skips any cast expression, but a cast like `(int) longValue` doesn't eliminate the subsequent lossy `int -> float` widening. I'll refine this in a follow-up to only suppress casts where the resulting type eliminates the lossy conversion (e.g., `(float) longValue`).

java-checks/src/main/java/org/sonar/java/checks/S9395Check.java
romainbrenguier: Valid point. The rule currently only checks primitive source types, so boxed types like `Integer` and `Long` that undergo implicit unboxing before the widening conversion are missed. I'll add wrapper type handling in a follow-up.

PR #6150, opened 2026-09-15
SONARJAVA-6848: Implement S9400: Method references not fulfilling the "Comparator" contract should not be used as comparators

nathsou:
## Summary

- Implement rule S9400 in SonarJava: Method references not fulfilling the "Comparator" contract should not be used as comparators.
- Add `InvalidComparatorMethodReferenceCheck` targeting `Math::min`, `Math::max`, `StrictMath::min`, `StrictMath::max`, `Integer::min`, `Integer::max`, `Long::min`, `Long::max`, `Float::min`, `Float::max`, `Double::min`, `Double::max` when used as `java.util.Comparator`.
- Add test class and test samples covering semantic and non-semantic analysis.
- Add generated rule metadata and activate in the `Sonar way` profile.

## Links

- Jira: https://sonarsource.atlassian.net/browse/SONARJAVA-6848
- RSPEC PR: https://github.com/SonarSource/rspec/pull/8186

## AI disclosure
- LLM model used for implementation: gemini-3.8-flash-high


java-checks/src/main/java/org/sonar/java/checks/InvalidComparatorMethodReferenceCheck.java
```diff
@@ -0,0 +1,79 @@
...
+      expressionName = ((IdentifierTree) expression).name();
+    } else if (expression.is(Tree.Kind.MEMBER_SELECT)) {
+      expressionName = ((MemberSelectExpressionTree) expression).identifier().name();
+    } else if (expression instanceof ExpressionTree expressionTree) {
+      expressionName = expressionTree.symbolType().name();
+    } else {
+      expressionName = expression.toString();
+    }
+    return expressionName + "::" + methodReference.method().name();
+  }
```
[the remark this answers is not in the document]
nathsou: Addressed in 96a54aa43d. Simplified `methodReferenceName` to check only `MEMBER_SELECT` vs `IDENTIFIER`, eliminating the unreachable dead-code branches and resolving the missing test coverage.

java-checks-test-sources/default/src/main/java/checks/InvalidComparatorMethodReferenceCheckSample.java
```diff
@@ -0,0 +1,67 @@
...
+    list.sort(Integer::min); // Noncompliant {{Replace this method reference; "Integer::min" returns an extremum rather than a comparison result, violating the "Comparator" contract.}}
+//            ^^^^^^^^^^^^
+    list.sort(Integer::max); // Noncompliant {{Replace this method reference; "Integer::max" returns an extremum rather than a comparison result, violating the "Comparator" contract.}}
+//            ^^^^^^^^^^^^
+    list.sort(StrictMath::min); // Noncompliant {{Replace this method reference; "StrictMath::min" returns an extremum rather than a comparison result, violating the "Comparator" contract.}}
+//            ^^^^^^^^^^^^^^^
+    list.sort(StrictMath::max); // Noncompliant {{Replace this method reference; "StrictMath::max" returns an extremum rather than a comparison result, violating the "Comparator" contract.}}
+//            ^^^^^^^^^^^^^^^
+    list.sort(java.lang.Math::min); // Noncompliant {{Replace this method reference; "Math::min" returns an extremum rather than a comparison result, violating the "Comparator" contract.}}
+//            ^^^^^^^^^^^^^^^^^^^
```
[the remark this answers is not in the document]
nathsou: Addressed in 96a54aa43d. In Java, `Long.min/max`, `Float.min/max`, and `Double.min/max` return `long`, `float`, and `double` respectively. Because none of these return types can be implicitly converted to `int` (the return type of `Comparator.compare`), expressions like `longs.sort(Long::max)` fail javac compilation with incompatible types. They can never be passed as a `java.util.Comparator` in valid Java. Rather than keeping dead types in `EXTREMUM_METHODS`, we pruned `Long`, `Float`, and `Double` from the matcher to keep it strictly aligned with compiling `Comparator` method references (`Math`, `StrictMath`, `Integer`).

java-checks/src/main/java/org/sonar/java/checks/InvalidComparatorMethodReferenceCheck.java
```diff
@@ -0,0 +1,68 @@
...
+@Rule(key = "S9400")
+public class InvalidComparatorMethodReferenceCheck extends IssuableSubscriptionVisitor {
+
+  private static final String MESSAGE = "Replace this method reference; \"%s\" returns an extremum rather than a comparison result, violating the \"Comparator\" contract.";
+
+  private static final MethodMatchers EXTREMUM_METHODS = MethodMatchers.create()
+    .ofTypes(
+      "java.lang.Math",
+      "java.lang.StrictMath",
+      "java.lang.Integer"
```
romainbrenguier, line 38: we could add Long, Double, Float

java-checks/src/main/java/org/sonar/java/checks/InvalidComparatorMethodReferenceCheck.java
```diff
@@ -0,0 +1,68 @@
...
+    )
+    .names("min", "max")
```
romainbrenguier: we could add `sum`, `bitCount`, `substractExact`, `addExact`, but it's difficult to be exhaustive.

romainbrenguier (approved):
This seems a bit restricted because we only checking for a few methods we know to not satisfy the contract, but it would be difficult to be general.

java-checks/src/main/java/org/sonar/java/checks/InvalidComparatorMethodReferenceCheck.java
nathsou: In Java, `Long.min/max`, `Double.min/max`, and `Float.min/max` return `long`, `double`, and `float` respectively. Since `Comparator<T>` requires `int compare(T o1, T o2)` and narrowing primitive conversions are not permitted in method reference assignment (JLS §15.13.2), references like `Long::min` or `Double::max` fail compilation with incompatible return type errors (`long cannot be converted to int`). They can never be passed as a `Comparator` in compiling Java code.

The only two-argument static methods on `Long`, `Double`, and `Float` returning `int` are `compare` and `compareUnsigned` (which are valid comparators), so matching them here would be dead code.

java-checks/src/main/java/org/sonar/java/checks/InvalidComparatorMethodReferenceCheck.java
nathsou: Note that `bitCount` only takes 1 argument (`int bitCount(int)`), so it cannot be assigned to a `Comparator`.

However, `Integer::sum`, `Math::addExact`, `Math::subtractExact`, and `Math::multiplyExact` (and their `StrictMath` counterparts) all compile as `Comparator<Integer>` while violating the general comparator contract (either by symmetry or overflow exception).

We have updated the rule in ef44f091f6 to cover these methods, updated the issue message to `"Replace this method reference; \"%s\" does not return a comparison result, violating the \"Comparator\" contract."`, and refreshed the RSPEC documentation and plugin metadata accordingly.

java-checks/src/main/java/org/sonar/java/checks/InvalidComparatorMethodReferenceCheck.java
nathsou: In Java, `Long.min/max`, `Double.min/max`, and `Float.min/max` return `long`, `double`, and `float` respectively. Since `Comparator<T>` requires `int compare(T, T)`, and method reference assignment conversions do not permit narrowing primitive conversions to `int`, usages like `longs.sort(Long::min)` or `doubles.sort(Double::max)` fail `javac` compilation (`incompatible types: bad return type in method reference`). Therefore they cannot be passed as a `Comparator` in compiling Java code.

java-checks/src/main/java/org/sonar/java/checks/InvalidComparatorMethodReferenceCheck.java
nathsou: Great suggestion! Note that `bitCount` only takes 1 argument (`int bitCount(int)`), so it cannot be assigned to a `Comparator`.

However, `Integer::sum`, `Math::addExact`, `Math::subtractExact`, and `Math::multiplyExact` (and their `StrictMath` counterparts) all compile as `Comparator<Integer>` while violating the general comparator contract (either by symmetry or overflow exception).

We have updated the rule in ef44f091f6 to cover these methods, updated the issue message to `"Replace this method reference; \"%s\" does not return a comparison result, violating the \"Comparator\" contract."`, and refreshed the RSPEC documentation and plugin metadata accordingly.

java-checks/src/test/java/org/sonar/java/checks/InvalidComparatorMethodReferenceCheckTest.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @Test
+  void test_without_semantic() {
+    CheckVerifier.newVerifier()
+      .onFile(TEST_FILE)
+      .withCheck(new InvalidComparatorMethodReferenceCheck())
+      .withoutSemantic()
+      .verifyIssues();
+  }
```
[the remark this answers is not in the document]
nathsou: In SonarJava's test runner, ECJ still resolves types from the standard Java runtime library (`java.base`), meaning `java.lang.Math`, `java.lang.StrictMath`, `java.lang.Integer`, and `java.util.Comparator` resolve even when `.withoutSemantic()` is configured (which omits external dependencies from the classpath). Both `MethodMatchers.matches(...)` and `symbolType().isSubtypeOf("java.util.Comparator")` successfully evaluate, and running `InvalidComparatorMethodReferenceCheckTest#test_without_semantic` passes with 0 failures and verifies all expected issues. Therefore, `.verifyIssues()` is the correct assertion.

java-checks/src/main/java/org/sonar/java/checks/InvalidComparatorMethodReferenceCheck.java
```diff
@@ -0,0 +1,68 @@
...
+import org.sonar.plugins.java.api.semantic.MethodMatchers;
+import org.sonar.plugins.java.api.tree.IdentifierTree;
+import org.sonar.plugins.java.api.tree.MemberSelectExpressionTree;
+import org.sonar.plugins.java.api.tree.MethodReferenceTree;
+import org.sonar.plugins.java.api.tree.Tree;
+
+@Rule(key = "S9400")
+public class InvalidComparatorMethodReferenceCheck extends IssuableSubscriptionVisitor {
+
+  private static final String MESSAGE = "Replace this method reference; \"%s\" does not return a comparison result, violating the \"Comparator\" contract.";
```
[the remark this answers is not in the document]
nathsou: Addressed in 9dec29eee6. `subtractExact` computes `a - b` and is anti-symmetric, making it conceptually a subtraction-based comparison rather than an operation returning non-comparison values (like extrema or commutative operations). Subtraction-based comparison issues with numeric overflow are already covered by S9354. We have removed `subtractExact` from `INVALID_METHODS`, removed the sample test cases, and updated the RSPEC documentation and plugin metadata accordingly.

PR #6158, opened 2026-09-15
SONARJAVA-6967: Implemented rule S9396 - Inner class fields should not shadow outer class fields

romainbrenguier:
This PR implements rule S9396 which detects when an inner class declares a field with the same name as a field in its enclosing outer class.

**What the rule detects:**
- Inner class fields that shadow outer class fields
- Static nested classes with shadowing fields
- Anonymous inner classes with shadowing fields
- Records with shadowing fields
- Deeply nested classes (3+ levels)

**What the rule excludes:**
- Static fields in outer classes (cannot be shadowed)
- serialVersionUID fields (common serialization convention)
- Top-level classes (no enclosing class)

**Files changed:**
-  - Rule implementation
-  - Test cases
-  - Test class
-  - Fixed access modifier bug
- Metadata files (HTML, JSON, profile)

## Agent workflow

Iterated on the PR with `uv run ci_loop.py` for 6 iterations. 

Addressed review comments in pr_report_6158.md using `uv run address_reviews.py pr_report_6158.md`


romainbrenguier:
gitar unblock

java-checks/src/main/java/org/sonar/java/checks/InnerClassShadowFieldCheck.java
```diff
@@ -0,0 +1,109 @@
...
+        break;
+      }
+      current = current.enclosingClass();
+    }
+  }
+
+  private static boolean isExplicitlyStatic(Symbol.TypeSymbol typeSymbol) {
+    ClassTree declaration = typeSymbol.declaration();
+    return declaration != null
+      && declaration.is(Kind.CLASS, Kind.INTERFACE, Kind.ANNOTATION_TYPE)
```
nathsou, line 106: Nested enums and records are implicitly static, but this declaration-kind filter excludes them. For example, `class Outer { int value; enum Nested { ITEM; int value; } }` has no `Outer.this`, so `Nested.value` cannot shadow the outer instance field and would be a false positive. Please treat enum and record boundaries like other static nested types both when entering a class and when walking outward.

java-checks/src/main/java/org/sonar/java/checks/InnerClassShadowFieldCheck.java
```diff
@@ -0,0 +1,109 @@
...
+
+    if (classSymbol.outermostClass().equals(classSymbol)) {
+      return;
+    }
+
+    if (isExplicitlyStatic(classSymbol)) {
+      return;
+    }
+
+    Symbol.TypeSymbol enclosingClass = classSymbol.enclosingClass();
```
nathsou, line 69: Local and anonymous classes declared inside static methods or initializers are checked against outer instance fields that are not accessible. For example, `class Outer { int value; static void create() { class Local { int value; } } }` would be reported even though no `Outer.this` exists in that context. Please account for static methods and initializers when determining the enclosing-instance boundaries.

nathsou (approved):
The two inline comments identify 2 false-positive cases

PR #6159, opened 2026-09-15
SONARJAVA-6970: Implemented rule S9397 "String.format()" should not be used when simpler alternatives exist

romainbrenguier:
Detect simple String.format() calls that can be replaced with String.valueOf() or string concatenation.

## Agent workflow

PR created using `uv run new_rule_implementation.py S9397 -j SONARJAVA-6970 -a codex -g AGENTS.md .claude/skills/new-rule/SKILL.md`

Iterated on the PR with `uv run ci_loop.py` for 5 iterations. 

Addressed review comments in pr_report_6159.md using `uv run address_reviews.py pr_report_6159.md`


java-checks/src/main/java/org/sonar/java/checks/StringFormatCheck.java
```diff
@@ -0,0 +1,100 @@
...
+    if (placeholders <= 0 || invocation.arguments().size() != formatIndex + placeholders + 1) {
+      return;
+    }
+    List<ExpressionTree> valueArguments = invocation.arguments().subList(formatIndex + 1, invocation.arguments().size());
+    for (ExpressionTree arg : valueArguments) {
+      if (arg.symbolType().isArray()) {
+        return;
+      }
+    }
+    if (hasLocale && valueArguments.stream().anyMatch(arg -> arg.symbolType().isSubtypeOf("java.util.Formattable"))) {
```
nathsou: Both String.format(String, Object...) and String.format(Locale, String, Object...) use Formattable.formatTo() for Formattable arguments. This guard only skips the locale overload, so a call such as String.format("%s", formattable) is reported even though replacing it with String.valueOf() or + changes the output to toString(). Please apply the Formattable exemption to both overloads and update the corresponding sample.

java-checks/src/main/java/org/sonar/java/checks/StringFormatCheck.java
```diff
@@ -0,0 +1,100 @@
...
+    int formatIndex = hasLocale ? 1 : 0;
+    if (invocation.arguments().size() <= formatIndex) {
+      return;
+    }
+    ExpressionTree formatArgument = invocation.arguments().get(formatIndex);
+    if (!formatArgument.is(Tree.Kind.STRING_LITERAL)) {
+      return;
+    }
+    LiteralTree literal = (LiteralTree) formatArgument;
+    int placeholders = countSimplePlaceholders(LiteralUtils.trimQuotes(literal.value()));
```
nathsou: LiteralUtils.trimQuotes(literal.value()) still leaves Java escape sequences in the source token. For example, String.format("%s / \\0451$04d", 7) decodes at runtime to a format containing %1$04d, but this parser sees only the simple %s and reports it. Please parse the decoded literal value or conservatively skip literals with unsupported escapes.

nathsou (approved):
Approved with two correctness findings to address: Formattable handling and escaped format-string parsing.

romainbrenguier:
Both review comments addressed in e523cce:

**Formattable handling (line 72):** Removed the `hasLocale &&` guard so the Formattable exemption now applies to both `String.format(String, Object...)` and `String.format(Locale, String, Object...)` overloads.

**Escape sequences (line 62):** The rule now conservatively skips any format string literal containing backslash escape sequences, avoiding false positives from octal/unicode escapes that could decode to `%` at runtime.

java-checks/src/main/java/org/sonar/java/checks/StringFormatCheck.java
romainbrenguier: Fixed: removed the `hasLocale &&` guard so the Formattable exemption now applies to both overloads.

java-checks/src/main/java/org/sonar/java/checks/StringFormatCheck.java
romainbrenguier: Fixed: the rule now conservatively skips any format string literal containing backslash escape sequences, avoiding false positives from octal/unicode escapes that could decode to `%` at runtime.

PR #6161, opened 2026-09-15
SONARJAVA-6969: Implement S9399 Instance fields should not be guarded by static locks

romainbrenguier:
## Summary
- Implements rule S9399 that detects `synchronized` blocks using a static lock to guard instance (non-static) fields
- Static locks cause all instances to compete for the same lock even though they access distinct data, providing no real protection and causing unnecessary contention
- Reports an issue on the lock expression with secondary locations on the accessed instance fields

## Test plan
- [x] Unit tests with semantic analysis pass
- [x] Unit tests without semantic analysis pass
- [ ] CI build passes
- [ ] Ruling test expectations updated if needed

## Agent workflow
PR created using `uv run new_rule_implementation.py S9399 -j SONARJAVA-6969 -a claude -g AGENTS.md .claude/skills/new-rule/SKILL.md`

Iterated on the PR with `uv run ci_loop.py` for 4 iterations. 

Addressed review comments in pr_report_6161.md using `uv run address_reviews.py pr_report_6161.md`


sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9399.html
```diff
@@ -0,0 +1,56 @@
+<p>This is an issue when instance-level member data is protected by a class-level lock or synchronized using a shared field that is common to all
+instances of the class.</p>
+<h2>Why is this an issue?</h2>
+<p>Locks in multi-threaded programming control access to shared data. When a lock protects a field, both the lock and the field must have the same
+scope.</p>
+<p>A class-level lock (also called a static lock) is shared across all instances of a class. It should only guard class-level fields (static fields)
+that are shared across all instances.</p>
+<p>An instance-level lock belongs to a specific object instance. It should guard instance fields that are specific to that object.</p>
+<p>When you use a class-level lock to guard an instance field, the lock provides no real protection. Different object instances will compete for the
```
nathsou, line 9: Request changes: this documentation overstates the defect. A consistently used static monitor does provide mutual exclusion and happens-before visibility for instance fields; the demonstrated problem is unnecessary contention, not that the lock provides no protection or inherently causes races. Also, replacing a static lock with an instance lock is unsafe for a block that accesses both instance and static state. Please align the RSPEC and generated documentation with the actual synchronization semantics and define the intended scope of the rule.

java-checks/src/main/java/org/sonar/java/checks/synchronization/S9399Check.java
```diff
@@ -0,0 +1,128 @@
...
+    private final List<IdentifierTree> instanceFieldAccesses = new ArrayList<>();
+
+    InstanceFieldAccessVisitor(Symbol ownerType) {
+      this.ownerType = ownerType;
+    }
+
+    @Override
+    public void visitIdentifier(IdentifierTree tree) {
+      Symbol symbol = tree.symbol();
+      if (!symbol.isUnknown() && symbol.isVariableSymbol() && symbol.owner().isTypeSymbol() && !symbol.isStatic()
```
nathsou, line 96: Request changes: `this` and `super` are represented by known, non-static variable symbols in this semantic model, so they satisfy this predicate. For example, `synchronized (LOCK) { consume(this); }` and `synchronized (LOCK) { super.toString(); }` are reported even though neither directly accesses an instance field. Exclude these synthetic receiver symbols from the collected field accesses.

java-checks/src/main/java/org/sonar/java/checks/synchronization/S9399Check.java
```diff
@@ -0,0 +1,128 @@
...
+    if (!isStaticLock(sst.expression())) {
+      return;
+    }
+    Tree enclosingClassTree = getEnclosingTree(sst, Tree.Kind.CLASS, Tree.Kind.ENUM, Tree.Kind.RECORD);
+    if (enclosingClassTree == null) {
+      return;
+    }
+    Symbol enclosingClassSymbol = ((ClassTree) enclosingClassTree).symbol();
+    InstanceFieldAccessVisitor visitor = new InstanceFieldAccessVisitor(enclosingClassSymbol);
+    sst.block().accept(visitor);
```
nathsou, line 58: Request changes: this traversal enters deferred executable bodies. A block such as `synchronized (LOCK) { return () -> count++; }` will report `count`, although the lambda runs after the synchronized statement has released the lock; local and anonymous class method bodies have the same issue. Stop traversal into lambda bodies and nested class methods while preserving direct accesses evaluated by the synchronized block.

java-checks/src/main/java/org/sonar/java/checks/synchronization/S9399Check.java
```diff
@@ -0,0 +1,128 @@
...
+      if (!symbol.isUnknown() && symbol.isVariableSymbol() && symbol.owner().isTypeSymbol() && !symbol.isStatic()
+        && isOwnedBySameOrEnclosingClass(symbol)) {
+        instanceFieldAccesses.add(tree);
+      }
+    }
+
+    @Override
+    public void visitMemberSelectExpression(MemberSelectExpressionTree tree) {
+      ExpressionTree expression = tree.expression();
+      if (isThis(expression)) {
```
nathsou, line 105: Request changes: qualified outer-instance accesses are missed. Inside an inner class, `synchronized (LOCK) { Outer.this.count++; }` is not reported because `isThis` only recognizes a bare `this`, and the fallback scans only the qualifier. Handle qualified `this`/`super` receivers so fields accessed through an explicit outer instance are analyzed.

nathsou (changes requested):
Request changes: the implementation still has false-positive and false-negative paths, and the rule documentation overstates the concurrency risk.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S9399.html
romainbrenguier: I will fix on the rspec side.

java-checks/src/main/java/org/sonar/java/checks/synchronization/S9399Check.java
romainbrenguier: Fixed: added `isThisOrSuper` check in `visitIdentifier` to exclude `this` and `super` identifiers from field access detection.

java-checks/src/main/java/org/sonar/java/checks/synchronization/S9399Check.java
romainbrenguier: Fixed: overrode `visitLambdaExpression` to stop traversal, and overrode `visitNewClass` to only scan constructor arguments (not the anonymous class body).

java-checks/src/main/java/org/sonar/java/checks/synchronization/S9399Check.java
romainbrenguier: Fixed: added `isQualifiedThis` helper that recognizes `Outer.this` and `Outer.super` member select expressions, and use it alongside `isThis` in `visitMemberSelectExpression`.

nathsou (approved):

PR #6174, opened 2026-09-16
SONARJAVA-6897 Implement parsing for Spring profile expressions

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **New Spring Profile Parser:**
    - Added `ProfileExpression` interface and record types to model `@Profile` conditions
    - Added `ProfileExpressionParser` for parsing Spring profile expressions with support for conjunctions, disjunctions, and negation


java-frontend/src/main/java/org/sonar/java/model/springcontext/ProfileExpression.java
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  /**
+   * Conjunction, as declared by {@code &} or by composing a class-level with a method-level {@code @Profile}.
+   *
+   * @param operands The expressions that must all match.
+   * @return {@link #UNKNOWN} if any operand is unknown; otherwise the conjunction of the operands, with
+   * {@link #UNCONDITIONAL} operands dropped and nested conjunctions flattened.
+   */
+  static ProfileExpression and(List<ProfileExpression> operands) {
```
asya-vorobeva: So we have associativity of & and |. Probably it makes sense to provide also commutativity (by switching to Set from List).

asya-vorobeva (changes requested):

asya-vorobeva (approved):
💯

PR #6176, opened 2026-09-16
SONARJAVA-7003: Implement S3037: Constructors should not access uninitialized values

nathsou:
## Summary
- Implement S3037: Constructors should not access uninitialized values in SonarJava.
- In record compact constructors, an explicit reference to a generated component accessor invokes the accessor method before the implicit component field assignments happen. The accessor returns default uninitialized field values rather than constructor parameters.
- Detect method invocations on `this` targeting record component accessors inside compact constructors and report an issue.

## Links
- Jira: https://sonarsource.atlassian.net/browse/SONARJAVA-7003
- RSPEC: Existing rule S3037 (https://rules.sonarsource.com/java/RSPEC-3037/)

## AI disclosure
- LLM model used for implementation: gpt-5.6-sol-low


java-checks/src/test/java/org/sonar/java/checks/ConstructorsShouldNotAccessUninitializedValuesCheckTest.java
```diff
@@ -0,0 +1,41 @@
...
+  @Test
+  void test() {
+    CheckVerifier.newVerifier()
+      .onFile(TestUtils.mainCodeSourcesPath("checks/ConstructorsShouldNotAccessUninitializedValuesCheckSample.java"))
+      .withCheck(new ConstructorsShouldNotAccessUninitializedValuesCheck())
+      .verifyIssues();
+  }
+
+  @Test
+  void test_without_semantic() {
```
utafrali, line 34: This test will fail. The rule's entire detection path requires semantic analysis — `methodSymbol.isUnknown()` returns `true` when semantic is absent, causing the guard on line 108 of the check to short-circuit and report nothing. But `.verifyIssues()` reads the `// Noncompliant` markers from the sample file and asserts that every annotated line has an issue, so the verifier will fail because no issues are emitted.

Change to `.verifyNoIssues()` to assert the rule produces no false positives when the semantic model is unavailable:
```java
@Test
void test_without_semantic() {
  CheckVerifier.newVerifier()
    .onFile(TestUtils.mainCodeSourcesPath("checks/ConstructorsShouldNotAccessUninitializedValuesCheckSample.java"))
    .withCheck(new ConstructorsShouldNotAccessUninitializedValuesCheck())
    .withoutSemantic()
    .verifyNoIssues();
}
```

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.html
```diff
@@ -0,0 +1,19 @@
+<h2>Why is this an issue?</h2>
```
utafrali, line 1: The documentation shows an abstract class with uninitialized fields, but the implementation exclusively targets record compact constructors. Someone reading the rule description will expect it to flag regular class constructors too, which it does not.

Update the HTML to describe the actual behaviour — record compact constructors run before the implicit component field assignments, so calling a synthetic accessor inside one reads a default/uninitialized value. A concrete noncompliant example with a record and compliant alternatives (direct parameter access or an explicit canonical constructor) would match what the check actually detects.

java-checks/src/main/java/org/sonar/java/checks/ConstructorsShouldNotAccessUninitializedValuesCheck.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Override
+    public void visitMethodInvocation(MethodInvocationTree tree) {
+      if (isInvocationOnThis(tree)) {
+        Symbol methodSymbol = tree.methodSymbol();
+        if (!methodSymbol.isUnknown()
+          && methodSymbol.isMethodSymbol()
+          && recordSymbol.equals(methodSymbol.owner())
+          && tree.arguments().isEmpty()
+          && ((Symbol.MethodSymbol) methodSymbol).parameterTypes().isEmpty()
```
utafrali: The `parameterTypes().isEmpty()` check on line 112 is redundant given `tree.arguments().isEmpty()` on line 111. Record component accessors are always zero-argument methods, and by the time the name is matched against `componentNames` (line 113) you already know the call site has no arguments. Removing line 112 keeps the condition set minimal and easier to follow:
```java
&& tree.arguments().isEmpty()
&& componentNames.contains(methodSymbol.name())
&& ((Symbol.MethodSymbol) methodSymbol).declaration() == null
```

java-checks-test-sources/default/src/main/java/checks/ConstructorsShouldNotAccessUninitializedValuesCheckSample.java
```diff
@@ -0,0 +1,134 @@
...
+    CompliantDirectAccess {
+      if (name.isBlank() || age < 0) { // Compliant
+        throw new IllegalArgumentException();
+      }
+    }
+  }
+
+  record CanonicalConstructor(String name) {
+    CanonicalConstructor(String name) {
+      if (name().isBlank()) { // Compliant, not a compact constructor
```
utafrali, line 60: The comment says "Compliant, not a compact constructor", but `name()` is invoked here *before* `this.name = name` runs, so at the time of the call `this.name` is still `null`. This is exactly the class of bug S3037 describes.

The current scope limitation (compact constructors only) means this case is intentionally skipped, which is a reasonable initial scope. Worth adding a TODO or a note in the ticket so the canonical-constructor case is tracked for a follow-up, and updating the comment to be explicit about the scope choice rather than implying the call is safe.

utafrali (commented):
The detection logic for compact constructors is correct and the test coverage is thorough, but the `test_without_semantic` method calls `.verifyIssues()` on a rule that emits nothing without semantic, which will cause a CI test failure. The HTML documentation also needs to be updated to describe records rather than abstract classes.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.html
romainbrenguier, line 1: There should be a paragraph to say when the rule raises. The rspec is also missing a How to fix section

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.html
```diff
@@ -0,0 +1,19 @@
...
+
+  String name;
+  String fname;
+  int hashCode;
+
+  public abstract String getValue();
+
+  public MyAbstractClass(String name) {
+    this.fname = this.name.split()[0];  // Noncompliant; this.name not assigned yet
+    this.hashCode = getValue().hashCode();  // Noncompliant; child class constructor hasn't run yet
```
romainbrenguier: The tests only check for the record case but the example is not a record. It is also missing a compliant example.

romainbrenguier (changes requested):
The RSPEC is very general and vague but the implementation seems to only check compact constructors of record classes, which is very narrow. The two need to be aligned.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.html
nathsou: Updated in commit 6dd3f43aec. Added an explanation of when the rule raises (invoking a record component accessor inside a compact constructor) and added a dedicated 'How to fix it' section recommending direct parameter access.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.html
nathsou: Updated in commit 6dd3f43aec. Replaced the abstract class example with noncompliant and compliant record examples matching the implementation and tests.

romainbrenguier (changes requested):
There should be a corresponding PR on the RSPEC repo

java-checks/src/main/java/org/sonar/java/checks/ConstructorsShouldNotAccessUninitializedValuesCheck.java
```diff
@@ -0,0 +1,150 @@
...
+          && componentNames.contains(methodSymbol.name())
+          && ((Symbol.MethodSymbol) methodSymbol).declaration() == null) {
```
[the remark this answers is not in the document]
nathsou: Agreed and fixed in a39683a589. Explicit accessors read the same unassigned field (I checked: an accessor returning `List.copyOf(items)` throws a `NullPointerException` when called from the compact constructor), so the `declaration() == null` filter is gone and explicit accessors are reported. The `ExplicitAccessor` sample case is now Noncompliant.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.json
```diff
@@ -0,0 +1,15 @@
...
+    "func": "Constant\/Issue",
+    "constantCost": "30min"
+  },
+  "tags": [],
+  "defaultSeverity": "Major",
+  "ruleSpecification": "RSPEC-3037",
+  "sqKey": "S3037",
+  "scope": "Main",
+  "quickfix": "unknown"
+}
```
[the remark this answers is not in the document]
nathsou: Fixed in a39683a589. The missing attribute came from the RSPEC metadata, so I added `"code": {"impacts": {"RELIABILITY": "MEDIUM"}, "attribute": "LOGICAL"}` (plus the `java16` tag used by the other record rules) in SonarSource/rspec#8229 and regenerated `S3037.json` with rule-api from that branch.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.html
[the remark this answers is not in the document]
nathsou: Fixed in a39683a589. The description is rewritten in SonarSource/rspec#8229 to match the implemented scope (record component accessors called before the fields are assigned), with compiling noncompliant/compliant examples, a "How to fix it" section and JLS resources. `S3037.html` is regenerated from that branch with rule-api.

java-checks/src/main/java/org/sonar/java/checks/ConstructorsShouldNotAccessUninitializedValuesCheck.java
```diff
@@ -0,0 +1,146 @@
...
+        if (ExpressionUtils.isThis(receiver)) {
+          return true;
+        }
+        if (receiver.is(Tree.Kind.MEMBER_SELECT)) {
+          MemberSelectExpressionTree memberSelect = (MemberSelectExpressionTree) receiver;
+          if (ExpressionUtils.isThis(memberSelect.identifier())) {
+            Symbol thisSymbol = memberSelect.identifier().symbol();
+            return !thisSymbol.isUnknown() && recordSymbol.equals(thisSymbol.enclosingClass());
+          }
+        }
```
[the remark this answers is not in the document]
nathsou: Fixed in a39683a589. As you pointed out, records are implicitly static, so a qualified `X.this` inside a record constructor always denotes the record. The enclosing-class comparison is removed and any `this` / `X.this` receiver is treated as the record instance.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.html
nathsou: Follow-up: my earlier edit (6dd3f43aec) changed the generated HTML directly, which was the wrong place. The description, including the "when the rule raises" paragraph and the "How to fix it" section, now lives in SonarSource/rspec#8229, and `S3037.html` is regenerated from it with rule-api in a39683a589.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.html
nathsou: Follow-up: the record examples (compact and explicit canonical constructor, each with a compliant solution) are now in SonarSource/rspec#8229, and `S3037.html` is regenerated from it in a39683a589.

nathsou:
@romainbrenguier thanks for the review. Addressing your two points:

**RSPEC PR:** https://github.com/SonarSource/rspec/pull/8229. The plugin HTML/JSON are now generated from it with rule-api (a39683a589), replacing my earlier hand edit.

**Aligning the spec and the implementation:** I narrowed the RSPEC to records instead of widening the check to every class. The new title is "Record component accessors should not be called before the fields are assigned". The rule now reports a component accessor (generated or explicitly declared) called on the record being constructed:
- anywhere in a compact constructor, because the fields are assigned after its body runs;
- in an explicit canonical constructor, before the `this.component = ...` assignment.

The implementation was widened slightly to cover explicit accessors and explicit canonical constructors.

Why not the general "any constructor reading an uninitialized value" scope from the old description:
- Calls to overridable methods from constructors, which the old example showed, are already covered by S1699.
- Reading a non-final field before assignment in an ordinary class compiles, but detecting it needs flow analysis across helper methods and super constructors. Fields initialized by `init()` helpers, or by overridden methods called from a super constructor, would make it noisy.
- Reading a final field before assignment is already a compile error.

In records, the accessor call is the only way to read a component field before it is assigned. javac rejects the direct read, and the accessor silently returns `null`/`0`/`false`. So the narrow scope is always a bug and cheap to detect: symbols plus source order, since a final field cannot be assigned in a loop. The full rationale is in the RSPEC PR description.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.html
```diff
@@ -0,0 +1,81 @@
...
+<h2>Why is this an issue?</h2>
+<p>The fields of a record are assigned only once, by its canonical constructor:</p>
+<ul>
+  <li>In a compact constructor, the fields are assigned implicitly, from the constructor parameters, <strong>after</strong> the body of the
+  constructor has run.</li>
+  <li>In an explicit canonical constructor, each field is assigned by a <code>this.component = …​</code> statement of the constructor body.</li>
+</ul>
+<p>Until a field is assigned, it holds the default value of its type: <code>null</code>, <code>0</code>, or <code>false</code>. Reading such a field
+by its simple name or through <code>this.</code> is rejected by the compiler, but calling the component accessor, for example <code>name()</code> or
+<code>this.name()</code>, compiles and silently reads the field.</p>
```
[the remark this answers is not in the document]
nathsou: Right, inside the constructor the simple name is the parameter. Fixed in SonarSource/rspec#8229 (only `this.` reads are mentioned now) and regenerated `S3037.html` with rule-api in c15a2eaef1.

java-checks/src/main/java/org/sonar/java/checks/ConstructorsShouldNotAccessUninitializedValuesCheck.java
```diff
@@ -0,0 +1,191 @@
...
+    public void visitMethodInvocation(MethodInvocationTree tree) {
+      Symbol methodSymbol = tree.methodSymbol();
+      String name = methodSymbol.name();
+      if (tree.arguments().isEmpty()
+        && componentNames.contains(name)
+        && recordSymbol.equals(methodSymbol.owner())
+        && isInvocationOnRecordInstance(tree)
+        && isBeforeFieldAssignment(name, tree)) {
+        reportIssue(ExpressionUtils.methodName(tree),
+          String.format("Replace this call to \"%s()\" with the \"%s\" parameter; the field is not assigned yet.", name, name));
```
romainbrenguier: I think you should replace String.format with a String concatenation. We should have a rule for that 😅

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.json
```diff
@@ -0,0 +1,23 @@
...
+  "code": {
+    "impacts": {
+      "RELIABILITY": "MEDIUM"
+    },
+    "attribute": "LOGICAL"
+  },
+  "status": "ready",
+  "remediation": {
+    "func": "Constant\/Issue",
+    "constantCost": "30min"
```
romainbrenguier: It seems a bit long for just replacing an accessor with a parameter.

java-checks/src/main/java/org/sonar/java/checks/ConstructorsShouldNotAccessUninitializedValuesCheck.java
```diff
@@ -0,0 +1,191 @@
...
+      }
+
+      @Override
+      public void visitAssignmentExpression(AssignmentExpressionTree tree) {
+        ExpressionTree variable = ExpressionUtils.skipParentheses(tree.variable());
+        if (variable.is(Tree.Kind.MEMBER_SELECT)) {
+          MemberSelectExpressionTree memberSelect = (MemberSelectExpressionTree) variable;
+          String name = memberSelect.identifier().name();
+          if (ExpressionUtils.isThis(memberSelect.expression()) && componentNames.contains(name)) {
+            assignmentEnds.putIfAbsent(name, Position.endOf(tree));
```
romainbrenguier, line 120: 🔧 [request]  `putIfAbsent` stores the first textual assignment, so an accessor in another branch can be missed. For example, if one if branch assigns this.name before calling name(), while the other branch calls name() before assigning it, the
  latter call is considered after the recorded position even though it still reads the uninitialized field. Assignment dominance or path-sensitive analysis is needed.

romainbrenguier (approved):

java-checks/src/main/java/org/sonar/java/checks/ConstructorsShouldNotAccessUninitializedValuesCheck.java
nathsou: Nice, new rule idea!

nathsou: Done in c847167fe1, the message is now built by concatenation. For the record, both styles are common in `java-checks` (about 140 checks use `String.format`, about 100 use concatenation), so this is a readability choice rather than a convention fix.

sonar-java-plugin/src/main/resources/org/sonar/l10n/java/rules/java/S3037.json
nathsou: Agreed, a one-token fix doesn't deserve 30 minutes. The value came from the rule's original generic description. I lowered `constantCost` to `5min` (the most common value for Java BUG rules) in SonarSource/rspec#8229 and regenerated `S3037.json` in c847167fe1.

java-checks/src/main/java/org/sonar/java/checks/ConstructorsShouldNotAccessUninitializedValuesCheck.java
nathsou: Good catch, the false negative is real: in your scenario `name()` returns `null` in the `else` branch and nothing is reported. The same happens for a call in a `catch` block after a `try` that assigns the field.

I don't think fixing it is worth it for now:
- **Cheap fixes introduce false positives.** Using the last assignment instead of the first flags calls placed after a real assignment in the other branch. Requiring the assignment's branch to enclose the call flags the common case where both branches assign before the call (`CanonicalBranches` in the sample), and `if (...) { this.name = ...; } else { throw ...; }`.
- **A correct fix needs a definite-assignment analysis** over the constructor CFG (`MethodTree.cfg()`): a fixpoint over the blocks, exceptional edges, and new tests for branches, `catch`/`finally`, `switch` and `?:`. That's roughly 40 to 70 more lines for a very rare case.
- **The benefit is close to zero.** Only explicit canonical constructors are affected, which are the minority (S6210 pushes towards compact constructors), and only when a field is assigned in several branches and its accessor is called in a branch before the assignment. Compact constructors, the main target, are not affected, and the approximation never produces false positives.

I documented the gap with a `// FN` case in the sample (c847167fe1). Happy to open a follow-up ticket for the CFG-based analysis if you think it's worth it.

PR #6183, opened 2026-09-17
SONARJAVA-7006 Record telemetry for Spring related properties

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **Telemetry:**
    - Added telemetry metrics for Spring properties including `java.spring.bean_count`, `java.spring.injection_point_count`, and model size in bytes
    - Implemented `SizeEstimator` to track memory footprint of `SpringContextModel` and its components


java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextModelMetricsCollector.java
```diff
@@ -0,0 +1,213 @@
...
+  private static final int ARRAY_HEADER = 16;
+  private static final int REFERENCE = 4;
+  private static final int INT = 4;
+  private static final int ALIGNMENT = 8;
+
+  private static final float HASH_TABLE_LOAD_FACTOR = 0.75f;
+  private static final int HASH_TABLE_MIN_CAPACITY = 16;
+  private static final int HASH_TABLE_MAX_CAPACITY = 1 << 30;
+
+  private final Set<Object> visited = Collections.newSetFromMap(new IdentityHashMap<>());
```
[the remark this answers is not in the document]
aurelien-coet-sonarsource: This is not a problem in practice: it only records references for each encountered object, and an empty model contains none, so it costs nothing then.

sonar-java-plugin/src/main/java/org/sonar/plugins/java/ProjectEndOfAnalysisSensor.java
```diff
@@ -23,7 +23,10 @@
 import org.sonar.api.batch.sensor.SensorDescriptor;
 import org.sonar.api.scanner.sensor.ProjectSensor;
 import org.sonar.java.jsp.Jasper;
+import org.sonar.java.model.springcontext.SpringContextModel;
+import org.sonar.java.model.springcontext.SpringContextModelMetrics;
 import org.sonar.java.telemetry.Telemetry;
+import org.sonar.java.telemetry.TelemetryKey;
 
 /**
  * Sensor that runs at the end of the project's analysis to send telemetry data.
```
asya-vorobeva, line 33: I think we should explicitly enforce ordering between this and SpringContextModelSensor to make sure that here we have SpringContextModel collected. See [here](https://github.com/SonarSource/sonar-plugin-api/blob/master/docs/influencing-extensions-execution-order.md?plain=1#L41) how to do it.

asya-vorobeva (changes requested):
Another thing that I forgot to mention. It would be great also to calculate and send the total time used for gathering Spring Context + executing SpringContextChecks. It shouldn't be a big deal (say we can do it in SpringContextModelGatherer + SpringContextModelSensor).

asya-vorobeva (changes requested):
Latest Gitar comment looks pretty reasonable. Let's not count time for Spring model gathering as it's nothing comparing to the main part (SpringContextModelGatherer's itself are considered as usual File visitors).

its/scanner-integration-tests/src/test/java/org/sonar/java/it/spring/SpringTelemetryTest.java
```diff
@@ -0,0 +1,46 @@
...
+  void spring_context_model_is_reported_as_telemetry() {
+    var project = analyzeProject(Path.of("ambiguous-dependencies-should-be-resolved"), "S9352").getProject();
+
+    assertThat(telemetryValue(project, "java.spring.bean_count")).isEqualTo("20");
+    assertThat(telemetryValue(project, "java.spring.bean_name_count")).isEqualTo("20");
+    assertThat(telemetryValue(project, "java.spring.injection_point_count")).isEqualTo("6");
+    assertThat(telemetryValue(project, "java.spring.component_scan_package_count")).isEqualTo("1");
+    assertThat(telemetryValue(project, "java.spring.context_model_gathering_time_ms")).matches("\\d+");
+    assertThat(telemetryValue(project, "java.spring.context_checks_time_ms")).matches("\\d+");
+    assertThat(Long.parseLong(telemetryValue(project, "java.spring.context_model_size_bytes"))).isPositive();
```
asya-vorobeva: isPositive looks too weak. I understand that we don't want to use isEqual, but let's compare with some minimal threshold.

java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextModelMetricsCollector.java
```diff
@@ -0,0 +1,255 @@
...
+    long estimatedSizeInBytes = collector.walk(model);
+    return new SpringContextModelMetrics(
+      collector.beanCount,
+      collector.beanNameCount,
+      collector.injectionPointCount,
+      collector.componentScanPackageCount,
+      estimatedSizeInBytes);
+  }
+
+  private long walk(SpringContextModel model) {
```
asya-vorobeva: I think about delegating size  calculations to individual sub-objects. We can create common interface for all model's parts (including SpringContextModel itself) with the method estimateSize and implement calculations there. Then we avoid providing mutable objects to the outer world and creating massive God class. WDYT?

asya-vorobeva (changes requested):

java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextModelGatherer.java
```diff
@@ -49,10 +77,81 @@ public boolean isCompatibleWithDependencies(Function<String, Optional<Version>>
...
+   * Gathers data from a node of one of the kinds returned by {@link #nodesToVisit()}.
+   *
+   * @param tree The node to visit.
+   */
+  protected void visitSpringNode(Tree tree) {
+    // by default, gatherers collect nothing from the nodes they visit
+  }
+
+  @Override
+  public final void leaveFile(JavaFileScannerContext context) {
```
asya-vorobeva, line 100: We use different names for signatures' parameters in this and child methods (context ones). Let's unify them

java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextModelMetricsCollector.java
aurelien-coet-sonarsource: Yes, I think it makes sense and would probably be cleaner. I especially dislike that we're exposing the mutable backing containers of the individual objects, so this should help us avoid doing this. I'll change that.

java-frontend/src/main/java/org/sonar/java/telemetry/SizeEstimator.java
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * Estimates object sizes for a 64-bit HotSpot JVM with compressed object pointers and 8-byte object alignment. Strings
+ * are assumed to use the Latin-1 coder. Immutable collections are approximated with the {@link java.util.HashMap}
+ * formulas used for the mutable indexes.
+ *
+ * <p>A single instance must be shared by all parts of an object graph so that shared instances are charged once. The
+ * result is a deterministic indicator suitable for comparing projects and tracking growth over time, not an exact
+ * retained-size measurement.
+ */
+public final class SizeEstimator {
```
asya-vorobeva, line 35: Argh, we're really struggling for the tests for this class :)

asya-vorobeva (changes requested):

java-frontend/src/main/java/org/sonar/java/model/springcontext/SpringContextModelMetrics.java
```diff
@@ -0,0 +1,58 @@
...
+  public static SpringContextModelMetrics of(SpringContextModel model) {
+    var beanDefinitions = model.getBeanDefinitionRegistry();
+    var dependencies = model.getTypeToDependenciesIndex();
+    var packageScan = model.getProjectPackageScan();
+    return new SpringContextModelMetrics(
+      beanDefinitions.beanCount(),
+      beanDefinitions.beanNameCount(),
+      dependencies.injectionPointCount(),
+      packageScan.packageCount(),
+      SizeEstimator.estimate(
```
asya-vorobeva: Maybe I miss something, but can't we also make SpringContextModel itself implementing SizeEstimable and simply call estimateSize of it?

its/scanner-integration-tests/src/test/java/org/sonar/java/it/spring/SpringTelemetryTest.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+class SpringTelemetryTest extends ScannerIntegrationAbstractTest {
+
+  private static final int EXPECTED_NAMED_BEAN_COUNT = 20;
+  private static final long HASH_MAP_NODE_SIZE_BYTES = 32L;
+  private static final long SINGLE_ELEMENT_LIST_SIZE_BYTES = 48L;
+  private static final long BEAN_DEFINITION_HOLDER_SHALLOW_SIZE_BYTES = 40L;
+  private static final long MINIMUM_NON_EMPTY_STRING_SIZE_BYTES = 48L;
+  private static final long MINIMUM_EXPECTED_CONTEXT_MODEL_SIZE_BYTES =
```
asya-vorobeva, line 75: I don't get why we do use this formula to estimate minimum size. Probably it comes from estimating below BeanDefinitionRegistry, but what about other components of the model? I guess we can be much more precise.

Another point is to estimate also from above. It could be rather illustrative (taking into account that we can have large enterprise monoliths with thousands (!) of beans defined and it would be good to see some benchmarking and not afraid that we become too heavy).

asya-vorobeva (changes requested):

asya-vorobeva (approved):
Generally 👍. Please don't forget to add explaining comment for  MINIMUM_EXPECTED_CONTEXT_MODEL_SIZE_BYTES field.

PR #6178, opened 2026-09-16
SONARJAVA-6897 Use the Spring profile expression parser in Spring context collection

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **Spring context collection:**
    - Refactored Spring context model to use `ProfileExpression` instead of raw strings for profile conditions
    - Updated `BeanDefinitionHolder` and related classes to support expression-aware profile parsing and serialization

<sub>This will update automatically on new commits.</sub>

PR #6184, opened 2026-09-17
SONARJAVA-6916 S9352: restrict ambiguous-bean candidate set to consumer's Spring context

asya-vorobeva:
We want restrict the candidate set to beans reachable from the consumer's own context (component-scan packages for its module, plus its dependencies) before deciding ambiguity.

I've made changes in a model:
- InjectionPoint gains a module field (set at gather time from
  context.getModuleKey(); not serialized to cache)
- TypeToBeanNamesIndex stores (name, module, beanPackage) per entry and
  exposes getNamesForType(type, consumerModule, scannedPackages) which
  returns only beans visible from the consumer's Spring context.

As a result AmbiguousDependencyCheck.execute()  is simplified and uses the information about dependencies' modules and component-scan packages for its module to properly decide about ambiguity.


aurelien-coet-sonarsource (approved):
LGTM ! 👍 Let's further improve `AmbiguousDependencyCheck` with profiles sensitivity in [this PR](https://github.com/SonarSource/sonar-java/pull/6178) once [that one](https://github.com/SonarSource/sonar-java/pull/6174) has been merged.

PR #6192, opened 2026-09-17
SONARJAVA-7012 Remove S9404 from Sonar way profile

nathsou:
S9404 currently produces false positives in the default Java Sonar way profile. Remove its profile marker while keeping the rule implementation and metadata available for explicit activation.

Companion RSPEC metadata PR: https://github.com/SonarSource/rspec/pull/8209

Verification:
- `mvn -pl sonar-java-plugin -am -Dtest=JavaSonarWayProfileTest -Dsurefire.failIfNoSpecifiedTests=false -DskipITs test`
- `mvn -pl sonar-java-plugin -am -Dtest=MetadataTest -Dsurefire.failIfNoSpecifiedTests=false -DskipITs test`
- Generated Sonar way profile contains 590 rules and no S9404.


romainbrenguier (changes requested):
Shouldn't we fix FPs instead of removing the rule? If it is not possible explain why (in PR or Jira).

romainbrenguier (approved):

PR #6197, opened 2026-09-18
SONARJAVA-7004: Implement S3949: Calculations should not overflow

nathsou:
## Summary

- detect high-confidence `int` and `long` overflow with exact and bounded range evaluation
- detect unsafe midpoint formulas while avoiding known-safe constants
- suppress overlap with S2184, S9354, and S2676
- add semantic and without-semantic tests with precise issue highlighting

## Links

- [SONARJAVA-7004](https://sonarsource.atlassian.net/browse/SONARJAVA-7004)
- [RSPEC PR](https://github.com/SonarSource/rspec/pull/8206)

## Validation

- `mvn -pl java-checks -am test -Dtest=S3949CheckTest -Dsurefire.failIfNoSpecifiedTests=false`
- generated checklist, metadata, rules-definition, and Sonar way profile tests
- overlap scan: no high-priority overlap remains

## AI disclosure

LLM model used for implementation: GPT-5.6 Sol

[SONARJAVA-7004]: https://sonarsource.atlassian.net/browse/SONARJAVA-7004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


its/ruling/src/test/resources/commons-beanutils/java-S3949.json
romainbrenguier, line 1: Just to note, the Ruling Diff comment is outdated

java-checks/src/test/java/org/sonar/java/checks/helpers/BoundedIntegerRangeTest.java
romainbrenguier, line 1: 🔧 [request] This test is deleted but it should be replaced  with a test for IntegerOverflowRange

java-checks/src/main/java/org/sonar/java/checks/IntegerOverflowCheck.java
```diff
@@ -0,0 +1,250 @@
...
+    Integer denominator = ExpressionUtils.skipParentheses(division.rightOperand()).asConstant(Integer.class).orElse(null);
+    Long longDenominator = ExpressionUtils.skipParentheses(division.rightOperand()).asConstant(Long.class).orElse(null);
+    Double floatingDenominator = LiteralUtils.doubleLiteralValue(ExpressionUtils.skipParentheses(division.rightOperand()));
+    if (!Integer.valueOf(2).equals(denominator) && !Long.valueOf(2L).equals(longDenominator)
+      && !Double.valueOf(2.0).equals(floatingDenominator)) {
+      return false;
+    }
+    Range left = IntegerOverflowRange.rangeOf(addition.leftOperand());
+    Range right = IntegerOverflowRange.rangeOf(addition.rightOperand());
+    return (left == null || right == null) && !isExact(left) && !isExact(right);
```
romainbrenguier, line 143: 🔧 **[request]**  (value + 1) / 2 can overflow when value == Integer.MAX_VALUE, but the condition returns false whenever either operand has an exact range. The sample explicitly marks ceilingHalf = (low + 1) / 2 compliant, so this real overflow is
  missed.

java-checks/src/main/java/org/sonar/java/checks/S3949Check.java
```diff
@@ -0,0 +1,250 @@
...
+import org.sonar.plugins.java.api.tree.MethodTree;
+import org.sonar.plugins.java.api.tree.NewClassTree;
+import org.sonar.plugins.java.api.tree.ReturnStatementTree;
+import org.sonar.plugins.java.api.tree.Tree;
+import org.sonar.plugins.java.api.tree.TypeCastTree;
+import org.sonar.plugins.java.api.tree.UnaryExpressionTree;
+import org.sonar.plugins.java.api.tree.VariableTree;
+
+@Rule(key = "S3949")
+public class S3949Check extends IssuableSubscriptionVisitor {
```
romainbrenguier: 🔧 **[request]** Give a more descriptive name to the class.

romainbrenguier (changes requested):

its/ruling/src/test/resources/commons-beanutils/java-S3949.json
nathsou: Good catch. That summary was posted on Sep 18, when the rule still reported bounded ranges that *could* overflow (182 issues). Since then, S3949 only reports calculations that always overflow, and it now raises 0 issues in every ruling project. The bot left the comment unchanged because it only picks up result files for rules that still raise issues, so it did not notice that the S3949 expectations became empty. I updated them to `{}` by hand in dbe24defe8, and Ruling QA is green. I also marked the old summary as outdated.

java-checks/src/test/java/org/sonar/java/checks/helpers/BoundedIntegerRangeTest.java
nathsou: Done in 1c9888af05. I added `IntegerOverflowRangeTest`, which covers the private constructor and `intSubtractionCannotOverflow`: it returns true for `long` constants and false when either operand is unknown. It also checks that exact overflows wrap like at runtime (`Integer.MAX_VALUE + 1` becomes `MIN_VALUE`, `Long.MIN_VALUE - 1L` becomes `Long.MAX_VALUE`), that a bounded range which may overflow becomes unknown (`s.length() + 1`) while one that fits keeps its bounds (`s.length() - 1`), and the difference between `exceeds` and `alwaysExceeds` for both `int` and `long`.

java-checks/src/main/java/org/sonar/java/checks/IntegerOverflowCheck.java
nathsou: You are right that `(value + 1) / 2` overflows when `value == Integer.MAX_VALUE`. Skipping it is deliberate, for three reasons:

- It is the same "overflows for one extreme input" case as a plain `value + 1`. After the last round, S3949 only reports calculations that always overflow; reporting "can overflow" cases on sizes and lengths produced all 182 ruling issues, and they were all false positives.
- `(n + 1) / 2` is usually ceiling division, e.g. `(size + 1) / 2`, not a midpoint.
- The midpoint check stays for `(low + high) / 2` when both bounds are unbounded, which is the well-known binary-search bug. Reporting `(n + 1) / 2` while not reporting `n + 1` would be inconsistent.

If you would still like `(n + 1) / 2` reported, I can add it, but it would bring back some of that noise. Let me know.

java-checks/src/main/java/org/sonar/java/checks/S3949Check.java
nathsou: Done in 1c9888af05. I renamed it to `IntegerOverflowCheck`, together with `IntegerOverflowCheckTest` and `IntegerOverflowCheckSample`. The name matches the `IntegerOverflowRange` helper it uses. The rule key is still S3949, and the plugin registration and metadata tests pass.

romainbrenguier (approved):

PR #6199, opened 2026-09-18
SONARJAVA-7016 Make S9352 profile expression-aware

aurelien-coet-sonarsource:
----
## Summary by Gitar

- **Spring check update:**
    - Updated `AmbiguousDependencyCheck` to evaluate candidates under combinations of profile expressions
    - Added comprehensive unit tests for profile-aware dependency disambiguation in `AmbiguousDependencyCheckTest`


java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -41,18 +41,19 @@
 @Rule(key = "S9352")
 public class AmbiguousDependencyCheck implements JavaCheck, SpringContextCheck {
 
+  private static final int MAX_ENUMERATED_PROFILE_NAMES = 12;
```
asya-vorobeva, line 44: A natural question arises there (I guess you expect it): why this number? In practice, maintainability breaks down if an application uses more than 5–10 active profiles simultaneously. So let's add at least some explaining comment to this field.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -66,44 +67,82 @@ public List<SpringContextIssue> execute(SpringContextModel model) {
...
+    if (profileNames.size() > MAX_ENUMERATED_PROFILE_NAMES) {
+      return Set.of();
+    }
+    Set<String> activeProfiles = new HashSet<>();
+    do {
+      Set<String> activeCandidates = candidates.stream()
+        .filter(candidate -> isActive(holdersByCandidate.get(candidate), activeProfiles))
+        .collect(Collectors.toUnmodifiableSet());
+      if (!hasUniqueOrPrimaryCandidate(activeCandidates, holdersByCandidate, activeProfiles)) {
+        return activeCandidates;
```
asya-vorobeva, line 105: Actually we shouldn't raise issues for all possible values of spring.profiles.active. The most typical situation that can happen in real app:

```
@Configuration
public class DataSourceConfig {

    @Bean
    @Profile("dev")
    public DataSource h2DataSource() {
        return new EmbeddedDatabaseBuilder().build();
    }

    @Bean
    @Profile("prod")
    public DataSource postgresDataSource() {
        return new PostgresDataSource();
    }
}
```

It's an absolutely valid case, and practically our application will be running with dev OR prod, not both. But we raise an issue.
As we don't have enough information to judge what profiles can be active, all becomes rather simple. To not produce FNs we should check only for subsets of size one for our profile set!

Another thing: if we find some profile for each the issue happens, it will be useful to note this profile in issue's message.

asya-vorobeva (changes requested):

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
aurelien-coet-sonarsource: Yes, I expected it 😅 The number is completely arbitrary. It is "high" to not miss too many cases.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
aurelien-coet-sonarsource: Ah, I actually considered only inspecting single profiles and not their combinations, but I feared this would miss too many cases. 
You're right though, we probably risk many FPs with the current approach. It will greatly simplify things to only consider single profiles, and this way we can probably also get rid of the hardcoded limit on the number of profiles considered, since there won't be any more risk of the state space blowing up. I'll make the change.

asya-vorobeva: In real world people almost never use multiple active profiles. I've never seen it. :)

java-checks/src/test/java/org/sonar/java/checks/spring/AmbiguousDependencyCheckTest.java
```diff
@@ -95,28 +96,28 @@ void one_resolved_injection_point_does_not_hide_another_ambiguous_one_of_the_sam
...
+    assertThat(check.execute(model)).hasSize(1);
   }
 
   @Test
-  void excluding_profiled_candidate_still_leaves_ambiguity_between_the_rest() {
+  void profiled_candidate_is_considered_with_unprofiled_candidates() {
     SpringContextModel model = buildModel(
       "PlainEventPublisherComponentA.java", "PlainEventPublisherComponentB.java", "ProfiledEventPublisherComponent.java", "EventPublisherConsumer.java");
     assertThat(check.execute(model)).hasSize(1);
   }
```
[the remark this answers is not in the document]
aurelien-coet-sonarsource: This was resolved by [8225d85](https://github.com/SonarSource/sonar-java/pull/6199/commits/8225d85a2776e74ddcd34189ad4c3126a96ee80c).

java-checks/src/test/java/org/sonar/java/checks/spring/AmbiguousDependencyCheckTest.java
```diff
@@ -154,9 +171,98 @@ void candidate_profiled_with_an_operator_expression_is_excluded_like_a_simply_pr
...
+    SpringContextModel model =
+      buildModel("SameProfileComponentA.java", "SameProfileComponentB.java", "SameProfileConsumer.java");
+    assertThat(check.execute(model))
+      .singleElement()
+      .extracting(SpringContextIssue::message)
+      .asString()
+      .contains("sameProfileComponentA", "sameProfileComponentB");
+  }
+
+  @Test
```
asya-vorobeva, line 190: The tests which explicitly construct model (via helping methods register...) were added initially mostly to cover cross-module case which is hard to cover using regular ones (via buildModel method). I'm sure that majority of such tests you add here are redundant (cover the same scenarios as previous regular tests do). Please inspect and reduce.

asya-vorobeva (changes requested):

java-checks/src/test/java/org/sonar/java/checks/spring/AmbiguousDependencyCheckTest.java
aurelien-coet-sonarsource: Most the tests check specific configurations that the others don't, but yes, it probably makes more sense to use `buildModel` here, since we don't need the cross-module aspect.

asya-vorobeva (approved):
💯

PR #6205, opened 2026-09-21
SONARJAVA-6999: Fix FP in S9148 when subtraction is used with Float.compare/Double.compare

romainbrenguier:
## Summary
- Fix false positive in S9148 where floating-point subtraction used to compute derived values (e.g. `this.a - this.b`) was flagged even when the result was properly compared using `Float.compare` or `Double.compare`
- When a comparison method body already contains a call to `Float.compare`/`Double.compare`, skip flagging subtraction operations (the developer is already using proper comparison)
- Relational operators (`<`, `>`, `<=`, `>=`) are still flagged regardless, as they always represent direct floating-point comparisons

## Test plan
- [x] Added compliant test cases for subtraction with `Float.compare` and `Double.compare`
- [x] Added compliant test case for lambda comparator with `Double.compare`
- [x] Added mixed test case: subtraction compliant but relational operator still flagged when both are in same method
- [x] Existing tests pass unchanged
- [ ] CI passes

## Agent workflow

Iterated on the PR with `uv run ci_loop.py` for 1 iterations. 


nathsou (changes requested):
The fix covers the ticket's case, but the suppression is narrow: `addNestedSubtractions` only walks through parentheses, casts and chains of `-`. As soon as the subtraction is nested in another operator or a method call, it is flagged again, even when the result flows into `Double.compare`/`Float.compare`. These are the same kind of FP as SONARJAVA-6999 (a derived sort key compared correctly), and I verified that each of them still raises issues on this branch:

```java
return Double.compare(Math.hypot(x - 1, y - 1), Math.hypot(o.x - 1, o.y - 1)); // 4 issues
double d1 = Math.abs(x - y); /* ... */ return Double.compare(d1, d2);          // flagged
return Float.compare(a - b + c, o.a - o.b + o.c);                               // 2 issues
double d1 = (a - b) * (a - b); /* ... */ return Double.compare(d1, d2);         // 2 issues
```

Distance-based and squared-delta comparators are common, so I expect these to show up in the same cluster.

Suggested options:
- **Minimal:** suppress any floating-point subtraction anywhere inside a compare argument (or inside the initializer of a tracked variable), not only along `-` chains.
- **More robust:** invert the logic and only flag a subtraction whose value becomes the method's int result (cast to `int`/`long`, returned, passed to `Math.signum`/`Math.round`, or assigned to an `int` variable). That matches what the RSPEC calls "subtraction-based comparison" and removes the whole family of FPs instead of allowlisting patterns.

Please also add test cases for these patterns.

romainbrenguier:
@nathsou Thanks for the review. I've addressed the feedback by adopting the "minimal" approach you suggested: instead of only walking minus-chains in `addNestedSubtractions`, the new `collectAllSubtractions` method walks the entire expression subtree of each `Float.compare`/`Double.compare` argument (and variable initializers that flow into them). This suppresses any floating-point subtraction nested anywhere inside a compare argument — whether wrapped in `Math.hypot`, `Math.abs`, addition, multiplication, or any other expression.

Added test cases for:
- `Math.hypot(x - 1, y - 1)` inside `Double.compare`
- `Math.abs(x - y)` stored in variable passed to `Double.compare`
- `a - b + c` inside `Float.compare`
- `(a - b) * (a - b)` stored in variable passed to `Double.compare`

I didn't go with the "more robust" inverted-logic approach (only flag subtractions whose value becomes the int return) as it would be a larger refactor and the minimal approach covers the FP cluster you identified.

java-checks/src/main/java/org/sonar/java/checks/FloatingPointComparisonCheck.java
```diff
@@ -38,27 +51,119 @@ public List<Tree.Kind> nodesToVisit() {
...
+
+    private void collectAllSubtractions(Tree tree) {
+      tree.accept(new SubtractionCollector());
+    }
+
+    private class SubtractionCollector extends ComparisonMethodUtils.SkipNestedTypesVisitor {
+      @Override
+      public void visitBinaryExpression(BinaryExpressionTree tree) {
+        if (tree.is(Tree.Kind.MINUS)) {
+          suppressedSubtractions.add(tree);
```
nathsou, line 151: Suppressing every subtraction inside a compare argument also suppresses `return Double.compare(x - other.x, 0)`. This still violates the comparator contract: comparing NaN against 1 and reversing the operands both return 1. A focused CheckVerifier probe reports the subtraction on the base revision but reports nothing on this PR. Distinguish comparison by subtraction from independently computed values before suppressing it.

java-checks/src/main/java/org/sonar/java/checks/FloatingPointComparisonCheck.java
```diff
@@ -38,27 +51,119 @@ public List<Tree.Kind> nodesToVisit() {
...
+    }
+
+    private static boolean allUsagesAreCompareArguments(Symbol symbol) {
+      for (IdentifierTree usage : symbol.usages()) {
+        Tree parent = usage.parent();
+        if (parent == null) {
+          return false;
+        }
+        if (parent.is(Tree.Kind.ARGUMENTS)) {
+          Tree grandParent = parent.parent();
```
nathsou: The usage check requires the identifier’s immediate parent to be `ARGUMENTS`. Consequently, `double a = x - y; double b = other.x - other.y; return Double.compare((a), (b));` still raises two false positives. Casting the arguments to `double` has the same result. Although argument collection unwraps parentheses and casts, this guard rejects those usages first. Walk through those wrappers here and add corresponding local-variable tests.

nathsou (approved):
Approved. I added two inline findings for follow-up: subtraction against a constant is still suppressed, and parenthesized/cast local-variable usages are not recognized as compare arguments.

PR #6211, opened 2026-09-22
SONARJAVA-6927 Fix FP for S9352: handle @Qualifier on the bean definition

asya-vorobeva:
When a @Bean method (or stereotype class) carries a @Qualifier whose value
differs from the bean name, a consumer that qualifies its injection point by
that value was incorrectly flagged as ambiguous. The rule now also checks the
@Qualifier value on each candidate bean when determining whether an injection
point is disambiguated by name.


java-frontend/src/main/java/org/sonar/java/serialization/BeanDefinitionHolderTypeAdapter.java
```diff
@@ -87,6 +89,7 @@ public BeanDefinitionHolder.InputFileData read(JsonReader in) throws IOException
     Boolean isPrimary = null;
     String profiles = null;
     boolean profilesRead = false;
+    String qualifier = null;
```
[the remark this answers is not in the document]
asya-vorobeva: This property is often legitimately absent (most beans have no @Qualifier), so making it required would invalidate all cached entries unnecessarily, just to fix a FP on a rare pattern. The optional approach — qualifier defaults to null when absent — is the right tradeoff. A file with a qualifying bean will get the fix when it's next re-parsed; that's acceptable.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
```diff
@@ -71,14 +71,27 @@ public List<SpringContextIssue> execute(SpringContextModel model) {
           continue;
         }
         Set<String> effectiveCandidates = excludeCandidatesWithProfile(candidates, registry);
-        if (!hasUniqueOrPrimaryCandidate(effectiveCandidates, registry) && !candidates.contains(point.name())) {
+        if (!hasUniqueOrPrimaryCandidate(effectiveCandidates, registry) && !matchesByNameOrQualifier(candidates, point.name(), registry)) {
```
aurelien-coet-sonarsource, line 74: Here we use the injection point's name when checking if there is a matching qualifier. However, if we have for example:
```java
@Bean
@Qualifier("someName")
public Foo firstFoo() {
    return new Foo();
}

@Bean
public Foo secondFoo() {
    return new Foo();
}

@Autowired
private Foo someName;
```
then Spring won't disambiguate the bean by matching the injection point's name with a qualifier annotation. We would need to have:
```java
@Autowired
@Qualifier("someName")
private Foo someName;
```
for it to work as expected.
We probably need to differentiate between an injection point's name and its qualifier annotation to properly handle this situation.

aurelien-coet-sonarsource (changes requested):
I think we need to update the model to account for qualifier annotations and injection points names separately.

java-checks/src/main/java/org/sonar/java/checks/spring/AmbiguousDependencyCheck.java
asya-vorobeva: I've checked the logic at first manually, then with the test. No problems, all works as expected. When it tries to find needed candidate it uses getBeanQualifier and properly finds someName. Why do you think it won't work?

aurelien-coet-sonarsource: The issue with the current implementation is that it misses some ambiguities and leads to false negatives. 
In the example from my previous comment, there is an ambiguity between the two beans of type `Foo`, because Spring won't actually match the injection point's name against `@Qualifier("someName")` on `firstFoo`. In practice this ambiguity will only be resolved if`@Qualifier("someName")` is also added to the `someName` field. Spring matches qualifiers on injection points against bean qualifiers or names, but injection point names only against bean names.
To catch these ambiguities, I think we would need to change the model to keep track of injection points names and qualifiers separately.

asya-vorobeva: The problem is that Spring uses @Qualifier annotation for two different things. In one case (major I'd say) it's indeed applied on autowired dependency, and helps to resolve ambiguity looking for a beans' names. But this case is different. It's applied on bean's definition and assigns a metadata label to the bean for dependency resolution. So it's used when you want to decouple the bean's lookup tag from its structural name.
So there is no FNs, works as expected.

aurelien-coet-sonarsource (approved):
LGTM 👍

